PR #3, opened 2019-01-11
Update openssl to 1.0.2q + link fix

milipili:
Just a small update to latest 1.0.x branch of openssl, as 1.1.x has API breaking changes (at least under Debian 10 testing)


sgatto:
Thank you @milipili.
We changed the build chain and we no longer use submodules, we updated the code tu use the latest API of OpenSSL, Curl, WxWidgets.
We close the issue.

PR #8, opened 2020-03-19
unit tests : modifying matrix + cmake tree

guilpier-code:
- First, we modify class matrix so that it can be unit-tested, at least saveToCSVFile
  method
- then we add a cmake tree to generate build tools for unit tests


guilpier-code:
PR not relevant any more. Replaced with another PR.

PR #51, opened 2020-09-23
poc for end to end test

jmkerloch:
Add end to end test as a proof of concept.

Very simple study are created and simulated using antares librairies.

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -2,7 +2,6 @@ cmake_minimum_required(VERSION 3.10)
 
 project(unit_tests_antares VERSION 1.0)
 
-
 # Finds boost.test or builds boost.test from github 
 add_subdirectory(ext)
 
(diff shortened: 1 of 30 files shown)


src/tests/CMakeLists.txt
```diff
@@ -2,11 +2,10 @@ cmake_minimum_required(VERSION 3.10)
 
 project(unit_tests_antares VERSION 1.0)
 
-# Defining TESTING preprocessor variable at compile time
-# This variable will soon be useless and droped. 
-add_definitions(-DTESTING)
```
jmkerloch, line 7: TESTING variable is now defined with `target_compile_definitions` only for tests that need the variable

src/tests/end-to-end/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+#bigobj support needed for windows compilation
+if(MSVC)
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
```
guilpier-code, line 3: Maybe use of 
`set_property(TARGET tests-simple-study PROPERTY /bigobj)`
or something like that would be more specific

jmkerloch: I used the option defined in `solver/CMakeList.txt`. This seems to be the way of adding bigobj support in Windows.

PR #66, opened 2020-10-14
Fix/58 bc marked skipped

guilpier-code:
- In panel "Binding constraint > Summary", the bullet related to a BC (binding constraint) must turn red in the following situations : 
   - if all links or clusters weights are nil
   - if the only non nil weights are clusters weights and these clusters are disabled or must-run
- Still in panel "Binding constraint > Summary", if a BC involves disabled or must-run clusters, they are marked as N/A by adding "x N/A" to their corresponding term in the BC.
- In the panels "Binding constraint > Weights > Thermal clusters" and  "Binding constraint > Offsets > Thermal clusters", disabled or must-run clusters are highlighted by "disabled" or "must-run" in the corresponding cells. These cells content cannot be changed unless user jumps to "Themal" tab and makes these clusters regular.
- In the previous panels, as well as in their "... > Links" counterparts, a BC should be marked as "Skipped" (on line "On/Off") if :
   - all links or clusters weights are nil
   - the only non nil weights are clusters weights and these clusters are disabled or must-run

### DIFF
--- a/src/libs/antares/study/constraint/constraint.cpp
+++ b/src/libs/antares/study/constraint/constraint.cpp
@@ -1243,9 +1243,7 @@ namespace Data
 				s << " + ";
 			SNPRINTF(tmp, sizeof(tmp), "%.2f", i->second);
 
-			s << '(' << (const char*)tmp << " x "
-				<< (i->first)->getName();
-			//<< (i->first)->from->name << '.' << (i->first)->with->name;
+			s << '(' << (const char*)tmp << " x " << (i->first)->getName();
 
 			auto at = pLinkOffsets.find(i->first);
 			if(at != pLinkOffsets.end())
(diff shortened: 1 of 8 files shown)


jmkerloch (changes requested):
Il y a une regression concernant le passage de Skipped à Yes.

Dans la version 7.2 on observe le comportement suivant:  
- suppression de la dernière contrainte renseignée => passage en Skipped
- ajout d'une nouvelle contrainte => passage en Yes

Dans la version de cette PR : 
- suppression de la dernière contrainte renseignée => passage en Skipped
- ajout d'une nouvelle contrainte => **maintien en Skipped + impossible de passer en Yes**

jmkerloch:
Erreur lors des tests. Il s'agisait pas de l'édition des poids mais des offsets.
Il est normal que les offsets n'aient pas d'effet sur l'activation des contraintes.

jmkerloch (changes requested):
Problème de rafraichissement du summary lors de la modification d'un cluster (passage en disabled par exemple).
L'affichage n'indique pas une éventuelle désactivation et l'indication de NA pour le cluster.

jmkerloch (approved):

PR #83, opened 2020-11-25
Feature/define deps install dir

jmkerloch:
- OR-Tools and sirius solver are not compiled by default (`BUILD_ortools=OFF` and `BUILD_sirius=OFF)`
- Add option to define dependency install dir `DEPS_INSTALL_DIR` default is **outside** current checkout directory so it can be used in another antares checkout directory.
- Update `INSTALL.md` with `DEPS_INSTALL_DIR` cmake option
- Update GitHub action CI to remove dependency build dir (needed in case of all dependency build because of available hard drive space available in GitHub action runners) and add OR-Tools and sirius solver compilation
- Remove not needed release.yml GitHub action (separated with 2 versions in develop : ubuntu and windows)
- Dependency compilation is now done in build directory


src/cmake/dependencies/CMakeLists.txt
```diff
@@ -63,21 +63,21 @@ if(WIN32 OR WIN64)
...
+	file(MAKE_DIRECTORY "${DEPS_INSTALL_DIR}/include/curl")
 	
 	#TODO : define VC version, at the moment it is hard coded with VC=14
 	#TODO : see for better WITH_PREFIX use
 
 	#Remarks : WITH_PREFIX compile option must use \ (backslash). Need to add multiple backslash because two backslashes (\\1) are required in CMake code to get a backslash through argument parsing
 	#We need 4 backslash because the BUILD_COMMAND option is passed to build_git_dependency then to another function inside:
-
-	set(BUILD_CALL "cd <SOURCE_DIR>/winbuild COMMAND nmake Makefile.vc mode=static VC=14 WITH_SSL=static WITH_DEVEL=${CMAKE_CURRENT_BINARY_DIR}/install/ WITH_PREFIX=..\\\\..\\\\..\\\\install ENABLE_SSPI=no ENABLE_IDN=no ENABLE_WINSSL=no GEN_PDB=no DEBUG=no MACHINE=x64")
+    string(REPLACE "/" "\\\\" DEPS_INSTALL_DIR_BACKSLASH ${DEPS_INSTALL_DIR})
```
flomnes, line 73: no need to escape / ?

jmkerloch: / must be replaced with 4 \

PR #103, opened 2021-01-11
Update label when using reverse and selection button and correction for status changes with numeric values

jmkerloch:


flomnes (approved):
Get feedback from @MDoquet and merge

src/ui/simulator/windows/options/playlist/playlist.cpp
```diff
@@ -207,6 +207,10 @@ namespace Options
 		Freeze();
```
guilpier-code, line 207: They are 2 things that does not seem right :
- suppose you have an activated year with a weight != 1, like 3 for instance. If you disable this year, everything stays ok, the year keeps its weight 3. Now activating the year again turn its weight to 1.
- suppose you have an activated year with a weight != 1, like 3 for instance. If you disable this year, the message next to the check box (above years list) keeps takings that year into account : the year still belong to the set of custom years. Why not ? Close the playlist window and reopen it, you'll see that this year is not anymore taken into account.

guilpier-code (changes requested):

src/ui/simulator/windows/options/playlist/playlist.cpp
jmkerloch: These behavior occur only when changing Status value with 0 for Skip and 1 for Active.

If you change value by typing Skip or Active the weight is not reset and the label is refreshed.

To Be Fixed.

jmkerloch: A break was missing in a switch statement. Numeric values for status column where also used for MC year weight.

flomnes (approved):

guilpier-code (approved):

PR #105, opened 2021-01-12
Change MC years weight to float instead of int

jmkerloch:


src/ui/simulator/windows/options/playlist/playlist.cpp
```diff
@@ -294,7 +294,7 @@ namespace Options
 				{
                     ++y;
 
-                    int weight = yearWeight[i];
+					float weight = yearWeight[i];
```
flomnes, line 297: indentation

src/libs/antares/study/parameters.cpp
```diff
@@ -1262,6 +1262,12 @@ namespace Data
                     result += yearsWeight[i];
                 }
             }
+
+			//Check if value is 0.0 then return 1.0 to avoid division by 0
+			if (result == 0.0)
```
flomnes, line 1267: indentation inconsistent with l. 1265

flomnes (changes requested):
- I suggest using `0.f` and `1.f` for float values
- Fix indentation

No other comments, the rest is good for me.

src/ui/simulator/toolbox/components/datagrid/renderer/mc-playlist.cpp
```diff
@@ -159,9 +159,9 @@ namespace Renderer
                 }
                 case MCPlaylistCol::WEIGHT:
                 {
-                    std::vector<int> yearsWeight = study->parameters.getYearsWeight();
+                    std::vector<float> yearsWeight = study->parameters.getYearsWeight();
                     assert(y < yearsWeight.size());
-                    return wxString::Format(wxT("%i"), yearsWeight[y]);
+                    return wxString::Format(wxT("%.3f"), yearsWeight[y]);
```
guilpier-code, line 164: Maybe we do not need that precision : 3 significant digits are enough.
2 digits after decimal points should be ok

src/ui/simulator/toolbox/components/datagrid/renderer/mc-playlist.cpp
```diff
@@ -159,9 +162,14 @@ namespace Renderer
...
-                    std::vector<int> yearsWeight = study->parameters.getYearsWeight();
+                    std::vector<float> yearsWeight = study->parameters.getYearsWeight();
                     assert(y < yearsWeight.size());
-                    return wxString::Format(wxT("%i"), yearsWeight[y]);
+
+                    std::ostringstream stream;
+                    stream << std::setprecision(3);
+                    stream << yearsWeight[y];
+
+                    return stream.str();
```
flomnes, line 172: why not just change the format to `"%.3f"` in the existing code ?

jmkerloch: This was used before : 
`return wxString::Format(wxT("%.3f"), yearsWeight[y])`
but this will display 100.00

With this stream it will display 100

guilpier-code (approved):

flomnes (approved):

PR #108, opened 2021-01-13
Scenario builder + hydro levels : bug fix

guilpier-code:
For a year and area, a scenario builder initial hydro level is meant to be the level applied at the initialization date of the heuristic.
It is not meant to be (like it was done before this commit) the initial level of the year and are, whatever the initialization date of the heuristic.


jmkerloch (approved):
OK for me

Issue #113, opened 2021-01-18
scenario builder + hydro : no symetry regarding automatic reframing of levels entries 


guilpier-code:
### GUI behaviours:
In the scenario builder hydro levels tab, user's input are reframed if they are not contained inside bounds 0 and 100.
Indeed, if user gives a number < 0 or > 100, the GUI automatic correction does not respond in the same way :
- if value<0 -> 0
- if value>100 ->100
- if values is not a `double` -> rand

### StudyReader behaviour:
- if value is <0 ->0
- if value >100  -> 100
- if values is not a `double` -> rand

PR #116, opened 2021-01-19
Fix/sc builder hydro  auto reframe entry

guilpier-code:
This PR is meant to fix the ticket #113

### DIFF
--- a/src/libs/antares/study/scenario-builder/hydroLevelsData.cpp
+++ b/src/libs/antares/study/scenario-builder/hydroLevelsData.cpp
@@ -28,6 +28,7 @@
 #include <sstream>
 #include <iomanip>
 #include "hydroLevelsData.h"
+#include "scBuilderUtils.h"
 
 namespace Antares
 {
@@ -43,7 +44,7 @@ namespace ScenarioBuilder
 		const uint nbYears = study.parameters.nbYears;
 
 		pHydroLevelsRules.reset(study.areas.size(), nbYears);
-		pHydroLevelsRules.fill(-1.);
+		pHydroLevelsRules.fill(default_value());
 
 		return true;
 	}
@@ -70,7 +71,7 @@ namespace ScenarioBuilder
 			{
 				const MatrixType::Type  value = col[y];
 				// Equals to zero means 'auto', which is the default mode
-				if (value < 0.)
+				if (value == default_value())
 					continue;
 				assert(index < study.areas.size());
 				value_into_string << value;
--- a/src/libs/antares/study/scenario-builder/hydroLevelsData.cpp
+++ b/src/libs/antares/study/scenario-builder/hydroLevelsData.cpp
@@ -44,7 +44,7 @@ namespace ScenarioBuilder
 		const uint nbYears = study.parameters.nbYears;
 
 		pHydroLevelsRules.reset(study.areas.size(), nbYears);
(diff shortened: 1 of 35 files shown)


src/libs/antares/study/scenario-builder/scBuilderUtils.h
```diff
@@ -39,6 +39,11 @@ namespace Data
 {
 namespace ScenarioBuilder
 {
+	const double NOT_A_DOUBLE = 12345;
```
jmkerloch, line 42: We could use [`std::nan`](https://en.cppreference.com/w/cpp/numeric/math/nan) as an undefined value instead of a antares custom on ?

src/libs/antares/study/scenario-builder/scBuilderUtils.cpp
```diff
@@ -59,11 +64,11 @@ namespace ScenarioBuilder
 		}
 		catch (std::invalid_argument&)
 		{
-			return -1.;
+			return not_a_double();
```
jmkerloch, line 67: In this case we will return `std::nan()`

src/libs/antares/study/scenario-builder/hydroLevelsData.cpp
```diff
@@ -70,7 +71,7 @@ namespace ScenarioBuilder
 			{
 				const MatrixType::Type  value = col[y];
 				// Equals to zero means 'auto', which is the default mode
-				if (value < 0.)
+				if (value == default_value())
```
jmkerloch, line 74: Check value with [`std::isnan`](https://en.cppreference.com/w/cpp/numeric/math/isnan)

src/libs/antares/study/scenario-builder/hydroLevelsData.cpp
```diff
@@ -43,7 +44,7 @@ namespace ScenarioBuilder
 		const uint nbYears = study.parameters.nbYears;
 
 		pHydroLevelsRules.reset(study.areas.size(), nbYears);
-		pHydroLevelsRules.fill(-1.);
+		pHydroLevelsRules.fill(default_value());
```
jmkerloch, line 47: Fill with [`std::nan`](https://en.cppreference.com/w/cpp/numeric/math/nan)

src/ui/simulator/toolbox/components/datagrid/renderer/scenario-builder-hydro-levels-renderer.cpp
```diff
@@ -47,7 +47,7 @@ namespace Renderer
 	{
 		const double d = cellNumericValue(x, y);
 
-		return (d < 0.)
+		return (is_not_a_double(d))
```
jmkerloch, line 50: Check value with [`std::isnan`](https://en.cppreference.com/w/cpp/numeric/math/isnan).
This should work even if `d` was multipled by `100.0` in `cellNumericValue`

jmkerloch (changes requested):
This should be better to use std::nan for undefined value instead of an Antares custom one.

jmkerloch:
There is also a mismatch in gui for rand value display between Hydro Levels definition and Solar for example : 
image
In Hydro rand value are not gray but 0 values are.
image
In Solar rand value are gray and defined values are not.

jmkerloch (approved):
OK with changes

PR #139, opened 2021-02-02
Antares Test : unfeasible problem pytest

jmkerloch:
closes #139

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -128,6 +128,21 @@ if(Python3_Interpreter_FOUND)
             output_tests(false, "sirius")
         endif()
         
+        # Specific tests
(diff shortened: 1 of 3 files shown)


src/tests/CMakeLists.txt
```diff
@@ -128,6 +128,21 @@ if(Python3_Interpreter_FOUND)
             output_tests(false, "sirius")
         endif()
         
+        # Specific tests
+        add_test(
+            NAME short-specific
```
flomnes, line 133: replace "specific" by "infeasibility" ?

jmkerloch: "specific" choose over "infeasibility" if we want to add more pytest for other functionnality

PR #124, opened 2021-01-21
Feature/xpansion  split mps

guilpier-code:
Antares for Xpansion : split MPS into fixed and variable part.

When asking Xpansion mode, the first week of the simulation dumps the fixed part of the first optimization problem on disk, and every week dumps the its own variable part of the optimization problem.

This PR regards more specially the split the division ot MPS data into its fixed part and variable part.

This PR is related to issue #121

PR #144, opened 2021-02-08
Mode xpansion - splitting MPS files depending on simulation week

guilpier-code:
In mode expansion, the first simulation week prints the fixed part of its optimization problem, and each week (first week included) prints te variable part of the simulation.
Thus the first week of simulation has to be spotted by the code.

This PR "merges" the 2 foolowing PRs : #124 and #110


guilpier-code:
**Questions on this issue** :
- Currently,  the expansion mode is a particular value of the parameter **simulation mode**. This parameter can take the following values : ecnonmy, adequacy, adequcy-draft and expansion (expansion cannot be set via GUI, it can only be set manually, in the **generaldata.ini file**). For the current work, we made the printings of new split MPS files relie on the value the paramater  **simulation mode**. It is set to Expansion then we split MPS into fixed and variable parts. Is this what we want ?
- Currently, in the expansion mode, user do not need to ask for printing special split MPS files, it is done automatically. Should the user be asking for exporting MPS in the study ? And otherwise (user does not need to ask), how does the : GUI > Configuration > optimization preferences > Export MPS should behave ?

flomnes:
Resolve conflicts

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -415,6 +430,236 @@ void OPT_EcrireResultatFonctionObjectiveAuFormatTXT(void* Prob,
...
+            }
+
+            il++;
+        }
+    }
+
+    free(Cder);
+
+    auto& study = *Data::Study::Current::Get();
+    Flot = study.createFileIntoOutputWithExtension("fixed-part", "mps", numSpace);
```
```suggestion
    Flot = study.createFileIntoOutputWithExtension("problem-fixed-part", "mps", numSpace);
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -415,6 +430,236 @@ void OPT_EcrireResultatFonctionObjectiveAuFormatTXT(void* Prob,
...
+
+    if (Nombre == NULL)
+    {
+        logs.fatal() << "Not enough memory";
+        AntaresSolverEmergencyShutdown();
+    }
+
+    // gp : to be changed
+    auto& study = *Data::Study::Current::Get();
+    Flot = study.createFileIntoOutputWithExtension("variable-part", "mps", numSpace);
```
```suggestion
    Flot = study.createFileIntoOutputWithExtension("problem-variable-part", "mps", numSpace);
```

PR #148, opened 2021-02-10
disable output compare on linux for unfeasible problem tests

jmkerloch:


flomnes (changes requested):
- differents -> different
- sys.platform returns "linux2" only with python <=3.3, but these obsolete versions are none of our concern. Thus, I suggest replacing `startswith` by `==`

flomnes (approved):

Issue #137, opened 2021-02-01
Separating dowload of tests and examples


guilpier-code:
Tests (from repo **Antares_Simulator_Tests**) and examples (from repo **Antares_Simulator_Examples**) are downloaded together at cmake configuration time (when option **-DBUILD_TESTING=ON** is given).

As examples are meant to be embedded in the Antares installer, they are not necessary for tests.

Examples should be downloaded at CPack time.

PR #151, opened 2021-02-11
Downloading tests and examples

guilpier-code:
Downloads of tests (Antares_Simulator_Tests) and examples (Antares_Simulator_Examples) used to be made together at cmake configuration time.
We keep on downloading tests at configuration time, but we download examples at packaging time (CPack).
This PR is related to #137


sgatto (approved):
OK, thanks!

PR #157, opened 2021-02-17
Skip connection to license server 2

guilpier-code:
When opening Antares GUI, the connection to license server dialog box is displayed.
Add a text informing the user of how to continue using Antares without connection to the server at all (offline usage),
for this session or the any further session.

### DIFF
--- a/src/ui/simulator/windows/proxy/proxysetup.cpp
+++ b/src/ui/simulator/windows/proxy/proxysetup.cpp
@@ -111,8 +111,7 @@ LicenseCouldNotConnectToInternetServer::LicenseCouldNotConnectToInternetServer(w
             "please click on \"Cancel\", then unfold Antares help (\"?\") menu and\n"
             "select \"Continue offline\"."),
         false,
-        false,
-        +1);
+        false);
     contentSizer->AddSpacer(20);
     contentSizer->Add(titlespacer, 0, wxALL | wxEXPAND);
     contentSizer->AddSpacer(40);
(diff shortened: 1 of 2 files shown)


src/ui/simulator/windows/proxy/proxysetup.cpp
```diff
@@ -114,6 +121,8 @@ LicenseCouldNotConnectToInternetServer::LicenseCouldNotConnectToInternetServer(w
     subtitlespacer->Add(title, 0, wxLEFT);
     subtitlespacer->AddSpacer(3);
     subtitlespacer->Add(subtitle, 0, wxLEFT);
+    subtitlespacer->AddSpacer(10);
+    subtitlespacer->Add(offline_title, 0, wxLEFT);
```
flomnes, line 125: `offline_title` is never deallocated, as well as other variables such as `title`. `subtitle`, etc.

More generally, this function has dozens of memory leaks...

For now, I suggest that you `delete` only `offline_title` and add a TODO note

jmkerloch (approved):

src/ui/simulator/windows/proxy/proxysetup.cpp
jmkerloch: I don't know if this is the case with wxWidget but with Qt you sometimes don't delete created pointer because they become children of another widget. 

The parent is then responsible for deleting the children

guilpier-code: @flomnes : you're right, there are things to destroy here, and probably in a lot of other files !
I destroyed the pointer you mention, and i will post an issue for this.

flomnes (approved):

PR #163, opened 2021-02-19
Add option BUILD_SWAP 

jmkerloch:


src/tools/CMakeLists.txt
```diff
@@ -2,7 +2,12 @@ add_subdirectory(batchrun)
 add_subdirectory(finder)
 add_subdirectory(updater)
 add_subdirectory(cleaner)
-add_subdirectory(yby-aggregator)
+
+#yby-aggregator need swap build
+if(BUILD_SWAP)
+    add_subdirectory(yby-aggregator)
+endif()
```
flomnes, line 9: yby-agregator's main.cpp has a few
```cpp
#ifdef ANTARES_SWAP_SUPPORT
// ...
#endif
```
So it should work without swap support, right ? Does yby-aggregator depend on libsomething-swap ?

flomnes: Consider adding the `ANTARES_SWAP_SUPPORT`flag instead

flomnes (changes requested):

sgatto (approved):

src/tools/CMakeLists.txt
jmkerloch: In CMakeLists.txt, yby-aggregator depends on 
```

set(YBY_AGGREGATOR_LIBS
	libantares-core-swap
	libantares-core-calendar
	yuni-static-core
	libantares-license
	${wxWidgets_LIBRARIES} ${CMAKE_THREADS_LIBS_INIT})
```

sgatto: I suspect that yby aggregator needs a swap support because it may need a lot of memory when running.
The `#ifdef` probably need to be reviewed

flomnes (approved):

guilpier-code (approved):

PR #179, opened 2021-02-25
Code cleaning : PROBLEME_HEBDO

flomnes:
* Remove unused members
* Sort used members into business problem/optimization problem/hydro management/implementation details
* Remove `OPT_OptimisationLineaire`'s unused parameters


jmkerloch (approved):

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -474,191 +474,82 @@ typedef struct
...
-    unsigned int HeureDansLAnnee;
-
-    char LeProblemeADejaEteInstancie;
-    char TypeDOptimisation;
-
     int NombreDePays;
     const char** NomsDesPays;
     int NombreDePaliersThermiques;
 
-    int NumberOfMarketOfferPool;
```
guilpier-code, line 524: **Caution :**
a bunch of variables were introduced here as a start for a future work : down flexibility of thermal energy.
Not sure it is a good idea too remove them. To be discussed.

guilpier-code (changes requested):

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: I've restored unused variables, with a comment and an `#if 0` condition surrounding their definition.

guilpier-code (approved):

PR #181, opened 2021-02-26
Add "playlist" parameter, refactor yearsFilter

flomnes:
New parameter playlist to be used with xpansion, see [this PR](https://github.com/AntaresSimulatorTeam/antares-xpansion/pull/50)


src/libs/antares/study/parameters.cpp
```diff
@@ -612,8 +610,14 @@ static bool SGDIntLoadFamily_N(Parameters& d, const String& key, const String& v
 
 static bool SGDIntLoadFamily_P(Parameters& d, const String& key, const String& value, uint)
 {
+    const auto warning = [&logs, &key]() {
```
guilpier-code: Instead of "Consider using playlist instead.", rather "Please save your study to upgrade it."

guilpier-code: This code does not compile on my machine : 
**parameters.cpp(613,38): error C3480: 'Antares::logs': a lambda capture must be a variable from an enclosing function scope**

guilpier-code (changes requested):

src/libs/antares/study/parameters.cpp
jmkerloch: @guilpier can you indicate MSVC version used ?

guilpier-code: This problem was solved for me in latest commit

src/libs/antares/study/parameters.cpp
```diff
@@ -1045,26 +1069,7 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
...
-                delete[] yearsFilter;
-                yearsFilter = newset;
-            }
-            else
-            {
-                yearsFilter = new bool[options.nbYears];
-                for (uint i = 0; i < options.nbYears; ++i)
-                    yearsFilter[i] = true;
-            }
-        }
```
flomnes, line 1067: Behavior is different when `options.nbYears <= nbYears`

flomnes: behavior is the same, my mistake

src/libs/antares/study/parameters.cpp
```diff
@@ -650,8 +658,26 @@ static bool SGDIntLoadFamily_P(Parameters& d, const String& key, const String& v
...
+        // Check indices
+        for (int y : in_playlist)
+            if ((uint)y >= d.nbYears || y < 0)
+                return false;
+        // All indices good, change d.yearsFilter
+        for (uint y = 0; y < d.nbYears; ++y)
+            d.yearsFilter[y] = false;
+        for (int y : in_playlist)
+            d.yearsFilter[y] = true;
+        return true;
```
jmkerloch, line 675: So if we want to activate all but deactivate a few years, we need to indicate all the activated years ?

src/libs/antares/study/parameters.cpp
```diff
@@ -697,6 +723,35 @@ static bool SGDIntLoadFamily_P(Parameters& d, const String& key, const String& v
...
+
+        if (weight_in < 0.f)
+        {
+            logs.warning() << "parameters: invalid MC year weight.Got '" << weight_in
+                           << "'. Value not used";
+            weight_in = 1.f;
+        }
+
+        d.setYearWeight(year_in, weight_in);
+        return true;
```
jmkerloch, line 752: Could we use a kind of a map ?
`{1: 2.3 , 2 : 4.5}`

src/libs/antares/study/parameters.cpp
```diff
@@ -1045,26 +1100,7 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
...
-                for (uint i = nbYears; i < options.nbYears; ++i)
-                    newset[i] = false;
-                delete[] yearsFilter;
-                yearsFilter = newset;
-            }
-            else
-            {
-                yearsFilter = new bool[options.nbYears];
-                for (uint i = 0; i < options.nbYears; ++i)
-                    yearsFilter[i] = true;
```
jmkerloch, line 1065: If there was no yearsFilter we init filter with true.

With new implementation resize is done with false value.

To be discussed.

src/libs/antares/study/parameters.cpp
```diff
@@ -1697,49 +1725,88 @@ void Parameters::saveToINI(IniFile& ini) const
...
-        bool weightEnabled = false;
-        for (uint i = 0; i != nbYears; ++i)
+        assert(not yearsFilter.empty());
+
+
+        if (version <= version800)
         {
-            if (yearsFilter[i])
+            // Playlist section must be added if at least one year is disable or one MC year weight
+            // sum is not equal to nbYears
```
jmkerloch, line 1737: report comment with correction for MC years weight check

src/tests/end-to-end/CMakeLists.txt
```diff
@@ -18,20 +34,46 @@ target_link_libraries(tests-simple-study
 						libantares-solver-simulation
 						libantares-solver-ts-generator
 						libmodel_antares
-						Boost::unit_test_framework
+                        Catch2::Catch2
```
jmkerloch: BOOST still used for simple-study.cpp

src/tests/end-to-end/playlist.cpp
```diff
@@ -0,0 +1,256 @@
...
+custom-scenario = false
+user-playlist = true
+
+[playlist]
+playlist_reset = false
+playlist_year += 0
+playlist_year += 1)");
+
+    StudyLoadOptions loadOpt;
+    loadOpt.nbYears = 5;
```
jmkerloch, line 105: We could add a test with a `loadOpt.nbyears` different than `nbyears `in `generaldata.ini`

jmkerloch (approved):

jmkerloch (changes requested):
Some questions and remarks before merge

src/libs/antares/study/parameters.cpp
flomnes: we could, but it's harder to parse and possibly less human-readable.

src/tests/end-to-end/playlist.cpp
flomnes: what is the expected behavior ?

src/libs/antares/study/parameters.cpp
flomnes: right

src/libs/antares/study/parameters.cpp
flomnes: done

src/libs/antares/study/parameters.cpp
flomnes: I'm not sure I understand, can you clarify ?

src/tests/end-to-end/playlist.cpp
flomnes:
```
/home/florian/Antares_Simulator/src/tests/end-to-end/playlist.cpp:110: FAILED:
  REQUIRE( p.nbYears == 5 )
with expansion:
  3 == 5
```

flomnes: Apparently, `loadOpt.nbYears` overrides `generaldata.ini`'s `nbyears`.

flomnes: I will leave `loadOpt.nbYears=5` for now. I do not know if we want to support other values than the one provided in `generaldata.ini`.

flomnes (commented):
I'll keep the current syntax as is for now
```ini
playlist_weight_year_1 = 2.2
playlist_weight_year_3 = 1.1
```
@jmkerloch suggests
```ini
playlist_weights = {1 : 2.2, 3 : 1.1}
```

flomnes:
After some discussions with the development team, the decision was made not to include this change to the next version of Antares Simulator.

Main reason : compatibility issues. Antares API should make these changes less useful in the future.

PR #185, opened 2021-03-01
correction of MC years weight write

### LINKED ISSUE #184: Weights not written when sum == number of total years
### How to reproduce

- Create a study with one node
- Select 2 MC years
- Select weights {0.8, 1.2} (not {1, 1}, but sum = number of total MC years)
- Save study

image

File **generaldata.ini** should contain weights 0.8, 1.2 but it doesn't.

### What to change
See **src/libs/antares/study/parameters.cpp**
```cpp
        bool addPlayListSection = effNbYears != nbYears;
        addPlayListSection |= weightSum != nbYears;
```

This flag should be set to true if **any** weight is `!= 1.f`, not if their sum is `!= nbYears`.

jmkerloch:
closes #184

### DIFF
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1699,18 +1699,17 @@ void Parameters::saveToINI(IniFile& ini) const
     {
         assert(yearsFilter);
         uint effNbYears = 0;
-        float weightSum = 0;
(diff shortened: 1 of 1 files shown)


flomnes:
Confirmed : fixes issue #184.

flomnes (approved):

PR #192, opened 2021-03-03
fix hydro hot start with refresh span enabled

### LINKED ISSUE #190: error on hydro study with playlist use and stochastic TS refresh span
There is an error when launching a hydro study with playlist use and refresh span (see study [069 Hydro Reservoir Model -enhanced-02-custom.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/6070000/069.Hydro.Reservoir.Model.-enhanced-02-custom.zip)) : 

image

This is seems to be cause by the use of playlist with stochastic TS refresh span : 

image

image

If refresh span is disabled or first MC year is activated there is no issue with the study.

jmkerloch:
closes #190

### DIFF
--- a/src/solver/simulation/solver.h
+++ b/src/solver/simulation/solver.h
@@ -167,8 +167,8 @@ private:
     HydroManagement pHydroManagement;
     //! Hydro hot start
     bool pHydroHotStart;
-    //! The first set of parallel year(s) was already run ?
-    bool pFirstSetParallelWasRun;
+    //! The first set of parallel year(s) with a performed year was already run ?
+    bool pFirstSetParallelWithAPerformedYearWasRun;
 
     //! Statistics about annual (system and solution) costs
     annualCostsStatistics pAnnualCostsStatistics;
--- a/src/solver/simulation/solver.hxx
+++ b/src/solver/simulation/solver.hxx
@@ -64,7 +64,7 @@ public:
             std::vector<unsigned int>& pYearsIndices,
             std::map<uint, bool>& pYearFailed,
             std::map<uint, bool>& pIsFirstPerformedYearOfASet,
-            bool pFirstSetParallelWasRun,
+            bool pFirstSetParallelWithAPerformedYearWasRun,
             unsigned int pNumSpace,
             randomNumbers& pRandomForParallelYears,
             bool pPerformCalculations,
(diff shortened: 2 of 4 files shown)


sgatto (approved):

src/solver/simulation/solver.hxx
```diff
@@ -1581,8 +1582,8 @@ void ISimulation<Impl>::loopThroughYears(uint firstYear,
         qs.wait(Yuni::qseIdle);
         qs.stop();
 
-        // At this point, the first set of parallel year(s) was run
-        if (!pFirstSetParallelWasRun)
+        // At this point, the first set of parallel year(s) was run with at least one year performed
+        if (!pFirstSetParallelWasRun && yearPerformed)
```
guilpier-code: - Ok, **pFirstSetParallelWasRun** has a bad name, and it should be replaced by its opposite : isFirstParallelSet, more clear (not sure it can be done within the current PR).
- We should even have a variable **isFirstParallelSetWithAPerformedYear**
- At least, if we should rename **pFirstSetParallelWasRun** as **pFirstSetParallelWithAPerformedYearWasRun**

guilpier-code (changes requested):

src/solver/simulation/solver.hxx
guilpier-code: For now, not sure we need an extra variable **yearPerformed**, we can use what already exist

guilpier-code (approved):

src/solver/simulation/solver.h
```diff
@@ -167,8 +167,8 @@ class ISimulation : public Impl
     HydroManagement pHydroManagement;
     //! Hydro hot start
     bool pHydroHotStart;
-    //! The first set of parallel year(s) was already run ?
-    bool pFirstSetParallelWasRun;
+    //! The first set of parallel year(s) with a performed year was already run ?
+    bool pFirstSetParallelWithAPerformedYearWasRun;
```
sgatto, line 171: I would correct the english
it is a paralleSet rather than a SetParallel
`pFirstParalleSetWithAPerformedYearWasRun;`

sgatto (commented):
I propose to change (again) the name of the variable

PR #217, opened 2021-04-01
Remove license - (1)   : removing 2 important files and consequences

guilpier-code:
The 2 files we want to empty/remove are :
- src/ui/simulator/application/main/check-license-at-startup.hxx
- src/internet/hello_web.cpp

Others files modifications are consequences of removing / emptying these 2 files


src/solver/main.cpp
```diff
@@ -568,8 +568,8 @@ bool SolverApplication::completeWithOnlineCheck()
     }
 
     // verify License online
-    if (not License::CheckOnlineLicenseValidity(Data::versionLatest, true))
-        return false;
+    // if (not License::CheckOnlineLicenseValidity(Data::versionLatest, true))
```
guilpier-code, line 571: The whole function is removed in the PRs that follow

flomnes (approved):

PR #218, opened 2021-04-07
Remove license - (2) : removing proxy and consequences

guilpier-code:
Before reviewing this PR, please until PR #217 is merged into branch **feature/remove_license**

PR #219, opened 2021-04-07
Remove license - (3) : moving limits to an existing constants file

guilpier-code:
Before reviewing this PR, please until PR #218 is merged into branch **feature/remove_license**

When reviewing, please have a particular care on how the content of **limits.h/cpp** was moved to **constant.h**.
This element is to be discussed.


src/libs/antares/constants.h
```diff
@@ -44,6 +44,11 @@
 /*! */
 /* # define ANTARES_USE_GLOBAL_MAXIMUM_COST 0 */
 
+// The maximum number of areas allowed in a study
+#define MAX_NUMBER_OF_AREAS 9999
+// The maximum number of thermal plants per area allowed in a study
+#define MAX_NUMBER_OF_THERMAL_CLUSTERS_PER_AREA 999
```
flomnes, line 50: It is my opinion that there should be no arbitrary limit on the number of areas or clusters per area.

While I agree that human-made studies will almost never go beyond these limits, computer-generated studies could.

src/analyzer/atsp/load.cpp
```diff
@@ -205,7 +205,7 @@ bool ATSP::loadFromINIFile(const String& filename)
                 }
 
                 pArea.push_back(info);
-                if (License::Limits::areaCount and pArea.size() > License::Limits::areaCount)
+                if (MAX_NUMBER_OF_AREAS and pArea.size() > MAX_NUMBER_OF_AREAS)
                     return false;
```
flomnes, line 209: remove this condition

flomnes (changes requested):

PR #230, opened 2021-04-20
Simplify ON_min computation

flomnes:
max(min(a,b),b)=b for all (a,b)

with
```
a = static_cast<uint>(Math::Floor(thermalClusterPMinOfTheClusterForYear[h]
    / currentCluster->pminOfAGroup[numSpace]));

b = static_cast<uint>(Math::Ceil(thermalClusterAvailableProduction
  / currentCluster->nominalCapacityWithSpinning));
```


flomnes:
`thermalClusterAvailableProduction` and `thermalClusterProduction` are different variables. This simplification is not valid (thanks, CI)

PR #248, opened 2021-05-06
Renewable internal structures, TS loading [1/3]

### LINKED ISSUE #220: Feature wind off-shore (or miscellaneus renewable)
 
## Antares main Window

## Scenario builder

## Input

### Building outputs

### Reading outputs
- Tab **General values**
- Tab **Renewable values**

## Solver

flomnes:


src/libs/antares/study/area/area.hxx
```diff
@@ -61,7 +61,9 @@ inline bool Area::storeTimeseriesNumbers(Study& study)
         // Thermal
     case timeSeriesThermal:
         return storeTimeseriesNumbersForThermal(study);
-
+        // Renewable
+    case timeSeriesRenewable:
```
guilpier-code, line 65: It is possible that current function (**Area::storeTimeseriesNumbers(Study& study)**) is never called for renewables.

src/libs/antares/study/area/list.cpp
```diff
@@ -1212,6 +1232,28 @@ bool AreaList::loadFromFolder(const StudyLoadOptions& options)
         }
     }
 
+    // Renewable data, specific to areas
+    {
+        logs.info() << "Loading renewable clusters...";
+        buffer.clear() << pStudy.folderInput << SEP << "renewables" << SEP << "areas.ini";
```
guilpier-code, line 1245: For thermal clusters, this areas.ini file contains unsupplied and spilled energy costs.
I don't know why this information is located inside thermal directory, as it does not regard thermal only. Maybe there is reason, but I do not see it now.
Anyway, for renewable, this information does not seem to have any meaning.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+                // adding the renewable cluster
+                if (not add(cluster))
+                {
+                    // This error should never happen
+                    logs.error() << "Impossible to add the renewable cluster '" << cluster->name()
+                                 << "'";
+                    delete cluster;
+                    continue;
+                }
+                // keeping track of the cluster
```
guilpier-code: For now, some load of modulations (marginal cost, capacity, ...) are missing. Will be added later.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+                mapping[cluster->id()] = cluster;
```
guilpier-code: Why do we keep 2 maps with clusters (**mapping** and the one we populate in **add(cluster)** instruction).
Maybe useful for thermal clusters (eventually to separate must-run clusters from others in my memory), not sure it is for renewable clusters.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+
+                cluster->flush();
```
guilpier-code: As flush function does nothing, why do we flush ?

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+            IniFile::Section* s = ini.addSection(cluster.name());
+
+            // The section must not be empty
+            // This key will be silently ignored the next time
+            s->add("name", cluster.name());
+
+            if (not cluster.group().empty())
+                s->add("group", cluster.group());
+            if (not cluster.enabled)
+                s->add("enabled", "false");
```
guilpier-code: For now, some parameters are missing (nominal capacity, marginal cost, ...), but they will be taken into account when specification are more advanced.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+    nominalCapacity = cluster.nominalCapacity;
+
+    // Making sure that the data related to the timeseries are present
+    if (not series)
+        series = new DataSeriesRenewable();
+
+    // timseries
+    series->series = cluster.series->series;
+    cluster.series->series.unloadFromMemory();
+    series->timeseriesNumbers.clear();
```
guilpier-code, line 101: ok, probably some missing parameters

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+        byIndex = nullptr;
+    }
+
+    auto end = mapping.end();
+    for (auto it = mapping.begin(); it != end; ++it)
+        delete it->second;
+    mapping.clear();
+
+    if (not cluster.empty())
+        cluster.clear();
```
guilpier-code: in the case of renewable clusters, **mapping** and **cluster** lists are probably redundants.
They are not destroyed in the same way but it's relevant, as values of these maps are addresses to clusters.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+Data::RenewableCluster* RenewableClusterList::find(const RenewableCluster* p)
+{
+    auto end = cluster.end();
+    for (auto i = cluster.begin(); i != end; ++i)
+    {
+        if (p == i->second)
+            return i->second;
+    }
+    return nullptr;
+}
```
guilpier-code: one of the 2 **find** methods is useless : if we comment the non const method, Antares still compiles.
Moreover, if **cluster.h** is touched, the recompilation is very long : a lot of files are recompiled. Why is that ? (maybe it's expected, to be checked).
Both remarks apply to thermal clusters as well, of course.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+    if (not newgrp)
+    {
+        groupID = renewableOther1;
+        pGroup.clear();
+        return;
+    }
+    pGroup = newgrp;
+    newgrp.toLower();
+
+    switch (newgrp[0])
```
guilpier-code: Cannot figure out why we start testing **newgrp[0]** before finally testing **newgrp**. Why not testing **newgrp** directly ? It would be more simple.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+                cluster->flush();
+            }
+        }
+
+        return ret;
+    }
+    return false;
+}
+
+int RenewableClusterListSaveDataSeriesToFolder(const RenewableClusterList* l,
```
guilpier-code: Not called yet

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+    for (auto it = l->cluster.begin(); it != end; ++it)
+    {
+        auto& cluster = *(it->second);
+        if (cluster.series)
+            ret = DataSeriesRenewableSaveToFolder(cluster.series, &cluster, folder) and ret;
+    }
+    return ret;
+}
+
+int RenewableClusterListSaveDataSeriesToFolder(const RenewableClusterList* l,
```
guilpier-code: It seems this function won't be called, as it is called for thermal clusters when storing generated TS. So it should be removed.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,837 @@
...
+    series->series = cluster.series->series;
+    cluster.series->series.unloadFromMemory();
+    series->timeseriesNumbers.clear();
+
+    // The parent must be invalidated to make sure that the clusters are really
+    // re-written at the next 'Save' from the user interface.
+    if (parentArea)
+        parentArea->invalidate();
+}
+
```
guilpier-code, line 108: Cluster and list of clusters are defined in the same files cluster.h/hxx/cpp. A separation would be clearer.

src/libs/antares/study/parts/renewable/cluster.hxx
```diff
@@ -0,0 +1,131 @@
...
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#ifndef __ANTARES_LIBS_STUDY_PARTS_RENEWABLE_CLUSTER_HXX__
+#define __ANTARES_LIBS_STUDY_PARTS_RENEWABLE_CLUSTER_HXX__
+
```
guilpier-code, line 29: As already said, we should split file cluster.* into cluster.* and clusterList.*

src/libs/antares/study/parts/renewable/container.h
```diff
@@ -0,0 +1,120 @@
...
+    ** \brief Removes disabled renewable clusters
+    **
+    ** All clusters with the flag 'enabled' turned to false will be removed from 'list'.
+    ** As a consequence, they will no longer be seen as renewable clusters
+    ** from the solver's point of view.
+    ** \warning This method should only be used from the solver
+    **
+    ** \return The number of disabled clusters found
+    */
+    uint removeDisabledClusters();
```
guilpier-code, line 84: For now, this function is not used

src/libs/antares/study/parts/renewable/container.cpp
```diff
@@ -0,0 +1,140 @@
...
+    for (auto i = list.begin(); i != end; ++i)
+    {
+        RenewableCluster* t = i->second;
+        t->areaWideIndex = idx;
+        clusters[idx] = t;
+        ++idx;
+    }
+}
+
+uint PartRenewable::removeDisabledClusters()
```
guilpier-code, line 95: why is this function so complicated, with 2 loops (**do** and **for**) intertwined ?

src/libs/antares/study/parts/renewable/series.h
```diff
@@ -0,0 +1,107 @@
...
+#include "defines.h"
+
+namespace Antares
+{
+namespace Data
+{
+/*!
+** \brief Data series (Renewable)
+*/
+class DataSeriesRenewable
```
guilpier-code: This class does not seem to be different from its thermal counterpart.

src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
```diff
@@ -93,6 +93,27 @@ static void InitializeTimeSeriesNumbers_And_ThermalClusterProductionCost(
             ptchro.Eolien
               = (data.series.width != 1) ? (long)data.timeseriesNumbers[0][year] : 0; // zero-based
         }
+        // Renewable
+        {
+            auto end = area.renewable.list.mapping.end();
```
guilpier-code: to be checked if **mapping** and **cluster** list are redundant

src/ui/simulator/toolbox/components/datagrid/renderer/simulation.cpp
```diff
@@ -33,6 +33,16 @@
 
 using namespace Yuni;
 
+// Anonymous namespace: global variable, local scope
+namespace
+{
+enum Antares::Data::TimeSeries mapping[] = {Data::timeSeriesLoad,
```
guilpier-code, line 39: This file is part of GUI, but ok. Why do we not add **Data::timeSeriesRenewable** then ?

guilpier-code (changes requested):

src/libs/antares/study/fwd.h
```diff
@@ -68,6 +68,8 @@ typedef Yuni::CString<ant_k_area_name_max_length * 2 + 1, false> AreaLinkName;
 //! Name of a single thermal
 typedef Yuni::CString<ant_k_cluster_name_max_length, false> ThermalClusterName;
 
+typedef Yuni::CString<ant_k_cluster_name_max_length, false> RenewableClusterName;
```
guilpier-code: **ClusterName** should be enough for both thermal and renewable

src/libs/antares/study/area/list.cpp
flomnes: Right.

This is only an hypothesis, but energy inadequacies (spilled or unsupplied) are used to determine which thermal cluster to use.

However, the hydro part is also controllable.

Therefore, spillage & unsupplied costs should be placed in a per-area directory.

src/libs/antares/study/parts/renewable/container.cpp
flomnes: These two loops are used to remove every disabled cluster from the list.

Iterator `list.end()` can change and become invalid if `list.remove(i);` is called. In this case, we get out of the `for` (with the `break` instruction) loop to reset this iterator.

Since `list.remove(i)` invalidates `list.end()`, the two loops are needed. See **src/libs/antares/study/parts/renewable/cluster.h**

```cpp
/*!
** \brief Remove a cluster represented by an iterator
**
** The renewable cluster will be removed from the list but _not_
** destroyed.
** The iterator should considered as invalid after using this method.
** \return void
*/
void remove(iterator i);
```

src/libs/antares/study/parts/renewable/container.h
flomnes: Fixed in commit 7b11174.

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: `mapping` is redundant indeed. Fixed in commit d9895347.

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: **part.h** is included in many places, it includes **cluster.h** indirectly. Thus, whenever it is changed, the recompilation takes time.

One method is `const`, the other is not. The non-`const` method is probably not used.

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: This is a clunky optimization. Removed in commit f6c8b85

src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
flomnes: `mapping` is redundant. Removed in commit d9895347

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: `mapping` removed in commit d9895347

src/libs/antares/study/area/area.hxx
flomnes: Fixed in commit 7b11174

flomnes: It is called after PR #251, added in commit 4be8e9f41.

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: In the swap flavor, some actions are required for time-series.

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: Right, fixed in commit 394a48b.

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: I disagree. For thermal clusters, there are 2 calls :
> libs/antares/study/study.importprepro.cpp:122:23:                ret = ThermalClusterListSaveDataSeriesToFolder(
> libs/antares/study/area/list.cpp:335:15:        ret = ThermalClusterListSaveDataSeriesToFolder(&area.thermal.list, buffer) and ret;

The first call is used to export generated TS into the input folder. It does not make sense for renewables, as long as no TS generation is possible.

The second call is used e.g when the user changes TS values in the GUI (function `AreaList::saveToFolder`, called by `Study::saveToFolder`). This call makes sense for renewable clusters.

See commit 394a48b.

src/libs/antares/study/parts/renewable/cluster.cpp
guilpier-code: why do we not remove it then ?

src/libs/antares/study/parts/renewable/cluster.cpp
guilpier-code: ok, but **cluster->flush()** still does nothing. According to me, this instruction should be removed and the definition : 
```
#ifndef ANTARES_SWAP_SUPPORT
inline void RenewableCluster::flush()
{
    // do nothing
}
#endif
```
should be removed as well, for renewable and as well as its thermal clusters counterpart.

src/libs/antares/study/parts/renewable/cluster.cpp
guilpier-code: ok, my mistake

src/libs/antares/study/study.cpp
```diff
@@ -1122,8 +1129,8 @@ bool Study::thermalClusterRename(ThermalCluster* cluster, ThermalClusterName new
     // Otherwise the values associated to the area will be lost.
     logs.info() << "  renaming thermal cluster '" << cluster->name() << "' into '" << newName
                 << "'";
-    //	ThermalClusterName oldId   = cluster->id();
```
guilpier-code, line 1125: Changes in commented lines, should not we simply remove these lines ?

src/libs/antares/study/parts/renewable/series.h
flomnes: Fixed in commits 949607a and bfb874d.

src/libs/antares/study/parts/common/series.h
```diff
@@ -0,0 +1,73 @@
...
+#include "../../fwd.h"
+
+namespace Antares
+{
+namespace Data
+{
+/*!
+** \brief Data series (Common)
+*/
+class DataSeriesCommon
```
guilpier-code, line 40: I would call this class DataSeriesCommonForClusters for instance, otherwise it could be about any TS

src/libs/antares/study/parts/common/series.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+using namespace Yuni;
+
+#define SEP IO::Separator
+
+namespace Antares
+{
+namespace Data
+{
+bool DataSeriesCommon::invalidate(bool reload) const
```
guilpier-code, line 43: Same remark as in the related *.h

src/libs/antares/study/parts/common/series.hxx
```diff
@@ -24,27 +24,27 @@
 **
 ** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
 */
-#ifndef __ANTARES_LIBS_STUDY_PARTS_THERMAL_TIMESERIES_HXX__
```
guilpier-code, line 27: Same remark as in the related *.h, except that, here, **__ANTARES_LIBS_STUDY_PARTS_COMMON_TIMESERIES_HXX__** should be named otherwise accordingly

src/libs/antares/study/parts/common/series.cpp
```diff
@@ -0,0 +1,75 @@
...
+    series.markAsModified();
+}
+
+void DataSeriesCommon::estimateMemoryUsage(StudyMemoryUsage& u, enum TimeSeries ts) const
+{
+    u.requiredMemoryForInput += sizeof(DataSeriesCommon);
+    timeseriesNumbers.estimateMemoryUsage(u, true, 1, u.years);
+    uint nbTimeSeries;
+    switch (ts) {
+    case timeSeriesThermal:
```
guilpier-code, line 60: As **estimateMemoryUsage** method differs depending on the kind of cluster we're on, we perhaps should make 2 data series classes (renewable and thermal), which both take a DataSeriesCommonCluster as ancester, but differ with this method only. If so, we woud not have the second argument and the switch in the body.
Further more, this file could contain specialized functions **DataSeriesSaveToFolder** and **DataSeriesLoadFromFolder**.
Is it worth it ?

src/libs/antares/study/parts/renewable/series.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+using namespace Yuni;
+
+#define SEP IO::Separator
+
+namespace Antares
+{
+namespace Data
+{
+int DataSeriesSaveToFolder(const DataSeriesCommon* t,
```
guilpier-code: See remark from **src/libs/antares/study/parts/common/series.cpp**

src/libs/antares/study/parts/renewable/series.h
```diff
@@ -0,0 +1,69 @@
...
+/*!
+** \brief Load renewable data series from a folder
+** \ingroup renewableseries
+**
+** \param t A renewable data series structure
+** \param ag A renewable cluster
+** \param folder The target folder
+** \return A non-zero value if the operation succeeded, 0 otherwise
+*/
+int DataSeriesLoadFromFolder(Study& s,
```
guilpier-code: See remark from **src/libs/antares/study/parts/common/series.cpp**

src/libs/antares/study/parts/thermal/series.cpp
```diff
@@ -40,9 +40,9 @@ namespace Antares
 {
 namespace Data
 {
-int DataSeriesThermalSaveToFolder(const DataSeriesThermal* t,
-                                  const ThermalCluster* ag,
-                                  const AnyString& folder)
+int DataSeriesSaveToFolder(const DataSeriesCommon* t,
```
guilpier-code: Same remark as above

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,805 @@
...
+    // assert
+    assert(parent and "A parent for a renewable dispatchable cluster can not be null");
+}
+
+Data::RenewableCluster::~RenewableCluster()
+{
+    delete series;
+}
+
+#ifdef ANTARES_SWAP_SUPPORT
```
guilpier-code: flush should be removed as well as its call

guilpier-code (changes requested):

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: Calling `flush` on a cluster (whatever the type) calls `flush` on the `series` and `timeseriesNumbers` objects. See **src/libs/antares/study/parts/common/series.hxx** :

```cpp
inline void DataSeriesCommon::flush()
{
#ifdef ANTARES_SWAP_SUPPORT
    series.flush();
    timeseriesNumbers.flush();
#endif
}
```

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: In case we might need it later ?

src/libs/antares/study/parts/common/series.h
flomnes: It may be extended to other modes as well (wind, solar, etc.)

src/libs/antares/study/parts/common/series.cpp
flomnes: Probably not

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: flush triggers some actions on the TS in the swap flavor. I insist it should be kept.

src/libs/antares/study/parts/renewable/container.cpp
flomnes: Re-written in pair programming.

src/libs/antares/study/parts/common/cluster.h
```diff
@@ -0,0 +1,143 @@
...
+    typedef std::set<Cluster*> SetPointer;
+    //! Map of renewable clusters
+    typedef std::map<ClusterName, Cluster*> Map;
+    //! Vector of renewable clusters
+    typedef std::vector<Data::Cluster*> Vector;
+
+public:
+    Cluster(Area* parent);
+
+    //! Get the thermal cluster ID
```
guilpier-code: some "thermal" (inside comments) left in this file

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -122,12 +120,12 @@ class ThermalCluster final : public Yuni::NonCopyable<ThermalCluster>
     /*!
     ** \brief Invalidate all data associated to the thermal cluster
     */
-    bool invalidate(bool reload) const;
```
guilpier-code, line 125: Few lines above, function **FlexibilityIsValid** is never used. to be removed

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -122,12 +120,12 @@ class ThermalCluster final : public Yuni::NonCopyable<ThermalCluster>
...
+    bool invalidate(bool reload) const override;
 
     /*!
     ** \brief Mark the thermal cluster as modified
     */
-    void markAsModified() const;
+    void markAsModified() const override;
 
     /*!
```
guilpier-code, line 123: few lines underneath, function **invalidateArea** should be moved to common cluster class

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -247,18 +231,6 @@ class ThermalCluster final : public Yuni::NonCopyable<ThermalCluster>
     */
```
guilpier-code, line 213: Above, function **isVisibleOnLayer** should be declared and defined in the cluster common part, isn't it ?

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -430,369 +387,9 @@ class ThermalCluster final : public Yuni::NonCopyable<ThermalCluster>
...
-** \param folder The target folder
-** \param msg Message to display
-** \return A non-zero value if the operation succeeded, 0 otherwise
-*/
-int ThermalClusterListSaveDataSeriesToFolder(const ThermalClusterList* l, const AnyString& folder);
-int ThermalClusterListSaveDataSeriesToFolder(const ThermalClusterList* l,
-                                             const AnyString& folder,
-                                             const YString& msg);
-
 struct CompareThermalClusterName final
```
guilpier-code: **CompareThermalClusterName** should be re-defined as **CompareClusterName** and moved to common cluster part

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,267 @@
...
+        groupID = renewableOther4;
+        return;
+    }
+    // assigning a default value
+    groupID = renewableOther1;
+}
+
+bool Data::RenewableCluster::invalidate(bool reload) const
+{
+    bool ret = true;
```
guilpier-code, line 171: why not invalidate modulation as well (like in thermal clusters) ?
`ret = modulation.invalidate(reload) and ret;`

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,267 @@
...
+    if (series)
+        ret = series->invalidate(reload) and ret;
+    return ret;
+}
+
+void Data::RenewableCluster::markAsModified() const
+{
+    if (series)
+        series->markAsModified();
```
guilpier-code, line 180: `modulation.markAsModified();` missing ?

src/libs/antares/study/parts/common/cluster.h
```diff
@@ -0,0 +1,143 @@
...
+    uint index;
+
+    //! Capacity of reference per unit (MW) (pMax)
+    double nominalCapacity;
+
+    //! The index of the thermal cluster from the area's point of view
+    //! \warning this variable is only valid when used by the solver
+    // (initialized in the same time that the runtime data)
+    uint areaWideIndex;
+
```
guilpier-code, line 118: **marginalCost** should be a common data, not only thermal cluster's

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -0,0 +1,267 @@
...
+}
+
+void Data::RenewableCluster::reset()
+{
```
guilpier-code, line 184: there should be a common part for reset, that is a **Cluster::reset()** and it should be called in both thermal and renewable cluster **reset()**

src/libs/antares/study/study.cpp
```diff
@@ -1071,7 +1078,7 @@ bool Study::areasThermalClustersMinStablePowerValidity(
     return resultat;
 }
 
-bool Study::thermalClusterRename(ThermalCluster* cluster, ThermalClusterName newName, bool)
+bool Study::thermalClusterRename(Cluster* cluster, ClusterName newName, bool)
```
guilpier-code, line 1081: - so the function should be renamed as **clusterRename**
- Why is this function part of the study class ? Shouldn't it be a method of common class for clusters ? Seems a bit complicated to move this method. To be checked.

src/libs/antares/study/parts/common/cluster.h
```diff
@@ -0,0 +1,143 @@
...
+{
+public:
+    //! Set of renewable clusters
+    typedef std::set<Cluster*, CompareClusterName> Set;
+    //! Set of renewable clusters (pointer)
+    typedef std::set<Cluster*> SetPointer;
+    //! Map of renewable clusters
+    typedef std::map<ClusterName, Cluster*> Map;
+    //! Vector of renewable clusters
+    typedef std::vector<Data::Cluster*> Vector;
```
guilpier-code: All these **typdef**s are useless, because unused.

src/libs/antares/CMakeLists.txt
```diff
@@ -195,22 +195,43 @@ set(SRC_STUDY_AREA_UI
 source_group("study\\area\\ui" FILES ${SRC_STUDY_AREA_UI})
 
 
+set(SRC_STUDY_PART_COMMON
+    study/parts/common/series.h
+	study/parts/common/series.hxx
+	study/parts/common/series.cpp
+    study/parts/common/cluster.cpp
+    study/parts/common/cluster.h
+    study/parts/common/cluster_list.h
```
guilpier-code, line 204: There is no **cluster_list.hxx** in **SRC_STUDY_PART_COMMON** ?

src/libs/antares/study/parts/common/cluster_list.hxx
```diff
@@ -0,0 +1,636 @@
+#ifndef __CLUSTER_LIST_COMMON_HXX__
```
guilpier-code: Not sure this file is useful, it seems never to be included (even in CMake files). Did I miss something ?
And the methods it defines are already defined in **cluster_list.cpp**

guilpier-code (changes requested):

src/libs/antares/study/parts/renewable/series.cpp
flomnes: May be used for solar, load, etc. in future revisions.

src/libs/antares/CMakeLists.txt
flomnes: no

src/libs/antares/study/parts/common/cluster_list.hxx
flomnes: removed

src/libs/antares/study/study.cpp
flomnes: For now, this function should be duplicated. It would be somewhat tricky to de-duplicate.

src/libs/antares/study/parts/thermal/cluster.h
flomnes: done

flomnes:
Local merge in PR #267

PR #250, opened 2021-05-10
Renewable output [3/3]

### LINKED ISSUE #220: Feature wind off-shore (or miscellaneus renewable)

flomnes:
Write renewable input data into the `output` directory 
* year by year
* synthesis
* record years

For each time granularity
* hourly
* daily
* weekly
* monthly
* annual


src/solver/variable/area.inc.hxx
```diff
@@ -241,6 +250,16 @@ void Areas<NEXTTYPE>::yearEndBuild(State& state, uint year, uint numSpace)
             // Variables
             variablesForArea.yearEndBuildForEachThermalCluster(state, year, numSpace);
         } // for each thermal cluster
+
+        // For each renewable cluster
+        for (uint j = 0; j != area.renewable.clusterCount; ++j)
+        {
+            state.renewableCluster = area.renewable.clusters[j];
+            state.yearEndReset();
```
guilpier-code: **yearEndReset** has already been called for thermal clusters. Calling it here again could set to zero some results computed for thermal clusters.

src/solver/variable/area.inc.hxx
```diff
@@ -229,7 +238,7 @@ void Areas<NEXTTYPE>::yearEndBuild(State& state, uint year, uint numSpace)
         // For each thermal cluster
         for (uint j = 0; j != area.thermal.clusterCount; ++j)
         {
-            state.cluster = area.thermal.clusters[j];
+            state.thermalCluster = area.thermal.clusters[j];
             state.yearEndReset();
 
             // Variables
```
guilpier-code, line 244: **yearEndBuildFromThermalClusterIndex** function name is very unclear on what the function really does.
This name should be changed, even if it has nothing to see with the current PR.

src/solver/variable/commons/links/links.hxx.inc.hxx
```diff
@@ -104,6 +104,21 @@ inline void Links::yearEndBuildPrepareDataForEachThermalCluster(State& state,
     }
 }
 
+inline void Links::yearEndBuildPrepareDataForEachRenewableCluster(State& state,
```
guilpier-code, line 107: This function is useless for now, but might be used someday (to be checked).

guilpier-code (changes requested):

src/solver/variable/commons/links/links.hxx.inc.hxx
flomnes: This function is used to recover hourly data for renewable clusters, see **src/solver/variable/economy/renewableGeneration.h:290**.
```cpp
void yearEndBuildPrepareDataForEachRenewableCluster(State& state,
                                                    uint year,
                                                    unsigned int numSpace)
{
    for (unsigned int i = 0; i <= state.study.runtime->rangeLimits.hour[Data::rangeEnd]; ++i)
    {
        state.renewableClusterProductionForYear[i]
          += pValuesForTheCurrentYear[numSpace][state.renewableCluster->areaWideIndex].hour[i];
    }
 // Next variable
    NextType::yearEndBuildPrepareDataForEachRenewableCluster(state, year, numSpace);
}
```

src/solver/variable/area.inc.hxx
flomnes: How about **yearEndRecoverThermalData** ?

src/solver/variable/area.inc.hxx
flomnes: I duplicated this function into two : one for thermal data, one for renewable data. See commit 9573759.

flomnes:
Local merge in PR #267

PR #251, opened 2021-05-10
Scenario builder for renewable [2/3]

### LINKED ISSUE #220: Feature wind off-shore (or miscellaneus renewable)

flomnes:
**DONE**: user-specified scenarii

**TODO**: configure intra/inter-modal TS selection


src/libs/antares/study/parameters.cpp
```diff
@@ -217,13 +219,15 @@ void Parameters::reset()
...
     nbTimeSeriesThermal = 1;
+    nbTimeSeriesRenewable = 1;
     // Time-series refresh
     timeSeriesToRefresh = 0; // None
     refreshIntervalLoad = 100;
     refreshIntervalSolar = 100;
     refreshIntervalHydro = 100;
     refreshIntervalWind = 100;
     refreshIntervalThermal = 100;
+    refreshIntervalRenewable = 100;
```
guilpier-code: **refreshIntervalRenewable** not needed : this parameter seems to be related only for stochastic TS, and renewable clusters don't have such time series

src/libs/antares/study/parameters.h
```diff
@@ -280,6 +282,8 @@ class Parameters final
     uint refreshIntervalWind;
     //! Refresh interval (in years) for timeSeries : Thermal
     uint refreshIntervalThermal;
+    //! Refresh interval (in years) for timeSeries : Renewable
```
guilpier-code: Same remark : **refreshIntervalRenewable** regards only stochastic TS, and renewable clusters don't have such TSs

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -187,6 +190,23 @@ static unsigned int CheckMatricesWidth(const Data::Study& study)
             }
             break;
         }
+        case Data::timeSeriesRenewable:
```
guilpier-code, line 193: Very strange function (I mean **CheckMatricesWidth**) : it runs a loop on areas (with a possible nested loop on clusters on thermal or renewable energy), gets the number of TSs for each, but seems to only use the last area / cluster.
From my point of view, something is absurd here.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -263,30 +286,34 @@ bool TimeSeriesNumbers::Generate(Data::Study& study)
...
+    uint32 draw[Data::timeSeriesCount] = {0, 0, 0, 0, 0, 0};
 
     const bool tsgen[Data::timeSeriesCount]
       = {0 != (Data::timeSeriesLoad & parameters.timeSeriesToRefresh),
          0 != (Data::timeSeriesHydro & parameters.timeSeriesToRefresh),
          0 != (Data::timeSeriesWind & parameters.timeSeriesToRefresh),
          0 != (Data::timeSeriesThermal & parameters.timeSeriesToRefresh),
-         0 != (Data::timeSeriesSolar & parameters.timeSeriesToRefresh)};
+         0 != (Data::timeSeriesSolar & parameters.timeSeriesToRefresh),
+         0 != (Data::timeSeriesRenewable & parameters.timeSeriesToRefresh)};
```
guilpier-code: ok, last element of **tsgen** will always be **false**, but I can't a way to do something else than that.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -316,20 +343,42 @@ bool TimeSeriesNumbers::Generate(Data::Study& study)
...
+                          = DRAW_A_RANDOM_NUMBER(Data::timeSeriesThermal,
+                                                 cluster->series->series,
+                                                 parameters.nbTimeSeriesThermal);
+                    }
                 }
-                else
+            }
+            // Renewables
+            {
+                auto end = area.renewable.list.mapping.end();
```
guilpier-code: **mapping** will disappear as a member of a list of renewable clusters

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -316,20 +343,42 @@ bool TimeSeriesNumbers::Generate(Data::Study& study)
...
-                                             parameters.nbTimeSeriesThermal);
+                    auto* cluster = i->second;
+                    if (!cluster->enabled)
+                    {
+                        study.runtime->random[Data::seedTimeseriesNumbers].next();
+                    }
+                    else
+                    {
+                        cluster->series->timeseriesNumbers.entry[0][y]
+                          = DRAW_A_RANDOM_NUMBER(Data::timeSeriesRenewable,
```
guilpier-code, line 378: This should **not** change the reference results because only the TS's seed is changed, and after the draws for load to thermal.

src/libs/antares/study/parameters.h
```diff
@@ -258,6 +258,8 @@ class Parameters final
     uint nbTimeSeriesWind;
     //! Nb of timeSeries : Thermal
     uint nbTimeSeriesThermal;
+    //! Nb of timeSeries : Renewable
+    uint nbTimeSeriesRenewable;
```
guilpier-code: We have a problem with **nbTimeSeriesRenewable** here : in the current state of the code, it will always be 1, as it is initialized to 1 and never set. Indeed, its thermal counterpart is the number of generated TS, and renewable clusters should not be concerned with this.
The problem is that this parameter is used elsewhere, for instance in the TS numbers draws.
A closer look should be given to this.
And a question is : if thermal has ready made TS, how does the TS number are drawn ?

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -316,20 +343,42 @@ bool TimeSeriesNumbers::Generate(Data::Study& study)
...
+                                                 cluster->series->series,
+                                                 parameters.nbTimeSeriesRenewable);
+                    }
                 }
             }
         });
```
guilpier-code, line 384: A bit further in the code from here, why don't we have a call of **BUILD_LOG_ENTRY** for reneawbles ?

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -405,6 +466,12 @@ bool TimeSeriesNumbers::Generate(Data::Study& study)
                     tsNumbers = &(area.wind.series->timeseriesNumbers);
                 else if (intermodal[TS_INDEX(Data::timeSeriesHydro)])
                     tsNumbers = &(area.hydro.series->timeseriesNumbers);
+                else if (intermodal[TS_INDEX(Data::timeSeriesThermal)]
```
guilpier-code, line 470: Why do we need to add a test on **intermodal** parameter for thermal cluster ? The code is supposed to work fine without it

guilpier-code (changes requested):

src/solver/simulation/timeseries-numbers.cpp
flomnes: I disagree. From the 2nd iteration, random generation structure `study.runtime->random[Data::seedTimeseriesNumbers]` may have a different state **if at least one renewable cluster is disabled**, from these lines :
```cpp
                auto end = area.renewable.list.mapping.end();
                for (auto i = area.renewable.list.mapping.begin(); i != end; ++i)
                {
                    auto* cluster = i->second;
                    if (!cluster->enabled)
                    {
                        study.runtime->random[Data::seedTimeseriesNumbers].next();
                    }
```

On the other hand, `study.runtime->random[Data::timeSeriesRenewable]` is unrelated and has no effect on existing random number draws (load, etc.). On that I agree.

src/solver/simulation/timeseries-numbers.cpp
flomnes: Without lines 467-470, when only thermal & renewable intemodal correlation is enabled, `tsNumbers = nullptr` and the assertion at l. 476 fails.

src/solver/simulation/timeseries-numbers.cpp
flomnes: Done

src/solver/simulation/timeseries-numbers.cpp
flomnes: Fixed

src/libs/antares/study/parameters.cpp
flomnes: Fixed

src/libs/antares/study/parameters.h
flomnes: Fixed

src/solver/simulation/timeseries-numbers.cpp
flomnes: I've run a few mental experiments on some width lists, here are the results.

Assume `TimeSeriesT = Data::timeSeriesLoad`. Very similar for clusters (thermal or renewable), the per-cluster lists are just concatenated into a big one, for which we can use the same reasoning as here.

### W=[1,1,1,1]
Four areas, each with a matrix of width 1. The function returns `w=1`.
### W=[2,2,2,2]
Four areas, each with a matrix of width 2. The function returns `w=2` after all iterations.
### W=[1,2,3,4]
Four areas, with matrices of width 1, 2, 3, 4 respectively. At the 3rd iteration, return 0.
### W=[4,3,2,1]
Four areas, with matrices of width 4, 3, 2, 1 respectively. At the 2nd iteration, return 0.

In conclusion, the function returns 0 whenever two width are different and different than 1. Otherwise, return the common value.

0 is an error value :
```cpp
nbTimeseries = CheckMatricesWidth<TimeSeriesT>(study);
if (!nbTimeseries)
{
    logs.error() << "Intra-modal correlation: Constraint violation: The number of time-series "
                    "must be identical for all areas";
    return false;
}
```

flomnes: More investigations. I wrote a small Python function to reproduce the behavior
```python3
def f(L):
    w = 1
    for p in L:
        if p != w:
            if p != 1 and w != 1:
                return 0
            if p > 1:
                w = p
    return w
```
For the testing part :
```
f([1, 1, 1, 1]) = 1
f([2, 2, 2, 2]) = 2
f([1, 2, 3, 4]) = 0
f([4, 3, 2, 1]) = 0
f([0, 0]) = 1
f([0, 1, 0]) = 1
f([0, 1, 2]) = 2
f([0, 2, 1]) = 2
f([2, 1, 0]) = 0 # PROBLEM : a permutation should not affect the result
```
There is a problem (see comment) on permutations.

Also, the function returns 1 when all matrices are empty (width=0), which is expected.

src/libs/antares/study/parameters.h
flomnes: >  And a question is : if thermal has ready made TS, how does the TS number are drawn ?

Here is the code when expanding the `DRAW_A_RANDOM_NUMBER` macro.

```cpp
if (intramodal[TS_INDEX(T)]) {
  cluster->series->timeseriesNumbers.entry[0][y] = draw[TS_INDEX(T)];
 } else {
  if (tsgen[TS_INDEX(T)]) {
    cluster->series->timeseriesNumbers.entry[0][y] = (uint32)(floor((double)(study.runtime->random[Data::seedTimeseriesNumbers].next()) * parameters.nbTimeSeriesThermal));
  } else {
    cluster->series->timeseriesNumbers.entry[0][y] = (uint32)(floor((double)(study.runtime->random[Data::seedTimeseriesNumbers].next()) * cluster->series->series.width));
  }
 }
```
When `tsgen[TS_INDEX(T)]=false`, TS generation is disabled, the ready-made series width is used.

I agree that since no TS generation is possible for renewable clusters, we should **always** use `cluster->series->series.width`.

flomnes: Note that
```cpp
if (intramodal[z])
{
    draw[z] = (uint32)(floor(study.runtime->random[Data::seedTimeseriesNumbers].next()
                             * nbTimeseries[z]));
}
```
where `nbTimeseries[z]` is obtained with function `CheckMatricesWidth` : 
- N if all matrices are of width N >= 1
- 1 if all matrices are of width 0
- 0 in error cases (inconsistent width)

src/solver/simulation/timeseries-numbers.cpp
flomnes: More on this : adding a renewable cluster changes the random generation process, even if this cluster is within an area that is not connected to the network.

But this behavior is the same with thermal clusters. And to me (and @sgatto) this is not a problem.

guilpier-code: ok, so we should expect a change in the results (compared to current reference) after the first year of simulation with the addition of renewable clusters

guilpier-code (approved):

flomnes:
Local merge in PR #267

PR #259, opened 2021-06-07
Renewable clusters main panel : adding modulation tab

guilpier-code:


src/ui/simulator/toolbox/components/datagrid/renderer/area/renewable-modulation.cpp
```diff
@@ -0,0 +1,227 @@
...
+        case Data::renewableModulationCapacity:
+            return (*pMatrix)[Data::renewableModulationCapacity][y];
+        case Data::renewableMinGenModulation:
+            return (*pMatrix)[Data::renewableMinGenModulation][y];
+        case (Data::renewableModulationCost + Data::renewableModulationMax):
+            // gp : "5" must be replaced with pCluster->unitCount when this last var is defined
+            return (*pMatrix)[Data::renewableModulationCapacity][y] * 5 * pCluster->nominalCapacity;
+        case (Data::renewableModulationCapacity + Data::renewableModulationMax):
+            // gp : "5" must be replaced with pCluster->unitCount when this last var is defined
+            return (*pMatrix)[Data::renewableMinGenModulation][y] * 5 * pCluster->nominalCapacity;
```
flomnes, line 124: Don't forget that. On branch **feature/renewable**, unitCount has been introduced. I suggest you merge this branch into yours and use it.

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -408,6 +408,20 @@ bool RenewableClusterListSaveToFolder(const RenewableClusterList* l, const AnySt
...
+            if (IO::Directory::Create(buffer))
+            {
+                buffer.clear() << folder << SEP << ".." << SEP << ".." << SEP << "prepro" << SEP
+                    << cluster.parentArea->id << SEP << cluster.id() << SEP
+                    << "modulation.txt";
+
+                ret = cluster.modulation.saveToCSVFile(buffer) and ret;
+            }
+            else
+                ret = 0;
```
flomnes, line 424: Move this to new function `RenewableClusterList::saveToFolder`

src/ui/simulator/toolbox/components/datagrid/renderer/area/renewable-modulation.h
```diff
@@ -0,0 +1,113 @@
...
+    }
+    virtual uint maxHeightResize() const
+    {
+        return 0;
+    }
+
+    virtual Date::Precision precision()
+    {
+        return Date::hourly;
+    }
```
flomnes, line 97: make sure to specify `override` so that the compiler will check for you that the derived-class method overrides the base-class method.

src/ui/simulator/toolbox/components/datagrid/renderer/area/renewable-modulation.h
```diff
@@ -0,0 +1,113 @@
...
+
+protected:
+    virtual void internalRenewableClusterChanged(Antares::Data::RenewableCluster* cluster);
+    virtual void onStudyClosed();
+
+private:
+    Antares::Data::RenewableCluster* pCluster;
```
flomnes, line 104: Maybe we could use the generic type `Antares::Data::Cluster*` here, and share most of the functions here with the thermal modulation class.

We would have a common base class with all common functions, and some specializations in 2 derived classes.

flomnes (commented):
The path to folder **../../prepro** is constructed 3 times. Consider writing a helper function to avoid code duplication.

PR #264, opened 2021-06-09
GUI & renewables - time series management

guilpier-code:
This PR concerns the **GUI > Simulation > Time-Series Management** grid.
A switch in **Configure > Advanced parameters > Other Preferences** must be added.
This switch allows user to choose if she wants to use old renewables (wind and Solar), or the renewable clusters.
In the second case,  **GUI > Simulation > Time-Series Management** grid is changed : Wind and Solar columns desappear and a
**Renewable** column appears.


flomnes (commented):
TODO :
- Take the parameter into account in **SIM**, when filling structures with TS values.
- Disable TS generation for wind and solar
- Disable scenario building for wind and solar
- Disable intra/intermodal TS number picking (see timeseries-numbers.cpp)

PR #269, opened 2021-06-11
Fix thermal cluster segfault

flomnes:
Cluster object shouldn't be `operator delete`d when removing disabled
clusters.

They're still used in the `area.thermal.list.mapping` object.

For what it's worth, this is the valgrind trace that lead me to this
conclusion :

```
[Fri Jun 11 20:02:38 2021][solver][infos] Preparing time-series numbers...
==489011== Invalid read of size 1
==489011==    at 0xFDAA39: Antares::Solver::TimeSeriesNumbers::Generate(Antares::Data::Study&)::{lambda(Antares::Data::Area&)#1}::operator()(Antares::Data::Area&) const (timeseries-numbers.cpp:353)
==489011==    by 0xFDCE67: void Antares::Data::AreaList::each<Antares::Solver::TimeSeriesNumbers::Generate(Antares::Data::Study&)::{lambda(Antares::Data::Area&)#1}>(Antares::Solver::TimeSeriesNumbers::Generate(Antares::Data::Study&)::{lambda(Antares::Data::Area&)#1} const&) (area.hxx:131)
==489011==    by 0xFDB316: Antares::Solver::TimeSeriesNumbers::Generate(Antares::Data::Study&) (timeseries-numbers.cpp:329)
==489011==    by 0xD7B08C: Antares::Solver::Simulation::ISimulation<Antares::Solver::Simulation::Economy>::run() (solver.hxx:361)
==489011==    by 0xD79BAC: void SolverApplication::runSimulation<Antares::Solver::Simulation::ISimulation<Antares::Solver::Simulation::Economy> >() (main.hxx:37)
==489011==    by 0xD75C8F: SolverApplication::runSimulationInEconomicMode() (economy.cpp:40)
==489011==    by 0xD5E9AA: SolverApplication::execute() (main.cpp:339)
==489011==    by 0xD5FA09: main (main.cpp:592)
==489011==  Address 0x6011d4c is 12 bytes inside a block of size 768 free'd
==489011==    at 0x483CFBF: operator delete(void*) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==489011==    by 0x1114853: Antares::Data::ThermalClusterList::remove(Yuni::CString<128u, false> const&) (cluster_list.cpp:725)
==489011==    by 0x110A79B: Antares::Data::PartThermal::removeDisabledClusters() (container.cpp:169)
==489011==    by 0x103BCF1: Antares::Data::StudyRuntimeInfos::removeDisabledThermalClustersFromSolverComputations(Antares::Data::Study&)::{lambda(Antares::Data::Area&)#1}::operator()(Antares::Data::Area&) const (runtime.cpp:635)
==489011==    by 0x103D2C8: std::_Function_handler<unsigned int (Antares::Data::Area&), Antares::Data::StudyRuntimeInfos::removeDisabledThermalClustersFromSolverComputations(Antares::Data::Study&)::{lambda(Antares::Data::Area&)#1}>::_M_invoke(std::_Any_data const&, Antares::Data::Area&) (std_function.h:285)
==489011==    by 0x103DB86: std::function<unsigned int (Antares::Data::Area&)>::operator()(Antares::Data::Area&) const (std_function.h:688)
==489011==    by 0x103BB76: Antares::Data::removeDisabledClusters(Antares::Data::Study&, char const*, std::function<unsigned int (Antares::Data::Area&)>) (runtime.cpp:618)
==489011==    by 0x103BD3A: Antares::Data::StudyRuntimeInfos::removeDisabledThermalClustersFromSolverComputations(Antares::Data::Study&) (runtime.cpp:634)
==489011==    by 0x103B33F: Antares::Data::StudyRuntimeInfos::loadFromStudy(Antares::Data::Study&) (runtime.cpp:493)
==489011==    by 0x1045088: Antares::Data::Study::initializeRuntimeInfos() (study.cpp:677)
==489011==    by 0xD5F498: SolverApplication::readDataForTheStudy(Antares::Data::StudyLoadOptions&) (main.cpp:502)
==489011==    by 0xD5D966: SolverApplication::prepare(int, char**) (main.cpp:143)
==489011==  Block was alloc'd at
==489011==    at 0x483BE63: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==489011==    by 0x1111883: Antares::Data::ThermalClusterList::loadFromFolder(Antares::Data::Study&, Yuni::CString<0u, true> const&, Antares::Data::Area*) (cluster_list.cpp:99)
==489011==    by 0x109AC5E: Antares::Data::AreaList::loadFromFolder(Antares::Data::StudyLoadOptions const&) (list.cpp:1237)
==489011==    by 0x105DCB8: Antares::Data::Study::internalLoadFromFolder(Yuni::CString<128u, true> const&, Antares::Data::StudyLoadOptions const&) (load.cpp:177)
==489011==    by 0x105D640: Antares::Data::Study::loadFromFolder(Yuni::CString<0u, true> const&, Antares::Data::StudyLoadOptions const&) (load.cpp:65)
==489011==    by 0xD5EF30: SolverApplication::readDataForTheStudy(Antares::Data::StudyLoadOptions&) (main.cpp:415)
==489011==    by 0xD5D966: SolverApplication::prepare(int, char**) (main.cpp:143)
==489011==    by 0xD5F9F9: main (main.cpp:591)
==489011==
```

### DIFF
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -463,9 +463,6 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
     // Invalidating the parent area
     c->parentArea->invalidate();
 
-    // delete the cluster
-    delete c;
-
     // Rebuilding the index
     rebuildIndex();
     return true;
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -222,7 +222,7 @@ void ClusterList<ClusterT>::rebuildIndex()
 }
 
 template<class ClusterT>
-typename std::shared_ptr<ClusterT> ClusterList<ClusterT>::add(ClusterT* newcluster)
+typename ClusterList<ClusterT>::SharedPtr ClusterList<ClusterT>::add(ClusterT* newcluster)
 {
     if (newcluster)
     {
@@ -230,7 +230,7 @@ typename std::shared_ptr<ClusterT> ClusterList<ClusterT>::add(ClusterT* newclust
             return cluster[newcluster->id()];
 
         newcluster->index = (uint)size();
-        cluster[newcluster->id()] = std::shared_ptr<ClusterT>(newcluster);
+        cluster[newcluster->id()] = SharedPtr(newcluster);
         ++(groupCount[newcluster->groupId()]);
         rebuildIndex();
         return cluster[newcluster->id()];
@@ -378,7 +378,7 @@ bool ClusterList<ClusterT>::rename(Data::ClusterName idToFind, Data::ClusterName
     if (it == cluster.end())
         return true;
 
-    std::shared_ptr<ClusterT> p = it->second;
+    SharedPtr p = it->second;
 
     if (idToFind == newID)
     {
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -88,7 +88,7 @@ ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id)
 template<class ClusterT>
 ClusterT* ClusterList<ClusterT>::detach(iterator i)
 {
-    auto c = i->second;
+    SharedPtr c = i->second;
     cluster.erase(i);
     return c.get();
 }
@@ -456,7 +456,7 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
         return false;
 
     // Getting the pointer on the cluster
-    auto c = i->second;
+    SharedPtr c = i->second;
 
     // Removing it from the list
     cluster.erase(i);
@@ -579,9 +579,9 @@ void ClusterList<ClusterT>::ensureDataTimeSeries()
     auto end = cluster.end();
     for (auto it = cluster.begin(); it != end; ++it)
     {
-        auto& cluster = *(it->second);
-        if (not cluster.series)
-            cluster.series = new DataSeriesCommon();
+        SharedPtr cluster = it->second;
+        if (not cluster->series)
+            cluster->series = new DataSeriesCommon();
     }
 }
 
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -75,22 +75,22 @@ template<class ClusterT>
 const ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
 {
     auto i = cluster.find(id);
-    return (i != cluster.end()) ? i->second : nullptr;
+    return (i != cluster.end()) ? i->second.get() : nullptr;
 }
 
 template<class ClusterT>
 ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id)
 {
     auto i = cluster.find(id);
-    return (i != cluster.end()) ? i->second : nullptr;
+    return (i != cluster.end()) ? i->second.get() : nullptr;
 }
 
 template<class ClusterT>
 ClusterT* ClusterList<ClusterT>::detach(iterator i)
 {
-    auto* c = i->second;
+    auto c = i->second;
     cluster.erase(i);
-    return c;
+    return c.get();
 }
 
 template<class ClusterT>
@@ -141,8 +141,8 @@ const ClusterT* ClusterList<ClusterT>::find(const ClusterT* p) const
     auto end = cluster.end();
     for (auto i = cluster.begin(); i != end; ++i)
     {
-        if (p == i->second)
-            return i->second;
+        if (p == i->second.get())
+            return i->second.get();
     }
     return nullptr;
 }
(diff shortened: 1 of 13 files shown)


src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -463,9 +463,6 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
     // Invalidating the parent area
     c->parentArea->invalidate();
```
guilpier-code, line 465: This method remove(id) is modified (si its behavior is modified) , but it is used elsewhere in the code, for example in : 
**src\ui\simulator\toolbox\input\thermal-cluster.cpp**
So this change could have nasty side effect
Maybe we should create a new method remove : 
```
bool ClusterList<ClusterT>::remove(const Data::ThermalClusterName& id)
{
    auto i = cluster.find(id);
    if (i == cluster.end())
        return false;
    // Removing it from the list
    cluster.erase(i);
}
```

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -721,9 +721,6 @@ bool ThermalClusterList::remove(const ClusterName& id)
         }
     }
 
-    // delete the cluster
```
guilpier-code, line 724: Same remark as above

guilpier-code (changes requested):

src/libs/antares/study/parts/common/cluster_list.cpp
flomnes: There are two calls for `remove`. One is in `removeDisabledClusters`, the other is where you mentioned.

The only side-effect is that some memory could still be accessible after closing the program (memory leak). So here is my suggestion :
- Do not free allocated memory in method `remove`
- Free everything in the destructor `~ClusterList`.

guilpier-code: I have a problem here : as I said, **remove** is called when deleting a thermal cluster from the UI (when you click on a cluster and ask to delete it). For me, the cluster should be deleted permanently at this stage, otherwise, it could be remaining when saving the study and be found in files on disk.

guilpier-code: correction to the previous remark : the deleted cluster will NOT be remaining in the list, so we should not have problem when saving the study, but we will have a memory leak during the user session.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -582,9 +579,9 @@ void ClusterList<ClusterT>::ensureDataTimeSeries()
     auto end = cluster.end();
     for (auto it = cluster.begin(); it != end; ++it)
     {
-        auto& cluster = *(it->second);
-        if (not cluster.series)
-            cluster.series = new DataSeriesCommon();
+        SharedPtr cluster = it->second;
```
guilpier-code, line 582: shared_ptr has a dereference operator *, so there should be no need to change this function. Is there a reason ?

src/libs/antares/study/parts/common/cluster_list.cpp
flomnes: There are two maps for thermal clusters
* `cluster`, defined in `common/cluster_list.h`
* mapping, defined in `renewable/cluster_list.h`
These maps contain copies of the same shared pointers. When the user adds a thermal cluster, the two following lines are executed
```cpp
auto added = pArea->thermal.list.add(cluster);
pArea->thermal.list.mapping[cluster->id()] = added;
```
In other words, both `mapping` and `cluster` contain the same pointer. There are **2** references to the pointed data.

When the user deletes a thermal cluster, the following line is executed
```cpp
pArea->thermal.list.remove(toDelete->id())
```
The remove method is `ThermalClusterList::remove()` :
```cpp
bool ThermalClusterList::remove(const ClusterName& id)
{
    auto i = cluster.find(id);
    if (i == cluster.end())
        return false;

    // Getting the pointer on the cluster
    SharedPtr c = i->second;

    // Removing it from the list
    cluster.erase(i);
    // Invalidating the parent area
    c->parentArea->invalidate();

    // Remove all cluster coupling
    if (not c->coupling.empty())
    {
        auto end = c->coupling.end();
        for (auto j = c->coupling.begin(); j != end; ++j)
        {
            auto* link = *j;
            link->parentArea->invalidate();
            link->coupling.erase(c.get());
        }
    }

    // Rebuilding the index
    rebuildIndex();
    return true;
}
```
Only **1** reference remains, from the shared pointer in `mapping`. However, when the user closes the GUI, the destructor of this structure is called, which calls the destructors of the `shared_ptr`, so there is **0** reference left. Therefore, the `ThermalCluster` pointer is automatically deallocated.

I agree that the user could fill the memory by creating/deleting a cluster several thousands of times in the GUI. This is a partial memory leak : the memory is only deallocated at the end.

src/libs/antares/study/parts/common/cluster_list.cpp
flomnes: While type inference (`auto`, `decltype`, etc.) is convenient, it could slow down compilation and make code harder to read. I'd rather avoid using it too often.

guilpier-code (approved):

PR #273, opened 2021-06-17
Digits for renewable clusters' time series

guilpier-code:
Because the **matrix of data series** cannot be of same type for thermal and renewable clusters, we have to split it in 2 specialized parts.
Actually, there is a minimum common part that remains.
This split has consequences in cluster and cluster lists classes of both types (thermal and renewable).


src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -8,6 +8,19 @@
...
+using namespace Yuni;
+
+namespace // anonymous
+{
+    struct TSNumbersPredicate
+    {
+        uint32 operator()(uint32 value) const
+        {
+            return value + 1;
+        }
```
flomnes: Defining functions in header files usually involves multiple definition issues at link time.

One solution is to use the `inline` keyword to tell the compiler not to create a symbol.

A better solution is to place function declarations in headers and function definitions in source (.cpp) files.

src/libs/antares/study/parts/renewable/cluster.h
```diff
@@ -167,10 +178,17 @@ class RenewableCluster final : public Cluster
     bool checkMinStablePowerWithNewModulation(uint index, double value);
     //@}
 
-    // TODO remove
-    void* prepro;
+    bool setTimeSeriesModeFromString(const YString& value);
+    int saveDataSeriesToFolder(const AnyString& folder) const;
+    int loadDataSeriesFromFolder(Study& s, const AnyString& folder);
```
flomnes: Mark overriding methods as `override`

flomnes (changes requested):

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -712,6 +700,111 @@ void ThermalClusterList::enableMustrunForEveryone()
...
+
+    each([&](const ThermalCluster& cluster) {
+        study.buffer = study.folderOutput;
+        study.buffer << SEP << "ts-numbers" << SEP << typeID() << SEP << cluster.parentArea->id
+            << SEP << cluster.id() << ".txt";
+        ret = cluster.series->timeseriesNumbers.saveToCSVFile(study.buffer, 0, true, predicate)
+            and ret;
+    });
+    return ret;
+}
```
flomnes: this method can be placed in ClusterList, `typeID()`, `cluster` all make sense there.

flomnes (changes requested):
A lot of unnecessary code re-duplication

flomnes:
Instead of re-duplicating code, I suggest introducing a pure virtual function for each cluster type
```cpp
virtual unsigned int precision() = 0;
```
This function would return 0 in class `ThermalClusterList` and e.g 4 in `RenewableClusterList`.

flomnes (approved):

PR #278, opened 2021-07-05
Add thermal groups other 2, 3, 4

### LINKED ISSUE #277: Create Thermal_1,...,4 groups

flomnes:
TODO : Change base branch when **feature/renewable** is merged.


src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -336,11 +336,31 @@ void Data::ThermalCluster::setGroup(Data::ClusterName newgrp)
             groupID = thermalDispatchGrpOil;
             return;
         }
+        else if (newgrp == "other")
```
guilpier-code: I understand that `newgrp == "other"` was left for compatibility with older study version.
But `"other 1"` is also a possible value for **newgrp**, and it's confusing that it does not appear in the switch...if...else block, and is set with default value.
I suggest : 
- `"other 1"` is also an item in the switch block.
- Regarding **other**, a comment should say that it is left for compatibility with older version.
- Regarding default value, it is historically other, I do not see why it should change.

We could also use **other** for **newgrp**, not **other 1**.
To be discussed.

guilpier-code (changes requested):

src/libs/antares/study/parts/thermal/cluster.cpp
flomnes: > "other 1" is also an item in the switch block.

OK
> Regarding other, a comment should say that it is left for compatibility with older version.

I recommend leaving value **other** (as opposed to **other 1**) in files for compatibility. But accepting **other 1** (in the switch...if...else block) if the user edits the file with a script or text editor.

> Regarding default value, it is historically other, I do not see why it should change.

The default value is `thermalDispatchGrpOther1`, which corresponds to **other**.

PR #287, opened 2021-07-09
De-duplicate clusterRename

flomnes:
From `thermalClusterRename` and `renewableClusterRename`. Use enum value.

### DIFF
--- a/src/libs/antares/study/study.cpp
+++ b/src/libs/antares/study/study.cpp
@@ -1078,8 +1078,9 @@ bool Study::areasThermalClustersMinStablePowerValidity(
     return resultat;
 }
 
-bool Study::thermalClusterRename(Cluster* cluster, ClusterName newName, bool)
+bool Study::clusterRename(Cluster* cluster, ClusterName newName, ClusterType clusterType)
 {
+    assert((clusterType == kThermal or clusterType == kRenewable) && "Wrong cluster type");
     // A name must not be empty
     if (!cluster or !newName)
         return false;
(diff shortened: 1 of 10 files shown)


flomnes:
I'm not completely satisfied with this PR. I may internalize the cluster type : each cluster will know its type (thermal or renewable) at construction, so it won't be necessary to provide it to the `clusterRename` function. Implementation : pure virtual function returning an enum value.

@guilpier-code do you agree ?

src/libs/antares/study/study.cpp
```diff
@@ -1102,100 +1102,27 @@ bool Study::thermalClusterRename(Cluster* cluster, ClusterName newName, bool)
...
+    Cluster* found = nullptr;
 
-    bool ret = true;
+    if (dynamic_cast<ThermalCluster*>(cluster))
+        found = area.thermal.list.find(newID);
+    else if (dynamic_cast<RenewableCluster*>(cluster))
+        found = area.renewable.list.find(newID);
 
-    // Archiving data
+    if (found)
```
flomnes: This is how we find the type of a cluster dynamically (essence of this PR)

src/libs/antares/study/study.cpp
```diff
@@ -1078,7 +1078,7 @@ bool Study::areasThermalClustersMinStablePowerValidity(
     return resultat;
 }
 
-bool Study::thermalClusterRename(Cluster* cluster, ClusterName newName, bool)
+bool Study::clusterRename(Cluster* cluster, ClusterName newName)
```
guilpier-code, line 1081: I know this would certainly be tricky but : function **clusterRename** should really not be here (I mean inside Study class).
Renaming a cluster should be done by a **clusterRename** function in common/ClusterList, that would contain calls to thermal or renewable cluster list function for some light parts.
Moreover, this function does more than just renaming a cluster (somehow udating scenario builder, reloading binding constraints, ...),  wouldn't it be split ?

src/libs/antares/study/study.cpp
```diff
@@ -1102,100 +1102,27 @@ bool Study::thermalClusterRename(Cluster* cluster, ClusterName newName, bool)
...
-    logs.info() << "  renaming thermal cluster '" << cluster->name() << "' into '" << newName
-                << "'";
-    //	ClusterName oldId   = cluster->id();
-    //	ClusterName oldName = cluster->name();
-    //	cluster->name(newName);
-    area.invalidateJIT = true;
+    Cluster* found = nullptr;
 
-    bool ret = true;
+    if (dynamic_cast<ThermalCluster*>(cluster))
```
guilpier-code, line 1118: If **clusterRename** was not coded in Study class, we wouldn't have to make those ugly **dynamic_cast** here.

guilpier-code (approved):

guilpier-code:
I made some comments, knowing that taking them into account would lead to rather tricky refactoring.
So they can be ignored for this PR.

src/libs/antares/study/study.cpp
flomnes: Using [RTTI](https://en.wikipedia.org/wiki/Run-time_type_information) is OK. But I agree with you : this function belongs in the cluster_list class.

Also, it would be possible replace the use of dynamic_cast with virtual functions.

PR #295, opened 2021-07-13
UI : Add installed capacity for renewable clusters

flomnes:
Main panel & inspector

### DIFF
--- a/src/ui/simulator/windows/inspector/accumulator.hxx
+++ b/src/ui/simulator/windows/inspector/accumulator.hxx
@@ -947,12 +947,12 @@ struct PAreaSpilledEnergyCost
 };
 
 // ----------------
-// THERMAL CLUSTERS
+// THERMAL/RENEWABLE CLUSTERS
 // ----------------
 struct PClusterEnabled
 {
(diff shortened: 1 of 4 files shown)


flomnes (commented):
Some more refactor is required to mutualise some structures.

src/ui/simulator/windows/inspector/accumulator.hxx
```diff
@@ -947,12 +947,12 @@ struct PAreaSpilledEnergyCost
 };
 
 // ----------------
-// THERMAL CLUSTERS
+// THERMAL/RENEWABLE CLUSTERS
```
guilpier-code, line 950: ok, but on line 991 starts the **THERMAL CLUSTERS** block, we should have a commented title on that line

guilpier-code (changes requested):

src/ui/simulator/windows/inspector/accumulator.hxx
flomnes: Done. I created 2 new sections : one for renewables, one for binding constraints.

PR #296, opened 2021-07-13
UI : allowing cloning renewable clusters

guilpier-code:
In renewable main panel, allowing to clone clusters


src/ui/simulator/toolbox/input/renewable-cluster.cpp
```diff
@@ -287,8 +287,7 @@ void RenewableCluster::internalDeletePlant(void*)
 
         // gp : uncomment this when GUI's scenario builder comes up 
         // We have to rebuild the scenario builder data, if required
```
flomnes, line 289: Remove this comment

flomnes (approved):
It would be nice to de-duplicate renewable/thermal classes. TODO later.

PR #299, opened 2021-07-15
UI : allowing to delete all renewable clusters of an area

guilpier-code:


src/ui/simulator/toolbox/input/renewable-cluster.cpp
```diff
@@ -404,11 +385,8 @@ void RenewableCluster::internalAddPlant(void*)
             sFl.clear() << "new cluster " << indx;
         }
 
-        // gp : to be adapted
-        /*
         // We have to rebuild the scenario builder data, if required
         ScenarioBuilderUpdater updaterSB(*study);
-        */
```
flomnes, line 411: Why uncomment these lines in this PR ? This PR is about deleting clusters, not creating new ones, right ?

PR #312, opened 2021-07-20
UI : hide or show tabs depending on renewable modeling parameter value

guilpier-code:
- On the left part of main window, Wind and Solar tabs must be hidden if we use renewable clusters.
Conversely, Renewable tab must be hidden if we use Wind and Solar as renewables.
- Same thing regarding the sceanrio builder pane

### DIFF
--- a/src/ui/simulator/application/main/main.cpp
+++ b/src/ui/simulator/application/main/main.cpp
@@ -781,15 +781,27 @@ void ApplWnd::onRenewableGenerationModellingChanged()
 
     if (study->parameters.renewableGeneration.rgModelling == Antares::Data::rgAggregated)
     {
+        // Main window
         pNotebook->show_page(wxString("wind"));
         pNotebook->show_page(wxString("solar"));
         pNotebook->hide_page(wxString("renewable"));
+
+        // Scenario builder pane
+        pScenarioBuilderNotebook->show_page(wxString("wind"));
+        pScenarioBuilderNotebook->show_page(wxString("solar"));
+        pScenarioBuilderNotebook->hide_page(wxString("renewable"));
     }
     else
     {
+        // Main window
         pNotebook->hide_page(wxString("wind"));
         pNotebook->hide_page(wxString("solar"));
         pNotebook->show_page(wxString("renewable"));
+
+        // Scenario builder pane
+        pScenarioBuilderNotebook->hide_page(wxString("wind"));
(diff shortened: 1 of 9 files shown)


src/ui/simulator/application/main/main.cpp
```diff
@@ -773,6 +773,38 @@ void ApplWnd::onSystemParametersChanged()
...
+        // Main window
+        pNotebook->hide_page(wxString("wind"));
+        pNotebook->hide_page(wxString("solar"));
+        pNotebook->show_page(wxString("renewable"));
+
+        // Scenario builder pane
+        pScenarioBuilderNotebook->hide_page(wxString("wind"));
+        pScenarioBuilderNotebook->hide_page(wxString("solar"));
+        pScenarioBuilderNotebook->show_page(wxString("renewable"));
+    }
```
flomnes: This could be simplified with a `setVisibility` method.

src/ui/simulator/toolbox/components/notebook/notebook.cpp
```diff
@@ -1023,6 +1023,18 @@ Notebook::Page* Notebook::find(const wxString& name, bool warn)
...
+    Page* page = find(name);
+    page->visible(false);
+}
+
+void Notebook::show_page(const wxString& name)
+{
+    Page* page = find(name);
+    page->visible(true);
+}
+
```
flomnes: Suggestion : use a slightly different interface
```cpp
void Notebook::setVisibility(const wxString& name, bool visible)
{
    Page* page = find(name);
    page->visible(visible);
}
```
Also, handle the case where `find` does not find, e.g by calling `return;` with an error in the logs.

flomnes:
```cpp
Notebook::Page* Notebook::find(const wxString& name, bool warn)
{
    foreach (auto* page, pPages)
    {
        if (page->name() == name)
            return page;
    }

#ifndef NDEBUG
    if (warn)
    {
        String sname;
        wxStringToString(name, sname);
        logs.warning() << "notebook: impossible to find '" << sname << "'";
    }
#endif
    return nullptr;
}
```

In debug mode, a warning is issued if find does not find, so it may not be necessary to display an error in the logs.

flomnes (changes requested):

flomnes (approved):

PR #319, opened 2021-08-03
UI : factor thermal/renewable area summary classes

guilpier-code:
Addition of a common area summary class that contains common data and behaviors of thermal/renewable area summary classes.
Consequently, thermal/renewable area summary are cleared from data/behaviors that are not specific.


flomnes (commented):
Remove comment at lines 66-99 of file **src/ui/simulator/toolbox/components/datagrid/renderer/area/renewable.areasummary.h**

flomnes:
Should be merged before #324

flomnes (approved):

PR #327, opened 2021-08-09
Include RES generation in hydro management

flomnes:


src/solver/hydro/management/management.cpp
```diff
@@ -168,6 +168,16 @@ void HydroManagement::prepareNetDemand(uint numSpace)
                 - ((ModeT != Data::stdmAdequacy) ? scratchpad.mustrunSum[hour]
                                                  : scratchpad.originalMustrunSum[hour]);
 
+            // RES
```
guilpier-code: In the comment, RES is too generic, I would mention clusters

guilpier-code (changes requested):

src/solver/hydro/management/management.cpp
flomnes: Agreed

PR #330, opened 2021-08-09
UI - factor classes for sorting clusters by certain order 

guilpier-code:


flomnes (commented):
I'm not sure why ordering requires anything else than a `name()` : the classes and methods don't need to know what's the nature (thermal or renewable) of a list of clusters.

flomnes (approved):

PR #332, opened 2021-08-10
Linux : allow ui to find solver when packaged in TGZ archive

flomnes:


guilpier-code:
Can we have a bit of context ?

flomnes:
When executing antares from an extracted .tgz package, the UI cannot find the solver (buttin "Run a Monte Carlo simulation").

This is because the solver is in the ROOT/bin, which is not looked up.

This PR adds this directory.

guilpier-code (approved):

PR #340, opened 2021-08-12
UI - clusters time series : extracting common code from cluster TS classes

guilpier-code:


flomnes (changes requested):
There is no need to differentiate data structures by thermal / renewable (l. 375 and after). We only use a common member `series`.

guilpier-code:
> There is no need to differentiate data structures by thermal / renewable (l. 375 and after). We only use a common member `series`.


Actually, it was not made on purpose : if you change 
`internalThermalClusterChanged(Antares::Data::ThermalCluster* cluster)`
into : 
`internalClusterChanged(Antares::Data::Cluster* cluster)`
and make that method common, you have to deal with the **notifier** (and the event **on<type>ClusterChanged**) in the classes constructor.
As changing **on<type>ClusterChanged** has consequances in many places, I think we should wait until other part of the code is factorized.

flomnes:
> > There is no need to differentiate data structures by thermal / renewable (l. 375 and after). We only use a common member `series`.
> 
> Actually, it was not made on purpose : if you change
> `internalThermalClusterChanged(Antares::Data::ThermalCluster* cluster)`
> into :
> `internalClusterChanged(Antares::Data::Cluster* cluster)`
> and make that method common, you have to deal with the **notifier** (and the event **onClusterChanged**) in the classes constructor.
> As changing **onClusterChanged** has consequances in many places, I think we should wait until other part of the code is factorized.

OK

flomnes (approved):

PR #342, opened 2021-08-12
Parameters are now associated to section in the parameter parser

### LINKED ISSUE #256: In generaldata.ini, section name is ignored
### Description of the issue
The two following **generaldata.ini** files are interpreted in the same way, hence they're equivalent :

> [general]
> mode = Economy

> [output]
> mode = Economy

This could lead to conflicts if two sections contained the same property : `section1/property` and `section2/property` would not be distinguished.

## Possible fix
In order to fix this, the section name should to be taken into account in source file **parameters.cpp** by adding this parameter to the relevant functions.

julza03:
The SGDIntLoadFamily_* functions, where * was a letter, were replaced… by functions where * correspond to a section name.
Thus, parameters are associated to their corresponding section.
Parameters kept for compatibility reasons are grouped in a single function

### DIFF
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -962,6 +962,7 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
         {
             // Continue on error
             logs.warning() << ini.filename() << ": '" << section->name << "': Unknown section name";
+            continue;
         }
 
         // Foreach properties in the section
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -955,13 +955,15 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
     {
         sectionName = section->name;
         sectionName.toLower();
-        handleAllKeysInSection = &sectionAssociatedToKeysProcess[section->name]; // @TODO Must check that the function exists
-        if (!handleAllKeysInSection)
+        try
+        {
+            handleAllKeysInSection = sectionAssociatedToKeysProcess.at(sectionName);
+        } catch (const std::out_of_range&)
         {
             // Continue on error
             logs.warning() << ini.filename() << ": '" << section->name << "': Unknown section name";
         }
-
+        
         // Foreach properties in the section
         for (const IniFile::Property* p = section->firstProperty; p; p = p->next)
         {
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -913,20 +913,10 @@ static bool SGDIntLoadFamily_Legacy(Parameters& d,
     return false;
 }
 
-static bool ParametersInternalLoadProperty(Parameters& params,
-                                           const String& sectionName,
-                                           const String& key,
-                                           const String& value,
-                                           const String& rawvalue,
-                                           uint version)
+bool firstKeyLetterIsValid(const String& name)
 {
-    char cKey = key.first();
-    char cSection = sectionName.first();
-    if (cKey < 'a' || cKey > 'z' || cSection < 'a' || cSection > 'z')
-        return false;
-
-
-
+    char firstLetter = name.first();
+    return (firstLetter >= 'a' && firstLetter <= 'z');
 }
 
 bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOptions& options)
@@ -969,6 +959,8 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
         for (const IniFile::Property* p = section->firstProperty; p; p = p->next)
         {
             if (p->key.empty())
+                continue;
+            if (!firstKeyLetterIsValid(p->key))
                 continue;
             // We convert the key and the value into the lower case format
             value = p->value;
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1756,7 +1756,7 @@ void Parameters::saveToINI(IniFile& ini) const
 
     // Seeds
     {
-        auto* section = ini.addSection("seeds - Mersenne Twister"); //@TODO : Why Capital letters are used here ?
+        auto* section = ini.addSection("seeds - Mersenne Twister");
         for (uint sd = 0; sd != (uint)seedMax; ++sd)
             section->add(SeedToID((SeedIndex)sd), seed[sd]);
     }
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -926,7 +926,11 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
     // A temporary buffer, used for the values in lowercase
     String value;
     String sectionName;
-    typedef bool (*Callback)(Parameters&, const String&, const String&, const String&, uint);
+    typedef bool (*Callback)(Parameters&, // [out] Parameter object to load the data into
+            const String&, // [in] Key, comes left to the '=' sign in the .ini file
+            const String&, // [in] Lowercase value, comes right to the '=' sign in the .ini file
+            const String&, // [in] Raw value as writtent right to the '=' sign in the .ini file
+            uint); // [in] Version of the simulator (such as 710)
 
     static const std::map<String,Callback> sectionAssociatedToKeysProcess = {
             {"general",&SGDIntLoadFamily_General},
@@ -969,7 +973,7 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
             // Deal with the current property
             // Do not forget the variable `key` and `value` are identical to
             // `p->key` and `p->value` except they are already in the lower case format
-            if (not (handleAllKeysInSection)(*this, p->key, value, p->value, version))
+            if (not handleAllKeysInSection(*this, p->key, value, p->value, version))
             {
                 if (not SGDIntLoadFamily_Legacy(*this, p->key, value, p->value, version))
                 {
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -969,7 +969,7 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
             // Deal with the current property
             // Do not forget the variable `key` and `value` are identical to
             // `p->key` and `p->value` except they are already in the lower case format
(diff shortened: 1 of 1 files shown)


src/libs/antares/study/parameters.cpp
```diff
@@ -691,327 +791,169 @@ static bool SGDIntLoadFamily_P(Parameters& d, const String& key, const String& v
...
-    }
-    else
-    {
-        int i = c - 'a';
-        if (jumper[i])
-            return (*(jumper[i]))(params, key, value, version);
-    }
-    return false;
+
+
```
flomnes: Rename this function to `checkKeySection(const String& s)` to check that the 1st character is in range a-z.

Call it whenever a section/property name is read.

src/libs/antares/study/parameters.cpp
```diff
@@ -1020,10 +962,33 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
...
+    static std::map<String,Callback> sectionAssociatedToKeysProcess = {
+            {"general",&SGDIntLoadFamily_General},
+            {"input",&SGDIntLoadFamily_Input},
+            {"output",&SGDIntLoadFamily_Output},
+            {"optimization",&SGDIntLoadFamily_Optimization},
+            {"other preferences",&SGDIntLoadFamily_OtherPreferences},
+            {"advanced parameters",&SGDIntLoadFamily_AdvancedParameters},
+            {"playlist",&SGDIntLoadFamily_Playlist},
+            {"variables selection",&SGDIntLoadFamily_VariablesSelection},
+            {"seeds",&SGDIntLoadFamily_Seeds}};
```
flomnes: The complete name is **seeds - mersenne twister**

src/libs/antares/study/parameters.cpp
```diff
@@ -1020,10 +962,33 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
...
+
+    Callback* handleAllKeysInSection;
```
flomnes: `Callback` is already a function pointer. This is equivalent to 
```cpp
function_t** Callback;
```
I suggest you remove one indirection level.

src/libs/antares/study/parameters.cpp
```diff
@@ -1020,10 +962,33 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
...
     // Foreach section on the ini file...
     for (auto* section = ini.firstSection; section; section = section->next)
     {
+        sectionName = section->name;
+        sectionName.toLower();
+        handleAllKeysInSection = &sectionAssociatedToKeysProcess[section->name]; // @TODO Must check that the function exists
```
flomnes: 1. Use `std::map<T>::count` to determine the existence of the key within the map.
2. Use `sectionName` for the key.
3. Use `std::map<T>::at`, it won't try to add elements to your map unlike `operator[]`. More details on [at](https://www.cplusplus.com/reference/map/map/at/) and [operator\[\]](https://www.cplusplus.com/reference/map/map/operator[]/).

src/libs/antares/study/parameters.cpp
```diff
@@ -1783,7 +1756,7 @@ void Parameters::saveToINI(IniFile& ini) const
 
     // Seeds
     {
-        auto* section = ini.addSection("seeds - Mersenne Twister");
+        auto* section = ini.addSection("seeds - Mersenne Twister"); //@TODO : Why Capital letters are used here ?
```
flomnes: I suggest you keep this as is for compatibility reasons.

Possible explanation [here](https://fr.wikipedia.org/wiki/Mersenne_Twister).

src/libs/antares/study/parameters.cpp
```diff
@@ -1020,10 +962,33 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
     reset();
     // A temporary buffer, used for the values in lowercase
     String value;
-
+    String sectionName;
+    typedef bool (*Callback)(Parameters&, const String&, const String&, const String&, const String&, uint);
+
+    static std::map<String,Callback> sectionAssociatedToKeysProcess = {
```
flomnes: `static const`

flomnes (changes requested):
Thank you for your contribution !

src/libs/antares/study/parameters.cpp
julza03: OK. Commit 55c18c4

src/libs/antares/study/parameters.cpp
julza03: OK. See [commit ](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/342/commits/a5a76f731abb174f417a1e70ec7f1fd996da2981)

src/libs/antares/study/parameters.cpp
julza03: Ok. Other solution adopted with try/catch block around .at use. If no section is found an exception, catched, will be raised. See the documentation if needed to confirm the try/catch block. 
See a8cb305

src/libs/antares/study/parameters.cpp
julza03: OK. Commits b03bbed and b5def8c

src/libs/antares/study/parameters.cpp
julza03: Partially corrected here : fa68674. 
I've made the correction only for parameter names (but not for sections). Indeed, section names are already checked in the map and I'm not sure that verifying it before will make any difference in the performances of the app. 
I think that this control is useless, even for parameters, with the new implementation (which is not based on the first letter). You may want to remove this function.

src/libs/antares/study/parameters.cpp
julza03: Ok. Commits 96a1f00 (fix 0a225bf).

julza03:
Other suggestions on this file parameters.cpp : 
- Before merging, you may want to apply a reformat code - I didn't know if my editor was conform to the rules of Antares so I didn't reformatted the code. 
- #include <stdio.h> is the C syntax. To adopt C++ style, it will be better to write : #include <cstdio> (You may check why cstdio is required, because iostream, already included, may answer to your needs. But CLion warns that these library are unused - you may want to do an analysis). 
- In SGDIntLoadFamily_Optimization, the if concerning "link-type" may be rewritten like this : 
if (key == "link-type")
    {
        CString<64, false> v = value;
        v.trim();
        v.toLower();
        if (value == "ac")
            d.linkType = ltAC;
        else
            d.linkType = ltLocal;
        return true;
    }
- As in the previous version, these comments, inLoadFromINI don't reflect the code : // We convert the key and the value into the lower case format. // Do not forget the variable `key` and `value` are identical to
            // `p->key` and `p->value` except they are already in the lower case format
Only the value is converted in lower case, I think that the key must be converted too because the comparisons in SGDIntLoadFamily are made with text in lowercase. (Maybe p->key is already in lower-case, I don't know.)
-saveToINI is a very long function. I think you may want to split it in several ones, each one in charge of saving one section.

src/libs/antares/study/parameters.cpp
flomnes: I agree

src/libs/antares/study/parameters.cpp
```diff
@@ -1020,24 +925,57 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
...
+                continue;
             // We convert the key and the value into the lower case format
             value = p->value;
             value.toLower();
 
             // Deal with the current property
-            if (not ParametersInternalLoadProperty(*this, p->key, value, p->value, version))
+            // Do not forget the variable `key` and `value` are identical to
+            // `p->key` and `p->value` except they are already in the lower case format
+            if (not (handleAllKeysInSection)(*this, p->key, value, p->value, version))
```
flomnes: Style : parenthesis around function name are not needed.

src/libs/antares/study/parameters.cpp
```diff
@@ -1020,24 +925,57 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
     reset();
     // A temporary buffer, used for the values in lowercase
     String value;
-
+    String sectionName;
+    typedef bool (*Callback)(Parameters&, const String&, const String&, const String&, uint);
```
flomnes: For clarity, I suggest you comment each parameter
Example :
```cpp
typedef bool (*Callback)(Parameters& // [out] Parameter object to load the data into
                         const String& // [in] Key, comes left to the '=' sign in the .ini file
                         const String& // [in] Lowercase value, comes right to the '=' sign in the .ini file
                                       // etc.
                        );

**Review by flomnes — COMMENTED** [review]
Formatting may be incorrect, I'll fix it before a merge.

To be added to Antares-next

File: src/libs/antares/study/parameters.cpp (hunk 8 above)
Comment (julza03) [reply src/libs/antares/study/parameters.cpp]: Ok : 0b051f3

File: src/libs/antares/study/parameters.cpp (hunk 7 above)
Comment (julza03) [reply src/libs/antares/study/parameters.cpp]: Ok : 0b051f3

File: src/libs/antares/study/parameters.cpp [hunk 9]
```diff
@@ -926,7 +926,11 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
     // A temporary buffer, used for the values in lowercase
     String value;
     String sectionName;
-    typedef bool (*Callback)(Parameters&, const String&, const String&, const String&, uint);
+    typedef bool (*Callback)(Parameters&, // [out] Parameter object to load the data into
+            const String&, // [in] Key, comes left to the '=' sign in the .ini file
+            const String&, // [in] Lowercase value, comes right to the '=' sign in the .ini file
+            const String&, // [in] Raw value as writtent right to the '=' sign in the .ini file
+            uint); // [in] Version of the simulator (such as 710)
```
flomnes: Upon closer examination, the version is read from the **study**, it is not the version of Antares Simulator.

**load.cpp**
```cpp
bool Study::internalLoadHeader(const String& path)
{
    // Header
    buffer.clear() << path << SEP << "study.antares";
    if (not header.loadFromFile(buffer))
    {
        logs.error() << path << ": impossible to open the header file";
        return false;
    }
// ...
}
```
**header.cpp**
```cpp
bool StudyHeader::internalLoadFromINIFile(const IniFile& ini, bool warnings)
{
    version = 0;
    const IniFile::Section* sect = ini.find("antares");
// ...
            // Version
            if (p->key == "version")
            {
                version = p->value.to<uint>(); // Property "version" in section "antares" from file study.antares
                continue;
            }
// ...
}
```

flomnes (commented):
One last change: comment at line 933

guilpier-code (approved):

flomnes (approved):

PR #349, opened 2021-08-17
UI : factor sc builder

flomnes:
Create a template for thermal & renewable sc. builder grid rendering.

### DIFF
--- a/src/ui/simulator/application/main/build/scenario-builder.cpp
+++ b/src/ui/simulator/application/main/build/scenario-builder.cpp
@@ -177,7 +177,7 @@ public:
     void create() override
     {
         page_ = createStdNotebookPage<Toolbox::InputSelector::Area>(
-          notebook_, Renderer::thermalScBuilderRenderer::internalLabel, Renderer::thermalScBuilderRenderer::InternalLabel);
+          notebook_, wxT("thermal"), wxT("Thermal"));
         createRenderer();
         control_->updateRules.connect(renderer_,
                                       &Renderer::thermalScBuilderRenderer::onRulesChanged);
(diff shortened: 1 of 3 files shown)


src/ui/simulator/application/main/build/scenario-builder.cpp
```diff
@@ -165,85 +165,69 @@ class solarScBuilderGrid : public basicScBuilderGrid
     }
 };
 
-// Thermal ...
-class thermalScBuilderGrid : public basicScBuilderGrid
+// Thermal & renewable
+template<class Renderer, class Label>
+class commonScBuilderGrid : public basicScBuilderGrid
```
guilpier-code: In **commonScBuilderGrid**, **common** can be a bit confusing because it suggest it suggests that this class is common to parent to all sc builder grid.
Maybe **clusterScBuilderGrid** is better.

src/ui/simulator/application/main/build/scenario-builder.cpp
```diff
@@ -165,84 +165,94 @@ class solarScBuilderGrid : public basicScBuilderGrid
...
+    virtual void createGrid() override
     {
         grid_ = new DatagridType(page_.first, renderer_);
     }
-    void addToNotebook()
+
+    virtual void addToNotebook() override
     {
-        page_.first->add(grid_, wxT("thermal"), wxT("Thermal"));
+        page_.first->add(grid_, label().first, label().second);
```
guilpier-code: **label().first** and **label().second** mean nothing to me at first sight, sorry...
They actually mean respectively a grid **name** and a grid **caption**;
So I would define 2 overloaded methods `name()` and `caption()`.

guilpier-code (changes requested):

guilpier-code (approved):

PR #351, opened 2021-08-18
UI - extracting common code in inspector

guilpier-code:


flomnes (commented):
While I understand your idea, I am not sure the new code is easier to read and maintain.

PR #354, opened 2021-08-19
Renewable clusters for RES as default on study creation

guilpier-code:


flomnes (changes requested):
Since "clusters" becomes the new default, I suggest changing the label of this choice in the UI **Advanced parameters**, see `src/ui/simulator/windows/options/advanced/advanced.cpp:1018`. And setting "cluster [default]" as the 1st entry.

src/libs/antares/study/study.cpp
```diff
@@ -159,6 +159,7 @@ void Study::createAsNew()
 
     // Simulations
     parameters.reset();
+    parameters.renewableGeneration.rgModelling = Antares::Data::rgClusters;
```
flomnes: Add a comment to explain why we set this value here, possibly

// GUI : use renewable clusters when creating a study

flomnes (approved):

PR #357, opened 2021-08-24
Add switch for TS generation for some thermal clusters

flomnes:
For each thermal cluster, add a property that can take 3 values :
- Use "generate thermal TS" property from study, as defined by parameters `generate` and `refreshtimeseries`
- Force generation of TS
- Force no generation of TS

There might be a distinction between generation and refresh, this needs to be discussed with the POs.


flomnes:
~~TODO : In the GUI, do not disable input for thermal refresh parameters.~~

Done

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -43,6 +43,26 @@ using namespace Antares;
...
+{
+namespace Data
+{
+bool ThermalCluster::doWeGenerateTS(bool global, bool refresh) const
+{
+    switch (tsGenBehavior)
+    {
+    // Generate if global tells us to
+    case TSGenerationBehavior::useGlobalParameter:
+        return global && refresh;
```
flomnes: Strange

hugo-antoine-rtei:
I conducted a small series of tests to make sure it works as intended. I created clusters, on to check the behavior of forced outage and one for planned outage. I ran the following combinations:
- Force generation with Ready-made TS in simulation window 
- Force generation with Stochastic TS in simulation window
- Force no generation with Ready-made TS in simulation window
- Force no generation with Stochastic TS in simulation window
- Use global parameter with Ready-made TS in simulation window
- Use global parameter with Stochastic TS in simulation window

In each of these, the input ts were modified and the output files were created as expected according to the options specified.

I also checked a couple of these using "run the time-series generators only" mode, behavior is also as expected.

When loading a random study (created y another antares version) the default "Generate TS" is set to "Use global parameter" as expected.

To sum up : ALL GOOD, no problems found

flomnes:
New look for thermal cluster parameters

image

PR #363, opened 2021-08-30
UI - input menu behaviour when changing renewable modelling

guilpier-code:


src/ui/simulator/application/main/main.cpp
```diff
@@ -489,8 +490,11 @@ void ApplWnd::evtOnUpdateGUIAfterStudyIO(bool opened)
         EnableItem(menu, mnIDViewSimulation, opened);
         EnableItem(menu, mnIDViewNotes, opened);
         EnableItem(menu, mnIDViewLoad, opened);
-        EnableItem(menu, mnIDViewSolar, opened);
-        EnableItem(menu, mnIDViewWind, opened);
+
+        EnableItem(menu, mnIDViewSolar, opened && not study->parameters.renewableGeneration.isClusters());
+        EnableItem(menu, mnIDViewWind, opened && not study->parameters.renewableGeneration.isClusters());
```
flomnes: use `isAggregated()`

flomnes: You could use a reference to `study->parameters.renewableGeneration` (auto& ref = ...) to lighten the code

guilpier-code: ok, but there is a problem here : 
In the current function, argument **opened** stands for "has the study been opened" and can be **false**, and in that case we haven't even constructed the study.
So defining a variable like : 
`bool renewableModellingIsAggregated = study->parameters.renewableGeneration.isAggregated();`
before the call to **Enable(...)** functions leads to a crash if **opened** is false.
However, an expression like : 
`opened && study->parameters.renewableGeneration.isAggregated()`
is OK in all cases : if **opened** is false, the second part of the expression is not evaluated and there is no crash

flomnes (approved):

PR #367, opened 2021-08-31
Remove libantares-ui-component-htmllistbox

### LINKED ISSUE #325: CMAKE : useless library ?
On one hand we have the **libantares-ui-component-htmllistbox** library.
On the other hand we have the bigger **libantares-ui-components** library, that includes the files that the previous library includes.
So the sames files are included in 2 libraries.
Is it normal or expected ? Or is it a wrong repetition ?

The problem is : 
if we change files in one library some day,
- we must change them in the other one.
- we could forget to add/change files in one of the 2 libraries

At least the common files should be gathered in a single cmake variable.
At best, one of the 2 lists of common files should be removed.

flomnes:
Fix #325


flomnes:
As discussed with @guilpier-code, remove the other target library.

PR #372, opened 2021-09-07
Refactor TS numbers

guilpier-code:


flomnes (commented):
TODO :
- [ ] Factor functions for different modes. Possibly use templates for the matrix type (Matrix<sint32>, etc.) and compiler template argument deduction for functions
- [ ] Decouple draws and storage of TS numbers
- [ ] Cut the Generate function into smaller function

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -120,155 +81,601 @@ namespace Antares
...
+        return false;
+    }
+
+    return true;
+}
+
+void drawTSnumbersForIntraModal(    uint32* intramodal_draws,
+                                    const bool* isTSintramodal, 
+                                    uint* nbTimeseriesByMode, 
+                                    Data::Study& study)
```
flomnes: study is not needed. Pass only `study.runtime->random` by non-const reference ?

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -120,155 +81,601 @@ namespace Antares
...
+                    * nbTimeseriesByMode[tsKind]));
         }
-        case Data::timeSeriesWind:
+    }
+}
+
+void storeTSnumbersForIntraModal(   const bool* isTSintramodal, 
+                                    uint32* intramodal_draws, 
+                                    uint year, 
+                                    Data::Study& study)
```
flomnes: Only `study.areas` is required. Pass it by non-const reference instead of `study` ?

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -280,111 +687,57 @@ bool TimeSeriesNumbers::Generate(Data::Study& study)
...
-            }
-        }
-
-        study.areas.each([&](Data::Area& area) {
-            assert(y < area.load.series->timeseriesNumbers.height);
-            area.load.series->timeseriesNumbers[0][y] = DRAW_A_RANDOM_NUMBER(
-              Data::timeSeriesLoad, area.load.series->series, parameters.nbTimeSeriesLoad);
+        // Intra-modal TS : draw and store TS numbres
+        drawTSnumbersForIntraModal(intramodal_draws, isTSintramodal, nbTimeseriesByMode, study);
+        storeTSnumbersForIntraModal(isTSintramodal, intramodal_draws, year, study);
```
flomnes: 1st and 2nd parameters of these 2 functions are inverted. Consider using the same order for consistency.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -120,155 +81,601 @@ namespace Antares
...
+public:
+    areaTSNumbersListRetriever(Data::Study& study) : study_(study) {}
+    virtual std::vector<uint> getAreaTimeSeriesNumber(const Data::Area& area) = 0;
+    virtual uint getGeneratedTimeSeriesNumber() = 0;
+protected:
+    Data::Study& study_;
+};
+
+class areaLoadTSNumbersListRetriever : public areaTSNumbersListRetriever
+{
```
flomnes: Keep the class interface, but use templates to name derived classes ?

Something like that
```cpp
class Interface {
 virtual void f() = 0;
};

template<int mode>
struct Derived;

template<>
struct Derived<timeSeriesThermal> : public Interface {
// ...
};
```

guilpier-code: it seems excessive to use the TS kind as a template parameter to make Derived instance more understandable.
Indeed, the template parameter is not meant to be used inside the class. 
After we talked about it, we decided to rename **areaTSNumbersListRetriever**'s children in such a way that the kind of TS is highlighted.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -120,155 +81,601 @@ namespace Antares
...
+        area.load.series->timeseriesNumbers.zero();
+        area.solar.series->timeseriesNumbers.zero();
+        area.wind.series->timeseriesNumbers.zero();
+        area.hydro.series->timeseriesNumbers.zero();
+
+        for (uint i = 0; i != area.thermal.clusterCount(); ++i)
+        {
+            auto& cluster = *(area.thermal.clusters[i]);
+            cluster.series->timeseriesNumbers.zero();
+        }
```
flomnes, line 100: Add renewable clusters ?

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -25,87 +25,48 @@
...
+};
+
+const map<TimeSeries, string> ts_to_tsTitle = {
+    {timeSeriesLoad, "load"},
+    {timeSeriesHydro, "hydro"},
+    {timeSeriesWind, "wind"},
+    {timeSeriesThermal, "thermal"},
+    {timeSeriesSolar, "solar"},
+    {timeSeriesRenewable, "renewable clusters"}
+};
```
flomnes, line 64: use pure virtual methods ?

flomnes (changes requested):
1. Replace macro `TS_INDEX` with `ts_to_tsIndex.at(...)`, remove macro definition.
2. Loop on timeSeries index

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: done

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: not done. Not sure if it is suitable.

flomnes: Ok I agree. Let's forget this idea.

PR #373, opened 2021-09-09
UI code factor ts management

guilpier-code:


flomnes:
PR #379 provides a better refactor.

PR #377, opened 2021-09-10
LPsolve measurements

flomnes:


flomnes (commented):
The average duration displayed for year `n` is in fact the average for years `1..n`. This is misleading to the user and must be changed.

In other words, display the total solve & update time for each year.

To be discussed.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -63,6 +65,31 @@ using namespace Yuni;
...
+        end_ = clock::now();
+    }
+
+    long long duration_ms() const {
+        return std::chrono::duration_cast<std::chrono::milliseconds>(end_ - start_).count();
+    }
+
+    std::string toString() const {
+        return std::to_string(duration_ms()) + "ms";
+    }
```
flomnes: Add space between number and units, see rule 15 [here](https://physics.nist.gov/cuu/Units/checklist.html).

PR #379, opened 2021-09-10
UI code factor - ts management : second try

guilpier-code:


src/ui/simulator/toolbox/components/datagrid/renderer/ts-management-clusters-as-renewables.cpp
```diff
@@ -65,461 +51,12 @@ wxString TSmanagementRenewableCluster::columnCaption(int colIndx) const
                                              wxT("   Thermal   "),
```
guilpier-code: The captions should be retrieved from de columns, not fixed one for all in the TS manager

src/ui/simulator/toolbox/components/datagrid/renderer/column.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// -------------------- -
+// Classic column
+// ---------------------
+classicColumn::classicColumn(TimeSeries ts) : Column()
+{
+	tsKind_ = ts;
+
+	cells_.push_back(make_unique<blankCell>());
```
guilpier-code: **make_unique** is unknown on Linux, compiler does not know it.
Could be solved (or not) in cmake files with : 
`set(CMAKE_CXX_STANDARD 17) `or `set(CMAKE_CXX_STANDARD 14)`

flomnes: No need for `std::unique_ptr` here, since the memory is handled by objects. We can just free the memory at object destruction.

src/ui/simulator/toolbox/components/datagrid/renderer/ts-management-clusters-as-renewables.cpp
flomnes: done

flomnes (approved):

PR #382, opened 2021-09-20
Feature/sonarcloud

jmkerloch:


flomnes (commented):
Why not use [this existing action](https://github.com/SonarSource/sonarcloud-github-action) ?

jmkerloch:
> 
> 
> Why not use [this existing action](https://github.com/SonarSource/sonarcloud-github-action) ?

It seems the action is not supported for C/C++ analysis.

I guess it's because we need some specific options for configure and build C/C++ project which requires build for analysis in sonarcloud.

PR #385, opened 2021-09-22
centos7 CI with docker image

jmkerloch:
- add docker image with installed dependency and pre-compiled libraries download
- add docker image with installed dependency, pre-compiled librairies download and sources checkout
- push docker images to docker.io (without source for every push _only if dockerfile changed_, with source on release creation)
- use docker image in github action CI
- remove unused dockerfiles


docker/build-docker-file-src.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+cat centos7 centos7-src > build-centos7-src
```
flomnes: Why do we need to concatenate these 2 files ?

Why not add file `build-centos7-src` directly to the repo ?

antares-version.json
```diff
@@ -0,0 +1,3 @@
+{
+    "antares_deps_version": "2.0.0-rc2"
+}
```
flomnes: I'm not sure why we need to keep track of antares_deps_version in a .json file.

If it's a submodule, the branch name is kept in file **.gitmodules**.

```ini
[submodule "src/antares-deps"]
	path = src/antares-deps
	url = https://github.com/AntaresSimulatorTeam/antares-deps.git
	branch = v1.0.2
```

docker/centos7
```diff
@@ -0,0 +1,26 @@
...
+
+# Install requirements : update repo
+RUN yum update -y
+
+# Install requirements : install epel
+RUN yum install -y epel-release 
+
+# Install requirements
+RUN yum install -y git redhat-lsb-core gcc gcc-c++ make wget centos-release-scl scl-utils rpm-build &&\
+    yum install -y cmake3 devtoolset-7
```
flomnes: Only one `yum` command ?

jmkerloch: We need to have 2 command because `devtoolset-7` need `centos-release-scl` to be installed first.

antares-version.json
jmkerloch: It's used in antares-xpansion to avoid definition of version in each github action .yml files.
We could look for github-action to get submodules tag.
⚠️ In this case we should only point to release tag because otherwise there won't be any release asset available to download

docker/build-docker-file-src.sh
jmkerloch: `centos7-src` file contains only `RUN` docker command needed to get project sources.
It's used only in case of release to have a docker image with already cloned repository. 
`cat` is used to avoid repeating contents of `centos7` docker file in `centos7-src`. We use the concatenation of both file when building the docker image.

.github/workflows/centos7-system.yml
```diff
@@ -10,49 +10,67 @@ on:
...
-          
-    - name: Install dependencies
-      run: |
-          pip3 install -r src/tests/examples/requirements.txt
 
     - name: Configure
       run: |
            source /opt/rh/devtoolset-7/enable
-           cmake3 -B _build -S src -DDEPS_INSTALL_DIR=rte-antares-deps-Release -DCMAKE_BUILD_TYPE=release -DBUILD_TESTING=ON -DBUILD_OUTPUT_TEST=ON
+           cmake3 -B _build -S src -DDEPS_INSTALL_DIR=/rte-antares-deps-Release -DCMAKE_BUILD_TYPE=release -DBUILD_TESTING=ON -DBUILD_OUTPUT_TEST=ON
```
flomnes: `-DDEPS_INSTALL_DIR=/rte-antares-deps-Release` is the default value if Antares_Simulator is placed in the fs root `/`. Remove this argument ?

flomnes (changes requested):
## git2
If git2 is required, install package **rh-git227-git** and enable it using `source /opt/rh/rh-git227/enable`.

## CMake tries to download Antares_Simulator_Tests
Either set `BUILD_TESTING=OFF` or add Antares_Simulator_Tests submodule to the docker image.

## CMake tries to build deps, although they're already present as pre-compiled libs
Use option `-DBUILD_not_system=OFF`

## GUI is probably not required
`-DBUILD_UI=OFF`

.github/workflows/centos7-system.yml
jmkerloch: the default value is `<antares_src>/../rte-antares-deps-Release`

docker/build-docker-file-src.sh
guilpier-code: We could name these scripts in a more expressive way, for instance **centos7-basic-requirements** and **centos7-clone-antares**

.github/workflows/centos7-system.yml
```diff
@@ -10,49 +10,68 @@ on:
...
+      with:
+        fetch-depth: 2
+        
+    - name: Get changed files
+      id: changed-files
+      uses: tj-actions/changed-files@v1.1.2
+      with:
+        files: |
+          docker/centos7
+          antares-version.json
```
flomnes: antares-deps-version.json ?

flomnes (approved):

.github/workflows/centos7-system.yml
jmkerloch: file is named `antares-version.json` like in `antares-xpansion`

.github/workflows/centos7-system.yml
jmkerloch: and we are doing git clone in .

flomnes: ok

.github/workflows/centos7-system.yml
flomnes: This file contains the version of antares-deps, right ?

jmkerloch: yes

docker/build-docker-file-src.sh
flomnes: That would be nice

.github/workflows/centos7-system.yml
flomnes: Then IMO it should be named antares-deps-version.json

.github/workflows/centos7-system.yml
```diff
@@ -10,49 +10,68 @@ on:
       - release/*
 jobs:
 
+  docker_publish:
+          
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+      with:
+        fetch-depth: 0
```
flomnes, line 19: Why would we want to fetch all commits from all branches here ?

jmkerloch: this is what is used for github action https://github.com/tj-actions/changed-files

flomnes (approved):

flomnes (approved):

docker/build-docker-file-src.sh
jmkerloch: file name changed

.github/workflows/centos7-system.yml
jmkerloch: file renamed

PR #386, opened 2021-09-22
Warn the user about disabled renewable clusters only when relevant

### LINKED ISSUE #381: Renewable cluster is not found when TS numbers are specified by the user.
**Description**
Renewable cluster is not found when TS numbers are specified by the user. A warning is displayed, no effect on the simulation.

**How to Reproduce**
Change renewable modelling to "clusters". Specify custom TS years. Change renewable modelling to "aggregated". Launch simulation.

**Expected behavior**
No warning should be displayed, since renewable clusters should be ignored when `rgModelling==aggregated`.

a64852b3-e080-4696-b220-39aca7901ae5

**OS and version**
 - OS: Windows
 - Antares Simulator version: 8.1

**Additional context**
None

flomnes:
Fixes #381

### DIFF
--- a/src/libs/antares/study/scenario-builder/rules.cpp
+++ b/src/libs/antares/study/scenario-builder/rules.cpp
@@ -162,7 +162,7 @@ void Rules::loadFromInstrs(Study& study,
         }
     }
 
-    if (kind_of_scenario == "r")
+    if (kind_of_scenario == "r" && study.parameters.renewableGeneration() == rgClusters)
     {
         if (clustername.empty())
             return;
--- a/src/libs/antares/study/scenario-builder/rules.cpp
+++ b/src/libs/antares/study/scenario-builder/rules.cpp
@@ -162,7 +162,7 @@ void Rules::loadFromInstrs(Study& study,
         }
     }
 
-    if (kind_of_scenario == "r" && study.parameters.renewableGeneration() == rgClusters)
+    if (kind_of_scenario == "r" && study.parameters.renewableGeneration.isClusters())
     {
         if (clustername.empty())
             return;


src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -162,7 +162,7 @@ void Rules::loadFromInstrs(Study& study,
         }
     }
 
-    if (kind_of_scenario == "r")
+    if (kind_of_scenario == "r" && study.parameters.renewableGeneration() == rgClusters)
```
guilpier-code: use **study.parameters.renewableGeneration.isClusters()** instead ?

guilpier-code (changes requested):

src/libs/antares/study/scenario-builder/rules.cpp
flomnes: done

guilpier-code: Maybe we could remove all occurences of **study.parameters.renewableGeneration() == rgClusters or rgAggregated** and replace them with **study.parameters.renewableGeneration.isClusters()** or **study.parameters.renewableGeneration.isAggregated()**, so that there would be no more reference to enum **RenewableGenerationModelling** in these files and possibly we could remove header inclusion from them

flomnes: Ok to be corrected after changes related to ts-numbers & unit tests

flomnes: About that, see #426

guilpier-code (approved):

PR #388, opened 2021-09-23
Pass typeVar instead of &Probleme, remove unused function argument

flomnes:
* change_MPSolver_variables_bounds doesn't need PROBLEME_SIMPLEXE*
  problemeSimplexe, only its member TypeDeVariable. Pass it directly.

* ORTOOLS_CorrigerLesBornes no longer needs argument Probleme, remove it

### DIFF
--- a/src/solver/utils/ortools_utils.cpp
+++ b/src/solver/utils/ortools_utils.cpp
@@ -20,8 +20,8 @@ void change_MPSolver_rhs(MPSolver* solver, double* rhs, char* sens, int nbRow);
 void change_MPSolver_variables_bounds(MPSolver* solver,
                                       double* bMin,
                                       double* bMax,
-                                      int nbVar,
-                                      PROBLEME_SIMPLEXE* problemeSimplexe);
+                                      int* typeVar,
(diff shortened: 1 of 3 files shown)


src/solver/utils/ortools_utils.cpp
```diff
@@ -354,11 +352,9 @@ extern "C"
                                    double* bMin,
                                    double* bMax,
                                    int* typeVar,
-                                   int nbVar,
-                                   PROBLEME_SIMPLEXE* Probleme)
+                                   int nbVar)
     {
-        Probleme->TypeDeVariable = typeVar;
-        change_MPSolver_variables_bounds(solver, bMin, bMax, nbVar, Probleme);
+        change_MPSolver_variables_bounds(solver, bMin, bMax, typeVar, nbVar);
```
flomnes: Remove function `change_MPSolver_variables_bounds` and place its implementation here ?

flomnes: Done

guilpier-code (approved):

Issue #404, opened 2021-10-07
Write function for optimization.ini does not use IniFile


flomnes:
See `src/libs/antares/study/area/list.cpp`
```cpp
        buffer.clear() << folder << SEP << "input" << SEP << "areas" << SEP << area.id << SEP
                       << "optimization.ini";
        IO::File::Stream file;
        if (file.openRW(buffer))
        {
            buffer.clear();
            buffer << "[nodal optimization]\n";
            buffer << "non-dispatchable-power = "
                   << ((bool)((area.nodalOptimization & anoNonDispatchPower) != 0) ? "true\n"
                                                                                   : "false\n");
            buffer << "dispatchable-hydro-power = "
                   << ((bool)((area.nodalOptimization & anoDispatchHydroPower) != 0) ? "true\n"
                                                                                     : "false\n");
            buffer << "other-dispatchable-power = "
                   << ((bool)((area.nodalOptimization & anoOtherDispatchPower) != 0) ? "true\n"
                                                                                     : "false\n");
            buffer << "spread-unsupplied-energy-cost = " << area.spreadUnsuppliedEnergyCost << '\n';
            buffer << "spread-spilled-energy-cost = " << area.spreadSpilledEnergyCost << '\n';

            buffer << '\n';
            buffer << "[filtering]\n";
            buffer << "filter-synthesis = ";
            AppendFilterToString(buffer, area.filterSynthesis);
            buffer << '\n';
            buffer << "filter-year-by-year = ";
            AppendFilterToString(buffer, area.filterYearByYear);
            buffer << '\n';
            buffer << '\n';
            file << buffer;
        }
```
We should use `IniFile::addSection` and `IniFile::add`.

flomnes:
Fixed by #419

PR #407, opened 2021-10-12
Links : improving save of properties.ini

guilpier-code:
Related to #404


guilpier-code:
**sonarQube analysis** : 
some code smell are difficult to change : if we choose to create a new file, with new functions, but contained in a existing **namespace** wich have a not compliant name (says sonarQube), we can't easily change the namespace

flomnes (changes requested):
Add tests to both old & new code (create branch on develop then merge into this branch)

PR #416, opened 2021-10-18
Test generation of links properties.ini files

guilpier-code:


flomnes (changes requested):
Finding the root dir using the `.git` directory is not robust. If we add a git submodule, this submodule could be found instead.

Furthermore, if we chose not to use git anymore, these tests would fail.

PR #425, opened 2021-10-21
Fix weekly->daily interpolation for min/max reservoir level

flomnes:


flomnes:
```
f(0) = levelInterpolBeg
f(23) = levelInterpolEnd
f(x) = a*x+b
```

PR #427, opened 2021-10-22
Add infeasibility analysis tool

flomnes:


src/solver/infeasible-diag/problem.cpp
```diff
@@ -0,0 +1,91 @@
...
+MPSolver::ResultStatus InfeasibleProblemDiag::Solve()
+{
+    return mSolver->Solve();
+}
+
+InfeasibleProblemReport InfeasibleProblemDiag::produceReport()
+{
+    addSlackVariables();
+    buildObjective();
+    assert(mSlackVariables.size() > 0);
```
flomnes: Absence of slack variables shouldn't be fatal : we only provide a analysis

src/solver/cmake/infeasible-diag.cmake
```diff
@@ -0,0 +1,12 @@
+set(SRC_INFEASIBLE_DIAG
+  infeasible-diag/problem.cpp
+  infeasible-diag/problem.h
+  infeasible-diag/report.h
+  infeasible-diag/report.cpp
+  )
+
+add_library(libinfeasible_diag ${SRC_INFEASIBLE_DIAG})
```
flomnes: Find a better name ?

flomnes (commented):
Check slack variables

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -345,6 +352,17 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, uint numSpace, int NumIn
         logs.error() << "* Hydro reservoir impossible to manage with cumulative options \"hard "
                         "bounds without heuristic\"";
 
+        Optimization::InfeasibleProblemDiag diag(&Probleme, "(.)*");
```
guilpier-code: we discussed it : **diagnostic** is not a good name for our analysis. Diag is worse.
Whatever name we choose, it should appear in full (not diag for diagnostic, and certainly not the short name for analysis)

guilpier-code: Smothing else on this line : regular expression pattern should be an inner of **InfeasibleProblemDiag**

src/solver/infeasible-diag/problem.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+    return mSolver->Solve();
+}
+
+InfeasibleProblemReport InfeasibleProblemDiag::produceReport()
+{
+    addSlackVariables();
+    if (mSlackVariables.size() == 0)
+        throw std::runtime_error(
+          "Cannot generate infeasibility report: no constraints have been pre-selected");
```
guilpier-code: For me, this message is unclear.
with the good regex pattern, is this possible that no slack variable was built ?
And what would it mean ?

src/solver/infeasible-diag/problem.cpp
```diff
@@ -0,0 +1,86 @@
...
+    buildObjective();
+    const MPSolver::ResultStatus s = Solve();
+    if (s != MPSolver::OPTIMAL && s != MPSolver::FEASIBLE)
+        throw std::runtime_error(
+          "Cannot generate infeasibility report: linear problem could not be solved");
```
guilpier-code: Would not this be more clear : "linear problem could not be solved, and infeasibility analysis could not help"

src/solver/infeasible-diag/report.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+void InfeasibleProblemReport::append(const std::string& constraintName, double value)
+{
+    // TODO check prior existence of constraintName
+    mConstraints.push_back({constraintName, value});
+}
+
+void InfeasibleProblemReport::printLargest(unsigned int n)
+{
+    std::sort(std::begin(mConstraints), std::end(mConstraints), compareSlackSolutions);
```
guilpier-code: Maybe missing some introductory message : "Infeasibility analysis gives following results :"

src/solver/infeasible-diag/report.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+namespace Optimization
+{
+void InfeasibleProblemReport::append(const std::string& constraintName, double value)
+{
+    // TODO check prior existence of constraintName
+    mConstraints.push_back({constraintName, value});
+}
+
+void InfeasibleProblemReport::printLargest(unsigned int n)
```
guilpier-code: **printLargest** : why not make this function's parameter a **InfeasibleProblemReport** member or constant ?
And in that case, we could call this method **print**

src/solver/cmake/infeasible-diag.cmake
```diff
@@ -0,0 +1,12 @@
+set(SRC_INFEASIBLE_DIAG
```
guilpier-code: Of course, if we change infeasible into something else, like analysis for instance, we should also change the name of the directory containing these files

guilpier-code (changes requested):

src/solver/infeasible-diag/problem.cpp
flomnes: With the right regex, we will match at least 1 constraint (hydro levels for example), even if the user has not defined any binding constraints.

This check is added for extra safety

src/solver/infeasible-diag/report.cpp
flomnes: I don't understand this remark, can you elaborate ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: Name: I agree and will rename this class to **InfeasibleProblemAnalysis**.

I don't understand the 2nd remark.

src/solver/infeasible-diag/report.cpp
flomnes: done

src/solver/cmake/infeasible-diag.cmake
flomnes: ok

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
guilpier-code: Sorry, let me be more clear : you pass the regex pattern **"(.)*"** to the constructor of **InfeasibleProblemAnalysis**.
This pattern should be a data of the class, and not be passed to this constructor, whatever this pattern is.

src/solver/infeasible-diag/report.cpp
guilpier-code: Sure : somewhere else in the code, you call : 
`printLargest(10);`
but the **n** (that is **10** here) that you pass, should it not be a inner data of the class **InfeasibleProblemReport** ?
I mean  : does the caller of **printLargest** have to known  that the function prints up to 10 incorrect constraints ?
And if you make this **n** an inner data, **printLargest** should be renamed **print()**

src/solver/infeasible-diag/problem.cpp
guilpier-code: ok, that said, user will probably not know what is a **pre-selected** constraint.
So the message will be obscure to him.
Maybe the message could be : "Cannot generate infeasibility report: no incorrect constraint found"

src/solver/infeasible-diag/report.cpp
flomnes: OK, understood & accepted

src/solver/infeasible-diag/problem.cpp
flomnes: I suggest: "no constraint have been selected". The message you suggest is misleading, it implies that the problem is feasible, even if it isn't.

src/solver/cmake/infeasible-diag.cmake
flomnes: done

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: Why not provide it to the constructor ?

flomnes:
Contained in https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/431

PR #430, opened 2021-10-26
Fix segfault when thermal clusters are disabled

### LINKED ISSUE #415: Calculation issues no output despite the fact that no error message is issued
**Description**
When launching the attached study (simulation for only MC year 348 as configured in the playlist), the simulation is apparently performed, but no output of the simulation is issued. The output folder only contains pre-processing data (like the ts-number folder). No error message is visible in the logs of the simulation.

Link to download the study: [Post'n'Go ](https://postngo.rte-france.com/?lang=fr&a=d&i=wkr1r9BRy9EoHymhEDse )

**OS and version**
 - OS: Windows
 - Antares Simulator version: 8.1

flomnes:
Return the number of **enabled** clusters in an area.


flomnes:
Replaced by #472

PR #431, opened 2021-10-27
Infeasible problem analysis: pretty print

flomnes:


src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -0,0 +1,89 @@
...
+                constraint->SetCoefficient(slack, 1.);
+                mSlackVariables.push_back(slack);
+            }
+
+            if (constraint->ub() != infinity)
+            {
+                MPVariable* slack;
+                slack = mSolver->MakeNumVar(0, infinity, constraint->name() + "::up");
+                constraint->SetCoefficient(slack, -1.);
+                mSlackVariables.push_back(slack);
```
flomnes, line 47: Careful with `push_pack`. Use `std::vector<>::reserve` ?

flomnes (commented):
Display which parameter could be wrong

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -338,12 +347,19 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, uint numSpace, int NumIn
...
             logs.info() << " Solver: Safe resolution failed";
         }
-        logs.error() << "Infeasible linear problem encountered. Possible causes:";
-        logs.error() << "* binding constraints,";
-        logs.error() << "* last resort shedding status,";
-        logs.error() << "* negative hurdle costs on lines with infinite capacity,";
-        logs.error() << "* Hydro reservoir impossible to manage with cumulative options \"hard "
-                        "bounds without heuristic\"";
+
+        Optimization::InfeasibleProblemAnalysis analysis(&Probleme, ".+::.+");
```
guilpier-code: the regex pattern **".+::.+"** should be an internal data of **InfeasibleProblemAnalysis**, and so should not be passed to its constructor.
Clients of **InfeasibleProblemAnalysis** should not have to know about this regex pattern.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -338,12 +347,19 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, uint numSpace, int NumIn
...
+        Optimization::InfeasibleProblemReport report;
+        try
+        {
+            report = analysis.produceReport(10);
```
guilpier-code: same thing as above : **10** should be an internal data of **InfeasibleProblemReport** and should not be passed to **produceReport** as argument, neither should it be passed to the constructor of **InfeasibleProblemReport**.
Clients of **InfeasibleProblemReport** should not have to know that the 10 first problematic constraints are spotted and raised to the user.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -338,12 +347,19 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, uint numSpace, int NumIn
...
+        }
+        catch (const std::logic_error& ex)
+        {
+            logs.error() << ex.what();
+        }
+
+        report.prettyPrint();
```
guilpier-code: Should not we have, instead : 
```
        Optimization::InfeasibleProblemAnalysis analysis(&Probleme, ".+::.+");
        try
        {
            Optimization::InfeasibleProblemReport  report = analysis.produceReport(10);
            report.prettyPrint();
        }
        catch (const Optimization::SlackVariablesEmptyr& ex)
        {
            logs.error() << ex.what();
        }
        ....
```

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -0,0 +1,89 @@
...
+    if (mSlackVariables.empty())
+        throw std::logic_error(
+          "Cannot generate infeasibility report: no constraints have been selected");
+    buildObjective();
+    const MPSolver::ResultStatus s = Solve();
+    if (s != MPSolver::OPTIMAL && s != MPSolver::FEASIBLE)
+        throw std::domain_error(
+          "Linear problem could not be solved, and infeasibility analysis could not help");
+
+    InfeasibleProblemReport r;
```
guilpier-code: **r** should be renamed into **report**

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+InfeasibleProblemReport InfeasibleProblemAnalysis::produceReport(
+  std::size_t nbSlackVariablesInReport)
+{
+    addSlackVariables();
+    if (mSlackVariables.empty())
+        throw std::logic_error(
+          "Cannot generate infeasibility report: no constraints have been selected");
+    buildObjective();
+    const MPSolver::ResultStatus s = Solve();
```
guilpier-code: **s** (status) should be renamed into something more expressive

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -0,0 +1,93 @@
...
+    if (s != MPSolver::OPTIMAL && s != MPSolver::FEASIBLE)
+        throw ProblemResolutionFailed(
+          "Linear problem could not be solved, and infeasibility analysis could not help");
+
+    InfeasibleProblemReport r;
+    for (const MPVariable* slack : mSlackVariables)
+    {
+        r.append(slack->name(), slack->solution_value());
+    }
```
guilpier-code: Appending slack variables to report should be done in the report constructor, so we should pass **mSlackVariables** to it.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+namespace Optimization
+{
+void InfeasibleProblemReport::append(const std::string& constraintName, double value)
+{
+    // TODO check prior existence of constraintName
+    mConstraints.emplace_back(constraintName, value);
+}
+
+void InfeasibleProblemReport::rawPrint() const
```
guilpier-code: **rawPrint** does not seem to be called. If so, we should remove it.

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    r.trimTo(nbSlackVariablesInReport);
```
guilpier-code: Does the analysis problem have to know that report trims slack variables ?
I recommend to make  **trimTo** a private method of the report and call it in its **prettyPrint** function.

src/solver/infeasible-problem-analysis/report.h
```diff
@@ -0,0 +1,29 @@
...
+{
+class InfeasibleProblemReport
+{
+public:
+    using SlackSolution = std::pair<std::string, double>;
+    InfeasibleProblemReport() = default;
+    void append(const std::string& constraintName, double value);
+    void trimTo(std::size_t nbSlackVariables);
+    void rawPrint() const;
+    void prettyPrint();
```
guilpier-code: For me, only the report **constructor** and the **prettyPrint** method should be public. The other functions should be private. **append** should called in the constructor, **rawPrint** seems to be never called,  **trimTo** should be called in **prettyPrint**.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+void InfeasibleProblemReport::trimTo(std::size_t n)
+{
+    std::sort(std::begin(mConstraints), std::end(mConstraints), compareSlackSolutions);
+
+    if (n > mConstraints.size())
+        n = mConstraints.size();
+
+    mConstraints.resize(n);
```
guilpier-code: Could possibly be : 
```
 mConstraints.resize(n > mConstraints.size() ? mConstraints.size() : n);
```

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -0,0 +1,75 @@
...
+    mConstraints.emplace_back(constraintName, value);
+}
+
+void InfeasibleProblemReport::rawPrint() const
+{
+    for (const auto& c : mConstraints)
+        Antares::logs.info() << c.getInput() << ": " << c.getSlackValue();
+}
+
+void InfeasibleProblemReport::prettyPrint()
```
guilpier-code: **prettyPrint** is a bit long, couldn't it be split into 2 parts (2 private methods calls) ? for instance : 
- **countTypes** updating 
- logging possible causes of infeasibility

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -119,7 +121,7 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* Pro
 
     for (Pdt = 0; Pdt < NombreDePasDeTempsPourUneOptimisation; Pdt++)
     {
-        int ts = study.runtime->weekInTheYear[numSpace] * 168 + Pdt;
+        int tsInYear = study->runtime->weekInTheYear[numSpace] * 168 + Pdt;
```
guilpier-code: For me **timeStepInYear** is better

src/solver/utils/ortools_utils.cpp
```diff
@@ -9,32 +9,35 @@
...
-        std::ostringstream oss;
-        oss << "x" << idxVar;
         double min_l = 0.0;
         if (bMin != NULL)
         {
             min_l = bMin[idxVar];
         }
         double max_l = bMax[idxVar];
-        auto x = solver->MakeNumVar(min_l, max_l, oss.str());
+        const MPVariable* x;
```
guilpier-code: **x** should be renamed into **var** for instance

guilpier-code (changes requested):

src/solver/utils/ortools_utils.cpp
```diff
@@ -51,9 +54,14 @@ void transferRows(MPSolver* solver, double* rhs, char* sens, int nbRow)
...
         }
-        std::ostringstream oss;
-        oss << "c" << idxRow;
-        MPConstraint* const ct = solver->MakeRowConstraint(bMin, bMax, oss.str());
+
+        std::string constraintName;
+        if (NomDesContraintes[idxRow].empty())
+          constraintName = "c" + std::to_string(idxRow);
+        else
+          constraintName = NomDesContraintes[idxRow];
```
flomnes: use curly braces here

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: ok

src/solver/infeasible-problem-analysis/problem.cpp
flomnes: ok

src/solver/infeasible-problem-analysis/problem.cpp
flomnes: ok

src/solver/infeasible-problem-analysis/report.h
flomnes: ok

src/solver/infeasible-problem-analysis/report.cpp
flomnes: I won't use a ternary, but change this. See code

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
flomnes: ok

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: ok, done

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: ok, done.

One idea could be to expose constraint types : bc_weekly, bc_daily, etc. And to allow the user to select what type of constraint they want to select. Not sure if useful though.

src/solver/infeasible-problem-analysis/report.cpp
flomnes: ok

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -0,0 +1,88 @@
...
+    {
+        if (c.extractItems() == 0)
+        {
+            return;
+        }
+        mTypes[c.getType()]++;
+    }
+}
+
+void InfeasibleProblemReport::prettyPrintImpl()
```
guilpier-code: For me, **prettyPrintImpl** is meaningless, **logViolatedContraints()** or something  like that would be better.

guilpier-code (changes requested):

src/solver/infeasible-problem-analysis/report.cpp
flomnes: done

guilpier-code (approved):

PR #432, opened 2021-10-27
NTC with multi time series : internal structures

guilpier-code:


src/libs/antares/study/area/links.cpp
```diff
@@ -125,19 +129,23 @@ void AreaLink::reverse()
...
+    memcpy(indirectCapacities[0], tmp, size);
     delete tmp;
 
-    data.markAsModified();
-    data.flush();
+    directCapacities.markAsModified();
+    indirectCapacities.markAsModified();
+
+    directCapacities.flush();
+    indirectCapacities.flush();
```
flomnes: Why do we need to swap memory content with deep copies ? Why not just swap pointers here ?

src/libs/antares/study/area/links.cpp
```diff
@@ -336,38 +344,58 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
-            // NTC
+            // NTC direct
             buffer.clear() << folder << SEP << link.with->id << SEP << "direct."
                            << study.inputExtension;
             ret = tmp.loadFromCSVFile(
                     buffer, 1, 8760, Matrix<>::optFixedSize | Matrix<>::optImmediate)
                   && ret;
-            link.data.pasteToColumn(fhlNTCDirect, tmp[0]);
+            link.directCapacities.pasteToColumn(0, tmp[0]);
+            link.directCapacities.markAsModified();
```
flomnes: Matrix `link.directCapacities` only contains link data. Therefore there is no need for a temporary matrix here.

src/libs/antares/study/area/links.cpp
```diff
@@ -377,34 +405,32 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
+            // Store data into link's data container
             for (int i = 0; i < HOURS_PER_YEAR; i++)
             {
-                link.data[fhlNTCDirect][i] = temp[fhlNTCDirect][i];
-                link.data[fhlNTCIndirect][i] = temp[fhlNTCIndirect][i];
-                link.data[fhlImpedances][i] = temp[fhlImpedances][i];
-                link.data[fhlHurdlesCostDirect][i] = temp[fhlHurdlesCostDirect][i];
-                link.data[fhlHurdlesCostIndirect][i] = temp[fhlHurdlesCostIndirect][i];
+                link.directCapacities[0][i] = tmpMatrix[0][i];
+                link.indirectCapacities[0][i] = tmpMatrix[1][i];
```
flomnes: Is this piece of code responsible for loading old studies ?  It cannot load new studies with more than NTC timeseries.

src/libs/antares/study/area/links.cpp
```diff
@@ -336,38 +344,58 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
 
         if (study.header.version < 320)
         {
-            link.data.resize(fhlMax, HOURS_PER_YEAR);
-            link.data.zero();
-            if (link.data.jit)
+            // gp : refactor this part : avoid code duplication
+            link.parameters.resize(fhlMax, HOURS_PER_YEAR);
```
flomnes: Consider a new function for old studies. This function is way too long

src/libs/antares/study/area/links.cpp
flomnes: See #439

src/libs/antares/study/area/links.h
```diff
@@ -100,6 +100,10 @@ class AreaLink final : public Yuni::NonCopyable<AreaLink>
     ** \brief Mark the data associated to the link as modified
     */
     void markAsModified() const;
+
+    void setPathToDataFile(const AnyString& path) { pathToDataFile = path; }
+    
+    bool loadDataFromCVSfile(uint loadOptions);
```
flomnes: ~~CVS~~ CSV

src/libs/antares/study/area/links.cpp
```diff
@@ -337,38 +346,58 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
-                link.data.markAsModified();
+                link.indirectCapacities.jit->options |= Matrix<>::optFixedSize;
+                link.indirectCapacities.markAsModified();
             }
 
             Matrix<double> tmp;
-            // NTC
+            // NTC direct
             buffer.clear() << folder << SEP << link.with->id << SEP << "direct."
                            << study.inputExtension;
```
flomnes: Check spec to use the right path

src/libs/antares/study/area/links.cpp
```diff
@@ -337,38 +346,58 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
             ret = tmp.loadFromCSVFile(
-                    buffer, 1, 8760, Matrix<>::optFixedSize | Matrix<>::optImmediate)
+                    buffer, 1, HOURS_PER_YEAR, Matrix<>::optFixedSize | Matrix<>::optImmediate)
                   && ret;
-            link.data.pasteToColumn(fhlNTCDirect, tmp[0]);
+            link.directCapacities.pasteToColumn(0, tmp[0]);
```
flomnes: There is no need to use a buffer. Load the file contents into `directMatrix` directly.

It wasn't possible previously, but now it is.

src/libs/antares/study/area/links.cpp
```diff
@@ -337,38 +346,58 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
+            // NTC indirect
             buffer.clear() << folder << SEP << link.with->id << SEP << "indirect."
                            << study.inputExtension;
             ret = tmp.loadFromCSVFile(
-                    buffer, 1, 8760, Matrix<>::optFixedSize | Matrix<>::optImmediate)
+                    buffer, 1, HOURS_PER_YEAR, Matrix<>::optFixedSize | Matrix<>::optImmediate)
                   && ret;
-            link.data.pasteToColumn(fhlNTCIndirect, tmp[0]);
+            link.indirectCapacities.pasteToColumn(0, tmp[0]);
+            link.indirectCapacities.markAsModified();
```
flomnes: Same as above: no need for a tmp buffer.

src/libs/antares/study/area/links.cpp
```diff
@@ -378,136 +407,95 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
             if (enabledModeIsChanged)
                 JIT::enabled = true; // Back to the previous loading mode.
         }
         else
         {
             buffer.clear() << folder << SEP << link.with->id << ".txt";
-            ret = link.data.loadFromCSVFile(buffer, fhlMax, HOURS_PER_YEAR, Matrix<>::optFixedSize)
-                  && ret;
+            link.setPathToDataFile(buffer);
+            ret = link.loadDataFromCVSfile(Matrix<>::optFixedSize | Matrix<>::optImmediate) && ret;
```
flomnes: Why do we need to add flag `Matrix<>::optImmediate` ?

src/libs/antares/study/area/links.cpp
```diff
@@ -591,23 +579,25 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
                 break;
             }
             }
         }
 
         // memory swap
-        link.data.flush();
+        link.parameters.flush();
+        link.directCapacities.flush();
+        link.indirectCapacities.flush();
```
flomnes: `link.flush();` does that in 1 line

src/solver/constraints-builder/cbuilder.cpp
```diff
@@ -87,15 +87,12 @@ bool CBuilder::update(bool applyCheckBox)
...
 
     for (auto i = pLink.begin(); i != pLink.end(); i++)
     {
-        // try to open the file
-        if (!(*i)->dataLink->loadFromCSVFile((*i)->ptr->data.jit->sourceFilename,
-                                             Data::fhlMax,
-                                             (uint)8760,
-                                             Matrix<>::optImmediate))
+        // Try to open the file
+        if (not (*i)->ptr->loadDataFromCVSfile(Matrix<>::optImmediate))
```
flomnes: You're de-referencing two pointers in a single instruction. This is dangerous since one of them could be `nullptr` or invalid.

---

It's unclear what `ptr` refers to, consider renaming this variable.

src/solver/constraints-builder/cbuilder.cpp
```diff
@@ -106,124 +103,107 @@ bool CBuilder::update(bool applyCheckBox)
...
-        // load the impedance
-        auto dataLink = k->dataLink;
 
+        // load the impedance
         // Can probably be improved (below) !!!
-        k->nImpedanceChanges = 0;
-        k->avgImpedance
-          = dataLink->entry[columnImpedance][0]; // or used a function created in matrix.h
+        (*i)->nImpedanceChanges = 0;
+        (*i)->avgImpedance = (*i)->ptr->parameters[columnImpedance][0];
```
flomnes: 2 de-reference in a single instruction. Consider using a variable for `*i`.

flomnes (changes requested):

src/libs/antares/study/area/links.cpp
guilpier-code: ok

src/libs/antares/study/area/links.cpp
guilpier-code: ok

src/libs/antares/study/area/links.cpp
guilpier-code: what do you mean ? (this is the path for direct capacities data file of a very old versions of the study)

src/libs/antares/study/area/links.cpp
guilpier-code: There's a problem otherwise : if we don't add this load option, the data is not loaded and code waits until it's the right time (jit : just in time) to load it (that is when we really use it). 
Because we want to update the new data structure (data is now split in 3 parts) now, we need to really load the old 8 columns matrix.
The way to update the new data structure without actually load the data might be possible, but at this stage of development, I keep it simple and load old data to construct the new one.

src/libs/antares/study/area/links.cpp
guilpier-code: you're right, I didn't call **flush()** on the link although I introduced this function in the link class

src/libs/antares/study/area/links.h
guilpier-code: done

src/solver/constraints-builder/cbuilder.cpp
guilpier-code: ok, **ptr** is a pointer to an object of type **AreaLink** (a study link actually).
For now, I do not rename that pointer, but I define a variable that makes this part of the code more clear.

src/solver/constraints-builder/cbuilder.cpp
guilpier-code: **(*i)** is now replaced with **linkInfo**

src/libs/antares/study/area/links.cpp
guilpier-code: Yes, here we load old studies (between 320 and 630), where NTC have a single timeseries per link

src/libs/antares/study/area/links.cpp
guilpier-code: A possible solution could be something like : 
```c++
linkLoaderInterface* linkLoader; // Imagine a smart pointer here, not a regular pointer

// A factory to build the right link loader
if (study.header.version < 320)
    linkLoader = new linkLoader_under320(...)
else if (study.header.version < 630)
    linkLoader = new linkLoader_320_to_630(...)
else if (study.header.version < 820)
   linkLoader = new linkLoader_630_to_820(...)
else
  linkLoader = new linkLoader_820(...)

linkLoader.run();
```
To be discussed...

src/libs/antares/study/area/links.cpp
guilpier-code: ok, done

src/libs/antares/study/area/links.cpp
guilpier-code: I extracted functions instead

src/libs/antares/study/area/links.cpp
```diff
@@ -83,12 +85,14 @@ void AreaLink::detach()
...
+    directCapacities.reset(1, HOURS_PER_YEAR, true);
+    indirectCapacities.reset(1, HOURS_PER_YEAR, true);
 
-    for (uint i = 0; i != data.height; ++i)
+    for (uint i = 0; i != HOURS_PER_YEAR; ++i)
     {
-        data[fhlNTCDirect][i] = 1.;
-        data[fhlNTCIndirect][i] = 1.;
+        directCapacities[0][i] = 1.;
+        indirectCapacities[0][i] = 1.;
```
flomnes, line 95: Use `void Matrix<T, ReadWriteT>::fillColumn(uint x, const T& value)` instead.

src/libs/antares/study/area/links.cpp
```diff
@@ -289,6 +291,127 @@ static bool AreaLinksInternalLoadFromProperty(Study& study,
...
 } // anonymous namespace
 
+void logLinkDataCheckError(Study& study, AreaLink& link)
+{
+    logs.error() << "Link (" << link.from->name << "/" << link.with->name
+        << "): Invalid values";
+    study.gotFatalError = true;
+}
+
+bool linkLoadTimeSeries_version_under_320(AreaLink& link, const AnyString& folder, Study& study)
```
flomnes, line 301: Mark as `static`

src/libs/antares/study/area/links.cpp
```diff
@@ -289,6 +291,127 @@ static bool AreaLinksInternalLoadFromProperty(Study& study,
...
+        << study.inputExtension;
+    Matrix<double> tmp;
+    ret = tmp.loadFromCSVFile(buffer, 1, HOURS_PER_YEAR, loadOptions) && ret;
+    link.parameters.pasteToColumn(fhlImpedances, tmp[0]);
+    link.parameters.markAsModified();
+
+    return ret;
+}
+
+bool linkLoadTimeSeries_version_320_to_630(AreaLink& link, const AnyString& folder)
```
flomnes, line 355: Mark as `static`

src/libs/antares/study/area/links.cpp
```diff
@@ -289,6 +291,127 @@ static bool AreaLinksInternalLoadFromProperty(Study& study,
...
+
+    return ret;
+}
+
+/*
+            buffer.clear() << folder << SEP << link.with->id << ".txt";
+            link.setPathToDataFile(buffer);
+            ret = link.loadDataFromCSVfile(Matrix<>::optFixedSize | Matrix<>::optImmediate) && ret;
+*/
+bool linkLoadTimeSeries_version_630_to_820(AreaLink& link, const AnyString& folder)
```
flomnes, line 403: Mark as `static`

PR #433, opened 2021-10-28
Fix Windows CI

flomnes:
Remove Chocolatey's CPACK. It interferes with CMake's CPACK


flomnes:
Wait for virtual-environments to be back to normal. Do not mess with upstream environment.

PR #437, opened 2021-11-02
CentOS 7: upgrade devtoolset-7 -> devtoolset-9

### LINKED ISSUE #436: Use gcc 8-9-10 on CentOS 7 ?
Using `devtoolset-<n>` would allow to use a more recent version of gcc, currently 7.3 (2015).

Would binaries be supported on CALIN ? To be tested.

flomnes:
GCC 9 is compatible with recent C++ features such as std::filesystem, etc.

### DIFF
--- a/docker/centos7-basic-requirements
+++ b/docker/centos7-basic-requirements
@@ -9,8 +9,8 @@ RUN yum update -y
 RUN yum install -y epel-release 
 
 # Install requirements
-RUN yum install -y git redhat-lsb-core gcc gcc-c++ make wget centos-release-scl scl-utils rpm-build &&\
-    yum install -y cmake3 devtoolset-7 &&\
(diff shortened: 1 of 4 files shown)


flomnes:
- [ ] Push image on internal docker

docker/centos7-basic-requirements
```diff
@@ -9,8 +9,8 @@ RUN yum update -y
 RUN yum install -y epel-release 
 
 # Install requirements
-RUN yum install -y git redhat-lsb-core gcc gcc-c++ make wget centos-release-scl scl-utils rpm-build &&\
```
flomnes, line 12: `gcc` and `gcc-c++` are not needed : they come from `devtoolset-9`.

guilpier-code (approved):

PR #439, opened 2021-11-03
Introduce and use Matrix<>::swap

flomnes:


src/libs/antares/array/matrix.hxx
```diff
@@ -1552,6 +1552,18 @@ inline void Matrix<T, ReadWriteT>::copyFrom(const Matrix<U, V>* rhs)
         copyFrom(*rhs);
 }
 
+template<class T, class ReadWriteT>
+template<class U, class V>
+void Matrix<T, ReadWriteT>::swap(Matrix<U, V>& rhs)
+{
+  static_assert(Yuni::Static::Type::StrictlyEqual<T, U>::Yes);
+  using std::swap;
```
guilpier-code, line 1560: You don't include **algorithm.h** for std::swap, meaning that it's included elsewhere... Strange

src/libs/antares/array/matrix.hxx
```diff
@@ -1552,6 +1552,18 @@ inline void Matrix<T, ReadWriteT>::copyFrom(const Matrix<U, V>* rhs)
         copyFrom(*rhs);
 }
 
+template<class T, class ReadWriteT>
+template<class U, class V>
+void Matrix<T, ReadWriteT>::swap(Matrix<U, V>& rhs)
```
guilpier-code: The type of **rhs** is different from local matrix's type. How can it be ?

src/libs/antares/array/matrix.hxx
```diff
@@ -1552,6 +1552,18 @@ inline void Matrix<T, ReadWriteT>::copyFrom(const Matrix<U, V>* rhs)
...
+{
+  static_assert(Yuni::Static::Type::StrictlyEqual<T, U>::Yes);
+  using std::swap;
+  swap(this->width, rhs.width);
```
guilpier-code, line 1561: If we swap matrices, their width and height must be equal, right ?
If not, we have a problem, don't you think ?

guilpier-code (changes requested):

src/libs/antares/array/matrix.hxx
flomnes: I have no idea how `std::swap` is defined here. However I included header `utility` for safety.

src/libs/antares/array/matrix.hxx
flomnes: `swap` should always succeed. Adding an assertion could potentially make it fail.

swapping matrices of different size makes sense. It is up to the user to check for size consistency.

guilpier-code (approved):

PR #447, opened 2021-11-09
Add data structures for link TS numbers

flomnes:


src/solver/simulation/sim_allocation_tableaux.cpp
```diff
@@ -156,6 +156,16 @@ void SIM_AllocationTableaux()
             }
         }
     }
+    NumeroChroniquesTireesParInterconnexion
+      = (NUMERO_CHRONIQUES_TIREES_PAR_INTERCONNEXION**)MemAlloc(study.maxNbYearsInParallel
+                                                                * sizeof(void*));
+    const auto intercoCount = study.runtime->interconnectionsCount;
```
guilpier-code: Why is **intercoCount** retrieved from **study.runtime** whereas it has not to be refreshed ? It's a study fixed input : **study.areas.areaLinkCount()**

flomnes: done

guilpier-code (approved):

PR #449, opened 2021-11-09
NTC: use TS numbers in optimization problem

flomnes:


src/solver/simulation/sim_calcul_adequation.cpp
```diff
@@ -274,11 +274,11 @@ void SIM_RenseignementValeursPourTouteLAnnee(const Antares::Data::Study& study,
     }
 }
 
-void SIM_RenseignementProblemeHoraireAdequation(long Heure)
+void SIM_RenseignementProblemeHoraireAdequation(long Heure, uint numSpace)
```
guilpier-code: **numSpace** is not a great name, and I guess it is not what you think it is : it's the number of the year in the set of years run in parallel. It is not an absolute number in the range of years of the simulation.
I guess what you're after is more :
`state[numSpace].year`
which is updated in solver.hxx (184)
So, when you use it to get a NTC ts number, use the previous instead.

In fact, a draft simulation is never run in parallel, that is the number of run parallel years is 1.
The reason for this is that draft computations are fast anyway.
So, in that case,  **numSpace** is always 0.

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -535,13 +535,15 @@ void SIM_RenseignementProblemeHebdo(PROBLEME_HEBDO& problem,
...
                 auto& lnk = *(studyruntime.areaLink[k]);
-                /* gp : these asserts will be done in another way later
-                assert((uint)indx < lnk.data.height);
-                assert((uint)fhlNTCDirect < lnk.data.width);
-                assert((uint)fhlNTCIndirect < lnk.data.width);
-                */
-                ntc->ValeurDeNTCOrigineVersExtremite[k] = lnk.directCapacities[0][indx];
-                ntc->ValeurDeNTCExtremiteVersOrigine[k] = lnk.indirectCapacities[0][indx];
+                const int tsIndex
+                  = NumeroChroniquesTireesParInterconnexion[numSpace][k].TransmissionCapacities;
```
guilpier-code, line 539: See remark in **sim_calcul_adequation.cpp**

guilpier-code (changes requested):

src/solver/simulation/sim_calcul_adequation.cpp
flomnes: ok

src/solver/simulation/sim_calcul_economique.cpp
flomnes: As discussed, I think `numSpace` is right here.

src/solver/simulation/sim_calcul_adequation.cpp
guilpier-code: my bad, **NumeroChroniquesTireesParInterconnexion** is updated before running each set of parallel years and its dimension is the number of years in a set of parallel years

flomnes: I'll keep it that way for Adequacy/Economy, and remove the unnecessary argument for Adequacy Draft, since `numSpace` is always 0 in this mode.

flomnes (commented):
I just re-factored `SIM_RenseignementProblemeHoraireAdequation` a bit and made it explicit that `numSpace=0` in this context.

I changed the ts-numbers only in Eco/Adq

guilpier-code (approved):

PR #463, opened 2021-11-16
Check consistency for NTC timeseries (solver only)

flomnes:


src/libs/antares/study/area/links.cpp
```diff
@@ -469,34 +477,44 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
-            for (int h = 0; h < HOURS_PER_YEAR; h++)
-            {
-                if (indirectCapacities[h] < 0. || indirectCapacities[h] + loopFlow[h] < 0)
+                // Checks on indirect capacities
+                for (int h = 0; h < HOURS_PER_YEAR; h++)
                 {
-                    logLinkDataCheckError(study, link);
-                    return false;
+                    if (indirectCapacities[h] < 0. || indirectCapacities[h] + loopFlow[h] < 0)
+                    {
```
flomnes, line 512: This should be in the base branch already.

flomnes: No, this is in the right place, my bad

guilpier-code (approved):

PR #464, opened 2021-11-17
fix TS number distribution for inter-modal

guilpier-code:
An inter-modal production/load gets 0 as time-series number if it owns only 1 time-series


guilpier-code:
re-opening this PR

PR #465, opened 2021-11-17
fix TS number distribution for inter-modal

guilpier-code:
An inter-modal production/load gets 0 as time-series number if it owns only 1 time-series


flomnes:
As discussed, this will need a bit more work. The final goal is
1. To have a data structure that will keep all number of TS for all areas & modes
2. To correct ts-numbers **after the draw**, so that is will be independent of intra/inter modal. This will avoid code duplication.

PR #471, opened 2021-11-22
NTC - scenario builder : reading scenariobuilder.dat

guilpier-code:


src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -488,6 +517,97 @@ bool renewableTSNumberData::reset(const Study& study)
...
+}
+
+void ntcTSNumberData::saveToINIFile(const Study& study, Yuni::IO::File::Stream& file) const
+{
+    // Prefix
+    CString<512, false> prefix;
+    prefix += get_prefix();
+
+    if (!pArea)
+        return;
```
flomnes: This could be the 1st instruction

src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -488,6 +517,97 @@ bool renewableTSNumberData::reset(const Study& study)
     return true;
 }
 
+
+// ================================
+// Transmission capacities ...
+// ================================
+bool ntcTSNumberData::reset(const Study& study)
+{
+    assert(&study != nullptr);
```
flomnes, line 543: This assert is probably useless

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -68,6 +69,7 @@ void Rules::saveToINIFile(Yuni::IO::File::Stream& file) const
         {
             thermal[i].saveToINIFile(study_, file);
             renewable[i].saveToINIFile(study_, file);
+            linksNtc[i].saveToINIFile(study_, file);
```
flomnes: linksNTC

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -257,6 +271,41 @@ void Rules::readHydroLevels(const AreaName::Vector& splitKey, String value, bool
     hydroLevels.set(area->index, year, val);
 }
 
+Data::AreaLink* Rules::getLink(const AreaName& areaName, const AreaName& targetAreaName, bool updaterMode)
```
flomnes: `areaFromName`, `areaToName`

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -257,6 +271,41 @@ void Rules::readHydroLevels(const AreaName::Vector& splitKey, String value, bool
...
+    if (!area)
+        return;
+
+    const Data::Area* targetArea = getArea(targetAreaName, updaterMode);
+    if (!targetArea)
+        return;
+
+    AreaLink* link = getLink(areaName, targetAreaName, updaterMode);
+    if (!link)
+        return;
```
flomnes: Variables `targetArea` and `area` are unused.

src/libs/antares/study/scenario-builder/rules.h
```diff
@@ -112,7 +112,7 @@ class Rules final : private Yuni::NonCopyable<Rules>
     //! Wind
     windTSNumberData wind;
 
-    // gp : change these 2 arrays into std containers, like std::vector 
+    // gp : change these 3 arrays into std containers, like std::vector
```
flomnes: That would be nice

flomnes (changes requested):
`Area::buildLinksIndexes` needs to be called from `AreaLink::reverse()`, for both areas.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
guilpier-code: True. Done.

guilpier-code:
> `Area::buildLinksIndexes` needs to be called from `AreaLink::reverse()`, for both areas.

I agree, done

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: ok, done

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: Are they ? There is an **if** statement involving these areas just after their definition.
That said, I can rename them into **fromArea** and **ToArea** --> Done

src/libs/antares/study/scenario-builder/rules.h
guilpier-code: I tried to do it with **std::vector**. I would have liked to do it with **std::array** but I don't see how to initialize the arrays size at declaration, as it seems to be mandatory.

Furthermore, trying to resize a **vector\<ntcTSNumberData\>** (see **Rules::reset()**) is forbidden by the compiler because all classes derived from **dataInterface** (like **ntcTSNumberData**) are non copyable (see class **dataInterface**).
We could have a vector\<ntcTSNumberData*\> instead, but we are compelled to delete (when needed) each element object manually, loosing the benefit of letting the program do it for us.

Finally, I do nothing regarding this point.
Please feel free to suggest something about it.

flomnes:
Careful, Sonar found 2 bugs

src/libs/antares/study/scenario-builder/TSnumberData.cpp
guilpier-code: ok, you probably refer to the only call of **ntcTSNumberData::reset()** made in **Rules::reset()**, where this assert is already called. But what if reset is called somewhere else someday ? We  could forget to call this assert then.
For me, this assert should be called in **ntcTSNumberData::reset()**.
Note that this assert is also called in **ntcTSNumberData** siblings (I mean her sisters, like **thermalTSNumberData**)

src/libs/antares/study/area/list.cpp
```diff
@@ -1535,6 +1535,9 @@ bool AreaList::renameArea(const AreaName& oldid, const AreaName& newid, const Ar
         assert(oldCount == a.links.size() and "We must have the same number of items in the list");
 #endif
     });
+    
+    // gp : do we really need to rebuild the area's link indexes here ?
+    area->buildLinksIndexes();
```
flomnes: Since `a` in the loop above could take the value `area`, its links could be changed at line 1532. So I think it is necessary to rebuild index here.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
flomnes: Ok

src/libs/antares/study/scenario-builder/rules.h
flomnes: Maybe `std::vector<std::shared_ptr<ntcTSNumberData>>` ? This can be addressed in another PR anyway.

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: ok, done

flomnes:
2 actions required before merging:
- Fix conflicts with the base branch
- Fix possible bug found by Sonar

src/libs/antares/study/area/links.cpp
```diff
@@ -776,30 +782,23 @@ bool AreaLinksSaveToFolder(const Area* area, const char* const folder)
...
-        f->detachLinkFromID(l->with->id);
-        f->detachLinkFromID(l->from->id);
-    }
+    assert(link->from);
+    assert(link->with);
 
-    Area* t = l->with;
-    if (t && !t->links.empty())
+    Area* areaFrom = link->from;
+    if (areaFrom && !areaFrom->links.empty())
```
guilpier-code, line 795: Note that if a link is removed, it is removed it from the origin area's list of links, and not from the extremity area's list of links

flomnes: of course

src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -488,6 +517,94 @@ bool renewableTSNumberData::reset(const Study& study)
...
+    for (auto i = pArea->links.begin(); i != pArea->links.end(); ++i)
+    {
+        auto* link = i->second;
+        for (uint y = 0; y != pTSNumberRules.height; ++y)
+        {
+            const uint val = pTSNumberRules[link->indexForArea][y];
+            // Equals to zero means 'auto', which is the default mode
+            if (!val)
+                continue;
+            file << prefix << pArea->id << "," << i->first << "," << y << " = " << val << "\n";
```
flomnes, line 580: This PR is called "reading scenariobuilder.dat". Consider renaming it to "read & save" since this bit of code saves.

src/libs/antares/study/scenario-builder/TSnumberData.h
```diff
@@ -264,6 +279,54 @@ inline CString<512, false> renewableTSNumberData::get_prefix() const
...
+        uint index = linksIndexMap.at(link);
+        return pTSNumberRules[index][year];
+    }
+
+    return 0;
+}
+
+inline CString<512, false> ntcTSNumberData::get_prefix() const
+{
+    return "tc,";
```
flomnes: Specific prefix to be discussed with Antares POs.

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -109,17 +100,28 @@ bool Rules::reset()
...
+
+
     return true;
 }
 
 Data::Area* Rules::getArea(const AreaName& areaname, bool updaterMode)
 {
     Data::Area* area = study_.areas.find(areaname);
-    if (!area)
+    if (!area && not updaterMode)
```
flomnes, line 121: Consider de-referencing `area` **after** checking it is not `nullptr`, not before.

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -257,6 +259,40 @@ void Rules::readHydroLevels(const AreaName::Vector& splitKey, String value, bool
     hydroLevels.set(area->index, year, val);
 }
 
+Data::AreaLink* Rules::getLink(const AreaName& fromAreaName, const AreaName& toAreaName, bool updaterMode)
+{
+    Data::AreaLink* link = study_.areas.findLink(fromAreaName, toAreaName);
+    if (!link && not updaterMode)
```
flomnes, line 274: Same remark as above.

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+// =================
+// Tests on Load
+// =================
+BOOST_AUTO_TEST_CASE(on_area2_and_on_year_18__load_TS_number_11_is_chosen__reading_OK)
+{
+	AreaName yearNumber = "18";
+	String tsNumber = "11";
+	AreaName::Vector splitKey = { "l", "area 2", yearNumber };
+	my_rule.readLine(splitKey, tsNumber, false);
```
flomnes: There is no way to know whether `readLine` was successful. Only a warning is displayed in the logs if it wasn't.

Consider adding a return value to `readLine`. You'd need to also add a return value to `getArea`, `getLink`, etc. Then test the return value here.

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+	BOOST_CHECK_EQUAL(my_rule.load.get_value(yearNumber.to<uint>(), area_2->index), tsNumber.to<uint>());
+
+	my_rule.apply();
```
flomnes: Same remark here. You'd need `ApplyToMatrix` to return a status. Test the result here.

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
```diff
@@ -0,0 +1,388 @@
...
+	BOOST_CHECK_EQUAL(link_13->timeseriesNumbers[0][yearNumber.to<uint>()], tsNumber.to<uint>() - 1);
+}
+
+BOOST_AUTO_TEST_CASE(on_link_area2_area3_and_on_year_19__ntc_TS_number_6_is_chosen__reading_OK)
+{
+	AreaName yearNumber = "19";
+	String tsNumber = "6";
+	AreaName::Vector splitKey = { "tc", "area 2", "area 3", yearNumber };
```
flomnes: What happens if we invert "area 2" and "area 3" ? Probably `readLine` fails because the link is not found. Could we maybe try to call `getLink` with inverted arguments in this case, as a fallback ? Or do we enforce strict rules ?

flomnes (changes requested):

src/libs/antares/study/scenario-builder/TSnumberData.cpp
guilpier-code: This PR was originally meant to cover the reading the NTC time series from the scenario builder file.
But eventually, I added : 
- Unit tests on reading (from the scenario file) TS numbers used for any production means (load, thermal, wind, ...).
- the (small) part regarding the NTC time series saving in this file.

So it seems to me that : 
- This PR is mainly dedicated to reading the scenario builder.
- It is big enough not to add the new coming tests on saving data on the scenario builder file.

For this reasons, I recommend not to change the name of this PR. Unit tests on savings should be treated by another PR.

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: **area** is never de-referenced

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: **link** is never de-referenced

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
guilpier-code: ok, done

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
guilpier-code: ok, done

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
guilpier-code: As **scenariobuilder.dat** is always created from a study save/save as, a line associated to a link is built with : 
- first the origin area
- then the pointed area
So we should keep strict rules, to be compliant with the way the existing code see things

src/libs/antares/study/scenario-builder/TSnumberData.h
guilpier-code: ok, done

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
guilpier-code: As `readLine` now returns a bool, we could add some more tests that ensures that this function returns **false** in situations where it has to.

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
guilpier-code: As `apply` now returns a bool, we could add some more tests that ensures that this function returns false in situations where it has to.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -144,13 +146,16 @@ static void ApplyToMatrix(uint& errors,
                         logs.warning() << "scenario-builder: " << logprefix
                                        << "value out of bounds for the year " << (y + 1);
                 }
+                status_to_return = status_to_return && false;
```
flomnes: - `true && false == false`
- `false && false == false`

You can safely replace this line with `status_to_return = false;`

I think that we should return `false` whenever there is an error, not only if `errors > maxErrors`.

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -156,25 +156,27 @@ void Rules::readThermalCluster(const AreaName::Vector& splitKey, String value, b
         {
             string clusterId = (area->id).to<string>() + "." + clustername.to<string>();
             disabledClustersOnRuleActive[clusterId].push_back(year);
+            return false;
```
flomnes, line 159: I don't think we should return in this case. This error is not fatal, it only gives a warning message.

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -156,25 +156,27 @@ void Rules::readThermalCluster(const AreaName::Vector& splitKey, String value, b
...
-void Rules::readRenewableCluster(const AreaName::Vector& splitKey, String value, bool updaterMode)
+bool Rules::readRenewableCluster(const AreaName::Vector& splitKey, String value, bool updaterMode)
 {
     const AreaName& areaname = splitKey[1];
     const uint year = splitKey[2].to<uint>();
     const ClusterName& clustername = splitKey[3];
 
     if (not study_.parameters.renewableGeneration.isClusters())
-        return;
+        return false;
```
flomnes, line 172: We can keep rules related to renewable clusters even if `isClusters() == false`. I think returning `true` would be better.

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -190,73 +192,80 @@ void Rules::readRenewableCluster(const AreaName::Vector& splitKey, String value,
         {
             string clusterId = (area->id).to<string>() + "." + clustername.to<string>();
             disabledClustersOnRuleActive[clusterId].push_back(year);
+            return false;
```
flomnes, line 195: Same remark as with thermal clusters.

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
flomnes: Ok, I agree.

flomnes (changes requested):

src/libs/antares/study/scenario-builder/TSnumberData.cpp
flomnes: ok

src/libs/antares/study/scenario-builder/rules.cpp
flomnes: In other words, this line makes a non-fatal error become fatal. This is probably not what we want.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
guilpier-code: I agree with the first part of your remark.
About the `errors > maxErrors` thing, this is not what I understand the code does : for me the code warns up to 20 errors (maxErrors is 20), and stops to warn after 20 errors. So, at most, 20 warnings are raised.
If at least 1 error is detected, **false** is returned.

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: Fortunately, this error is not fatal, as the caller of **readLine** function (see **Sets::internalLoadFromINIFile(...)**) does not use its returned status.

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: As the caller of **readLine** function (see **Sets::internalLoadFromINIFile(...)**) does not use **readLine**'s returned status, this little error has no consequence. So I think we should keep the code as it is here.
To be discussed.

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: Same reply as above

flomnes (approved):

PR #473, opened 2021-11-24
Use average hourly NTC values for quadratic optimization

flomnes:


src/solver/simulation/adequacy.cpp
```diff
@@ -371,13 +371,24 @@ AvgExchangeResults* Adequacy::callbackRetrieveBalanceData(Data::Area* area)
     return balance;
 }
 
+AvgNTC* Adequacy::callbackRetrieveAverageNTC(Data::AreaLink* link)
```
guilpier-code: The 2 callbacks and the **simulationEnd()** function are the same in the **adequacy** and **economy** mode.
Besides, I am not sure that the callback system is needed here.
So maybe we could call directly  : 
- `variables.retrieveResultsForArea<Variable::Economy::VCardBalance>`
- `variables.retrieveResultsForLink<Variable::Economy::VCardNTC>`
inside function **RecalculDesEchangesMoyens**, and not pass callback as arguments down to this latter function.

src/solver/simulation/economy.cpp
```diff
@@ -245,13 +245,25 @@ AvgExchangeResults* Economy::callbackRetrieveBalanceData(Data::Area* area)
     return balance;
 }
 
+AvgNTC* Economy::callbackRetrieveAverageNTC(Data::AreaLink* link)
```
guilpier-code: Same remark as in the **adequacy.cpp**

src/solver/simulation/economy.cpp
```diff
@@ -245,13 +245,25 @@ AvgExchangeResults* Economy::callbackRetrieveBalanceData(Data::Area* area)
...
-        CallbackBalanceRetrieval callback;
-        callback.bind(this, &Economy::callbackRetrieveBalanceData);
-        PerformQuadraticOptimisation(study, *pProblemesHebdo[0], callback, pNbWeeks);
+        CallbackBalanceRetrieval callbackBalance;
+        callbackBalance.bind(this, &Economy::callbackRetrieveBalanceData);
+
+        CallbackNTCRetrieval callbackNTC;
+        callbackNTC.bind(this, &Economy::callbackRetrieveAverageNTC);
+
+        PerformQuadraticOptimisation(
```
guilpier-code: **PerformQuadraticOptimisation** is not a good name here : at this level (of abstraction), what we want to do is compute the average flow, right ? Quadratic optimization is only a means to do it.
So I would reserve "quadratic optimization" for a lower level.

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -58,7 +59,7 @@ static void RecalculDesEchangesMoyens(Data::Study& study,
 
         for (uint j = 0; j < study.areas.size(); ++j)
         {
-            auto* balance = callback(study.areas.byIndex[j]);
+            auto* balance = callbackBalance(study.areas.byIndex[j]);
```
guilpier-code: callbackBalance is called for every time step and area. Is there a way to call it once for each area ? For example by inverting the 2 loops.

src/solver/variable/economy/links/ntc.h
```diff
@@ -0,0 +1,344 @@
...
+            }
+        }
+    };
+
+}; // class VCard
+
+/*!
+** \brief NTC
+*/
+template<class NextT = Container::EndOfList>
```
guilpier-code: Please add this file to the appropriate cmake file, so that global search over Antares simulator allow searching in the current file.

src/solver/variable/economy/links/ntc.h
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    Antares::Memory::Stored<double>::ConstReturnType retrieveRawHourlyValuesForCurrentYear(
+      uint column,
+      uint numSpace) const
+    {
+        return pValuesForTheCurrentYear[numSpace][column].hour;
+    }
+
+    void localBuildAnnualSurveyReport(SurveyResults& results,
```
guilpier-code: Is the content of this function useful ? To be checked.

guilpier-code (changes requested):

guilpier-code:
I have a study that crashes on this branch : see attached zip
[Uploading 021 Four areas - DC law-v8.2.zip…]()

My bad : it's an well known assertion in **alea_tirage_au_sort_chroniques.cpp** : 183, it will be fixed by a merge with the main trunk branch of NTC evolution

src/solver/variable/economy/links/ntc.h
```diff
@@ -0,0 +1,344 @@
...
+        }
+    };
+
+}; // class VCard
+
+/*!
+** \brief NTC
+*/
+template<class NextT = Container::EndOfList>
+class NTC : public Variable::IVariable<NTC<NextT>, NextT, VCardNTC>
```
guilpier-code: I am not a fan of adding this new output variable in order to make some average computations for the quadratic optimization. For me, it's a kind of over-engineering and, as this variable is not a real output variable (it's not printed as a result), it can be confusing. I understand that this way of doing things is quite easy, and was made on the model of what is currently made for **balance** regarding the quadratic optimization. But still, there should be another way of making things.
By the way, this variable appears in the thematic trimming tab, and we don't want that, I think.

guilpier-code (changes requested):

src/solver/simulation/economy.cpp
flomnes: I agree and renamed this function `ComputeFlowQuad`.

src/solver/simulation/economy.cpp
flomnes: I no longer use callbacks, it was a poor design choice

src/solver/simulation/adequacy.cpp
flomnes: I have removed the callback

src/solver/variable/economy/links/ntc.h
flomnes: I no longer use a new output variable.

src/solver/simulation/common-eco-adq.cpp
flomnes: See #493

guilpier-code (approved):

PR #481, opened 2021-12-09
[UI] Search solver in the build directory first

### LINKED ISSUE #480: Built Antares uses locally installed antares solver
**Description**
Probably not an major issue, as this is mostly about convenience. If I really want to use the built solver, I can uninstall antares.

If you have installed Antares on your machine. And afterwards, you build another version of Antares, launch a GUI session through the _build_name/ui/simulator/antares-8.1-ui-simulator_ file. Then, the solver used on this session won't be the built one, but the installed one.

**How to Reproduce**
- Install antares 8.1
- build the antares on the branch https://github.com/AntaresSimulatorTeam/Antares_Simulator/tree/feature/thermal-ts-gen-switch
- open the built antares `./antares-8.1-ui-simulator`
- start an empty study, add an area, a thermal cluster, and put its _Generate TS_ parameter to _Force Generation_
- Run it
- You will get the following error, which doesn't trigger when actually using the solver of the built antares.

> \`new cluster\`/\`gen-ts\`: The property is unknown and ignored

**Expected behavior**
The built antares to use the built solver.

The logs of the failed simulation, where the solver used was in /usr/bin
[solver-20211208-190845.log](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/7678044/solver-20211208-190845.log)

**OS and version**
 - OS: Ubuntu 20.04.3
 - Antares Simulator version: 8.1 & branch feature/thermal-ts-gen-switch

flomnes:
Do not prioritize the system directory in 2 cases:
- Hand-made build
- Github-generated .tar.gz archive

Instead, prefer the solver from the build/archive.

### DIFF
--- a/src/libs/antares/solver.cpp
+++ b/src/libs/antares/solver.cpp
@@ -74,10 +74,10 @@ bool FindLocation(String& location, Data::Version /*version*/, Solver::Feature f
     }
     else
     {
-        searchpaths.directories.push_back(s.clear() << "/usr/local/bin/");
-        searchpaths.directories.push_back(s.clear() << "/usr/bin/");
         searchpaths.directories.push_back((s = root) << "/../../solver");
         searchpaths.directories.push_back(root); // TGZ package
+        searchpaths.directories.push_back(s.clear() << "/usr/local/bin/");
+        searchpaths.directories.push_back(s.clear() << "/usr/bin/");
     }
 
     bool success = false;


hugo-antoine-rtei (approved):
I tested it both ways and it works (used solver is on line 5):

- [local_build_uses_local_solver.log](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/7715546/local_build_uses_local_solver.log)
- [installed_antares_uses_installed_solver.log](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/7715548/installed_antares_uses_installed_solver.log)

 So it looks all good to me !

hugo-antoine-rtei (approved):
I conducted a small series of tests to make sure it works as intended. I created clusters, on to check the behavior of forced outage and one for planned outage. I ran the following combinations:
- Force generation with Ready-made TS in simulation window 
- Force generation with Stochastic TS in simulation window
- Force no generation with Ready-made TS in simulation window
- Force no generation with Stochastic TS in simulation window
- Use global parameter with Ready-made TS in simulation window
- Use global parameter with Stochastic TS in simulation window

In each of these, the input ts were modified and the output files were created as expected according to the options specified.

I also checked a couple of these using "run the time-series generators only" mode, behavior is also as expected.

When loading a random study (created y another antares version) the default "Generate TS" is set to "Use global parameter" as expected.

To sum up : ALL GOOD, no problems found

PR #483, opened 2021-12-15
Fix when loading links NTC time series from files

guilpier-code:
- bug found when making manual tests : when loading NTC time series from an empty file, a crash occurs because associated matrix is empty after read.
- The condition of an assertion is wrong

### DIFF
--- a/src/libs/antares/study/area/links.cpp
+++ b/src/libs/antares/study/area/links.cpp
@@ -220,15 +220,16 @@ bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& fol
 
     // Read link's parameters times series
     filename.clear() << folder << SEP << with->id << "_parameters.txt";
-    success = parameters.loadFromCSVFile(filename) && success;
+    const uint matrixWidth = 6;
+    success = parameters.loadFromCSVFile(filename, matrixWidth, HOURS_PER_YEAR, Matrix<>::optFixedSize | Matrix<>::optImmediate) && success;
 
     // Read link's direct capacities time series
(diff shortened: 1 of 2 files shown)


flomnes (commented):
Why use `Matrix<>::optImmediate` ? For load, the code looks like this:
```cpp
series.loadFromCSVFile(buffer, 1, HOURS_PER_YEAR, &study.dataBuffer) && ret;
```

I'm afraid we'll load plenty of time-series in the GUI, even when it's not needed. This could lead to longer study loading times.

guilpier-code:
Regarding the previous comment on loading the NTC time series, I passed a study containing **100 load TS per area** with a debugger. After instruction (which loads **Load** TSs) : 
`series.loadFromCSVFile(buffer, 1, HOURS_PER_YEAR, &study.dataBuffer) && ret; `
I was sorry to see that all 100 are loaded.
So if we want to postpone the load of TSs, this is not the way to take.

Furthermore, for **NTC TS**s, instruction : 
`directCapacities.loadFromCSVFile(filename, 1, HOURS_PER_YEAR, Matrix<>::optFixedSize)`
does not work : an error is raised (`Invalid format: Too many entry for the row 0 (offset: 5byte)`).
And instruction : 
`directCapacities.loadFromCSVFile(filename, 1, HOURS_PER_YEAR, Matrix<>::optNone)`
immediately loads the NTC time series. 
Still, not ways to postpone the loading of TSs.

The problem seems to be that `JIT::enabled` is always **false** (see **jit.cpp** : line 33).
If we turn it to **true**, big troubles occur, the application crashes at first "light" loading action.

flomnes (approved):

flomnes:
See ticket #485 when the GUI is ready.

PR #484, opened 2021-12-16
NTC - scenario builder - save scenariobuilder.dat

guilpier-code:


flomnes (commented):
Fix CMake to generate the executable

Add tests for other modes, incl. links

flomnes (commented):
Create a .h+.cpp to create a study in-memory ?

flomnes (approved):

PR #491, opened 2021-12-21
Fix hydro level discontinuities

### LINKED ISSUE #490: Hydro reservoir level goes from 0% to 100% instantly
**Description**
On some studies, hydro reservoirs have become full in one hourly time-step.

**How to Reproduce**
This bug is pretty rare. It has been successfully reproduced in a study named BP50_M2_DE_VE_batt_10_2051, which cannot be included for confidentiality reasons.

**Expected behavior**
The reservoir level should not vary so fast: it isn't in accordance with the physics of the problem. We expect a Lipschitz-continuous function.

image

**OS and version**
 - OS: Windows, GNU/Linux
 - Antares Simulator version: 8.0.x

**Additional context**
None

flomnes:
Level is between 0 and capacity, not 0 and 1.

### DIFF
--- a/src/solver/simulation/sim_structure_probleme_economique.h
+++ b/src/solver/simulation/sim_structure_probleme_economique.h
@@ -369,11 +369,8 @@ public:
 
         level = level + inflows[step] - turb[step] + pumpRatio * pump[step];
 
-        if (Yuni::Math::Zero(level))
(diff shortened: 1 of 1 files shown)


flomnes:
image

guilpier-code (approved):

PR #493, opened 2021-12-23
Remove callback for balance retrieval

flomnes:
Small refactor

### DIFF
--- a/src/solver/simulation/adequacy.cpp
+++ b/src/solver/simulation/adequacy.cpp
@@ -373,7 +373,7 @@ void Adequacy::simulationEnd()
         std::vector<AvgExchangeResults*> balance(nbAreas, nullptr);
         for (uint areaIndex = 0; areaIndex < nbAreas; ++areaIndex)
         {
-            Data::Area* area = study.areas.byIndex[areaIndex];
+            const Data::Area* area = study.areas.byIndex[areaIndex];
             variables.retrieveResultsForArea<Variable::Economy::VCardBalance>(&balance[areaIndex],
                                                                               area);
         }
--- a/src/solver/simulation/adequacy.cpp
+++ b/src/solver/simulation/adequacy.cpp
@@ -364,20 +364,20 @@ void Adequacy::incrementProgression(Progression::Task& progression)
         ++progression;
 }
 
-AvgExchangeResults* Adequacy::callbackRetrieveBalanceData(Data::Area* area)
-{
-    AvgExchangeResults* balance = nullptr;
-    variables.retrieveResultsForArea<Variable::Economy::VCardBalance>(&balance, area);
-    return balance;
-}
 void Adequacy::simulationEnd()
 {
     if (!preproOnly && study.runtime->interconnectionsCount > 0)
     {
-        CallbackBalanceRetrieval callbackBalance;
-        callbackBalance.bind(this, &Adequacy::callbackRetrieveBalanceData);
-
-        ComputeFlowQuad(study, *pProblemesHebdo[0], callbackBalance, pNbWeeks);
+        // Retrieve weighted average balance for each area
+        const uint nbAreas = study.areas.size();
+        std::vector<AvgExchangeResults*> balance(nbAreas, nullptr);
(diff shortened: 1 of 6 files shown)


src/solver/simulation/economy.cpp
```diff
@@ -238,20 +238,20 @@ void Economy::incrementProgression(Progression::Task& progression)
...
-        ComputeFlowQuad(study, *pProblemesHebdo[0], callbackBalance, pNbWeeks);
+        // Retrieve weighted average balance for each area
+        const uint nbAreas = study.areas.size();
+        std::vector<AvgExchangeResults*> balance(nbAreas, nullptr);
+        for (uint areaIndex = 0; areaIndex < nbAreas; ++areaIndex)
+        {
+            const Data::Area* area = study.areas.byIndex[areaIndex];
+            variables.retrieveResultsForArea<Variable::Economy::VCardBalance>(&balance[areaIndex],
+                                                                              area);
+        }
```
guilpier-code: We could put line 246 to 253 into a function, named **Economy::retrieveBalance** and then have : 
```c++
void Economy::simulationEnd()
{
    if (!preproOnly && study.runtime->interconnectionsCount > 0)
    {
        std::vector<AvgExchangeResults*> balance = retrieveBalance(study);
        ComputeFlowQuad(study, *pProblemesHebdo[0], balance, pNbWeeks);
    }
}
```

src/solver/simulation/adequacy.cpp
```diff
@@ -364,20 +364,20 @@ void Adequacy::incrementProgression(Progression::Task& progression)
...
-        ComputeFlowQuad(study, *pProblemesHebdo[0], callbackBalance, pNbWeeks);
+        // Retrieve weighted average balance for each area
+        const uint nbAreas = study.areas.size();
+        std::vector<AvgExchangeResults*> balance(nbAreas, nullptr);
+        for (uint areaIndex = 0; areaIndex < nbAreas; ++areaIndex)
+        {
+            const Data::Area* area = study.areas.byIndex[areaIndex];
+            variables.retrieveResultsForArea<Variable::Economy::VCardBalance>(&balance[areaIndex],
+                                                                              area);
+        }
```
guilpier-code: Save remark as for economy

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -58,12 +58,11 @@ static void RecalculDesEchangesMoyens(Data::Study& study,
...
-            if (balance)
+            assert(balance[j] && "Impossible to find the variable");
+            if (balance[j])
             {
                 problem.SoldeMoyenHoraire[i]->SoldeMoyenDuPays[j]
-                  = balance->avgdata.hourly[decalPasDeTemps];
+                  = balance[j]->avgdata.hourly[decalPasDeTemps];
             }
             else
             {
```
guilpier-code, line 68: About the call to **retrieveAverageNTC** : 
```c++
int ret = retrieveAverageNTC(study, link, avgDirect, avgIndirect);
```
we could 2 calls to a single function : 
```c++
int ret = retrieveAverageNTC(study, link->directCapacities, avgDirect);
ret = retrieveAverageNTC(study, link->indirectCapacities, avgIndirect) && ret;
```
as the treatment is very similar for both direct and indirect capacities.
Or we can consider it is going too far in code factorization.

guilpier-code (changes requested):

src/solver/simulation/common-eco-adq.cpp
flomnes: That looks good, I will do that

src/solver/simulation/adequacy.cpp
flomnes: done

src/solver/simulation/economy.cpp
flomnes: done

flomnes (commented):
Accepted all remarks and applied the corresponding changes.

guilpier-code (approved):

PR #496, opened 2021-12-28
GUI - links and NTC time series

guilpier-code:
Inside the notebook tab **links** : 
- Summary tab remains untouched
- **Details** tabs is replaced with **Parameters** tab. This tab does not contain anymore time series for direct/indirect capacities (NTC)
- A new tab **Transmission capacities** was created : it contains 2 grids of several time series for direct and indirect NTC


src/ui/simulator/windows/hydro/levelsandvalues.cpp
```diff
@@ -71,14 +71,6 @@ void LevelsAndValues::createComponents()
     wxBoxSizer* ssGrids = new wxBoxSizer(wxHORIZONTAL);
     pSupport->SetSizer(ssGrids);
 
-    const wxSize ourDefaultSize(55, wxDefaultSize.GetHeight());
```
guilpier-code, line 74: This code is useless. I removed it.

src/ui/simulator/windows/constraints-builder/constraintsbuilder.cpp
```diff
@@ -476,16 +476,7 @@ void ConstraintsBuilderWizard::onIncludeLoopFlow(wxCommandEvent& evt)
     pCBuilder->setLoopFlowInclusion(value);
     pCBuilder->setUpToDate(false);
     pCBuilder->saveCBuilderToFile();
-
-    // update all lines in the study
```
guilpier-code, line 480: I only removed commented code in this file.

src/ui/simulator/windows/hydro/levelsandvalues.cpp
flomnes: ok

flomnes (commented):
It is not possible to change the matrix size for NTCs. Consider overloading method `maxWidthResize` to return a large value, e.g 1000.

Another problem is that the user must resize both matrices. Resizing should be done on both matrices at the same time.

src/ui/simulator/toolbox/components/datagrid/dbgrid.cpp
```diff
@@ -669,9 +669,21 @@ void DBGrid::onScroll(wxScrollWinEvent& evt)
     Renderer::IRenderer* r = ((Component*)GetParent())->renderer();
     if (r)
         r->onScroll();
+
+    // Synchronize scrolling with another grid
+    if (otherGrid_ != nullptr)
+    {
+        otherGrid_->Scroll(GetScrollPos(wxHORIZONTAL), GetScrollPos(wxVERTICAL));
+    }
```
flomnes, line 680: I suggest making the horizontal scrolling optional. Something like

```cpp
int horizontalPos;
if (horizontalScroll)
{
    horizontalPos = GetScrollPos(wxHORIZONTAL);
}
else 
{
    horizontalPos = otherGrid_->GetScrollPos(wxHORIZONTAL);
}
```
Add `horizontalScroll` as an argument of `setOtherGrid` and save it.

guilpier-code: So you recommend deactivating horizontal simultaneous scrolling, right ?
A user could have a need to compare the same TS for direct and indirect NTC.
So it seems to me horizontal scrolling should be activated, especially when time series are numerous.

flomnes: I recomment making horizontal scrolling optional if we decide to use this sync in other components that may not have the same width (hydro for example).

guilpier-code: ok, for information, you and I decided to treat this point later : there is an existing issue on this #507

flomnes (approved):

PR #497, opened 2021-12-31
GUI : scenario builder for links NTC

guilpier-code:


src/ui/simulator/toolbox/components/datagrid/renderer/scenario-builder-renderer-base.cpp
```diff
@@ -144,11 +113,61 @@ IRenderer::CellStyle ScBuilderRendererBase::cellStyle(int x, int y) const
 
 void ScBuilderRendererBase::onStudyClosed()
 {
-    pArea = nullptr;
+    // pArea = nullptr;
```
flomnes: Why comment this line ?

flomnes (changes requested):

flomnes (changes requested):
To be merged after #499

flomnes:
Fix 3 bugs and possibly a few issues

src/ui/simulator/toolbox/components/datagrid/renderer/scenario-builder-renderer-base.cpp
guilpier-code: Because dealing with areas is now done by class **ScBuilderRendererAreasAsRows**, not by **ScBuilderRendererBase** (parent of the latter).
**pArea** is not part of **ScBuilderRendererBase** anymore.

flomnes (approved):

PR #500, opened 2022-01-05
Remove clusterIndexMap and linksIndexMap, make Matrix<>'s move operations noexcept

flomnes:
also format code.

### DIFF
--- a/src/libs/antares/array/matrix.hxx
+++ b/src/libs/antares/array/matrix.hxx
@@ -1594,6 +1594,8 @@ inline Matrix<T, ReadWriteT>& Matrix<T, ReadWriteT>::operator=(Matrix<T, ReadWri
     {
         entry = rhs.entry;
     }
+    // Prevent spurious de-allocation from rhs's destructor
+    rhs.entry = nullptr;
     return *this;
 }
 
--- a/src/libs/antares/array/matrix.hxx
+++ b/src/libs/antares/array/matrix.hxx
@@ -1584,7 +1584,7 @@ inline Matrix<T, ReadWriteT>& Matrix<T, ReadWriteT>::operator=(Matrix<T, ReadWri
     width = rhs.width;
     height = rhs.height;
     jit = rhs.jit;
-    if (0 == width or 0 == height)
+    if (0 == width || 0 == height)
     {
         entry = nullptr;
         width = 0;
--- a/src/libs/antares/array/matrix.hxx
+++ b/src/libs/antares/array/matrix.hxx
@@ -237,6 +237,13 @@ Matrix<T, ReadWriteT>::Matrix(const Matrix<T, ReadWriteT>& rhs) :
 }
 
 template<class T, class ReadWriteT>
+Matrix<T, ReadWriteT>::Matrix(Matrix<T, ReadWriteT>&& rhs) noexcept
+{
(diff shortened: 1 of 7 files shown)


src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -195,12 +193,13 @@ bool loadTSNumberData::apply(Study& study)
         const MatrixType::ColumnType& col = pTSNumberRules[areaIndex];
 
         logprefix.clear() << "Load: Area '" << area.name << "': ";
-        status_to_return = status_to_return && ApplyToMatrix(errors, logprefix, *area.load.series, col, tsGenCountLoad);
+        status_to_return
```
guilpier-code: Personally, I prefer when an instruction holds in one line, given that this line is not too long, which wasn't the case I think. The automatic formatter (slang) is a bit too restrictive on this I think.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -603,18 +602,19 @@ bool ntcTSNumberData::apply(Study& study)
     // Alias to the current area
     assert(pArea != nullptr);
     assert(pArea->index < study.areas.size());
-    Area& area = *(study.areas.byIndex[pArea->index]);
+    const Area& area = *(study.areas.byIndex[pArea->index]);
 
     const uint ntcGeneratedTScount = get_tsGenCount(study);
 
-    for (auto i = pArea->links.begin(); i != pArea->links.end(); ++i)
+    for (auto i : pArea->links)
```
guilpier-code: Why not have `for (const auto& i : pArea->links)`, like your correction above (line 572) ?

src/libs/antares/array/matrix.hxx
```diff
@@ -1571,6 +1578,25 @@ inline Matrix<T, ReadWriteT>& Matrix<T, ReadWriteT>::operator=(const Matrix<T, R
...
+    jit = rhs.jit;
+    if (0 == width || 0 == height)
+    {
+        entry = nullptr;
+        width = 0;
+        height = 0;
+    }
+    else
+    {
+        entry = rhs.entry;
```
guilpier-code, line 1595: Question : I don't master move semantics, but since **rhs** is "moved", shouldn't we  leave **rhs.entry** to **nullptr** at the end of the copy operator ?

flomnes: ok, I also added a comment to explain why we set `rhs.entry = nullptr` after the move.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
flomnes: I'll try

src/libs/antares/study/scenario-builder/TSnumberData.cpp
flomnes: `status_to_return` -> `ret`

src/libs/antares/study/scenario-builder/TSnumberData.cpp
flomnes: done

guilpier-code (approved):

PR #502, opened 2022-01-06
Do not try to de-allocate global arrays if they were not allocated

flomnes:
In some cases, when loading the study fails, `SIM_AllocationTableaux` is not called, but `SIM_DesallocationTableaux` is called. Trying to free address `0x0` triggers a segfault.

The goal of this PR is to make sure this doesn't happen.

### DIFF
--- a/src/solver/main.cpp
+++ b/src/solver/main.cpp
@@ -214,8 +214,6 @@ bool SolverApplication::prepare(int argc, char* argv[])
             // Alias de la zone courant
             auto& area = *(pStudy->areas.byIndex[i]);
 
-            auto NombreDePaliersThermiques = area.thermal.list.size();
-
             for (uint l = 0; l != area.thermal.clusterCount(); ++l) //
             {
                 auto& cluster = *(area.thermal.clusters[l]);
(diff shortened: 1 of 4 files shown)


src/solver/main.cpp
```diff
@@ -588,8 +583,16 @@ int main(int argc, char** argv)
...
-    if (application->prepare(argc, argv))
-        ret = application->execute();
+    try
+    {
+        application->prepare(argc, argv);
+    }
+    catch (const std::runtime_error& e)
+    {
+        logs.error() << e.what();
+        AntaresSolverEmergencyShutdown();
```
guilpier-code, line 593: **logs.error()** does not stop Antares' run, and it's ok since it does not do something it's not meant for.
So you're right to call **AntaresSolverEmergencyShutdown()** here

src/solver/main.cpp
```diff
@@ -214,8 +214,6 @@ bool SolverApplication::prepare(int argc, char* argv[])
             // Alias de la zone courant
             auto& area = *(pStudy->areas.byIndex[i]);
 
-            auto NombreDePaliersThermiques = area.thermal.list.size();
```
guilpier-code, line 217: **NombreDePaliersThermiques** totally useless indeed

guilpier-code (approved):

PR #505, opened 2022-01-10
Remove calls to exit()

flomnes:


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -561,8 +561,7 @@ void OPT_EcrireJeuDeDonneesLineaireAuFormatMPS(void* Prob, uint numSpace, char T
                     "des sens reconnus\n",
                     __FUNCTION__,
                     Sens[Cnt]);
-            AntaresSolverEmergencyShutdown();
-            exit(0);
+            AntaresSolverEmergencyShutdown(2);
```
guilpier-code, line 564: Is it expected to "emergency" exit with 2 as though the previous exit code was 0 ?

flomnes: Anything related to the optimization problem should return 2 by convention.

PR #508, opened 2022-01-13
Add option --exclude to study cleaner

flomnes:


src/tools/cleaner/main.cpp
```diff
@@ -118,6 +121,11 @@ int main(int argc, char* argv[])
 
         options.addFlag(optMrProper, ' ', "mrproper", "Suppress the outputs and logs files");
 
+        options.add(optExclude,
```
guilpier-code: Maybe we should add a short example of use, like : 
--exclude folder-path:folder-path-2/some-file.txt:folder-path-3/some-other-file.ini

guilpier-code (approved):

PR #511, opened 2022-01-13
Add "view results" button and function

### LINKED ISSUE #333: Add a "view output" button when a simulation ends
In this window, allow the user to view the output associated to a simulation.

image

flomnes:
Add a "view results" button in the "Simulation" modal. Clicking this button, the user opens the output for that run.

This should save the user a few clicks.

image

### DIFF
--- a/src/ui/simulator/windows/message.h
+++ b/src/ui/simulator/windows/message.h
@@ -92,6 +92,8 @@ public:
         btnUpgrade,
         //! Standard button: continue
         btnContinue,
+        //! Standard button: view simulation results
+        btnViewResults,
         //! Standard button: quit
         btnQuit
     };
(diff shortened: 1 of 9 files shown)


src/ui/simulator/windows/message.h
```diff
@@ -92,6 +92,8 @@ class Message final : public wxDialog
         btnUpgrade,
         //! Standard button: continue
         btnContinue,
+        //! Standard button: view simulation results
+        btnViewResults,
```
guilpier-code, line 96: This change is ok for me.
It's just that **pReturnStatus** should not be defined as a **uint**, but more as a item of an **enum class**, otherwise one can't swiftly know it's related to an (old) **enum**

guilpier-code (approved):

src/ui/simulator/windows/message.h
flomnes: Done. I cleaned up the code and added a template to remove an ugly double cast (`void*` -> `size_t` -> `uint`).

PR #521, opened 2022-01-14
Fix segfault when study folder does not exist / refactor error handling in solver's main()

flomnes:
When the study folder does not exist, no `study` is created. Trying to de-reference it results in a segfault.

Add safeguard in `SolverApplication::execute()`


src/libs/antares/exception/LoadingError.hpp
```diff
@@ -0,0 +1,148 @@
...
+class StudyFolderDoesNotExist : public std::runtime_error
+{
+public:
+    explicit StudyFolderDoesNotExist(const Yuni::String& folder);
+};
+
+class ReadingStudy : public std::runtime_error
+{
+public:
+    explicit ReadingStudy();
```
guilpier-code: This constructor has no parameter. Why the **explicit** keyword then ?
The question stands for any other constructor that has no parameter.

src/solver/main.cpp
```diff
@@ -104,8 +104,9 @@ bool SolverApplication::prepare(int argc, char* argv[])
     options.usedByTheSolver = true;
 
     // Parse arguments and store usefull values
-    if (not GrabOptionsFromCommandLine(argc, argv, pSettings, options))
-        return false;
+    int ret = GrabOptionsFromCommandLine(argc, argv, pSettings, options);
```
guilpier-code: variable **ret** not very useful here, as it is not used elsewhere in this function.
Maybe the old code is more concise and clear :
```c++
if (not GrabOptionsFromCommandLine(argc, argv, pSettings, options))
        return false;
```

src/solver/main.cpp
```diff
@@ -200,8 +194,8 @@ bool SolverApplication::prepare(int argc, char* argv[])
                   << it->second
                   << ". Conflict between Min Stable Power, Pnom, spinning and capacity modulation.";
             }
-
-            return false;
+            // TODO : create & throw exception
```
guilpier-code: I see, what's preventing you from throwing an exception here is that there a loop. It should be a problem though, and it would allow this function to return **void**

src/solver/main.h
```diff
@@ -56,7 +56,7 @@ class SolverApplication
     **
     ** \return False if the operation failed.
     */
-    bool prepare(int argc, char* argv[]);
+    int prepare(int argc, char* argv[]);
```
guilpier-code: **prepare(...)** could return **void** eventually

src/solver/misc/options.cpp
```diff
@@ -219,11 +220,12 @@ bool GrabOptionsFromCommandLine(int argc,
...
     {
 #ifdef GIT_SHA1_SHORT_STRING
-        std::cout << ANTARES_VERSION_STR <<  " (revision " << GIT_SHA1_SHORT_STRING << ")" << std::endl;
+        std::cout << ANTARES_VERSION_STR << " (revision " << GIT_SHA1_SHORT_STRING << ")"
+                  << std::endl;
 #else
         std::cout << ANTARES_VERSION_STR << std::endl;
 #endif
-        return false;
+        return 1;
```
guilpier-code: We should make this function return void, as **return** is only called twice : here and at the end of the current function. Exceptions within it would be caught 2 levels above.

src/solver/misc/options.h
```diff
@@ -65,9 +65,9 @@ class Settings
 /*!
 ** \brief Interprete options given in command line
 */
-bool GrabOptionsFromCommandLine(int argc,
-                                char* argv[],
-                                Settings& settings,
-                                Antares::Data::StudyLoadOptions& options);
+int GrabOptionsFromCommandLine(int argc,
```
guilpier-code: Should return **void**

src/solver/main.cpp
guilpier-code: Moreover, **GrabOptionsFromCommandLine** could return **void**, in which case an **if** statement is not necessary

guilpier-code (changes requested):

src/solver/application.cpp
```diff
@@ -0,0 +1,466 @@
...
+    logs.debug() << "tsGenThermal = " << tsGenThermal;
+
+    if (tsGenThermal)
+    {
+        std::map<int, YString> areaClusterNames;
+        if (!(pStudy->areasThermalClustersMinStablePowerValidity(areaClusterNames)))
+        {
+            Error::ThermalClusterHelper helper;
+            for (const auto& it : areaClusterNames)
+                helper.append(it.second);
```
flomnes: Remove append, helper. Provide list of clusters to `InvalidParametersForThermalClusters`

src/libs/antares/exception/LoadingError.hpp
flomnes: I placed it here by habit, but it's not really useful in this case.

src/libs/antares/exception/LoadingError.cpp
```diff
@@ -0,0 +1,125 @@
...
+              + " : simplex daily optimization and use heuristic target == no are not compatible")
+{
+}
+
+WritingProgressFile::WritingProgressFile(const Yuni::String& file) :
+ LoadingError(std::string("I/O error: impossible to write ") + file.c_str())
+{
+}
+
+static std::string buildInvalidParametersMessage(const std::map<int, Yuni::String>& clusterNames)
```
guilpier-code: Why not defining a method **makeMessage()** private to **InvalidParametersForThermalClusters** class instead of an external function ?

src/solver/main/adequacy-draft.cpp
```diff
@@ -25,17 +25,24 @@
...
+{
+namespace Solver
+{
+void Application::runSimulationInAdequacyDraftMode()
 {
     // Type of the simulation
     typedef Solver::Simulation::ISimulation<Solver::Simulation::AdequacyDraft> SimulationType;
-
-    runSimulation<SimulationType>();
+    SimulationType simulation(*pStudy, pSettings);
```
guilpier-code, line 41: There is code duplication here : we have same lines of code for draft, adequacy and economy

src/solver/misc/options.cpp
```diff
@@ -45,151 +45,134 @@
...
+    {
+        availableSolverListStr += *it + ";";
+    }
+    // Remove last semicolumn
+    if (!availableSolverListStr.empty())
+        availableSolverListStr.pop_back();
+    return availableSolverListStr;
+}
+
+void GrabOptionsFromCommandLine(int argc,
```
guilpier-code: As I can understand, this function makes 2 things : 
- It defines the admissible solver command line options
- It reads the command line options the user actually chose at execution (that's the reason why we have **argc** and **argv** as an argument of this function)

We could rename this function (ex : **defineAdmissibleOptions**), remove user options parsing from it (2 last lines) and it could return the parser (note **argc** & **argv** are no more needed as arguments ). 
After the call of this function, we parse user arguments as it is done now in this function.

guilpier-code (changes requested):

src/solver/main/adequacy-draft.cpp
flomnes: Right, except that `SimulationType` is not the same. I removed the function template.

src/libs/antares/exception/LoadingError.cpp
flomnes: Done

src/solver/misc/options.cpp
flomnes: This idea is great, I tried to implement it. Many thanks

src/solver/main/adequacy-draft.cpp
flomnes: Function templates require an **.hxx** file, adding some complexity and compilation time. I'd rather not use them if possible.

src/solver/main.cpp
flomnes: done

src/solver/main.cpp
flomnes: done

src/solver/main.h
flomnes: done

src/solver/misc/options.cpp
flomnes: done

src/solver/misc/options.h
flomnes: done

src/libs/antares/exception/LoadingError.hpp
flomnes: done

src/solver/main/adequacy-draft.cpp
flomnes: Interface + factory function

guilpier-code (approved):

PR #525, opened 2022-01-18
Tests : activating end-to-end tests on Windows OS

guilpier-code:


src/tests/examples/antares_test_utils.py
```diff
@@ -90,46 +58,118 @@ def enable_study_output(study_path, enable):
...
-
+def skip_folder(folder):
+    if basename(folder) == 'grid':
+        return True
+    return False
+
+def skip_file(file):
+    if file in ['id-daily.txt', 'id-hourly.txt']:
+        return True
+    return False
```
flomnes: `return file in ['id-daily.txt', 'id-hourly.txt']`

guilpier-code: As the function is named **skip_file**, it may return **true** or **false**. returning anything else could be confusing for the caller

flomnes:
```python
>>> 1 in [1,2,3]
True
>>> 4 in [1,2,3]
False
```

src/tests/examples/antares_test_utils.py
```diff
@@ -90,46 +58,132 @@ def enable_study_output(study_path, enable):
...
+
+
+def get_tolerances():
+    if sys.platform == "linux":
+        return Linux_tolerances()
+    elif sys.platform == "win32":
+        return Win_tolerances()
+    else:
+        print("Unknown OS")
+        raise AssertionError()
```
flomnes: raise AssertionError("Unknown OS")

src/tests/examples/output_compare_test.py
```diff
@@ -7,139 +7,158 @@
 ## TESTS ##
 @pytest.mark.short
 def test_001_one_node_passive(use_ortools, ortools_solver, solver_path):
+    print()
+    print("Running the tests")
```
flomnes: Remove this print

src/tests/examples/output_compare_test.py
```diff
@@ -513,95 +597,107 @@ def test_071_hydro_reservoir_model_enhanced_04(use_ortools, ortools_solver, solv
...
 
 @pytest.mark.short
 def test_playlist_23(use_ortools, ortools_solver, solver_path):
     study_path = ALL_STUDIES_PATH / "short-tests" / "playlist-23"
     enable_study_output(study_path, True)
     run_study(solver_path, study_path, use_ortools, ortools_solver)
     enable_study_output(study_path, False)
-    check_output_values(study_path)
+    tolerances = get_tolerances()
+    check_output_values(study_path, tolerances)
```
flomnes: `check_output_values(study_path, get_tolerances())`

src/tests/examples/antares_test_utils.py
guilpier-code: Sorry. Done.

src/tests/examples/output_compare_test.py
guilpier-code: Done

src/tests/examples/output_compare_test.py
guilpier-code: ok, done

src/tests/examples/antares_test_utils.py
guilpier-code: Indeed, that would be better. But option **--tb=no** when calling **output_compare_test.py** prevents any trace back report for assertions to be printed.
We could use  **--tb=line** instead (one line of trace back per failure).
But in that case we have a report like this (2 tests are failing) : 
```
================================== FAILURES ===================================
D:\Users\pierregui\Desktop\ant-open-rel\src\tests\examples\antares_test_utils.py:178: AssertionError: Comparison failed
D:\Users\pierregui\Desktop\ant-open-rel\src\tests\examples\antares_test_utils.py:178: AssertionError: Comparison failed
=========================== short test summary info ===========================
FAILED output_compare_test.py::test_065_pumped_storage_plant_explicit_model_01
FAILED output_compare_test.py::test_067_pumped_storage_plant_explicit_model_03
================= 2 failed, 1 passed, 82 deselected in 34.11s =================
```
We have one trace back per failure, but failed assertions are not associated to a test. So we could insert the test name in the assertion message.

flomnes (approved):

PR #532, opened 2022-01-26
Adq Patch Inputs step 1

hugo-antoine-rtei:
A first pull request for creating the inputs:
- When creating a new area, the adequacy_patch.ini file will be created containing False (PR#1)
- Add the true/false option in the Optimization Parameters (to use or not the adq patch) (PR#2)

### DIFF
--- a/src/libs/antares/study/area/list.cpp
+++ b/src/libs/antares/study/area/list.cpp
@@ -348,7 +348,7 @@ bool saveAreaAdequacyPatchIniFile(const Area& area, const Clob& buffer)
         value = 2;
         break;
     default:
-        value = 0;
+        value = 1; //default adqmPhysicalAreaOutsideAdqPatch
         break;
     }
     section->add("adequacy-patch-mode", value);
@@ -1098,7 +1098,7 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
                         area.adequacyPatchMode = Data::adqmPhysicalAreaInsideAdqPatch;
                         break;
                     default:
-                        area.adequacyPatchMode = Data::adqmVirtualArea;
+                        area.adequacyPatchMode = Data::adqmPhysicalAreaOutsideAdqPatch;
                         break;
                     }
                     continue;
--- a/src/libs/antares/study/area/list.cpp
+++ b/src/libs/antares/study/area/list.cpp
@@ -203,6 +203,11 @@ static bool AreaListSaveToFolderSingleArea(const Area& area, Clob& buffer, const
                    << "optimization.ini";
     ret = saveAreaOptimisationIniFile(area, buffer) and ret;
 
+    //Adequacy ini
+    buffer.clear() << folder << SEP << "input" << SEP << "areas" << SEP << area.id << SEP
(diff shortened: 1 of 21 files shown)


src/libs/antares/study/area/list.cpp
```diff
@@ -323,6 +328,15 @@ bool saveAreaOptimisationIniFile(const Area& area, const Clob& buffer)
     return ini.save(buffer);
 }
 
+bool saveAreaAdequacyPatchIniFile(const Area& area, const Clob& buffer)
+{
+    IniFile ini;
+    IniFile::Section* section = ini.addSection("adequacy-patch");
+    bool bUseAdequacyPatch = false; //adq: this should be later replaced with area.bUseAdequacyPatch
+    section->add("use-adequacy-patch", static_cast<bool>(bUseAdequacyPatch));
```
flomnes: No need to cast, `bUseAdequacyPatch` is a bool.

src/libs/antares/study/cleaner/cleaner-v20.cpp
```diff
@@ -195,6 +195,8 @@ static void PreflightVersion20_area(PathList& e, PathList& p, const Area* area,
     e.add(buffer);
     buffer.clear() << "input/areas/" << id << "/optimization.ini";
     e.add(buffer);
+    buffer.clear() << "input/areas/" << id << "/adequacy_patch.ini";
```
flomnes, line 198: Well done ! This tool us easily forgotten. Adding this here should prevent the study cleaner from removing these newly introduced files.

src/libs/antares/study/parameters.cpp
```diff
@@ -548,6 +549,8 @@ static bool SGDIntLoadFamily_Optimization(Parameters& d,
         return value.to<bool>(d.include.reserve.primary);
     if (key == "include-exportmps")
         return value.to<bool>(d.include.exportMPS);
+    if (key == "include-adequacypatch")
```
flomnes, line 553: Check indentation, use the `clang-format` tool

flomnes:
I'm not sure we need both a global and a local (one for each area) flag. Maybe local flags are enough, the global flag can be deduced from the local flags. The adq patch should be enabled for the simulation if it is set to `true` for any area.

flomnes (changes requested):

src/libs/antares/study/parameters.cpp
hugo-antoine-rtei: Should be correct now

src/libs/antares/study/area/list.cpp
hugo-antoine-rtei: corrected

hugo-antoine-rtei:
I check with Jean-Marc for your comment:
> I'm not sure we need both a global and a local (one for each area) flag. Maybe local flags are enough, the global flag can be deduced from the local flags. The adq patch should be enabled for the simulation if it is set to `true` for any area.
He finds it better to have a global parameter, so as not to have to modify each area when he wants to run the adq patch.

PR #533, opened 2022-01-27
Do not set up Python, use system Python3

flomnes:


flomnes:
We still get an error from the fact that multiple versions of Python are included in Github's [virtual-environments](https://github.com/actions/virtual-environments). CMake and scripts don't mention a specific version of Python, so we have no guarantee that the same version is called.

#### Python
| Version | Architecture |
| ------- | ------------ |
| 2.7.18  | x64, x86     |
| 3.6.8   | x64, x86     |
| 3.7.9 (Default) | x64, x86     |
| 3.8.10  | x64, x86     |
| 3.9.9   | x64, x86     |
| 3.10.1  | x64, x86     |

[source](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#python)

`pytest` and `numpy` are installed in `c:\hostedtoolcache\windows\python\3.7.9\x64\lib\site-packages` by the Github Actions script.

CMake on the other hand finds  `c:/hostedtoolcache/windows/Python/3.10.1/x64/python3.exe`, for which neither `pytest` nor `numpy` are installed.

We need a way to make sure that the same installation of Python is used.

src/tests/CMakeLists.txt
```diff
@@ -154,7 +154,7 @@ if(Python3_Interpreter_FOUND)
         )
         set_property(TEST non-regression PROPERTY LABELS non-regression sirius)
   else()
```
guilpier-code, line 156: Instead of having an **if (numpy and pytest)** statement embracing a big block, we should have : 
```cmake
if (PYTHON_MODULE_pytest_FOUND AND PYTHON_MODULE_numpy_FOUND)
    message(FATAL_ERROR "Module pytest or numpy not installed : ..." )
```
Same remark for the : 
```cmake
if(Python3_Interpreter_FOUND)
```
as long as : 
```cmake
find_package(Python3 COMPONENTS Interpreter)
```
does not already do the job, of course : maybe **cmake** stops with an error if the **find_package** does not find anything.

.github/workflows/windows-vcpkg.yml
```diff
@@ -27,11 +27,6 @@ jobs:
       
     steps:
     - uses: actions/checkout@v2    
-
```
guilpier-code, line 30: Question : on **ubuntu**, we keep asking for python 3.8. Does this mean that no default python is installed ?

guilpier-code (changes requested):

.github/workflows/windows-vcpkg.yml
flomnes: Python 3.8.10 is installed on ubuntu-latest (20.04), see [here](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#language-and-runtime).

PR #534, opened 2022-01-27
Improve UI for new thermal parameter "tsGenBehavior"

### LINKED ISSUE #526: In the "thermal" tab, show that generation/no-generation is forced

flomnes:
### Add a **G**, **NG** or both when some thermal clusters force TS generation, force absence of TS generation or the two.
image

### Add a "Generate" icon when the TS gen behavior for a cluster is set to "force TS gen"
image

### Add a "No Generate" icon when TS gen behavior for a cluster is set to "force no TS gen"
image

### Remove "Th" and "Re" icons


guilpier-code:
Depending on the global thermal generation status (see TS management for thermal),  which can be **we generate TS** or **we don't generate TS**, we could narrow the available local behavior of each thermal cluster : 
- if global status is **generation**, the user choice is either **Use global parameter** or **force no generation**
-  if global status is **no generation**, the user choice is either **Use global parameter** or **force generation**

This evolution could be costly. to be discussed.

flomnes:
> Depending on the global thermal generation status (see TS management for thermal), which can be **we generate TS** or **we don't generate TS**, we could narrow the available local behavior of each thermal cluster :
> 
>     * if global status is **generation**, the user choice is either **Use global parameter** or **force no generation**
> 
>     * if global status is **no generation**, the user choice is either **Use global parameter** or **force generation**
> 
> 
> This evolution could be costly. to be discussed.

What should happen if the user changes the global parameter ? We'd have to change local parameters, which is a dangerous side effect IMO.

guilpier-code:
Maybe I missed a point but I don't see what **[G|NG]** is for. I see that information as a deviation from the global parameter, for some clusters.
If the global parameter is turned to **Generation**, then the clusters in **no generation** make the **[NG]** appear, ok. 
But the clusters turned to **force generation** should not add a  **G** to the latter insertion.

guilpier-code:
Changing a cluster generation status does not make the **[G]**, **[NG]** or **[G|NG]** insertion change immediately, user has to click on an other area to refresh it. For me, it can be misleading.
Besides, I'm not sure that putting this insertion next to the study name is a good idea : it is too specific to thermal clusters. I would put that insertion more next to the Thermal tab, no ?

guilpier-code:
> What should happen if the user changes the global parameter ? We'd have to change local parameters, which is a dangerous 
> side effect IMO.
We would have to change the choices that the user have to changes the status of a thermal cluster regarding the generation.

guilpier-code:
Last thing (sorry, I should have ordered my remarks/question in one comment) : I don't understand why we have 3 choices of generation policy for a cluster. 2 would be enough : either we generate, or we don't. By default, we follow the global generation status.

flomnes:
> Depending on the global thermal generation status (see TS management for thermal), which can be **we generate TS** or **we don't generate TS**, we could narrow the available local behavior of each thermal cluster :
> 
>     * if global status is **generation**, the user choice is either **Use global parameter** or **force no generation**
> 
>     * if global status is **no generation**, the user choice is either **Use global parameter** or **force generation**
> 
> 
> This evolution could be costly. to be discussed.

As discussed, we use 3 values to have a consistent default value (use global parameter). As you mention it, we could use only 2, but then we would need to change local parameters when changing the global parameter, which I find misleading for the user.

flomnes:
> Changing a cluster generation status does not make the **[G]**, **[NG]** or **[G|NG]** insertion change immediately, user has to click on an other area to refresh it. For me, it can be misleading. Besides, I'm not sure that putting this insertion next to the study name is a good idea : it is too specific to thermal clusters. I would put that insertion more next to the Thermal tab, no ?

I agree. This change was applied.

guilpier-code:
In GUI, when the user changes the generation status of a thermal cluster, the icon **G** or **NG** immediately shows up on the cluster line, and that's good. 
But it does not show up on the area line. Clicking on another area does not make that icon appear either. The only way I found to make it appear (still on the area line) is to save the study.

flomnes:
> In GUI, when the user changes the generation status of a thermal cluster, the icon **G** or **NG** immediately shows up on the cluster line, and that's good. But it does not show up on the area line. Clicking on another area does not make that icon appear either. The only way I found to make it appear (still on the area line) is to save the study.

Yes we need to refresh the panel when the user changes some cluster data. This is a pre-existing bug.

src/ui/simulator/toolbox/components/htmllistbox/item/cluster-item.cpp
```diff
@@ -60,24 +60,30 @@ void ClusterItem::preloadImages()
     {
         pIconFileEnabled = getIconFilePath("images/16x16/light_green.png");
         pIconFileDisabled = getIconFilePath("images/16x16/light_orange.png");
-        pClusterIconFilePath = getClusterIconFilePath();
     }
 }
 
+void ClusterItem::addAdditionalIcons(wxString& out) const
```
guilpier-code, line 66: An empty method is a code smell. **addAdditionalIcons(...)** should be a private function of class **ThermalClusterItem**.
To do this, we could : 
- split **ClusterItem::HtmlContent(...)** into **ClusterItem**'s protected functions with appropriate names,
- make **ClusterItem::HtmlContent(...)** abstract
- define a **HtmlContent** method for each **ClusterItem**'s child (thermal or renewable), and call the **ClusterItem**'s protected functions depending on the child. **ThermalClusterItem** could call its own **addAdditionalIcons**.

src/ui/simulator/toolbox/components/htmllistbox/item/thermal-cluster-item.cpp
```diff
@@ -38,36 +38,54 @@ namespace HTMLListbox
...
-    return getIconFilePath("images/16x16/thermal.png");
+    pIconFileRefresh = getIconFilePath("images/16x16/G.png");
+    pIconFileNoRefresh = getIconFilePath("images/16x16/NG.png");
 }
 
 wxString ThermalClusterItem::htmlContentTail()
 {
     wxString tail_to_return;
-    tail_to_return << Math::Round(pThermalClusterMarketBidCost, 3)
+    tail_to_return << Math::Round(pThermalCluster->marketBidCost, 3)
```
guilpier-code, line 68: I totally agree with that change : market bid cost is not a **ThermalClusterItem**'s data member anymore.

src/ui/simulator/toolbox/components/htmllistbox/item/thermal-cluster-item.cpp
```diff
@@ -38,36 +38,54 @@ namespace HTMLListbox
...
                    << wxT(" <font color=\"#DD3311\">\u20AC/MWh</font></font></td>")
                    << wxT("<td width=5></td>");
     return tail_to_return;
 }
 
+void ThermalClusterItem::addAdditionalIcons(wxString& out) const
```
guilpier-code, line 74: See remark in **cluster-item.cpp**

src/ui/simulator/windows/inspector/property.update.cpp
```diff
@@ -924,28 +924,29 @@ bool InspectorGrid::onPropertyChanging_ThermalCluster(wxPGProperty*,
     if (name == "cluster.gen-ts")
     {
         long index = value.GetLong();
-
-        Data::LocalTSGenerationBehavior behavior
-          = Data::LocalTSGenerationBehavior::useGlobalParameter;
+        using Behavior = Data::LocalTSGenerationBehavior;
+        auto behavior = Behavior::useGlobalParameter;
```
guilpier-code, line 928: Maybe **behavior** should be renamed to be more expressive, even if it's a local variable in a local context

guilpier-code (changes requested):

src/ui/simulator/toolbox/components/htmllistbox/item/cluster-item.cpp
flomnes: It's a code smell if you leave the body of the function empty, which isn't the case here.

I agree that we could split functions, but we would lose a lot of  expressiveness. How is the child supposed to implement `HtmlContent_part1`, `HtmlContent_part2` and so on ? From children classes, these functions' meaning could be hard to understand.

We don't want to stick too much to the implementation of the base class here.

src/ui/simulator/windows/inspector/property.update.cpp
flomnes: It' clear from lines 925 and 928 that this a TS generation behavior.

guilpier-code (approved):

PR #541, opened 2022-02-04
Remove duplicate bool StudyLoadOptions::useOrtools

flomnes:
* Having 2 attributes results in OR-Tools not being used even if the flag `--use-ortools` is present.
* `ortoolsUsed` is read in file **parameters.cpp**.
* This is a regression introduced by #521


flomnes:
`ortoolsUsed` is read in file **parameters.cpp**.

guilpier-code (approved):

PR #550, opened 2022-02-16
[UI] Fix segfault on area copy-paste

### LINKED ISSUE #549: Antares V8.2-rc1 - Copy-pasting of an area causes a crash
**Description**
On the enclosed study, when trying to duplicate the area "Area" via copy-paste, Antares crashes with no error message (the window simply closes.

**How to Reproduce**
Try to copy-paste the area "Area" in the enclosed study.

**Expected behavior**
The area should be correctly copy-pasted to a duplicate "Area-1" area

[Initial_study.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/8078236/Initial_study.zip)

**OS and version**
 - OS: Windows

flomnes:
use smartptr for `Action::Context::study`

Fix #549

### DIFF
--- a/src/libs/antares/study/action/context.h
+++ b/src/libs/antares/study/action/context.h
@@ -57,9 +57,9 @@ public:
 public:
     //! \name Constructor & Destructor
     //@{
-    explicit Context(Data::Study& targetStudy, const size_t layer = 0);
+    explicit Context(Data::Study::Ptr targetStudy, const size_t layer = 0);
     //! Destructor
-    ~Context();
+    ~Context() = default;
     //@}
 
     /*!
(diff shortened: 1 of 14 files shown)


src/libs/antares/study/action/context.h
```diff
@@ -69,12 +69,9 @@ class Context
 
 public:
     //! The target study
-    // TODO : use smart ptr here
-    Data::Study& study;
+    Data::Study::Ptr study;
     //! An optional external study, which will be destroyed with this class
-    // TODO : use smart ptr here
-    Data::Study* extStudy;
-    bool shouldDestroyExtStudy;
```
guilpier-code, line 77: You removed **shouldDestroyExtStudy**.
I guess we don't need it anymore because **extStudy** is now a shared pointer, that destroys the study when the scope containing the last reference to this pointer closes.

guilpier-code (approved):

src/libs/antares/study/action/context.h
flomnes: That's right, flags like this one are quite error-prone : memory can be leaked or de-allocated twice.

PR #552, opened 2022-02-16
Remove optImmediate from links' timeseries

### LINKED ISSUE #485: Make sure loading of links TS is just-in-time
TS should only be loaded when necessary to save time and I/O access. Such cases include :

- In the GUI, the user selects a data table
- In the solver, the data is required for computations

In particular, we don't want to load all TS when opening a study in the GUI since it would make loading time high.

flomnes:
Fix #485


flomnes:
Fix #485

guilpier-code (approved):

PR #556, opened 2022-02-18
Restore results for non-regression tests

flomnes:
In this PR, we change the code so as to remove **pathological differences**. Specifically, we partially revert
- aa186d0c Re-implement removeDisabledClusters()
- d170afc1 Take renewableGeneration into account for must run prod
- c36b913d Fix hydro heuristic (#375)

Please note that the two following PRs introduce **legitimate differences** :
- Fix hydro level discontinuities (#491)
- Fix weekly->daily interpolation for min/max reservoir level (#425)

We may still need to re-generate results for non-regression tests to include those legitimate differences.

### DIFF
--- a/src/solver/hydro/management/management.cpp
+++ b/src/solver/hydro/management/management.cpp
@@ -161,21 +161,26 @@ void HydroManagement::prepareNetDemand(uint numSpace)
             auto realmonth = study.calendar.months[month].realmonth;
             auto dayYear = study.calendar.hours[hour].dayYear;
 
-            double netdemand
-              = +scratchpad.ts.load[ptchro.Consommation][hour] - scratchpad.miscGenSum[hour]
-                - ror[hour]
-                - ((ModeT != Data::stdmAdequacy) ? scratchpad.mustrunSum[hour]
-                                                 : scratchpad.originalMustrunSum[hour]);
+            double netdemand = 0;
 
             // Aggregated renewable production: wind & solar
             if (parameters.renewableGeneration.isAggregated())
             {
-                netdemand -= scratchpad.ts.solar[ptchro.Solar][hour]
-                             + scratchpad.ts.wind[ptchro.Eolien][hour];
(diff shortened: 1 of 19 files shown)


src/solver/hydro/management/management.cpp
```diff
@@ -161,21 +161,26 @@ void HydroManagement::prepareNetDemand(uint numSpace)
...
-              = +scratchpad.ts.load[ptchro.Consommation][hour] - scratchpad.miscGenSum[hour]
-                - ror[hour]
-                - ((ModeT != Data::stdmAdequacy) ? scratchpad.mustrunSum[hour]
-                                                 : scratchpad.originalMustrunSum[hour]);
+            double netdemand = 0;
 
             // Aggregated renewable production: wind & solar
             if (parameters.renewableGeneration.isAggregated())
             {
-                netdemand -= scratchpad.ts.solar[ptchro.Solar][hour]
```
guilpier-code, line 173: There is code duplication here (load, miscGen, ror, mustRun), but if this makes non-regression disappear, why not ?

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -597,24 +597,25 @@ void SIM_RenseignementProblemeHebdo(PROBLEME_HEBDO& problem,
             }
 
             uint tsFatalIndex = (uint)tsIndex.Hydraulique < ror.width ? tsIndex.Hydraulique : 0;
-            problem.AllMustRunGeneration[j]->AllMustRunGenerationOfArea[k]
-              = scratchpad.miscGenSum[indx] + ror[tsFatalIndex][indx] + scratchpad.mustrunSum[indx];
-
+            double& mustRunGen = problem.AllMustRunGeneration[j]->AllMustRunGenerationOfArea[k];
             if (parameters.renewableGeneration.isAggregated())
             {
-                problem.AllMustRunGeneration[j]->AllMustRunGenerationOfArea[k]
```
guilpier-code, line 605: Same remark as in **management.cpp**

PR #561, opened 2022-02-21
FLOW QUAD : use batches to compute the average hourly NTC

flomnes:
This PR intends to speed up the computation of FLOW QUAD results.

The archetypal example is the following : N MC years with all the same NTC timeseries, with weight = 1 for all MC years. There is no need to compute `(capa+capa+...+capa)/N`. Instead, we compute `N*capa/N`, which is roughly N times faster.


flomnes:
Add test & solve conflicts.

guilpier-code (approved):

PR #563, opened 2022-02-21
Doc/replace docx by markdown generation

pl-buiquang:


flomnes (commented):
I think that doc generation in Windows is tricky. However, we could generate the .pdf in Ununtu and dispatch it to other platforms as an artifact, right ?

PR #568, opened 2022-02-21
Fix infinite capacity bug

### LINKED ISSUE #567: Antares V8.2-rc1 - Infinite NTCs are not taken into account in the calculation
**Description**
In the attached study, NTC are set as infinite in the unique link represented. However, in the results, it seems like this is not taken into account, and the NTC values from the Time-Series are used. This is true when NTC are set to infinite locally or as a global parameter of the study.

**How to Reproduce**
See attached study.

**Expected behavior**
When NTCs are set to infinite, there should be no restriction on the maximum flow on the corresponding link.

[study_ntc_example.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/8110111/study_ntc_example.zip)

**OS and version**
 - OS: [e.g. Windows, GNU/Linux]
 - Antares Simulator version: [e.g. 6, 7.2, 8]

**Additional context**

flomnes:
Fix #567

### DIFF
--- a/src/libs/antares/study/area/links.cpp
+++ b/src/libs/antares/study/area/links.cpp
@@ -711,8 +711,8 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
             {
                 // Copper plate mode
                 auto infinity = +std::numeric_limits<double>::infinity();
-                link.directCapacities.fillColumn(0, +infinity);
-                link.indirectCapacities.fillColumn(0, +infinity);
+                link.directCapacities.fill(+infinity);
+                link.indirectCapacities.fill(+infinity);
                 break;
             }
             }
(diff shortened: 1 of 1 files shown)


src/libs/antares/study/area/links.cpp
```diff
@@ -711,8 +711,8 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
             {
                 // Copper plate mode
                 auto infinity = +std::numeric_limits<double>::infinity();
-                link.directCapacities.fillColumn(0, +infinity);
-                link.indirectCapacities.fillColumn(0, +infinity);
+                link.directCapacities.fill(+infinity);
```
guilpier-code: There 2 symbols **"+"** here : one in the definition of **infinity**, and one when using **infinity** as argument of the **fill** function. Not sure the second one is required.
But otherwise, ok !

Issue #566, opened 2022-02-21
Antares V8.2-rc1 - Global link parameters not refreshed in GUI


PPlessiez:
**Description**
In the GUI, when a global link parameter (use of transmission capacities, hurdle costs, asset type...) is changed in one of the "Parameters"/"Transmission capacities" tabs, the change is not reflected in the other tab unless the study is saved.

**Expected behavior**
When one of the parameters is changed in one of the "Parameters"/"Transmission capacities" tabs, the change should be visible in the other tab without having to save the study.

**OS and version**
 - OS: Windows
 - Antares Simulator version: 8.2-rc1

flomnes:
Bug is confirmed.

@guilpier-code There seems to be a few missing refreshes.

PR #572, opened 2022-02-22
Link upper banner : update the banner (or board) when changing of tab

guilpier-code:
When we change a link property (transmission capacity usage, hurdle costs usage, ...) in the upper banner, and than
change of tab  (for example from Transmission capacities to Parameters), his upper banner is not properly refreshed.
This is fixed here.

See issue #566

### DIFF
--- a/src/ui/simulator/windows/connection.cpp
+++ b/src/ui/simulator/windows/connection.cpp
@@ -99,10 +99,10 @@ void linkNTCgrid::add(wxBoxSizer* sizer,
 }
 
 // Events to update a link property in all Interconnection objects (upper banner for any link view) 
-Yuni::Event<void(Antares::Data::AreaLink*)> onTransmissionCapacitiesUsageChanges;
-Yuni::Event<void(Antares::Data::AreaLink*)> onHurdleCostsUsageChanges;
-Yuni::Event<void(Antares::Data::AreaLink*)> onAssetTypeChanges;
-Yuni::Event<void(Antares::Data::AreaLink*)> onLinkCaptionChanges;
+Yuni::Event<void(const Antares::Data::AreaLink*)> onTransmissionCapacitiesUsageChanges;
+Yuni::Event<void(const Antares::Data::AreaLink*)> onHurdleCostsUsageChanges;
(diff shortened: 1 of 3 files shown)


src/ui/simulator/windows/connection.cpp
```diff
@@ -98,6 +98,13 @@ void linkNTCgrid::add(wxBoxSizer* sizer,
     gridIndirect->setOtherGrid(gridDirect);
 }
 
+// Events to update a link property in all Interconnection objects (upper banner for any link view)
```
guilpier-code, line 101: I tried to put the definition of this events inside class  **Interconnection**, where it belongs I think.
But it does not work : It compiles but these events are not caught, and so the view update is not done.
Strange. Maybe it should be investigated.

src/ui/simulator/windows/connection.cpp
```diff
@@ -148,6 +155,7 @@ Interconnection::Interconnection(wxWindow* parent,
         sizer_flex_grid->Add(button, 0, wxLEFT | wxALIGN_LEFT | wxALIGN_CENTER_VERTICAL);
     }
     // Caption
+    onLinkCaptionChanges.connect(this, &Interconnection::updateLinkCaption);
```
guilpier-code, line 158: Possible improvement : each button of the upper banner (hurdle costs usage, transmission capacities usage, ...) should be objects inner to the class  **Interconnection**, probably created outside this class and passed to its constructor. All these little button classes have things in common, they could inherit from an mother abstract class.

guilpier-code: Another thing about the previous refactoring : passing button objects to the **Interconnection** constructor would have another advantage than modularity. these objects could be shared by the 2 **Interconnection** objects (that is te 2 tabs **Parameters** and **Transmission capacities**), and we should not need to throw any events to update the button values in the other tab.

flomnes (approved):
Can be merged as is. Create an issue to solve this later ?

flomnes (changes requested):
De-duplicate objects ?

flomnes (approved):

PR #574, opened 2022-02-23
Fix asset type enum

flomnes:


flomnes:
This fix has been obsoleted by #572

PR #582, opened 2022-03-01
Adequacy first step initialization

hugo-antoine-rtei:
Runs the optimization once for the adq patch, gathers the DENS value and then runs it again. See [our internal PR](https://github.com/rte-i/Antares_Simulator/pull/10) for more info.

### DIFF
--- a/src/solver/simulation/sim_calcul_economique.cpp
+++ b/src/solver/simulation/sim_calcul_economique.cpp
@@ -63,6 +63,8 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
     problem.hydroHotStart
       = (parameters.initialReservoirLevels.iniLevels == Antares::Data::irlHotStart);
 
+    problem.AdequacyFirstStep = parameters.include.adequacyPatch;
+
     problem.WaterValueAccurate
       = (study.parameters.hydroPricing.hpMode == Antares::Data::HydroPricingMode::hpMILP)
           ? OUI_ANTARES
--- a/src/solver/simulation/sim_calcul_economique.cpp
+++ b/src/solver/simulation/sim_calcul_economique.cpp
@@ -114,6 +114,7 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
         auto& area = *(study.areas[i]);
 
         problem.NomsDesPays[i] = area.id.c_str();
+        problem.AreaAdequacyPatchMode[i] = area.adequacyPatchMode;
 
         problem.CoutDeDefaillancePositive[i] = area.thermal.unsuppliedEnergyCost;
 
--- a/src/solver/simulation/sim_calcul_economique.cpp
+++ b/src/solver/simulation/sim_calcul_economique.cpp
@@ -63,6 +63,7 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
     problem.hydroHotStart
       = (parameters.initialReservoirLevels.iniLevels == Antares::Data::irlHotStart);
 
+    problem.UseAdequacyPatch = parameters.include.adequacyPatch;
     problem.AdequacyFirstStep = parameters.include.adequacyPatch;
 
     problem.WaterValueAccurate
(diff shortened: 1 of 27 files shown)


src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -63,6 +63,9 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
     problem.hydroHotStart
       = (parameters.initialReservoirLevels.iniLevels == Antares::Data::irlHotStart);
 
+    problem.UseAdequacyPatch = parameters.include.adequacyPatch;
+    problem.AdequacyFirstStep = parameters.include.adequacyPatch;
+
```
flomnes, line 68: Maybe you could use a smart ptr here, something like

```cpp
struct AdequacyPatch {
bool firstStep = true;
// ...
};
std::unique_ptr<AdequacyPatch> adqPatch = nullptr;
```
So that you can write something like
```cpp
if (problem.adqPatch && problem.adqPatch->firstStep)
// ...
```

Replace `unique_ptr` by `shared_ptr` if `problem` needs to be copied.

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -478,6 +479,7 @@ struct PROBLEME_HEBDO
     char OptimisationAvecCoutsDeDemarrage;
     int NombreDePays;
     const char** NomsDesPays;
+    int* AreaAdequacyPatchMode;
```
flomnes, line 482: Consider using an enum here to avoid conversions back and forth.

src/solver/variable/economy/all.h
```diff
@@ -59,6 +59,7 @@
 #include "waterValue.h"
 #include "hydroCost.h"
 #include "unsupliedEnergy.h"
+#include "domesticUnsuppliedEnergy.h"
 #include "spilledEnergy.h"
```
flomnes, line 63: Make sure that the simulation mode is **economy** when using the adequacy patch ?

src/solver/variable/economy/all.h
```diff
@@ -214,30 +216,32 @@ typedef // Prices
...
-                                                  LOLP,
-
+                                                  LOLD,
                                                   Common::SpatialAggregate<
-                                                    AvailableDispatchGen,
+                                                    LOLP,
+
                                                     Common::SpatialAggregate<
-                                                      DispatchableGenMargin,
+                                                      AvailableDispatchGen,
```
flomnes, line 229: Can you explain this change ?

src/solver/variable/economy/domesticUnsuppliedEnergy.h
```diff
@@ -0,0 +1,288 @@
...
+namespace Variable
+{
+namespace Economy
+{
+struct VCardDomesticUnsuppliedEnergy
+{
+    //! Caption
+    static const char* Caption()
+    {
+        return "LOCAL UNSP. ENRG";
```
flomnes, line 45: I suggest a shorter name, such as "DENS" because the columns are narrow in the GUI, so I'm quite sure this wouldn't fit.

src/solver/variable/economy/domesticUnsuppliedEnergy.h
```diff
@@ -0,0 +1,288 @@
...
+    }
+    //! Unit
+    static const char* Unit()
+    {
+        return "MWh";
+    }
+    //! The short description of the variable
+    static const char* Description()
+    {
+        return "Domestic Energy Not Served (demand that cannot be satisfied before exchange)";
```
flomnes, line 55: Without interconnections/links. I suggest sticking to the Antares vocabulary : unsupplied energy.

This is essentially the same concept, so it would only make it harder to understand.

flomnes (changes requested):

src/solver/variable/economy/all.h
hugo-antoine-rtei: We're adding DomesticUnsuppliedEnergy to the list of variables. I don't know why everything else got deleted&added like this, but that's the only actual change.

PR #584, opened 2022-03-04
UI : fix assertion failures in debug mode

flomnes:
Fix assertion failures in
- Study logs viewer
- Resource monitor
- Optimization preferences
- Kirchoff constraint builder

wxWidgets complains because in a vertical sizer there is no need to center vertically (resp. horizontal, horizontally).


flomnes:
TODO : Kirchoff constraint generator

guilpier-code (approved):

PR #591, opened 2022-03-10
Add noise in hydro allocation cost vectors

flomnes:


flomnes:
TODO : display the cost value without noise in the cost vector.

PR #594, opened 2022-03-11
Disable JIT for link parameters

### LINKED ISSUE #593: Crash when opening links
**Description**
The GUI crashes when trying to open the links.

**How to Reproduce**
1. Have multiple NTC timeseries
2. Run a study
3. Navigate in the outputs
4. Go back to the inputs
5. Select the panel "Links"
6. The GUI crashes

**Expected behavior**
No crash.

**OS and version**
 - OS: Windows

flomnes:
Keeping it enabled may cause some issues when all coefficients are 0.

### DIFF
--- a/src/libs/antares/study/area/links.cpp
+++ b/src/libs/antares/study/area/links.cpp
@@ -217,6 +217,13 @@ bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& fol
     String filename;
     bool success = true;
 
+    bool enabledModeIsChanged = false;
+    if (JIT::enabled)
+    {
+        JIT::enabled = false; // Allowing to read the area's daily max power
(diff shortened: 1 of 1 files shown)


src/libs/antares/study/area/links.cpp
```diff
@@ -217,14 +217,23 @@ bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& fol
     String filename;
     bool success = true;
 
+    bool enabledModeIsChanged = false;
```
guilpier-code, line 220: Class Matrix is so messy that we're compelled to do things like that !
We should clean this class some day

flomnes: I completely agree. Matrix + JIT is a mess.

PR #595, opened 2022-03-11
Save NTC timeseries even if all coefficients are zero

### LINKED ISSUE #592: It is not possible to have multiple zero NTC TS
**Description**
Create a link between two areas. Set all coefficients to 0 for the NTC. The matrix is resized to 0x0, thus the number of direct and indirect TS is not equal. The study crashes at runtime.

**How to Reproduce**
See above.

**Expected behavior**
All coefficients should be saved, even if they're all 0.

**OS and version**
 - OS: All

flomnes:


src/libs/antares/array/matrix.hxx
```diff
@@ -1202,7 +1205,7 @@ void Matrix<T, ReadWriteT>::saveToBuffer(std::string& data,
         isDecimal = Static::Type::IsDecimal<ReadWriteType>::Yes,
     };
 
-    if (not print_dimensions and containsOnlyZero(predicate))
+    if ((not print_dimensions and containsOnlyZero(predicate)) && !saveEvenIfAllZero)
```
guilpier-code: I would put it that way : 
```suggestion
    if (not print_dimensions and (containsOnlyZero(predicate) and not saveEvenIfAllZero))
```
more clear to me

guilpier-code (changes requested):

src/libs/antares/array/matrix.hxx
flomnes: Operator `and` is associative _ie_ `(x and y) and z == x and (y and z)`, but I agree.

In such cases, you can suggest changes using this button 
image

guilpier-code (approved):

PR #597, opened 2022-03-15
Adequacy Patch - Local Matching Rule 15/03

hugo-antoine-rtei:
Hello Florian,

Here's the latest version of the adequacy patch we have at RTEi. I think we took care of most of what you brought to our attention last time. If you have some time we would like to get some feedback from it, as well as get an idea for the next steps that we should follow.

We also created a test study ([Test_study (1-week).zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/8249945/Test_study.1-week.zip)), and we'd like to have a bit of insight on how to make it compliant with the ones found in the [test repository](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests). Then I guess we just need to create a PR to add it to the repository and to the src/tests/examples/example_test.py file, and that part should be taken care of.


src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
```diff
@@ -176,11 +178,39 @@ void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(PROBLEME_HEBDO* Prob
...
+            else if (SetToZeroLinkNTCForAdequacyPatchFirstStep == Data::AdequacyPatch::setOrigineExtremityToZero)
+            {
+                Xmax[Var] = 0.;
+                Xmin[Var] = -(ValeursDeNTC->ValeurDeNTCExtremiteVersOrigine[Interco]);
+            }
+            else if (SetToZeroLinkNTCForAdequacyPatchFirstStep == Data::AdequacyPatch::setExtremityOrigineToZero)
+            {
+                Xmax[Var] = ValeursDeNTC->ValeurDeNTCOrigineVersExtremite[Interco];
+                Xmin[Var] = 0.;
+            }
```
flomnes: The code would be easier to read with something like that :
```cpp
if (ProblemeHebdo->adqPatch)
{
    setBoundsAdqPatch(Xmin, Xmax/*, etc */);
} 
else 
{
    setBoundsNoAdqPatch(Xmin, Xmax/*, etc */);
}
```

Leave lines 214-234 (handling of infinite values) as they are right now. Don't be shy : create new functions where needed.

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
```diff
@@ -401,9 +431,18 @@ void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(PROBLEME_HEBDO* Prob
                 else
                     Xmax[Var] = 0.;
 
+                // adq patch: update ENS <= DENS in 2nd run
+                if (ProblemeHebdo->adqPatch && ProblemeHebdo->adqPatch->AdequacyFirstStep == false)
+                    if (ProblemeHebdo->AreaAdequacyPatchMode[Pays]
```
flomnes: These two `if`s can be merged into one.

src/solver/simulation/economy.cpp
```diff
@@ -146,7 +146,40 @@ bool Economy::year(Progression::Task& progression,
...
+                          pProblemesHebdo[numSpace]->ResultatsHoraires[pays]->ValeursHorairesDENS,
+                          0,
+                          pProblemesHebdo[numSpace]->NombreDePasDeTemps * sizeof(double));
+                }
+
+                // TODO check if we need to cut SIM_RenseignementProblemeHebdo and just pick out the
+                // part that we need
+                ::SIM_RenseignementProblemeHebdo(
+                  *pProblemesHebdo[numSpace], state, numSpace, hourInTheYear);
+                OPT_OptimisationHebdomadaire(pProblemesHebdo[numSpace], numSpace);
```
flomnes: Create a new function called `OPT_OptimisationHebdomadaireAdqPatch`, call it here.

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -478,11 +489,14 @@ struct PROBLEME_HEBDO
...
     const char** NomsDesPays;
+    AdequacyPatchMode* AreaAdequacyPatchMode;
     int NombreDePaliersThermiques;
 
     int NombreDInterconnexions;
     int* PaysOrigineDeLInterconnexion;
     int* PaysExtremiteDeLInterconnexion;
+    AdequacyPatchMode* StartAreaAdequacyPatchType;
+    AdequacyPatchMode* EndAreaAdequacyPatchType;
     COUTS_DE_TRANSPORT** CoutDeTransport;
```
flomnes: I may be wrong, but I think `AreaAdequacyPatchMode`, `StartAreaAdequacyPatchType` and `EndAreaAdequacyPatchType` can be defined in `AdequacyPatchParameters`. Then they would only be allocated if needed : when the adq-patch is enabled.

Otherwise memory consumption would increase even if the adq-patch is disabled, which is not desirable.

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -469,6 +473,13 @@ typedef struct
     double* CoutsMarginauxHorairesDeLaReserveParZone;
 } COUTS_MARGINAUX_ZONES_DE_RESERVE;
 
+struct AdequacyPatchParameters
+{
+    bool AdequacyFirstStep;
+    bool Ntc12;
+    bool Ntc11;
```
flomnes: Please rename these two variables to make them more expressive.

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: **Origin** and **extremity** would be more appropriate than **Start** and **End** in my opinion.

src/ui/simulator/windows/inspector/property.update.cpp
```diff
@@ -172,6 +172,29 @@ bool InspectorGrid::onPropertyChanging_A(wxPGProperty*,
...
+
+                if (s == "virtual area" || s == "0")
+                    area->adequacyPatchMode = Data::AdequacyPatch::adqmVirtualArea;
+                else if (s == "physical area outside patch" || s == "1")
+                    area->adequacyPatchMode = Data::AdequacyPatch::adqmPhysicalAreaOutsideAdqPatch;
+                else if (s == "physical area inside patch" || s == "2")
+                    area->adequacyPatchMode = Data::AdequacyPatch::adqmPhysicalAreaInsideAdqPatch;
+            }
+        }
+        OnStudyAreaUseAdequacyPatchChanged();
```
flomnes, line 195: This event isn't connected to any function call. We usually bind functions to such events to refresh other UI components, but there are no other components to refresh in this case. Consider removing `OnStudyAreaUseAdequacyPatchChanged`.

src/ui/simulator/windows/options/optimization/optimization.cpp
```diff
@@ -308,6 +323,52 @@ Optimization::Optimization(wxWindow* parent) :
...
+        button = new Component::Button(this, wxT("true"), "images/16x16/light_green.png");
+        button->SetBackgroundColour(bgColor);
+        button->menu(true);
+        onPopup.bind(this,
+                     &Optimization::onPopupMenuSpecify,
+                     PopupInfo(study.parameters.include.adequacyPatch, wxT("true")));
+        button->onPopupMenu(onPopup);
+        s->Add(label, 0, wxRIGHT | wxALIGN_RIGHT | wxALIGN_CENTER_VERTICAL);
+        s->Add(button, 0, wxLEFT | wxALIGN_LEFT | wxALIGN_CENTER_VERTICAL);
+        pBtnAdequacyPatch = button;
```
flomnes: I suggest creating a window dedicated to the adq-patch. Something similar to
image


image

In the same window.

src/ui/simulator/windows/inspector/property.update.cpp
```diff
@@ -172,6 +172,29 @@ bool InspectorGrid::onPropertyChanging_A(wxPGProperty*,
...
+                wxStringToString(value.GetString(), s);
+                s.toLower();
+                s.trim();
+
+                if (s == "virtual area" || s == "0")
+                    area->adequacyPatchMode = Data::AdequacyPatch::adqmVirtualArea;
+                else if (s == "physical area outside patch" || s == "1")
+                    area->adequacyPatchMode = Data::AdequacyPatch::adqmPhysicalAreaOutsideAdqPatch;
+                else if (s == "physical area inside patch" || s == "2")
+                    area->adequacyPatchMode = Data::AdequacyPatch::adqmPhysicalAreaInsideAdqPatch;
```
flomnes, line 192: s will never take values 0, 1 or 2. Remove `|| s == "0"`, etc.

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -469,6 +473,13 @@ typedef struct
     double* CoutsMarginauxHorairesDeLaReserveParZone;
 } COUTS_MARGINAUX_ZONES_DE_RESERVE;
 
+struct AdequacyPatchParameters
```
flomnes, line 499: Rename to `AdequacyPatchRuntimeData`

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -59,9 +59,15 @@ void SIM_AllocationProblemeHebdo(PROBLEME_HEBDO& problem, int NombreDePasDeTemps
         problem.BruitSurCoutHydraulique[p] = (double*)MemAlloc(8784 * sizeof(double));
 
     problem.NomsDesPays = (const char**)MemAlloc(nbPays * sizeof(char*));
+    problem.AreaAdequacyPatchMode
+      = (Data::AdequacyPatch::AdequacyPatchMode*)MemAlloc(nbPays * sizeof(int));
```
flomnes: To be placed in the constructor of `AdequacyPatchRuntimeData`
```cpp
AdequacyPatchRuntimeData::AdequacyPatchRuntimeData(uint nbCountries, uint nbLinks) {
/// ...
}
```
You may use `std::vector<AdequacyPatchMode>` instead of `int*` and leave memory management to the standard library.

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: Also, instead of having two arrays for the link types (`StartAreaAdequacyPatchType` and `EndAreaAdequacyPatchType`), you could have a single array/vector containing objects of this new type
```cpp
struct {
   AdequacyPatchMode origin;
   AdequacyPatchMode extremity;
} AdqPatchLinkType;

File: src/solver/optimisation/adequacy_patch.cpp [hunk 11]
```diff
@@ -0,0 +1,75 @@
...
+*/
+
+#include "../simulation/simulation.h"
+#include "adequacy_patch.h"
+
+using namespace Antares;
+using namespace Antares::Data;
+using namespace Antares::Data::AdequacyPatch;
+
+NTC SetNTCForAdequacyFirstStep(bool AdequacyFirstStep,
```
flomnes: This function should be placed inside a specific namespace.

Also in my opinion, `using` directives should be kept to a minimum since `AdaquecyPatch::NTC` is more expressive than `NTC`.

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,75 @@
...
+            if (SetToZero11LinksForAdequacyPatch)
+                return setToZero;
+            else
+                return leaveLocalValues;
+        }
+        else
+            return leaveLocalValues;
+    }
+    else
+        return leaveLocalValues;
```
flomnes: This piece of code is hard to read because of nested if/else blocks. Maybe build a static `std::map` like the following ?

`std::map<std::pair<AdequacyPatchMode, AdequacyPatchMode>, NTC> behaviorMap;`

Then fill this map depending on `SetToZero12LinksForAdequacyPatch` and `SetToZero11LinksForAdequacyPatch`.

As pointed out in another comment, `std::pair<AdequacyPatchMode, AdequacyPatchMode>` could be a dedicated type.

Another possibility is to break it down into smaller functions.

src/solver/optimisation/adequacy_patch.h
```diff
@@ -0,0 +1,56 @@
...
+ * @param SetToZero12LinksForAdequacyPatch bool: Switch to cut links from nodes of adq type 1
+ * towards nodes of adq type 2
+ *
+ * @param SetToZero11LinksForAdequacyPatch bool: Switch to cut links from nodes of adq type 1
+ * towards nodes of adq type 1
+ *
+ * @return uint from an enumeration that describes the type of restrictions to put on this link for
+ * adq purposes.
+ */
+NTC SetNTCForAdequacyFirstStep(bool AdequacyFirstStep,
```
flomnes: `NTC` is pretty vague, consider using something more expressive. Also place this type in an adq-patch namespace.

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -423,6 +426,7 @@ typedef struct
 typedef struct
 {
     double* ValeursHorairesDeDefaillancePositive;
+    double* ValeursHorairesDENS; // adq patch DENS
```
flomnes: domestic unsupplied energy (comment)

flomnes (changes requested):
Indeed, it seems to me that the adq-patch is only for `studyMode == economy`. Please make sure that `antares-solver` won't run if adq-patch is enabled and the study mode is not **economy** (adequacy or adequacy-draft). See **src/solver/application.cpp**, lines 47-141 for how to achieve that. Make sure to define a new exception in **src/libs/antares/exception/LoadingError.hpp**, derived from `Error::LoadingError` so that it will be caught appropriately.

src/ui/simulator/windows/inspector/property.update.cpp
flomnes: Sorry, after checking how it's done for other enums, you need to call `value.GetLong()` and then `switch..case`. See **src/ui/simulator/windows/inspector/property.update.cpp:905-924** for an example.

So the value will never be "virtual area", etc.

src/ui/simulator/windows/options/optimization/optimization.cpp
hugo-antoine-rtei: Can you elaborate on this point ? How would that window be accessed ? Is there a piece of code we can look at with similar behavior ?

flomnes: I suggest that a new window be created, that could be accessed from the menu "Configure". It would improve the ease of use for this new feature by making all the info accessible from one place.

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
hugo-antoine-rtei: solved with [r2 review changes part 1 ](https://github.com/rte-i/Antares_Simulator/pull/30)

src/solver/simulation/economy.cpp
hugo-antoine-rtei: solved with [r2 review changes part 1 ](https://github.com/rte-i/Antares_Simulator/pull/30)

src/solver/simulation/sim_structure_probleme_economique.h
hugo-antoine-rtei: solved with [r2 review changes part 1 ](https://github.com/rte-i/Antares_Simulator/pull/30)

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
hugo-antoine-rtei: solved in [#31](https://github.com/rte-i/Antares_Simulator/pull/31)

src/solver/simulation/sim_structure_probleme_economique.h
hugo-antoine-rtei: solved in [#31](https://github.com/rte-i/Antares_Simulator/pull/31)

src/solver/optimisation/adequacy_patch.cpp
hugo-antoine-rtei: solved in [#31](https://github.com/rte-i/Antares_Simulator/pull/31)

src/solver/optimisation/adequacy_patch.cpp
hugo-antoine-rtei: solved in [#31](https://github.com/rte-i/Antares_Simulator/pull/31)

src/solver/optimisation/adequacy_patch.h
hugo-antoine-rtei: solved in [#31](https://github.com/rte-i/Antares_Simulator/pull/31)

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,139 @@
...
+        return search->second;
+    else
+        return leaveLocalValues;
+}
+
+std::map<adqPair, AdequacyPatch::LinkCapacityForAdequacyPatchFirstStep> GenerateLinkRestrictionMapForAdqFirstStep(
+  bool SetToZero12LinksForAdequacyPatch,
+  bool SetToZero11LinksForAdequacyPatch)
+{
+    std::map<adqPair, AdequacyPatch::LinkCapacityForAdequacyPatchFirstStep> behaviorMap;
```
flomnes: using namespace AdequacyPatch;

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,139 @@
...
+    // No transfer of energy possible between physical areas part of the adequacy patch
+    behaviorMap.insert(std::make_pair(
+      std::make_pair(AdequacyPatch::adqmPhysicalAreaInsideAdqPatch, AdequacyPatch::adqmPhysicalAreaInsideAdqPatch), AdequacyPatch::setToZero));
+    // all else remains the same
+    behaviorMap.insert(
+      std::make_pair(std::make_pair(AdequacyPatch::adqmVirtualArea, AdequacyPatch::adqmVirtualArea), AdequacyPatch::leaveLocalValues));
+    behaviorMap.insert(std::make_pair(
```
flomnes: Go for if..else or switch, and divide into smaller functions.

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
flomnes:
```cpp
class AdequacyPatchRuntimeData
{
private:
   std::vector<double> AreaMode;
   std::vector<int> StartAreaAdequacyPatchType;
   /// ... etc
public:
   void initialize(uint nbCountries, uint nbLinks);
};
```
In `PROBLEME_HEBDO`, add another member of this type.

Here, call 
```cpp
if(adqPatch)
probleme.adqPatchRuntimeData.initialize(nbPays, linkCount);
```

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -845,6 +856,7 @@ void SIM_DesallocationProblemeHebdo(PROBLEME_HEBDO& problem)
         MemFree(problem.PaliersThermiquesDuPays[k]->PuissanceDisponibleEtCout);
         MemFree(problem.PaliersThermiquesDuPays[k]);
         MemFree(problem.ResultatsHoraires[k]->ValeursHorairesDeDefaillancePositive);
+        MemFree(problem.ResultatsHoraires[k]->ValeursHorairesDENS);
```
flomnes, line 848: Keep this

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: Forget that remark.

src/solver/application.cpp
```diff
@@ -84,6 +84,16 @@ void checkSimplexRangeHydroHeuristic(Antares::Data::SimplexOptimization optRange
     }
 }
 
+// Adequacy Patch can only be used with Economy Study/Simulation Mode.
+void checkAdqPatchStudyModeEconomuOnly(const bool adqPatchOn,
```
flomnes: Economy

src/solver/application.cpp
```diff
@@ -239,6 +249,8 @@ void Application::prepare(int argc, char* argv[])
 
     checkSimplexRangeHydroHeuristic(pParameters->simplexOptimizationRange, pStudy->areas);
 
+    checkAdqPatchStudyModeEconomuOnly(pParameters->include.adequacyPatch, pParameters->mode);
+
```
flomnes, line 274: Add a check here : if adq-patch is on and there is no area in the adq-patch : throw an error.

src/solver/optimisation/adequacy_patch.h
```diff
@@ -0,0 +1,89 @@
...
+
+/*!
+ * Generates map containing all possible relations of start&end area adq patch mode taking into
+ * consideration user interface input options.
+ *
+ * @param SetToZero12LinksForAdequacyPatch bool: Switch to cut links from nodes of adq type 1
+ * towards nodes of adq type 2
+ *
+ * @param SetToZero11LinksForAdequacyPatch bool: Switch to cut links from nodes of adq type 1
+ * towards nodes of adq type 1
```
flomnes: Type 1 ? Type 2 ? Explain here.

src/solver/optimisation/adequacy_patch.cpp
hugo-antoine-rtei: I'll take a look again

PR #604, opened 2022-03-17
Add noise to the cost in the daily & monthly allocation problem

flomnes:


src/solver/hydro/daily2/h2o2_j_donnees_optimisation.h
```diff
@@ -134,38 +135,44 @@ typedef struct
 
 } PROBLEME_HYDRAULIQUE_ETENDU;
 
+namespace Constants
+{
+constexpr double denom = 1e3;
```
flomnes: `noiseAmplitude = 1e-3;`

src/solver/hydro/monthly/h2o_m_resoudre_le_probleme_lineaire.c
```diff
@@ -102,7 +102,7 @@ void H2O_M_ResoudreLeProblemeLineaire(DONNEES_ANNUELLES* DonneesAnnuelles, int N
     Probleme->NombreMaxDIterations = -1;
```
guilpier-code, line 102: by the way, **H2O_M_EcrireJeuDeDonneesLineaireAuFormatMPS** function (in this file), is useless.
We should think about removing it.

src/solver/hydro/daily/Test.c
```diff
@@ -44,6 +44,7 @@ main(int argc, char** argv)
     NbReservoirs = 100;
```
guilpier-code: This file is a ghost : it's part of the git repo, but it's not in the **cmake** system (never compiled, or scanned when searching strings over the whole Visual studio solution).
It calls the **H2O_J_EcrireJeuDeDonneesLineaireAuFormatMPS** function, which is not called anywhere else.
It only prints an **mps** file (probably for visual check).
We should consider compiling it and do something with it, or remove it otherwise.

flomnes: Good idea, I'll remove these parts of the code.

src/solver/hydro/monthly/h2o_m_resoudre_le_probleme_lineaire.c
flomnes: Done

PR #609, opened 2022-03-18
Write 1 in the ts-numbers when series.width == 1

### LINKED ISSUE #587: Antares V8.2-rc2 - Incorrect TS numbers in the ts-numbers output directory
**Description**
In the attached study containing 2 links, there is only one Time-Series for the link area-area 3, and the scenario builder NTC TS are set at random for the 4 MC years of the study.
However, when running a simulation, the ts-numbers output directory contains the following vector for the area - area 3 link: (4, 4, 2, 3).

**Expected behavior**
The ts-numbers output text files should contain only TS numbers corresponding to the input TS numbers for a given links (in this specific case, (1, 1, 1, 1)).

[Initial_study_for_datastorm.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/8196045/Initial_study_for_datastorm.zip)

**OS and version**
 - OS: Windows
 - Antares Simulator version: 8.2-RC2

flomnes:
Before that patch, random numbers were drawn, ignored in the simulation, but still written in ts-numbers.

### DIFF
--- a/src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
+++ b/src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
@@ -70,7 +70,7 @@ static void InitializeTimeSeriesNumbers_And_ThermalClusterProductionCost(
             if (data.series.width == 1)
                 data.timeseriesNumbers[0][year] = 0;
 
-            ptchro.Consommation = (long)data.timeseriesNumbers[0][year];
+            ptchro.Consommation = data.timeseriesNumbers[0][year];
         }
         // Solar
         {
@@ -80,7 +80,7 @@ static void InitializeTimeSeriesNumbers_And_ThermalClusterProductionCost(
             if (data.series.width == 1)
                 data.timeseriesNumbers[0][year] = 0;
 
-            ptchro.Solar = (long)data.timeseriesNumbers[0][year];
+            ptchro.Solar = data.timeseriesNumbers[0][year];
         }
         // Hydro
         {
@@ -90,7 +90,7 @@ static void InitializeTimeSeriesNumbers_And_ThermalClusterProductionCost(
             if (data.count == 1)
                 data.timeseriesNumbers[0][year] = 0;
 
-            ptchro.Hydraulique = (long)data.timeseriesNumbers[0][year];
+            ptchro.Hydraulique = data.timeseriesNumbers[0][year];
             // Hydro - mod
             memset(ptvalgen.HydrauliqueModulableQuotidien, 0, nbDaysPerYearDouble);
         }
@@ -102,7 +102,7 @@ static void InitializeTimeSeriesNumbers_And_ThermalClusterProductionCost(
             if (data.series.width == 1)
                 data.timeseriesNumbers[0][year] = 0;
 
-            ptchro.Eolien = (long)data.timeseriesNumbers[0][year];
+            ptchro.Eolien = data.timeseriesNumbers[0][year];
         }
         // Renewable
         {
@@ -122,7 +122,7 @@ static void InitializeTimeSeriesNumbers_And_ThermalClusterProductionCost(
                 if (data.series.width == 1)
                     data.timeseriesNumbers[0][year] = 0;
 
-                ptchro.RenouvelableParPalier[index] = (long)data.timeseriesNumbers[0][year];
+                ptchro.RenouvelableParPalier[index] = data.timeseriesNumbers[0][year];
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
```diff
@@ -65,33 +65,44 @@ static void InitializeTimeSeriesNumbers_And_ThermalClusterProductionCost(
 
         // Load
         {
-            const Data::DataSeriesLoad& data = *area.load.series;
+            Data::DataSeriesLoad& data = *area.load.series;
             assert(year < data.timeseriesNumbers.height);
-            ptchro.Consommation
-              = (data.series.width != 1) ? (long)data.timeseriesNumbers[0][year] : 0; // zero-based
+            if (data.series.width == 1)
```
guilpier-code: As **timeseriesNumbers** was set previously in **timeseries-numbers.cpp**, it should already be set to **0** if there is only one available time series for some given area and year.
Said in another way, there should be a need for setting **timeseriesNumbers** here.

guilpier-code (changes requested):

src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
flomnes: What you mean is that `timeseriesNumbers` should be set to 0 if `width==1` in **timeseries-numbers.cpp**. Then we would introduce a dependency to timeseries in **timeseries-numbers.cpp**, I'm not sure we want to do that.

flomnes: Actually, that dependency already exists in **timeseries-numbers.cpp**. Therefore, the TS numbers provided by `TimeSeriesNumbers::Generate` are wrong when `width==1`. I agree : the fix should be applied there, not here.

guilpier-code:
As you now set **timeseriesNumbers** to **0** when there is only **1** TS for given year, area, ... in **timeseries-numbers.cpp**, file **alea_tirage_au_sort_chroniques.cpp** can be simplified. 
Example : 
```c++
ptchro.Consommation = data.timeseriesNumbers[0][year];
```

src/tests/src/solver/simulation/tests-ts-numbers.cpp
```diff
@@ -354,7 +353,6 @@ BOOST_AUTO_TEST_CASE(two_areas_3_renew_clusters_with_respectively_4_4_1_ready_ma
 	// TS number checks
 	uint year = 0;
 	BOOST_CHECK_EQUAL(rnCluster_12->series->timeseriesNumbers[0][year], rnCluster_11->series->timeseriesNumbers[0][year]);
-	BOOST_CHECK_EQUAL(rnCluster_21->series->timeseriesNumbers[0][year], rnCluster_11->series->timeseriesNumbers[0][year]);
```
guilpier-code, line 357: ok, so before the patch, **timeseriesNumbers** is to something != 0 even if the TS actual size is **1**.
The patch in this test makes **rnCluster_21** useless. And we should rename that test.
And maybe make another test that show te new behavior.

src/tests/src/solver/simulation/tests-ts-numbers.cpp
```diff
@@ -233,7 +233,6 @@ BOOST_AUTO_TEST_CASE(two_areas_3_thermal_clusters_with_respectively_4_1_4_ready_
 
 	// TS number checks
 	uint year = 0;
-	BOOST_CHECK_EQUAL(thCluster_12->series->timeseriesNumbers[0][year], thCluster_11->series->timeseriesNumbers[0][year]);
```
guilpier-code, line 236: Same remark as below (for renewable clusters)

guilpier-code (approved):

src/tests/src/solver/simulation/tests-ts-numbers.cpp
guilpier-code: bla

src/tests/src/solver/simulation/tests-ts-numbers.cpp
flomnes: I will just remove these obsolete clusters and rename the corresponding tests. My fix does not change the behavior of `Antares::Solver::TimeSeriesNumbers::Generate`.

The correct way to fix the bug would be to change `Antares::Solver::TimeSeriesNumbers::Generate`, but unfortunately it is called before the timeseries generator (the one that generates the coefficients of the TS, not the numbers). A large refactor would be necessary to include that change in `TimeSeriesNumbers::Generate`.

guilpier-code (approved):

PR #610, opened 2022-03-21
Fix scenario builder data loss when renaming area

### LINKED ISSUE #605: Loss of sc-builder data / assertion failure when renaming areas
### Description
In some cases, when renaming areas that are connected by a link, one of the following happens :

- In debug mode : assertion failure **src/libs/antares/study/scenario-builder/TSnumberData.cpp:556** 
```cpp
#ifndef NDEBUG
    if (pTSNumberRules.width)
    {
        assert(pTSNumberRules.width == pArea->links.size());
    }
#endif
```
- In release mode, `assert`s are not run. Instead, valgrind indicates invalid reads. The TS numbers are reset to `rand`.

### How to reproduce

1. Create an empty study
2. Create 2 areas, keep their default names "Area 1" and "Area 2"
3. Create a link between those 2 areas
4. Rename "Area 1" to "XX"
5. Rename "Area 2" to "AA"

flomnes:
Fixes #605

### DIFF
--- a/src/libs/antares/study/scenario-builder/TSnumberData.cpp
+++ b/src/libs/antares/study/scenario-builder/TSnumberData.cpp
@@ -561,15 +561,11 @@ void ntcTSNumberData::saveToINIFile(const Study& study, Yuni::IO::File::Stream& 
     {
         const Data::AreaLink* link = i.second;
         /*
-          When renaming an area, it may happen that from->id and with->id
-          are not in the expected order. In that case, we need to swap them prior to writing.
+          When renaming an area, it may happen that i.first is not the name
+          of the supporting area. We only trust from->id and to->id.
         */
-        Data::AreaName fromID = link->from->id;
-        Data::AreaName withID = link->with->id;
-        if (fromID > withID)
-        {
-            swap(fromID, withID);
-        }
+        const Data::AreaName& fromID = link->from->id;
+        const Data::AreaName& withID = link->with->id;
         for (uint y = 0; y != pTSNumberRules.height; ++y)
         {
             const uint val = pTSNumberRules[link->indexForArea][y];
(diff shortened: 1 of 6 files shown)


src/libs/antares/study/study.cpp
```diff
@@ -1022,12 +1022,12 @@ bool Study::areaRename(Area* area, AreaName newName)
     // Updating all hydro allocation
     areas.each([&](Data::Area& areait) { areait.hydro.allocation.rename(oldid, newid); });
 
+    ScenarioBuilderUpdater updaterSB(*this);
```
flomnes, line 1025: The destructor is called later, in particular after `areas.fixOrientationForAllInterconnections` (l. 1044). That's when the sc-builder tmp data is loaded into memory.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -560,13 +560,23 @@ void ntcTSNumberData::saveToINIFile(const Study& study, Yuni::IO::File::Stream&
     for (const auto& i : pArea->links)
     {
         const Data::AreaLink* link = i.second;
+        /*
```
guilpier-code, line 563: As you know, this is a workaround. 
This added block may not belong here, because : 
- it tries to do a job that should have been done before in the execution,
- it makes the current task less clear than it should.

Isn't it possible to make that job (for instance) in the **AreaList::renameArea(...)** function ?
If it's not, we should consider a refactoring to make it possible, I think.

guilpier-code (changes requested):

src/libs/antares/study/scenario-builder/TSnumberData.cpp
flomnes: You're right. I think lines 569-572 are not mandatory, I just need to make sure that areas have already been renamed when saving the temporary file sc-builder data.


The order is the following when renaming an area :
1. Rename areas
2. Save the sc. builder data in a temporary file
3. Change data in memory
4. Load the sc. builder data from the temporary file

We should not use `i.first` (l. 579) there is no way map `Area::links` is up-to-date at step 1, since it is updated (and must be updated) at step 3.

guilpier-code (approved):

PR #617, opened 2022-03-23
Do not perform quadratic opt if FLOW QUAD. is disabled

### LINKED ISSUE #614: Disable quadratic optimization if FLOW QUAD is disabled in thematic trimming

flomnes:
in the thematic trimming section.


JMJ-rte:
In order to benefit largely of this simplification, we should be default considering that flow-quad results are not required

JMJ-rte:
In order to benefit largely of this simplification, we should consider by default that flow-quad results are not required

guilpier-code (approved):

PR #618, opened 2022-03-24
Issue an info if misc. gen is used

### LINKED ISSUE #608: Deprecate "Misc. Gen"
This feature is no longer necessary since renewable generation now allows to de-aggregate fatal production.

flomnes:
Printing logs in case **misc. gen.** is used in a study


flomnes:
Write an `info` instead ?

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -485,6 +486,23 @@ void StudyRuntimeInfos::checkThermalTSGeneration(Study& study)
     });
 }
 
+static void checkMiscGenUsed(const AreaList& areas)
```
guilpier-code, line 482: - We have this kind of log message : 
log-misc-gen
so there is a log line for each area name.
For more tightness, we should print a list of area names in a row.
- Why not print a warning instead of a classic log ?
As is, users do not know they should use renewable clusters instead of misc. Gen.

guilpier-code (changes requested):

src/libs/antares/study/runtime/runtime.cpp
flomnes: - OK
- When loading the study, unless the user asks to ignore warnings, all warnings and errors are fatal. I'm not sure we'd want to go as far as this.

guilpier-code: Regarding second bullet, it seems that warnings do not disable stages optimization and results print.
I just made the test.
Besides, ticking option **ignore warnings** in the **Run the simulation** window do not seem to have an effect : warnings are still shown in a separate window at end of simulation.

flomnes: Done. All in one line + warning

PR #619, opened 2022-03-24
Fix ignored resize with certain sizes

### LINKED ISSUE #261: UI : Times series for clusters (thermal & renewable)
**Description**
This problem apply for any cluster, thermal or renewable.
In GUI, for a given cluster, trying to resize the number of time series to the number of columns that the time series table contains before the change + 3, has no effect.
For instance, if a cluster has 2 time series and we want to resize the number of TS to 5, nothing is done.
See section **Additional context** for a possible cause.

**How to Reproduce**
- Click on a cluster
- Select the **Time series** tab 
- Select **Dataset** (not **Values**)
- just on the right of **Dataset**, select **Resize columns to**
- then set number of columns to : current number of columns + 3

**Expected behavior**
For the above example (2 TS, resized to 5), we would expect to get 8 columns : 
- 5 columns of TS
- 3 columns for **average**, **min** and **max**

**OS and version**
 - OS: Windows, to be tested on Linux
 - Antares Simulator version: 8.0.2

**Additional context**
A possible reason causing this bug : 
before changing the number of TS, a count on how many columns is done.
But this count includes columns **average**, **min** and **max**, though it should not.
As the new number of columns equals the current one, nothing is changed.

flomnes:
Fix #261


src/ui/simulator/toolbox/components/datagrid/renderer.h
```diff
@@ -118,6 +118,13 @@ class IRenderer : public Yuni::IEventObserver<IRenderer>
     */
     virtual int width() const = 0;
 
+    //! \name Sizes
+    //@{
+    /*!
+    ** \brief Width excluding agregated columns such as min, max, average, etc.
+    */
+    virtual int widthExcludingAgregatedColumns() const;
```
guilpier-code: Aggregated means nothing to me. **widthExcludingExtraStatisticsColumns** ?

src/ui/simulator/toolbox/components/datagrid/modifiers.hxx
```diff
@@ -348,7 +348,7 @@ struct ModifierOperatorsData<modifierDataset>
         }
         case opResizeColumns:
         {
-            uint width = (uint)renderer->width();
+            const auto width = (uint)renderer->widthExcludingAgregatedColumns();
```
guilpier-code: It seems to fix the problem if we just remove the block : 
```C++
 if (newwidth == width)
       break; // nothing to do
```

guilpier-code (changes requested):

guilpier-code (approved):

PR #624, opened 2022-03-29
Antares solver options : fix a bug on Windows

guilpier-code:


src/solver/misc/options.h
```diff
@@ -27,6 +27,7 @@
 #ifndef __SOLVER_MISC_GETOPT_H__
 #define __SOLVER_MISC_GETOPT_H__
 
+#include<memory>
```
flomnes, line 30:
```suggestion
#include <memory>
```

flomnes (changes requested):
You could use a `unique_ptr`, right ?

flomnes (approved):

Issue #398, opened 2021-09-30
Log report window for some study


guilpier-code:
For study **MAF2025**, log report contains very large lines.
![study-log-report](https://user-images.githubusercontent.com/62292552/135484481-6b22336b-b247-4813-964c-f4bad24d4792.PNG)

There are 2 things we should enhance : 
- we should shorten the larges lines (no need for them to be shown thouroughly, as they can be found in the log file)
- the vertical scrolling is very slow (see if it is caused by the large lines that must be often refreshed or not)

Study can be found here : 
[case study.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/7260744/case.study.zip)

flomnes:
@guilpier-code If you want to reproduce the issue without having to run the case study for 30 minutes, you can add logs with something like this
```cpp
for (uint ii = 0; ii < 1000; ++ii)
    logs.warning() << "Lorem ipsum";
```

and run a simple study instead. This will save you a lot of time.

flomnes:
See **rules.cpp**. Limit that loop to 10 iterations
```cpp
for (; itv != scenariiForCurrentCluster.end(); itv++)
```

flomnes:
Vertical scrolling speed is now OK. However it would be nice to have an horizontal scroll / resizable window.

Issue #613, opened 2022-03-23
[Antares V8.2] Output log screen not adjusted to screen width


PPlessiez:
**Description**
When an Antares study is completed, the output pop-up screen is displayed showing the possible warnings/errors. In case the size of a warning is very large, the width of the pop-up is adjusted so this message can be displayed entirely. It is then very inconvenient to slide the pop-up to be able to reach the "Continue" button at the bottom-right, or to be able to read the message. It would be more appropriate to adapt the size of this screen to the width of the user screen.

**How to Reproduce**
Output pop-up log screen in case there is a long warning message (e.g. many areas with incorrect scenario builder configuration, which happens a lot in studies for which many areas are virtual areas).

**Expected behavior**
It would be more user-friendly to adapt the size of the log screen to the user screen rather than to the length of messages/

**Screenshots and additional files**
Example of a log screen output:
![screenshot_log_screen](https://user-images.githubusercontent.com/53614116/159661066-0adca749-69a3-4f5d-8ead-00888c3c898e.png)

**OS and version**
 - OS: Windows
 - Antares Simulator version: 8.2 (but the issue already exists in previous versions)

flomnes:
@PPlessiez Can you post a .zip containing the study ?

PPlessiez:
Here is an example of a study where the problem occurs:
https://postngo.rte-france.com/?lang=fr&a=d&i=uaRMclSRJAbtvjy5G0IF

guilpier-code:
Thanks Paul, but wow ! Big study ! :-)

PR #626, opened 2022-03-30
Scenario builder warnings too long

guilpier-code:
Warnings from scenario builder can be very long :
this happens when a cluster (thermal or renewable) is disabled and when some TS numbers are assigned to this cluster in the scenario builder.
As the cluster is disabled, Antares warns the user of all years where a TS number is set.
When this year collection grows, the size of the warning line grows accordingly.

This work fixes the issues #398 and #613 (both pointing the same problem)

### DIFF
--- a/src/libs/antares/study/scenario-builder/rules.cpp
+++ b/src/libs/antares/study/scenario-builder/rules.cpp
@@ -364,21 +364,22 @@ void Rules::sendWarningsForDisabledClusters()
         int nbScenariiForCluster = (int)scenariiForCurrentCluster.size();
         vector<uint>::iterator itv = scenariiForCurrentCluster.begin();
 
-        // Listing the 10 first scenarii numbers (years numbers) where current cluster is refered
-        // to. Notice that these scenarii could be less then 10, but are at least 1.
-        string listScenarii = to_string(*itv + 1);
+        // Listing the 10 first years for which the current cluster was given a specific TS number 
+        // in the scenario builder.
+        // Note that this list of years size could be less then 10, but are at least 1.
+        string listYears = to_string(*itv + 1);
         itv++;
-        for (; itv != scenariiForCurrentCluster.end(); itv++)
-            listScenarii += ", " + to_string(*itv + 1);
+        for (; itv != scenariiForCurrentCluster.end() && *itv + 1 <= 10; itv++)
+            listYears += ", " + to_string(*itv + 1);
 
         // Adding last scenario to the list
         if (nbScenariiForCluster > 10)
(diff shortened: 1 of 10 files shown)


src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -364,21 +364,22 @@ void Rules::sendWarningsForDisabledClusters()
...
-        // to. Notice that these scenarii could be less then 10, but are at least 1.
-        string listScenarii = to_string(*itv + 1);
+        // Listing the 10 first years for which the current cluster was given a specific TS number 
+        // in the scenario builder.
+        // Note that this list of years size could be less then 10, but are at least 1.
+        string listYears = to_string(*itv + 1);
         itv++;
-        for (; itv != scenariiForCurrentCluster.end(); itv++)
-            listScenarii += ", " + to_string(*itv + 1);
+        for (; itv != scenariiForCurrentCluster.end() && *itv + 1 <= 10; itv++)
```
flomnes: No message is displayed if a study contains e.g 50 years, but the 10 first years are disabled. I recommend using a container, then aggregate it into a string.

guilpier-code: You're right, changes done

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -364,21 +364,22 @@ void Rules::sendWarningsForDisabledClusters()
...
-        // to. Notice that these scenarii could be less then 10, but are at least 1.
-        string listScenarii = to_string(*itv + 1);
+        // Listing the 10 first years for which the current cluster was given a specific TS number 
+        // in the scenario builder.
+        // Note that this list of years size could be less then 10, but are at least 1.
+        string listYears = to_string(*itv);
         itv++;
-        for (; itv != scenariiForCurrentCluster.end(); itv++)
-            listScenarii += ", " + to_string(*itv + 1);
+        for (int year_count = 1; itv != scenariiForCurrentCluster.end() && year_count < 10; itv++, year_count++)
```
flomnes, line 372: Starting at 1, you'll only get 9 years. Start at 0 for 10 ?

flomnes:
Resolve conflicts, set number of displayed years to 10 (currently 9) and merge.

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: Before the loop **for**, **listYears** is filled with 1 year.
In the loop, **listYears** is filled with 9 more years.

PR #627, opened 2022-03-31
Avoid error messages on TS preprocessor content when thermal TS are ready made

### LINKED ISSUE #616: [Antares V8.2] Check on FO/PO duration is performed even if thermal TS are ready-made
**Description**
When an Antares simulation is launched, Antares checks if the FO/PO durations are realistic (> 0).  This check is performed even when the thermal TS are set to "ready-made", which is not useful as these parameters will not be used.

**How to Reproduce**
Launch an Antares simulation with FO/PO durations equal to 0 for some clusters, and thermal TS status set to "ready-made".

**Expected behavior**
The FO/PO duration consistency check should not be performed for thermal clusters for which no thermal TS will be generated.

**OS and version**
 - OS: Windows

### LINKED ISSUE #642: Wrong error messages when checking thermal prepro data
**Description**
Error messages regarding data used to build thermal TS (prepro) can be wrong.
These wrong messages occur for PO durations grid.
The messages currently point line numbers in the PO grid.
These numbers can be wrong.

**How to Reproduce**
Create a study with a thermal cluster.
For this cluster, set **0** on the first day in column **FO** duration.
Then set **0** on the second day in column **PO** duration.
Make sure thermal TS are generated. 
Run the study.

**Expected behavior**
The simulation stops with correct error messages.
Line number pointed by the error message for **PO** duration should be **2**. You get **1** (see screenshot)

error-msg

**Fixing the current bug**
The fix is tiny and will be integrated in PR #627

guilpier-code:


src/libs/antares/study/parts/thermal/prepro.h
```diff
@@ -116,9 +114,9 @@ class PreproThermal
     **
     ** This method should only be used by the solver
     */
-    bool normalizeAndCheckNPO(const AnyString& areaname,
-                              const AnyString& clustername,
-                              uint clusterSize);
+    bool normalizeAndCheckNPO();
+private:
+    ThermalCluster* itsThermalCluster = nullptr;
```
flomnes:
```suggestion
    const ThermalCluster* cluster = nullptr;
```

flomnes:
Confirmed : this PR fixes the described bug on a simple study.

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -719,9 +719,9 @@ void ThermalClusterList::ensureDataPrepro()
     auto end = cluster.end();
     for (auto it = cluster.begin(); it != end; ++it)
     {
-        auto& c = *(it->second);
-        if (not c.prepro)
-            c.prepro = new PreproThermal();
+        auto c = it->second;
+        if (not c->prepro)
+            c->prepro = new PreproThermal(c.get());
```
flomnes: Storing the underlying address of a smart ptr is **extremely dangerous**. The managed object could be destroyed, leading to a segfault. I strongly suggest that we keep a `std::shared_ptr<const ThermalCluster>` here.

flomnes (changes requested):
Please store a `std::shared_ptr` in class PreproThermal

src/libs/antares/study/parts/thermal/cluster_list.cpp
guilpier-code: Done

src/libs/antares/study/parts/thermal/prepro.h
guilpier-code: Done

src/libs/antares/study/parts/thermal/prepro.cpp
```diff
@@ -43,7 +43,11 @@ namespace Antares
 {
 namespace Data
 {
-PreproThermal::PreproThermal(ThermalCluster* thermalCluster) : itsThermalCluster(thermalCluster)
+PreproThermal::PreproThermal(const ThermalCluster* thermalCluster) : cluster_(thermalCluster)
+{
+}
```
flomnes: Is this constructor required ?

flomnes (approved):

PR #628, opened 2022-04-04
Use std::unique_ptr for GetOpt::Parser

flomnes:


flomnes:
Let's keep using `std::unique_ptr<T>(new T())` for now, then switch to C++17 in a separate commit.

Note : it might be necessary to upgrade OR-Tools (currently 8) to 9.2.

PR #632, opened 2022-04-05
Fix build when BUILD_SWAP=OFF

flomnes:
Previously, linking would fail because libmodel_antares-swap didn't exist


flomnes:
Crash in the GUI
```
0x0000555555f77d71 in Antares::Matrix<double, double>::fill (this=0x7fffcc1363e0, v=@0x7fffd3748ea0: 1)
    at Antares_Simulator/src/libs/antares/array/../study/constraint/../area/../../array/matrix.hxx:337
337	            column[j] = v;
```

guilpier-code (approved):

Issue #639, opened 2022-04-11
Thermal cluster TS generation management prevents runs in parallel mode


guilpier-code:
**How the reproduce** : 
- Run any study that has no reason not to run MC years on several threads.
- Look into the log file (**logs/solver-yyyymmdd-hhmmss**)
- Search for **parallel batch size :** 
- result is **1** for any batch of parallel years

You can test on this study : 
[hydro levels--heuristic-4y-v8.2.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/8465090/hydro.levels--heuristic-4y-v8.2.zip)

PR #640, opened 2022-04-11
Thermal clusters TS generation management prevents parallel running

guilpier-code:
Handles ticket #639


flomnes (changes requested):
- Can you explain the differences ?
- Why choose this base branch ?

flomnes (approved):

PR #651, opened 2022-04-19
Use std::enable_shared_from_this

flomnes:
Relevant example
```cpp
#include <iostream>
#include <memory>

class Resource : public std::enable_shared_from_this<Resource> {
public:
  explicit Resource(const char* msg) {
    std::cout << "Resource(" << msg << ")" << std::endl;
  }
  ~Resource() {
    std::cout << "~Resource()" << std::endl;
  }
};

int main() {
  auto s1 = std::make_shared<Resource>("Hello");
  auto ptr = s1.get();
  auto s2 = ptr->shared_from_this();

  std::cout << s1.use_count() << ' ' << s2.use_count() << std::endl;

  /*  Possible output
  Resource(Hello)
  2 2
  ~Resource()
  */
}
```

See [cppreference.com](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this) about this C++11 feature.

### DIFF
--- a/src/libs/antares/study/parts/thermal/cluster_list.cpp
+++ b/src/libs/antares/study/parts/thermal/cluster_list.cpp
@@ -721,7 +721,7 @@ void ThermalClusterList::ensureDataPrepro()
     {
         auto c = it->second;
         if (not c->prepro)
-            c->prepro = new PreproThermal(c.get());
+            c->prepro = new PreproThermal(c);
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -721,7 +721,7 @@ void ThermalClusterList::ensureDataPrepro()
     {
         auto c = it->second;
         if (not c->prepro)
-            c->prepro = new PreproThermal(c.get());
+            c->prepro = new PreproThermal(c);
```
guilpier-code, line 724: Why not : 
```c++
c->prepro = new PreproThermal(c->shared_from_this());
```
?

flomnes: Because `c` is already a `shared_ptr`, there is no need to create a new one.

guilpier-code: ok, I'll approve then

guilpier-code (approved):

PR #654, opened 2022-04-20
Cosmetic changes to the adequacy patch (pt. 1 - local matching rule)

flomnes:


src/solver/application.cpp
```diff
@@ -96,19 +98,14 @@ void checkAdqPatchStudyModeEconomyOnly(const bool adqPatchOn,
...
-        {
-            const auto& area = *(areas.byIndex[i]);
-            if (area.adequacyPatchMode
-                == Antares::Data::AdequacyPatch::adqmPhysicalAreaInsideAdqPatch)
-            {
-                containsAdqArea = true;
-                break;
-            }
-        }
+        const bool containsAdqArea
```
guilpier-code, line 104: Why not directly : 

```C++
if (! std::any_of(areas.cbegin(), areas.cend(), 
         [](const std::pair<AreaName, Area*>& area) 
         {
                return area.second->adequacyPatchMode
                          == AdequacyPatch::adqmPhysicalAreaInsideAdqPatch;
         }
     )
  )
{
      throw Error::NoAreaInsideAdqPatchMode();
}
```
?

guilpier-code (changes requested):

PR #657, opened 2022-04-21
Adequacy Patch - Local matching

flomnes:


src/libs/antares/study/area/list.cpp
```diff
@@ -1040,6 +1069,44 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
         area.renewable.list.flush();
     }
 
+    // Adequacy patch
+    if (study.header.version >= 820)
+    {
+        buffer.clear() << study.folderInput << SEP << "areas" << SEP << area.id << SEP
+                       << "adequacy_patch.ini";
+        IniFile ini;
```
guilpier-code: We could make a function/method with the following code block : 
```c++
void readAdqPatchMode(area&, buffer)
```

src/libs/antares/study/parameters.h
```diff
@@ -502,6 +505,14 @@ class Parameters final
     SimplexOptimization simplexOptimizationRange;
     //@}
 
+    //! Transmission capacities from physical areas outside adequacy patch (area type 1) to physical
+    //! areas inside adequacy patch (area type 2). NTC is set to null (if true) only in the first
+    //! step of adequacy patch local matching rule.
+    bool setToZero12LinksForAdequacyPatch;
```
guilpier-code: "12" or "11" are not expressive for anyone who did not read adequacy patch documentation.
We could rename these variables like : **setToZeroNTCfromOutToIn_AdqPatch**, which is not much longer.

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
```diff
@@ -177,9 +178,10 @@ void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(PROBLEME_HEBDO* Prob
             Var = CorrespondanceVarNativesVarOptim->NumeroDeVariableDeLInterconnexion[Interco];
             CoutDeTransport = ProblemeHebdo->CoutDeTransport[Interco];
 
-            Xmax[Var] = ValeursDeNTC->ValeurDeNTCOrigineVersExtremite[Interco];
-
-            Xmin[Var] = -(ValeursDeNTC->ValeurDeNTCExtremiteVersOrigine[Interco]);
+            if (ProblemeHebdo->adqPatch && ProblemeHebdo->adqPatch->AdequacyFirstStep)
```
guilpier-code: As you could see, this source file (like many others in the modeler) is messy because functions are too big.
We should avoid to add more lines to functions.
Particularly here : instead of having a **if... else...**, we should have a call to **setNTCbounds(...)**,
and put the **if... else...** inside.
Please also see the remarks regarding file **src\solver\optimisation\adequacy_patch.cpp**, they complete what I mean.

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+    LinkCapacityForAdequacyPatchFirstStep SetToZeroLinkNTCForAdequacyPatchFirstStep;
+
+    SetToZeroLinkNTCForAdequacyPatchFirstStep = SetNTCForAdequacyFirstStep(
+      ProblemeHebdo->adequacyPatchRuntimeData.originAreaType[Interco],
+      ProblemeHebdo->adequacyPatchRuntimeData.extremityAreaType[Interco],
+      ProblemeHebdo->adqPatch->LinkCapacityForAdqPatchFirstStepFromAreaOutsideToAreaInsideAdq,
+      ProblemeHebdo->adqPatch->LinkCapacityForAdqPatchFirstStepBetweenAreaOutsideAdq);
+
+    if (SetToZeroLinkNTCForAdequacyPatchFirstStep == setToZero)
```
guilpier-code: Here a **switch** would be better than a **if... elsif...elsif...else**, especially because of the **enum** variable to be tested.

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,146 @@
...
+        Xmin = -(ValeursDeNTC->ValeurDeNTCExtremiteVersOrigine[Interco]);
+    }
+    else if (SetToZeroLinkNTCForAdequacyPatchFirstStep == setExtremityOrigineToZero)
+    {
+        Xmax = ValeursDeNTC->ValeurDeNTCOrigineVersExtremite[Interco];
+        Xmin = 0.;
+    }
+    else
+    {
+        Xmax = ValeursDeNTC->ValeurDeNTCOrigineVersExtremite[Interco];
```
guilpier-code: Here you have a code duplication with the content of **setBoundsNoAdqPatch(...)**.
So you should consider calling this latter function from **setBoundsAdqPatch(...)**. 
Or, even better if it's possible, you could set **Xmin** and **Xmax** from **no adq patch** case as the default values before the **switch** statement, so that you would not have to treat the **no adq patch**
case in the **switch**.

src/libs/antares/study/fwd.h
```diff
@@ -557,4 +557,45 @@ class Sets;
...
+namespace AdequacyPatch
+{
+
+/*!
+** \brief Types of Adequacy patch mode
+*/
+enum AdequacyPatchMode
+{
+    //! Virtual area in adq patch
+    adqmVirtualArea = 0,
```
guilpier-code: What does the **adqm** suffixing each item mean ?
If you put **AdequacyPatchMode::** before items when using it, you probably won't need  **adqm**

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
```diff
@@ -401,9 +403,18 @@ void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(PROBLEME_HEBDO* Prob
                 else
                     Xmax[Var] = 0.;
 
+                // adq patch: update ENS <= DENS in 2nd run
```
guilpier-code, line 385: To me, the code format is unclear. I would put it like this : 
```c++
if (ProblemeHebdo->adqPatch && 
    ProblemeHebdo->adqPatch->AdequacyFirstStep == false &&
    ProblemeHebdo->adequacyPatchRuntimeData.areaMode[Pays] == AdequacyPatch::adqmPhysicalAreaInsideAdqPatch)
{
    Xmax[Var] = min(Xmax[Var], ProblemeHebdo->ResultatsHoraires[Pays]->ValeursHorairesDENS[PdtHebdo]);
}
```
You can even make use of **using namespace** upper in the file to shorten enum values references here.

src/solver/simulation/economy.cpp
```diff
@@ -146,7 +174,16 @@ bool Economy::year(Progression::Task& progression,
 
         try
         {
-            OPT_OptimisationHebdomadaire(pProblemesHebdo[numSpace], numSpace);
+            if (pProblemesHebdo[numSpace]->adqPatch)
```
guilpier-code: The choice of which problem to run should not be made here, but more in a construction phase. 
For instance in the **simulationBegin** function (upper in this file), or even better, in the **Economy** class constructor (if possible). 

A possible way of doing things would be  : 
- Creating an interface class **weeklyProblemInterface** with a **run** method.
- Creating 2 classes implementing **weeklyProblemInterface** :  **weeklyProblem** and **weeklyProblemWithAdequacyPatch**, with their own run method.
- As the "classic" **run** method would be called from the **weeklyProblemWithAdequacyPatch** (to avoid code duplication), this classic run could be implemented in the interface class, and called **"as is"** from the **weeklyProblem**'s run method.

Anyway, here we would only have : 
```c++
theWeeklyProblem->run(/*some arguments ?*/)
```
and that's it

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
guilpier-code, line 385: Furthermore, I guess you had to find a loop on areas (Pays) to insert bound setting on ENS.
Still to simplify the code here, you could extract this code into a separate function, even if you would have to extract the loop on areas within it.

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -63,6 +63,21 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
     problem.hydroHotStart
       = (parameters.initialReservoirLevels.iniLevels == Antares::Data::irlHotStart);
 
+    if (parameters.include.adequacyPatch)
+    {
+        problem.adqPatch = std::unique_ptr<AdequacyPatchParameters>(new AdequacyPatchParameters());
```
guilpier-code: Use of **make_unique** ?

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -317,6 +321,28 @@ typedef struct
...
 
+class AdequacyPatchRuntimeData
+{
+public:
+    std::vector<AdequacyPatchMode> areaMode;
+    std::vector<AdequacyPatchMode> originAreaType;
+    std::vector<AdequacyPatchMode> extremityAreaType; 
+    void initialize(Antares::Data::Study& study)
+    {
+        for (uint i = 0; i != study.areas.size(); ++i)
```
guilpier-code, line 333: - on one hand, we have area**Mode**, on the other we have originArea**Type**. From my understanding, these vectors contain the same kind of values. How about : **areasAdqPatchModes** and **originAreasAdqPatchModes** ?
- As there is 2 loops making different things, we could have 2 private methods to make them... Or not

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+void setBoundsAdqPatch(double& Xmax,
+                       double& Xmin,
+                       VALEURS_DE_NTC_ET_RESISTANCES* ValeursDeNTC,
+                       const int Interco,
+                       PROBLEME_HEBDO* ProblemeHebdo)
+{
+    LinkCapacityForAdequacyPatchFirstStep SetToZeroLinkNTCForAdequacyPatchFirstStep;
+
+    SetToZeroLinkNTCForAdequacyPatchFirstStep = SetNTCForAdequacyFirstStep(
```
guilpier-code, line 96: For me : 
- Too many arguments in this function, and it makes difficult to understand what the call does. Arguments here should be developed inside the function.
- The name of the function starts with "Set", but actually it is more a "get".
- This name could be shorten because, at this stage, we know we're in the adq patch first step, and it's about links. So "Links" and "first step" can be removed from the function name. 

I would put it this way :   
```c++
ntcToZeroStatusForAdqPatch = getNTCtoZeroStatus(Interco, ProblemeHebdo)
```

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -469,6 +496,13 @@ typedef struct
     double* CoutsMarginauxHorairesDeLaReserveParZone;
 } COUTS_MARGINAUX_ZONES_DE_RESERVE;
 
+struct AdequacyPatchParameters
+{
+    bool AdequacyFirstStep;
+    bool LinkCapacityForAdqPatchFirstStepFromAreaOutsideToAreaInsideAdq;
```
guilpier-code: Name a bit long and unclear for me.
Suggestion :  
- setToZeroNTCfromInToOut_AdqPatchStep1
- setToZeroNTCbetweenOutsideAreas_AdqPatchStep1

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,146 @@
...
+  bool SetToZero12LinksForAdequacyPatch)
+{
+    LinkCapacityForAdequacyPatchFirstStep returnNTC;
+
+    switch (ExtremityNodeAdequacyPatchType)
+    {
+    case adqmPhysicalAreaInsideAdqPatch:
+        returnNTC = setToZero;
+        break;
+    case adqmPhysicalAreaOutsideAdqPatch:
```
guilpier-code: Based on the documentation that I have, for a link of type **Area Inside** => **Area Outside** (that is 2 -> 1), its NTC is always set to zero.  So here, I suppose we should have : 
 ```C++
switch (ExtremityNodeAdequacyPatchType)
{
    case adqmPhysicalAreaInsideAdqPatch:
        returnNTC = setToZero;
        break;
    case case adqmPhysicalAreaOutsideAdqPatch:
       returnNTC = setExtremityOrigineToZero;
       break;
    default:
        returnNTC = leaveLocalValues;
        break;
}
```

File: src/libs/antares/study/fwd.h [hunk 14]
```diff
@@ -557,4 +557,45 @@ class Sets;
...
+    adqmPhysicalAreaOutsideAdqPatch,
+    //! Physical Area inside the adq-patch
+    adqmPhysicalAreaInsideAdqPatch
+
+}; // enum AdequacyPatchMode
+
+/*!
+** \brief Setting Link Capacity (NTC) for Adequacy patch first step
+*/
+enum LinkCapacityForAdequacyPatchFirstStep
```
guilpier-code, line 582: Wouldn't **ntcSetToZeroStatus_AdqPatchStep1** be better (more clear) ?

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,146 @@
...
+        returnNTC
+          = SetNTCForAdequacyFirstStepOriginNodeOutsideAdq(ExtremityNodeAdequacyPatchType,
+                                                           SetToZero12LinksForAdequacyPatch,
+                                                           SetToZero11LinksForAdequacyPatch);
+        break;
+    default:
+        returnNTC = leaveLocalValues;
+        break;
+    }
+    return returnNTC;
```
guilpier-code: You don't need a local variable **returnNTC**,  you should return directly the result of functions or return  **leaveLocalValues**

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeInsideAdq(
+  AdequacyPatchMode ExtremityNodeAdequacyPatchType,
+  bool SetToZero12LinksForAdequacyPatch)
+{
+    LinkCapacityForAdequacyPatchFirstStep returnNTC;
```
guilpier-code: As above, you don't need **returnNTC**

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return returnNTC;
+}
+
+LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeOutsideAdq(
+  AdequacyPatchMode ExtremityNodeAdequacyPatchType,
+  bool SetToZero12LinksForAdequacyPatch,
+  bool SetToZero11LinksForAdequacyPatch)
+{
+    LinkCapacityForAdequacyPatchFirstStep returnNTC;
```
guilpier-code: As above, you don't need **returnNTC**

src/ui/simulator/application/study.cpp
```diff
@@ -90,6 +90,7 @@ Event<void(Data::AreaLink*)> OnStudyLinkDelete;
 
 Event<void()> OnStudySimulationSettingsChanged;
 Event<void()> OnStudyNodalOptimizationChanged;
+Event<void()> OnStudyAreaUseAdequacyPatchChanged;
```
guilpier-code: The event **OnStudyAreaUseAdequacyPatchChanged** does not seem to be connected to a function / treatment / behavior. So, does it have any effect ?

src/ui/simulator/windows/inspector/accumulator.hxx
```diff
@@ -906,6 +909,28 @@ struct PAreaUnsuppliedEnergyCost
...
 };
 
+struct PAdequacyPatchMode
+{
+    typedef Data::AdequacyPatch::AdequacyPatchMode Type;
+    static Type Value(const Data::Area* area)
+    {
+        return area->adequacyPatchMode;
+    }
+    static wxString ConvertToString(const Type v)
```
guilpier-code, line 633: Why not : 
```C++
static wxString ConvertToString(const Type v)
{
     return adequacyPatchMode[v];
}
```
You defined **adequacyPatchMode** array earlier in this file.
Note that having a value out of the **enum** range is very unlikely to happen, precisely because **enum** types do not accept values out of their range. So instruction : 
```C++
return wxEmptyString
```
is dead code

src/ui/simulator/windows/inspector/property.update.cpp
```diff
@@ -172,6 +172,29 @@ bool InspectorGrid::onPropertyChanging_A(wxPGProperty*,
...
+        {
+            auto* area = *i;
+            if (area)
+            {
+                String s;
+                wxStringToString(value.GetString(), s);
+                s.toLower();
+                s.trim();
+
+                if (s == "virtual area" || s == "0")
```
guilpier-code: - we could avoid **if...else if...else if**, replacing it with a switch statement
- Apparently, **s** cannot contain anything else than 1 character : "0", "1" or "2", I tested it.

So we could have : 
```C++
switch (s.to<int>())
{
    case 0:
        area->adequacyPatchMode = Data::AdequacyPatch::adqmVirtualArea;
        break;
     case 1:
         area->adequacyPatchMode = Data::AdequacyPatch::adqmPhysicalAreaOutsideAdqPatch;
         break;
      case 2:
          area->adequacyPatchMode = Data::AdequacyPatch::adqmPhysicalAreaInsideAdqPatch;
          break;
      default:
          break;
}
``` 
I tested it, it works fine.

src/ui/simulator/windows/options/optimization/optimization.h
```diff
@@ -115,6 +116,9 @@ class Optimization final : public wxDialog
     Component::Button* pBtnSimplexOptimizationRange;
 
     Component::Button* pBtnExportMPS;
+    Component::Button* pBtnAdequacyPatch;
+    Component::Button* pBtnAdequacyPatchNTC12;
```
guilpier-code: "12" and "11" unclear. Prefer **pBtnNTCfromOutToInAdqPatch**

src/ui/simulator/windows/options/optimization/optimization.cpp
```diff
@@ -46,6 +46,21 @@ namespace Window
 {
 namespace Options
 {
+static void SubTitle(wxWindow* parent, wxSizer* sizer, const wxChar* text, bool margintop = true)
```
guilpier-code: - **SubTitle** is not a very expressive name about what the function does. Maybe **addLabelAdqPatch** would be better.
- As this function is used once  with **margintop** as **true**, **margintop** is probably useless. Consider removing this argument. Inside the function, keep what's inside the **if** but remove the **if** test.

guilpier-code:
From what I see, setting the adequacy patch status of an area (virtual area, physical area inside or outside the adequacy patch domain domain) can only be done from the inspector.
Am I right ?
If this is the case, it does seem to be a "compliant" use of GUI, maybe we should consider this.

guilpier-code (changes requested):

src/solver/application.cpp
```diff
@@ -84,6 +84,36 @@ void checkSimplexRangeHydroHeuristic(Antares::Data::SimplexOptimization optRange
...
+    {
+        throw Error::IncompatibleStudyModeForAdqPatch();
+    }
+}
+// When Adequacy Patch is on at least one area must be inside Adequacy patch mode.
+void checkAdqPatchContainsAdqPatchArea(const bool adqPatchOn, const Antares::Data::AreaList& areas)
+{
+    if (adqPatchOn)
+    {
+        bool containsAdqArea = false;
```
guilpier-code: Not sure this variable is useful : the exception could be directly raised from inside the loop over areas

guilpier-code (changes requested):

src/solver/application.cpp
flomnes: See #654

src/libs/antares/study/area/list.cpp
```diff
@@ -326,6 +331,30 @@ bool saveAreaOptimisationIniFile(const Area& area, const Clob& buffer)
...
 
+bool saveAreaAdequacyPatchIniFile(const Area& area, const Clob& buffer)
+{
+    IniFile ini;
+    IniFile::Section* section = ini.addSection("adequacy-patch");
+    int value;
+    switch (area.adequacyPatchMode)
+    {
+    case Data::AdequacyPatch::adqmVirtualArea:
+        value = 0;
```
guilpier-code: I can see that when saving any area's **adequacy-patch.ini**, we have something like : 
```ini
[adequacy-patch]
adequacy-patch-mode = 2
```
The value **2** is not very expressive. We could have **inside adq patch domain** for instance

guilpier-code (changes requested):

src/solver/simulation/sim_calcul_economique.cpp
flomnes: Only available in c++17

src/solver/optimisation/adequacy_patch.cpp
flomnes: I agree with @guilpier-code's remark

src/solver/optimisation/adequacy_patch.cpp
flomnes: done

flomnes (commented):
- [small] 11, 12 : not clear. Make sure to use 11 -> OutsideOutside, 12 -> OutsideInside
- [small] UI : Subtitle function : remove last argument whose value is always `true`
- [small] UI : switch case + no conversion to `YString` to parse enum from menu
- [medium] Interface class for the weekly problem (factory method ?), at least add explanatory comments

src/libs/antares/study/parameters.h
flomnes: Fixed in "last remarks..."

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
flomnes: See issue #716

src/solver/optimisation/adequacy_patch.cpp
flomnes: See #717

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -469,6 +497,13 @@ typedef struct
     double* CoutsMarginauxHorairesDeLaReserveParZone;
 } COUTS_MARGINAUX_ZONES_DE_RESERVE;
 
+struct AdequacyPatchParameters
```
guilpier-code, line 500: ok

guilpier-code (commented):
ok

PR #659, opened 2022-04-26
Test and use JSON for end-to-end tests

flomnes:
### TODO
- [x] `pytest-xdist`
- [ ] Run non-reg tests on release (ubuntu + windows)
- [ ] Run tests on centOS
- [ ] Write a script to download & unzip batches, to run tests locally


flomnes:
### TODO
- [ ] `pytest-xdist`
- [ ] Run non-reg tests on release (ubuntu + windows)
- [ ] Run tests on centOS

src/tests/run-study-tests/check_on_results/tolerances.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def set_relative(self, tol_collection):
+        for key, value in tol_collection.items():
+            self.RTOL[key] = value
+
+    def set_absolute(self, tol_collection = {}):
+        for key, value in tol_collection.items():
+            self.ATOL[key] = value
+
+class Linux_tolerances(Tolerances):
```
guilpier-code, line 50: We probably don't need these tolerances anymore. To be checked.

src/tests/run-study-tests/test_unfeasible_problem.py
```diff
@@ -0,0 +1,68 @@
...
+from check_on_results.unfeasible_problem import unfeasible_problem, weeks_in_year
+from check_on_results.sim_return_code import simulation_success
+
+from fixtures import *
+
+ALL_STUDIES_PATH = Path('../resources/Antares_Simulator_Tests').resolve()
+
+# --------------------------------------------------------------
+# Example of a test with output AND check integrity comparisons
+# --------------------------------------------------------------
```
guilpier-code: Is this comment relevant ?

.github/workflows/windows-release.yml
```diff
@@ -96,10 +97,41 @@ jobs:
...
+      with:
+        path: 'simtest.json'
+        prop_path: 'version'
+
+    - name: Run medium-tests
+      uses: ./.github/workflows/run-tests
+      with:
+        simtest-tag: ${{steps.simtest-version.outputs.prop}}
+        batch-name: medium-tests
+        os: ${{ matrix.testref }}
```
guilpier-code: About **testref**, could we make a bit clearer ?
For instance : **os-name-to-download-tests**

.github/workflows/windows-vcpkg.yml
```diff
@@ -21,6 +21,7 @@ jobs:
           - os: windows-latest
             triplet: x64-windows
             vcpkgPackages: wxwidgets boost-test
+            testref: windows-2022
```
guilpier-code: see remark on **testref** in **.github/workflows/windows-release.yml**

.gitmodules
```diff
@@ -11,6 +11,3 @@
 [submodule "vcpkg"]
 	path = vcpkg
 	url = https://github.com/microsoft/vcpkg.git
-[submodule "src/tests/resources/NonRegTests"]
```
guilpier-code, line 14: ok, I suppose that we get rid of non regression repository as a sub module because we don't play these tests yet / anymore

src/tests/examples/antares_test_utils.py
```diff
@@ -52,18 +52,68 @@ def run_study(solver_path, study_path, use_ortools = False, ortools_solver = "si
     if "Solver returned error" in output[0].decode('utf-8'):
         raise_assertion("Solver returned error")
 
-def enable_study_output(study_path, enable):
-    st = Study(str(study_path))
-    st.check_files_existence()
+# ===============================================================
```
guilpier-code: I guess we still need this source file. Why do we need it for ?

guilpier-code (changes requested):

src/tests/examples/antares_test_utils.py
flomnes: It is no longer needed, I will remove it

src/tests/run-study-tests/test_unfeasible_problem.py
flomnes: agreed, changed this comment to make it appropriate

src/tests/run-study-tests/check_on_results/tolerances.py
flomnes: I disagree. If we didn't provide default values for tolerances, we'd have to specify them in each .json config file, which would be quite redundant.

.github/workflows/windows-release.yml
flomnes: Ok, will use `test-platform`

src/tests/run-study-tests/check_on_results/tolerances.py
guilpier-code: I rather meant that, as each platform has now its own reference results, we should try to make tests pass without any tolerance and see what we get.

guilpier-code (approved):

PR #666, opened 2022-04-29
Remove SUBSTITUTION_DE_LA_VARIABLE_MPLUS

flomnes:


guilpier-code (changes requested):
We have some regressions here, but these 3 commits does not seem to be responsible for that , I tested it.
We'll have to investigate further more.
Between reference and this version results, the difference is 1 for most output variables.
But there are some bigger differences, see for instance : 
**test_069_hydro_reservoir_model_enhanced_02** : 
      mc-all\areas\@ all\values-daily.txt : H. STOR.2
             12917.0  10562.0 -2355.0

PR #667, opened 2022-04-29
Remove GROSSES_VARIABLES

flomnes:


src/solver/optimisation/opt_gestion_des_couts_couts_demarrage.cpp
```diff
@@ -155,36 +155,5 @@ void OPT_InitialiserLesCoutsLineaireCoutsDeDemarrage(PROBLEME_HEBDO* ProblemeHeb
...
-                if (Var >= 0 && Var < ProblemeAResoudre->NombreDeVariables)
-                {
-                    ProblemeAResoudre->CoutLineaire[Var] = 0;
-                }
-            }
-        }
-    }
-#endif
-
     return;
```
guilpier-code: The **return** is not required

guilpier-code (changes requested):

guilpier-code (changes requested):
Same remarks as in #666 about the regressions

PR #668, opened 2022-04-29
Remove PminDuPalierThermiquePendantUneSemaine

flomnes:


guilpier-code (approved):
Still this regression problem, as in #666 and #667

PR #670, opened 2022-04-29
Feature/optimization discrete variables

HugoKulesza:


src/solver/optimisation/opt_construction_variables_couts_demarrages.cpp
```diff
@@ -136,18 +144,21 @@ void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeLineaireCoutsDeDemarra
...
               = NombreDeVariables;
             ProblemeAResoudre->TypeDeVariable[NombreDeVariables] = VARIABLE_NON_BORNEE;
+            ProblemeAResoudre->VariablesEntieres[NombreDeVariables] = false;
             NombreDeVariables++;
 
             CorrespondanceVarNativesVarOptim->NumeroDeGrosseVariableDefaillanceNegative[Pays]
               = NombreDeVariables;
             ProblemeAResoudre->TypeDeVariable[NombreDeVariables] = VARIABLE_NON_BORNEE;
+            ProblemeAResoudre->VariablesEntieres[NombreDeVariables] = false;
             NombreDeVariables++;
```
flomnes: Not needed

src/solver/utils/ortools_utils.cpp
```diff
@@ -145,20 +160,38 @@ static void extract_from_MPSolver(const MPSolver* solver,
     auto& variables = solver->variables();
     int nbVar = problemeSimplexe->NombreDeVariables;
 
+    const bool isMIP = std::any_of(problemeSimplexe->VariablesEntieres.cbegin(),
+                                   problemeSimplexe->VariablesEntieres.cend(),
+                                   [](bool x) { return x; });
+
```
flomnes: Duplicate code : add a const method in `PROBLEME_SIMPLEXE_NOMME`.

src/libs/antares/study/fwd.cpp
```diff
@@ -293,8 +293,10 @@ UnitCommitmentMode StringToUnitCommitmentMode(const AnyString& text)
     s.trim();
     s.toLower();
     if (s == "fast")
-        return ucHeuristic;
-    if (s == "accurate") // mixed integer linear problem
+        return ucFast;
+    if (s == "accurate") 
+        return ucAccurate;
```
flomnes: Use `ucHeuristicFast`, `ucHeuristicAccurate`. When we come back to the code in a few months, we'll understand easily that "Fast" and "Accurate" both use heuristics, whereas MILP doesn't.

flomnes: You may use `sed`, `xarg` and `find` to search & replace across the whole codebase. Something like
```
find -type f \( -name "*.cpp" -or -name "*.h" \) | xargs sed -i "s/search/replace/g"
```

src/solver/application.cpp
```diff
@@ -42,6 +42,26 @@ void printVersion()
...
+    {
+        if (!ortoolsUsed)
+        {
+            throw Error::IncompatibleMILPWithoutOrtools();
+        }
+        else if (ortoolsSolverUsed == Antares::Data::OrtoolsSolver::sirius)
+        {
+            throw Error::IncompatibleMILPOrtoolsSolver();
+        }
+    }
```
```suggestion
                       )
{
    using namespace Antares::Data;
    if (ucMode == UnitCommitmentMode::ucMILP)
    {
        if (!ortoolsUsed)
        {
            throw Error::IncompatibleMILPWithoutOrtools();
        }
        else if (ortoolsSolverUsed == OrtoolsSolver::sirius)
        {
            throw Error::IncompatibleMILPOrtoolsSolver();
        }
    }
```

src/solver/application.cpp
```diff
@@ -55,12 +75,13 @@ void checkSimplexRangeHydroPricing(Antares::Data::SimplexOptimization optRange,
...
 // CHECK incompatible de choix simultané des options « simplex range= daily » et «
-// unit-commitment = MILP ».
+// unit-commitment = Accurate » ou « unit-commitment = MILP ».
 void checkSimplexRangeUnitCommitmentMode(Antares::Data::SimplexOptimization optRange,
                                          Antares::Data::UnitCommitmentMode ucMode)
 {
     if (optRange == Antares::Data::SimplexOptimization::sorDay
-        && ucMode == Antares::Data::UnitCommitmentMode::ucMILP)
+        && (ucMode == Antares::Data::UnitCommitmentMode::ucAccurate
+            || ucMode == Antares::Data::UnitCommitmentMode::ucMILP))
```
```suggestion
        && (ucMode == UnitCommitmentMode::ucAccurate
            || ucMode == UnitCommitmentMode::ucMILP))
```
Add `using namespace Antares::Data;` as the 1st instruction in the function.

src/solver/variable/state.cpp
```diff
@@ -254,263 +255,285 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
...
+                // on amorce une descente : thermalClusterDispatchedUnitsCountForYear[h] 
+                // peut être supérieur à  ON_min[h]
+                
+                portee = 0;
+                nivmin = ON_min[h];
 
-                    (ON_max[h] < ON_opt[h - 1]) ? nivmax = ON_max[h] : nivmax = ON_opt[h - 1];
+                (ON_max[h] < thermalClusterDispatchedUnitsCountForYear[h - 1]) 
+                        ? nivmax = ON_max[h] 
+                        : nivmax = thermalClusterDispatchedUnitsCountForYear[h - 1];
```
flomnes:
```suggestion
                 ON_max[h] = Math::Min<uint>(ON_max[h], thermalClusterDispatchedUnitsCountForYear[h - 1]);
```

src/solver/variable/state.cpp
```diff
@@ -254,263 +255,285 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
...
+            switch (unitCommitmentMode)
             {
-                // When the cluster is in must-run mode, the production value
-                // directly comes from the time-series
-                thermalClusterProduction
-                  = thermalClusterAvailableProduction; // in mustrun, production==available
-                                                       // production
-            }
-            else
+            case Antares::Data::UnitCommitmentMode::ucFast:
```
flomnes: This `switch..case` is used to determine `ON_min[h]`. It would be nice to make it visible here with a function
```cpp
ON_min[h] = minimalClustersON(unitCommitmentMode, currentCluster, /* ... */);
```

Lines 334-370.

src/solver/variable/state.cpp
```diff
@@ -254,263 +255,285 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
...
+            if (currentCluster->minStablePower > 0.)
+            {
+                maxUnitNeeded = static_cast<uint>(
+                Math::Floor(thermalClusterProduction / currentCluster->minStablePower));
+                if (ON_max[h] > maxUnitNeeded)
+                    ON_max[h] = maxUnitNeeded;
             }
+
+            if (ON_max[h] < ON_min[h])
+                ON_max[h] = ON_min[h];
```
flomnes: Same remark as above
```cpp
ON_max[h] = maximalClustersON(currentCluster, /* ... */);
```
If you can think of a better name, please use it.

src/solver/utils/ortools_utils.cpp
```diff
@@ -143,7 +143,10 @@ MPSolver* convert_to_MPSolver(
     // Create the linear solver instance
     MPSolver* solver = new MPSolver("simple_lp_program", solverType);
 
-    tuneSolverSpecificOptions(solver);
+    if (study.parameters.ortoolsParamsString == "")
```
```suggestion
    if (study.parameters.ortoolsParamsString.empty())
```

src/solver/optimisation/opt_alloc_probleme_a_optimiser.cpp
```diff
@@ -162,10 +162,15 @@ static void optimisationAllocateProblem(PROBLEME_HEBDO* ProblemeHebdo, const int
     NbIntervalles
       = (int)(ProblemeHebdo->NombreDePasDeTemps / NombreDePasDeTempsPourUneOptimisation);
 
-    ProblemeAResoudre->ProblemesSpx = (PROBLEMES_SIMPLEXE*)MemAlloc(sizeof(PROBLEMES_SIMPLEXE));
-    ProblemeAResoudre->ProblemesSpx->ProblemeSpx = (void**)MemAlloc(NbIntervalles * sizeof(void*));
+    ProblemeAResoudre->ProblemesSpx1 = (PROBLEMES_SIMPLEXE*)MemAlloc(sizeof(PROBLEMES_SIMPLEXE));
+    ProblemeAResoudre->ProblemesSpx1->ProblemeSpx = (void**)MemAlloc(NbIntervalles * sizeof(void*));
```
flomnes: Simplification
```cpp
...->ProblemeSpx[OptIndex]->ProblemeSpx = ...
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -103,15 +103,21 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, uint numSpace, int NumIn
...
     PremierPassage = OUI_ANTARES;
-    MPSolver* solver;
 
-    ProbSpx = (PROBLEME_SPX*)(ProblemeAResoudre->ProblemesSpx->ProblemeSpx[(int)NumIntervalle]);
-    solver = (MPSolver*)(ProblemeAResoudre->ProblemesSpx->ProblemeSpx[(int)NumIntervalle]);
+    PROBLEMES_SIMPLEXE* ProblemesSpx;
+    if (Probleme.NumeroOptimisation == 1)
+        ProblemesSpx = ProblemeAResoudre->ProblemesSpx1;
+    else if (Probleme.NumeroOptimisation == 2)
+        ProblemesSpx = ProblemeAResoudre->ProblemesSpx2;
```
flomnes: L'utilisation d'un index d'optimisation (0 ou 1) permet d'éliminer le if...else

src/solver/optimisation/opt_nombre_min_groupes_demarres_couts_demarrage.cpp
```diff
@@ -206,23 +206,32 @@ void OPT_AjusterLeNombreMinDeGroupesDemarresCoutsDeDemarrage(PROBLEME_HEBDO* Pro
...
-                          = NombreMinDeGroupesEnMarcheDuPalierThermique[PdtHebdo];
-
-                    if (PminDUnGroupeDuPalierThermique
-                          * NombreMaxDeGroupesEnMarcheDuPalierThermique[PdtHebdo]
-                        > PuissanceDisponibleDuPalierThermique[PdtHebdo])
-                        PuissanceDisponibleDuPalierThermique[PdtHebdo]
-                          = PminDUnGroupeDuPalierThermique
-                            * NombreMaxDeGroupesEnMarcheDuPalierThermique[PdtHebdo];
+                    NombreMaxDeGroupesEnMarcheDuPalierThermique[PdtHebdo]
+                      = NombreMinDeGroupesEnMarcheDuPalierThermique[PdtHebdo];
```
flomnes: Ça me paraît violent comme changement, on en discute

src/solver/utils/ortools_utils.cpp
```diff
@@ -165,21 +195,36 @@ static void extract_from_MPSolver(const MPSolver* solver,
...
         auto& row = constraints[idxRow];
-        problemeSimplexe->CoutsMarginauxDesContraintes[idxRow] = row->dual_value();
+        if (isMIP && problemeSimplexe->NumeroOptimisation == 1)
+        {
+            problemeSimplexe->CoutsMarginauxDesContraintes[idxRow] = 0;
+        }
+        else
+        {
+            problemeSimplexe->CoutsMarginauxDesContraintes[idxRow] = row->dual_value();
+        }
```
flomnes: Si MIP, il faut lire la valeur de la 2nde optimisation, qui est un LP résolu exprès pour avoir les valeurs duales

src/libs/antares/study/parameters.cpp
```diff
@@ -318,6 +318,8 @@ void Parameters::reset()
 
     ortoolsUsed = false;
     ortoolsEnumUsed = OrtoolsSolver::sirius;
+    ortoolsParamsString = "";
```
guilpier-code: `ortoolsParamsString.clear()` would be better

src/libs/antares/study/parameters.h
```diff
@@ -552,8 +552,12 @@ class Parameters final
     //@{
     //! Define if ortools is used
     bool ortoolsUsed;
-    //! Ortool solver used for simulation
+    //! Ortools solver used for simulation
     OrtoolsSolver ortoolsEnumUsed;
+    //! Specific parameters string for ortools solver
+    string ortoolsParamsString;
```
guilpier-code: `string ortoolsParamsString;` : 
What this parameter is for exactly ? It's not obvious when reading its name.
Maybe it means something to someone that knows about or-tools, but not to me. 
We already know it's a string, so we don't need it in the name, and **Params** in the name does not give us extra clue, so it's not needed as well.

src/solver/utils/ortools_utils.cpp
```diff
@@ -126,21 +133,39 @@ MPSolver* convert_to_MPSolver(
     auto& study = *Data::Study::Current::Get();
```
guilpier-code: `MPSolver* convert_to_MPSolver(const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe)` : 
Like many functions/methods in Antares, there is a dependency problem with function : it depends on the whole study, while it should take a few values as argument (for instance **ortoolsEnumUsed**, or the new **ortoolsParamsString**).
Passing these parameters would break the dependency to study, and as a side effect, not being recompiled when we touch the Study

src/libs/antares/study/parameters.h
guilpier-code: `string ortoolsParamsString;` (again) :
If this parameter can have several possible values, making it an **enum** would be better.
**To be noticed** : there is no check of any kind on its value, unless this check is done inside or-tools. Theoretically, this parameter's value could be any string (most checks are done in class **Parameters**).

src/solver/utils/ortools_utils.cpp
```diff
@@ -126,21 +133,39 @@ MPSolver* convert_to_MPSolver(
...
 
     // Define solver used depending on study option
-    MPSolver::OptimizationProblemType solverType
-      = OrtoolsUtils().getLinearOptimProblemType(study.parameters.ortoolsEnumUsed);
+
+    MPSolver::OptimizationProblemType solverType;
```
guilpier-code: In this file, we have 3 **if** statements following the pattern : 
```C++
if (isMIP && problemeSimplexe->NumeroOptimisation == 1)
{
     ....
}
else
{
     ....
}
``` 
This is the sign that we should use polymorphism and encapsulate **MIP** problem and **LP** problem in separate classes.
Let's create an interface (or a base class to factorize the common code of both problems), and have a factory to instantiate the right problem in a base pointer, etc.

If we do that (and if don't want to touch code but in the current source file), we will be lead to  : 
- change the entry point, that is the **ORTOOLS_Simplexe(...)** function's body : call the factory and the resolution there
- extract the **solver type** determination (sirius, glop, xpress, ...) from **convert_to_MPSolver(...)** and put it in the factory
- make several functions of this file methods of LP / MIP classes

src/solver/utils/ortools_utils.cpp
```diff
@@ -16,11 +16,15 @@ static void transferVariables(MPSolver* solver,
                               const double* bMax,
                               const double* costs,
                               int nbVar,
-                              const std::vector<std::string>& NomDesVariables)
-{
+                              const std::vector<std::string>& NomDesVariables,
+                              const std::vector<bool>& VariablesEntieres,
```
guilpier-code: If we refactor as suggested below, **VariablesEntieres** will be used only in the **MIP** problem

src/solver/simulation/solver.hxx
```diff
@@ -852,7 +852,8 @@ void ISimulation<Impl>::estimateMemoryForOptimizationPb(Antares::Data::StudyMemo
     }
 
     // Mode accurate
```
guilpier-code: `// Mode accurate` ==> `// Mode accurate and MILP`

src/solver/application.cpp
```diff
@@ -286,6 +307,9 @@ void Application::prepare(int argc, char* argv[])
 
     // Some more checks require the existence of pParameters, hence of a study.
     // Their execution is delayed up to this point.
+    checkOrtoolsUsage(
```
guilpier-code, line 143: Does the comment : 
```C++    
// Some more checks require the existence of pParameters, hence of a study.
// Their execution is delayed up to this point.
```
regard also the **checkOrtoolsUsage**'s call ?
I ask it because the comment used to regard the following call.
If the comment regards **checkOrtoolsUsage**'s call as well, where are the other checks ? And couldn't we gather all checks in one function, called at the right place ?

src/ui/simulator/windows/options/advanced/advanced.cpp
```diff
@@ -841,41 +841,65 @@ void AdvancedParameters::onUnitCommitmentMode(Component::Button&, wxMenu& menu,
...
+    it = Menu::CreateItem(&menu, wxID_ANY, text, "images/16x16/tag.png");
+    menu.Connect(it->GetId(),
+                 wxEVT_COMMAND_MENU_SELECTED,
+                 wxCommandEventHandler(AdvancedParameters::onSelectUCMILP),
                  nullptr,
                  this);
 }
 
-void AdvancedParameters::onSelectUCHeuristic(wxCommandEvent& /* evt */)
+void AdvancedParameters::onSelectUCHeuristicFast(wxCommandEvent& /* evt */)
```
guilpier-code: the 3 callback functions associated to each unit commitment type are very similar and differ by this type only. We should avoid code duplication by creating a template function.

src/ui/simulator/windows/options/advanced/advanced.cpp
```diff
@@ -841,41 +841,65 @@ void AdvancedParameters::onUnitCommitmentMode(Component::Button&, wxMenu& menu,
     wxMenuItem* it;
     wxString text;
 
-    text = wxStringFromUTF8(UnitCommitmentModeToCString(Data::ucHeuristic)); // Fast
+    text = wxStringFromUTF8(UnitCommitmentModeToCString(Data::ucHeuristicFast)); // Fast
```
guilpier-code, line 809: There is code duplication when creating unit commitment items. Each creation is similar to the other. Could we factorize that code in a single function ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -103,15 +103,21 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, uint numSpace, int NumIn
...
+
```
guilpier-code: So many problems : 
- `PROBLEME_HEBDO* ProblemeHebdo`
- `PROBLEME_SIMPLEXE_NOMME Probleme`
- `PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre;`
- `PROBLEMES_SIMPLEXE* ProblemesSpx;`
-  `PROBLEME_SPX* ProbSpx`
-  `MPSolver* solver`
- `ProblemeAResoudre->ProblemesSpx1;`
- `ProblemeAResoudre->ProblemesSpx2;`

Do we really need them all ? 
It's very likely that there is unnecessary complexity here.
We **really** should do something to simplify this. It's always painful to read this area of code.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -103,15 +103,21 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, uint numSpace, int NumIn
...
+                                                   ProblemeAResoudre->NomDesContraintes,
+                                                   ProblemeAResoudre->VariablesEntieres,
+                                                   ProblemeAResoudre->NumeroDOptimisation);
     PremierPassage = OUI_ANTARES;
-    MPSolver* solver;
 
-    ProbSpx = (PROBLEME_SPX*)(ProblemeAResoudre->ProblemesSpx->ProblemeSpx[(int)NumIntervalle]);
-    solver = (MPSolver*)(ProblemeAResoudre->ProblemesSpx->ProblemeSpx[(int)NumIntervalle]);
+    PROBLEMES_SIMPLEXE* ProblemesSpx;
+    if (Probleme.NumeroOptimisation == 1)
```
guilpier-code: **PREMIERE_OPTIMISATION** and **DEUXIEME_OPTIMISATION** pre-processor defined constants exist. Should not we use them ?

guilpier-code: **Question** : does it mean we allocate 2 problems instead of one ? Is needed RAM growing ?

guilpier-code (changes requested):

src/solver/optimisation/opt_structure_probleme_a_resoudre.h
```diff
@@ -140,6 +141,9 @@ typedef struct
     /* Nommage des variables & contraintes */
     std::vector<std::string> NomDesVariables;
     std::vector<std::string> NomDesContraintes;
+
+    std::vector<bool> VariablesEntieres; // true = int, false = continuous
```
guilpier-code, line 118: There is already a std::vector\<bool\> **VariablesEntieres** as a member of class **PROBLEME_SIMPLEXE_NOMME**.
It looks like code duplication to me.

src/solver/optimisation/opt_alloc_probleme_a_optimiser.cpp
```diff
@@ -178,6 +183,7 @@ static void optimisationAllocateProblem(PROBLEME_HEBDO* ProblemeHebdo, const int
 
     ProblemeAResoudre->NomDesVariables.resize(ProblemeAResoudre->NombreDeVariables);
     ProblemeAResoudre->NomDesContraintes.resize(ProblemeAResoudre->NombreDeContraintes);
+    ProblemeAResoudre->VariablesEntieres.resize(ProblemeAResoudre->NombreDeVariables);
```
guilpier-code: `ProblemeAResoudre->VariablesEntieres.resize(...)` : 
So here we reserve memory even if a **MILP** problem is not required.

src/solver/optimisation/opt_construction_variables_couts_demarrages.cpp
```diff
@@ -79,6 +79,8 @@ void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeLineaireCoutsDeDemarra
                   = NombreDeVariables;
                 ProblemeAResoudre->TypeDeVariable[NombreDeVariables]
                   = VARIABLE_BORNEE_DES_DEUX_COTES;
+                if (ProblemeHebdo->OptimisationAvecVariablesEntieres)
```
guilpier-code: This source file is originally about the kind of bounds we put on variables regarding the number of  working dispatchable units.
What we add here : whether the problem to solve is a **LP** or **MILP**, these variables are flagged as discrete or not.
So, on flagging variables like this, we : 
- mix problem **LP** and problem **MILP**, which contradict separation of concerns
- mix some operations that are not really of the same nature

We could create a new function that only gives a flag (discrete or not) variables related to number of  working dispatchable units, and call it as follows : 
```C++
if (ProblemeHebdo->OptimisationAvecVariablesEntieres)
    OPT_makeNODUvariablesDiscrete(ProblemeHebdo);
```

src/solver/optimisation/opt_nombre_min_groupes_demarres_couts_demarrage.cpp
```diff
@@ -206,23 +206,35 @@ void OPT_AjusterLeNombreMinDeGroupesDemarresCoutsDeDemarrage(PROBLEME_HEBDO* Pro
                     NombreMinDeGroupesEnMarcheDuPalierThermique[PdtHebdo] = (int)ceil(X);
                 }
```
guilpier-code, line 178: We should think of whether we want to keep **ProblemeHebdo->Expansion**, and remove it otherwise.
Here we would remove dozens of lines

src/solver/optimisation/opt_nombre_min_groupes_demarres_couts_demarrage.cpp
```diff
@@ -206,23 +206,35 @@ void OPT_AjusterLeNombreMinDeGroupesDemarresCoutsDeDemarrage(PROBLEME_HEBDO* Pro
...
 
-                OPT_PbLineairePourAjusterLeNombreMinDeGroupesDemarresCoutsDeDemarrage(
-                  ProblemeHebdo, NombreMinDeGroupesEnMarcheDuPalierThermique, Pays, Index);
-
-                for (PdtHebdo = 0; PdtHebdo < NombreDePasDeTempsProblemeHebdo; PdtHebdo++)
+                if (ProblemeHebdo->OptimisationAvecVariablesEntieres == NON_ANTARES)
```
guilpier-code: Same remark as in **src/solver/optimisation/opt_construction_variables_couts_demarrages.cpp**, we should think of separating **LP** and **MILP** regarding fixing Min number of working dispatchable units

guilpier-code (dismissed):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: We should use the constants.

The 2 problems are required for the 2nd resolution, which is LP and allows us to recover dual values/reduced costs for continuous variables.

flomnes: Although with a bit of work, we may re-use the existing problem for the 2 resolutions, changing only variable types

- Use the real variable type (continuous/integer)
- Recover values for integer values
- Fix integer variables to the above
- Solve as an LP problem, profit

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: I agree ! It should be quite easy when/if we switch to OR-Tools only, discarding the "native/sirius" mode.

src/ui/simulator/windows/options/advanced/advanced.cpp
flomnes: Yes, agreed

src/solver/utils/ortools_utils.cpp
```diff
@@ -175,21 +164,36 @@ static void extract_from_MPSolver(const MPSolver* solver,
...
     }
 
     auto& constraints = solver->constraints();
     int nbRow = problemeSimplexe->NombreDeContraintes;
     for (int idxRow = 0; idxRow < nbRow; ++idxRow)
     {
         auto& row = constraints[idxRow];
-        problemeSimplexe->CoutsMarginauxDesContraintes[idxRow] = row->dual_value();
+        if (isMIP && problemeSimplexe->NumeroOptimisation == 1)
+        {
```
flomnes: Use `PREMIERE_OPTIMISATION`, etc.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: Can't do this with OR-Tools, since MPSolver's LP/MIP status is immutable. Unfortunately, we'll need a copy of the MPSolver object.

src/solver/optimisation/opt_nombre_min_groupes_demarres_couts_demarrage.cpp
flomnes: Xpansion sometimes uses it. And it is outside the scope of this PR

src/solver/optimisation/opt_structure_probleme_a_resoudre.h
flomnes: The `VariablesEntieres` defined in `PROBLEME_SIMPLEXE_NOMME` is a const reference to this one.

src/solver/optimisation/opt_alloc_probleme_a_optimiser.cpp
```diff
@@ -81,9 +81,11 @@ void OPT_AllocateFromNumberOfVariableConstraints(PROBLEME_ANTARES_A_RESOUDRE* Pr
...
     ProblemeAResoudre->Pi.assign(nbVariables, 0.);
     ProblemeAResoudre->Colonne.assign(nbVariables, 0);
-
-    ProblemeAResoudre->NomDesVariables.resize(nbVariables);
-    ProblemeAResoudre->NomDesContraintes.resize(nbConstraints);
+    // Names
+    ProblemeAResoudre->NomDesVariables.resize(ProblemeAResoudre->NombreDeVariables);
+    ProblemeAResoudre->NomDesContraintes.resize(ProblemeAResoudre->NombreDeContraintes);
+    // Integer variables ? (MILP)
+    ProblemeAResoudre->VariablesEntieres.resize(ProblemeAResoudre->NombreDeVariables);
```
flomnes: Partially revert to use `nbVariables` for consistency

src/libs/antares/study/parameters.h
```diff
@@ -433,11 +432,13 @@ class Parameters final
         PowerFluctuations fluctuations;
     } power;
 
-    struct
+    struct UCMode
     {
         //! Unit Commitment Mode
         UnitCommitmentMode ucMode;
-    } unitCommitment;
+        void addExcludedVariables(std::vector<std::string>&) const;
```
sylvlecl, line 441: Not obvious: we could add a comment why this is needed

src/libs/antares/study/fwd.h
```diff
@@ -464,8 +464,9 @@ SheddingPolicy StringToSheddingPolicy(const AnyString& text);
 
 enum UnitCommitmentMode
 {
-    ucHeuristic = 0,
-    ucMILP, // mixed integer linear problem
+    ucHeuristicFast = 0,
```
sylvlecl, line 469: Would be nice to have some explanatory comments on that enum values too, I think.
The old `ucMILP` is the new `ucHeuristicAccurate`, that's right ?

src/solver/application.cpp
```diff
@@ -31,6 +31,10 @@ using namespace Antares::Check;
 
 namespace
 {
+// CHECK nécessaire d'un choix simultané entre " unit-commitment = MILP " et " ortools-used "
```
sylvlecl: I think you have handled that TODO

sylvlecl (commented):
Nice:
it would be great to have a small in memory end-to-end test as in `simple-study.cpp`, I think it's possible with one of the solvers available with ortools ?

src/libs/antares/study/parameters.h
flomnes: done

src/libs/antares/study/fwd.h
flomnes: yes, added some comments

The old ucMILP wasn't MILP

src/solver/application.cpp
flomnes: It is now handled, I will now remove this TODO note.
```
[2023-10-12 10:05:03][solver][fatal] Unit Commitment mode 'milp' must be used with an OR-Tools solver 
```

src/libs/antares/study/parameters.h
flomnes: Removed

src/solver/utils/ortools_utils.cpp
flomnes: Removed, the `Study` singleton has been removed.

src/solver/utils/ortools_utils.cpp
flomnes: This bit of code has been removed

sylvlecl (approved):

PR #680, opened 2022-05-11
Fix logging of performed years

### LINKED ISSUE #678: Parallel batches of years while running in sequential mode
**Description**
While running a study in sequential mode, some MC year are run in parallel.

**How to Reproduce**
Run the study attached and see the simulation.log file

**Expected behavior**
All years should be run alone, sequentially.

[NTC-2-areas.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/8667976/NTC-2-areas.zip)

**OS and version**
 - OS: Windows, Ubuntu
 - Antares Simulator version: 8.2.1

**Additional context**
None

guilpier-code:
Logging years can be misleading under some circumstances : when there are skipped years in the playlist.
Some skipped years can be embedded among years being performed in the progression bar and in the log file.
In case of sequential execution, it lets user think that several years are run in parallel, although only one is actually performed and the others are actually skipped.
This PR intends to make logging more clear on that matter.
May fix #678

### DIFF
--- a/src/solver/simulation/solver.hxx
+++ b/src/solver/simulation/solver.hxx
@@ -61,7 +61,6 @@ class yearJob final : public Yuni::Job::IJob
 public:
     yearJob(ISimulation<Impl>* pSimulationObj,
             unsigned int pY,
-            std::vector<unsigned int>& pYearsIndices,
             std::map<uint, bool>& pYearFailed,
             std::map<uint, bool>& pIsFirstPerformedYearOfASet,
             bool pFirstSetParallelWithAPerformedYearWasRun,
@@ -73,7 +72,6 @@ public:
             bool pYearByYear) :
      simulationObj(pSimulationObj),
      y(pY),
-     yearsIndices(pYearsIndices),
      yearFailed(pYearFailed),
      isFirstPerformedYearOfASet(pIsFirstPerformedYearOfASet),
      firstSetParallelWithAPerformedYearWasRun(pFirstSetParallelWithAPerformedYearWasRun),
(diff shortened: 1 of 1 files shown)


src/solver/simulation/solver.hxx
```diff
@@ -1547,13 +1549,12 @@ void ISimulation<Impl>::loopThroughYears(uint firstYear,
...
 
-            // If the year has not to be rerun, we skip the computation of the year Note that, when
-            // we enter for the first time in the "for" loop, all years of the set have to be rerun
+            // If the year has not to be rerun, we skip the computation of the year. 
+            // Note that, when we enter for the first time in the "for" loop, all years of the set have to be rerun
             // (meaning : they must be run once). if(!set_it->yearFailed[y]) continue;
 
             qs.add(new yearJob<ImplementationType>(this,
                                                    y,
-                                                   yearsIndicesCopy,
```
flomnes, line 1556: `yearIndicesCopy` is no longer required, do not declare.

flomnes (changes requested):
Remove `yearsIndicesCopy`, it is no longer used.

flomnes:
@guilpier-code from my understanding, this PR only affects logging. However, #678 seems to be related to batches themselves.

Is #678 related caused by logging, or to the construction of MC years batches ?

src/solver/simulation/solver.hxx
```diff
@@ -1134,7 +1121,10 @@ uint ISimulation<Impl>::buildSetsOfParallelYears(
             }
         }
         else
+        {
             set->isYearPerformed[y] = false;
+            set->isFirstPerformedYearOfASet[y] = false;
```
flomnes: Remove this useless instruction

flomnes (changes requested):

src/solver/simulation/solver.hxx
guilpier-code: Done

src/solver/simulation/solver.hxx
guilpier-code: Done

flomnes (approved):

PR #684, opened 2022-05-16
Use antares-deps 2.0.2

### LINKED ISSUE #37: Many shared library files are generated, only provide one ?
OS: Ubuntu
```
install/lib$ ls libortools.so*
libortools.so  libortools.so.9  libortools.so.9.2.9999
```

Keep only `libortools.so.9` ? This creates some confusion since only `libortools.so.9.2.9999` one is mentioned in **ortoolsTargets-release.cmake**, as the location.

```cmake
set_property(TARGET ortools::ortools APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
set_target_properties(ortools::ortools PROPERTIES
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libortools.so.9.2.9999"
  IMPORTED_SONAME_RELEASE "libortools.so.9"
  )
```

flomnes:
### TODO
- [x] Make it possible to compile in debug (see below). Note that RTE does not provide debug install packages for OR-Tools.
- [x] Make it possible to compile OR-Tools from sources, automatically
- [x] ~~Fix https://github.com/rte-france/or-tools/issues/37 in OR-Tools~~
- [x] Used static libortools
- [x] Update doc

We would need to have or-tools as a submodule, to be checked out at configure time.


ortools.json
```diff
@@ -0,0 +1,3 @@
+{
+    "ortools": "v9.2-rte1.2"
```
JasonMarechal25: Shouldn't it be v9.2-rte1.3 ? There is a mismatch with .yml files
Is this file necessary ?

JasonMarechal25 (changes requested):

ortools.json
flomnes: I added file `ortools_tag` to the repository. I could have added a JSON file, but I need CMake to parse it. Parsing JSON is natively supported since Cmake 3.19, which is not easily available in Ubuntu 20.04.

JasonMarechal25 (approved):

PR #702, opened 2022-06-03
Fix segfault occurring when calling antares-solver with --version

flomnes:


src/solver/application.cpp
```diff
@@ -493,8 +493,10 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
 
 void Application::saveElapsedTime()
 {
-    pTotalTimer.stop();
+    if (!pStudy)
```
guilpier-code, line 495: ok. It seems that the value of data member **shouldExecute** is redundant with the value of **pStudy**. We should try to find a way to remove it.

guilpier-code (changes requested):

src/solver/application.cpp
flomnes: Done ! It works

guilpier-code (approved):

PR #705, opened 2022-06-09
Fix version check of Boost in CMake files

flomnes:


guilpier-code (approved):

guilpier-code:
Works on my machine. I approve.

PR #711, opened 2022-06-10
Statistics : add info related to the study

### LINKED ISSUE #708: Add study-related info to statistics
- Number of areas
- Number of links
- Number of enabled thermal clusters
- Number of enabled binding constraints
- etc.

flomnes:
TODO
- [x] Fill the `StudyInfo` structure from the study using the `initializeFromStudy` method, probably just after the study has been loaded
- [x] Write the data into a separate file, named `study-info.txt` in the output directory


src/libs/antares/benchmarking/studyinfo.cpp
```diff
@@ -1,12 +1,53 @@
...
+		auto end = study_.areas.end();
+		for (auto i = study_.areas.begin(); i != end; ++i)
+		{
+			Area& area = *(i->second);
+			auto end = area.thermal.list.end();
+			for (auto i = area.thermal.list.begin(); i != end; ++i)
+			{
+				ThermalCluster* cluster = i->second.get();
+				if (cluster->enabled)
+					nbEnabledThermalClusters_++;
```
flomnes: No need to get the raw pointer to de-reference, `std::unique_ptr<T>` defines a structure dereference operator (`->`)
```suggestion
				if (i->second->enabled)
					nbEnabledThermalClusters_++;
```

src/libs/antares/benchmarking/studyinfo.h
```diff
@@ -1,33 +1,40 @@
...
-    unsigned int UnitComitmentMode;
+public:   
+    StudyInfo(const Antares::Data::Study& study);
+    void collect();
+    void flush(Yuni::String& filePath);
+
+private:
+    // Methods
+    void getNbPerformedYears();
+    void getEnabledThermalClusters();
```
flomnes: I suggest passing `study` to each method, not to the class constructor.
```suggestion
    StudyInfo();
    void collect(const Antares::Data::Study& study);
    void flush(Yuni::String& filePath);

private:
    // Methods
    void getNbPerformedYears(const Antares::Data::Study& study);
    void getEnabledThermalClusters(const Antares::Data::Study& study);
```

src/solver/application.cpp
```diff
@@ -234,6 +235,16 @@ void Application::prepare(int argc, char* argv[])
     // Loading the study
     readDataForTheStudy(options);
 
+    // Collecting info from the study for a further record
+    Benchmarking::StudyInfo study_info(*pStudy);
+    study_info.collect();
+
+    // Flush previous info into a record file
+    Yuni::String filePath; 
+    filePath.clear() << pStudy->folderOutput << Yuni::IO::Separator << "study-info.txt";;
```
flomnes: You may use pStudy->buffer

See **study.h**
```cpp
class Study final : public Yuni::NonCopyable<Study>, public IObject
{
// ...
mutable YString buffer;
// ...
};
```

src/solver/application.cpp
```diff
@@ -234,6 +235,16 @@ void Application::prepare(int argc, char* argv[])
...
+    study_info.flush(filePath);
```
flomnes: I suggest splitting responsabilities :

- 1 class stores data
- Multiple classes can write data in CSV, human-readable format, JSON, etc.
- Possibly one interface class for those writers

src/libs/antares/benchmarking/studyinfo.cpp
```diff
@@ -1,12 +1,53 @@
...
-  }
+	void StudyInfo::flush(Yuni::String& filePath)
+	{
+		if (!outputFile_.openRW(filePath))
+		{
+			throw Antares::Error::CreatingStudyInfoFile(filePath);
+		}
+
+		outputFile_ << "Number of performed years" << "\t" << nbPerformedYears_ << "\n";
+		outputFile_ << "Number of enabled thermal clusters" << "\t" << nbEnabledThermalClusters_ << "\n";
```
flomnes: Since this data is meant to be parsed for statistical analysis, I suggest making it less human-readable, and more easily parsable. Avoid spaces, use short names, etc.

src/libs/antares/benchmarking/studyinfo.h
```diff
@@ -1,33 +1,40 @@
...
+    // Member data
+    const Antares::Data::Study& study_;
+    Yuni::IO::File::Stream outputFile_;
+
+    unsigned int nbPerformedYears_ = 0;
+    unsigned int nbEnabledThermalClusters_ = 0;
+    unsigned int nbEnabledBindingConstraints_ = 0;
+    unsigned int nbAreas_ = 0;
+    unsigned int nbLinks_ = 0;
+    unsigned int UnitComitmentMode_;
```
flomnes:
```suggestion
    unsigned int UnitComitmentMode_ = 0;
```

src/libs/antares/benchmarking/studyinfo.h
```diff
@@ -1,33 +1,40 @@
...
 
     // Parallel execution
-    unsigned int maxNbYearsInParallel;
-    unsigned int minNbYearsInParallel;
-    unsigned int nbCoreMode;
+    unsigned int maxNbYearsInParallel_ = 0;
+    unsigned int minNbYearsInParallel_ = 0;
+    unsigned int nbCoreMode_;
```
flomnes:
```suggestion
    unsigned int nbCoreMode_ = 0;
```

src/libs/antares/benchmarking/studyinfo.cpp
guilpier-code: ok, but we could leave the code as is for more clarity : 
```C++
if (cluster->enabled)
```
Is quite clear I guess

src/libs/antares/benchmarking/studyinfo.cpp
guilpier-code: I agree. Could you be more specific about this file format ?

flomnes:
```
areas <number of areas>
years <number of years>
thermal_clusters <number of thermal clusters>
etc.
```

flomnes (commented):
- Since we don't manipulate numerical data, you could just convert all data to std::string before adding it to the container (no need for a mixed-type container, nice interface though !)
- The data related to the optimization problem is missing (number of variables, constraints, non-zero terms in the constraint matrix, etc.), I'll try to add it.

src/libs/antares/benchmarking/studyinfo.h
```diff
@@ -0,0 +1,130 @@
...
+    void collectEnabledBindingConstraintsCount();
+    void collectUnitCommitmentMode();
+    void collectMaxNbYearsInParallel();
+    void collectOptimizationInfo();
+    void collectSolverVersion();
+
+    template<class ClassItem, typename T>
+    void addItem(std::string name, T value)
+    {
+        items_.push_back(std::unique_ptr<ClassItem>(new ClassItem(name, value)));
```
flomnes:
```suggestion
        items_.emplace_back(new ClassItem(name, value));
```

src/libs/antares/benchmarking/studyinfo.cpp
```diff
@@ -0,0 +1,161 @@
...
+		auto end = study_.areas.end();
+		for (auto i = study_.areas.begin(); i != end; ++i)
+		{
+			Area& area = *(i->second);
+			auto end = area.thermal.list.end();
+			for (auto i = area.thermal.list.begin(); i != end; ++i)
+			{
+				ThermalCluster* cluster = i->second.get();
+				if (cluster->enabled)
+					nbEnabledThermalClusters++;
```
flomnes:
```suggestion
				auto& cluster = i->second;
				if (cluster->enabled)
					nbEnabledThermalClusters++;
```

PR #712, opened 2022-06-10
Add interface class + implementations for economy weekly problem

flomnes:


guilpier-code (approved):

guilpier-code:
ok, that was exactly the idea

PR #554, opened 2022-02-17
Fix GCC warnings

flomnes:

PR #721, opened 2022-06-15
UI - fix a bug on map layers

### LINKED ISSUE #720: Error with the hydro upgraded release (based on 8.2.1)
**Description**
The simulation runs correctly and produces the expected results but does not give you your hand at the end : you need to interrupt Antares

**How to Reproduce**
Execute the simulation of the Use case attached below.

**Expected behavior**
You should have on the screen the final pop-up and your hand again.

None

**OS and version**
 - OS: Windows
 - Antares Simulator version: https://github.com/AntaresSimulatorTeam/Antares_Simulator/suites/6762934244/artifacts/258672145

**Additional context**
None
[079 Zero  Power Balance - Type 1-V8.2.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/8892176/079.Zero.Power.Balance.-.Type.1-V8.2.zip)

guilpier-code:
Before this fix :
- impossible to create a layer correctly :
    + either the new layer has a wrong name (for instance "Map 6" as though there only All and "Map 1" already exist)
    + either it's not possible to set areas as visible or not in the layer
- After a simulation is run, Antares GUI never gives back the control to user : GUI must be killed.

This PR fixes #721

### DIFF
--- a/src/ui/simulator/toolbox/components/map/control.cpp
+++ b/src/ui/simulator/toolbox/components/map/control.cpp
@@ -107,7 +107,7 @@ Control::Control(wxWindow* parent, Component& component, size_t uID) :
  pCurrentMousePositionGraph(INT_MAX, INT_MAX),
  pCurrentClientSize(0, 0),
  pLastMousePosition(INT_MAX, INT_MAX),
- uid(newUID++)
+ uid(uID)
 {
     while (newUID <= uID)
         newUID++;


flomnes:
This was merged in #554, which was never shipped into an official release AFAIK. There is another constructor for this class, it seems to be correct.

flomnes (approved):

JMJ-rte:
Is this bug linked to the one we have with the release that includes Michel hydro development : for this release some simulations does not end and after Guillaume first analysis it seems that it is link to layer management

flomnes:
> Is this bug linked to the one we have with the release that includes Michel hydro development : for this release some simulations does not end and after Guillaume first analysis it seems that it is link to layer management

No, it's completely unrelated in my opinion.

PR #725, opened 2022-06-21
Clarify label for output column 'Profit by plant'

flomnes:
Files details-xxx.txt
### Current
```
<name of the cluster>
Euro
1.21
2.22
1.22
...
```

### Proposed
```
<name of the cluster>
Profit - Euro
1.21
2.22
1.22
...
```

### DIFF
--- a/src/solver/variable/economy/profitByPlant.h
+++ b/src/solver/variable/economy/profitByPlant.h
@@ -47,7 +47,7 @@ namespace Economy
         //! Unit
         static const char* Unit()
         {
-            return "Euro";
+            return "Profit - Euro";
         }
 
         //! The short description of the variable


src/solver/variable/economy/profitByPlant.h
```diff
@@ -47,7 +47,7 @@ namespace Economy
         //! Unit
         static const char* Unit()
         {
-            return "Euro";
```
guilpier-code, line 50: for me, the unit is "Euro".
I know that for some other similar output variables (like **NonProportionalCostByDispatchablePlant**), the unit is "NP Cost - Euro", but it does not seem right to me.

guilpier-code (changes requested):

src/solver/variable/economy/profitByPlant.h
guilpier-code: After discussing it, I'm ok with that.

guilpier-code (approved):

PR #735, opened 2022-06-28
Profit by plant : only compute average

flomnes:
+ format code

### DIFF
--- a/src/solver/variable/economy/profitByPlant.h
+++ b/src/solver/variable/economy/profitByPlant.h
@@ -37,345 +37,342 @@ namespace Variable
 {
 namespace Economy
 {
-    struct VCardProfitByPlant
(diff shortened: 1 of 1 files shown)


src/solver/variable/economy/profitByPlant.h
```diff
@@ -37,345 +37,342 @@ namespace Variable
...
+    {
+        return "Profit for thermal units";
+    }
 
-        //! The VCard to look for calculating spatial aggregates
-        typedef VCardProfitByPlant VCardForSpatialAggregate;
+    //! The expected results
+    typedef Results<R::AllYears::Average< // The average values throughout all years
+      >>
+      ResultsType;
```
flomnes, line 62: Only non-formatting change is here

PR #740, opened 2022-06-30
Refactoring info collect and saving on files

### LINKED ISSUE #708: Add study-related info to statistics

guilpier-code:
See comments below


guilpier-code:
This is a try to remove the duplicate code between : 
- The way we collect and print data from the study and the simulation in **study-info.txt** in one hand
- The way we collect and print durations form the simulation in **time_measurement.txt** in one the other hand

Until now we roughly used, for both files, the following (collaborating) classes : 
- The sources of data : the study and the simulation
- The file content
- The file writer

In one case (**study-info.txt**), the file content class was polluted with some data collecting, so this class makes too many things.
In the other case (**time_measurement.txt**), the collecting of durations (or times if you like) consists in passing the file content to the simulation and  let a **Timer** class fill the file content on the fly as the simulation runs. I guess a file content should not be passed to the simulation.

So the main idea here is to : 
- extract some information collector classes, whose mission is to collect data (as their name suggests)
- make one file content and one file writer classes used for both printed files.

The way we collect information is different depending on the file we print.
That's why you'll find here 3 data collectors : 
- one to collect data from study (nb of areas, nb of links, ...)
- one to collect data from simulation (nb of variables, constraints, ...)
- one to collect durations from simulation.

The 2 last collectors are passed to the simulation and the last does its job with the help of a very simple timer.

For now, 2 ways of collecting **durations** are **coexisting** (so don't be surprised) : 
- the way we did before this PR (for **time_measurement.txt**)
- a way with the help of a duration collector (so we print another time measurement file : **time_measurement_draft.txt**)

The second is meant to replace the first, or much rather is a start for a discussion.
This way of collecting durations is **not complete**, but gives a good idea of what is looked for.
So, in the code, you'll find many comments with the **draft** keyword as a kind of tag.

They are still "**// gp** :" comments in the code to point out things that remain to be done.  

The little classes representing file content lines and inheriting the same interface were removed. They were more complicated and unclear than they should. So the **FileContent** class was changed as well. 

Reminder : 
- Getting the duration collector retrieve all data it's supposed
- duration collector retrieves data inside each MC year. In case of a study with several years, make sure that the number of calls in **time_measurement.txt** is not **1** but more.
- Check if there is a concurrence issue in filling the duration collector inside years played in parallel
- Take tags "**gp :**" comments into account 
-

src/libs/antares/benchmarking/info_collectors.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include <vector>
+#include <map>
+#include <string>
+
+#include <antares/study.h>
+#include "file_content.h"
+
+using namespace std;
```
flomnes: We don't want to have this `using namespace` whenever we include the header.

flomnes (commented):
1.  IMO, in the file contents, we should keep at least a key and value.

```cpp
struct FileContent {
struct KeyValue {
  const std::string key;
  const std::string value;
  std::write(std::ostream& os, const std::string& sep) {
    os << key << sep << value << '\n';
}
};
std::vector<KeyValue> lines_;
};
```
So that different formats can be used (CSV, JSON, etc.)

2. Remove existing data structures, only pass the new one to the Simulation object.
3. The number of calls should be in a different key-value pair
```txt
mc_years_time = 221
mc_years_nb_calls = 5
```

4. Add sections ?

```cpp
std::map<
std::string, // section name
std::vector<KeyValue> // key-value pairs
>
```

src/libs/antares/benchmarking/info_collectors.cpp
```diff
@@ -0,0 +1,158 @@
...
+			int64_t duration_sum = 0;
+			for (auto& duration : durations)
+			{
+				duration_sum += duration;
+			}
+
+			file_content.addDurationLine(name, (unsigned int)duration_sum, (int)durations.size());
+		}
+
+		file_content.addNameValueLine("total", (unsigned int)total_duration_);
```
flomnes: I'm not sure we want to sum all durations in this way. There may be gaps in our time measurements, so this time measure will be inaccurate.

src/libs/antares/benchmarking/info_collectors.h
guilpier-code: Done

src/libs/antares/benchmarking/info_collectors.cpp
guilpier-code: Done

src/solver/main.cpp
```diff
@@ -124,8 +124,7 @@ int main(int argc, char** argv)
 
     application.execute();
 
-    application.writeElapsedTime();
-    application.writeStudyInfos();
+    application.writeInfoOnExecution();
```
flomnes: Nice !

src/libs/antares/benchmarking/info_collectors.h
```diff
@@ -0,0 +1,100 @@
...
+public:
+    virtual void addDuration(std::string name, int64_t duration) = 0;
+};
+
+class NullDurationCollector : public IDurationCollector
+{
+public:
+    NullDurationCollector() = default;
+    void addDuration(std::string name, int64_t duration) override { /* Do nothing */ }
+};
```
flomnes: I suggest moving this definition to **tests/end-to-end/simple-study.cpp** where it is used. 

Also, comment out `name` and `duration` to avoid "warning : unused parameter".

src/libs/antares/benchmarking/info_collectors.h
```diff
@@ -0,0 +1,100 @@
...
+private:
+    const OptimizationInfo& opt_info_;
+};
+
+
+
+class IDurationCollector
+{
+public:
+    virtual void addDuration(std::string name, int64_t duration) = 0;
```
flomnes:
```suggestion
    virtual void addDuration(const std::string& name, int64_t duration) = 0;
```

src/libs/antares/benchmarking/file_content.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+	void FileContent::addItemToSection(string section, string key, string value)
+	{
+		sections_[section][key] = value;
+	}
+
+	void FileContent::addDurationItem(std::string name, unsigned int duration, int nbCalls)
+	{
+		sections_["Durations"][name + "--duration"] = to_string(duration);
+		sections_["Durations"][name + "--nb_calls"] = to_string(nbCalls);
```
flomnes:
```suggestion
		addItemToSection("Durations", name + "_duration", duration);
		addItemToSection("Durations", name + "_nb_calls", nbCalls);
```

src/libs/antares/benchmarking/file_content.cpp
```diff
@@ -0,0 +1,32 @@
...
+	{
+		sections_[section][key] = to_string(value);
+	}
+
+	void FileContent::addItemToSection(string section, string key, string value)
+	{
+		sections_[section][key] = value;
+	}
+
+	void FileContent::addDurationItem(std::string name, unsigned int duration, int nbCalls)
```
flomnes: 1. Pass strings by const-ref
2. Use `string` version in the `int` implementation.

```suggestion
	void FileContent::addItemToSection(const std::string& section, const std::string& key, int value)
	{
		this->addItemToSection(section, key, std::to_string(value));
	}

	void FileContent::addItemToSection(const std::string& section, const std::string& key, const std::string& value)
	{
		sections_[section][key] = value;
	}

	void FileContent::addDurationItem(const std::string& name, unsigned int duration, int nbCalls)
```

src/libs/antares/benchmarking/file_content.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <string>
+#include <map>
+
+#include <memory>
```
flomnes:
```suggestion
```

src/libs/antares/benchmarking/file_content.h
```diff
@@ -0,0 +1,31 @@
...
+    iterator firstSection();
+    iterator endSections();
+
+    void addItemToSection(std::string section, std::string key, int value);
+    void addItemToSection(std::string section, std::string key, std::string value);
+    void addDurationItem(std::string name, unsigned int duration, int nbCalls);
+
+private:
+    // Data of the file content
+    std::map<std::string, std::map<std::string, std::string>> sections_;
```
flomnes:
```suggestion
    std::map<std::string, // Section name
    std::map<std::string, std::string> // Key-values
    > sections_;
```

src/libs/antares/benchmarking/file_content.h
```diff
@@ -0,0 +1,31 @@
...
+{
+public:
+    FileContent() = default;
+
+    using iterator = std::map<std::string, std::map<std::string, std::string>>::iterator;
+    iterator firstSection();
+    iterator endSections();
+
+    void addItemToSection(std::string section, std::string key, int value);
+    void addItemToSection(std::string section, std::string key, std::string value);
```
flomnes: Pass `std::strings` by const ref

src/libs/antares/benchmarking/file_writer.cpp
```diff
@@ -0,0 +1,32 @@
...
+	{
+		Antares::IniFile ini;
+
+		FileContent::iterator it_section = fileContent_.firstSection();
+		for (; it_section != fileContent_.endSections(); it_section++)
+		{
+			Antares::IniFile::Section* section = ini.addSection(it_section->first);
+			for (pair<string, string> line : it_section->second)
+				section->add(line.first, line.second);
+		}
```
flomnes: Use `auto` whenever possible ?

src/solver/application.cpp
```diff
@@ -525,32 +525,30 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
...
 
     // Flush previous info into a record file
-    Yuni::String filePath; 
-    filePath.clear() << pStudy->folderOutput << Yuni::IO::Separator << "study-info.txt";
-    Benchmarking::StudyInfoCSVwriter study_info_writer(filePath, study_info);
-    study_info_writer.flush();
+    Yuni::String filePath;
+    filePath.clear() << pStudy->folderOutput << Yuni::IO::Separator << "execution_info.ini";
+    Benchmarking::iniFilewriter ini_file_writer(filePath, file_content);
+    ini_file_writer.flush();
```
flomnes, line 551: The interface class `FileWriter` isn't very useful for now since `iniFileWriter` is its only child, and there is no need for dynamic polymorphism. Remove it ?

src/solver/application.h
```diff
@@ -50,8 +50,7 @@ class Application : public Yuni::IEventObserver<Application, Yuni::Policy::Singl
     */
     void resetProcessPriority() const;
 
-    void writeElapsedTime();
-    void writeStudyInfos();
+    void writeInfoOnExecution();
```
flomnes: writeExectutionInfo

src/libs/antares/benchmarking/info_collectors.h
```diff
@@ -0,0 +1,100 @@
...
+    void enabledBindingConstraintsCountToFileContent(FileContent& file_content);
+    void unitCommitmentModeToFileContent(FileContent& file_content);
+    void maxNbYearsInParallelToFileContent(FileContent& file_content);
+    void solverVersionToFileContent(FileContent& file_content);
+
+    // Member data
+    const Antares::Data::Study& study_;
+
+    // ----------------------------------------------------------------------------------------
+    // TODO : add some more pieces of information about study : see following data members.
```
flomnes: Do we want to add more info ? Or is this an old TODO note ?

src/libs/antares/benchmarking/info_collectors.cpp
```diff
@@ -0,0 +1,152 @@
...
+	{		
+		for (pair<string, vector<int64_t>> element : duration_items_)
+		{
+			std::string name = element.first;
+			vector<int64_t> durations = element.second;
+			int64_t duration_sum = 0;
+			for (auto& duration : durations)
+			{
+				duration_sum += duration;
+			}
```
flomnes: Use `auto&` to avoid unnecessary copies. Also, you could keep only the sum of durations, removing the need for loop at lines 137-140.

src/libs/antares/benchmarking/info_collectors.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+			file_content.addDurationItem(name, (unsigned int)duration_sum, (int)durations.size());
+		}
+	}
+
+	void DurationCollector::addDuration(std::string name, int64_t duration)
+	{	
+		const std::lock_guard<std::mutex> lock(mutex_);
+		duration_items_[name].push_back(duration);
```
flomnes, line 145:
```suggestion
		duration_items_[name] += duration;
```

src/libs/antares/benchmarking/info_collectors.h
```diff
@@ -0,0 +1,100 @@
...
+class DurationCollector : public IDurationCollector
+{
+public:
+    DurationCollector() = default;
+
+    void toFileContent(FileContent& file_content);
+    void addDuration(std::string name, int64_t duration) override;
+
+private:
+    map<string, vector<int64_t>> duration_items_;
```
flomnes, line 85:
```suggestion
    map<string, int64_t> duration_items_;
```
There is no need to keep all durations, only their sum.

src/libs/antares/benchmarking/info_collectors.cpp
```diff
@@ -0,0 +1,152 @@
...
+		file_content.addItemToSection("from study", "antares version", version);
+	}
+
+	// Collecting data from simulation
+	// ---------------------------------
+	void SimulationInfoCollector::toFileContent(FileContent& file_content)
+	{
+		file_content.addItemToSection("from simulation", "variables", opt_info_.nbVariables);
+		file_content.addItemToSection("from simulation", "constraints", opt_info_.nbConstraints);
+		file_content.addItemToSection("from simulation", "non-zero coefficients", opt_info_.nbNonZeroCoeffs);
```
flomnes: Section "optimization problem"

src/libs/antares/benchmarking/info_collectors.cpp
```diff
@@ -0,0 +1,152 @@
...
+		enabledThermalClustersCountToFileContent(file_content);
+		enabledBindingConstraintsCountToFileContent(file_content);
+		unitCommitmentModeToFileContent(file_content);
+		maxNbYearsInParallelToFileContent(file_content);
+        solverVersionToFileContent(file_content);
+	}
+
+	void StudyInfoCollector::areasCountToFileContent(FileContent& file_content)
+	{
+		file_content.addItemToSection("from study", "areas", study_.areas.size());
```
flomnes: Section "study"

src/solver/application.cpp
```diff
@@ -421,14 +419,16 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
...
 
     if (study.loadFromFolder(pSettings.studyFolder, options) && !study.gotFatalError)
     {
         logs.info() << "The study is loaded.";
         logs.info() << LOG_UI_DISPLAY_MESSAGES_OFF;
     }
-    loadTimer.stop();
+
+    timer.stop();
+    pDurationCollector.addDuration("study_loading", timer.get_duration());
```
flomnes, line 431: [Style] You could pass `timer` as the 2nd argument, and call `get_duration` inside `addDuration`

flomnes (commented):
I wonder if in some cases the simulation may get slowed down by the `mutex`, when executing years in parallel. I'm talking about one thread waiting for the other, etc.

Also, we need to make sure that there is one single instance of `mutex`, not one per thread. Otherwise the lock mechanism won't lock.

src/libs/antares/benchmarking/info_collectors.h
guilpier-code: 1. moving **NullDurationCollector** to end-to-end tests : if we do that, we separate **NullDurationCollector** from its class hierarchy and we can't easily reuse **NullDurationCollector** somewhere else. To be discussed.
2. Done

flomnes: Do we really want to re-use this class elsewhere ?

src/libs/antares/benchmarking/file_content.cpp
guilpier-code: Done

src/libs/antares/benchmarking/info_collectors.h
guilpier-code: Done

src/libs/antares/benchmarking/file_content.cpp
guilpier-code: Done

src/libs/antares/benchmarking/file_content.h
guilpier-code: Done

src/libs/antares/benchmarking/file_content.h
guilpier-code: Done

src/libs/antares/benchmarking/file_content.h
guilpier-code: Done

src/libs/antares/benchmarking/file_writer.cpp
guilpier-code: Done

src/solver/application.cpp
guilpier-code: If we are sure we want an ini file, we can remove it. But :
- We might want a Json file eventually
- Like this, we separate the essence of a writer from a particular implementation.

To be discussed

src/libs/antares/benchmarking/info_collectors.h
guilpier-code: We wanted more pieces of information : they are pointed by the unused data members below the comment.
Do we still want them ?

src/solver/application.cpp
flomnes: Ok ! Let's keep it but then, make sure to use dynamic polymorphism in the application. Something like
```cpp
// application.h
std::unique_ptr<FileWriter> writer;
// application.cpp
writer = std::make_unique<iniFileWriter>(/* args */);
```

Also, I suggest renaming `FileWriter` to `IFileWriter`

src/libs/antares/benchmarking/info_collectors.cpp
guilpier-code: Using **auto&** to avoid unnecessary copies : done
For the other point : I' m not sure to understand your remark. If you mean we don't need vectors as value of map **duration_items_** (of type `map<string, vector<int64_t>>`), but **duration_items_** should be defined as a `map<string, unsigned int>` to relate a name to a duration sum, I disagree : we need the vector size below the loop.

I replaced the loop with an accumulation though, to make code more concise.

src/libs/antares/benchmarking/info_collectors.cpp
guilpier-code: See remark above

src/libs/antares/benchmarking/info_collectors.h
guilpier-code: See remark above

src/libs/antares/benchmarking/info_collectors.cpp
guilpier-code: Done

src/libs/antares/benchmarking/info_collectors.cpp
guilpier-code: Done

src/solver/application.cpp
guilpier-code: We could, but the code is clear as it is now. If we pass the `Timer` in the **addDuration** method and call stop() inside that method, it's shorter but less clear I think.

flomnes: We can keep the call to `Timer::stop` outside, but place the call to `Timer::get_duration` inside `addDuration`

src/libs/antares/benchmarking/info_collectors.h
flomnes: No, you may remove these data members.

src/libs/antares/benchmarking/info_collectors.cpp
flomnes: I understand. But if we wanted to know the number of calls, we could just increment a counter. On the other hand, it doesn't really matter since we don't expect large vectors, so memory usage won't be an issue.

src/solver/application.h
guilpier-code: Done

src/libs/antares/benchmarking/info_collectors.h
guilpier-code: Done

flomnes (approved):

PR #755, opened 2022-07-05
Replace "Flat" view with "By area" view + refactor

flomnes:


src/ui/simulator/toolbox/input/connection.cpp
```diff
@@ -173,6 +173,121 @@ void Connections::internalBuildSubControls()
...
+    auto localRootId = listbox->AppendItem(rootId, wxString(wxT("By area")), 2);
+    listbox->SetItemBold(localRootId, true);
+
+    // For each area, list the links related to it. For each link, the area is either
+    // its origin or extremity.
+    using ListOfLinks
+      = std::vector<std::pair<Data::AreaLink* /*pointer to link*/,
+                              bool /*true if area is the origin of the link, false otherwise*/>>;
+    // 1. Build hierarchy
+    std::map<Data::AreaName, ListOfLinks> areaToListOfLinks;
```
guilpier-code: What if, instead of using a `map<AreaName, vector<pair<Data::AreaLink*, bool>>`, we use a `map<AreaName, vector<Data::AreaName>>` ?
So **areaToListOfLinks** would change into **areaToListOfAreas**
If so, I guess it would result in a bit more simple code (see further).

src/ui/simulator/toolbox/input/connection.cpp
```diff
@@ -173,6 +173,121 @@ void Connections::internalBuildSubControls()
...
+        Data::Area* area = namedArea.second;
+        if (area->isVisibleOnLayer(layerID))
+        {
+            for (const auto& namedLink : area->links)
+            {
+                Data::AreaLink* lnk = namedLink.second;
+                if (lnk->isVisibleOnLayer(layerID))
+                {
+                    areaToListOfLinks[area->name].push_back({lnk, true});
+                    areaToListOfLinks[lnk->with->name].push_back({lnk, false});
```
guilpier-code: Here we would have : 
```C++
areaToListOfAreas[area->name].push_back(lnk->with->name);
areaToListOfAreas[lnk->with->name].push_back(area->name);
```

src/ui/simulator/toolbox/input/connection.cpp
```diff
@@ -173,6 +173,121 @@ void Connections::internalBuildSubControls()
...
+    for (const auto& al : areaToListOfLinks)
+    {
+        // Reference to the area
+        wxTreeItemId id;
+        // Foreach Interconnection for the area
+        const auto end = al.second.end();
+        for (auto i = al.second.begin(); i != end; ++i)
+        {
+            Data::AreaLink* lnk = i->first;
+            const bool isAreaOriginOfLink = i->second;
```
guilpier-code: We would not need : **const bool isAreaOriginOfLink**

src/ui/simulator/toolbox/input/connection.cpp
```diff
@@ -173,6 +173,121 @@ void Connections::internalBuildSubControls()
...
+                                         1,
+                                         1);
+                listbox->SetItemBold(id, true);
+            }
+            // Adding the item for the interconnection
+            /* wxTreeItemId itemID = */ listbox->AppendItem(
+              id, /*parent*/
+              // caption
+              wxString() << wxT(' ')
+                         << wxStringFromUTF8(isAreaOriginOfLink ? lnk->with->name : lnk->from->name)
```
guilpier-code, line 275: we would not need the ternary operator (a ? b : c)  here

src/ui/simulator/toolbox/input/connection.cpp
```diff
@@ -173,6 +173,121 @@ void Connections::internalBuildSubControls()
...
+    {
+        for (const auto& namedArea : areas)
+        {
+            Data::Area* area = namedArea.second;
+            if (area->isVisibleOnLayer(layerID))
+            {
+                wxTreeItemId id;
+                // Foreach Interconnection for the area
+                const Data::AreaLink::Map::iterator end = area->links.end();
+                for (Data::AreaLink::Map::iterator i = area->links.begin(); i != end; ++i)
```
guilpier-code: Isn't it possible, for the inner loop on links, to have : 
```C++
for (const auto& lnk : area->links)
{
     ....
}
```
?

guilpier-code (changes requested):

src/ui/simulator/toolbox/input/connection.cpp
flomnes: Keeping pointers is easier. If we keep names, we need to lookup names in a table.

When selecting an item, we need to find the link. The correspondance (areaName1, areaName2) -> pointer to link isn't obvious.

src/ui/simulator/toolbox/input/connection.cpp
flomnes: Done

guilpier-code (approved):

PR #767, opened 2022-07-11
Disable DENS output variable if adq-patch is disabled

### LINKED ISSUE #765: Output variable "DENS" should not appear if adq-patch is disabled

flomnes:
- Move `include.adequacyPatch` to `adqPatch.include`
- Add `Parameters::AdequacyPatch::excludedVariables`
- Exclude "DENS" if adequacy patch is disabled

### DIFF
--- a/src/libs/antares/study/parameters.h
+++ b/src/libs/antares/study/parameters.h
@@ -458,7 +458,7 @@ public:
     {
         //! Renewable generation mode
         RenewableGenerationModelling rgModelling;
-        std::vector<std::string> excludedVariables() const;
+        void addExcludedVariables(std::vector<std::string>&) const;
         RenewableGenerationModelling operator()() const;
         void toAggregated();
         void toClusters();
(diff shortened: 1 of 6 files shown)


src/libs/antares/study/parameters.h
```diff
@@ -521,7 +518,9 @@ class Parameters final
             //! rule.
             bool setToZeroOutsideOutsideLinks = true;
         };
+        bool include;
```
guilpier-code, line 521: I guess `bool include` inside `struct AdequacyPatch` replaces former `bool adequacyPatch`.
**include** is unclear to me, maybe **activated** or **used** would be better.
I know that **include** is a part of **include-adequacypatch** in the **generaldata.ini**, but still, in the code (but not in **generaldata.ini**), the associated parameter could be renamed.

guilpier-code (changes requested):

src/libs/antares/study/parameters.h
flomnes: Done. Kept the same name in the study, renamed in the code.

guilpier-code: You made the change **include** --> **enabled** in PR #768, but I also approuve this one so that it can be closed.

guilpier-code (approved):

PR #768, opened 2022-07-11
Make adequacy patch parameters future-proof

### LINKED ISSUE #762: Adequacy patch parameters in wrong section
TODO : move parameters to section **adequacy patch**, rename some to match RTE-i's naming conventions.

flomnes:
Move adequacy-patch parameters to section **adequacy patch** (current : **optimization**). This is to match RTE-i's development branch.

### DIFF
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1435,10 +1435,9 @@ void Parameters::prepareForSimulation(const StudyLoadOptions& options)
     const auto excluded_vars_rg = renewableGeneration.excludedVariables();
     const auto excluded_vars_adq_patch = adqPatch.excludedVariables();
     std::vector<std::string> excluded_vars;
-    excluded_vars.insert(excluded_vars.end(), excluded_vars_rg.begin(),
-                         excluded_vars_rg.end());
-    excluded_vars.insert(excluded_vars.end(), excluded_vars_adq_patch.begin(),
-                         excluded_vars_adq_patch.end());
+    excluded_vars.insert(excluded_vars.end(), excluded_vars_rg.begin(), excluded_vars_rg.end());
+    excluded_vars.insert(
+      excluded_vars.end(), excluded_vars_adq_patch.begin(), excluded_vars_adq_patch.end());
     variablesPrintInfo.prepareForSimulation(thematicTrimming, excluded_vars);
 
     switch (mode)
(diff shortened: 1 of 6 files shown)


src/libs/antares/study/parameters.cpp
```diff
@@ -623,6 +624,22 @@ static bool SGDIntLoadFamily_Optimization(Parameters& d,
     }
     return false;
 }
+static bool SGDIntLoadFamily_AdqPatch(Parameters& d,
+                                      const String& key,
+                                      const String& value,
+                                      const String&,
+                                      uint)
```
guilpier-code, line 631: For information, the last argument of **every SGDIntLoadFamily_some_section(..., uint)** is useless. We should remove it for every function.
Either we correct it in this PR or we create a ticket. Correcting it now would be very quick.

guilpier-code (changes requested):

src/libs/antares/study/parameters.cpp
flomnes: The legacy function uses this argument as a study version number.

https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/eacccf28ca43f2d714365c1da73a41ee4136c57e/src/libs/antares/study/parameters.cpp#L923-L927

guilpier-code: ok. 
here I correct my previous remark.
Note that this new remark leads to a possible refactor. Let's discuss if we need to take this remark into account.

Every **Callback** function that is a value of the map **sectionAssociatedToKeysProcess** do not need the last **uint version** argument. the **Callback** type should also loose the unnecessary argument.
So **SGDIntLoadFamily_Legacy** is kind of apart from the  **SGDIntLoadFamily**.

Another point : the 4th argument **const String&** (which the raw value, before the value is turned to lower case) is only used by the body of **SGDIntLoadFamily_General**. So there is a kind of repetition in the list of arguments of this family of function, example : 
```C++
handleAllKeysInSection(*this, p->key, value, p->value, version)
```
We could try to remove one of the **value** argument as well.

guilpier-code (changes requested):

src/libs/antares/study/parameters.cpp
flomnes: See #774

guilpier-code (approved):

PR #776, opened 2022-07-13
Polish the doc related to the adq-patch

flomnes:


PPlessiez (approved):
I proposed some slight changes of formulation + the addition of the DENS variable in the list of outputs, but besides that it seems good!

PR #780, opened 2022-07-17
Save output to .zip archive

flomnes:
Save output to one single .zip file
### TODO
- [ ] Include all files to this archive
- [ ] Create a new option : save as .zip (new), or as files (legacy)
- [ ] Make sure scripts (R, python) can read files easily when they're in a zip archive
- [ ] Also discuss this with the Xpansion team : save all MPS files in one archive. This is an easy test-case, and useful in the short-term since it would may allow to solve the disk usage issue.
- [ ] Create a generic interface allowing other output formats : SQL, HDF5, etc. and to avoid introducing those dependencies deep into our codebase
- [ ] Check technical limitations (RAM consumption, does libzip write when calling `close` or whenever data is added ?, etc.)
- [ ] Check licensing issues : is ZIP free to use ? Can we use all underlying components (libzippp, libzip, etc.) ?
- [ ] Assess performance : is it shorter to write a .zip than multiple files then zip them all ?
```cpp
// We don't want this everywhere in our code
switch (outputType) {
case file:
   result.saveAsFile();
   break;
case archive:
   result.saveAsZip();
   break;
case sql:
   sqlDB->connect();
   result.saveAsSQLTable(sqlDB);
   break;
}

// We want this
result.write(output);
```

### DIFF
--- a/.github/workflows/centos7-system-deps-build.yml
+++ b/.github/workflows/centos7-system-deps-build.yml
@@ -26,7 +26,7 @@ jobs:
            yum install -y git redhat-lsb-core gcc gcc-c++ make wget centos-release-scl scl-utils rpm-build
            yum install -y cmake3 devtoolset-9
            yum install -y rh-git227-git
-           yum install -y unzip libuuid-devel wxGTK3-devel boost-test boost-devel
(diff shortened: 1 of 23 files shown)


flomnes:
See #794

PR #792, opened 2022-07-25
Add option to set link capacity to null/infinity for physical links only

flomnes:


flomnes:
@JMJ-rte Tested with adq-patch=ON and wasn't satisfied with the results. Investigate interaction with adq-patch/CSR (current and expected).

### EDIT

1. Set capacities to 0/infinity according to the global/local values
2. Set capacities to 0/current according to adequacy patch

So in adq patch mode, virtual links should be ignored if new options are used (set to null (physical links), set to infinite (physical links)). This is the right behavior.

payetvin (approved):

src/ui/simulator/windows/options/optimization/optimization.cpp
```diff
@@ -120,6 +120,14 @@ static void ResetButton(Component::Button* button, Data::TransmissionCapacities
         button->image("images/16x16/infinity.png");
         button->caption(wxT("set to infinite"));
         break;
+    case Data::tncInfinitePhysical:
```
guilpier-code: When we select "set to null (physical links)", this selected item goes out of the window **Optimization preferences**, we should consider enlarging this window.

docs/reference-guide/13-file-format.md
```diff
@@ -3,6 +3,14 @@ This is a list of all recent changes that came with new Antares Simulator featur
 
 ## v8.4.0
 ### Input
+#### Zero/infinite capacity for physical links only
+In file **settings/generaldata.ini**, in section `optimization`, add 2 values for existing property `transmission-capacities` [str]:
+
+* `false-physical` sets the capacity of physical links to 0
```
guilpier-code: In the file **generaldata.ini**, for key **transmission-capacities**, the possible value **false-physical** is unclear to me, I would (at least) rename it into **null-physical**. A better value would be **null-for-physical-link**.
I guess that the **false** in  **false-physical** comes from the historic value **false** (which stands for _null for any link_).
We can even step further with the 4 following values : 
- **null-for-any-link** (or **false** for compatibility with old studies)
- **infinite-for-any-link** (or **true** for compatibility with old studies)
- **null-for-physical-link**
- **infinite-for-physical-link**

If we want to go a bit further, we could make a renaming change internally, that is in the associated **enum** : 
- **tncEnabled** --> **enabled**
- **tncIgnore** --> **nullForAnyLink**
- **tncInfinite** --> **infiniteForAnyLink**
- **tncIgnorePhysical** --> **nullForPhysicalLink**
- **tncInfinitePhysical** --> **infiniteForPhysicalLink**

Note that I consider prefix **tnc** as useless, as the enum is named **TransmissionCapacities**

src/libs/antares/study/area/links.cpp
```diff
@@ -233,6 +233,60 @@ bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& fol
...
+    case tncInfinitePhysical: // Use 'infinity' only for physical links
+        return tncInfinite;
+
+    // tncIgnore, tncInfinite
+    default:
+        return tncGlobal; // Use the global property
+    }
+}
+
+static TransmissionCapacities overrideVirtual(TransmissionCapacities tncGlobal,
```
guilpier-code: Maybe more clear in the following excerpt, as we know right away when to change a virtual link NTC) : 
```C++
static TransmissionCapacities overrideVirtual(TransmissionCapacities tncGlobal,
                                              TransmissionCapacities tncLocal)
{
    switch (tncGlobal)
    {
        // We change the NTC
        case tncIgnore:
        case tncInfinite: 
            return tncGlobal;

        // We don't not change the NTC
        default:
             return tncLocal;
    }
}

```

src/libs/antares/study/parameters.cpp
```diff
@@ -618,6 +618,10 @@ static bool SGDIntLoadFamily_Optimization(Parameters& d,
         v.toLower();
         if (v == "infinite")
             d.transmissionCapacities = tncInfinite;
+        else if (v == "infinite-physical")
```
guilpier-code: Please see remarks on file **13-file-format.md**

src/ui/simulator/windows/inspector/accumulator.hxx
```diff
@@ -430,6 +430,8 @@ struct PLinkCopperPlate
             return wxT("set to zero");
         case Data::tncInfinite:
             return wxT("set to infinite");
+        default:
```
guilpier-code: This change is about `struct PLinkCopperPlate`, which does not seem to be used ! At least I can't see a place where it is used

src/ui/simulator/windows/options/optimization/optimization.cpp
```diff
@@ -708,23 +716,39 @@ void Optimization::onPopupMenuTransmissionCapacities(Component::Button&, wxMenu&
                           wxEmptyString);
     menu.Connect(it->GetId(),
                  wxEVT_COMMAND_MENU_SELECTED,
-                 wxCommandEventHandler(Optimization::onSelectTransCapInclude),
+                 wxCommandEventHandler(Optimization::onSelectTransmissionCapacity<Data::tncEnabled>),
```
guilpier-code: The body of **onPopupMenuTransmissionCapacities** could be 4 lines, that is 4 calls to a template function that creates item and put it in a menu, like it was recently done for MPS writers (see PR #872)

guilpier-code (changes requested):

docs/reference-guide/13-file-format.md
flomnes: I agree on everything. To avoid the prefix, we can use an **enum class**.

We should keep existing values in the read functions for compatibility
```cpp
if (key == "null-for-any-link" || key == "false") {
   prop = tnc::nullForAllLinks;
}
```

src/ui/simulator/windows/options/optimization/optimization.cpp
flomnes: OK, will fix if easy

src/ui/simulator/windows/options/optimization/optimization.cpp
flomnes: Yes, we can go one step forward & simplify even more

src/ui/simulator/windows/inspector/accumulator.hxx
flomnes: True, this was to avoid missing cases after adding new enum values. This is no longer necessary since there now 2 enums (local & global)

src/ui/simulator/windows/options/optimization/optimization.cpp
flomnes: Cannot reproduce in Linux

src/libs/antares/study/area/links.cpp
```diff
@@ -238,6 +238,50 @@ bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& fol
     return success;
 }
 
+bool AreaLink::isLinkPhysical() const
```
guilpier-code, line 240: Or you can do it shorter : 
```C++
bool AreaLink::isLinkPhysical() const
{
    switch (assetType)
    {
    case atVirt:
        return false;
    default:
        return true;
    }
}
```

src/libs/antares/study/area/links.cpp
```diff
@@ -387,7 +431,8 @@ static bool AreaLinksInternalLoadFromProperty(Study& study,
         bool copperPlate;
         if (value.to<bool>(copperPlate))
         {
-            link.transmissionCapacities = (copperPlate) ? Data::tncInfinite : Data::tncEnabled;
+            using LT = Data::LocalTransmissionCapacities;
```
guilpier-code: `using LocalNTCtype = Data::LocalTransmissionCapacities;` ?

src/libs/antares/study/area/links.cpp
```diff
@@ -450,15 +495,16 @@ static bool AreaLinksInternalLoadFromProperty(Study& study,
     }
     if (key == "transmission-capacities")
     {
+        using LT = Data::LocalTransmissionCapacities;
```
guilpier-code: Same as above

src/libs/antares/study/fwd.cpp
```diff
@@ -410,5 +410,68 @@ mpsExportStatus stringToMPSexportStatus(const AnyString& value)
     return mpsExportStatus::UNKNOWN_EXPORT;
 }
 
+std::string GlobalTransmissionCapacitiesToString(
+  GlobalTransmissionCapacities transmissionCapacities)
+{
+    switch (transmissionCapacities)
+    {
+        using GT = GlobalTransmissionCapacities;
```
guilpier-code: `using GlobalNTCtype = GlobalTransmissionCapacities;` ?

src/libs/antares/study/fwd.h
```diff
@@ -128,19 +128,41 @@ enum SimplexOptimization
...
+  GlobalTransmissionCapacities transmissionCapacities);
+std::string GlobalTransmissionCapacitiesToString_Display(
+  GlobalTransmissionCapacities transmissionCapacities);
+bool stringToGlobalTransmissionCapacities(const AnyString& value,
+                                          GlobalTransmissionCapacities& out);
+
+/*!
+** \brief Transmission capacities mode (applies to individual links)
+*/
+enum class LocalTransmissionCapacities
```
guilpier-code, line 159: Separating local NTC type from global NTC type : you did well !

src/libs/antares/study/links/property-links-helper.h
```diff
@@ -7,13 +7,13 @@ namespace Antares
 {
 namespace Data
 {
-std::string transmissionCapacitiesToString(const TransmissionCapacities& tc)
+std::string transmissionCapacitiesToString(const LocalTransmissionCapacities& tc)
```
guilpier-code: I'm not a fan of this source **.h** file.
Indeed, for other parameters, the **enum** and the functions to act on the value of type **enum** are in **fwd.h/cpp**. Here, functions are defined in **property-links-helper.h**.

A data, some functions to interact with the data. It's a class.
I guess every parameters should be encapsulated in a class defined in a separate source file.

src/ui/simulator/toolbox/components/datagrid/renderer/links/summary.cpp
```diff
@@ -178,12 +183,12 @@ bool Summary::cellValue(int x, int y, const String& value)
         return true;
     case 1:
         if (pTmp == "enabled" || pTmp == "e")
-            link->transmissionCapacities = Data::tncEnabled;
+            link->transmissionCapacities = LocalTransmissionCapacities::enabled;
         else if (pTmp == "infinite" || pTmp == "set to infinite" || pTmp == "inf" || pTmp == "i")
-            link->transmissionCapacities = Data::tncInfinite;
+            link->transmissionCapacities = LocalTransmissionCapacities::enabled;
```
guilpier-code: It seems there an error here : 
```C++
else if (pTmp == "infinite" || pTmp == "set to infinite" || pTmp == "inf" || pTmp == "i")
      link->transmissionCapacities = LocalTransmissionCapacities::enabled; // error here, should be "infinite"
```

src/ui/simulator/windows/inspector/accumulator.hxx
guilpier-code: So let's remove it

guilpier-code (changes requested):

src/libs/antares/study/area/links.cpp

src/libs/antares/study/area/links.cpp
flomnes: done

src/libs/antares/study/area/links.cpp

src/libs/antares/study/fwd.cpp
flomnes: done

src/libs/antares/study/fwd.h
flomnes: thanks

src/ui/simulator/toolbox/components/datagrid/renderer/links/summary.cpp
flomnes: done

guilpier-code (approved):

PR #793, opened 2022-07-25
Fix wrong number of cores + code de-duplication

### LINKED ISSUE #789: Simulation cores parameter doesn't work without saving
**Description**
Changing the "Simulation cores" parameter inside the Advanced parameters window, is not taken into account when running the study, if you don't save the study.

**How to Reproduce**
- Create an empty study with "minimum" as "Simulation cores"
- Add one area
- Save the study
- Click "Run a simulation", select parallel, the screen prints Simulation cores: 1 as expected
- Change "Simulation cores" to "maximum" in the Advanced parameters
- Click "Run a simulation", select parallel, the screen still prints Simulation cores: 1 <- it depends on your system but it should be higher

**Expected behavior**
Either we should get a prompt to save the study before running it, this would solve this issue as when you save the study, the correct number of simulation cores will be displayed in the "Run simulation window". Or, the number of simulation cores should dynamically evolve with the option chosen.

**OS and version**
 - OS: reported to me on Windows, reproduced on Ubuntu
 - Antares Simulator version: reported on to me on 8.0.4, reproduced on 8.3.0

flomnes:


src/ui/simulator/windows/options/advanced/advanced.cpp
```diff
@@ -941,74 +941,45 @@ void AdvancedParameters::onNumberOfCores(Component::Button&, wxMenu& menu, void*
...
-    auto& study = *Data::Study::Current::Get();
+    auto study = Data::Study::Current::Get();
 
-    if (study.parameters.nbCores.ncMode != Data::ncMin)
+    if (study->parameters.nbCores.ncMode != ncMode)
     {
-        study.parameters.nbCores.ncMode = Data::ncMin;
+        study->parameters.nbCores.ncMode = ncMode;
+        // Force refresh for study->nbYearsParallelRaw
+        study->getNumberOfCores(false, 1 /* ignored */);
```
flomnes, line 954: Actual change is here, the rest is refactor

src/ui/simulator/windows/options/advanced/advanced.cpp
```diff
@@ -941,74 +941,45 @@ void AdvancedParameters::onNumberOfCores(Component::Button&, wxMenu& menu, void*
                  this);
 }
 
-void AdvancedParameters::onSelectNCmin(wxCommandEvent& /* evt */)
+void AdvancedParameters::onSelectNC(Data::NumberOfCoresMode ncMode)
```
guilpier-code: Not really relevant to your dev, but  **Data::NumberOfCoresMode** isn't a good name, **NumberOfCoresLevel** would be better.
And so **onSelectNC** should be renamed **onSelectNumberOfCoresLevel**

guilpier-code (approved):

PR #794, opened 2022-07-27
Add option to write results into a single zip archive

flomnes:
This features adds the possibility to write results to a single .zip file. There was a lot of refactoring to centralize output, since before that everything was based on files & directories. I created an interface class & a factory method, depending on a new parameter.

### TODO
- [x] Move minizip-ng to antares-deps #814
- [x] MPS files, and other Xpansion-related files #807
- [x] Annual statistics (checkIntegrity.txt, etc.) #808
- [x] Export about-the-study #804 
- [x] simulation.log #805
- [x] hydro debug files #809
- [x] Export generated timeseries to the .zip archive if "store in output" is enabled
- [x] Export generated timeseries when the `-g` option is provided
- [x] Avoid copies when possible, switch from (buffer, size) to `std::string` ?
- [x] Fix failing linking in the Sonar job #810
- [x] Performance tests : check if we actually save time by saving the results into a single zip archive.
- [x] Document the feature (command-line option & study parameter) #813
- [x] Document the [study format change](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/develop/docs/reference-guide/13-file-format.md) #813
- [x] Handle errors from minizip-ng (not enough space on disk, incorrect path, etc.) #822
- [ ] Track and fix bugs
- [x] Make sure that a valid .zip file is produced the program is terminated (Ctrl+C, etc.), see [SO](https://stackoverflow.com/questions/2350489/how-to-catch-segmentation-fault-in-linux) #827
- [x] Write an Architecture Decision Record (ADR)
- [ ] Add unit tests
- [X] TS numbers
- [ ] Handle case where archive / directory already exists


src/libs/antares/study/fwd.h
```diff
@@ -540,6 +540,14 @@ const char* DayAheadReserveManagementModeToCString(DayAheadReserveManagement daR
 */
 DayAheadReserveManagement StringToDayAheadReserveManagementMode(const AnyString& text);
 
+// Format of results
+enum ResultFormat
+{
+    // Store outputs as files inside directories
+    legacyFilesDirectories = 0,
```
guilpier-code, line 530: I see what you mean with **legacy**, but someone else does not necessarily know. It is supposed to mean text files.
**legacyFilesDirectories** ==> **textFiles** ?

src/libs/antares/study/parameters.h
```diff
@@ -557,6 +557,8 @@ class Parameters final
     //! Ortool solver used for simulation
     OrtoolsSolver ortoolsEnumUsed;
     //@}
+    // Format of results. Currently, only single files or zip archive are supported
+    ResultFormat resultFormat = legacyFilesDirectories;
```
guilpier-code, line 562: There are often many result**s** (with an **s**), even if that code is already understandable

src/libs/antares/study/study.cpp
```diff
@@ -774,22 +775,23 @@ bool Study::prepareOutput()
...
 
     // Settings
+    // TODO : use writer
     buffer.clear() << folderOutput << SEP << "about-the-study";
-    if (not IO::Directory::Create(buffer))
-    {
-        logs.error() << "I/O: impossible to create the directory " << buffer;
-        return false;
-    }
+    IO::Directory::Create(buffer);
```
guilpier-code: Why did you get rid of : 
```C++
if (not IO::Directory::Create(buffer))
{
        logs.error() << "I/O: impossible to create the directory " << buffer;
        return false;
 }
```
?

src/solver/writer/i_writer.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Antares
+{
+namespace Solver
+{
+class IResultWriter
+{
+public:
+    using Ptr = std::shared_ptr<IResultWriter>;
+    virtual void addJob(const std::string& entryPath, Yuni::Clob& entryContent) = 0;
```
guilpier-code: Functions **addJob** actually flushes the results on disk, into a file format or another (.txt or .zip). Why calling them **addJob** rather than **flush** or **save** or **dump** ?

src/solver/simulation/solver.hxx
```diff
@@ -265,7 +264,10 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
  pHydroManagement(study),
  pFirstSetParallelWithAPerformedYearWasRun(false),
  pAnnualCostsStatistics(study),
- pDurationCollector(duration_collector)
+ pDurationCollector(duration_collector),
+ qs(std::make_shared<Yuni::Job::QueueService>()),
+ pResultWriter(
+   resultWriterFactory(study.parameters.resultFormat, study.folderOutput, qs, duration_collector))
```
guilpier-code: **qs** ==> **qs_**, **_qs** or **pQs**

src/solver/writer/zip_writer.cpp
```diff
@@ -0,0 +1,106 @@
...
+      new ZipWriteJob<std::string>(*this, entryPath, entryContent, pDurationCollector),
+      Yuni::Job::priorityLow);
+}
+
+void ZipWriter::addJob(const std::string& entryPath, Antares::IniFile& entryContent)
+{
+    EnsureQueueStartedIfNeeded ensureQueue(this, pQueueService);
+    std::string buffer;
+    entryContent.saveToString(buffer);
+    pQueueService->add(new ZipWriteJob<std::string>(*this, entryPath, buffer, pDurationCollector),
```
guilpier-code: Can't we replace ini files **addJob**'s body with something like : 
```C++
void ZipWriter::addJob(const std::string& entryPath, Antares::IniFile& entryContent)
{
    std::string buffer;
    entryContent.saveToString(buffer);
    addJob(entryPath, buffer);
}
```
and this to avoid some kind of code duplication ?

src/solver/writer/zip_writer.cpp
```diff
@@ -0,0 +1,106 @@
...
+ pDurationCollector(duration_collector)
+{
+}
+
+template<class ContentT>
+void ZipWriteJob<ContentT>::onExecute()
+{
+    mz_zip_file file_info;
+
+    Benchmarking::Timer timer_wait;
```
guilpier-code, line 54: Here we have 3 kinds of tasks got done : time measurement and compress job.
We could separate them like this (it's an example) : 
```C++
template<class ContentT>
void ZipWriteJob<ContentT>::onExecute()
{
    // ok, I don't see how to split this block
    Benchmarking::Timer timer_wait;
    std::lock_guard<std::mutex> guard(pZipMutex); // Wait
    timer_wait.stop();
    pDurationCollector->addDuration("zip_wait", timer_wait.get_duration());

    Benchmarking::Timer timer_write;
    mz_zip_file file_info = prepareFileInfo();  // I guess preparing file is part of compressing, so it's measured as well
    zip_content(file_info);
    timer_write.stop();
    pDurationCollector->addDuration("zip_write", timer_write.get_duration());
}
```
where **prepareFileInfo()** and **zip_content(...)** would be private method of class **ZipWriteJob**

src/solver/writer/zip_writer.h
```diff
@@ -0,0 +1,68 @@
...
+class ZipWriter : public IResultWriter
+{
+public:
+    ZipWriter(std::shared_ptr<Yuni::Job::QueueService> qs,
+              const char* archivePath,
+              Benchmarking::IDurationCollector* duration_collector);
+    ~ZipWriter();
+    void addJob(const std::string& entryPath, Yuni::Clob& entryContent) override;
+    void addJob(const std::string& entryPath, std::string& entryContent) override;
+    void addJob(const std::string& entryPath, Antares::IniFile& entryContent) override;
```
guilpier-code: So **addJob** methods are made to run jobs. 
Calling them **addJob** could be confusing, it gives the idea that we add some job to the queue and that we run the jobs later. 
Why not call them **run(...)** or **write(...)** ?

src/libs/antares/benchmarking/file_content.h
```diff
@@ -19,13 +19,16 @@ class FileContent
     void addItemToSection(const std::string& section, const std::string& key, const std::string& value);
     void addDurationItem(const std::string& name, unsigned int duration, int nbCalls);
 
+    template<class DestT>
+    DestT to();
```
guilpier-code: What does this function do exactly (I mean : **to()**) ?
After having looked, I see that it converts the content into an ini file content.
- The name is unclear to me, we should rename this function (**convertIntoFormat\<TargetFormatType\>()**)
- Why not dropping the template parameter and call the function **convertIntoIniContent()** 
- By adding this function, we make the class grow, it starts doing something it shouldn't do, that is converting the content into another kind of content (ini content). We should start to think of extracting the conversion into another function or class, for example by introducing a **decorator** that would make the conversions. If some other conversion is made, we would not have to touch the **FileContent** class.

src/solver/writer/zip_writer.h
```diff
@@ -0,0 +1,68 @@
...
+#include <yuni/core/string.h>
+
+#include "i_writer.h"
+
+namespace Antares
+{
+namespace Solver
+{
+class ZipWriter;
+template<class ContentT>
```
guilpier-code, line 25: **ContentT** ==> **ContentType**

src/solver/simulation/solver.hxx
```diff
@@ -265,7 +264,10 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
  pHydroManagement(study),
  pFirstSetParallelWithAPerformedYearWasRun(false),
  pAnnualCostsStatistics(study),
- pDurationCollector(duration_collector)
+ pDurationCollector(duration_collector),
+ qs(std::make_shared<Yuni::Job::QueueService>()),
+ pResultWriter(
```
guilpier-code: I would have put **pResultWriter** initialization in the constructor body to highlight it a bit more, and because it is a more complex initialization than just affecting a value, but it's a question of taste I suppose.

src/solver/main/economy.cpp
```diff
@@ -51,6 +51,7 @@ void Application::runSimulationInEconomicMode()
 
         this->pOptimizationInfo = simulation.getOptimizationInfo();
     }
+    this->pResultWriter = simulation.pResultWriter;
```
guilpier-code: I think we should not make this kind of initialization (line 54), as if it was done behind the scene.
In this case, **pResultWriter** should be initialized in a method class **Application**, as early as possible in the execution, and then passed to the simulation : 
```C++
   // See line 42
   SimulationType simulation(*pStudy, pSettings, &pDurationCollector, pResultWriter);
```
Same remark for mode **adequacy** and **draft**.

guilpier-code (changes requested):

src/solver/main/economy.cpp
flomnes: I agree. But since the output folder isn't known when `Application` is created, we'd have to provide an `initialize` method, instead if providing it to the constructor.

src/solver/writer/zip_writer.h
guilpier-code: But a question bothers me : 
Why do we use jobs to do the work ? Jobs are launched in a different thread than the calling thread. Why don't we execute the work of adding the zip archive a piece of data in the same thread than the calling thread ?
And why don't we run a new job for writing classic txt files ?

src/solver/writer/immediate_file_writer.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
guilpier-code, line 1: Why calling the txt result writer the immediate result writer ? Why **immediate** ?
Can't we just call it **txtResultsWriter** ?

guilpier-code: Same remark for the source file name

src/solver/writer/immediate_file_writer.cpp
```diff
@@ -0,0 +1,78 @@
...
+    auto output = prepareDirectoryHierarchy(pOutputFolder, entryPath);
+    IOFileSetContent(output, entryContent);
+}
+
+// Write to file immediately, creating directories if needed
+void ImmediateFileResultWriter::addJob(const std::string& entryPath, Antares::IniFile& entryContent)
+{
+    auto output = prepareDirectoryHierarchy(pOutputFolder, entryPath);
+    std::string buffer;
+    entryContent.saveToString(buffer);
```
guilpier-code: Regarding : 
```C++
std::string buffer;
entryContent.saveToString(buffer);
```
We could expect : 
```C++
std::string buffer = entryContent.intoString();
``` 
I guess a call to std::string move consrtuctor will be made when returning a std::string in **intoString()**

src/libs/antares/study/mps.cpp
```diff
@@ -37,9 +37,9 @@ namespace Antares
 {
 namespace Data
 {
-FILE* Study::createFileIntoOutputWithExtension(const YString& prefix,
-                                               const YString& extension,
-                                               uint numSpace) const
+std::string Study::createFileIntoOutputWithExtension(const YString& prefix,
```
guilpier-code: We should rename this function to stick to what it really does, for instance into **createOutputFileName(...)**

src/solver/application.cpp
```diff
@@ -551,6 +556,9 @@ void Application::writeExectutionInfo()
     pTotalTimer.stop();
     pDurationCollector.addDuration("total", pTotalTimer.get_duration());
 
+    if (!pResultWriter)
```
guilpier-code: We could have : 
```C++
pTotalTimer.stop(); // If we have a embarrassing return, we should stop first the timer, right ?

if (!pStudy || !pResultWriter)
    return;

pDurationCollector.addDuration("total", pTotalTimer.get_duration());
```

src/solver/simulation/solver.hxx
guilpier-code: Furthermore, why making **qs** a smart pointer ? Why not a reference data member ?

src/solver/simulation/solver.hxx
guilpier-code: Reminder : I said somewhere else that, **pResultWriter** is shared with (at least) **Application** class, it should be imported from outside the **ISimulation** constructor

src/solver/ts-generator/xcast/xcast.cpp
```diff
@@ -85,21 +89,18 @@ void XCast::exportTimeSeriesToTheOutput(Progression::Task& progression, Predicat
         String filename;
```
guilpier-code, line 88: I don't know why the current function **exportTimeSeriesToTheOutput**(...)  in a member of class **XCast** : it does not use a data / function member of  **XCast** (apart from **study**, ..., and now **pWriter**, ok).
Maybe we should make it an external function and so make **XCast** thinner. Ok, not a priority.

src/solver/ts-generator/hydro.cpp
```diff
@@ -311,19 +314,23 @@ bool GenerateHydroTimeSeries(Data::Study& study, uint currentYear)
...
+                {
+                    std::string storage_buffer;
                     output.clear() << study.buffer << SEP << "storage.txt";
-                    area.hydro.series->storage.saveToCSVFile(output);
-
-                    area.hydro.series->flush();
+                    writer->addJob(output.c_str(), storage_buffer);
                 }
 
+                area.hydro.series->flush();
```
guilpier-code: As far I can see, **flush()** does nothing (empty body), let's remove it

src/solver/ts-generator/hydro.cpp
```diff
@@ -311,19 +314,23 @@ bool GenerateHydroTimeSeries(Data::Study& study, uint currentYear)
...
-                                     << "hydro" << SEP << "mc-" << currentYear << SEP << area.id;
-                if (IO::Directory::Create(study.buffer))
+                study.buffer.clear() << "ts-generator" << SEP << "hydro" << SEP << "mc-"
+                                     << currentYear << SEP << area.id;
+
                 {
+                    std::string ror_buffer;
                     output.clear() << study.buffer << SEP << "ror.txt";
-                    area.hydro.series->ror.saveToCSVFile(output);
+                    writer->addJob(output.c_str(), ror_buffer);
```
guilpier-code: I don't understand how the following instruction can work : 
```c++
writer->addJob(output.c_str(), ror_buffer);
```
`area.hydro.series->ror` should be turned into a buffer, and then passed to **addJob**

src/solver/ts-generator/hydro.cpp
```diff
@@ -311,19 +314,23 @@ bool GenerateHydroTimeSeries(Data::Study& study, uint currentYear)
...
+                }
 
+                {
+                    std::string storage_buffer;
                     output.clear() << study.buffer << SEP << "storage.txt";
-                    area.hydro.series->storage.saveToCSVFile(output);
-
-                    area.hydro.series->flush();
+                    writer->addJob(output.c_str(), storage_buffer);
```
guilpier-code: Same as above

src/solver/ts-generator/thermal.cpp
```diff
@@ -157,42 +166,22 @@ void GeneratorTempData::writeResultsToDisk(const Data::Area& area,
...
-    folder.reserve(study.folderOutput.size() + 256);
-
-    study.areas.each([&](Data::Area& area) {
-        if (archive)
-        {
-            folder.clear();
-            folder << study.folderOutput << SEP << "ts-generator" << SEP << "thermal" << SEP
-                   << "mc-" << year << SEP << area.id;
+        pWriter->addJob(pTempFilename.c_str(), buffer);
+    }
```
guilpier-code, line 182: Please remove **prepareOutputFoldersForAllAreas** declaration in class **GeneratorTempData**

guilpier-code (changes requested):

src/solver/utils/mps_utils.cpp
```diff
@@ -16,14 +17,14 @@ using namespace Antares::Data;
 
 constexpr size_t OPT_APPEL_SOLVEUR_BUFFER_SIZE = 256;
 
-static void printHeader(FILE* Flot, int NombreDeVariables, int NombreDeContraintes)
+static void printHeader(Clob& Flot, int NombreDeVariables, int NombreDeContraintes)
```
guilpier-code: In this source file, we could replace **Clob** with **std::string**, we would get rid of Yuni and the code would be faster to execute (no attention to performance issues was given in **Yuni::Clob**)

src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -139,7 +139,7 @@ void InternalExportDigestLinksMatrix(const Data::Study& study,
     }
 }
 
-void ExportGridInfosAreas(const Data::Study& study, const String& folder)
+static void ExportGridInfosAreas(const Data::Study& study, IResultWriter::Ptr writer)
 {
     Clob out;
```
guilpier-code, line 119: We should try to remove **Yuni::Clob** from this function, and from this file as well (if it's easy to do and test), and replace it with **std::string**, faster at execution

guilpier-code (changes requested):

src/solver/utils/mps_utils.cpp
flomnes: The problem is that as of C++17, `std::string` provides no formatting capabilities, for example `"%4.4f"`

src/solver/variable/surveyresults/surveyresults.cpp
flomnes: Sure

src/solver/writer/i_writer.h
guilpier-code: Maybe you checked it, but I does not seem that we need **addJob** in **Yuni::Clob** version, or we can avoid it. I did not make a full check...
In case I'm right, we would get rid of a method and get rid of the **Yuni** dependance

guilpier-code (changes requested):

src/solver/main/economy.cpp
guilpier-code: ok. As I say in PR #804, the study itself needs the writer to write the **about-the-study** things.
So I guess we should initialize the writer somewhere in the study as soon as the kind of writer is known.

src/solver/simulation/solver.hxx
flomnes: To avoid dangling references

src/libs/antares/study/study.cpp
flomnes: Because directories are created automatically

src/solver/writer/i_writer.h
flomnes: Sadly, it's needed.

flomnes: > Functions **addJob** actually flushes the results on disk, into a file format or another (.txt or .zip). Why calling them **addJob** rather than **flush** or **save** or **dump** ?

Because the file content isn't saved immediately, or flushed.

It's only written immediately if the "files" version is used.

src/solver/writer/zip_writer.cpp
flomnes: Done, in another PR. Thanks

src/solver/writer/zip_writer.h
flomnes: > Why don't we execute the work of adding the zip archive a piece of data in the same thread than the calling thread ?

Because zip files can't be written concurrently, _ie_ since the lib isn't thread-safe. All operations on it are `mutex`ed. If we wrote results immediately in this case, a lot of CPU time would be wasted waiting for write operations to be completed. We don't want that, so we give the responsability of writing data to a separate thread.

src/solver/writer/i_writer.h
guilpier-code: ok

src/solver/simulation/solver.hxx
guilpier-code: Ok

PR #804, opened 2022-08-17
Save about-the-study and files therein using generic result writer

flomnes:


src/libs/antares/study/study.cpp
```diff
@@ -773,51 +779,76 @@ bool Study::prepareOutput()
...
+  Antares::IniFile header_ini;
+  header.CopySettingsToIni(header_ini, false);
+  pResultWriter->addJob(path.c_str(), header_ini);
+
+  // copying the generaldata.ini
+  // Input : absolute path
+  path.clear() << folderSettings << SEP << "generaldata.ini";
+  logs.notice() << path;
+  std::string generaldata_buffer;
+  Yuni::IO::File::LoadFromFile(generaldata_buffer, path);
```
guilpier-code: What does a call to a **load** function do inside a **save** function ? Does it do a copy (as expected) ?

src/libs/antares/study/study.cpp
```diff
@@ -773,51 +779,76 @@ bool Study::prepareOutput()
...
-        f << "\ntimestamp = " << now;
-        f << "\n\n";
-    }
-    else
-        logs.error() << "I/O: impossible to write " << buffer;
-
     return true;
 }
 
+void Study::saveMiscFilesIntoOutput()
```
guilpier-code: I would rename this method into **save_about_the_study()**

src/libs/antares/study/study.cpp
```diff
@@ -145,6 +146,11 @@ void Study::clear()
     gotFatalError = false;
 }
 
+void Study::setWriter(Solver::IResultWriter::Ptr writer)
```
guilpier-code: Why this function (**setWriter**) ?
I said in a previous review (#794 : src/solver/main/economy.cpp) that the writer should be created in the Application class, where it is first needed, and then passed to any component that needs it (simulation, xcast, ...)
Now, I would go a bit further : the writer should be created in the study, as soon as **study->parameters** gets the kind of writer should be buit, and then it should be passed to Application, simulation, and so on.

guilpier-code (changes requested):

src/libs/antares/study/study.h
```diff
@@ -372,6 +372,8 @@ class Study final : public Yuni::NonCopyable<Study>, public IObject
     */
     bool prepareOutput();
 
+    void saveMiscFilesIntoOutput();
```
guilpier-code: See **study.cpp** for remarks

src/solver/application.cpp
```diff
@@ -377,6 +377,10 @@ void Application::execute()
         break;
     }
 
+    // Save about-the-study files (comments, notes, etc.)
+    pStudy->setWriter(pResultWriter);
```
guilpier-code: Please see comment in **study.cpp**

src/libs/antares/study/simulation.cpp
```diff
@@ -68,6 +68,14 @@ bool Simulation::saveToFolder(const AnyString& folder) const
     return false;
 }
 
+void Simulation::saveUsingWriter(Solver::IResultWriter::Ptr writer, const AnyString& folder) const
```
guilpier-code: In method **Simulation::saveToFolder(...)**, instruction `b = folder;` is run twice.

src/libs/antares/study/simulation.h
```diff
@@ -64,6 +65,8 @@ class Simulation final
     */
     bool saveToFolder(const AnyString& folder) const;
 
+    void saveUsingWriter(Solver::IResultWriter::Ptr writer, const AnyString& folder) const;
```
guilpier-code, line 68: Current class and source file have a terrible name (**Simulation**).
Actually, it's all about **user's notes**, loading and saving them (modifying them is handled in : 
src/ui/simulator/windows/notes.cpp).
It can be misleading. So this class should renamed **UserNotes**. Same for source file.
The content of user's notes should be renamed from **comments** into **content**.

guilpier-code: Once "simulation" renamed, **saveUsingWriter** will be able to keep its name.

src/libs/antares/study/study.cpp
```diff
@@ -773,51 +779,76 @@ bool Study::prepareOutput()
...
     return true;
 }
 
+void Study::saveMiscFilesIntoOutput()
+{
+  String path;
+  path.reserve(1024);
+
+  path.clear() << "about-the-study";
+  simulation.saveUsingWriter(pResultWriter, path);
```
guilpier-code: As said in **simulation.h**'s review, we should rename study's **simulation** data member into **user_notes_**

src/solver/application.cpp
```diff
@@ -377,6 +377,10 @@ void Application::execute()
...
+    pStudy->saveMiscFilesIntoOutput();
```
guilpier-code: Reminder : 
in case we decide to initialize the writer in the study (and broadcast it in the rest of the program),  the call to **saveMiscFilesIntoOutput()** should be made (if **study.parameters** has the writer's info) where it was originally, that next to call to **prepareOutput()**

guilpier-code (changes requested):

src/solver/application.cpp
flomnes: yes

flomnes (commented):
@guilpier-code DONE

src/libs/antares/study/study.cpp
flomnes: I've renamed it to `SimulationMetadata`

flomnes:
@guilpier-code DONE

src/libs/antares/study/study.cpp
guilpier-code: I'm sorry but **simulation meta data** doesn't speak to me. I still think **user's notes** is better. 
Question : how is it called in Antares Web ?
We should also change the name of the source file (simulation.h is unrelated to what this class does)

flomnes: Renamed to `simulationComments`

src/libs/antares/study/study.cpp
flomnes: Yes, the idea is to dump the data into a file, then load the content into a buffer. I agree that this is not good, I'll change it.

flomnes:
@guilpier-code DONE

PR #805, opened 2022-08-17
Use generic writer to write simulation.log

flomnes:


src/libs/antares/study/importlogs.cpp
```diff
@@ -41,22 +41,17 @@ void Study::importLogsToOutputFolder() const
     if (!logs.logfile())
         return;
 
-    String buffer;
-    buffer.reserve(this->folderOutput.size() + 20);
+    String path;
```
guilpier-code: In the end, **path** contains a only file name.
so : 
- it should be renamed (for instance) **logFileName**
- it should be an **std::string**

And in these conditions, what is the point of block : 
```C++
    path.reserve(this->folderOutput.size() + 20);

    // Double check: Since this method might be call after some memory starvation,
    // we should make sure that we have a valid buffer
    if (!path.data())
    {
        logs.error() << "I/O Error: Impossible to create the folder '" << path << "'";
        return;
    }
```
?

src/libs/antares/study/importlogs.cpp
```diff
@@ -67,24 +62,28 @@ void Study::importLogsToOutputFolder() const
...
+    switch (IO::File::LoadFromFile(log_buffer, from))
     {
     case IO::errNone:
         break;
-    case IO::errOverwriteNotAllowed:
-        logs.error() << from << ": File already exists";
-        buffer.clear().shrink();
+    case IO::errNotFound:
+        logs.error() << from << ": file does not exist";
+        path.clear().shrink();
```
guilpier-code: After the previous remark, we should remove instructions `path.clear().shrink();`

guilpier-code (changes requested):

PR #807, opened 2022-08-18
Use generic writer to write MPS files, criterion, interco, etc.

flomnes:
- Remove `Application::pResultWriter`
- Add `Study::getWriter`
- In MPS-writing functions, replace `fprintf` by `IResultWriter::addJob` from a buffer.
- In function `Study::createFileIntoOutputWithExtension`, do not check for the existence of files. Instead, pass the resolution number (1st or 2nd resolution) to ensure that file names are unique, removing the need for this check.


src/libs/antares/study/mps.cpp
```diff
@@ -39,43 +39,25 @@ namespace Data
 {
 std::string Study::createFileIntoOutputWithExtension(const YString& prefix,
                                                      const YString& extension,
-                                                     uint numSpace) const
+                                                     uint numSpace,
+                                                     int n) const
```
guilpier-code: What is argument **n** ? We should give it a more explicit name. As it has a default value (=0) and is never given another value, why appending a "-0" in the file name.

src/libs/antares/study/mps.cpp
```diff
@@ -39,43 +39,25 @@ namespace Data
 {
 std::string Study::createFileIntoOutputWithExtension(const YString& prefix,
                                                      const YString& extension,
-                                                     uint numSpace) const
+                                                     uint numSpace,
```
guilpier-code: The function name is not related to what it does, we should rename it **createWeeklySimulationPath**.
The source file hasn't got a good name as well.

src/solver/utils/mps_utils.cpp
```diff
@@ -19,9 +19,9 @@ constexpr size_t OPT_APPEL_SOLVEUR_BUFFER_SIZE = 256;
 
 static void printHeader(Clob& Flot, int NombreDeVariables, int NombreDeContraintes)
```
guilpier-code: **Flot** does mean anything to me. Why not **buffer** or **memoire_tampon** (french)

src/solver/utils/mps_utils.cpp
```diff
@@ -246,14 +240,11 @@ void OPT_dump_spx_fixed_part(const PROBLEME_SIMPLEXE* Pb, uint numSpace)
...
-    // auto archive = study->pZipArchive;
-    // archive->addData(filename,
-    //                  Flot.c_str(),
-    //                  Flot.size());
-
-    // archive->close();
-    // archive->open(libzippp::ZipArchive::Write);
+    auto study = Data::Study::Current::Get();
+    const auto filename
+      = study->createFileIntoOutputWithExtension("problem-fixed-part", "mps", numSpace);
```
guilpier-code: As I said in **mps.cpp** review, prepending output folder was removed from the file path

src/solver/optimisation/opt_export_structure.cpp
```diff
@@ -66,14 +66,39 @@ void OPT_ExportInterco(const Antares::Data::Study& study,
                        PROBLEME_HEBDO* ProblemeHebdo,
                        uint numSpace)
 {
-  // TODO
+    // Interco are exported only once for first year
+    if (ProblemeHebdo->firstWeekOfSimulation)
+    {
+        Yuni::Clob Flot;
```
guilpier-code: **Flot** ==> **buffer** ? **memoire_tampon** ?

src/solver/simulation/solver.hxx
```diff
@@ -281,6 +281,8 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
         pYearByYear = false;
 
     pHydroHotStart = (study.parameters.initialReservoirLevels.iniLevels == Data::irlHotStart);
+
+    study.setWriter(pResultWriter);
```
guilpier-code: As said earlier review made on the "zip" matter, the **study->parameters** already has the info to create a writer, we should create it in the study after reading this info, and pass this writer to component that need it.

src/solver/utils/mps_utils.cpp
```diff
@@ -290,17 +276,14 @@ void OPT_dump_spx_variable_part(const PROBLEME_SIMPLEXE* Pb, uint numSpace)
...
-    // archive->open(libzippp::ZipArchive::Write);
+    auto study = Data::Study::Current::Get();
+    const auto filename
+      = study->createFileIntoOutputWithExtension("problem-variable-part", "mps", numSpace);
+    auto writer = study->getWriter();
+    writer->addJob(filename, Flot);
 }
 
-void OPT_EcrireJeuDeDonneesLineaireAuFormatMPS(void* Prob, uint numSpace)
+void OPT_EcrireJeuDeDonneesLineaireAuFormatMPS(void* Prob, uint numSpace, int n)
```
guilpier-code: Parameter **n** is unused (and should be renamed [**optimNumber** ?])

src/libs/antares/study/mps.cpp
```diff
@@ -39,43 +39,25 @@ namespace Data
...
+                                                     int n) const
 {
-    static std::map<YString, int> count;
-
     // Empty log entry
     logs.info();
 
-    // Preparing the filename where to write the MPS file
```
guilpier-code: As this function does not need to get the output folder from the study, it's not related to the study anymore and should not be a method of the study : it belongs more to the place it is used : for instance **mps_utils.h/cpp**

guilpier-code (changes requested):

src/solver/utils/mps_utils.cpp
flomnes: It's now used

src/solver/simulation/solver.hxx
flomnes: Will do it elsewhere

src/libs/antares/study/mps.cpp
flomnes: name was changed, more expressive

src/libs/antares/study/mps.cpp
flomnes: Renamed into `getFilenameWithExtension`.

flomnes:
@guilpier-code DONE

src/libs/antares/study/mps.cpp
flomnes: This will be obsolete with the MPS writer interface

guilpier-code (approved):

PR #809, opened 2022-08-18
Use generic writer for hydro debug files

flomnes:


src/solver/hydro/management/daily.cpp
```diff
@@ -251,8 +241,15 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
     std::shared_ptr<DebugData> debugData(nullptr);
 
     if (study.parameters.hydroDebug)
```
guilpier-code, line 243: Please put braces "{ }" after the **if** statement

src/solver/hydro/management/daily.cpp
```diff
@@ -112,104 +116,90 @@ struct DebugData
         CoutTotal.fill(0);
     }
 
-    void writeTurb(const String& folderOutput, const String& filename, uint y) const
+    void writeTurb(const String& filename, uint y) const
     {
-        String folder;
-        folder << folderOutput << SEP << "debug" << SEP << "solver" << SEP << (1 + y);
-        if (IO::Directory::Create(folder))
+        Yuni::Clob buffer;
```
guilpier-code: Why not trying to make **buffer** an **std::string** ?

guilpier-code (changes requested):

flomnes:
@guilpier-code DONE

src/solver/hydro/management/daily.cpp
```diff
@@ -112,104 +117,89 @@ struct DebugData
...
+                                * 100;
                 }
+                buffer << '\n';
+
+                dayMonth++;
             }
-            file.close();
         }
+        auto buffer_str = buffer.str();
+        pWriter->addJob(path.str(), buffer_str);
```
guilpier-code, line 202: why not directly : 
```C++
        pWriter->addJob(path.str(), buffer.str());
```
instead of : 
```C++
        auto buffer_str = buffer.str();
        pWriter->addJob(path.str(), buffer_str);
```
?

src/solver/hydro/management/monthly.cpp
```diff
@@ -244,60 +244,55 @@ void HydroManagement::prepareMonthlyOptimalGenerations(double* random_reservoir_
...
 #endif
-
-        if (study.parameters.hydroDebug)
+        auto writer = study.getWriter();
+        if (study.parameters.hydroDebug && writer)
         {
-            String folder;
-            folder << study.folderOutput << SEP << "debug" << SEP << "solver" << SEP << (1 + y);
-            if (IO::Directory::Create(folder))
+            Yuni::Clob file;
```
guilpier-code: Does **file** have to be a **Yuni::Clob** ? Or it can be a **std::string** as well ?
About about `String path;` ? **std::string path;** is better

src/solver/hydro/management/daily.cpp
```diff
@@ -417,7 +417,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
         {
             String filename;
             filename << "daily." << area.name << ".txt";
-            debugData->writeTurb(study.folderOutput, filename, y);
+            debugData->writeTurb(filename, y);
```
guilpier-code, line 419: If **filename** becomes a **std::string**, we could have : 
```C++
        std::string filename;
        filename << "daily." << area.to<std::string>() << ".txt";
        //  area.to<std::string>()  or  area.c_str()
        debugData->writeTurb(filename, y);
```

src/solver/hydro/management/daily.cpp
```diff
@@ -112,104 +117,89 @@ struct DebugData
         CoutTotal.fill(0);
     }
 
-    void writeTurb(const String& folderOutput, const String& filename, uint y) const
+    void writeTurb(const String& filename, uint y) const
```
guilpier-code: **String& filename** becomes **std::string& filename**

guilpier-code (changes requested):

src/solver/hydro/management/daily.cpp
flomnes: We would need `addJob(..., std::string&& content);`, which would complexify the code even more

flomnes:
@guilpier-code DONE

src/solver/hydro/management/daily.cpp
guilpier-code: Then why not have a **const** qualifier for 2nd parameter of **addJob**, like it's done for the 1st ? 
I guess the **content** is not modified by **addJob** anyway

PR #812, opened 2022-08-18
Binding Constr marginal price : handling synthesis results

guilpier-code:
See PR's title


src/solver/variable/economy/bindingConstraintsMarginalCost.h
```diff
@@ -44,7 +44,7 @@ struct VCardBindingConstMarginCost
     //! Caption
     static const char* Caption()
     {
-        return "MARG. COST by BC";
+        return "BC. MARG. COST";
```
flomnes, line 47: Since it's obvious from the filename that it's a binding constraint, I suggest "MARG. COST"

src/solver/variable/economy/bindingConstraintsMarginalCost.h
```diff
@@ -323,12 +323,42 @@ class BindingConstMarginCost
...
+    // Private methods
+    // ---------------
     std::string getBindConstraintCaption() const
     {
         return associatedBC_->name + " (" + associatedBC_->operatorType + ")";
     }
 
+    bool isCurrentOutputNonApplicable(int precision) const
+    {
+        return precision < pow(2, associatedBC_->type - 1);
```
flomnes:
```suggestion
        // Explanatory comment
        return precision < (1 << (associatedBC_->type - 1));
```

src/solver/variable/economy/bindingConstraintsMarginalCost.h
```diff
@@ -323,12 +323,42 @@ class BindingConstMarginCost
...
+    }
+
+    // Private data mambers
```
flomnes, line 364: Use dedicated keyword ?

src/solver/variable/info.h
```diff
@@ -575,6 +575,19 @@ struct VariableAccessor<ResultsT, Category::singleColumn /* The default */>
...
 
+    template<class VCardType>
+    static SurveyResults::CaptionType getCaption(SurveyResults& results)
+    {
+        if (VCardType::categoryDataLevel & Category::bindingConstraint)
+        {
+            // In case the current output variable is associated to a BC, its caption is
+            // already set in the output variable class.
+            return results.variableCaption;
+        }
```
flomnes: I don't have access to the code, however for thermal cluster details, the situation is identical : the column name depends on the object. We should check how it's done, maybe the implementation is more simple.

flomnes (changes requested):

src/solver/variable/info.h
flomnes: Ok, so there is a template
```cpp
template<class ResultsT, int ColumnCountT>
struct VariableAccessor
```
and 3 (partial) specializations
```cpp
template<class ResultsT>
struct VariableAccessor<ResultsT, Category::dynamicColumns>
// ...
template<class ResultsT>
struct VariableAccessor<ResultsT, Category::singleColumn /* The default */>
// ...
template<class ResultsT>
struct VariableAccessor<ResultsT, Category::noColumn>
```
Most functions have 4 implementations (1 in the generic template, 3 in specializations), most are `static` functions.

These functions attempt to use cluster names in the "multiple columns" specialization. 
```cpp
    174:                results.variableCaption = VCardT::Multiple::Caption(i);
    194:                results.variableCaption = VCardType::Multiple::Caption(i);
    214:                results.variableCaption = VCardType::Multiple::Caption(i);
    380:                results.variableCaption = thermal.clusters[i]->name();
    399:            results.variableCaption = thermal.clusters[idx]->name();
    405:            results.variableCaption = renewable.clusters[idx]->name();
    573:            results.variableCaption = VCardT::Caption();
    585:            return results.variableCaption;
    600:            results.variableCaption = getCaption<VCardType>(results);
    614:            results.variableCaption = VCardType::Caption();
```

I think we should also call `getCaption` in `BuildAnnualSurveyReport` (l. 614).

guilpier-code: Actually, **BuildAnnualSurveyReport** (in **info.h**) is called only when printing the year by year **district** report for an **area** output variable. As you know, the new output is not related to an area, this function is not called from the new output variable class.

src/solver/variable/economy/bindingConstraintsMarginalCost.h
guilpier-code: As far as I know, the function **Caption()** (in the case of our new output variable) is not used when writing reports.
It could be called (to be checked) form GUI to make the list of variables to be printed in the **Thematic trimming**.
In this case, we need to keep it that way to distinguish it from any other marginal price output variable.

src/solver/variable/economy/bindingConstraintsMarginalCost.h
guilpier-code: What do you mean ?

flomnes:
```suggestion
    private:
```

src/solver/variable/economy/bindingConstraintsMarginalCost.h
flomnes: Also check that `associatedBC_->type > 0`. 
```cpp
    enum Type
    {
        //! Unknown status
        typeUnknown = 0,
        //! The binding constraint has hourly values
        typeHourly,
        //! The binding constraint has daily values
        typeDaily,
        //! The binding constraint has weekly values
        typeWeekly,
        //! The maximum number of types
        typeMax
    };
```

src/solver/variable/economy/bindingConstraintsMarginalCost.h
guilpier-code: the private keyword is already there, but there is one **private** for data and methods

src/solver/variable/economy/bindingConstraintsMarginalCost.h
guilpier-code: Already done in : **src\libs\antares\study\constraint\constraint.cpp** : 760

PR #813, opened 2022-08-19
Document zip output feature

flomnes:


docs/reference-guide/13-file-format.md
```diff
@@ -1,6 +1,13 @@
 # Study format changes
 This is a list of all recent changes that came with new Antares Simulator features. The main goal of this document is to lower the costs of changing existing interfaces, both GUI and scripts.
 
+## v8.4.0
+### Input
+In file **settings/generaldata.ini**, in existing section `output`, add property `result-format` [str]. Default value = `files`. If this property is set to `zip`, all results are written into a single zip archive, instead of multiple files.
```
guilpier-code: Are you sure that the value **files** for item **result-format** is a good value name ?
Would **txt-files** not be better ? Or **full-txt-format** ?

guilpier-code (changes requested):

guilpier-code (approved):

PR #814, opened 2022-08-22
Use antares-deps to provide minizip-ng

flomnes:


guilpier-code:
I can't find branch **feature/minizip-use-deps**, it has been probably removed

flomnes:
> I can't find branch **feature/minizip-use-deps**, it has been probably removed

Yes, it was merged

PR #801, opened 2022-08-09
BC marginal price

guilpier-code:
This PR covers : 
- In the output static list, inserting  an output variable class associated to binding constraints.
- This PR aims at handling the year by year results, but not the synthesis results.
- N/A status in year by year results should be handled correctly : for a weekly BC, hourly results are N/A 

TODO
- [x] Make sure that new files are created & written
- [x] Handle constraints of type "< and >". Add direction as 2nd line ?
- [x] Synthesis

PR #819, opened 2022-08-22
Binding constraints : filtering each constraint results

guilpier-code:
Filtering results of marginal price related each binding constraints, depending on parameters in **input\bindingconstraints\bindingconstraints.ini**

For each binding constraint, we add 2 parameters (**filter-year-by-year** and **filter-synthesis**).

Example : 
```ini
[0]
name = my-first-bind-const
...
filter-year-by-year = hourly, daily, annual
filter-synthesis = daily, weekly, monthly
...

[1]
name = my-second-bind-const
...
```

### DIFF
--- a/src/solver/variable/economy/bindingConstraintsMarginalCost.h
+++ b/src/solver/variable/economy/bindingConstraintsMarginalCost.h
@@ -311,6 +311,9 @@ public:
       int precision /* printed results : hourly, daily, weekly, ...*/,
       unsigned int numSpace) const
     {
+        if (!(precision & associatedBC_->filterYearByYear_))
+            return;
+        
         // Initializing external pointer on current variable non applicable status
         results.isCurrentVarNA[0] = isCurrentOutputNonApplicable(precision);
 
@@ -330,6 +333,8 @@ public:
     {
         // Building syntheses results
         // ------------------------------
+        if (!(precision & associatedBC_->filterSynthesis_))
+            return;
 
         // And only if we match the current data level _and_ precision level
         if ((dataLevel & VCardType::categoryDataLevel) && (fileLevel & VCardType::categoryFileLevel)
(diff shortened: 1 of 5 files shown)


src/solver/variable/economy/bindingConstraintsMarginalCost.h
```diff
@@ -311,6 +311,9 @@ class BindingConstMarginCost
       int precision /* printed results : hourly, daily, weekly, ...*/,
       unsigned int numSpace) const
     {
+        if (!(precision & associatedBC_->filterYearByYear_))
```
flomnes, line 314:
```suggestion
        if (!associatedBC_ || !(precision & associatedBC_->yearByYearFilter()))
```

src/solver/variable/economy/bindingConstraintsMarginalCost.h
```diff
@@ -330,6 +333,8 @@ class BindingConstMarginCost
     {
         // Building syntheses results
         // ------------------------------
+        if (!(precision & associatedBC_->filterSynthesis_))
+            return;
```
flomnes, line 337:
```suggestion
        if (!associatedBC_ || !(precision & associatedBC_->synthesisFilter()))
            return;
```

flomnes (changes requested):

src/solver/variable/economy/bindingConstraintsMarginalCost.h
guilpier-code: With changes done on PR #801, we don't have to make this change

src/solver/variable/economy/bindingConstraintsMarginalCost.h
guilpier-code: Same answer as above

PR #823, opened 2022-08-23
BC marginal price and thematic trimming (select output for printing or not)

guilpier-code:
User can disable output related to BC marginal price)


flomnes (approved):
Nice !
image

PR #827, opened 2022-08-24
Finalize write upon receiving SIGINT / SIGTERM

flomnes:
This should ensure that a valid zip archive is produced even if the
user pushes "Ctrl+C".


src/solver/application.h
```diff
@@ -54,6 +54,8 @@ class Application : public Yuni::IEventObserver<Application, Yuni::Policy::Singl
 
     void writeExectutionInfo();
 
+    void installSignalHandlers() const;
```
guilpier-code, line 57: **installInterruptSignalHandlers()** ?

why **install** ? Rather **initialize** ?

guilpier-code: Why would we have to make  **installSignalHandlers** a method of class **Application** and make this class grow ?

src/solver/main.cpp
```diff
@@ -111,6 +111,9 @@ int main(int argc, char** argv)
     argv = AntaresGetUTF8Arguments(argc, argv);
 
     Antares::Solver::Application application;
+
+    application.installSignalHandlers();
```
guilpier-code: I guess initializing interruption signals whereas the study and the writer are not created yet is dangerous, as catching such a signal would lead to use them, right ?
Could we not postpone this initialization ?

src/solver/signal-handling/common.cpp
```diff
@@ -0,0 +1,33 @@
+#include <antares/logs.h>
+#include <antares/study.h>
+#include "common.h"
+
+using namespace Antares;
+
+static void finalizeWrite()
+{
+    auto study = Data::Study::Current::Get();
```
guilpier-code, line 9: Please read remarks on **application.h** first (we suppose here that initializing signals handling is not made inside class Application).

I'm not a fan of using the singleton to get the study. It hides the dependency to the study.
What we really need here is the writer (the study only gives the writer).
So I would define a class **SignalHanlder**, create an instance of it in **application::prepare(...)** (after study is created), passing the **study::writer** to its constructor.
**SignalHanlder** could even be a parent class, with children for handling signals on Windows and Linux; we would have a factory like : 
```C++
SignalHanlder signalHandlerFactory(iResultWriterPtr writer)
{
    #ifdef YUNI_OS_WINDOWS
    return winSignalHanlder(writer);
    #else
    return linuxSignalHanlder(writer);
    #endif
}

```
Just an idea, let's discuss it

src/solver/writer/zip_writer.cpp
```diff
@@ -70,12 +78,14 @@ ZipWriter::ZipWriter(std::shared_ptr<Yuni::Job::QueueService> qs,
...
 ZipWriter::~ZipWriter()
 {
-    mz_zip_writer_close(pZipHandle);
-    mz_zip_writer_delete(&pZipHandle);
+    this->finalize(false);
 }
 
 void ZipWriter::addJob(const std::string& entryPath, Yuni::Clob& entryContent)
 {
+    if (pState != ZipState::can_receive_data)
```
guilpier-code: We cannot make **addJob** a template method (with **entryContent**'s type as template parameter) because of polymorphism with parent class **IResultWriter**.
But what we could do is have a template private method : 
```C++
class ZipWriter : public IResultWriter
{
....
private:
    template<ContentType>
    void runJob(const std::string& entryPath, Yuni::Clob& entryContent);
}
....

template<class ContentType>
ZipWrite<ContentType>::runJob(const ContentT& entryPath, Yuni::Clob& entryContent)
{
    if (pState != ZipState::can_receive_data)
        return;

    EnsureQueueStartedIfNeeded ensureQueue(this, pQueueService);
    pQueueService->add(
      new ZipWriteJob<ContentType>(*this, entryPath, entryContent, pDurationCollector),
      Yuni::Job::priorityLow);
}
```
and call this template method inside **addJob** methods.
ok, we would have to specialize **runJob** for **IniFile** type...

src/solver/writer/zip_writer.cpp
```diff
@@ -34,6 +35,12 @@ ZipWriteJob<ContentT>::ZipWriteJob(ZipWriter& writer,
 template<class ContentT>
 void ZipWriteJob<ContentT>::onExecute()
 {
+    // Don't write data if finalize() has been called
+    if (pState != ZipState::can_receive_data)
```
guilpier-code, line 39: In case of ctrl + c, if the goal here is let the started jobs finish but prevent others to start, 
maybe the block `if (pState != ZipState::can_receive_data) {...}` is not necessary, as we have such block in the **ZipWriter::addJob(...)** ?

guilpier-code (changes requested):

src/solver/writer/zip_writer.cpp
flomnes: We want to prevent new jobs from being added to the queue. This job is already in the queue, `onExecute` is called when the job is, well, executed. If the job is executed after `finalize` has been called, it does nothing.

src/solver/writer/zip_writer.cpp
flomnes: templates don't work well with virtual methods, [see this](https://stackoverflow.com/a/7968213)

src/solver/main.cpp
flomnes: Ok, this method doesn't need to be in class `Application`.

src/solver/signal-handling/common.cpp
flomnes: However, we can't pass an `IWriter` object to the functions being called.

You can't provide parameters to signal-handling functions, see this [SO answer](https://stackoverflow.com/a/6970238). Also, signal handlers are global variables.

flomnes: As for the factory method, I disagree since we don't want `winSignalHanlder` compiled on Linux and so on.

flomnes (commented):
### Will do
- Take out `initializeSignalHandlers` from class `Application`

### Won't do

- Use templates in a virtual class (impossible)
- Factory method for signal handlers (need too many #ifdef...#endif)
- Avoid using `Study::Get` (impossible). The best I can do is to declare an `IResultWriter::Ptr` as a global variable, and get it from there.

flomnes:
@guilpier-code DONE
Using templatized function for addJob, see ae0fa71b

src/solver/writer/zip_writer.cpp
guilpier-code: We agree on that, my comment does not contradict yours

flomnes: I've done the requested change [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/827/commits/ae0fa71bf6a01b0e7766d462548cbeb1d2b5f976)

src/solver/writer/zip_writer.cpp
guilpier-code: ok

src/solver/signal-handling/common.cpp
guilpier-code: ok, so you don't mind initializing signal handler even if the study is not ready yet. If study is not created when user presses ctrl + c, Antares Simulator stops with an  error log message

guilpier-code (approved):

PR #838, opened 2022-08-26
Statistics: use std::mutex / std::atomic<> to prevent concurrent writes

flomnes:
`FileContent` and `optimizationStatistics` objects are shared among threads. Prevent concurrent writes to `int/map` by using `std::atomic/std::mutex`.

### DIFF
--- a/src/libs/antares/benchmarking/file_content.h
+++ b/src/libs/antares/benchmarking/file_content.h
@@ -6,28 +6,27 @@
 
 namespace Benchmarking
 {
-
 class FileContent
 {
 public:
     FileContent() = default;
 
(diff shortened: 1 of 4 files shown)


src/libs/antares/benchmarking/file_content.h
```diff
@@ -20,6 +21,7 @@ class FileContent
     void addDurationItem(const std::string& name, unsigned int duration, int nbCalls);
 
 private:
+    std::mutex pMutex;
```
a-zakir: may be renaming pMutex to pSectionsMutex would be clearer.

flomnes: sure

a-zakir (approved):

a-zakir (approved):

PR #851, opened 2022-09-02
Remove everything related to deprecated output file digest.txt

### LINKED ISSUE #840: Segfault when exporting digest, new variable "profit by plant" suspected
v8.3.1, BP21 variant

flomnes:
Since the digest file is deprecated, and sometimes causes segfaults, we decide to remove it completely.

Hopefully, it will also simplify the output module (**src/solver/variable**) and allow the dev team to work on higher-value topics.

### DIFF
--- a/src/solver/variable/area.h
+++ b/src/solver/variable/area.h
@@ -171,8 +171,6 @@ public:
                                  int precision,
                                  uint numSpace) const;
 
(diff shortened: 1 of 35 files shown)


flomnes:
Too many users use mc-all/digest.txt, we cannot remove it.

PR #852, opened 2022-09-05
Digest: ignore thermal, renewable clusters, binding constraints

### LINKED ISSUE #840: Segfault when exporting digest, new variable "profit by plant" suspected

flomnes:
Avoid out of bound indices.


src/solver/variable/storage/average.h
```diff
@@ -142,8 +142,12 @@ struct Average : public NextT
...
     void buildDigest(SurveyResults& report, int digestLevel, int dataLevel) const
     {
-        if ((dataLevel & Category::area || dataLevel & Category::setOfAreas)
-            && digestLevel & Category::digestAllYears)
+        const bool isCluster = (VCardT::categoryFileLevel & Category::de)
+                               || (VCardT::categoryFileLevel & Category::de_res);
+        const bool isBindingConstraint = VCardT::categoryFileLevel & Category::bc;
+        const bool isDigest = digestLevel & Category::digestAllYears;
+        if ((dataLevel & Category::area || dataLevel & Category::setOfAreas) && isDigest
+            && !isCluster && !isBindingConstraint)
```
guilpier-code, line 150: Suggestion : 
```C++
const bool isCluster = (VCardT::categoryFileLevel & Category::de)
                               || (VCardT::categoryFileLevel & Category::de_res);
const bool isBindingConstraint = VCardT::categoryFileLevel & Category::bc;

// Here we could add a comment on why we must avoid clusters and BCs
if (isCluster || isBindingConstraint)
    return;

const bool isDigest = digestLevel & Category::digestAllYears;
if ((dataLevel & Category::area || dataLevel & Category::setOfAreas) && isDigest)
{
    ...
}
```

flomnes: We want to call `NextType::buildDigest`, so we can't just `return`

guilpier-code: I see

guilpier-code (approved):

PR #853, opened 2022-09-05
Fix potential segfault with many BCs, refactor

flomnes:


src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -562,24 +563,40 @@ static inline void WriteIndexHeaderToFileDescriptor(int precisionLevel,
...
+      - Production
+      - NODU,
+      - NP Costs
+      - Net profit
+    */
+    const uint nbVariablesPerDetailRenewableCluster = 1;
+    // Production
+
+    const uint nbVariablesPerInequalityBindingConstraint = 1;
+    // Marginal price
```
flomnes, line 586: Determine at compilation ?

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -661,6 +663,19 @@ void StudyRuntimeInfos::initializeMaxClusters(const Study& study)
       = maxNumberOfClusters<CompareAreasByNumberOfClusters::renewable>(study);
 }
 
+void StudyRuntimeInfos::initializeInequalityBindingConstraints()
```
guilpier-code: Do we initialize, here ? I guess we rather store inequality binding constraints numbers, do we not ?
If so, I suggest we call this method **getInequalityBindingConstraintsNumbers()**. 
We do not need to store these numbers, so we just have to return them (vector<uint>)

src/solver/variable/bindConstraints.hxx
```diff
@@ -102,25 +100,10 @@ inline void BindingConstraints<NextT>::provideInformations(I& infos)
...
-        }
-    }
-    return bindConstGlobalNumbers;
-}
-
 template<class NextT>
 void BindingConstraints<NextT>::initializeFromStudy(Data::Study& study)
 {
-    std::vector<uint> InequalityBCnumbers = getInequalityBindConstraintGlobalNumbers(study);
+    const std::vector<uint>& InequalityBCnumbers = study.runtime->bindingConstraintGlobalNumbers;
```
guilpier-code: I'm not at ease with getting a data directly in the runtime class like this.
According to remarks made in source file **runtime.cpp**, we can either have : 
- **getInequalityBindConstraintGlobalNumbers**(study) computing the global numbers and return a vector of them (which it not costly with the vector move constructor)
-  **getInequalityBindConstraintGlobalNumbers**(study) could return a reference to the the global numbers as a data member of runtime class

src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -562,24 +563,43 @@ static inline void WriteIndexHeaderToFileDescriptor(int precisionLevel,
...
+
+    const auto max = [](uint a, uint b, uint c, uint d) { return std::max({a, b, c, d}); };
+
+    return max(maxVars,
+               static_cast<uint>(nbVariablesPerDetailThermalCluster
+                                 * runtime->maxThermalClustersForSingleArea),
+               static_cast<uint>(nbVariablesPerDetailRenewableCluster
+                                 * runtime->maxRenewableClustersForSingleArea),
+               static_cast<uint>(nbVariablesPerInequalityBindingConstraint
+                                 * runtime->bindingConstraintGlobalNumbers.size()));
```
guilpier-code: Same remark as in **bindConstraints.hxx**

guilpier-code (changes requested):

src/libs/antares/study/runtime/runtime.h
```diff
@@ -175,6 +179,9 @@ class StudyRuntimeInfos
     //! The maximum number of renewable clusters for each area
     size_t maxRenewableClustersForSingleArea = 0;
 
+    //! The global indices for inequality ('<' or '>') binding constraints
+    std::vector<uint> bindingConstraintGlobalNumbers;
```
guilpier-code: **bindingConstraintGlobalNumbers** seems unused

guilpier-code (changes requested):

flomnes:
@guilpier-code DONE

flomnes:
@guilpier-code DONE

guilpier-code (approved):

PR #863, opened 2022-09-15
MPS writer refactoring : defining an mps writer interface and consequences

guilpier-code:
Each way of printing MPS files (full historic, or-tools way, split format or none) is now isolated inside a different class. Each of these ways can be removed or changed without impacting the client code. A factory function contains the logic that leads to the creation of a particular MPS writer.


src/solver/utils/mps_utils.h
```diff
@@ -8,7 +8,104 @@ extern "C"
...
+		MPSolver* solver,
+		uint thread_number);
+
+	std::unique_ptr<I_MPS_writer> create();
+	std::unique_ptr<I_MPS_writer> createOnOptimizationError();
+
+private:
+	// Member functions...
+	std::unique_ptr<I_MPS_writer> createFullmpsWriter();
+	bool doWeExportMPS();
```
flomnes, line 100:
```suggestion
	bool doWeExportMPS() const;
```

flomnes (approved):

flomnes (changes requested):
Some tests still fail

PR #864, opened 2022-09-15
Move writer creation to Study

flomnes:
- Create the queue earlier (in `Study`)
- Create the writer earlier (in `Study::prepareWriter`)
- Pass both to `Application` as `shared_ptr`s.
- Remove setter/getter in `Study`, expose writer as a public member.

### DIFF
--- a/src/solver/simulation/solver.hxx
+++ b/src/solver/simulation/solver.hxx
@@ -39,11 +39,14 @@
 #include <antares/emergency.h>
 #include "../ts-generator/generator.h"
 #include <antares/memory/memory.h>
+#include <antares/exception/InitializationError.hpp>
 
 #include "../hydro/management.h" // Added for use of randomReservoirLevel(...)
 
 #include <yuni/core/system/suspend.h>
 #include <yuni/job/job.h>
+
+
 
 #define SEP Yuni::IO::Separator
 #define HYDRO_HOT_START 0
@@ -264,7 +267,8 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
  pHydroManagement(study),
  pFirstSetParallelWithAPerformedYearWasRun(false),
  pDurationCollector(duration_collector),
- pQueueService(study.pQueueService)
+ pQueueService(study.pQueueService),
+ pResultWriter(study.resultWriter)
 {
     // Ask to the interface to show the messages
     logs.info();
@@ -279,8 +283,15 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
 
     pHydroHotStart = (study.parameters.initialReservoirLevels.iniLevels == Data::irlHotStart);
 
-    // Result writer
-    pResultWriter = study.resultWriter;
+    if (!pQueueService)
+    {
+        throw Solver::Initialization::Error::NoQueueService();
+    }
+
+    if (!pResultWriter)
+    {
+        throw Solver::Initialization::Error::NoResultWriter();
+    }
 }
 
 template<class Impl>
--- a/src/solver/simulation/solver.hxx
+++ b/src/solver/simulation/solver.hxx
@@ -264,7 +264,7 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
  pHydroManagement(study),
  pFirstSetParallelWithAPerformedYearWasRun(false),
  pDurationCollector(duration_collector),
- pQueueService(std::make_shared<Yuni::Job::QueueService>())
+ pQueueService(study.pQueueService)
 {
     // Ask to the interface to show the messages
     logs.info();
@@ -279,8 +279,7 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
 
     pHydroHotStart = (study.parameters.initialReservoirLevels.iniLevels == Data::irlHotStart);
 
-    // Initialize writer
-    study.resultWriter = resultWriterFactory(study.parameters.resultFormat, study.folderOutput, pQueueService, duration_collector);
+    // Result writer
     pResultWriter = study.resultWriter;
 }
 
--- a/src/solver/simulation/solver.hxx
+++ b/src/solver/simulation/solver.hxx
@@ -264,9 +264,8 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
  pHydroManagement(study),
  pFirstSetParallelWithAPerformedYearWasRun(false),
  pDurationCollector(duration_collector),
- qs(std::make_shared<Yuni::Job::QueueService>()),
- pResultWriter(
-   resultWriterFactory(study.parameters.resultFormat, study.folderOutput, qs, duration_collector))
+ pQueueService(std::make_shared<Yuni::Job::QueueService>()),
+ pResultWriter(study.resultWriter)
 {
     // Ask to the interface to show the messages
     logs.info();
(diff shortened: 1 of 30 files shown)


src/solver/simulation/solver.hxx
```diff
@@ -281,7 +279,9 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
 
     pHydroHotStart = (study.parameters.initialReservoirLevels.iniLevels == Data::irlHotStart);
 
-    study.setWriter(pResultWriter);
+    // Initialize writer
+    study.resultWriter = resultWriterFactory(study.parameters.resultFormat, study.folderOutput, pQueueService, duration_collector);
+    pResultWriter = study.resultWriter;
```
flomnes: When `study` is instantiated, `pQueueService` doesn't exist yet since it is instantiated by `simulation` object of type `ISimulation<Impl>` (e.g **src/solver/main/economy.cpp**)

src/solver/simulation/solver.hxx
```diff
@@ -281,7 +279,8 @@ inline ISimulation<Impl>::ISimulation(Data::Study& study,
 
     pHydroHotStart = (study.parameters.initialReservoirLevels.iniLevels == Data::irlHotStart);
 
-    study.setWriter(pResultWriter);
+    // Result writer
+    pResultWriter = study.resultWriter;
```
guilpier-code: Why don't we keep initialization of **pResultWriter** in the initialization list, like **pQueueService** ?

src/solver/simulation/solver.hxx
```diff
@@ -1042,6 +1041,11 @@ void ISimulation<Impl>::regenerateTimeSeries(uint year)
     // * The option "Preprocessor" is checked in the interface _and_ year == 0
     // * Both options "Preprocessor" and "Refresh" are checked in the interface
     //   _and_ the refresh must be done for the given year (always done for the first year).
+    if (!pResultWriter)
```
guilpier-code: There are few tests on nullity of **pResultWriter** before running an action inside class **ISimulation<Impl>**.
Could we test nullity of **pResultWriter** once for all when initializing it (here in the constructor of **ISimulation<Impl>**) ?

On the contrary, **pQueueService** (which is also a smart pointer) is never tested, is that intended ?

src/solver/utils/mps_utils.cpp
```diff
@@ -296,7 +296,7 @@ void OPT_dump_spx_fixed_part(const PROBLEME_SIMPLEXE* Pb, uint numSpace)
 
     auto study = Data::Study::Current::Get();
     const auto filename = getFilenameWithExtension("problem-fixed-part", "mps", numSpace);
-    auto writer = study->getWriter();
+    auto writer = study->resultWriter;
```
guilpier-code: Here, we are in **OPT_dump_spx_fixed_part** and this function needs the writer. So, for me, instead of getting the study by the **Get()** function and then get the study's writer, we should pass the writer to the function.
It's seems logical (as the function needs the writer) and more readable. Furthermore, we avoid to pull the whole study to get the reader only. 
By doing this, we only move a problem : the study will still have to be pulled at a higher level (when we call **OPT_dump_spx_fixed_part**).
If we do that as soon as we can with other functions or classes, it will be easier (when the time comes) to break the study god class into pieces.

guilpier-code (changes requested):

src/solver/simulation/solver.hxx
flomnes: Both will be tested at construction, throwing an exception in case of failure

src/solver/simulation/solver.hxx
flomnes: done

src/solver/utils/mps_utils.cpp
flomnes: I agree. Though we'd have to pass the writer to about 4-5 functions

flomnes: `getFilenameWithExtension` needs `study->runtime` for context info (week/year).

flomnes: Could we use your factory to pass the writer ?

guilpier-code: > `getFilenameWithExtension` needs `study->runtime` for context info (week/year).

**y** (year) and **w** (week) should come from above : they should be parameters of **getFilenameWithExtension**, 
**getFilenameWithExtension** should not pull the study.
As soon as we know about **y** and **w**, we should give them to the entity that need them (for example a mps writer, as soon as merges allow it).

Functions that call **getFilenameWithExtension** (for example **OPT_dump_spx_variable_part**) make 3 things : 
1. Build the file name.
2. Build the content to be printed
3. Print the file.

There should be 3 functions, or at least 2, to separate building filename form the rest.
To enhance the code, my suggestion would be something like : 
```C++
void OPT_dump_spx_variable_part(const PROBLEME_SIMPLEXE* Pb, int y, int w, opt_number)
{
    std::string outputFileName = buildOutputFileName("problem-fixed-part", "mps", y, w, opt_number);
    std::string fileContent = build_spx_variable_part(Pb);
    printFile(outputFileName, fileContent);
}
// "buildOutputFileName" is a better name that "getFilenameWithExtension" in my opinion
```
As you can see, parameter **numSpace** disappeared from **OPT_dump_spx_variable_part**'s parameters list.

Furthermore, we should think about removing : `study->runtime->currentYear` and `study->runtime->weekInTheYear`, they exist only because we don't want to pass them as argument to functions that need them.

guilpier-code: > Could we use your factory to pass the writer ?

What do you mean ?

flomnes: We could pass the writer the MPS writer factory, but I'm not sure it would help much

src/solver/simulation/solver.hxx
```diff
@@ -461,7 +469,7 @@ void ISimulation<Impl>::writeResults(bool synthesis, uint year, uint numSpace)
         }
 
         // Dumping
-        if (IO::Directory::Create(newPath))
+        if (IO::Directory::Create(newPath) && pResultWriter)
```
guilpier-code: As we know **pResultWriter** is not null (checked in **ISimulation<Impl>**'s constructor), do we need to test it here ? Shouldn't we only have : 
```C++
if (IO::Directory::Create(newPath))
{
    ...
}
```
?

src/solver/simulation/solver.hxx
```diff
@@ -1550,7 +1557,7 @@ void ISimulation<Impl>::loopThroughYears(uint firstYear,
     {
         int numThreads = pNbMaxPerformedYearsInParallel;
         // If the result writer uses the job queue, add one more thread for it
-        if (pResultWriter->needsTheJobQueue())
+        if (pResultWriter && pResultWriter->needsTheJobQueue())
```
guilpier-code: Same as above : we **pResultWriter** is not null : 
```C++
if (pResultWriter->needsTheJobQueue())
{
   ...
}
```
should be enough.

src/solver/simulation/solver.hxx
```diff
@@ -1683,7 +1690,7 @@ void ISimulation<Impl>::loopThroughYears(uint firstYear,
     } // End loop over sets of parallel years
 
     // Writing annual costs statistics
-    if (not study.parameters.adequacyDraft())
+    if (not study.parameters.adequacyDraft() && pResultWriter)
```
guilpier-code: Same as above

guilpier-code (changes requested):

PR #866, opened 2022-09-16
[DEV] Refactored Layers by adding a LayerData class

### LINKED ISSUE #862: Extract layers, activeLayerID, showAllLayer from class Study
Create a dedicated class (see **src/libs/antares/study/study.h**)
### Current design
```cpp
class Study final : public Yuni::NonCopyable<Study>, public IObject
{
public:    //! \name Layers
    //@{
    //! All available layers
    std::map<size_t, std::string> layers;
    //@}
    size_t activeLayerID;
    bool showAllLayer;
    // A lot of stuff
private:
    //! Load all layers
    bool saveLayers(const AnyString& filename);
    void loadLayers(const AnyString& filename);
};
```

### Proposed design
```cpp
class LayerData {
    //! \name Layers
    //@{
    //! All available layers
    std::map<size_t, std::string> layers;
    //@}
    size_t activeLayerID;
    bool showAllLayer;
protected:
    bool saveLayers(const AnyString& filename);
    void loadLayers(const AnyString& filename);
};
// ...
class Study : public LayerData {
// ...
};

### PR DESCRIPTION
**Pull request by payetvin**
Added a LayerData class that the Study class inherits from
#862

### DIFF
--- a/src/libs/antares/study/layerdata.cpp
+++ b/src/libs/antares/study/layerdata.cpp
@@ -36,9 +36,7 @@
 
 using namespace Yuni;
 
-namespace Antares
-{
-namespace Data
+namespace Antares::Data
 {
 
 void LayerData::loadLayers(const AnyString& filename)
@@ -114,5 +112,4 @@ bool LayerData::saveLayers(const AnyString& filename)
     return false;
 }
 
-} // namespace Data
-} // namespace Antares
+} // namespace Antares::Data
--- a/src/libs/antares/study/layerdata.cpp
+++ b/src/libs/antares/study/layerdata.cpp
@@ -85,11 +85,9 @@ bool LayerData::saveLayers(const AnyString& filename)
     {
         CString<256, true> data;
         data << "[layers]\n";
-        for (auto iterator = layers.begin(); iterator != layers.end(); iterator++)
-        {
-            data << iterator->first << " = " << iterator->second;
-            data << '\n';
-        }
+
+        for (const auto& [key, value] : layers)
+            data << key << " = " << value << '\n';
 
         data << "[activeLayer]\n";
         data << "activeLayerID"
--- a/src/libs/antares/study/layerdata.cpp
+++ b/src/libs/antares/study/layerdata.cpp
@@ -42,55 +42,50 @@ namespace Antares::Data
 void LayerData::loadLayers(const AnyString& filename)
 {
     IniFile ini;
-    if (std::ifstream(filename.c_str()).good()) // check if file exists
-        if (ini.open(filename))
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/libs/antares/study/load.cpp [hunk 1]
```diff
@@ -456,7 +456,7 @@ bool Study::reloadXCastData()
     return ret;
 }
 
-void Study::loadLayers(const AnyString& filename)
+void LayerData::loadLayers(const AnyString& filename)
 {
     IniFile ini;
     if (std::ifstream(filename.c_str()).good()) // check if file exists
```
flomnes: Move to new source/header file

src/libs/antares/study/study.cpp
```diff
@@ -70,6 +70,8 @@ static inline void FreeAndNil(T*& pointer)
 }
 
 Study::Study(bool forTheSolver) :
+ //LayerData class constructor
+ LayerData(0, true),
```
flomnes: Remove useless comment

flomnes (changes requested):

src/libs/antares/study/layerdata.h
```diff
@@ -0,0 +1,63 @@
...
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#ifndef __ANTARES_LIBS_STUDY_LAYERDATA_H__
+#define __ANTARES_LIBS_STUDY_LAYERDATA_H__
+
+#include <map>
+
```
flomnes, line 32: Also include `string` for `std::string`

flomnes (changes requested):
LGTM, add missing header and we're good

src/libs/antares/study/layerdata.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+bool LayerData::saveLayers(const AnyString& filename)
+{
+    if (IO::File::Stream file; file.openRW(filename))
+    {
+        CString<256, true> data;
+        data << "[layers]\n";
+        for (auto iterator = layers.begin(); iterator != layers.end(); iterator++)
+        {
+            data << iterator->first << " = " << iterator->second;
```
flomnes: Use structured bindings from C++17
```suggestion
        for (const auto& [key, value] : layers)
        {
            data << key << " = " << value << '\n';
        }
```

flomnes (changes requested):
All good, just one last detail

flomnes (approved):

PR #872, opened 2022-09-21
MPS files printing : restore print of 1st and 2nd optimization, whether split or not  

guilpier-code:
The goal is to enrich the set of choices the user can make regarding the MPS export : 
Before this current PR, the user can only ask to print : 
- no MPS files 
- MPS files from 2nd optimization (1st optimization are overridden), in full or split format.

The user should be able to choose to print either : 
- no MPS
- MPS from 1st optimization
- MPS from 2nd optimization
- All optimizations MPS files

Note that, since the work is done, the **objective function** files (**criterion....txt**) are printed for every optimization as soon as we do  not ask for **no MPS** (that is for the 3 other choices).


src/solver/utils/mps_utils.h
```diff
@@ -36,10 +36,11 @@ class I_MPS_writer
 class fullMPSwriter final : public I_MPS_writer
 {
 public:
-	fullMPSwriter(PROBLEME_SIMPLEXE_NOMME* named_splx_problem, uint thread_number);
+	fullMPSwriter(PROBLEME_SIMPLEXE_NOMME* named_splx_problem, int currentOptimNumber, uint thread_number);
```
guilpier-code, line 39: Some data members are common to all MPS writers but the null writer.
Do we create a class **MPSwriter** containing the common data members ?

src/libs/antares/study/fwd.cpp
```diff
@@ -400,5 +400,62 @@ const char* DayAheadReserveManagementModeToCString(DayAheadReserveManagement daR
     return "";
 }
 
+
+std::string mpsExportStatusToString(const mpsExportStatus& mps_export_status)
```
flomnes, line 404:
```suggestion
std::pair<bool, std::string> mpsExportStatusToString(const mpsExportStatus& mps_export_status)
```

src/libs/antares/study/fwd.h
```diff
@@ -540,9 +540,28 @@ const char* DayAheadReserveManagementModeToCString(DayAheadReserveManagement daR
...
+    NO_EXPORT = 0,
+    EXPORT_FIRST_OPIM = 1,
+    EXPORT_SECOND_OPIM = 2,
+    EXPORT_BOTH_OPTIMS = 3,
+    UNKNOWN_EXPORT = 4
+};
+
+std::string mpsExportStatusToString(const mpsExportStatus& mps_export_status);
+mpsExportStatus stringToMPSexportStatus(const AnyString& value);
+const char* mpsExportIcon(const mpsExportStatus& mps_export_status);
```
flomnes: Move the last function to **src/ui** ?

src/ui/simulator/windows/options/optimization/optimization.cpp
```diff
@@ -729,6 +734,75 @@ void Optimization::onPopupMenuTransmissionCapacities(Component::Button&, wxMenu&
...
+                        wxID_ANY,
+                        mpsExportStatusToString(value),
+                        mpsExportIcon(value),
+                        wxEmptyString);
+        menu.Connect(it->GetId(),
+            wxEVT_COMMAND_MENU_SELECTED,
+            wxCommandEventHandler(Optimization::onSelectExportMPSbothOptimizations),
+            nullptr,
+            this);
+    }
```
flomnes: for loop ? Like here https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/832

src/ui/simulator/windows/options/optimization/optimization.cpp
```diff
@@ -923,6 +997,46 @@ void Optimization::onSelectLinkTypeAC(wxCommandEvent&)
...
+
+void Optimization::onSelectExportMPSsecondOptimization(wxCommandEvent&)
+{
+    onSelectExportMPS(Data::mpsExportStatus::EXPORT_SECOND_OPIM);
+}
+
+void Optimization::onSelectExportMPSbothOptimizations(wxCommandEvent&)
+{
+    onSelectExportMPS(Data::mpsExportStatus::EXPORT_BOTH_OPTIMS);
+}
```
flomnes: Templatize this ?

src/solver/utils/mps_utils.h
flomnes: Not really required

src/solver/utils/mps_utils.cpp
```diff
@@ -514,16 +521,18 @@ mpsWriterFactory::mpsWriterFactory(
...
 bool mpsWriterFactory::doWeExportMPS()
 {
-    // Current optimization number is not involved yet in the return value
-    // but will be soon
-    return export_mps_;
+    if (export_mps_ == Data::mpsExportStatus::EXPORT_BOTH_OPTIMS)
+        return true;
+    if ((int)export_mps_ == current_optim_number_)
+        return true;
+    return false;
```
flomnes: Return false if NO_EXPORT

flomnes (changes requested):

src/solver/utils/mps_utils.cpp
flomnes: Also, avoid converting `enums` to `int`, expand this `if (export_mps_ == FIST_OPTIM && current_optim_number_ == 1)` etc.

flomnes (changes requested):
Note : split MPS + OR-Tools are incompatible, check these options at the start

src/libs/antares/study/fwd.cpp
guilpier-code: Could you be more specific ?

src/libs/antares/study/fwd.h
guilpier-code: Not sure we get a great benefit to separate declaration / definition of **mpsExportIcon** from the rest of the MPS stuff in fwd.h/.cpp.
We could place **mpsExportIcon** closer to the GUI, but it does not use wxWidget function, so ti's more a "utils" function. 
What we could do is (as it is done for unfeasible problem behavior) is regroup everything about MPS in mps_utils.h/.cpp (I tried this but I met problems so I gave up), or even make a little separate lib (**mps.lib**) to be linked wherever it's needed.

flomnes: I definitely that anything related to the GUI should be in **src/ui**, so for me it would be best to define `mpsExportIcon` as `static` in the source (optimization.cpp) file where it is called. The 2 other functions should be declared/defined in **parameters.h/.cpp**

src/solver/utils/mps_utils.cpp
guilpier-code: > Return false if NO_EXPORT

What for ? If **NO_EXPORT**, this function returns already **false**

guilpier-code: > Also, avoid converting `enums` to `int`, expand this `if (export_mps_ == FIST_OPTIM && current_optim_number_ == 1)` etc.

To convert an **enum class** value into an **int**, I now use `static_cast<int>(...)`, safer

src/ui/simulator/windows/options/optimization/optimization.cpp
guilpier-code: Done. It's better templated

src/ui/simulator/windows/options/optimization/optimization.cpp
guilpier-code: I did not introduce a loop here, but a template function makes thing more clear and avoids duplicated code

src/solver/utils/mps_utils.cpp
flomnes: It would be best not to convert

src/ui/simulator/windows/options/optimization/optimization.cpp
flomnes: Ok, much better indeed

src/libs/antares/study/fwd.cpp
flomnes: The UNKNOWN enum value is only required to return an error in case the conversion wasn't successful. We can eliminate this value using an error-handling mechanism

docs/reference-guide/13-file-format.md
```diff
@@ -2,14 +2,27 @@
...
 
 ## v8.4.0
-### Input
+### **Writing mps files**
+mps files of first optimization used to be overwritten by mps files of second optimization. Not anymore.
+Now user can choose to print :
+- no mps file,
+- mps files related to the first opimization,
+- mps files related to the second opimization,
+- mps files related to the both opimizations.
```
flomnes:
```suggestion
- no MPS file,
- MPS files related to the first opimization,
- MPS files related to the second opimization,
- MPS files related to the both opimizations.
```

docs/reference-guide/13-file-format.md
```diff
@@ -2,14 +2,27 @@
...
+
+In the **generaldata.ini** input file, corresponding values for **include-exportmps** are : **none**, **optim-1**, **optim-2**, **both-optims**.
+
+Values **none** and **both-optims** can be replaced respectively by **false** and **true**, so that Antares Simulator **v8.4.0** can read studies of older versions.
```
flomnes: Compatibility with existing values is ensured (true = both-optims, false=none)

docs/reference-guide/13-file-format.md
```diff
@@ -2,14 +2,27 @@
...
+### **Marginal price of a binding constraint**
+**Input** :
 In the context of the addition of a new output variable (marginal price associated to a binding constraint), file **input/bindingconstraints/bindingconstraints.ini** get 2 new parameters for each binding constraint.
 They constrol which marginal price time granularity is printed, either regarding year by year or synthesis results.
 
 * `filter-year-by-year`. Default value = hourly, daily, weekly, monthly, annual
 * `filter-synthesis`. Default value = hourly, daily, weekly, monthly, annual
 
-### Output
+**Output** :
```
flomnes: No need to make titles bold

```md

# Title 1

## Title 2

### Title 3

#### Title 4
```

# Title 1

## Title 2

### Title 3

#### Title 4

docs/reference-guide/13-file-format.md
```diff
@@ -2,14 +2,27 @@
 This is a list of all recent changes that came with new Antares Simulator features. The main goal of this document is to lower the costs of changing existing interfaces, both GUI and scripts.
 
 ## v8.4.0
-### Input
+### **Writing mps files**
```
flomnes: No need for bold titles

flomnes (changes requested):

src/libs/antares/study/fwd.cpp
flomnes: See also std::optional

guilpier-code: **mpsExportStatusToString** converts the status to a string, and its return value is based on its argument, which is necessarily in the range of values of the enum. So I guess we don't need this default returned value : it is useless.
We can either leave it as is, or return an empty string. What do you think ?

docs/reference-guide/13-file-format.md
guilpier-code: Done

docs/reference-guide/13-file-format.md
guilpier-code: Done

src/libs/antares/study/fwd.h
guilpier-code: I moved **mpsExportIcon** into **src/ui**, where it is used. I didn't touch the 2 others functions for now to conform to the way it's done for other parameters.

docs/reference-guide/13-file-format.md
guilpier-code: Done

docs/reference-guide/13-file-format.md
guilpier-code: Done

flomnes (approved):

PR #881, opened 2022-09-26
Use average for variable MARG. COST (links)

flomnes:


src/solver/variable/economy/links/marginalCost.h
```diff
@@ -79,7 +79,7 @@ struct VCardMarginalCost
         //! Number of columns used by the variable (One ResultsType per column)
         columnCount = 1,
         //! The Spatial aggregation
-        spatialAggregate = Category::spatialAggregateSum,
+        spatialAggregate = Category::spatialAggregateAverage,
```
guilpier-code: Are you sure about that ? To me, this change is about districts : you change the way that area variables are aggregated in a district, it seems. As a matter of fact, this change should not have any effect, as the current variable is related to links. 

Maybe more something like that ? : 
```C++
void yearEnd(unsigned int year, unsigned int numSpace)
{
    // Here is the change I would try
    pValuesForTheCurrentYear[numSpace].computeAVGstatisticsForCurrentYear();

    NextType::yearEnd(year, numSpace);
}
```

guilpier-code (changes requested):

src/solver/variable/economy/links/marginalCost.h
flomnes: @guilpier-code Tested & approved ! DONE

guilpier-code (approved):

PR #892, opened 2022-10-04
No or empty generaldata.ini

### LINKED ISSUE #848: Empty generaldata.ini leads to an assertion error
```
> antares-8.3-solver study
[Wed Aug 31 15:15:18 2022][solver][infos]   simulation mode: Economy
[Wed Aug 31 15:15:18 2022][solver][infos]   simplex optimization range: week
/home/omnesflo/Projects/Antares_Simulator/src/libs/antares/study/parameters.cpp:1373: void Antares::Data::Parameters::prepareForSimulation(const Antares::Data::StudyLoadOptions&): Assertion `yearsFilter && "The array yearsFilter must be valid at this point"' failed.
Abandon (core dumped)
```

Default values exist for each key. So providing an empty generaldata.ini should work : the application should use these default values. Instead, a segfault occurs.

guilpier-code:
Refers to ticket #848


src/libs/antares/study/parameters.cpp
```diff
@@ -205,6 +207,18 @@ void Parameters::resetAdqPatchParameters()
...
+    // Set the number of years
+    nbYears = std::min(nbOfYears, maximumMCYears);
+
+    // Reset the filter
+    delete[] yearsFilter;
+    yearsFilter = new bool[nbYears];
+    for (uint i = 0; i != nbYears; ++i)
+        yearsFilter[i] = true;
+}
+
```
flomnes: How about creating a dedicated class here ? With a dedicated API, so that we could add unit tests

```cpp
class Playlist {
public:
  void initialize(uint nbYears);
  bool setYearPlayed(uint y, bool played);
  bool isYearPlayed(uint y) const;
private:
  std::vector<bool> mYears;
};

**Review by flomnes — APPROVED** [review]

**Review by flomnes — CHANGES_REQUESTED** [review]
Add tests

File: src/libs/antares/study/parameters.cpp [hunk 2]
```diff
@@ -49,11 +51,10 @@ namespace Antares
...
 {
-enum
-{
-    //! Hard coded maximum number of MC years
-    maximumMCYears = 100000,
-};
+
+//! Hard coded maximum number of MC years
+const uint maximumMCYears = 100000;
+
```
flomnes: We should check the number of MC years as an input from the user and assume it's valid later.

flomnes:
See #1043

flomnes (approved):

PR #893, opened 2022-10-06
[TEST] Refactoring study's loader

kathvargasr:


src/libs/antares/study/load-params.cpp
```diff
@@ -0,0 +1,185 @@
...
+    return true;
+
+}
+
+class SetHandlerAreas
+{
+public:
+    SetHandlerAreas(Study& study) : pStudy(study)
+    {
+    }
```
flomnes: This class only uses `study.areas`. I recommend saving a reference to this object instead.

src/libs/antares/study/load-params.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    void clear(Study::SingleSetOfAreas& set)
+    {
+        set.clear();
+    }
+
+    uint size(Study::SingleSetOfAreas& set)
+    {
+        return (uint)set.size();
+    }
```
flomnes: mark as `const`

src/libs/antares/study/study.cpp
```diff
@@ -320,10 +322,10 @@ void Study::ensureDataAreAllInitialized()
     StudyEnsureDataThermalPrepro(this);
 }
 
-std::map<std::string, uint> Study::getRawNumberCoresPerLevel()
+
+void Study::computeRawNbParallelYear()
 {
     std::map<std::string, uint> table;
```
flomnes: - No need for `std::strings` in the `std::map`, just use `Antares::Data::NumberOfCoresMode` as a key, so at line 437, you can use `table.at(parameters.nbCores.ncMode)`. Don't forget to `catch` exception [`std::out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range)
- `table["min"]` is always equal to `1`, so you may place it before the `switch..case` block
- Move line 331 (`logs.fatal() << ...`) to `case 0` in the `switch..case` block

src/libs/antares/study/study.cpp
```diff
@@ -479,13 +477,13 @@ void Study::getNumberOfCores(const bool forceParallel, const uint nbYearsParalle
...
+}
 
-    if (TSlimit < maxNbYearsInParallel)
-        maxNbYearsInParallel = TSlimit;
+std::vector<std::vector<uint>> Study::computeMinNbYearsInParallelYearSet(){
 
-    // Limiting the number of parallel years by the total number of years
-    if (p.nbYears < maxNbYearsInParallel)
-        maxNbYearsInParallel = p.nbYears;
+    auto& p = parameters;
```
flomnes: `const`

src/libs/antares/study/study.cpp
```diff
@@ -547,6 +545,63 @@ void Study::getNumberOfCores(const bool forceParallel, const uint nbYearsParalle
...
+    return setsOfParallelYears;
+
+}
+
+void Study::setGUIMinMaxNbYearsInParallel(const uint min, const uint max){
+    minNbYearsInParallel_save = min;
+    maxNbYearsInParallel_save = max;
+}
+
+bool Study::allSetsParallelYearsHaveSameSize(const std::vector<std::vector<uint>> setsOfParallelYears){
```
flomnes:
```suggestion
bool Study::allSetsParallelYearsHaveSameSize(const std::vector<std::vector<uint>>& setsOfParallelYears){
```

src/libs/antares/study/study.cpp
```diff
@@ -547,6 +545,63 @@ void Study::getNumberOfCores(const bool forceParallel, const uint nbYearsParalle
...
+
+    if (parameters.initialReservoirLevels.iniLevels == Antares::Data::irlHotStart
+        && setsOfParallelYears.size() && maxNbYearsInParallel > 1)
+    {
```
flomnes:
```suggestion
    if (parameters.initialReservoirLevels.iniLevels == Antares::Data::irlHotStart
        && !setsOfParallelYears.empty() && maxNbYearsInParallel > 1)
    {
```

src/libs/antares/study/study.cpp
```diff
@@ -547,6 +545,63 @@ void Study::getNumberOfCores(const bool forceParallel, const uint nbYearsParalle
...
+        uint currentSetSize = (uint)setsOfParallelYears[0].size();
+        if (setsOfParallelYears.size() > 1)
```
flomnes: This condition is already checked at line 560

src/libs/antares/study/study.cpp
```diff
@@ -547,6 +545,63 @@ void Study::getNumberOfCores(const bool forceParallel, const uint nbYearsParalle
...
+            This number is limited by the smallest refresh span (if at least
+            one type of time series is generated)
+    */
+    computeRawNbParallelYear();
+
+    maxNbYearsInParallel = getNbYearsParallelRaw();
+
+    // In case solver option '--force-parallel n' is used, previous computation is overridden.
+    if (forceParallel)
+        maxNbYearsInParallel = nbYearsParallelForced;
```
flomnes: Don't call `getNbYearsParallelRaw` in this case, since its results are overridden ?

src/libs/antares/study/study.cpp
```diff
@@ -562,7 +617,7 @@ void Study::getNumberOfCores(const bool forceParallel, const uint nbYearsParalle
     // GUI : storing minimum number of parallel years (in a set of parallel years).
     //		 Useful in the run window's simulation cores field in case parallel mode is enabled
     // by user.
-    minNbYearsInParallel_save = minNbYearsInParallel;
+    // minNbYearsInParallel_save = minNbYearsInParallel;
```
flomnes: Remove commented-out code

src/tests/src/libs/antares/study/loader/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+target_include_directories(test-study-load-internal
+								PRIVATE
+									"${src_libs_antares_study}"
+)
+target_link_libraries(test-study-load-internal
+						PRIVATE
+							Boost::unit_test_framework
+							libantares-core
+							libmodel_antares
+)
```
flomnes: Fix formatting here

src/tests/src/libs/antares/study/loader/test-study-load-internal.cpp
```diff
@@ -0,0 +1,105 @@
...
+ 	
+#include <fstream>
+#include <string>
+#include <filesystem>
+
+#include <fwd.h>
+#include <study.h>
+#include <area/constants.h>
+#include "../wrappers/study-wrapper.h"
+// #include "files-generator.h"
```
flomnes: Remove this commented-out line

flomnes (changes requested):
- Function `Study::internalLoadFromFolder` is still way too long, consider cutting it into smaller functions

src/tests/src/libs/antares/study/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 set(src_tests_src_libs_antares_study "${CMAKE_CURRENT_SOURCE_DIR}")
 
 add_subdirectory(area)
-add_subdirectory(scenario-builder)
\ No newline at end of file
+add_subdirectory(scenario-builder)
+add_subdirectory(loader)
```
flomnes, line 5: study-loader

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
```diff
@@ -0,0 +1,79 @@
...
+        return pstudy->internalLoadSets();
+    }
+    //@}
+
+    bool initializeInternalData(const StudyLoadOptions& options){
+        return pstudy->initializeInternalData(options);
+    }
+    void initializeSetsData(){
+        pstudy->initializeSetsData();
+    }
```
flomnes, line 53: Added for tests, not tested though. Either remove from this [attorney class](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Friendship_and_the_Attorney-Client), or test.

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
```diff
@@ -0,0 +1,79 @@
...
+    //@{
+    //! Reset the input extension according the study version
+    void inputExtensionCompatibility(){
+        pstudy->inputExtensionCompatibility();
+    }
+    //! Release all unnecessary buffers
+    void reduceMemoryUsage(){
+        pstudy->reduceMemoryUsage();
+    }
+    //@}
```
flomnes, line 65: Same as above

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
```diff
@@ -0,0 +1,79 @@
...
+    pStudy pstudy;
+    
+
+}; // class StudyWrapper
+
+} // namespace Data
+} // namespace Antares
+
+#include "runtime.h"
+
```
flomnes, line 79: Include at top ? Or don't include

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
```diff
@@ -0,0 +1,79 @@
...
+namespace Antares
+{
+namespace Data
+{
+class Study;
+/*!
+** \brief Antares StudyWrapper
+*/
+
+using pStudy = std::shared_ptr<Study>;
```
flomnes, line 13: `Study::Ptr` is already an alias to `shared_ptr<Study>`

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
```diff
@@ -0,0 +1,79 @@
...
+class StudyWrapper
+{
+
+public:
+    using Ptr = std::shared_ptr<StudyWrapper>;
```
flomnes, line 18: Not needed in my opinion. This object can be allocated on the stack

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
```diff
@@ -0,0 +1,79 @@
+
+#include <study.h>
+
```
flomnes, line 3: **study-wrapper.h** isn't part of any CMakeFile. It may result in an incomplete dependency graph.

flomnes (changes requested):

guilpier-code:
Why not leave a clear introduction to the work ? Ok, I often forget to leave one but it's still useful

**Caution** : the comments I leave here are not necessarily mandatory, we should discuss them

src/libs/antares/study/study.cpp
```diff
@@ -320,151 +322,127 @@ void Study::ensureDataAreAllInitialized()
...
+        nbYearsParallelRaw = nbYearsParallelForced;
+        return;
+    }
 
+    std::map<NumberOfCoresMode, uint> table;
     uint nbLogicalCores = Yuni::System::CPU::Count();
-    if (!nbLogicalCores)
-        logs.fatal() << "Number of logical cores available is 0.";
+
+    table[ncMin] = 1;
```
guilpier-code, line 337: We should rename **table** into **number_of_MC_year_threads** (for instance), table is meaningless.
Besides, **uint** in `std::map<NumberOfCoresMode, uint>` is overstated. **int** is enough.

src/libs/antares/study/study.cpp
```diff
@@ -547,6 +525,50 @@ void Study::getNumberOfCores(const bool forceParallel, const uint nbYearsParalle
...
+void Study::computeNumberOfCores(const StudyLoadOptions& options)
+{
+    /*
+            Getting the number of parallel years based on the number
+            of cores level.
+            This number is limited by the smallest refresh span (if at least
+            one type of time series is generated)
+    */
+    computeRawNbParallelYear(options.forceParallel, options.maxNbYearsInParallel);
+    maxNbYearsInParallel = getNbYearsParallelRaw();
```
guilpier-code, line 559: Why not : 
```C++
   maxNbYearsInParallel = computeRawNbParallelYear(options.forceParallel, options.maxNbYearsInParallel);
```
?

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
guilpier-code: And **pStudy** is very often (if not every time) a class data member variable that contains an instance of Study, not a type as here. There can be a confusion.

src/tests/src/libs/antares/study/wrappers/study-wrapper.h
```diff
@@ -0,0 +1,79 @@
...
+{
+namespace Data
+{
+class Study;
+/*!
+** \brief Antares StudyWrapper
+*/
+
+using pStudy = std::shared_ptr<Study>;
+class StudyWrapper
```
guilpier-code, line 14: What is the point of this study wrapper ? (It may have a real benefit, but I can't see which one)

I can see that this wrapper is only used (for now) in the newly created tests. So, if it is meant to be used in the context of tests, maybe we should move it next to the tests.

src/libs/antares/study/study.cpp
```diff
@@ -48,6 +48,8 @@
 #include <yuni/core/system/cpu.h> // For use of Yuni::System::CPU::Count()
 #include <math.h>                 // For use of floor(...) and ceil(...)
 
+#include <string>
```
guilpier-code, line 51: `#include <string>` : is it useful ?

src/libs/antares/study/study.cpp
```diff
@@ -320,151 +322,127 @@ void Study::ensureDataAreAllInitialized()
     StudyEnsureDataThermalPrepro(this);
 }
 
-std::map<std::string, uint> Study::getRawNumberCoresPerLevel()
+
+void Study::computeRawNbParallelYear(const bool forceParallel, const uint nbYearsParallelForced)
```
guilpier-code, line 326: In this file, all changes are related to parallel MC years, right ?
We could try to extract all the logic (computing sets of parallel years) from the study, in order to : 
- empty a little bit the god class **Study**. Said in another way, it would the number of responsibilities of **Study**
- this extracted code could be tested in isolation.
- avoid currently existing code duplication with **solver.hxx** : **buildSetsOfParallelYears(...)**

We could have something like : 
```C++
class setsOfParallelYearsCalculator // Or some name more accurate
{
public:
     setsOfParallelYearsCalculator (
             int nbYears,
             bool userPlaylist,
             int timeSeriesToGenerate,
             int timeSeriesToRefresh,
             bool forceParallel,
             int forcedNbOfParallelYears // Today called "maxNbYearsInParallel"
             );
     void build(); // Main function, it would do what "computeNumberOfCores" does
     bool allSetsParallelYearsHaveSameSize();
private:
     // Private methods
     computeRawNbParallelYear();
     computeMinNbYearsInParallelYearSet();
     
     // Private data members
     std::vector<std::vector<uint>> setsOfParallelYears_;
}
```
**Note** : 
- In this proposal, the class does not take the whole object **parameters** as an argument for its constructor. Instead, we would pass the constructor exactly the list of data it really needs. Indeed, class parameters can be split or can change. The data we give won't.
-

src/libs/antares/study/load-params.cpp
```diff
@@ -0,0 +1,185 @@
...
+    SetHandlerAreas(AreaList& _areas) : areas(_areas)
+    {
+    }
+
+    void clear(Study::SingleSetOfAreas& set)
+    {
+        set.clear();
+    }
+
+    uint size(Study::SingleSetOfAreas& set) const
```
guilpier-code, line 101: `uint size(Study::SingleSetOfAreas& set) const` : I don't see the point of this function.
Instead of calling `sets_of_areas_handler.size(set_of_areas)`, let's simply call `sets_of_areas.size()`

src/libs/antares/study/load-params.cpp
```diff
@@ -0,0 +1,185 @@
...
+    if (usedByTheSolver && !checkHydroHotStart())
+        return false;
+
+    // Reducing memory footprint
+    reduceMemoryUsage();
+    return true;
+
+}
+
+class SetHandlerAreas
```
guilpier-code, line 89: Each method of class **SetHandlerAreas** is given a `SingleSetOfAreas& set` (which is a `std::set<Area*>`), and makes something with this argument.
Why not put the **set** in the class and rename the class **setOfAreas** ?
There would be some side effects to that change : it will bring some other changes in the code.
It remains to evaluate the size of these changes.

src/libs/antares/study/load-params.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+using namespace Yuni;
+
+
+namespace Antares
+{
+namespace Data
+{
+
+bool Study::initializeInternalData(const StudyLoadOptions& options)
```
guilpier-code, line 38: Apparently, method **initializeInternalData** does 3 (at least) things : 
- it initializes **parameters** (it should be done by a method of class **Parameters**, or better : in the **Parameters** constructor, if possible)
- it initializes the calendar (same remark as above, with class **Date::Calendar**)
- it makes a check (should be moved with other checks : to be explored)

Each bullet should be separated from the others in a function / method
This split would have few consequences **test-study-load-internal.cpp**.

guilpier-code (changes requested):

kathvargasr:
replaced by #1017 and  #1057

PR #895, opened 2022-10-07
Remove error message printed on -h/--help

### LINKED ISSUE #894: solver in command line : help menu
When running : 
`> .\antares-8.3-solver.exe --help`
we get the list of options that the solver offers, as expected, but we also get an error : 
```
[Thu Oct  6 17:49:13 2022][solver][error] A study folder is required. Use '--help' for more information
[Thu Oct  6 17:49:13 2022][solver][fatal] Aborting now. (warning: no file log available)
[Thu Oct  6 17:49:13 2022][solver][warns] No log file available
```
Note that the error message ask the user to run precisely the command that generates the error message !

flomnes:
- Use a return code with 3 values : `OK`, `ERROR` and `HELP`
- Don't display an error message if that return code is `HELP`
- [x] Fix compilation errors on Windows


flomnes (commented):
`ERROR` does not compile because of some `#define` in Windows. Since it is in an `enum class`, I suggest renaming it to `error`

Issue #452, opened 2021-11-10
The code contains several global variables


flomnes:
See **sim_variables_globales.cpp**
```cpp
DONNEES_PAR_PAYS** DonneesParPays;

char* Positionnement;
char* PositionnementSV;
char* LigneACompter;
char* LigneLue;
char AppelEnModeSimulateur;

VALEURS_GENEREES_PAR_PAYS*** ValeursGenereesParPays;
NUMERO_CHRONIQUES_TIREES_PAR_PAYS*** NumeroChroniquesTireesParPays;

PROBLEME_HORAIRE_ADEQUATION ProblemeHoraireAdequation;
VALEURS_ANNUELLES** ValeursAnnuellesAdequation;

RESULTATS_PAR_INTERCONNEXION** ResultatsParInterconnexion;
RESULTATS_PAR_CONTRAINTE_COUPLANTE** ResultatsParContrainteCouplante;

COMPTE_RENDU CompteRendu;
```

PR #896, opened 2022-10-07
Remove unused global variables

flomnes:
- Remove unused global variables
- Fix style for `for` loops


flomnes:
Partially fixes #452

PR #903, opened 2022-10-10
Synchronize latest RTE-i/adq-patch branch with the RTE/do-not-merge-csr-generate-executables

Milos-RTEi:


src/libs/antares/study/parameters.cpp
```diff
@@ -636,6 +683,19 @@ static bool SGDIntLoadFamily_AdqPatch(Parameters& d,
...
+    // Include Hurdle Cost
+    if (key == "include-hurdle-cost-csr")
+        return value.to<bool>(d.adqPatch.curtailmentSharing.includeHurdleCost);
+    // Thresholds
+    if (key == "threshold-initiate-curtailment-sharing-rule")
+        return value.to<double>(d.adqPatch.curtailmentSharing.thresholdInitiate);
+    if (key == "threshold-display-local-matching-rule-violations")
+        return value.to<double>(d.adqPatch.curtailmentSharing.thresholdDisplayViolations);
+    if (key == "threshold-csr-variable-bounds-relaxation")
+        return value.to<int>(d.adqPatch.curtailmentSharing.thresholdVarBoundsRelaxation);
```
flomnes, line 698: This function feeds d.adqPatch. I suggest making it a member function of `struct AdequacyPatch`. We'd have something like this here
```cpp
return d.adqPatch.setParameter(key, value);
```

flomnes:
TODO

- [ ] Fix merge conflicts
- [ ] Document format changes (see https://antares-simulator.readthedocs.io/en/latest/reference-guide/13-file-format/, file **docs/reference-guide/13-file-format.md**)

docs/reference-guide/03-commands.md
```diff
@@ -211,9 +211,19 @@ given back their regular capacities (infinite for those being set on "set to inf
     - _Export mps (false/true)_
     - _Simplex optimization range [^4] (day / week)_
     - _Unfeasible problems behavior (Error Dry/ Error Verbose/ Warning Dry/ Warning Verbose_
-    - _Enable Adequacy patch (false / true)_
-    - _NTC from physical areas outside to physical areas inside adequacy patch (set to null / local values)_
-    - _NTC between physical areas outside adequacy patch (set to null / local values)_
+
```
guilpier-code, line 214: Documentation seems incomplete here

docs/reference-guide/03-commands.md
```diff
@@ -211,9 +211,19 @@ given back their regular capacities (infinite for those being set on "set to inf
...
+- **Adequacy Patch** Auxiliary window [Options] Defines a set of options related to the adequacy patch.
+The set of preferences is study-specific; it can be changed at any time and saved along with study data.
+Auxiliary window [Areas] Opens a window in which a choice can be made regarding the individual area adequacy patch mode.
+
+- _Enable Adequacy patch (false / true)_
+- _NTC from physical areas outside to physical areas inside adequacy patch (set to null / local values)_
+- _NTC between physical areas outside adequacy patch (set to null / local values)_
+- _Price taking order (DENS / Load)_
+- _Include hurdle cost in CSR optimization (false / true)_
+- _Thresholds:_
```
guilpier-code, line 224: Items under **Threshold** should be indented

src/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set(ANTARES_WEBSITE "https://antares.rte-france.com")
 set(ANTARES_ONLINE_DOC "https://antares-simulator.readthedocs.io/")
 
 # Beta release
-set(ANTARES_BETA 0)
+set(ANTARES_BETA 7)
```
guilpier-code, line 17: Why beta **7** ?

src/libs/antares/study/parameters.cpp
```diff
@@ -177,6 +178,39 @@ const char* StudyModeToCString(StudyMode mode)
     }
     return "Unknown";
 }
+bool StringToPriceTakingOrder(const AnyString& text, AdequacyPatch::AdqPatchPTO& out)
```
guilpier-code, line 181: Let's call the 2 arguments differently : **PTO_as_string** and **PTO_as_enum** for instance

src/solver/optimisation/adequacy_patch.h
```diff
@@ -37,75 +37,91 @@ namespace Data
...
-    * @return uint from an enumeration that describes the type of restrictions to put on this link for
-    * adq purposes.
-    */
-    LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStep(
-        AdequacyPatchMode OriginNodeAdequacyPatchType,
-        AdequacyPatchMode ExtremityNodeAdequacyPatchType,
-        bool SetToZero12LinksForAdequacyPatch,
-        bool SetToZero11LinksForAdequacyPatch);
+//! A default threshold value for initiate curtailment sharing rule
+const double defaultValueThresholdInitiateCurtailmentSharingRule = 0.0;
```
guilpier-code, line 41: Would **defaultThresholdToRunCurtailmentSharing** be better ?
**Initiate** is unclear for me : we run the CSR algorithm.

src/solver/optimisation/adequacy_patch.h
```diff
@@ -37,75 +37,91 @@ namespace Data
...
+//! A default threshold value for display local matching rule violations
+const double defaultValueThresholdDisplayLocalMatchingRuleViolations = 0.0;
```
guilpier-code, line 43: **defaultThresholdDisplayLocalMatchingRuleViolations** is enough (I only took off **Value**)

src/libs/antares/study/parameters.h
```diff
@@ -524,6 +528,22 @@ class Parameters final
         };
         bool enabled;
         LocalMatching localMatching;
+
+        struct CurtailmentSharing
+        {
+            //! PTO (Price Taking Order) for adequacy patch. User can choose between DENS and Load.
+            Data::AdequacyPatch::AdqPatchPTO priceTakingOrder;
+            //! Threshold to initiate curtailment sharing rule
+            double thresholdInitiate;
```
guilpier-code, line 537: **thresholdInitiate** : not fan of **initiate**, **run** is better to me. 
```C++
   //! Threshold to run curtailment sharing rule algorithm
   double threshold;
```

src/libs/antares/study/parameters.cpp
```diff
@@ -201,12 +235,25 @@ void Parameters::resetSeeds()
...
+    // Initialize all thresholds values for adequacy patch
+    adqPatch.curtailmentSharing.thresholdInitiate
+      = defaultValueThresholdInitiateCurtailmentSharingRule;
+    adqPatch.curtailmentSharing.thresholdDisplayViolations
+      = defaultValueThresholdDisplayLocalMatchingRuleViolations;
+    adqPatch.curtailmentSharing.thresholdVarBoundsRelaxation
+      = defaultValueThresholdVarBoundsRelaxation;
+}
 
 void Parameters::resetAdqPatchParameters()
```
guilpier-code, line 249: We should have something like : 
```C++
void Parameters::resetAdqPatchParameters()
{
    adqPatch.enabled = false;    
    this->resetAdqPatch_LocalMatching();    
    this->resetAdqPatch_CurtailmentSharing();
}

````

File: src/libs/antares/study/parameters.cpp [hunk 10]
```diff
@@ -636,6 +683,19 @@ static bool SGDIntLoadFamily_AdqPatch(Parameters& d,
...
     if (key == "set-to-null-ntc-between-physical-out-for-first-step")
         return value.to<bool>(d.adqPatch.localMatching.setToZeroOutsideOutsideLinks);
+    // Price taking order
+    if (key == "price-taking-order")
+        return StringToPriceTakingOrder(value, d.adqPatch.curtailmentSharing.priceTakingOrder);
+    // Include Hurdle Cost
+    if (key == "include-hurdle-cost-csr")
+        return value.to<bool>(d.adqPatch.curtailmentSharing.includeHurdleCost);
+    // Thresholds
+    if (key == "threshold-initiate-curtailment-sharing-rule")
```
guilpier-code, line 693: Renaming : 
```C++
    if (key == "threshold-run-curtailment-sharing-rule")
        return value.to<double>(d.adqPatch.curtailmentSharing.threshold);
```

src/libs/antares/study/parameters.cpp
```diff
@@ -1762,7 +1822,18 @@ void Parameters::saveToINI(IniFile& ini) const
                      adqPatch.localMatching.setToZeroOutsideInsideLinks);
         section->add("set-to-null-ntc-between-physical-out-for-first-step",
                      adqPatch.localMatching.setToZeroOutsideOutsideLinks);
+        section->add("price-taking-order",
+                     PriceTakingOrderToString(adqPatch.curtailmentSharing.priceTakingOrder));
+        section->add("include-hurdle-cost-csr", adqPatch.curtailmentSharing.includeHurdleCost);
+        // Threshholds
```
guilpier-code, line 1828: `// Threshholds` ==> `// Thresholds`

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -26,35 +26,45 @@
 */
 
 #include "../simulation/simulation.h"
+#include "opt_fonctions.h"
 #include "adequacy_patch.h"
+#include <math.h>
+#include "../simulation/sim_structure_probleme_economique.h"
```
guilpier-code, line 32:
```C++
#include "../simulation/sim_structure_probleme_economique.h"
```
This line is useless

src/solver/simulation/economy.h
```diff
@@ -47,7 +47,7 @@ class EconomyWeeklyOptimization
 {
 public:
     using Ptr = std::unique_ptr<EconomyWeeklyOptimization>;
-    virtual void solve(Variable::State& state, int hourInTheYear, uint numSpace) = 0;
+    virtual void solve(Variable::State& state, int hourInTheYear, uint numSpace, uint week) = 0;
     void initializeProblemeHebdo(PROBLEME_HEBDO** pProblemesHebdo);
```
guilpier-code, line 51: Do we need method **initializeProblemeHebdo** ? **pProblemesHebdo** should be given as argument to the constructor of base class **EconomyWeeklyOptimization**

src/solver/simulation/economy.cpp
```diff
@@ -46,7 +48,6 @@ namespace Simulation
 {
 enum
 {
-
     nbHoursInAWeek = 168,
 };
```
guilpier-code, line 53: In **EconomyWeeklyOptimization::create()** I would use **std..make_unique<...>()** : 
```C++
std::unique_ptr<EconomyWeeklyOptimization> EconomyWeeklyOptimization::create(bool adqPatchEnabled)
{
    if (adqPatchEnabled)
        return make_unique<AdequacyPatchOptimization>();
    else
        return make_unique<NoAdequacyPatchOptimization>();
    return nullptr;
}
```
I did not test but this should work, right ?

src/solver/simulation/economy.h
```diff
@@ -47,7 +47,7 @@ class EconomyWeeklyOptimization
 {
 public:
     using Ptr = std::unique_ptr<EconomyWeeklyOptimization>;
```
guilpier-code, line 49: Not sure this definition is useful : 
`using Ptr = std::unique_ptr<EconomyWeeklyOptimization>;`

It prevents the reader from seing it's a **std::unique_ptr**, and when used, it's almost as long : `EconomyWeeklyOptimization::Ptr `

src/solver/simulation/economy.cpp
```diff
@@ -69,7 +70,10 @@ EconomyWeeklyOptimization::Ptr EconomyWeeklyOptimization::create(bool adqPatchEn
 
 // Adequacy patch
 AdequacyPatchOptimization::AdequacyPatchOptimization() = default;
```
guilpier-code, line 72: No need of the following line : 
`AdequacyPatchOptimization::AdequacyPatchOptimization() = default;`
Tag the constructor as **default** in the header file **.h**

src/solver/simulation/economy.cpp
```diff
@@ -93,11 +97,13 @@ void AdequacyPatchOptimization::solve(Variable::State& state, int hourInTheYear,
     // part that we need
     ::SIM_RenseignementProblemeHebdo(*problemeHebdo, state, numSpace, hourInTheYear);
     OPT_OptimisationHebdomadaire(problemeHebdo, numSpace);
+
+    solveCSR(state, numSpace, w);
 }
 
 // No adequacy patch
 NoAdequacyPatchOptimization::NoAdequacyPatchOptimization() = default;
```
guilpier-code, line 105: As above, the `= default` instruction should not be here

src/solver/simulation/economy.h
```diff
@@ -59,14 +59,19 @@ class AdequacyPatchOptimization : public EconomyWeeklyOptimization
...
-    void solve(Variable::State& state, int hourInTheYear, uint numSpace) override;
+    void solve(Variable::State& state, int hourInTheYear, uint numSpace, uint week) override;
+private:
+    void solveCSR(const Variable::State& state, uint numSpace, uint week);
+    vector<double> calculateENSoverAllAreasForEachHour(uint numSpace) const;
+    std::set<int> identifyHoursForCurtailmentSharing(vector<double> sumENS, uint numSpace) const;
+    std::set<int> getHoursRequiringCurtailmentSharing(uint numSpace) const;
 };
 
 class NoAdequacyPatchOptimization : public EconomyWeeklyOptimization
```
guilpier-code, line 70: We should not name this optimization **NoAdequacyPatchOptimization** : if a 3rd optimization (let's name it **my-3rd-opt**) shows up, we would be forced to name it **NoAdequacyPatchAndNoMy3rdOptOptimization**.
I suggest naming it  **weeklyOptimization** and name the parent class **baseWeeklyOptimization** or **interfaceWeeklyOptimization** or **IWeeklyOptimization**

src/solver/simulation/economy.h
```diff
@@ -47,7 +47,7 @@ class EconomyWeeklyOptimization
 {
 public:
     using Ptr = std::unique_ptr<EconomyWeeklyOptimization>;
-    virtual void solve(Variable::State& state, int hourInTheYear, uint numSpace) = 0;
+    virtual void solve(Variable::State& state, int hourInTheYear, uint numSpace, uint week) = 0;
```
guilpier-code, line 50: **economy.h** and **economy.cpp** start to grow. They mix **economy** mode and weekly optimization.
In order to make code more clear we should split economy.h/cpp into 4 parts : 
- **economy**.h/cpp containing only class **economy** and 
- **baseWeeklyOptimization**.h/cpp for weekly optimization base class 
- **weeklyOptimization**.h/cpp for "normal" or "classic" optimization class
- **adqPatchWeeklyOptimization**.h/cpp for adequacy patch weekly optimization

src/solver/simulation/economy.cpp
```diff
@@ -219,7 +280,7 @@ bool Economy::year(Progression::Task& progression,
 
         try
         {
-            weeklyOptProblem->solve(state, hourInTheYear, numSpace);
+            weeklyOptProblem->solve(state, hourInTheYear, numSpace, w);
```
guilpier-code, line 283: In the call of the **solve(...)** method, there are several things to be pointed out : 
- the argument **state** is only there because its member `state.year` is used inside the call. So, instead of passing the entire state, we should consider passing the year only.
- the argument **numSpace** (**thread_number** would be a better name) is passed here and is also passed through several levels of calls inside **solve(...)**. It is used  : 

    - once to call the resolution of the classic optimization ( `OPT_OptimisationHebdomadaire(problemeHebdo, numSpace)`)
    - many times to reach a particular year optimization problem (`pProblemesHebdo[numSpace]`). If we store directly `pProblemesHebdo[numSpace]` as a member data (of **AdequacyPatchOptimization**), it makes **numSpace** useless in many cases and we don't have to pass it to many private methods.

    I strongly recommend to 
    - store the **numSpace** as a data member of the 2 weekly optimization classes
    - store the corresponding weekly problem (`pProblemesHebdo[numSpace]`) as a data member directly, instead of storing the table **pProblemesHebdo** with method **initializeProblemeHebdo**. 
 
To sum up what I mean : 
```C++
bool Economy::simulationBegin()
{
   ....
    // Please remove this block
    // if (weeklyOptProblem)
    // {
    //      weeklyOptProblem->initializeProblemeHebdo(pProblemesHebdo);
    // }
}
...

bool Economy::year(/* many arguments */)
{
    ... 
    // Store numSpace, pProblemesHebdo[numSpace] and numSpace
    weeklyOptProblem->initializeProblemeHebdo(state.year, numSpace, pProblemesHebdo);
    ...
    for (uint w = 0; w != pNbWeeks; ++w)
    {
         .... 
         weeklyOptProblem->solve(hourInTheYear, w);
         ....
    }
}
```

src/solver/simulation/economy.cpp
```diff
@@ -185,6 +191,61 @@ bool Economy::simulationBegin()
...
+    {
+        if (pProblemesHebdo[numSpace]->adequacyPatchRuntimeData.areaMode[area]
+            == Data::AdequacyPatch::physicalAreaInsideAdqPatch)
+            addArray(sumENS,
+                     pProblemesHebdo[numSpace]
+                       ->ResultatsHoraires[area]
+                       ->ValeursHorairesDeDefaillancePositive);
+    }
+    return sumENS;
+}
```
guilpier-code, line 207: Suggestion (if other remarks of this file are taken into account) : 
```C++
vector<double> AdequacyPatchOptimization::calculateENSoverAllAreasForEachHour() const
{
    std::vector<double> sumENS(nbHoursInAWeek, 0.0);
    for (int area = 0; area < problemeHebdo_->NombreDePays; ++area)
    {
        if (problemeHebdo_->isAreaInsideAdqPatchDomain(area))
        {
             double * ENS = problemesHebdo_->ResultatsHoraires[area]
                                       ->ValeursHorairesDeDefaillancePositive;
             for (uint h = 0; h < nbHoursInAWeek; ++h)
                 sumENS[i] += ENS[i];
        }
    }
    return sumENS;
}
```

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -67,72 +77,239 @@ LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeInside
...
+        densNew = Math::Max(0.0, ensInit + netPositionInit);
+
+    return std::make_pair(netPositionInit, densNew);
 }
 
-void setBoundsNoAdqPatch(double& Xmax,
-                         double& Xmin,
-                         VALEURS_DE_NTC_ET_RESISTANCES* ValeursDeNTC,
-                         const int Interco)
+void addArray(std::vector<double>& A, const double* B)
```
guilpier-code, line 231: We may not need the function **addArray** : see remarks in **economy.cpp**, more precisely on function **calculateENSoverAllAreasForEachHour**

src/solver/simulation/economy.cpp
```diff
@@ -185,6 +191,61 @@ bool Economy::simulationBegin()
...
+
+std::set<int> AdequacyPatchOptimization::identifyHoursForCurtailmentSharing(vector<double> sumENS,
+                                                                            uint numSpace) const
+{
+    double threshold
+      = pProblemesHebdo[numSpace]->adqPatchParams->ThresholdInitiateCurtailmentSharingRule;
+    std::set<int> triggerCsrSet;
+    for (int i = 0; i < nbHoursInAWeek; ++i)
```
guilpier-code, line 215: For hours indices, prefer **h**, or better : **hour**

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -707,4 +715,41 @@ struct PROBLEME_HEBDO
     double maxPminThermiqueByDay[366];
 };
 
+// hourly CSR problem structure
```
guilpier-code, line 718: About class **HOURLY_CSR_PROBLEM** declaration : each method of this class are defined in **src/solver/optimisation/adequacy_patch.cpp**, so why not move this declaration in corresponding **adequacy_patch.h** ?
As it is done (declaration in the current file, and definition in another), if we touch the class, more things will have to be re-compiled

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -707,4 +715,41 @@ struct PROBLEME_HEBDO
...
+    void buildProblemConstraintsLHS();
+    void buildProblemConstraintsRHS();
+    void setProblemCost();
+    void solveProblem(uint week, int year);
+public:
+    void run(uint week, int year);
+    
+    int hourInWeekTriggeredCsr;
+    double belowThisThresholdSetToZero;
+    PROBLEME_HEBDO* pWeeklyProblemBelongedTo;
```
guilpier-code, line 735: About **pWeeklyProblemBelongedTo** : everywhere in the code, this object called **problemHebdo**, could not we do the same here, to avoid confusion ?

src/solver/simulation/economy.cpp
```diff
@@ -185,6 +191,61 @@ bool Economy::simulationBegin()
...
+    const std::set<int> hoursRequiringCurtailmentSharing
+      = getHoursRequiringCurtailmentSharing(numSpace);
+    for (int hourInWeek : hoursRequiringCurtailmentSharing)
+    {
+        logs.info() << "[adq-patch] CSR triggered for Year:" << state.year + 1
+                    << " Hour:" << w * nbHoursInAWeek + hourInWeek + 1;
+        HOURLY_CSR_PROBLEM hourlyCsrProblem(hourInWeek, problemeHebdo);
+        hourlyCsrProblem.run(w, state.year);
+    }
+    double totalLmrViolation = checkLocalMatchingRuleViolations(problemeHebdo);
```
guilpier-code, line 244: As **AdequacyPatchOptimization** owns **problemeHebdo** already, it would make sense to make **checkLocalMatchingRuleViolations()** a method of **AdequacyPatchOptimization**, instead of a free function. 
And it's the same thing for **LmrViolationArea()**.

src/solver/optimisation/adequacy_patch_csr/construct_problem_variables.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeQuadratique_CSR(
+  PROBLEME_HEBDO* ProblemeHebdo,
+  HOURLY_CSR_PROBLEM& hourlyCsrProblem)
+{
+    logs.debug() << "[CSR] variable list:";
+
+    const PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre;
+    ProblemeAResoudre = ProblemeHebdo->ProblemeAResoudre;
+    assert(ProblemeAResoudre != NULL);
```
guilpier-code, line 155:
```C++
assert(ProblemeHebdo->ProblemeAResoudre != NULL);
```
could replace 3 lines 153 -> 155

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -67,72 +77,239 @@ LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeInside
...
+}
+
+void HOURLY_CSR_PROBLEM::resetProblem()
+{
+    OPT_LiberationProblemesSimplexe(pWeeklyProblemBelongedTo);
+}
+
+void HOURLY_CSR_PROBLEM::buildProblemVariables()
+{
+    OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeQuadratique_CSR(pWeeklyProblemBelongedTo,
```
guilpier-code, line 276: We don't need to pass **pWeeklyProblemBelongedTo** to **OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeQuadratique_CSR**, because **this** already owns this object (**this->pWeeklyProblemBelongedTo**). So the first argument is useless.

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_quadratique.cpp
```diff
@@ -32,6 +32,8 @@
 #include "../simulation/sim_extern_variables_globales.h"
 #include "opt_fonctions.h"
 
+using namespace Antares::Data;
```
guilpier-code, line 35: As the file barely changed, is `using namespace Antares::Data;` still required ?

src/solver/optimisation/opt_fonctions.h
```diff
@@ -36,20 +36,26 @@ void OPT_NumeroDeJourDuPasDeTemps(PROBLEME_HEBDO*);
 void OPT_NumeroDIntervalleOptimiseDuPasDeTemps(PROBLEME_HEBDO*);
 void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeLineaire(PROBLEME_HEBDO*);
 void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeQuadratique(PROBLEME_HEBDO*);
+void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeQuadratique_CSR(PROBLEME_HEBDO*, HOURLY_CSR_PROBLEM&);
```
guilpier-code, line 39: All added functions here don't need two arguments, the 2nd one is enough

guilpier-code (changes requested):

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -707,4 +715,41 @@ struct PROBLEME_HEBDO
...
+    void buildProblemVariables();
+    void setVariableBounds();
+    void buildProblemConstraintsLHS();
+    void buildProblemConstraintsRHS();
+    void setProblemCost();
+    void solveProblem(uint week, int year);
+public:
+    void run(uint week, int year);
+    
+    int hourInWeekTriggeredCsr;
```
guilpier-code, line 733: `int hourInWeekTriggeredCsr` : **hour_in_week_** should be enough because, as we are inside class **HOURLY_CSR_PROBLEM**, we know the problem was triggered and that it is a CSR problem, right ?

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -67,72 +77,239 @@ LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeInside
...
+                                                                         *this);
+}
+
+void HOURLY_CSR_PROBLEM::buildProblemConstraintsLHS()
```
guilpier-code, line 280: Private method of **HOURLY_CSR_PROBLEM** : many of them contain only a call to a free function, then why not remove a "floor" and directly put the free function's body on the method ?

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -67,72 +77,239 @@ LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeInside
...
+                             ->ValeursHorairesDeDefaillanceNegative[hour];
+
+            densNewValues[Area] = densNew;
+            rhsAreaBalanceValues[Area] = ensInit + netPositionInit - spillageInit;
+        }
+    }
+    return;
+}
+
+void HOURLY_CSR_PROBLEM::resetProblem()
```
guilpier-code, line 269: **pWeeklyProblemBelongedTo** : here an weekly linear problem is used to build and run an hourly quadratic problem, it is strange. 
In fact, the weekly linear problem of the simulation has nothing to see with our CSR quadratic problem.
We should instantiate a new quadratic problem,  either as a data member of class **HOURLY_CSR_PROBLEM**, or in **HOURLY_CSR_PROBLEM::run()**
**pWeeklyProblemBelongedTo** should not have to be passed to many free functions in private methods of **HOURLY_CSR_PROBLEM**

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -67,72 +77,239 @@ LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeInside
...
+{
+    OPT_LiberationProblemesSimplexe(pWeeklyProblemBelongedTo);
```
guilpier-code, line 271: So, according to previous comments, `HOURLY_CSR_PROBLEM::resetProblem()` should not exist, because **pWeeklyProblemBelongedTo** shoukd not be used to build the CSR quadratic problem.

**pWeeklyProblemBelongedTo** should be used to build the parameters of the CSR quadratic problem.

src/solver/optimisation/adequacy_patch.h
```diff
@@ -1,6 +1,6 @@
 /*
-** Copyright 2007-2018 RTE
-** Authors: Antares_Simulator Team
+** Copyright 2007-2022 RTE
+** Authors: RTE-international / Redstork / Antares_Simulator Team
```
guilpier-code, line 3: This file (and its **.cpp** counterpart) should be split into **adqPatch-localMatching.h/cpp** and **adqPatch-curtailmentSharing.h/cpp**

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -707,4 +715,41 @@ struct PROBLEME_HEBDO
...
+    double belowThisThresholdSetToZero;
+    PROBLEME_HEBDO* pWeeklyProblemBelongedTo;
+    HOURLY_CSR_PROBLEM(int hourInWeek, PROBLEME_HEBDO* pProblemeHebdo)
+    {
+        hourInWeekTriggeredCsr = hourInWeek;
+        pWeeklyProblemBelongedTo = pProblemeHebdo;
+        belowThisThresholdSetToZero
+          = pProblemeHebdo->adqPatchParams->ThresholdCSRVarBoundsRelaxation;
+    };
+    std::map<int, int> numberOfConstraintCsrEns;
```
guilpier-code, line 743: **numberOfConstraintCsrEns** seems useless

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -707,4 +715,41 @@ struct PROBLEME_HEBDO
...
+    std::map<int, int> numberOfConstraintCsrAreaBalance;
+    std::map<int, int> numberOfConstraintCsrFlowDissociation;
+    std::map<int, int> numberOfConstraintCsrHourlyBinding; // length is number of binding constraint
```
guilpier-code, line 746: The 3 previous map are all used inside the CSR quadratic problem.
They should be data member of an object of class **csrQuadraticProblem** (see remarks in **adequacy_patch.cpp**).
Same remark for **varToBeSetToZeroIfBelowThreshold**

src/solver/optimisation/adequacy_patch.cpp
```diff
@@ -67,72 +77,239 @@ LinkCapacityForAdequacyPatchFirstStep SetNTCForAdequacyFirstStepOriginNodeInside
...
+{
+    OPT_InitialiserLesCoutsQuadratiques_CSR(pWeeklyProblemBelongedTo, *this);
+}
+
+void HOURLY_CSR_PROBLEM::solveProblem(uint week, int year)
+{
+    ADQ_PATCH_CSR(pWeeklyProblemBelongedTo->ProblemeAResoudre, *this, week, year);
+}
+
+void HOURLY_CSR_PROBLEM::run(uint week, int year)
```
guilpier-code, line 305: Many methods of **HOURLY_CSR_PROBLEM** are about building and running a CSR quadratic problem.
We should have a class **csrQuadraticProblem**, and **HOURLY_CSR_PROBLEM::run(...)** should run an instance of it.
In code word, we should have something like : 
```C++
void HOURLY_CSR_PROBLEM::run(uint week, int year)
{
    computeCSRquadraticProblemParameters();

    csrQuadraticProblem csr_quad_pb(hourInWeek,
                                    densNewValues, 
                                    rhsAreaBalanceValues); // and maybe other args
    csr_quad_pb.buildVariables();
    csr_quad_pb.buildProblemConstraints();
    ....
    csr_quad_pb.solve(week, year);
}
```

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_LHS.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    Pi = (double*)MemAlloc(ProblemeAResoudre->NombreDeVariables * sizeof(double));
+    Colonne = (int*)MemAlloc(ProblemeAResoudre->NombreDeVariables * sizeof(int));
+
+    ProblemeAResoudre->NombreDeContraintes = 0;
+    ProblemeAResoudre->NombreDeTermesDansLaMatriceDesContraintes = 0;
+
+    setConstraintsOnFlows(ProblemeHebdo, hourlyCsrProblem, Pi, Colonne);
+    setNodeBalanceConstraints(ProblemeHebdo, hourlyCsrProblem, Pi, Colonne);
+    setBindingConstraints(ProblemeHebdo, hourlyCsrProblem, Pi, Colonne);
```
guilpier-code, line 301: These 3 previous methods should be private methods of a class **csrQuadraticProblem**

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_LHS.cpp
```diff
@@ -0,0 +1,305 @@
...
+                  Pi,
+                  Colonne,
+                  NombreDeTermes,
+                  MatriceDesContraintesCouplantes->SensDeLaContrainteCouplante);
+            }
+        }
+    }
+}
+
+void OPT_ConstruireLaMatriceDesContraintesDuProblemeQuadratique_CSR(
```
guilpier-code, line 283: According to review of : 
- **adequacy_patch.cpp**, 
- **sim_structure_probleme_economique.h**

this function should be a method of a class  **csrQuadraticProblem**, and have no argument.

guilpier-code (changes requested):

src/CMakeLists.txt
flomnes: will be fixed before merge

src/libs/antares/study/parameters.cpp
flomnes: Renaming parameters will force users to change their studies

src/solver/simulation/economy.h
flomnes: See #954

src/solver/simulation/economy.cpp
flomnes: See #954 (again)

src/solver/simulation/economy.cpp
flomnes: The ctor is no longer the default one (#954)

src/solver/simulation/economy.cpp
flomnes: - We'd need to remove `struct h2oValueWorkVarsType`, use local variables instead.
- We'd need to pass `state.weekInTheYear` instead of `state`

src/CMakeLists.txt
payetvin: It's ok to remove it ?

flomnes: Yes, revert this change

src/solver/optimisation/adequacy_patch.cpp
payetvin: this is not valid in OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeQuadratique_CSR
I'll keep it this way

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_quadratique.cpp
payetvin: Indeed, removed it

src/solver/simulation/sim_structure_probleme_economique.h
payetvin: there's already a hoursInWeek var, maybe it' better to keep this name in order to differentiate them

src/solver/optimisation/adequacy_patch.cpp
payetvin: I dont' understand where's the free you're talking about

src/solver/optimisation/adequacy_patch.h
payetvin: The file is not huge and I'm not sure how to split it

src/solver/optimisation/adequacy_patch.cpp
payetvin: This would make HOURLY_CSR_PROBLEM just a wrapper of the csrQuadraticProblem class adding complexity, I don't see the point

guilpier-code:
Actually, the more up to date version of adequacy patch work can be found in #961. 
The branch associated to the current pull request is a bit older. 
Therefore, there are a few more remarks in that pull request.
We should take them into account.

src/solver/simulation/economy.h
guilpier-code: Done in #968

src/solver/simulation/economy.cpp
guilpier-code: Done in #968

src/solver/simulation/economy.cpp
guilpier-code: Done in #971

src/solver/optimisation/adequacy_patch.cpp
guilpier-code: Done in https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/971

src/solver/simulation/economy.cpp
guilpier-code: Done in #971

src/solver/simulation/sim_structure_probleme_economique.h
guilpier-code: Sorry I dont see another **hoursInWeek** as a data member of class **HOURLY_CSR_PROBLEM**, do I miss something ?

PR #922, opened 2022-10-26
[TEST] Remove free functions for areas' mem allocation

kathvargasr:


src/ui/simulator/application/study.cpp
```diff
@@ -224,8 +224,7 @@ class JobOpenStudy final : public Toolbox::Jobs::Job
 
         // Postflight
         logs.info();
-        study->ensureDataAreAllInitialized();
-
+        study->areas.ensureDataIsInitialized(study->parameters, options.loadOnlyNeeded);
```
flomnes, line 227: Possible change of behavior here. Previously, `loadOnlyNeeded` was implicitly set to `false`. Now it depends on the context.

src/libs/antares/study/area/list.cpp
```diff
@@ -1212,6 +1212,42 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
 }
 
+void AreaList::ensureDataIsInitialized(Parameters& params, bool loadOnlyNeeded)
+{
+    AreaListEnsureDataLoadTimeSeries(this);
+    AreaListEnsureDataSolarTimeSeries(this);
+    AreaListEnsureDataWindTimeSeries(this);
+    AreaListEnsureDataHydroTimeSeries(this);
+    AreaListEnsureDataThermalTimeSeries(this);
+    AreaListEnsureDataRenewableTimeSeries(this);
```
flomnes, line 1222: Make all these function private methods of class `AreaList`
- `AreaListEnsureDataLoadTimeSeries`
- ...
- `AreaListEnsureDataRenewableTimeSeries`
- `AreaListEnsureDataLoadPrepro`
- ...
- `AreaListEnsureDataThermalPrepro`

flomnes: You could also create a specific class for Prepro, and another one for TimeSeries

flomnes (changes requested):
Possible change of behavior, to be investigated

flomnes (approved):

PR #939, opened 2022-11-08
[CI] Centos ccache

### LINKED ISSUE #890: Oracle8 CI : use ccache + docker volume persistency
Since the build is set up within a docker image, persistence is a bit tricky.

However, mounting volumes allows persistence of data.

General idea https://stackoverflow.com/questions/39650056/using-ccache-when-building-inside-of-docker

Github workflow configuration https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

The only missing part seems to be the download/upload of cache data.

payetvin:
close #890


flomnes (changes requested):
@payetvin The cache must be downloaded/upladed on Github.

Issue #920, opened 2022-10-24
BC result aggregation


flomnes:
- Hourly BC : **OK**
- Daily & Weekly : **KO**
  Average, not sum. For each hour, take the daily MRG price.

PR #940, opened 2022-11-08
BC's marginal prices output correction

guilpier-code:
This fixes issue #920 

**Note** (important) : 
- This current PR fixes incorrect marginal prices results. 
Reference results are wrong, and that's why we get a CI failure.
We must correct reference results, but to do that, we need a new tag for Antares Simulator (like **v8.4.0-rc1** for instance). The need for an Antares new tag is part of the **adding new tests** or **correct reference results** procedure.
Therefore, before correcting the reference results, we should merge the current branch into the **develop** and create a tag.
- At this occasion, we should add some more tests regarding binding constraints' marginal prices


guilpier-code:
Please See **Note** section in current PR's introductory comment.

Issue #685, opened 2022-05-16
Move Kirchhoff constraint generator to an executable


flomnes:
This is required for AntaresWeb, since it's not (yet ?) possible to run a wxWidget app in the browser.

Search results for "Kirchhoff" show that this feature is very adherent to the current GUI
```bash
ag --nocolor --literal --line-number --smart-case --nogroup --column --stats -- Kirchhoff .
ui/simulator/application/main/menu.cpp:402:35:                     wxT("Run the Kirchhoff's constraints generator"),
ui/simulator/application/main/menu.cpp:404:35:                     wxT("Run the Kirchhoff's constraints generator"));
ui/simulator/windows/constraints-builder/constraintsbuilder.cpp:95:20:        this, wxT("Kirchhoff's constraints generator"), "images/32x32/network.png", wxT("")),
distrib/changelog.txt:419:3:- Kirchhoff's constraint generator: fixed several GUI issues
distrib/changelog.txt:427:21:- Implementation of Kirchhoff's laws (DC approximation),
distrib/changelog.txt:430:13:  dedicated Kirchhoff's constraints generator is now available
6 matches
3 files contained matches
```

The idea is to create a new executable CMake target that reads a config file, and produces the same constraints.

PR #948, opened 2022-11-14
Fix kirchhoff CI 

payetvin:
#685


flomnes (changes requested):
The CI still fails...

PR #960, opened 2022-11-21
Using clang

### LINKED ISSUE #951: Compile with clang/llvm
- Yuni defines weird compile flags, we'll need to clean them up
- It is impossible to build the project in Mac OS, since it uses clang
- Compare build/execution time

payetvin:
This PR aims at using clang instead of GCC.
The changes were made to prevent clang compiling error

close #951 

To use clang:
export CC=/usr/bin/clang
export CXX=/usr/bin/clang++
Then build in a new folder

# Benchmark:

run 1 and 2 on power, run 3 on battery
Each bench was run with a starting CPU temp around 40°C to avoid thermal throttling.

## Clang

run 1: 4m 28s
run 2: 4m 18s
run 3: 4m 28s

## GCC

run 1: 6m 22s
run 2: 6m 22s
run 3: 6m 49s

### DIFF
--- a/src/ext/yuni/src/yuni/core/system/stdint.h
+++ b/src/ext/yuni/src/yuni/core/system/stdint.h
@@ -135,8 +135,8 @@ typedef unsigned int uint;
 #ifdef YUNI_OS_CLANG
 /* Try to fix issues with clang which does not yet support __float128, and produces
 ** compilation error in the STL */
-typedef struct
-{
-    double x, y;
-} __float128;
+/* typedef struct __float128 */
+/* { */
+/*     double x, y; */
+/* } __float128; */
 #endif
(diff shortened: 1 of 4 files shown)


src/ext/yuni/src/yuni/core/system/stdint.h
```diff
@@ -135,8 +135,8 @@ typedef unsigned int uint;
...
 ** compilation error in the STL */
-typedef struct
-{
-    double x, y;
-} __float128;
+/* typedef struct __float128 */
+/* { */
+/*     double x, y; */
+/* } __float128; */
 #endif
```
flomnes: Remove the whole block, don't leave commented code

flomnes (changes requested):

flomnes (approved):

PR #785, opened 2022-07-20
Adequacy patch - CSR

hugo-antoine-rtei:
Generating an executable for beta 4 of adq patch

PR #961, opened 2022-11-21
Adequacy patch CSR - final

flomnes:
See #785 and #903 for remarks


src/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set(ANTARES_WEBSITE "https://antares-simulator.org/")
 set(ANTARES_ONLINE_DOC "https://antares-simulator.readthedocs.io/")
 
 # Beta release
-set(ANTARES_BETA 0)
+set(ANTARES_BETA 10)
```
guilpier-code, line 17: To be removed

src/solver/simulation/economy.h
```diff
@@ -47,7 +47,8 @@ class EconomyWeeklyOptimization
 {
 public:
     using Ptr = std::unique_ptr<EconomyWeeklyOptimization>;
-    virtual void solve(Variable::State& state, int hourInTheYear, uint numSpace) = 0;
+    virtual void solve(Variable::State& state, int hourInTheYear, uint numSpace, uint week) = 0;
+    virtual void solveCSR(const Variable::State& state, uint numSpace, uint week) = 0;
```
guilpier-code, line 51: We should not force the definition of **solveCSR** if it can be empty (in case of no adequacy patch).
**solve(...)** as the only pure virtual method is enough, and **solveCSR** should be called from `AdequacyPatchOptimization::solve(...)`

guilpier-code (changes requested):

src/solver/simulation/economy.cpp
```diff
@@ -71,7 +72,10 @@ AdequacyPatchOptimization::AdequacyPatchOptimization(PROBLEME_HEBDO** problemesH
  EconomyWeeklyOptimization(problemesHebdo)
 {
 }
-void AdequacyPatchOptimization::solve(Variable::State& state, int hourInTheYear, uint numSpace)
+void AdequacyPatchOptimization::solve(Variable::State& state,
```
guilpier-code, line 75: In **AdequacyPatchOptimization::solve(...)**, we should highlight the 3 steps of adq patch first step (local matching + classic optimization + curtailment sharing). We should have : 
```C++
void AdequacyPatchOptimization::solve(...)
{
     this->computeLocalMatchingRules();  // Step 1
     this->solveClassicOptimization(); // Step 2
     // Step 3
     this->prepareCurtailmentSharing(); // Actually computing DTG margin
     this->computeCurtailmentSharing();
}
```

guilpier-code (changes requested):

src/solver/simulation/economy.cpp
```diff
@@ -35,6 +35,8 @@
 #include "simulation.h"
 #include "../optimisation/opt_fonctions.h"
 #include "common-eco-adq.h"
+#include <set>
+#include "../optimisation/adequacy_patch.h"
```
guilpier-code, line 39: In function **Economy::simulationBegin()**, we can see that we create several **PROBLEME_HEBDO**, whereas only one **weeklyOptProblems** is created. 
Indeed, for each year in parallel, we reuse the same **weeklyOptProblems**. 
We should have one **weeklyOptProblems** per **PROBLEME_HEBDO**

guilpier-code (changes requested):

src/solver/optimisation/adequacy_patch.h
```diff
@@ -1,6 +1,6 @@
 /*
-** Copyright 2007-2018 RTE
-** Authors: Antares_Simulator Team
+** Copyright 2007-2022 RTE
+** Authors: RTE-international / Redstork / Antares_Simulator Team
```
guilpier-code, line 3: We should split **adequacy_patch.h/.cpp**  into : 
- **adq_patch_local_matching.h/.cpp**
- **adq_patch_curtailment_sharing.h/.cpp**

guilpier-code (changes requested):

PR #964, opened 2022-11-22
Refactor adq patch (part 2)

guilpier-code:
In the delivered adequacy patch work, the current PR reduces dependencies to big objects **state** and **study**.
Indeed, in many situations, it's overkill to pass these objects whereas we only need to use a small a part of them.


src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -987,19 +987,18 @@ void PartHydro::copyFrom(const PartHydro& rhs)
...
-    if ((int)(workVar.levelUp) == (int)(workVar.levelDown))
-        waterValueToReturn = waterValues[(int)(workVar.levelUp)][day];
+    if ((int)(levelUp) == (int)(levelDown))
+        waterValueToReturn = waterValues[(int)(levelUp)][day];
     else
         waterValueToReturn
-          = waterValues[(int)(workVar.levelUp)][day] * (level - workVar.levelDown)
-            + waterValues[(int)(workVar.levelDown)][day] * (workVar.levelUp - level);
+          = waterValues[(int)(levelUp)][day] * (level - levelDown)
+            + waterValues[(int)(levelDown)][day] * (levelUp - level);
```
flomnes, line 1001: nice !

src/solver/simulation/economy.cpp
```diff
@@ -106,12 +105,12 @@ weeklyOptimization::weeklyOptimization(PROBLEME_HEBDO** problemesHebdo) :
...
+void weeklyOptimization::solve(uint, int, uint numSpace)
 {
     auto problemeHebdo = pProblemesHebdo[numSpace];
     OPT_OptimisationHebdomadaire(problemeHebdo, numSpace);
 }
-void weeklyOptimization::solveCSR(const Variable::State& state, uint numSpace, uint week)
+void weeklyOptimization::solveCSR(Antares::Data::AreaList&, uint, uint, uint)
 {
     return;
 }
```
flomnes: No need for an abstract method. Use an empty method in the base class

flomnes (changes requested):

src/solver/simulation/economy.cpp
guilpier-code: Done

flomnes (approved):

flomnes (approved):

PR #967, opened 2022-11-23
[TEST] Activate automatic gcov report on SonarCloud

kathvargasr:
Sensor Generic Coverage Report on SonarCloud does not recognize the paths generated by custom target code-coverage.

More precisely, when running gcovr, paths are smoothen to its source directory. 

So by running gcovr through a custom target, we'll always have relative paths, and Sonar's coverage sensor only recognizes absolute paths

To bypass this issue, we're now running gcovr directly on workflow.


.github/workflows/sonarcloud.yml
```diff
@@ -128,9 +129,9 @@ jobs:
           cd $GITHUB_WORKSPACE/_build
           ctest -C Release --output-on-failure -L "unit"
 
-      - name: Compile coverage reports
+      - name: Collect coverage into one XML report
         run: |
-          cmake --build $GITHUB_WORKSPACE/_build --target code-coverage
+          gcovr --sonarqube > coverage.xml
```
flomnes: From the doc
```suggestion
          gcovr --sonarqube --output coverage.xml
```

sonar-project.properties
```diff
@@ -12,7 +12,8 @@ sonar.sources=src
 sonar.sourceEncoding=UTF-8
 
 sonar.exclusions=src/ext/**
+sonar.coverage.exclusions=src/ext/**,src/tests/**,src/analyzer/**,src/distrib/**,src/solver/**,src/tools/**,src/ui/**
```
flomnes: Don't exclude **src/solver**. The rest is OK.

.github/workflows/sonarcloud.yml
```diff
@@ -85,6 +85,7 @@ jobs:
         with:
           antares-deps-version: ${{steps.antares-deps-version.outputs.prop}}
           os: ${{matrix.os}}
+          buildtype: Debug
```
flomnes, line 88: unused, consider removing

flomnes (changes requested):

.github/workflows/sonarcloud.yml
kathvargasr: The example for cfamily from the docs suggested the > but we can try the option and see if it doesn't smooth the paths wrongly

flomnes: From `man gcovr`
```
       -o OUTPUT, --output OUTPUT
              Print output to this filename. Defaults to stdout.  Required for
              --html-details.
```
So redirection from stdout to `coverage.xml` and this option should be equivalent.

.github/workflows/sonarcloud.yml
kathvargasr: seems that dependencies cannot be built correctly without this. cmake fails with:

```
CMake Error at CMakeLists.txt:249 (message):
  Sirius solver not found.  Sirius solver can be compiled with
  -DBUILD_sirius=ON or you can specify previous dependency install directory
  with -DCMAKE_PREFIX_PATH or -DDEPS_INSTALL_DIR
```

flomnes: you are right, sorry

flomnes (approved):

PR #968, opened 2022-11-24
Adequacy patch : refactor economy (weekly optimization)

guilpier-code:
In the delivered adequacy patch work, this PR focuses on refactoring of source file **economy.h/.cpp**.
- We extract **weeklyOptimization** (classic opt) and **adqPatchWeeklyOptimization** from **economy.h/.cpp** into new source files
- We create one **weeklyOptimization** (adq patch or classic) per created **PROBLEME_HEBDO**.
- We also make corrections after a first review.


src/solver/simulation/economy.cpp
```diff
@@ -168,10 +169,13 @@ bool Economy::simulationBegin()
...
                 return false;
             }
-        }
 
-        weeklyOptProblem
-          = interfaceWeeklyOptimization::create(study.parameters.adqPatch.enabled, pProblemesHebdo);
+            weeklyOptProblems_.push_back(
+                    interfaceWeeklyOptimization::create(study.parameters.adqPatch.enabled,
+                                                        pProblemesHebdo[numSpace], 
+                                                        numSpace)
```
flomnes: I'm not sure quadratic opt are run in parallel, to be checked. It would be nice if they were.

With the current  implementation, there is no guarantee that `weeklyOptProblems_` contains the problem associated to index `numSpace`. I'd prefer something like this
```cpp
// before the for loop
weeklyOptProblems_.resize(pNbMaxPerformedYearsInParallel);
// inside the for loop
weeklyOptProblems_[numSpace] = interfaceWeeklyOptimization::create(/* ... */);
```

flomnes (changes requested):

src/solver/simulation/economy.cpp
guilpier-code: ok, I'll check for the quadratic optimization.
For the rest of your remark, I was thinking that the creation of a **PROBLEME_HEBDO** could even be done inside **WeeklyOptimization** constructor.  I guess it would make sense. What do you think ?
We would have something like : 
```C++
bool Economy::simulationBegin()
{
    if (!preproOnly)
    {
        {

            weeklyOptProblems_.push_back(
                    interfaceWeeklyOptimization::create(study.parameters.adqPatch.enabled, numSpace)
            );
        }
        SIM_InitialisationResultats();
    }
    ...
    return true;
}
``` 

But it would be too much work, as **pProblemesHebdo[numSpace]** is spread all over **economy.cpp**. 
Too bad.

flomnes: Yes, that's too ambitious for now. See also **adequacy.cpp**

guilpier-code: about `weeklyOptProblems_.resize(...)`, done

flomnes (approved):

PR #970, opened 2022-11-24
Using Clang on CI

payetvin:
This PR branched from the #960


.github/workflows/ubuntu.yml
```diff
@@ -105,11 +106,19 @@ jobs:
...
+      # if: ${{ matrix.ortools-xpress == 'ON' }}
       run: |
         git clone -b master https://${{ secrets.XPRS_USERNAME }}:${{ secrets.XPRS_TOKEN }}@github.com/rte-france/xpress-mp ${{ env.XPRESSDIR }}
         ls -la ${{ env.XPRESSDIR }}
 
+    - name: Configure compiler
+      run: |
+        if [ "${{ matrix.compiler }}" == "clang" ]; then
+          export CC=/usr/bin/clang
+          export CXX=/usr/bin/clang++
```
flomnes: Those env variables won't be available to the next steps.

See [here](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable).

.github/workflows/ubuntu.yml
```diff
@@ -55,13 +55,14 @@ jobs:
     - name: ccache
       uses: hendrikmuhs/ccache-action@v1.2
       with:
-          key: xpress-${{ matrix.ortools-xpress }}-${{ matrix.os }}
+          key: xpress-${{ matrix.compiler }}-${{ matrix.os }}
```
flomnes: OK

.github/workflows/ubuntu.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04]
         test-platform: [ubuntu-20.04]
-        ortools-xpress: [ON, OFF]
+        compiler: [gcc, clang]
```
flomnes: Please keep `ortools-xpress`

flomnes (changes requested):

.github/workflows/ubuntu.yml
```diff
@@ -55,13 +55,14 @@ jobs:
...
       with:
-          key: xpress-${{ matrix.ortools-xpress }}-${{ matrix.os }}
+          key: xpress-${{ matrix.compiler }}-${{ matrix.os }}
 
     - name: Install libraries
       run: |
            sudo apt-get update
            sudo apt-get install uuid-dev libwxgtk3.0-gtk3-dev
            sudo apt-get install libboost-test-dev
+           sudo apt-get install clang
```
flomnes: Only install if required

.github/workflows/ubuntu.yml
```diff
@@ -73,13 +74,13 @@ jobs:
...
             echo "ORTOOLS_URL=https://github.com/rte-france/or-tools/releases/download/v9.2-rte1.3/ortools_cxx_ubuntu-20.04_static_sirius_xprs.zip" >> $GITHUB_ENV
             echo "LABEL=-ortools-xpress" >> $GITHUB_ENV
-          else
-            echo "ORTOOLS_URL=https://github.com/rte-france/or-tools/releases/download/v9.2-rte1.3/ortools_cxx_ubuntu-20.04_static_sirius.zip" >> $GITHUB_ENV
-            echo "LABEL=" >> $GITHUB_ENV
-          fi
+          # else
+          #   echo "ORTOOLS_URL=https://github.com/rte-france/or-tools/releases/download/v9.2-rte1.3/ortools_cxx_ubuntu-20.04_static_sirius.zip" >> $GITHUB_ENV
+          #   echo "LABEL=" >> $GITHUB_ENV
+          # fi
```
flomnes: Keep this

flomnes (changes requested):

.github/workflows/ubuntu.yml
payetvin: It only needs to be there for configuration, it's stored in the Cmake cache after that

JasonMarechal25:
There is conflict managing Ortools Xpress [on|off] on one hand and compiler [gcc|clang] on the other hand.

.gitmodules
```diff
@@ -4,7 +4,7 @@
 [submodule "src/antares-deps"]
 	path = src/antares-deps
 	url = https://github.com/AntaresSimulatorTeam/antares-deps.git
-	branch = v2.0.2
+	branch = fix/clang-boost
```
flomnes: It shouldn't be necessary to change antares-deps. Just pass `-DBUILD_BOOST` to cmake's configure command ?

.github/workflows/ubuntu.yml
```diff
@@ -86,19 +90,32 @@ jobs:
...
+           fi
+
            cmake -B _build -S src \
            -DCMAKE_C_COMPILER_LAUNCHER=ccache \
            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
            -DDEPS_INSTALL_DIR=${{github.workspace}}/rte-antares-deps-Release \
            -DCMAKE_BUILD_TYPE=release \
            -DBUILD_TESTING=ON \
            -DBUILD_not_system=OFF \
+           -DBUILD_BOOST=ON \
```
flomnes, line 105: This should not be required

PR #978, opened 2022-12-01
[DEV] Refactored ortools_utils

### LINKED ISSUE #974: OR-Tools : some cleaning required
See file **src/solver/utils/ortools_utils.cpp**, class `OrtoolsUtils` allows to query solvers available through OR-Tools, using maps

- `_solverLinearProblemOptimStringMap`
- `_solverMixedIntegerProblemOptimStringMap`

These functions should be available in OR-Tools, see [here](https://github.com/rte-france/or-tools/blob/main/ortools/linear_solver/linear_solver.h) for these functions.

Two features should be implemented:

- List solvers
- See if the solver asked by the user is available

https://github.com/rte-france/or-tools/blob/8fe4afda604de5f69dccce50036667c702ac88dd/ortools/linear_solver/linear_solver.h#L265

payetvin:
close #974

Removing RTE wrapper functions 
Using native ortools functions instead


src/libs/antares/study/load-options.h
```diff
@@ -90,7 +90,7 @@ class StudyLoadOptions
     bool ortoolsUsed;
 
     //! Ortool solver used for simulation
-    OrtoolsSolver ortoolsEnumUsed = OrtoolsSolver::sirius;
+    operations_research::MPSolver::OptimizationProblemType ortoolsEnumUsed = operations_research::MPSolver::OptimizationProblemType::SIRIUS_LINEAR_PROGRAMMING;
```
flomnes:
```suggestion
    auto ortoolsEnumUsed = operations_research::MPSolver::SIRIUS_LINEAR_PROGRAMMING;
```

src/libs/antares/study/parameters.cpp
```diff
@@ -353,7 +353,7 @@ void Parameters::reset()
     hydroDebug = false;
 
     ortoolsUsed = false;
-    ortoolsEnumUsed = OrtoolsSolver::sirius;
+    ortoolsEnumUsed = operations_research::MPSolver::OptimizationProblemType::SIRIUS_LINEAR_PROGRAMMING;
```
```suggestion
    ortoolsEnumUsed = operations_research::MPSolver::SIRIUS_LINEAR_PROGRAMMING;
```

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +347,30 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
...
+
+    if (MPSolver::SupportsProblemType(MPSolver::SIRIUS_LINEAR_PROGRAMMING))
+        result.push_back((std::string)operations_research::ToString(MPSolver::SIRIUS_LINEAR_PROGRAMMING));
+    if (MPSolver::SupportsProblemType(MPSolver::SIRIUS_MIXED_INTEGER_PROGRAMMING))
+        result.push_back((std::string)operations_research::ToString(MPSolver::SIRIUS_MIXED_INTEGER_PROGRAMMING));
+
+    if (MPSolver::SupportsProblemType(MPSolver::CLP_LINEAR_PROGRAMMING))
+        result.push_back((std::string)operations_research::ToString(MPSolver::CLP_LINEAR_PROGRAMMING));
+    if (MPSolver::SupportsProblemType(MPSolver::CBC_MIXED_INTEGER_PROGRAMMING))
+        result.push_back((std::string)operations_research::ToString(MPSolver::CBC_MIXED_INTEGER_PROGRAMMING));
```
```cpp
for (auto solverType : {MPSolver::CBC_MIXED_INTEGER_PROGRAMMING,
/* ... */
MPSolver::CBC_MIXED_INTEGER_PROGRAMMING})
{
if (MPSolver::SupportsProblemType(solverType))
    result.push_back(solverType);
}

**Review by flomnes — CHANGES_REQUESTED** [review]

File: src/libs/antares/study/load-options.h (hunk 1 above)
Comment (payetvin) [reply src/libs/antares/study/load-options.h]: error: non-static data member declared with placeholder ‘auto’

**Comment by flomnes** [comment]
This work is a great improvement from the previous implementation. But I'm afraid we still need a small wrapper around OR-Tools. Let me explain why. Each solver comes in two flavors: linear programming (LP) and mixed programming (MILP or MIP).

We don't want the user to choose between these two flavors because it doesn't make sense. This decision depends on the presence of integer variables in the problem.

In the end, we want something like this

1. The user chooses a solver (for now xpress/sirius/cbc)
2. Antares Simulator chooses the flavor: LP or MILP

Based on that, we provide the real enum value e.g `SIRIUS_LINEAR_PROGRAMMING`.

NOTE: A MIP solver can solve an LP problem, but not the other way around. Though I'm not sure we want to go down this way.

**Comment by flomnes** [comment]
- Handle strings only for I/O : "sirius", "xpress", "coin" (linear => "clp" mixed => "cbc"), "glpk". No more custom enums
- **When loading the study**, if `ortoolsUsed == true`, check that the requested solver is available using `MPSolver::SupportsProblemType` and `MPSolver::ParseSolverType`

```cpp
// Entrée xpress/sirius/etc. (SANS précision MIXED/LINEAR)
// Choix de la variante

  enum Solver {
               sirius,
               xpress,
               cbc
  };

  const std::map<std::string, std::pair<std::string, std::string>> kMAP
    {
    {"xpress", std::pair("xpress_lp", "xpress")},
    {"sirius", std::pair("sirius_lp", "xpress")}
    };

  // Valider les chaînes de caractères dans la propriété 
  
MPSolver* factory()
{
  // Recherche dans la map, puis choix de la variante LP/MILP
  std::string type = "sirius_lp";
  return MPSolver::CreateSolver(type);
}
```

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +344,66 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
...
-    MPSolver::OptimizationProblemType::CBC_MIXED_INTEGER_PROGRAMMING;
-
-    _solverLinearProblemOptimMap[OrtoolsSolver::xpress]       =
-    MPSolver::OptimizationProblemType::XPRESS_LINEAR_PROGRAMMING;
-    _solverMixedIntegerProblemOptimMap[OrtoolsSolver::xpress] =
-    MPSolver::OptimizationProblemType::XPRESS_MIXED_INTEGER_PROGRAMMING;
+    {"xpress", std::pair("xpress_lp", "xpress")},
+    {"sirius", std::pair("sirius_lp", "sirius")},
+    {"coin", std::pair("clp", "cbc")},
+    {"glpk", std::pair("glpk_lp", "glpk")}
```
```suggestion
    {"xpress", {"xpress_lp", "xpress"}},
    {"sirius", {"sirius_lp", "sirius"}},
    {"coin", {"clp", "cbc"}},
    {"glpk", {"glpk_lp", "glpk"}}
```

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +344,66 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
...
-}
-
-MPSolver::OptimizationProblemType OrtoolsUtils::getLinearOptimProblemType(
-  const OrtoolsSolver& ortoolsSolver)
-{
-    // TODO JMK : see how we can get optimization problem type with current ortools RTE branch
-    // (can't use enum because of compile switch) MPSolver::OptimizationProblemType result =
-    // _solverLinearProblemOptimMap.at(ortoolsSolver)
+        MPSolver::OptimizationProblemType solverType;
+        MPSolver::ParseSolverType(solverName.second.first, &solverType);
```
flomnes: Check return value
`true` if success, `false` otherwise.

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +344,66 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
...
-    MPSolver::OptimizationProblemType result
-      = MPSolver::OptimizationProblemType::GLOP_LINEAR_PROGRAMMING;
+MPSolver* factory(const Antares::Optimization::PROBLEME_SIMPLEXE_NOMME *probleme, std::string solverName)
+{
+    std::list<std::string> solverList = getAvailableOrtoolsSolverName();
 
-    if (!MPSolver::ParseSolverType(_solverMixedIntegerProblemOptimStringMap[ortoolsSolver],
-                                   &result))
+    auto it = std::find(solverList.begin(), solverList.end(), solverName);
+    if (it != solverList.end())
```
flomnes: C++17 allows this

```suggestion
    if (auto it = std::find(solverList.begin(), solverList.end(), solverName); it != solverList.end())
```

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +344,66 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
...
     {
-        throw AssertionError("Unsupported Mixed Integer OrtoolsSolver for solver "
-                             + Enum::toString(ortoolsSolver));
+        Antares::logs.fatal() << "Solver doesn't exist";
+        return NULL;
```
flomnes: Throw as previously

src/solver/utils/ortools_utils.cpp
flomnes: There's no need to loop in this list.

`CreateSolver` returns `nullptr` in case the solver isn't supported.

You can do the following

1. Call `CreateSolver`
2. If equal to `nullptr`, return an error
3. Return the newly allocated object otherwise

src/libs/antares/CMakeLists.txt
```diff
@@ -567,6 +564,7 @@ target_link_libraries(libantares-core
 		libantares-core-calendar
 		libantares-solver-variable-info
         result_writer
+        ortools::ortools
```
flomnes: This shouldn't be necessary if we keep a `std::string` in class `Parameters`.

It was necessary when we kept an `MPSolver::OptimizationType`

src/solver/utils/ortools_utils.cpp
```diff
@@ -125,12 +125,8 @@ MPSolver* convert_to_MPSolver(
...
     auto& study = *Data::Study::Current::Get();
 
-    // Define solver used depending on study option
-    MPSolver::OptimizationProblemType solverType
-      = OrtoolsUtils().getLinearOptimProblemType(study.parameters.ortoolsEnumUsed);
-
-    // Create the linear solver instance
-    MPSolver* solver = new MPSolver("simple_lp_program", solverType);
+    // Create the MPSolver
+    MPSolver* solver = factory(problemeSimplexe, study.parameters.ortoolsSolver);
```
flomnes: 1. Using a factory here is a good idea
2. `factory` is too generic, choose a more specific name (it can contain factory)

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +344,66 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
...
-    _solverLinearProblemOptimMap[OrtoolsSolver::glop_cbc]       =
-    MPSolver::OptimizationProblemType::GLOP_LINEAR_PROGRAMMING;
-    _solverMixedIntegerProblemOptimMap[OrtoolsSolver::glop_cbc] =
-    MPSolver::OptimizationProblemType::CBC_MIXED_INTEGER_PROGRAMMING;
-
-            */
-}
-
-bool OrtoolsUtils::isOrtoolsSolverAvailable(OrtoolsSolver ortoolsSolver)
+std::list<std::string> getAvailableOrtoolsSolverNameLinearAndMixed()
```
flomnes: This method isn't used. Consider removing it.

src/solver/utils/ortools_utils.h
```diff
@@ -13,64 +12,33 @@ using namespace operations_research;
...
-     *
-     *  \return List of available ortools solver
-     */
-    std::list<OrtoolsSolver> getAvailableOrtoolsSolver();
+/*!
+ *  \brief Return list of available ortools solver name, linear and mixed
+ *
+ *  \return List of available ortools solver name
+ */
+std::list<std::string> getAvailableOrtoolsSolverNameLinearAndMixed();
```
flomnes: Remove this unused function

flomnes (changes requested):

src/solver/utils/ortools_utils.cpp
payetvin: We're parsing the strings defined in map which are the same in ortools so this call can't fail

flomnes (changes requested):
Command `solver840 ~/short/021\ Four\ areas\ -\ DC\ law/ --use-ortools --ortools-solver xpress`

Uses sirius, should use xpress.

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +344,46 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
     delete solver;
 }
 
-using namespace Antares::Data;
 
-OrtoolsUtils::OrtoolsUtils()
+const std::map<std::string, std::pair<std::string, std::string>> OrtoolsUtils::solverMap =
```
flomnes: here instead of a std::pair, use a small struct members LPSolverName and MIPSolverName.

This will make the rest of the code easier to understand than first/second

src/solver/utils/ortools_utils.cpp
```diff
@@ -348,178 +344,46 @@ void ORTOOLS_LibererProbleme(MPSolver* solver)
...
-    // (can't use enum because of compile switch) MPSolver::OptimizationProblemType result =
-    // _solverMixedIntegerProblemOptimMap.at(ortoolsSolver)
+    MPSolver *solver;
+    std::list<std::string> solverList = getAvailableOrtoolsSolverName();
 
-    // For now we are using string to get optimization problem type
-    MPSolver::OptimizationProblemType result
-      = MPSolver::OptimizationProblemType::GLOP_LINEAR_PROGRAMMING;
+    if (probleme->isMIP())
+        solver = MPSolver::CreateSolver((OrtoolsUtils::solverMap.at(solverName)).second);
```
flomnes: at may throw if key is not found. Please catch that error and call AntaresEmergencyShutdown

Also do this if solver==nullptr at the ens

flomnes (approved):

PR #1004, opened 2022-12-08
[FIX] Add var to avoid calling ortools function

payetvin:


flomnes (changes requested):
1. Rename `isXpress` to `solverSupportsWarmStart`
2. Add condition `&& solverSupportsWarmStart` for call to `SetStartingLpBasisInt` (see ortools_utils.cpp:289)

flomnes (approved):

PR #1011, opened 2022-12-14
[UI] Allow more than 100 thermal clusters

flomnes:
See #886

### DIFF
--- a/src/ui/simulator/windows/inspector/property.cluster.update.cpp
+++ b/src/ui/simulator/windows/inspector/property.cluster.update.cpp
@@ -101,26 +101,15 @@ bool ClusterUpdater::changeGroup(const wxVariant& value)
 bool ClusterUpdater::changeUnit(const wxVariant& value)
 {
     uint d = static_cast<uint>(value.GetLong());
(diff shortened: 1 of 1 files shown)


src/ui/simulator/windows/inspector/property.cluster.update.cpp
```diff
@@ -101,26 +101,15 @@ bool ClusterUpdater::changeGroup(const wxVariant& value)
 bool ClusterUpdater::changeUnit(const wxVariant& value)
 {
     uint d = static_cast<uint>(value.GetLong());
```
guilpier-code: Rename **d** into **nbUnits**

guilpier-code (changes requested):

src/ui/simulator/windows/inspector/property.cluster.update.cpp
flomnes: @guilpier-code DONE

payetvin (approved):

guilpier-code (approved):

PR #1015, opened 2022-12-14
Fix BC MARG COST values in parallel

### LINKED ISSUE #1013: BC MARG COST values incorrect when run in parallel
When running in parallel, results for individual MC years are different (typically 0) on some years.

Values when running in sequential are correct.

See enclosed study and results within [bc.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/10226793/bc.zip)

flomnes:


src/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
```diff
@@ -229,6 +226,7 @@ class BindingConstMarginCost
         if (!isInitialized())
             return;
 
+        auto numSpace = state.numSpace;
```
guilpier-code, line 229: I agree on every change of this PR.
That said, I have remarks not exactly related to it : 
- **BindingConstMarginCost** is the only output variable to actually do something in  **weekBegin** and **hourEnd**. Notice that **weekBegin** is not defined in any other output variable and variables for which **hourEnd** is defined, it does nothing. About **hourEnd**, we should remove its definition for output variables that define it.
-  **weekBegin** and **hourEnd** have a need for **numSpace** because they don't get it passed as argument. If we want to be consistent with  **numSpace** usage in other variables, we should consider passing **numSpace** to these 2 functions : it would not be much work. Otherwise, we have now  **numSpace** as a public member of **state**, and I guess (to be checked) there are functions defined in output variables that don't need  **numSpace** as argument.
- About **state** : it really seems like state is related to a year of a parallel set. So ideally, it should attached to a year. For now, it is an element of a vector of **State** objects instantiated in a high level function `ISimulation<Impl>::run()` (see **src\solver\simulation\solver.hxx**). We should not instantiate a vector of **State** at this level because it does not belong to this level : instead, we should instantiate a simple **State** object in `Economy::year(...)` (**economy.cpp**), where we are inside a year scope.

If we agree on these remarks, there are tickets to be created.

flomnes: I agree on all but the 3rd point.

Using a `State` for each year of a batch only allows to re-use the same data structure.

We don't need a `State` for each year, since after a batch is run and results have been exported, the data it contains is no longer useful. It can be used for the next batch.

This "hack" just saves us a few `malloc/free` behind the scenes.

guilpier-code (approved):

PR #1017, opened 2022-12-14
[DEV] Refactoring parallelization computation and remove duplication from solver

kathvargasr:
src/solver/simulation/solver.hxx's method buildSetsOfParallelYears and Study's getNumberOfCores had code duplication.
Parallelization computation is now managed by a separate class: SetsOfParallelYearCalculator

Although, due to dependencies on loading and calculations performed later at the study class, SetsOfParallelYearCalculator's results are still managed by the study and passed to the solver through study's member.


src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+**
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+
+#include <vector>
+#include <climits>
+#include <assert.h>
```
flomnes: C headers are deprecated in C++

```suggestion
#include <cassert>
```

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+}
+
+
+void SetsOfParallelYearCalculator::computeRawNbParallelYear()
+{
+    if (forceParallel)
+        return;
+
+    std::map<NumberOfCoresMode, int> numberOfMCYearThreads;
+    uint nbLogicalCores = Yuni::System::CPU::Count();
```
flomnes:
```suggestion
    const uint nbLogicalCores = Yuni::System::CPU::Count();
```

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+        areaNames.push_back(lname);
+    }
+
+}
+
+
+void SetsOfParallelYearCalculator::computeRawNbParallelYear()
+{
+    if (forceParallel)
+        return;
```
flomnes: Add a comment on why we don't need this computation if `forceParallel` is `true`

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,159 @@
...
+    // En fonction d'une �ventuelle play-list, certaines seront jou�es et d'autres non.
+
+public:
+    std::vector<unsigned int> setsSizes;
+    // Numeros des annees en parallele pour ce lot (certaines ne seront pas jou�es en cas de
+    // play-list "trou�e")
+    std::vector<unsigned int> yearsIndices;
+
+    // Une annee doit-elle �tre rejou�e ?
+    std::map<uint, bool> yearFailed;
```
flomnes: fix encoding here, use e.g [iconv](https://linux.die.net/man/1/iconv)

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+    }
+    
+}
+
+
+void SetsOfParallelYearCalculator::limitNbOfParallelYearsbyMinRefreshSpan()
+{
+    uint TSlimit = UINT_MAX;
+    if ((p.timeSeriesToGenerate & timeSeriesLoad) && (p.timeSeriesToRefresh & timeSeriesLoad))
+        TSlimit = p.refreshIntervalLoad;
```
flomnes, line 236: This one (load) is different from the others (thermal, solar, etc.). But it's the 1st one so I don't think it's a mistake.

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+    if ((p.timeSeriesToGenerate & timeSeriesSolar) && (p.timeSeriesToRefresh & timeSeriesSolar))
+        TSlimit = (p.refreshIntervalSolar < TSlimit) ? p.refreshIntervalSolar : TSlimit;
```
flomnes: This would be easier to read IMO
```suggestion
        TSlimit = std::min(p.refreshIntervalSolar,  TSlimit);
```
Same remark for the other ones

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+        TSlimit = (p.refreshIntervalWind < TSlimit) ? p.refreshIntervalWind : TSlimit;
+    if ((p.timeSeriesToGenerate & timeSeriesThermal) && (p.timeSeriesToRefresh & timeSeriesThermal))
+        TSlimit = (p.refreshIntervalThermal < TSlimit) ? p.refreshIntervalThermal : TSlimit;
+
+    if (TSlimit < forcedNbOfParallelYears)
+        forcedNbOfParallelYears = TSlimit;
+
+    // Limiting the number of parallel years by the total number of years
+    if (p.nbYears < forcedNbOfParallelYears)
+        forcedNbOfParallelYears = p.nbYears;
```
flomnes:
```cpp
forcedNbOfParallelYears = std::min({p.nbYears, TSlimit, forcedNbOfParallelYears});
```

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+}
+
+bool SetsOfParallelYearCalculator::isRefreshNeededForCurrentYear(uint y)
+{
+    bool refreshing = false;
+    refreshing = (p.timeSeriesToGenerate & timeSeriesLoad)
+                    && (p.timeSeriesToRefresh & timeSeriesLoad)
+                    && (!y || ((y % p.refreshIntervalLoad) == 0));
```
flomnes:
```suggestion
                    && ((y % p.refreshIntervalLoad) == 0);
```
since `!y` implies `(y % p.refreshIntervalLoad) == 0`, this is equivalent.

src/solver/simulation/solver.hxx
```diff
@@ -39,6 +39,7 @@
 #include "apply-scenario.h"
 #include <antares/emergency.h>
 #include "../ts-generator/generator.h"
+#include "../../libs/antares/study/parallel-years.h"
```
flomnes, line 42: This will invalidate CMake's dependency graph

src/solver/simulation/solver.utils.h
```diff
@@ -211,11 +179,14 @@ class annualCostsStatistics
         buffer << to_scientific(criterionCost1.costStdDeviation) << "\n";
         buffer << to_scientific(criterionCost1.costMin) << "\n";
         buffer << to_scientific(criterionCost1.costMax) << "\n";
+        // buffer << to_scientific(criterionCost1.getNbPerformedYears()) << "\n";
+
```
flomnes: remove this commented-out code

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+#define WIN32_LEAN_AND_MEAN
+
+#include <boost/test/included/unit_test.hpp>
+
+#include <antares/study/study.h>
+#include <antares/study/parallel-years.h>
+
+#include <iostream>
+#include <stdio.h>
```
flomnes: Probably not needed

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,291 @@
...
+{
+    uint Count()
+    {
+        // The number of processors online (capable of running processes)
+        return nbLogicalCores_;
+    }
+
+}
+}
+}
```
flomnes:
```suggestion
namespace Yuni::System::CPU
{
    uint Count()
    {
        // The number of processors online (capable of running processes)
        return nbLogicalCores_;
    }
}
```

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+{
+    
+    uint maxNbYearsOverAllSets = 0;
+    for (uint s = 0; s < setsOfParallelYears.size(); s++)
+    {
+        if (setsOfParallelYears[s].setsSizes.size() > maxNbYearsOverAllSets)
+            maxNbYearsOverAllSets = (uint)setsOfParallelYears[s].setsSizes.size();
+    }
+
+    forcedNbOfParallelYears = maxNbYearsOverAllSets; //j'aime pas
```
flomnes: ?

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,159 @@
...
+**
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#pragma once
+
+#include <vector>
+#include "../../../solver/simulation/solver.utils.h" // KVR se debarraser de ce include
```
flomnes: TODO

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+#include <antares/utils.hxx>
+#include "parameters.h"
+#include "parallel-years.h"
+
+namespace Antares{
+
+bool TempAreaListHolder::checkThermalTSGeneration(YString folder_)
+{
+    inputFolder = folder_;
+    loadAreaList();
```
flomnes: Add comment on why this is needed

flomnes (changes requested):

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,426 @@
...
+            if (ini.open(thermalPlant))
+            {
+                if(ini.firstSection)
+                {
+                    for (IniFile::Section* section = ini.firstSection; section; section = section->next)
+                    {
+                        for (IniFile::Property* property = section->firstProperty; property; property = property->next)
+                        {
+                            return property->key == "gen-ts";
+                        }
```
flomnes: Only the 1st loop iteration is executed because of the `return` statement.

src/libs/antares/CMakeLists.txt
```diff
@@ -447,6 +447,7 @@ set(SRC
 	study.h
 	study/estimate.memory-footprint.cpp
 	study/study.cpp
+	study/parallel-years.cpp
```
guilpier-code, line 450: Why **parallel-years.h** is not a part of the solution as well ?
Actually it should be, so that this source file is taken into account when searching key words over the whole  solution (for example under VS code).

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+void SetsOfParallelYearCalculator::computeRawNbParallelYear()
+{
+    // In case solver option '--force-parallel n' is used, this computation is not needed
+    // and n will remain the forcedNbOfParallelYears
+    if (forceParallel)
+        return;
+
+    std::map<NumberOfCoresMode, int> numberOfMCYearThreads;
+    const uint nbLogicalCores = Yuni::System::CPU::Count();
```
guilpier-code: As determining the number of logical cores **nbLogicalCores** is not strictly the mission (or responsibility) of your class **SetsOfParallelYearCalculator**, it should be done out of it, and the number of available cores could be given to the  **SetsOfParallelYearCalculator**'s constructor.
I guess it would make the tests on this call easier : for example you would not need to define a new imitating `Yuni::System::CPU::Count()` function, right ?

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,377 @@
...
+namespace utf = boost::unit_test;
+namespace tt = boost::test_tools;
+
+using namespace Yuni;
+using namespace Antares::Data;
+
+
+uint nbLogicalCores_{2};
+
+namespace Yuni::System::CPU
```
guilpier-code: See the remark for **src/libs/antares/study/parallel-years.cpp** about `Yuni::System::CPU::Count()`.

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,377 @@
...
+        // using default params
+        params.reset();
+	}
+
+	~Fixture(){}
+
+    //creating a study, maybe not needed
+    Study::Ptr study = std::make_shared<Study>();
+    Parameters params{};
+    Data::StudyLoadOptions options;
```
guilpier-code: About `Data::StudyLoadOptions options` : 
Why do we need this object ?
It really looks like we only need it when feeding the **SetsOfParallelYearCalculator**'s constructor in all tests. I understand that it gives a default value to **StudyLoadOptions**'s data (like **forceParallel** for instance), because of **StudyLoadOptions**'s constructor.
Even if we have to give each **StudyLoadOptions**'s data a value, and in each test (or we can find another solution based on fixture to set default values), I think a dependency of **Data::StudyLoadOptions** in the test source file is not good.
To be debated if you like.

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,377 @@
...
+	Fixture() 
+	{
+        // using default params
+        params.reset();
+	}
+
+	~Fixture(){}
+
+    //creating a study, maybe not needed
+    Study::Ptr study = std::make_shared<Study>();
```
guilpier-code: I commented the line : 
```C++
Study::Ptr study = std::make_shared<Study>();
```
and it still compiles. Seemingly we don't need the study in these tests.
So I also tried to remove `#include <antares/study/study.h>` on line 7.
But something strange happened : 
Line 8 (`#include <antares/study/parallel-years.h>`) does not compile anymore.
So I guess **parallel-years.h** is missing some header inclusions to be used independently from **study.h**.
Please see remarks about **parallel-years.h**

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,168 @@
...
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#pragma once
+
+#include <vector>
+
+namespace Antares
+{
+
+using namespace Data;
```
guilpier-code, line 34: As already pointed in the review of **test-parallel-years-calculator.cpp**, this file cannot be included (in another file) alone : it's missing some header inclusions. And probably **parallel-years.cpp** too.

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,377 @@
...
+    Parameters params{};
```
guilpier-code, line 30: In the same purpose of removing dependencies than in the remark below, I wonder if **params** (line 43), is needed as a dependency ?
**SetsOfParallelYearCalculator** needs **params** because it uses internally a few of its data.
**Question** : if we want to break **SetsOfParallelYearCalculator**'s dependency with the whole **Parameters** class, shouldn't we extend the list of arguments to be passed to the  **SetsOfParallelYearCalculator** 's constructor with all pieces of data that come from **Parameters** class ? 
I counted them, they are 12.
We could regroup them in one or more storing structures before passing them to the **SetsOfParallelYearCalculator** 's constructor. 
To be discussed.

src/libs/antares/study/parallel-years.h
guilpier-code: I tried : 
- including `#include <yuni/yuni.h>` **in parallel-years.h**
- including `#include "fwd.h"` **in parallel-years.cpp**

It should work

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,168 @@
...
+        bool forceParallel_,
+        bool enableParallel_,
+        uint forcedNbOfParallelYears_,
+        bool thermalTSRefresh_, // Temporary parameter needed. To be removed once TSGen will be detached from simulator
+        Parameters& params_)
+    : forceParallel{forceParallel_},
+    enableParallel {enableParallel_},
+    forcedNbOfParallelYears{forcedNbOfParallelYears_},
+    thermalTSRefresh{thermalTSRefresh_},
+    p{params_}{
```
guilpier-code: Maybe we should make the **build()** public and call it right after calling the constructor, so that the algorithm is removed from the constructor.

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,168 @@
...
+    std::vector<Yuni::String> areaNames;
+
+};
+
+class SetsOfParallelYearCalculator
+{
+public:
+    
+    SetsOfParallelYearCalculator(
+        bool forceParallel_,
```
guilpier-code: Usually, from what I know about C++ conventions, trailing underscores apply to classes' internal data.
And arguments of a constructor does not have this trailing underscore.
Here we have the reverse.

src/ui/simulator/windows/options/advanced/advanced.cpp
```diff
@@ -930,7 +930,7 @@ void AdvancedParameters::onSelectNumberOfCoresLevel(Data::NumberOfCoresMode ncMo
     {
         study->parameters.nbCores.ncMode = ncMode;
         // Force refresh for study->nbYearsParallelRaw
-        study->getNumberOfCores(false, 1 /* ignored */);
+        study->getNumberOfCores(false, false, 1 /* ignored */);
```
guilpier-code, line 919: **Question** : why do we suppose that the **getNumberOfCores**'s second argument (which is new here) is necessarily **false** ?

src/ui/simulator/application/study.cpp
```diff
@@ -320,7 +320,7 @@ class JobSaveStudy final : public Toolbox::Jobs::Job
 
         // Updating the number of logical cores to use when saving the study
         // so that the run window is up to date.
-        study->getNumberOfCores(false, 0);
+        study->getNumberOfCores(false, false, 0);
```
guilpier-code, line 322: Same question as in **advanced.cpp**'s review.

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,168 @@
...
+
+using namespace Data;
+
+struct setOfParallelYears
+{
+    // Un lot d'année à exécuter en parallèle.
+    // En fonction d'une éventuelle play-list, certaines seront jouées et d'autres non.
+
+public:
+    std::vector<unsigned int> setsSizes;
```
guilpier-code: **setsSizes** is used internally only. Shouldn't it be private ?
Besides, I do not understand the purpose of this variable.
Do we have to add this variable ?

guilpier-code (changes requested):

src/libs/antares/study/parallel-years.cpp
JasonMarechal25: The number of core is no injected to SetsOfParallelYearCalculator constructor

flomnes: now ?

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
JasonMarechal25: It is a valid point. I've removed the dependency.
The Data::StudyLoadOptions structure is kind of a bag to dump thing that don't feel like belonging with one another. In our case parallelism option are not so much as study option than execution/performance options

src/libs/antares/study/parallel-years.h
JasonMarechal25: By doing this you encounter a code smell known as temporal coupling : https://blog.ploeh.dk/2011/05/24/DesignSmellTemporalCoupling/

src/ui/simulator/windows/options/advanced/advanced.cpp
JasonMarechal25: I don't know. Honestly it feels like the lesser issue of "getNumberOfCores"
- It is in the study
- It's called "get" but doesn't return anything
- Last parameter is ignored if the first one is false

src/libs/antares/study/parallel-years.h
JasonMarechal25: setOfParallelYears is a struct (not a class) so it behaves as a "bag of data" and doesn't really have usage for private members.
However the point on the utility of setsSizes is good. It acts as a "computed value"/cache value to avoid recalculating the value every time we need it. We only ever use the size of the vector as meaningful value so I change the type from vector to a simple uint counter.

src/libs/antares/study/study.cpp
```diff
@@ -224,284 +224,25 @@ uint64 Study::memoryUsage() const
...
-        }
-    } // End if hot start
+void Study::getNumberOfCores(const bool forceParallel, const bool enableParallel, const uint nbYearsParallelForced) {
+
+    TempAreaListHolder holder{};
+    bool thermalTSRefresh = holder.checkThermalTSGeneration(folderInput);
+
+    SetsOfParallelYearCalculator setsBuilder(forceParallel,
+                                             enableParallel,
+                                             nbYearsParallelForced, Yuni::System::CPU::Count(),
```
guilpier-code: Move `Yuni::System::CPU::Count()` to next line, for more readability

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,165 @@
...
+    bool isRefreshNeededForCurrentYear(uint y);
+  
+    std::vector<setOfParallelYears> setsOfParallelYears;
+   
+    bool forceParallel_;
+    bool enableParallel_;
+    uint forcedNbOfParallelYears_;
+    uint number_of_cores_;
+    bool thermalTSRefresh_;
+    Parameters& p;
```
guilpier-code, line 166: **p** should be renamed **parameters_**

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,165 @@
...
+{
+
+using namespace Data;
+
+struct setOfParallelYears
+{
+    // Un lot d'année à exécuter en parallèle.
+    // En fonction d'une éventuelle play-list, certaines seront jouées et d'autres non.
+
+    unsigned int setsSizes = 0;
```
guilpier-code, line 41: We don't need **setsSizes**, we already have **nbPerformedYears** as a data member in this **struct**.
It contains exactly the same information and has a better name.
Besides, the use of this data member in class **SetsOfParallelYearCalculator** is a regression compared to the previous version of the parallel years orchectrator  : the previous counterpart of **setsSizes** was the size of a set, containing also years not performed (disabled by the **playlist**).

A question is : what should we use instead of **setsSizes** ? 
Should we use the set's size or the **set->nbPerformedYears** ? (Especially if we deal with the **hydro hot start**, see function **allSetsParallelYearsHaveSameSize**)

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,165 @@
...
+  
+    std::vector<setOfParallelYears> setsOfParallelYears;
+   
+    bool forceParallel_;
+    bool enableParallel_;
+    uint forcedNbOfParallelYears_;
+    uint number_of_cores_;
+    bool thermalTSRefresh_;
+    Parameters& p;
+    uint nbYearsReallyPerformed{0};
```
guilpier-code, line 167: **nbYearsReallyPerformed** : 
To be strict, this data should not belong to class **SetsOfParallelYearCalculator**.
The job of this class is to define the years that will be run at the same time (each in a different thread).
This class does not mean counting the performed years over the whole simulation.
But this class currently takes advantages of its own treatments to compute this number.

By the way, this number is already computed in other contexts : 
- **src/libs/antares/benchmarking/info_collectors.cpp** : see function **performedYearsCountToFileContent(...)**
- **parameters.cpp** : see function **prepareForSimulation(...)**, especially look for local variable **effectiveNbYears** 

We should remove  **nbYearsReallyPerformed** from the class and compute it somewhere else.
For now, I can't see another location to hold and compute it than the **parameters.cpp**, because it already contains the data member **yearsFilter** (see this data).

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+        {
+            buildNewSet = true;
+            foundFirstPerformedYearOfCurrentSet = false;
+            if (set->nbPerformedYears > forcedNbOfParallelYears_)
+                forcedNbOfParallelYears_ = set->nbPerformedYears;
+        }
+        else
+            buildNewSet = false;
+
+     // End of loop over years
```
guilpier-code, line 369: This comment  : 
`// End of loop over years`
should be moved to the same line we close the bracket of the loop

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+        if (p.mode == Antares::Data::stdmAdequacyDraft){
+            forcedNbOfParallelYears_ = 1;
+        }
+
+        // In case parallel mode was not chosen, only 1 core is allowed
+        if (!enableParallel_ && !forceParallel_){
+            forcedNbOfParallelYears_ = 1;
+        }
+
+        if (indexSpace == forcedNbOfParallelYears_ - 1 || y == p.nbYears - 1)
```
guilpier-code, line 360: It seems that some regressions were introduced here.
We should take a look at what was done before.
What makes me think this is that the following if : 
```C++
if (set->nbPerformedYears > forcedNbOfParallelYears_)
     forcedNbOfParallelYears_ = set->nbPerformedYears;
``` 
seems impossible to reach, as  `set->nbPerformedYears` cannot be > than `forcedNbOfParallelYears_`, which is ill named by the way : it is supposed to be the max number of years in a set of parallel years.

guilpier-code (changes requested):

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,165 @@
...
+    [[nodiscard]] uint computeMinNbParallelYears() const;
+    void computeForcedNbYearsInParallelYearSet();
+
+    bool isRefreshNeededForCurrentYear(uint y);
+  
+    std::vector<setOfParallelYears> setsOfParallelYears;
+   
+    bool forceParallel_;
+    bool enableParallel_;
+    uint forcedNbOfParallelYears_;
```
guilpier-code, line 162: **forcedNbOfParallelYears_** : 
Following the computation of this data member through the **build()** method, we realize that we compute the maximum possible number of **performed** years we can find in a set of parallel years.
So it is a limit max for the number of performed years in a set of parallel years.
And it should not to be confused with the actual maximum number of performed year over all sets.
So why not call this data **limitForNbOfPerformedYearsInASet_** ?

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+        // In case the study is run in the draft mode, only 1 core is allowed
+        if (p.mode == Antares::Data::stdmAdequacyDraft){
+            forcedNbOfParallelYears_ = 1;
+        }
+
+        // In case parallel mode was not chosen, only 1 core is allowed
+        if (!enableParallel_ && !forceParallel_){
+            forcedNbOfParallelYears_ = 1;
+        }
```
guilpier-code, line 357: About the code snippet : 
```C++
// In case the study is run in the draft mode, only 1 core is allowed
  if (p.mode == Antares::Data::stdmAdequacyDraft){
         forcedNbOfParallelYears_ = 1;
  }

 // In case parallel mode was not chosen, only 1 core is allowed
 if (!enableParallel_ && !forceParallel_){
     forcedNbOfParallelYears_ = 1;
 }
```
This **forcedNbOfParallelYears_** update for special cases should really not be inside de loop over the years : doing this, we update it every year with the same value.
At least, we should move it at the beginning of the current function.
A better choice would be to place these updates in a special method, called from the **build()** method.

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    [[nodiscard]] uint getNbYearsReallyPerformed() const
+    {
+        return nbYearsReallyPerformed;
+    }
+
+private:
+    
+    void build();
```
guilpier-code, line 146: About the **build()** method : 
- Make the **build()** public
- We should not call it from the constructor
- but call it from outside, after an object of type **SetsOfParallelYearCalculator** is instantiated

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+
+        if (indexSpace == forcedNbOfParallelYears_ - 1 || y == p.nbYears - 1)
+        {
+            buildNewSet = true;
+            foundFirstPerformedYearOfCurrentSet = false;
+            if (set->nbPerformedYears > forcedNbOfParallelYears_)
```
guilpier-code, line 363: About snippet : 
```C++
   if (set->nbPerformedYears > forcedNbOfParallelYears_)
      forcedNbOfParallelYears_ = set->nbPerformedYears;
```
This piece of code seems to be useless and even error prone : **forcedNbOfParallelYears_** is the max number of years in a set of parallel years. It was computed earlier in the execution and it is a limit for the sizes of sets. So it should not be changed for any reason while we are building these sets.
Besides, I can't see how the condition in the **if** can ever be **true** : this **if** is located inside another one, and the conditions of these 2 **if** statements are incompatible.
So this snippet should be removed.

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+            // Number of actually performed years in the current set (up to now).
+            set->nbPerformedYears++;
+            // Index of the MC year's space (useful if this year is actually run)
+            indexSpace = set->nbPerformedYears - 1;
+
+            set->isYearPerformed[y] = true;
+            set->performedYearToSpace[y] = indexSpace;
+            set->spaceToPerformedYear[indexSpace] = y;
+
+            if (!foundFirstPerformedYearOfCurrentSet)
```
guilpier-code, line 336: About the snippet : 
```C++
if (!foundFirstPerformedYearOfCurrentSet)
 {
       set->isFirstPerformedYearOfASet[y] = true;
       foundFirstPerformedYearOfCurrentSet = true;
 }

...

// And when we reach the end of the current set, and we have to start another one : 
foundFirstPerformedYearOfCurrentSet = false;

```
This is a complicated mechanism to mark the first performed year of a set.
A simpler way to do this would be to insert few lines above :  
```C++
if (buildNewSet)
 {
       // Start a new set here                
       ...

       // Initializations here
       //...
      if (performCalculations)
           set->isFirstPerformedYearOfASet[y] = true;
     ...
}
```
And that's it. So no need for local variable **foundFirstPerformedYearOfCurrentSet**

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+        uint setSize = (uint)setsOfParallelYears[s].setsSizes;
+        // Empty sets are not taken into account because, on the solver side,
+        // they will contain only skipped years
+        if (setSize && (setSize < minNbYearsInParallel))
+            minNbYearsInParallel = setSize;
+    }
+    return minNbYearsInParallel;
+}
+
+void SetsOfParallelYearCalculator::computeForcedNbYearsInParallelYearSet()
```
guilpier-code, line 403: About `computeForcedNbYearsInParallelYearSet()` : 
this function is ill named, it should be called : `computeActualMaxNbOfPerformedYearsInASet()`
Indeed, before calling this function from the **build()** function, previously called functions define the sets of parallel years by constraining the max number of **performed** (enabled in the playlist) years in a set with a limit, which is **forcedNbOfParallelYears_** (but should be renamed into **limitForNbOfPerformedYearsInASet_** to be accurate).
The real intention of the current function (according to its body) is to compute the max number **performed** years actually contained in a set. 
Note that this number can be different from  **limitForNbOfPerformedYearsInASet_** (for example if the study does not contain enough years to reach the limit).
And yet, this computed number is stored in  **limitForNbOfPerformedYearsInASet_** (or **forcedNbOfParallelYears_**, sorry for the all these names), which is not right.

So to sum up, I think we should : 
- rename the function into `computeActualMaxNbOfPerformedYearsInASet()`
- Put the result of its computation into a new data member called `actualMaxNbOfPerformedYearsInASet_`

guilpier-code (changes requested):

guilpier-code:
Also, I spotted a bug (probably a bug in the GUI)  :
A study is attached, with **5** MC years.
Trying to run it in parallel mode.
Just before running, look at the **simulation cores** : it is **0**.
When the study is run, a set of 4 years is run, then a set of 1 year, as expected.

Please find the study here : 
[021 Four areas - DC law.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/10548737/021.Four.areas.-.DC.law.zip)

src/ui/simulator/windows/options/advanced/advanced.cpp
guilpier-code: I do agree.
See review of **src/libs/antares/study/study.cpp** (https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1017#discussion_r1093524571) for my opinion on this

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+        = ((p.timeSeriesToGenerate & timeSeriesThermal)
+                        && (p.timeSeriesToRefresh & timeSeriesThermal)) || thermalTSRefresh_;
+    refreshing
+        = refreshing || (haveToRefreshTSThermal && (y % p.refreshIntervalThermal == 0));
+
+    return refreshing;
+
+}
+
+void SetsOfParallelYearCalculator::buildSetsOfParallelYears()
```
guilpier-code, line 278: I think we could make this algorithm drier, and reject some data member assignement in a further method.
For instance, `set->nbYears` can be computed later, as well as `set->yearFailed` or 
`set->isFirstPerformedYearOfASet` (not exhaustive list).
It would clarify this algorithm.

guilpier-code (changes requested):

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+BOOST_AUTO_TEST_CASE(default_params_no_force)
+{
+    bool thermalTSRefresh = false;
+
+    params.nbCores.ncMode = Antares::Data::NumberOfCoresMode::ncHigh;
+
+    SetsOfParallelYearCalculator builder(forceParallel,
+                                         enableParallel,
+                                         maxNbYearsInParallel, 1,
```
guilpier-code, line 47: The number of available cores on the machine is quite hidden : here it's **1**.
I would recommend to declare a new variable `uint machine_nb_cores = 1;` and pass it to the constructor below.

guilpier-code (changes requested):

src/tests/src/libs/antares/study/parallel-years/test-parallel-years-calculator.cpp
```diff
@@ -0,0 +1,354 @@
...
+    // Create playlist
+    params.userPlaylist = true;
+    params.yearsFilter.reserve(params.nbYears);
+    for (uint i = 0; i != params.nbYears; ++i)
+        params.yearsFilter[i] = false;
+    params.yearsFilter[2] = true;
+    params.resetYearsWeigth();
+    params.setYearWeight(0,4);
+    params.setYearWeight(1,10);
+    params.setYearWeight(2,3);
```
guilpier-code, line 294: Why do we need to set some years' weight to test the parallel years orchestrator (or builder) ?
It seems to me that year weights and  parallel years orchestrator are not related.

guilpier-code (changes requested):

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+namespace Antares
+{
+
+/*!
+** \brief Checks if clusters have the "gen-ts" param
+**
+** \warning TODO: REMOVE THIS METHOD ONCE THE TSGEN HAS BEEN DECOUPLED FROM SOLVER
+*/
+
+bool TempAreaListHolder::checkThermalTSGeneration(YString folder_)
```
guilpier-code, line 54: Thermal clusters are special regarding TS refresh. 
As for any other production mean, clusters have their own global parameter to enable time series refreshment. But user can ask the refresh to any particular cluster, that is not to obey to the global parameter.
Here we check if **any** cluster has to refresh its TS,  and we return true in that case, false otherwise.

**Question** : why get this information form input files, and not from memory ?
Probably because when we build the parallel years orchestrator, this information on clusters is not loaded yet.
In that case, why not build the parallel years orchestrator after the information on clusters is loaded ?

guilpier-code (changes requested):

src/libs/antares/study/study.cpp
```diff
@@ -224,284 +224,25 @@ uint64 Study::memoryUsage() const
...
+                                             parameters);
+    // For GUI
+    minNbYearsInParallel_save = setsBuilder.getMinNbParallelYearsForGUI();
+    maxNbYearsInParallel_save = setsBuilder.getForcedNbOfParallelYears();
+
+    // For the solver
+    maxNbYearsInParallel = setsBuilder.getForcedNbOfParallelYears();
+    parameters.allSetsHaveSameSize = setsBuilder.allSetsParallelYearsHaveSameSize();
+    setsOfParallelYears = setsBuilder.getSetsOfParallelYears();
+    pNbYearsReallyPerformed = setsBuilder.getNbYearsReallyPerformed();
```
guilpier-code, line 247: About : 
```C++
pNbYearsReallyPerformed = setsBuilder.getNbYearsReallyPerformed();
```
If we agree that computing the number of performed years in the study is not part of the responsibility of the parallel years orchestrator, this line should be removed.

guilpier-code:
About the file **run.cpp**, which has not been touched in the PR, but should be commented as well.
This file is about the run simulation window in the GUI.
Consider the snippet : 
```C++
    // Needed For RAM estimation
    uint& maxNbCores = Data::Study::Current::Get()->maxNbYearsInParallel;
    uint maxNbCoresParallelMode = Data::Study::Current::Get()->maxNbYearsInParallel_save;

    // Needed for run window's simulation cores field
    uint& minNbCores = Data::Study::Current::Get()->minNbYearsInParallel;
    uint minNbCoresParallelMode = Data::Study::Current::Get()->minNbYearsInParallel_save;

    if (featuresAlias[pFeatureIndex] == Solver::parallel)
    {
        maxNbCores = maxNbCoresParallelMode; // For RAM estimation
        minNbCores = minNbCoresParallelMode; // For run window's simulation cores field
        ...
    }
    else
    {
        maxNbCores = 1; // For RAM estimation
        minNbCores = 1; // Not needed (because simulation cores field is hidden here)
        ....
    }
```
This snippet sets the variables **minNbCores** and **maxNbCores**, which are later used in the window to print a line of information like : 

**Simulation cores : 4 (smallest batch size : 1)** 

So back to the snippet : **minNbCores** is set to **1** and **maxNbCores** is set to **4**.
Note that these 2 variables are references to actual variables inside the study. 
So changing them here has an effect on the study state.
But what's the point of changing the study regarding these variables in the study ? 
Indeed, if we run the study, these 2 values will be re-computed by the solver in a separate process.
So there is no need to change the study internally.
For me we should have : 
```C++
    uint minNbCores = 0;
    uint maxNbCores = 0;
    if (featuresAlias[pFeatureIndex] == Solver::parallel)
    {
        maxNbCores = Data::Study::Current::Get()->maxNbYearsInParallel; // For RAM estimation
        minNbCores = Data::Study::Current::Get()->minNbYearsInParallel; // For run window's simulation cores field
        ...
    }
        else
    {
        // No need to set minNbCores or maxNbCores here
        ....
    }
```

src/libs/antares/study/study.cpp
```diff
@@ -224,284 +224,25 @@ uint64 Study::memoryUsage() const
...
+    bool thermalTSRefresh = holder.checkThermalTSGeneration(folderInput);
+
+    SetsOfParallelYearCalculator setsBuilder(forceParallel,
+                                             enableParallel,
+                                             nbYearsParallelForced, Yuni::System::CPU::Count(),
+                                             thermalTSRefresh,
+                                             parameters);
+    // For GUI
+    minNbYearsInParallel_save = setsBuilder.getMinNbParallelYearsForGUI();
+    maxNbYearsInParallel_save = setsBuilder.getForcedNbOfParallelYears();
```
guilpier-code, line 241: About : 
```C++
maxNbYearsInParallel_save = setsBuilder.getForcedNbOfParallelYears();
```
The data member **maxNbYearsInParallel_save**  is useless, as it is the same as **maxNbYearsInParallel**, except that it is used by the GUI to print some information in the run simulation window.
We should remove this line.

guilpier-code (changes requested):

src/libs/antares/study/study.cpp
```diff
@@ -224,284 +224,25 @@ uint64 Study::memoryUsage() const
...
+    TempAreaListHolder holder{};
+    bool thermalTSRefresh = holder.checkThermalTSGeneration(folderInput);
+
+    SetsOfParallelYearCalculator setsBuilder(forceParallel,
+                                             enableParallel,
+                                             nbYearsParallelForced, Yuni::System::CPU::Count(),
+                                             thermalTSRefresh,
+                                             parameters);
+    // For GUI
+    minNbYearsInParallel_save = setsBuilder.getMinNbParallelYearsForGUI();
```
guilpier-code, line 239: We don't need to name **minNbYearsInParallel_save** as it is. 
**minNbYearsInParallel** is enough.
Note this value is only read by GUI to print some information about the minimum size there is over all set (of parallel years)

src/libs/antares/study/study.cpp
```diff
@@ -224,284 +224,25 @@ uint64 Study::memoryUsage() const
...
-            {
-                if (setsOfParallelYears[s].size() != currentSetSize)
-                {
-                    parameters.allSetsHaveSameSize = false;
-                    break;
-                }
-            }
-        }
-    } // End if hot start
+void Study::getNumberOfCores(const bool forceParallel, const bool enableParallel, const uint nbYearsParallelForced) {
```
guilpier-code, line 227: The function **getNumberOfCores(...)** updates some **Study** data members that are used elsewhere either in the GUI or in the solver to run the simulation properly.
Ideally, we should have an instance of  **SetsOfParallelYearCalculator** that would contain and would be able to give all information needed about parallel sets to any part of the code that requests them.
I can't see anywhere else in the code to put this instance than as a **Study's** data member. 
Indeed, as a singleton, the study can be reached from anywhere in the code, and deliver information.
But if some part of code needs information about parallel years, it should retrieve them through the instance `study->parallelYearCalculator`, and not directly from the study itself.

Furthermore, instead of having **getNumberOfCores(...)**, `study->parallelYearCalculator` should have a different interface : 
- We could have at least 2 **updates(...)** methods : 
    - one for a default **update()** : calculating the sets of parallel years with the characteristics of the study
    - one **update(....)** in case we have external information to compute the sets of parallel years (I mean information from the solver command line option to force the number of cores used)
- some get functions to give information to any part of code requiring it.
- Any other methods should be private, especially those that are used to compute the sets of parallel years

With this, we would encapsulate the information about parallel years, and furthermore we would avoid some unclean lines that can be found in the code like : 
`study->getNumberOfCores(false, false, 0); ` 
Instead, we would have : 
`study->parallelYearCalculator->update()`
which is cleaner I guess.

guilpier-code (changes requested):

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+            buildNewSet = false;
+
+     // End of loop over years
+
+    }
+
+}
+
+
+bool SetsOfParallelYearCalculator::allSetsParallelYearsHaveSameSize()
```
guilpier-code, line 375: About **allSetsParallelYearsHaveSameSize()** : 
This function answer the question : does all sets contains the same number of performed years ?
For now It is only called in case the study enables the **hydro hot start**. 
And that is reflected in the body of this function.
Well it should not : the function should stick to its name and answer the simple question without knowing if it's called in the context of **hydro hot start** or not.
So the body of this function should be : 
```C++
bool SetsOfParallelYearCalculator::allSetsParallelYearsHaveSameSize()
{
    uint currentSetSize = (uint)setsOfParallelYears[0].setsSizes;
    return all_of(setsOfParallelYears.begin(), setsOfParallelYears.end(),
                    [currentSetSize](const setOfParallelYears& v){ return v.setsSizes == currentSetSize; } );
}
```
and that's it.
Another benefit of this simplification is that  `p.initialReservoirLevels.iniLevels` is not required anymore in class **SetsOfParallelYearCalculator**, so we loose a dependency on a study parameter.

guilpier-code (changes requested):

src/libs/antares/study/parallel-years.h
```diff
@@ -0,0 +1,165 @@
...
+    number_of_cores_{number_of_cores},
+    thermalTSRefresh_{thermalTSRefresh},
+    p{params}{
+        this->build();
+    }
+
+        
+    bool allSetsParallelYearsHaveSameSize();
+
+    [[nodiscard]] uint getForcedNbOfParallelYears() const
```
guilpier-code, line 119: As already explained in review of **parallel-years.cpp**, we return **actualMaxNbOfPerformedYearsInASet_**, and rename this getter accordingly

guilpier-code (changes requested):

src/libs/antares/study/parallel-years.cpp
```diff
@@ -0,0 +1,431 @@
...
+    }
+    catch(const std::out_of_range& e)
+    {
+        logs.fatal() << "Simulation cores level not correct : " << (int)p.nbCores.ncMode;
+    }
+    
+}
+
+
+void SetsOfParallelYearCalculator::limitNbOfParallelYearsbyMinRefreshSpan()
```
guilpier-code, line 232: About functions **limitNbOfParallelYearsbyMinRefreshSpan** and **isRefreshNeededForCurrentYear** : 
- there are some duplication code in both functions
- they imply a dependency of class **SetsOfParallelYearCalculator** to **study->parameter**, making this class more difficult to test in isolation that it should.

Just a first think (to be digged)
So we could have a parent class **refreshSpanHandler** : 
```C++
class refreshSpanHandler
{
     getSpan() = 0;
     doWeRefresh(int year) = 0;
}
```
Then derive it for all kinds of production or load.
Maybe build them (create instance of them) outside the class **SetsOfParallelYearCalculator**.
Add them to the parallel years calculator, that is put them in a vector of parent class **refreshSpanHandler**, let's say **refreshSpanHandlerList**)
Then we could traverse this vector with an std::algorithm.
example : 
```C++
 refreshing = std::any_of(refreshSpanHandlerList.begin(), refreshSpanHandlerList.end(), [](const refreshSpanHandler h) { return h->doWeRefresh(y)});
```
Note we would avoid class **SetsOfParallelYearCalculator** a dependency to **study.parameter** in that way.

guilpier-code (changes requested):

src/libs/antares/study/parallel-years.h
flomnes: In general, having `using ...` is a bad idea in headers because it forces it on the .cpp file that `#includes`. I suggest removing this clause.

flomnes:
> Also, I spotted a bug (probably a bug in the GUI) : A study is attached, with **5** MC years. Trying to run it in parallel mode. Just before running, look at the **simulation cores** : it is **0**. When the study is run, a set of 4 years is run, then a set of 1 year, as expected.
> 
> Please find the study here : [021 Four areas - DC law.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/10548737/021.Four.areas.-.DC.law.zip)

**smallest batch size** is incorrect as well (always 1)

guilpier-code:
This PR still has some interest, its comments / remarks should be read, and the subject should be taken care of.

PR #1040, opened 2022-12-27
Refactor : about optimization number

guilpier-code:
This is a refactoring work about the optimization number.
- Information of optimization number is spread at 2 locations : it can be found inside **ProblemeHebdo** and inside **ProblemeHebdo->ProblemeAResoudre**
- This piece of information does not need to be nested in these structures 

So it is now declared and defined once, and locally to high level function **OPT_OptimisationLineaire**, and then passed to lower functions when needed.

Note that we used to define variable **ProblemeHebdo->numeroOptimisation** as an array of which dimension is 7 (the number of days in a week). This array is not necessary, so when relocating the optimization number, we do without an array anymore.

I hope the code is more simple and easier to understand now.

### DIFF
--- a/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
+++ b/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
@@ -95,7 +95,9 @@ private:
     clock::time_point end_;
 };
 
-bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, int NumIntervalle)
+bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo,
+                         int NumIntervalle,
+                         const int numeroOptimisation)
 {
     int Var;
     int Cnt;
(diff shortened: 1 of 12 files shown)


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -95,7 +95,9 @@ class TimeMeasurement
     clock::time_point end_;
 };
 
-bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo, int NumIntervalle)
+bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* ProblemeHebdo,
+                         int NumIntervalle,
+                         const int numeroOptimisation)
```
flomnes: No more French, please. When creating new variables, arguments, etc. prefer English

flomnes (changes requested):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp

flomnes (approved):

PR #1042, opened 2022-12-29
De-duplicate OPT_Export_add_variable by using `std::optional<int>`

flomnes:
### Before
Two functions, one with an additional `int` argument `secondVal`

### After
One single function, with an `optional` argument. It works much like a pointer
```cpp
std::optional<int> p;
if (p) {
   int& v = *p;
}
// equivalent
if (p.has_value())
{
   int& v = p.value();
}
```

### DIFF
--- a/src/solver/optimisation/opt_export_structure.cpp
+++ b/src/solver/optimisation/opt_export_structure.cpp
@@ -69,9 +69,9 @@ void OPT_ExportInterco(const Antares::Solver::IResultWriter::Ptr writer,
     for (int i(0); i < ProblemeHebdo->NombreDInterconnexions; ++i)
     {
         Flot.appendFormat("%d %d %d\n",
-                            i,
-                            ProblemeHebdo->PaysOrigineDeLInterconnexion[i],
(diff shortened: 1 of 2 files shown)


src/solver/optimisation/opt_export_structure.cpp
```diff
@@ -93,30 +93,18 @@ void OPT_Export_add_variable(std::vector<std::string>& varname,
...
-                             int firstVal)
+                             std::optional<int> secondVal)
 {
     if ((int)varname.size() > Var && varname[Var].empty())
     {
         std::stringstream buffer;
         buffer << Var << " ";
         buffer << Antares::Data::Enum::toString(structDict) << " ";
         buffer << firstVal;
+        if (secondVal)
```
guilpier-code: I like the **has_value()**... **value()** version

guilpier-code (changes requested):

src/solver/optimisation/opt_export_structure.cpp
flomnes: It's a bit more expressive, I agree

guilpier-code (approved):

PR #1044, opened 2022-12-29
Fix output overwrite when using zip output

### LINKED ISSUE #1002: Ensure the creation of a unique output directory
When using files, a prefix is added if the directory already exists.

It is not the case when a .zip file is used for result output.

## How to reproduce

Execute a short study twice with option `-z`. The 1st zip archive is overwritten.

flomnes:


payetvin (approved):
Security problem according to sonarcloud with tmp filename
It's located in tests so fine for me

PR #1059, opened 2023-01-04
Feature/cleanup yuni

JasonMarechal25:


JasonMarechal25:
Need to add definition of NOMINMAX globally (through cmake) to prevent windows.h to define min and max macro

JasonMarechal25:
Closing for now. To reopen when/if development restart

PR #1079, opened 2023-01-10
Restore warm start for OR-Tools+XPRESS

### LINKED ISSUE #1078: warm start is disabled with OR-Tools+XPRESS
# Hot-start disabled for all weeks but the 1st in v8.4.1

# Probable cause
`Probleme->solverSupportsWarmStart` is assigned `true` at the 1st week, and `false` for other weeks.

flomnes:


src/solver/utils/ortools_utils.cpp
```diff
@@ -116,6 +116,20 @@ static void tuneSolverSpecificOptions(MPSolver* solver)
     }
 }
 
+static bool solverSupportsWarmStart(const MPSolver* solver)
```
guilpier-code, line 119: **Warm** is better than **hot** ? 
I would have no problem with that, but as everybody in the project say **hot start** when she/he talks about that, it can be confusing for a second (is there a **hot start** and a **warm start** ? Are they different ? Ok, no, they are the same)

src/solver/utils/ortools_utils.cpp
```diff
@@ -116,6 +116,20 @@ static void tuneSolverSpecificOptions(MPSolver* solver)
...
+    if (!solver)
+        return false;
+
+    switch (solver->ProblemType())
+    {
+    case MPSolver::XPRESS_LINEAR_PROGRAMMING:
+        return true;
+    default:
+        return false;
+    }
```
guilpier-code, line 130: So, there are no **warm start** for any solver apart from **xpress** behind **ortools**. I guess it's expected, but why ?

payetvin (approved):

src/solver/utils/ortools_utils.cpp
flomnes: I agree that we mostly say "hot start". But it seems that we're wrong since Google returns many more results for "warm start".

src/solver/utils/ortools_utils.cpp
flomnes: Yes, this is expected. Methods `SetStartingLpBasisInt` and `GetFinalLpBasisInt` are implemented only for the LP XPRESS solver. For other solvers, a fatal log is emitted.

guilpier-code (approved):

PR #1081, opened 2023-01-11
Feature/simulteur as lib

JasonMarechal25:
The goal of this PR is to export simulator (mainly the solver) as a lib tu be used by xpansion


JasonMarechal25:
Current status:
- Lib package is created properly.
- Headers are exported properly
- Lib can be used by xpansion with find_package

Limitation:
- Only a couple of headers are exported
- When used as a lib headers pointing across "modules" are not resolved properly

What to do next:
- Rename "src" folder as "antares" (or any name we want for the lib)
- Update includes path in source file to use full path from antares/ when trying to find header across repository/modules

Ideas:
- Transform each "module" as a proper cmake package instead of a cmake file included in a top CmakeList.txt file

sylvlecl:
> Ideas:
> 
>     * Transform each "module" as a proper cmake package instead of a cmake file included in a top CmakeList.txt file

I upvote for this! :)

flomnes:
Too many merge conflicts

PR #1098, opened 2023-01-17
Pass hour instead of hourlyCsrProblem

flomnes:


payetvin (approved):

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
```diff
@@ -154,8 +152,8 @@ void OPT_InitialiserLesCoutsQuadratiques_CSR(PROBLEME_HEBDO* ProblemeHebdo,
     memset((char*)ProblemeAResoudre->CoutLineaire,
            0,
            ProblemeAResoudre->NombreDeVariables * sizeof(double));
-
-    setQuadraticCost(ProblemeHebdo, hourlyCsrProblem);
+    const auto hour = hourlyCsrProblem.hourInWeekTriggeredCsr;
```
guilpier-code: could not we have : 
```C++
void OPT_InitialiserLesCoutsQuadratiques_CSR(PROBLEME_HEBDO* ProblemeHebdo,
                                             int hour)
```
instead ?

guilpier-code (changes requested):

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
flomnes: Good idea ! Done

PR #1099, opened 2023-01-18
Preparatory refactor to use own instance of PROBLEME_A_RESOUDRE

flomnes:
* Rename `HOURLY_CSR_PROBLEM` -> `HourlyCSRProblem`
* Move file `csr_quadratic_problem.cpp` to directory `src/solver/optimisation/adequacy_patch_csr`
* Rename `OPT_ConstruireLaMatriceDesContraintesDuProblemeQuadratique_CSR` -> `buildConstraintMatrix`
* Use `std::vector<>` for temporary pointers instead of `malloc/free`
* To compute the objective function (cost), use an instance of `PROBLEME_POINT_INTERIEUR` (quadratic problem to be solved) instead of `PROBLEME_ANTARES_A_RESOUDRE` (temporary structure used to fill `PROBLEME_POINT_INTERIEUR`)
* Use one single instance of `HourlyCSRProblem` for all hours, add method `setHour`


src/solver/optimisation/adequacy_patch_csr/solve_problem.cpp
```diff
@@ -265,20 +249,20 @@ bool ADQ_PATCH_CSR(PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre,
                    uint weekNb,
                    int yearNb)
 {
-    double costPriorToCsr = calculateCsrCostFunctionValue(ProblemeAResoudre, hourlyCsrProblem);
     auto Probleme = buildInteriorPointProblem(ProblemeAResoudre);
+    double costPriorToCsr = calculateCsrCostFunctionValue(Probleme, hourlyCsrProblem);
```
flomnes: `Probleme` should be built from scratch, not from `ProblemeAResoudre` ?

src/solver/optimisation/adequacy_patch_csr/solve_problem.cpp
```diff
@@ -161,7 +150,7 @@ void storeOrDisregardInteriorPointResults(const PROBLEME_ANTARES_A_RESOUDRE* Pro
           << ". hour: " << weekNb * hoursInWeek + hourlyCsrProblem.hourInWeekTriggeredCsr + 1;
 }
 
-double calculateCsrCostFunctionValue(const PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre,
+double calculateCsrCostFunctionValue(const std::unique_ptr<PROBLEME_POINT_INTERIEUR>& Probleme,
```
sylvlecl: Design suggestion, from the c++ core guidelines:
we should use `const PROBLEME_POINT_INTERIEUR&` as a parameter here (or `PROBLEME_POINT_INTERIEUR&`, if modified). Using a std::unique_ptr should be reserved to transferring ownership, and there is no point in requiring a unique_ptr here instead of another kind of pointer/reference.

Reference is the right choice compared to plain pointer because here we expect a non-null value.

flomnes: [done](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1099/commits/088a498fa7e6286e2dc50bda2a5325de4e01317e)

src/solver/optimisation/adequacy_patch_csr/csr_quadratic_problem.h
```diff
@@ -27,21 +27,21 @@
...
 
 class CsrQuadraticProblem
 {
 public:
-    CsrQuadraticProblem(PROBLEME_HEBDO *prob, HOURLY_CSR_PROBLEM& hourly) :
-        ProblemeHebdo (prob),
-        hourlyCsrProblem (hourly)
-        {}
+    CsrQuadraticProblem(PROBLEME_HEBDO* prob, HourlyCSRProblem& hourly) :
+     ProblemeHebdo(prob), hourlyCsrProblem(hourly)
```
guilpier-code: I suggest : 
**ProblemeHebdo** --> **problemeHebdo_**
**hourlyCsrProblem** --> **hourlyCsrProblem_**

guilpier-code (changes requested):

src/solver/optimisation/adequacy_patch_csr/csr_quadratic_problem.h
flomnes: done

src/solver/optimisation/hourly_csr_problem.h
```diff
@@ -50,14 +44,18 @@ class HOURLY_CSR_PROBLEM
 
     int hourInWeekTriggeredCsr;
```
sylvlecl, line 45: Those fields should be private I think.

src/solver/optimisation/hourly_csr_problem.h
```diff
@@ -50,14 +44,18 @@ class HOURLY_CSR_PROBLEM
...
     double belowThisThresholdSetToZero;
-    PROBLEME_HEBDO* problemeHebdo;
-    HOURLY_CSR_PROBLEM(int hourInWeek, PROBLEME_HEBDO* pProblemeHebdo) :
-        hourInWeekTriggeredCsr(hourInWeek),
-        problemeHebdo(pProblemeHebdo)
+    PROBLEME_HEBDO* problemeHebdo_;
```
sylvlecl, line 47: I would remove the trailing "_" for consistency (or rename others)

flomnes: Conflicting reviews

src/solver/optimisation/hourly_csr_problem.h
flomnes: Yes, I'll add a TODO note

PR #1105, opened 2023-01-19
Use our own instance of `PROBLEME_ANTARES_A_RESOUDRE`

flomnes:
- Pass `ProblemeHebdo` by const pointer
- Pass HourlyCSRProblem's `PROBLEME_ANTARES_A_RESOUDRE` by pointer. Use a different instance than the member of `PROBLEME_HEBDO` since it has nothing in common (weekly adequacy problem _vs_ hourly CSR problem)
- Use anonymous namespace according to @sylvlecl suggestion (avoid exposing implementation details, better encapsulation)

### TODO
- [X] `HourlyCSRProblem::setVariableBounds()`
- [X] `HourlyCSRProblem::setProblemCost()`
- [x] `HourlyCSRProblem::buildProblemConstraintsRHS()`
- [x] `HourlyCSRProblem::buildProblemConstraintsLHS()`
- [x] Handle memory allocation/deallocation
- [x] Invesitate `OPT_LiberationProblemesSimplexe(problemeHebdo_);` : not sure we want to call this


src/solver/optimisation/hourly_csr_problem.h
```diff
@@ -45,6 +45,7 @@ class HourlyCSRProblem
     int hourInWeekTriggeredCsr;
     double belowThisThresholdSetToZero;
     PROBLEME_HEBDO* problemeHebdo_;
+    PROBLEME_ANTARES_A_RESOUDRE problemeAResoudre_;
```
guilpier-code, line 51: **Question** : 
We have a **problemeHebdo_** as a member of class **HourlyCSRProblem**. To me, it should not be (an weekly problem should not be part of a hourly problem). Is is a transitional stage of the code ? Is it intended to remain in the final state ?

src/solver/optimisation/adequacy_patch_csr/solve_problem.cpp
```diff
@@ -256,20 +256,20 @@ void handleInteriorPointError(const PROBLEME_POINT_INTERIEUR& Probleme,
...
 }
 
-bool ADQ_PATCH_CSR(PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre,
-                   HourlyCSRProblem& hourlyCsrProblem,
+bool ADQ_PATCH_CSR(HourlyCSRProblem& hourlyCsrProblem,
                    uint weekNb,
                    int yearNb)
 {
-    auto Probleme = buildInteriorPointProblem(ProblemeAResoudre);
+    auto problemeAResoudre = &hourlyCsrProblem.problemeAResoudre_;
```
guilpier-code: Why not : 
```C++
    auto problemeAResoudre = hourlyCsrProblem->problemeAResoudre_;
```
?

guilpier-code (changes requested):

src/solver/optimisation/hourly_csr_problem.h
flomnes: Yeah we need it for many things:
- Adequacy patch parameters
- Adequacy patch runtime data (e.g areas inside/outside/virtual)
- Variable/constraint correspondence arrays necessary for
  - Accessing (read & write) results from the adequacy simulation
  - Accessing the hourly binding constraint indices (input data)

It would be nice to remove it though.

src/solver/optimisation/adequacy_patch_csr/solve_problem.cpp
flomnes: Because the function after takes a pointer as an argument. It could take a non-const ref.

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
```diff
@@ -113,47 +116,50 @@ void setLinearCost(PROBLEME_HEBDO* ProblemeHebdo, int hour)
...
+ 
+void OPT_InitialiserLesCoutsQuadratiques_CSR(const PROBLEME_HEBDO* ProblemeHebdo,
+                                             PROBLEME_ANTARES_A_RESOUDRE& ProblemeAResoudre,
+                                             int hour)
 {
     logs.debug() << "[CSR] cost";
 
-    PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre = ProblemeHebdo->ProblemeAResoudre;
-    memset((char*)ProblemeAResoudre->CoutLineaire,
+    memset((char*)ProblemeAResoudre.CoutLineaire,
```
sylvlecl: I know it's not part of the PR, put this cast to `char*` seems suspect to me ?

src/solver/optimisation/adequacy_patch_csr/construct_problem_variables.cpp
```diff
@@ -65,14 +67,14 @@ void constructVariableENS(PROBLEME_HEBDO* ProblemeHebdo, HourlyCSRProblem& hourl
...
+void constructVariableSpilledEnergy(const PROBLEME_HEBDO* ProblemeHebdo,
+                                    PROBLEME_ANTARES_A_RESOUDRE& ProblemeAResoudre,
                                     HourlyCSRProblem& hourlyCsrProblem)
 {
     int hour = hourlyCsrProblem.hourInWeekTriggeredCsr;
-    PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre = ProblemeHebdo->ProblemeAResoudre;
-    CORRESPONDANCES_DES_VARIABLES* CorrespondanceVarNativesVarOptim
-        = ProblemeHebdo->CorrespondanceVarNativesVarOptim[hour];
-    int& NumberOfVariables = ProblemeAResoudre->NombreDeVariables;
+    CORRESPONDANCES_DES_VARIABLES* CorrespondanceVarNativesVarOptim;
```
sylvlecl: It's considered better practice to have declaration and assignment on the same line

sylvlecl (approved):

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
flomnes: It does

src/solver/optimisation/adequacy_patch_csr/construct_problem_variables.cpp
flomnes: done

PR #1107, opened 2023-01-19
Fix infinite loop, possible other bug

flomnes:


src/solver/optimisation/csr_quadratic_problem.cpp
```diff
@@ -149,20 +152,22 @@ void CsrQuadraticProblem::setNodeBalanceConstraints(double* Pi, int* Colonne)
...
-                NombreDeTermes++;
-                logs.debug()
-                    << "E-Interco number: [" << std::to_string(Interco) << "] between: ["
-                    << ProblemeHebdo->NomsDesPays[Area] << "]-["
-                    << ProblemeHebdo
-                    ->NomsDesPays[ProblemeHebdo->PaysOrigineDeLInterconnexion[Interco]]
-                    << "]";
+                Var = CorrespondanceVarNativesVarOptim
+                        ->NumeroDeVariableDeLInterconnexion[Interco]; // flow (2 -> A)
+                if (Var >= 0)
```
flomnes, line 159: Before that it's possible that `Var` is from a previous iteration. But we probably don't want that.

src/solver/optimisation/csr_quadratic_problem.cpp
```diff
@@ -124,8 +124,11 @@ void CsrQuadraticProblem::setNodeBalanceConstraints(double* Pi, int* Colonne)
         while (Interco >= 0)
         {
             if (ProblemeHebdo->adequacyPatchRuntimeData.extremityAreaMode[Interco]
-                    != Data::AdequacyPatch::physicalAreaInsideAdqPatch)
+                != Data::AdequacyPatch::physicalAreaInsideAdqPatch)
+            {
+                Interco = ProblemeHebdo->IndexSuivantIntercoOrigine[Interco];
                 continue;
+            }
```
flomnes, line 131: There was an infinite loop here because `Interco` wasn't updated

payetvin (approved):

PR #1109, opened 2023-01-23
Refactor optimization post process

guilpier-code:
The main idea is to customize the post processes to be run, in regard to the weekly optimization to be run : the adequacy patch weekly optimization is not followed by the same post processes than the default weekly optimization. 

To do that : 
- **post process commands** : each post process used to be performed by a free function. This call was moved in the **execute()** methods of a command design pattern. These commands are meant to be added to a vector of post-process commands.
- **Post-process list (vector)** : depending of the kind of weekly optimization (adequacy path or default), we create a vector of post-process. This list is not supposed to be the same in either cases.
- **Running the post processes** : running the post processes is done in one line inside a simulation year.


src/solver/optimisation/post_process_commands.cpp
```diff
@@ -0,0 +1,28 @@
...
+        AreaList& areas) 
+    : basePostProcessCommand(problemeHebdo), 
+      thread_number_(thread_number), 
+      area_list_(areas)
+{
+}
+
+void DispatchableMarginPostProcessCmd::acquireOptRuntimeData(const struct optRuntimeData& opt_runtime_data)
+{
+    hourInYear_ = opt_runtime_data.year;
```
flomnes: Is it a year or an hour ?

guilpier-code: Exactly the problem of regression that I had and fixed !
Good eyes !
So this is corrected now

src/solver/simulation/base_post_process.h
```diff
@@ -0,0 +1,69 @@
...
+    optRuntimeData(unsigned int y, unsigned int w, unsigned int h) :
+        year(y), week(w), hourInTheYear(h)
+    {}
+
+    unsigned int year = 0;
+    unsigned int week = 0;
+    unsigned int hourInTheYear = 0;
+};
+
+class basePostProcessCommand
```
sylvlecl, line 26: For namings I would suggest to:
 - use consistently `UpperCamelCase` for classes/interfaces
 - remove the words `base` or `interface` from the root of hierarchies: it only hurts readability with longer names without real added value in the context of their use (we don't need to know it's an interface when we use it)
 - In this particular case, I would suggest to remove "command" word from the name, which is most often used for command line: maybe just `PostProcessor` ?

src/solver/cmake/simulation.cmake
```diff
@@ -34,13 +34,14 @@ set(SRC_SIMULATION
...
 		simulation/common-hydro-remix.cpp
 		simulation/common-hydro-levels.cpp
-		simulation/common-dispatchable-margin.cpp
 		simulation/adequacy-draft.h
 		simulation/adequacy-draft.cpp
 		simulation/adequacy.h
 		simulation/adequacy.cpp
 		simulation/economy.h
 		simulation/economy.cpp
+		simulation/base_post_process.h
```
guilpier-code, line 43: Rename **base_post_process.h/.cpp** into **post_process.h/.cpp** (according to remark of @sylvlecl )

src/solver/optimisation/post_process_commands.h
```diff
@@ -0,0 +1,23 @@
...
+{
+
+class DispatchableMarginPostProcessCmd : public basePostProcessCommand
+{
+public:
+    DispatchableMarginPostProcessCmd(PROBLEME_HEBDO* problemeHebdo,
+                                     unsigned int thread_number,
+                                     AreaList& areas);
+    void acquireOptRuntimeData(const struct optRuntimeData& opt_runtime_data) override;
+    void run() override;
```
guilpier-code: Rename **run()** into **execute()**, to stick with the command design pattern, so that it can be more recognizable

src/solver/optimisation/optim_post_process_list.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include "../simulation/base_post_process.h"
+
+
+namespace Antares::Solver::Simulation
+{
+
+class OptPostProcessList : public interfacePostProcessList
```
guilpier-code, line 9: Rename **OptPostProcessList** into **DefaultOptPostProcessList**.
Rename file **optim_post_process_list.h** into **default_opt_post_process_list.h**

src/solver/optimisation/adq_patch_post_process_list.h
```diff
@@ -0,0 +1,23 @@
...
+{
+public:
+    AdqPatchPostProcessList(PROBLEME_HEBDO* problemeHebdo,
+                            uint thread_number,
+                            AreaList& areas,
+                            SheddingPolicy sheddingPolicy,
+                            SimplexOptimization splxOptimization,
+                            Calendar& calendar);
+
+    ~AdqPatchPostProcessList() = default;
```
flomnes: Make this function `virtual`

src/solver/simulation/base_post_process.cpp
```diff
@@ -0,0 +1,48 @@
...
+        return std::make_unique<AdqPatchPostProcessList>(
+            problemeHebdo, thread_number, areas, sheddingPolicy, splxOptimization, calendar);
+    else
+        return std::make_unique<OptPostProcessList>(
+            problemeHebdo, thread_number, areas, sheddingPolicy, splxOptimization, calendar);
+}
+
+void interfacePostProcessList::runAll(const struct optRuntimeData& opt_runtime_data)
+{
+    for (auto it = post_process_list.begin(); it != post_process_list.end(); it++)
```
flomnes:
```suggestion
    for (auto command : post_process_list)
```

src/solver/simulation/base_post_process.cpp
```diff
@@ -0,0 +1,48 @@
...
+    {
+        (*it)->acquireOptRuntimeData(opt_runtime_data);
+        (*it)->run();
```
flomnes:
```suggestion
        cmd->run(context);
```

flomnes (changes requested):
IMO there is no need for function `acquireOptRuntimeData`. Just call `run` with whatever context is necessary.

src/solver/optimisation/optim_post_process_list.cpp
```diff
@@ -0,0 +1,44 @@
...
+        thread_number));
+    post_process_list.push_back(std::make_unique<HydroLevelsUpdatePostProcessCmd>(
+        problemeHebdo_,
+        areas,
+        true,
+        false));
+    post_process_list.push_back(std::make_unique<InterpolateWaterValuePostProcessCmd>(
+        problemeHebdo_,
+        areas,
+        calendar));
```
flomnes, line 39: I know you don't like if..else much, but IMO it would be clearer to have
```cpp
post_process_list.push_back(...);
if (adqPatch)
    post_process_list.push_back(std::make_unique<DTGmarginForAdqPatchPostProcessCmd>(...);
post_process_list.push_back(...);
```
It would avoid code duplication.

flomnes: You could hide this in a factory function
```cpp
postProcessListFactory(bool adqPatch) {
// ...
}
```

flomnes (changes requested):

src/solver/simulation/common-eco-adq.h
```diff
@@ -88,16 +88,16 @@ void ComputeFlowQuad(Data::Study& study,
...
 */
-void RemixHydroForAllAreas(const Data::Study& study,
+void RemixHydroForAllAreas(const Data::AreaList& areas,
                            PROBLEME_HEBDO& problem,
+                           Data::SheddingPolicy sheddingPolicy,
+                           Data::SimplexOptimization splxOptimization,
                            uint numSpace,
-                           uint hourInYear,
-                           uint nbHours);
+                           uint hourInYear);
```
flomnes, line 100: This is context info

flomnes:
undefined reference to `Antares::Solver::Simulation::RemixHydroForAllAreas(Antares::Data::AreaList const&, PROBLEME_HEBDO&, Antares::Data::SheddingPolicy, Antares::Data::SimplexOptimization, unsigned int, unsigned int)`

guilpier-code:
> undefined reference to `Antares::Solver::Simulation::RemixHydroForAllAreas(Antares::Data::AreaList const&, PROBLEME_HEBDO&, Antares::Data::SheddingPolicy, Antares::Data::SimplexOptimization, unsigned int, unsigned int)`

Yes, only on Ubuntu, **antares-solver** is built, but **tests-simple-study** fails to compile

src/solver/optimisation/post_process_commands.h
guilpier-code: Done

src/solver/simulation/base_post_process.cpp
guilpier-code: Done

src/solver/simulation/base_post_process.cpp
guilpier-code: You're right, better that way.
Done

src/solver/simulation/common-eco-adq.h
guilpier-code: could you give some more explanations about what you mean ?

src/solver/optimisation/adq_patch_post_process_list.h
guilpier-code: Done : I made the destructor **virtual**

src/solver/optimisation/optim_post_process_list.cpp
guilpier-code: This should be discussed more deeply. 
Before taking a decision, let's look at **adequacy.cpp**, which has not been touched yet.

src/solver/simulation/common-eco-adq.h
flomnes: It should be picked up from `optRuntimeData`

guilpier-code: This free function **RemixHydroForAllAreas(...)** is called from a post process command, which retrieves  **hourInYear** from **optRuntimeData**.
We could move the body of this function to the **run(...)** method of this post process command. 
But it would have some consequences, as  **RemixHydroForAllAreas(...)** is also called in **adequacy.cpp**. 
So I preferred postponing this refactoring.

PR #1111, opened 2023-01-23
Remove `OPT_*_CSR` functions

flomnes:


src/solver/optimisation/adequacy_patch_csr/adq_patch_curtailment_sharing.cpp
```diff
@@ -197,8 +197,11 @@ void HourlyCSRProblem::allocateProblem()
 
 void HourlyCSRProblem::buildProblemVariables()
 {
-    OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeQuadratique_CSR(
-      problemeHebdo_, problemeAResoudre_, *this);
+    logs.debug() << "[CSR] variable list:";
```
guilpier-code, line 200: At some point, we should rename this file into **curtailment_sharing.cpp**, as it would be sufficient, and also because it is contained in directory **adequacy_patch_csr**

payetvin (approved):

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
```diff
@@ -69,86 +61,82 @@ void setRHSnodeBalanceValue(const PROBLEME_HEBDO* ProblemeHebdo,
...
+              = problemeHebdo_->MatriceDesContraintesCouplantes[CntCouplante];
 
-            int Cnt = bingdingConstraintNumber[CntCouplante];
+            int Cnt = numberOfConstraintCsrHourlyBinding[CntCouplante];
 
             // 1. The original RHS of bingding constraint
-            SecondMembre[Cnt]
-                = MatriceDesContraintesCouplantes->SecondMembreDeLaContrainteCouplante[hour];
+            SecondMembre[Cnt] = MatriceDesContraintesCouplantes
+                                  ->SecondMembreDeLaContrainteCouplante[hourInWeekTriggeredCsr];
```
guilpier-code: About **hourInWeekTriggeredCsr** : because here we obviously are in the context of CSR, we could shorten this name into  something like **triggeredHour**. It would remain expressive.

guilpier-code (changes requested):

src/solver/optimisation/adequacy_patch_csr/adq_patch_curtailment_sharing.cpp
flomnes: Right. I suggest even something like

```
adequacy_patch/
├── csr
└── lmr
```

PR #1116, opened 2023-01-25
Move tests outside of sources

JasonMarechal25:
- [ ] Move "tests" outside of src. Sources and tests directory are generally separated
- [ ] Have "src" be a proper Cmake modul with it's cmakelist
- [ ] Move down code from top CMakeList that belong to src/CMakeList


JasonMarechal25:
May be difficult at the moment. Include chain while building tests doesn't find include properly like yuni

PR #1120, opened 2023-01-26
Export yuni installed file at root level

JasonMarechal25:
Currently yuni is installed with the structure : yuni/1.1/yuni/<content>
Since include path are of the form <yuni/<contentent> when simulator is used as a library the path are not resolved properly.

With this PR it is possible to resolve include path to yuni properly


guilpier-code (approved):

flomnes:
@JasonMarechal25 What's the new path for Yuni ?

If this path changes, why don't we change in antares also ? I would expect to see

```diff
- #include <old/path/to/yuni/header.h>
+ #include <new/path/to/yuni/header.h>
```

JasonMarechal25:
@flomnes 
Currently when using the "install" target, yuni is installed in: /path/to/install/include/yuni/1.1/yuni
It's changed to /path/to/install/include/yuni/

It doesn't change anything for Simulator that build against yuni internally. It only affect components that would look for yuni headers in an installation prefix path.

flomnes (approved):

PR #1141, opened 2023-01-31
Use dedicated structure to test if CSR was triggered for area, hour

### LINKED ISSUE #1129: Do not use `ResultatsHoraires[Area]->ValeursHorairesDtgMrgCsr[hour]` to take decisions
## Current state
```cpp
    for (int Area = 0; Area < problemeHebdo_->NombreDePays; Area++)
    {
        if (problemeHebdo_->adequacyPatchRuntimeData.areaMode[Area]
            == Antares::Data::AdequacyPatch::physicalAreaInsideAdqPatch)
        {
            // set DTG MRG CSR in all areas inside adq-path for all CSR triggered hours to -1.0
            problemeHebdo_->ResultatsHoraires[Area]->ValeursHorairesDtgMrgCsr[hour] = -1.0;
            // ...
        }
    }
```

```cpp
            if (dtgMrgCsr == -1.0) // area is inside adq-patch and it is CSR triggered hour
            {
                dtgMrgCsr = std::max(0.0, dtgMrg - ens);
                ens = std::max(0.0, ens - dtgMrg);
                // set MRG PRICE to value of unsupplied energy cost, if LOLD=1.0 (ENS>0.5)
                if (ens > 0.5)
                    mrgCost = -area_list_[Area]->thermal.unsuppliedEnergyCost;
            }
            else
                dtgMrgCsr = dtgMrg;
```

## Suggestion
Do not use `ResultatsHoraires[Area]->ValeursHorairesDtgMrgCsr[hour]` to take decisions. Only use it to store results.

flomnes:
**Depends on #1146 to fix segfault related to `memset`. Otherwise `memset ` sets two `std::shared_ptr` to `nullptr`**

* Use dedicated structure to test if CSR was triggered for area, hour
* Dedicated .h/.cpp for Adequacy Patch runtime data
* Use `std::shared_ptr` for adq-patch runtime data & parameters
* Run tests on adq-patch CSR on push (not release)


src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -324,6 +337,9 @@ class AdequacyPatchRuntimeData
             extremityAreaMode.push_back(link.with->adequacyPatchMode);
         }
     }
+
+private:
+    std::set<std::pair<int, int>> csrTriggered_;
```
flomnes: Maybe there's a better structure for insert/find execution times, but I can't think of any.

flomnes: See [this discussion](https://gist.github.com/flomnes/5614997a44660ab7b8836ef3cd6497be)

src/solver/simulation/adequacy_patch_runtime_data.h
```diff
@@ -0,0 +1,48 @@
...
+#include <antares/study/fwd.h>
+#include <antares/study/study.h>
+#include <vector>
+#include <set>
+
+class AdequacyPatchRuntimeData
+{
+private:
+    using adqPatchParamsMode = Antares::Data::AdequacyPatch::AdequacyPatchMode;
+    std::vector<std::set<int>> csrTriggered;
```
guilpier-code: **csrTriggered** unclear to me. 
We could rename it **csrTriggeredHoursPerArea_**

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -597,8 +577,8 @@ struct PROBLEME_HEBDO
     OptimizationStatistics optimizationStatistics[2];
 
     /* Adequacy Patch */
-    std::unique_ptr<AdequacyPatchParameters> adqPatchParams = nullptr;
-    AdequacyPatchRuntimeData adequacyPatchRuntimeData;
+    std::shared_ptr<AdequacyPatchParameters> adqPatchParams = nullptr;
```
guilpier-code: I guess we don't need to initialize smart pointers to **nullptr**, as it is their default value anyway

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -58,7 +59,7 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
     if (parameters.adqPatch.enabled)
     {
         problem.adqPatchParams
-          = std::unique_ptr<AdequacyPatchParameters>(new AdequacyPatchParameters());
+          = std::make_shared<AdequacyPatchParameters>();
```
guilpier-code: I guess we should have : 
```C++
if (parameters.adqPatch.enabled)
{
     problem.adqPatchParams = std::make_shared<AdequacyPatchParameters>(parameters.adqPatch);
}
```
and let the constructor do its job.

Besides, it seems like `problem.adqPatchParams` and `parameters.adqPatch` contain very much the same things. 
Why are there 2 different types then ? 
- one type for an instance in **Parameters** (type : **AdequacyPatch**)
- one in **PROBLEME_HEBDO** (type : **AdequacyPatchParameters**)

There is some kind of code duplication here.
If this is true, we could simply store a reference to `parameters.adqPatch` inside `PROBLEME_HEBDO`

src/solver/simulation/adequacy_patch_runtime_data.cpp
```diff
@@ -0,0 +1,27 @@
...
+{
+    return csrTriggered[area].count(hour) > 0;
+}
+
+void AdequacyPatchRuntimeData::addCSRTriggeredAtAreaHour(int area, int hour)
+{
+    csrTriggered[area].insert(hour);
+}
+
+void AdequacyPatchRuntimeData::initialize(Antares::Data::Study& study)
```
guilpier-code: It seems we call this structure **RuntimeData** because we use supposed study runtime data to initialize it.
The problem is that this study runtime data does not seem to require some change at runtime.
In comparison, `problemeHebdo_->adqPatchParams->AdequacyFirstStep ` is a real runtime data (its value actually changes during runtime), but it  belongs to a structure that is not supposed to be used for a runtime usage (according to its name). 

Ideally (it should be the purpose of a further github ticket), we should : 
- extract **interconnectionsCount** and **areaLink** from **study->runtime** if they turn out not to be actual runtime pieces of data.
- remove from **AdequacyPatchRuntimeData** members that are not runtime data : **areaMode**, **originAreaMode** and **extremityAreaMode** to put them in the other adq patch structure, like `problemeHebdo_->adqPatchParams` for instance 
- add real runtime data like **AdequacyFirstStep** to **AdequacyPatchRuntimeData**

src/solver/simulation/sim_calcul_economique.cpp
guilpier-code: _**But eventually**_ : 

Honestly, **PROBLEME_HEBDO** should not contain an **adqPatchParams** data member.
Configuration of adq patch should not belong to **PROBLEME_HEBDO**. We access adequacy patch **fixed** parameters (not even runtime adq patch properties) through **PROBLEME_HEBDO** to solve.... adq patch problems !
How strange !
Wouldn't it be more simple to build these adq patch problems with the  **parameters.adqPatch** directly ?
For example, this is the way it goes in the curtailment sharing problem : 
instead of building it with the adq patch study parameters, we get the values of these parameters through the  **PROBLEME_HEBDO**

src/solver/optimisation/adequacy_patch_csr/adq_patch_curtailment_sharing.cpp
```diff
@@ -29,6 +29,7 @@
 #include "../opt_fonctions.h"
 #include "csr_quadratic_problem.h"
 #include "count_constraints_variables.h"
+#include "../simulation/adequacy_patch_runtime_data.h"
```
guilpier-code, line 32: Here the problem of curtailment sharing needs the **problemeHebdo** to access : 
- the optimization results
- the adq patch parameters

I don't discuss the first point, which seems natural.
However, the second is more questionable.
Why would the CSR need to access adq patch parameters through **problemeHebdo** ?
We should build the CSR problem directly with adq patch from study, whether these parameters are adq patch runtime ones or not.
Besides, as said in review of **adequacy_patch_runtime_data.cpp**, the fact the runtime parameters are real runtime parameters is also questionable (see this review).

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
```diff
@@ -63,7 +64,7 @@ void HourlyCSRProblem::setRHSnodeBalanceValue()
 
     for (int Area = 0; Area < problemeHebdo_->NombreDePays; Area++)
```
guilpier-code, line 65: We should not loop over areas with the help of **problemeHebdo_**, but with the help of **study->areas**

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
```diff
@@ -63,7 +64,7 @@ void HourlyCSRProblem::setRHSnodeBalanceValue()
...
     {
-        if (problemeHebdo_->adequacyPatchRuntimeData.areaMode[Area]
+        if (problemeHebdo_->adequacyPatchRuntimeData->areaMode[Area]
```
guilpier-code, line 67: As well as we should not access adq patch parameters from the  **problemeHebdo_**, but from the **study->parameters->adqpatch**

src/solver/optimisation/adequacy_patch_local_matching/adq_patch_local_matching.cpp
```diff
@@ -98,9 +101,9 @@ static NtcSetToZeroStatus_AdqPatchStep1 getNTCtoZeroStatus(PROBLEME_HEBDO* probl
                                                            int Interco)
 {
```
guilpier-code, line 102: About **problemeHebdo->adqPatchParams->AdequacyFirstStep**, we should not store this runtime data in the **problemeHebdo**, but rather where it belongs to : the local matching problem.
Besides, we should not name it **AdequacyFirstStep** but **FirstStep** witch is enough, due to the context (adqPatchParams or whatever it will be after the change I advice)

guilpier-code (changes requested):

guilpier-code:
Some of the remarks made here go out of the scope of this PR.
To fix these remarks, an issue was created (see #1152)

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
flomnes: I agree. The runtime data should be provided as an argument. parameters->adqpatch as well (there are 2 structures)

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
flomnes: Yes, but currently we have 2 layers
- One C++ layer defined in libs/antares/study
- One C layer defined in solver/ SIM

The C++ structures are translated into SIM, so that OPT/SIM doesn't use C++ structures. Do we want to change this logic ?

src/solver/simulation/adequacy_patch_runtime_data.cpp
flomnes: It's runtime data because it's initialized from runtime data (study->runtime).

Your remark was true before this patch (AdequacyPatchRuntimeData doesn't hold runtime data), but it becomes obsolete after new data member `csrTriggered` is added.

src/solver/optimisation/adequacy_patch_csr/adq_patch_curtailment_sharing.cpp
flomnes: Yes, see #1152

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
guilpier-code: When I made this remark, I was worried about the dependency that the **CSR problem** has to the **weekly problem**, especially to the **adq patch** part of the **weekly problem**.
I still am.
After your last answer, I see that the weekly problem also contains some information about areas, links, number of clusters, ... as though this pieces of information already exist in the study. 
So it looks like some code duplication. 
But I don't think now is the time to address this (supposed) code duplication in the weekly problem.
That said, we could avoid the **weekly problem** dependency in the **CSR problem** by directly use **study->areas**

flomnes: OK, agreed. Let's not use SIM in OPT_AdequacyPatch. See #1162

src/solver/optimisation/adequacy_patch_local_matching/adq_patch_local_matching.cpp
flomnes: I agree

guilpier-code (approved):

PR #1149, opened 2023-02-02
Remove unused MPS sources & global variable

flomnes:


src/solver/CMakeLists.txt
```diff
@@ -59,9 +59,6 @@ add_executable(antares-solver
 	misc/system-memory.h
 	misc/system-memory.cpp
 
-	mps/mps_define.h
```
guilpier-code, line 62: ok, these files were moved to **antares-deps** at some point.

guilpier-code (approved):

src/solver/CMakeLists.txt
flomnes: no, they were never used

Issue #1152, opened 2023-02-03
adq patch : move adq patch params from problem hebdo


guilpier-code:
This issue is about taking care of the remarks that would not taken care of in the PR #1141.
~~Please refer to this PR to solve the current issue.~~

- extract interconnectionsCount and areaLink from study->runtime if they turn out not to be actual runtime pieces of data.
- remove from AdequacyPatchRuntimeData members that are not runtime data : areaMode, originAreaMode and extremityAreaMode to put them in the other adq patch structure, like problemeHebdo_->adqPatchParams for instance
- add real runtime data like AdequacyFirstStep to AdequacyPatchRuntimeData

Here the problem of curtailment sharing needs the problemeHebdo to access :

- the optimization results (OK)
- the adq patch parameters (not OK)

Why would the CSR need to access adq patch parameters through problemeHebdo ?
We should build the CSR problem directly with adq patch from study, whether these parameters are adq patch runtime ones or not.
Besides, as said in review of adequacy_patch_runtime_data.cpp, the fact the runtime parameters are real runtime parameters is also questionable.

PR #1154, opened 2023-02-06
adq patch : refactor parameters management

### LINKED ISSUE #1131: Parameters : split parameters.cpp in several pieces
**parameters.h/.cpp** and **fwd.h** are growing bigger and bigger.
We should split them into finer grained files which could be called **some-subject-params.h/.cpp**.
Note that definitions of **enum**  could be moved to **some-subject-params.h** : see if this implies a lot of work or not

### Themes
- [ ] Adequacy patch, see #1154
- [ ] Random seeds (`seed[seedMax]`, `resetSeeds`)
- [ ] Hydro (heuristic policy, pricing mode, initial reservoir level, hydro debug)
- [ ] Timeseries (number of timeseries, ~~refresh interval~~)
- [ ] Simulation mode (economy/adequacy/expansion
- [ ] Playlist / `yearsFilter` / section "playlist"
- [ ] Years weights (search for "weight", 4-5 results)
- [ ] TS generation (intra-modal / inter-modal, refresh intervals, `storeTimeseriesNumbers`, `timeSeriesAccuracyOnCorrelation`)
- [ ] `variablesPrintInfo` / section "variables selection"
- [ ] OR-Tools (ON/OFF, what solver to use)
- [ ] Scenario builder (`activeRulesScenario`, `useCustomScenario`)
- [ ] Calendar (`simulationDays`, `dayOfThe1stJanuary`, `firstWeekday`, `firstMonthInYear`, `leapYear`
- [ ] Result export (`yearByYear`, `synthesis`, `noOutput`, `resultFormat`)

### When using parameters
Instead of passing `parameters`, pass only the required part, e.g `parameters.adqPatchParams`, `parameters.hydroParams`, etc.

guilpier-code:
Extracting adequacy patch parameters management logic form class **Paramaters** to class **AdqPatchParams**.
Code was moved from **parameters.h/.cpp** and **fwd.h** to files **adq-patch-params.h/.cpp**

This PR is a fix for issue #1131


src/libs/antares/study/parameters/adq-patch-params.h
```diff
@@ -0,0 +1,107 @@
...
+    //! CSR Variables relaxation threshold
+    int thresholdVarBoundsRelaxation;
+    //! Include hurdle cost in CSR cost function
+    bool includeHurdleCost;
+    //! Check CSR cost function prior & after CSR optimization
+    bool checkCsrCostFunction;
+    
+    
+    void reset();
+    void resetThresholds();
```
guilpier-code, line 104: About **resetThresholds()** : 
It is called inside **CurtailmentSharing::reset()** and nowhere else.
So we should make struct **CurtailmentSharing** a class and make **resetThresholds()** a private method.

flomnes: Agreed

guilpier-code: Done

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -53,26 +53,26 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
...
+          = parameters.adqPatchParams.curtailmentSharing.checkCsrCostFunction;
         problem.adqPatchParams->ThresholdRunCurtailmentSharingRule
-          = parameters.adqPatch.curtailmentSharing.thresholdRun;
+          = parameters.adqPatchParams.curtailmentSharing.thresholdRun;
         problem.adqPatchParams->ThresholdDisplayLocalMatchingRuleViolations
-          = parameters.adqPatch.curtailmentSharing.thresholdDisplayViolations;
-        double temp = pow(10, -parameters.adqPatch.curtailmentSharing.thresholdVarBoundsRelaxation);
+          = parameters.adqPatchParams.curtailmentSharing.thresholdDisplayViolations;
+        double temp = pow(10, -parameters.adqPatchParams.curtailmentSharing.thresholdVarBoundsRelaxation);
         problem.adqPatchParams->ThresholdCSRVarBoundsRelaxation = temp < 0.1 ? temp : 0.1;
```
flomnes: The next step is to remove `problem.adqPatchParams`, and pass `parameters.adqPatchParams` where needed.

guilpier-code: I agree that it should be done.
This task is the subject of #1152, which depends on PR #1141.
So we could first merge end close PR #1141.

src/libs/antares/study/parameters/adq-patch-params.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+bool AdqPatchParams::updateFromKeyValue(const String& key, const String& value)
+{
+    if (key == "include-adq-patch")
+        return value.to<bool>(enabled);
+    if (key == "set-to-null-ntc-from-physical-out-to-physical-in-for-first-step")
+        return value.to<bool>(localMatching.setToZeroOutsideInsideLinks);
+    if (key == "set-to-null-ntc-between-physical-out-for-first-step")
+        return value.to<bool>(localMatching.setToZeroOutsideOutsideLinks);
+    // Price taking order
```
flomnes: You could go one step further and add `CurtailmentSharing::updateFromKeyValue` & `LocalMatching::updateFromKeyValue`. Then this method looks like
```cpp
bool AdqPatchParams::updateFromKeyValue(const String& key, const String& value)
{
   if (key == "include-adq-patch")
        return value.to<bool>(enabled);
   return localMatching.updateFromKeyValue(key, value) || curtailmentSharing.updateFromKeyValue(key, value);
}

File: src/libs/antares/study/parameters/adq-patch-params.cpp [hunk 4]
```diff
@@ -0,0 +1,145 @@
...
+}
+
+void AdqPatchParams::saveToINI(IniFile& ini) const
+{
+    auto* section = ini.addSection("adequacy patch");
+    section->add("include-adq-patch", enabled);
+    section->add("set-to-null-ntc-from-physical-out-to-physical-in-for-first-step",
+        localMatching.setToZeroOutsideInsideLinks);
+    section->add("set-to-null-ntc-between-physical-out-for-first-step",
+        localMatching.setToZeroOutsideOutsideLinks);
```
flomnes: Same remark, add `LocalMatching::addProperties(IniFile::Section& section)` and `CurtailmentSharing::addProperties(IniFile::Section& section)`, so that this method becomes
```cpp
auto* section = ini.addSection("adequacy patch");
section->add("include-adq-patch", enabled);
localMatching.addProperties(section);
curtailmentSharing.addProperties(section);
```

flomnes (commented):
I don't understand the changes related to `Parameters::horizon`. Care to explain ?

flomnes (changes requested):
`parameters.adqPatch` looks better than `parameters.adqPatchParameters` in my opinion.

src/libs/antares/study/parameters/adq-patch-params.cpp
guilpier-code: I do agree

src/libs/antares/study/parameters/adq-patch-params.cpp
guilpier-code: I agree as well

src/libs/antares/study/parameters/adq-patch-params.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+
+bool AdqPatchParams::updateFromKeyValue(const String& key, const String& value)
+{
+    if (key == "include-adq-patch")
+        return value.to<bool>(enabled);
+
+    localMatching.updateFromKeyValue(key, value);
```
flomnes: We want to make sure that one single property was updated. For me the proper behavior is to return the XOR of the 2 `updateFromKey` operations. It's not exact but it does the job.

flomnes (changes requested):

src/libs/antares/study/parameters/adq-patch-params.cpp
flomnes: Curiously enough, XOR for bools can be expressed as `a != b`.

src/libs/antares/study/parameters/adq-patch-params.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+
+bool AdqPatchParams::updateFromKeyValue(const String& key, const String& value)
+{
+    if (key == "include-adq-patch")
+        return value.to<bool>(enabled);
+
+    return curtailmentSharing.updateFromKeyValue(key, value) || localMatching.updateFromKeyValue(key, value);
```
flomnes:
```suggestion
    return curtailmentSharing.updateFromKeyValue(key, value) != localMatching.updateFromKeyValue(key, value); // XOR
```

flomnes (changes requested):

src/libs/antares/study/parameters/adq-patch-params.cpp
guilpier-code: I guess this was corrected, isn't it ?

src/libs/antares/study/parameters/adq-patch-params.cpp
guilpier-code: Sorry, I don't see the benefit of **XOR** operator compared to **| |** operator, is there any ?
It is less readable to me

flomnes: - If we leave operator `||`, it means that we accept the key to be added to both `localMatching` and `curtailmentSharing`.
- If we instead use `!=`, it means that the key must be added only to one of these, exclusively

In my opinion, the 2nd option is closer to what we want.

guilpier-code: This is subtle !
I pushed your suggestion.

flomnes (changes requested):
Can you add unit tests for the loading of the adq-patch properties ? Similar to **src/tests/src/libs/antares/study/area/test-save-area-optimization-ini.cpp**.

IMO there is no need to test the save (used only in the GUI). But the load is important.

guilpier-code:
> Can you add unit tests for the loading of the adq-patch properties ? Similar to **src/tests/src/libs/antares/study/area/test-save-area-optimization-ini.cpp**.
> 
> IMO there is no need to test the save (used only in the GUI). But the load is important.

Can we create a ticket for this ?

flomnes:
Sure ! #1185

flomnes (approved):

PR #1155, opened 2023-02-08
Add coefficient for hurdle cost in the CSR problem

flomnes:
If hurdle costs are enabled, use `1/max(unsupplied cost origin area, unsupplied cost destination area)` as a multiplicative coefficient on them to avoid side-effects.

### DIFF
--- a/src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
+++ b/src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
@@ -113,7 +113,8 @@ void HourlyCSRProblem::setLinearCost()
         {
             continue;
         }
-        const double coeff = problemeHebdo_->adequacyPatchRuntimeData.linkCostCoefficient[Interco];
+        const double coeff
+          = problemeHebdo_->adequacyPatchRuntimeData.hurdleCostCoefficients[Interco];
         TransportCost = problemeHebdo_->CoutDeTransport[Interco];
         // flow
         var = CorrespondanceVarNativesVarOptim->NumeroDeVariableDeLInterconnexion[Interco];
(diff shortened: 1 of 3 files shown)


src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
```diff
@@ -113,7 +113,8 @@ void HourlyCSRProblem::setLinearCost()
         {
```
guilpier-code, line 113: The file **hourly_csr_problem.h** is not referenced in the any cmake file.
I just run a symbol search over the solution and realized that **hourly_csr_problem.h** was not searched in.

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -301,27 +301,47 @@ typedef struct
...
 } ENERGIES_ET_PUISSANCES_HYDRAULIQUES;
 
+// TODO[FOM] move to new .h/.cpp
 class AdequacyPatchRuntimeData
 {
 private:
+    // TODO[FOM] out of class
+    static constexpr double thresholdForCostCoefficient = 1.e-9;
+    // TODO[FOM] free function
+    static double computeHurdleCostCoefficient(double a, double b)
```
guilpier-code:
```C++
static double hurdleCostCoefficient(double unsuppThermalCost1, double unsuppThermalCost2)
{
    ....
}
```

guilpier-code (changes requested):

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
flomnes: fixed in #1165

guilpier-code (approved):

Issue #927, opened 2022-10-27
Computing output files' max number of columns before print


guilpier-code:
Currently, computing a max number of columns before print is made too roughly and spread in several steps at different moment of the simulation. 
This is done in file **src/solver/variable/surveyresults/surveyresults.cpp** in function **initializeMaxVariables(...)**.
Consequently : 
- the code that computes this max number of columns should be gathered in one place
- if possible, this computation should be more accurate

PR #1159, opened 2023-02-08
Max number of columns in an output file

guilpier-code:
The current PR was first an attempt the issue #927 

It now contains a preparatory refactoring work before addressing the real problem raised by #927.
Issue #927 is addressed in #1172. 

**The original purpose was** : 
The purpose is to compute more precisely the max number of columns that a output report can have than it is today.
There is only one container that holds the results (see `SurveyResults::maxVariables`), and it has to be sized at the max number of columns a report can have.
As the synthesis output reports have more columns than the year-by-year reports, we build this max number from their number of columns.
The goal here can be summarized as follows : 
```C++
void SurveyResults::initializeMaxVariables(const Data::Study& s)
{
     // Getting the any report's max number of columns     
     maxVariables = s.parameters.variablesPrintInfo.getMaxColumnsCount();
     logs.debug() << "  (for " << maxVariables << " columns)";
}
```

**Tests** : 
To validate this work, tests are passed : on [this repo](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR), in directory **valid-filter-outputs**.

**Eventually** : 
This first attempt is a half-failure.
It gave the opportunity to make some refactoring and cleaning but the purpose of this issue could not be reached.
For further developments allowing to thoroughly reach the target, see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1159#issuecomment-1429581033).


src/libs/antares/study/variable-print-info.cpp
```diff
@@ -230,8 +185,8 @@ bool AllVariablesPrintInfo::isPrinted(std::string var_name) const
 
 void AllVariablesPrintInfo::setAllPrintStatusesTo(bool b)
 {
-    for (uint i = 0; i < size(); ++i)
-        allVarsPrintInfo[i]->enablePrint(b);
+    for (VariablePrintInfo* v : allVarsPrintInfo)
+        v->enablePrint(b);
 }
 
 void AllVariablesPrintInfo::computeMaxColumnsCountInReports()
```
guilpier-code, line 220: About **computeMaxColumnsCountInReports()** : 
This computation is strange.
There are 2 degrees of freedom : **FileLevel** (value, id, details, ...) and **DataLevel** (area, links, ...).
But here we make just one loop. Should not we make 2 nested loops ?

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -186,42 +171,12 @@ void AllVariablesPrintInfo::prepareForSimulation(bool userSelection,
...
-    // The search for a variable starts from the beginning of the variable print info list.
-
-    for (; it_info != allVarsPrintInfo.end(); it_info++)
-    {
-        if ((*it_info)->name() == var_name)
-        {
-            return (*it_info)->isPrinted();
-        }
-    }
+    auto has_name = [&](VariablePrintInfo* v) { return v->name() == var_name; };
```
guilpier-code: About **isPrinted** and **setPrintStatus** : 
These functions let think that **allVarsPrintInfo** should be a **std::map** instead of a **std::vector**.
If so, we could have : 
```C++
bool AllVariablesPrintInfo::isPrinted(std::string var_name) const
{
    return VariablePrintInfo[var_name]->isPrinted();
}
```
To be thought of (it may not be that easy)

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -86,12 +86,6 @@ void variablePrintInfoCollector::add(const AnyString& name,
 // ============================================================
```
guilpier-code, line 88: About **clear()** function : 
Currently, we need a **clear()** in class **AllVariablesPrintInfo** because the vector of **VariablePrintInfo** expects pointers to **VariablePrintInfo**.
Instead of : 
```C++
  std::vector<VariablePrintInfo*> allVarsPrintInfo;
```
we should have : 
```C++
  std::vector<VariablePrintInfo> allVarsPrintInfo;
```
This would allow to remove the **clear()** function and the **destructor** to be a **default** one.

guilpier-code:
As said the current PR's description, we could not reach the purpose of this PR.
The reason is that the operation of visiting all output variables (`RetrieveVariableList(collector)`) is static : it is made at compilation time.
When compiling, there is no way to know about the size required by the **dynamic** output variables 
(`columnCount = Category::dynamicColumns`).

A possible solution : 
- Today, **variablePrintInfoCollector** is built at compilation time and configured at runtime. But more importantly, besides being used to handle the print status of each variables, it is also required to collect the max size of each output variable, which it cannot do accurately (because of dynamic columns variables). So it should not be used to collect max size of variables.
- As we have to collect max size of each output variables at runtime, we could do what is done for other reasons elsewhere : defining and calling another function to traverse the static list of output variables at runtime. The function's prototype could be `void getMaxColumnsNumbe(SizeCollector &)`.
- A said in https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1159#discussion_r1105707798, the today's place to collect the max number of columns is not suitable. A possible suitable place where to instantiate the **SizeCollector** and run the **getMaxColumnsNumbe(...)** could be the `void ISimulation<Impl>::run()` method, right after the call of **variables.initializeFromStudy(...)**.

src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -504,10 +504,16 @@ static inline void WriteIndexHeaderToFileDescriptor(int precisionLevel,
     s += '\n';
 }
 
-uint initializeMaxVariables(uint maxVars, const Data::StudyRuntimeInfos* runtime)
+void SurveyResults::initializeMaxVariables(const Data::Study& s)
 {
+    const auto* runtime = s.runtime;
```
guilpier-code: Note that getting the max column count here is inefficient : located here, this operation is done for each yearly results and for synthesis results. But it could be done only once.
We should do it elsewhere : see https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1159#issuecomment-1429581033

flomnes: OK I identified a few steps

1. Compute the matrix size required to store all results
2. Allocate memory
3. Write to that memory
4. Dump that memory into files
5. Deallocate memory

What interests us here is step 1. Currently, it is done both statically and dynamically.

Static: Number of variables with fixed number of columns (mostly 1, sometimes more)
Dynamic: Number of clusters * number of variables by cluster (hard-coded)

Then we take the max of all this to make sure that every result fits (details-xxx.txt, values-xxx.txt, etc.).

For every file, there is a category id (de, de_res, etc.) that's stored in attribute `categoryFileLevel`. Example
```cpp
// src/solver/variable/economy/profitByPlant.h
categoryFileLevel = ResultsType::categoryFile & (Category::de),
// src/solver/variable/economy/dispatchable-generation-margin.h
categoryFileLevel = ResultsType::categoryFile & (Category::id | Category::va),
```

For every category, we need to compute the width & height of the associated matrix for yby & synthesis files.

Why not attach a function to every variable ? It could return 1 (most cases), another constant (thermal/renewable clusters), or a number depending on the study ("this variable produces one variable per thermal cluster")
```cpp
categoryFileLevel = ResultsType::categoryFile & (Category::de),
columnCalculator = [](const Study& study) { // return the number of thermal clusters };
```

Or, if you don't like lambda functions, we could have a small library of functions (return a constant, return the number of thermal clusters, return the number of inequality binding constraints).

guilpier-code: - Currently, the computation of max number of columns is done before the clusters or binding constraints are loaded.
So in the state of the current code, it would not work. 
That said, we could change the order in the data load.
- **Question** : we will have to traverse all output variables anyway whether we call a classic method or a lambda function, right ? How do you think we could do ?

flomnes: > Currently, the computation of max number of columns is done before the clusters or binding constraints are loaded.

No, see function `initializeMaxVariables` & constructor of class `SurveyResults`

guilpier-code: > No, see function `initializeMaxVariables` & constructor of class `SurveyResults`

`initializeMaxVariables` is based on information collected before the clusters and BCs are loaded.
`initializeMaxVariables` only gets their value.

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Done in #1172

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -186,42 +171,12 @@ void AllVariablesPrintInfo::prepareForSimulation(bool userSelection,
...
-    {
-        if ((*it_info)->name() == var_name)
-        {
-            return (*it_info)->isPrinted();
-        }
-    }
+    auto has_name = [&](VariablePrintInfo* v) { return v->name() == var_name; };
+    auto it = std::find_if(allVarsPrintInfo.begin(), allVarsPrintInfo.end(), has_name);
+    if (it != allVarsPrintInfo.end())
+        return (*it)->isPrinted();
```
flomnes: Using [std::any_of](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of) will save you a few lines of code

flomnes (changes requested):

flomnes:
Please resolve conflicts

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -76,161 +79,141 @@ variablePrintInfoCollector::variablePrintInfoCollector(AllVariablesPrintInfo* al
...
 
-void AllVariablesPrintInfo::add(VariablePrintInfo* v)
+void AllVariablesPrintInfo::add(std::string name, VariablePrintInfo v)
 {
-    allVarsPrintInfo.push_back(v);
+    std::string upperCaseName = to_uppercase(name);
+    if (not exists(upperCaseName))
+    {
+        index_to_name[(unsigned int)allVarsPrintInfo.size()] = upperCaseName;
+        allVarsPrintInfo.insert(std::pair<std::string, VariablePrintInfo>(upperCaseName, v));
```
flomnes: Use `try_emplace`, see https://www.fluentcpp.com/2018/12/11/overview-of-std-map-insertion-emplacement-methods-in-cpp17/

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -239,56 +222,58 @@ void AllVariablesPrintInfo::computeMaxColumnsCountInReports()
...
         {
-            if ((*it)->isPrinted() && (*it)->getFileLevel() & CFileLevel
-                && (*it)->getDataLevel() & CDataLevel)
-                currentColumnsCount += (*it)->getMaxColumnsCount();
+            uint currentColumnsCount = 0;
+            for (auto& [name, variable] : allVarsPrintInfo)
+            {
+                if (variable.isPrinted() &&
+                    variable.getFileLevel() & CFileLevel &&
+                    variable.getDataLevel() & CDataLevel)
```
flomnes:
```suggestion
                    variable.isFileLevel(CFileLevel) &&
                    variable.isDataLevel(CDataLevel))
```
Apply the `&` operator inside the method instead. Looks a bit more clear

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -239,56 +222,58 @@ void AllVariablesPrintInfo::computeMaxColumnsCountInReports()
...
 }
 
 void AllVariablesPrintInfo::countSelectedAreaVars()
 {
-    resetInfoIterator();
-    for (; it_info != allVarsPrintInfo.end(); it_info++)
+    for (auto& [name, variable] : allVarsPrintInfo)
     {
-        if ((*it_info)->isPrinted() && (*it_info)->getDataLevel() == Category::area)
+        if (variable.isPrinted() && variable.getDataLevel() == Category::area)
```
flomnes: Use `std::count_if`

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -239,56 +222,58 @@ void AllVariablesPrintInfo::computeMaxColumnsCountInReports()
...
 }
 
 void AllVariablesPrintInfo::countSelectedLinkVars()
 {
-    resetInfoIterator();
-    for (; it_info != allVarsPrintInfo.end(); it_info++)
+    for (auto& [name, variable] : allVarsPrintInfo)
     {
-        if ((*it_info)->isPrinted() && (*it_info)->getDataLevel() == Category::link)
+        if (variable.isPrinted() && variable.getDataLevel() == Category::link)
```
flomnes: Use `std::count_if`

src/libs/antares/study/variable-print-info.h
```diff
@@ -44,38 +44,35 @@ namespace Data
...
-        return dataLevel;
-    }
-    uint getFileLevel()
-    {
-        return fileLevel;
-    }
+    void setMaxColumns(uint maxColumnsNumber);
+
+    uint getDataLevel() { return dataLevel_; }
+    uint getFileLevel() { return fileLevel_; }
```
flomnes:
```suggestion
    bool isDataLevel(uint dataLevel) const { return dataLevel_ & dataLevel; }
    bool isFileLevel(uint fileLevel) const { return fileLevel_ & fileLevel; }
```

src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -489,54 +489,21 @@ static inline void WriteIndexHeaderToFileDescriptor(int precisionLevel,
...
-               static_cast<uint>(nbVariablesPerDetailRenewableCluster
-                                 * runtime->maxRenewableClustersForSingleArea),
-               maxNbVariablesPerInequalityBindingConstraint
-                 * runtime->getNumberOfInequalityBindingConstraints());
-}
-
 // TOFIX - MBO 02/06/2014 nombre de colonnes fonction du nombre de variables
-SurveyResults::SurveyResults(uint maxVars,
-                             const Data::Study& s,
+SurveyResults::SurveyResults(const Data::Study& s,
```
flomnes: You may now remove this comment

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -76,161 +79,141 @@ variablePrintInfoCollector::variablePrintInfoCollector(AllVariablesPrintInfo* al
...
-    }
-    return false;
+void AllVariablesPrintInfo::setMaxColumns(std::string varname, uint maxColumnsNumber)
+{
+    allVarsPrintInfo.at(to_uppercase(varname)).setMaxColumns(maxColumnsNumber);
 }
 
-void AllVariablesPrintInfo::prepareForSimulation(bool userSelection,
-                                                 const std::vector<std::string>& excluded_vars)
+std::string AllVariablesPrintInfo::name_of(unsigned int index) const
```
flomnes, line 147:
```suggestion
[[deprecated("Only needed by the GUI, to be removed")]]
std::string AllVariablesPrintInfo::name_of(unsigned int index) const
```

src/libs/antares/study/variable-print-info.h
```diff
@@ -97,30 +94,30 @@ class AllVariablesPrintInfo
...
+    ~AllVariablesPrintInfo() = default;
 
-    void add(VariablePrintInfo* v);
+    void add(std::string name, VariablePrintInfo v);
     void clear();
-    VariablePrintInfo* operator[](uint i) const;
+    VariablePrintInfo& operator[](uint i);
     size_t size() const;
-    bool isEmpty() const;
+    bool exists(std::string name);
```
flomnes, line 104: Use `const std::string&` or `std::string_view`

src/libs/antares/study/variable-print-info.h
```diff
@@ -132,27 +129,32 @@ class AllVariablesPrintInfo
...
 
-private:
-    void setAllPrintStatusesTo(bool b);
     void computeMaxColumnsCountInReports();
+    void setAllPrintStatusesTo(bool b);
+    void reverseAll();
+
+    unsigned int numberOfEnabledVariables();
+    std::vector<std::string> namesOfEnabledVariables();
+    std::vector<std::string> namesOfDisabledVariables();
```
flomnes, line 138: Avoid copies (`std::string_view` are non-owning references to char ranges)
```suggestion
    std::vector<std::string_view> namesOfEnabledVariables();
    std::vector<std::string_view> namesOfDisabledVariables();
```

flomnes (changes requested):
Minor changes required. LGTM

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Due to changes, remark not relevant anymore

src/libs/antares/study/variable-print-info.h
guilpier-code: I tried to replace **std::string** by **std::string_view** or  **const std::string&** in arguments of member functions : 
- **add(...)**
- **exists(...)**
- **setPrintStatus(...)**
- **isPrinted(...)**

but they all call **to_uppercase(std::string& str)** internally, that makes changes on the **std::string**.
So I leave the code as it is regarding this.

src/libs/antares/study/variable-print-info.h
guilpier-code: The returned **std::vector\<std::string\>** is an **rvalue**, meaning a temporary value.
Do we really want to make it a **std::vector\<std::string_view\>** ?

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Done

src/solver/variable/surveyresults/surveyresults.cpp
guilpier-code: Done

src/libs/antares/study/variable-print-info.h
guilpier-code: Done

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Done

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Done

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Done

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Done

guilpier-code:
As you suggested, I removed : 
- **VariablePrintInfo::hasDataLevel(...)**
- **VariablePrintInfo::hasFileLevel(...)**

flomnes (approved):

PR #1160, opened 2023-02-09
replaced g++11 by g++17 in common-settings.cmake

hugo-antoine-rtei:


src/cmake/common-settings.cmake
```diff
@@ -170,8 +170,8 @@ if(MSVC)
 	#SET(CMAKE_EXE_LINKER_FLAGS_DEBUG   "/debug /VERSION:${ANTARES_VERSION_HI}.${ANTARES_VERSION_LO}")
 	#SET(CMAKE_EXE_LINKER_FLAGS_RELEASE "/VERSION:${ANTARES_VERSION_HI}.${ANTARES_VERSION_LO}")
 else()
-	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -std=gnu++11")
-	set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG}   -std=gnu++11")
+	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -std=gnu++17")
+	set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG}   -std=gnu++17")
```
flomnes: I think we should remove lines 172-174 (maybe the ones above as well, but later on)

JasonMarechal25: To expand on the subject. The choice of the standard is dictated by the use of CMAKE_CXX_STANDARD
Flags should not be used directly.
Moreover you should not redefined c++ standard in submodules but leave it to the top CMakeList file to do it.

hugo-antoine-rtei: I deleted lines 172-174

flomnes (approved):

PR #1163, opened 2023-02-14
Short-term storage

### LINKED ISSUE #1065: Add short term storage objects
https://gopro-tickets.rte-france.com/browse/ANT-598

flomnes:
### TODO

- [x] Check that STS are available also in Adequacy mode -> seems fine, need to test with output files PR #1283 
- [x] Fix units for file "values" #1248 
- [x] Do not try to load STS if the version number is < 860 -> done in #1239
- [x] Output files #1238 
- [ ] Add unit & end-to-end tests / unit tests for loading OK
- [x] Leap year testing / cycles
- [x] Year starting in a month != January / cycles
- [ ] Cycles for days simulations instead of weeks


src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -153,6 +151,35 @@ typedef struct
...
 
+namespace ShortTermStorage
+{
+struct SINGLE_OBJECT_INPUT
+{
+    double capacity;
+    double injectionCapacity;
+    double withdrawalCapacity;
+    double efficiency;
+};
```
flomnes: Pointer to input data ?

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -514,6 +545,8 @@ struct PROBLEME_HEBDO
 
     PALIERS_THERMIQUES** PaliersThermiquesDuPays = nullptr;
     ENERGIES_ET_PUISSANCES_HYDRAULIQUES** CaracteristiquesHydrauliques = nullptr;
+    std::vector<::ShortTermStorage::AREA_INPUT>* ShortTermStorage = nullptr;
```
sylvlecl: Is it necessary to use a pointer instead of plain vector?

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -433,7 +462,9 @@ typedef struct
     double* debordementsHoraires;
 
     double* CoutsMarginauxHoraires;
-    PRODUCTION_THERMIQUE_OPTIMALE** ProductionThermique;
+    PRODUCTION_THERMIQUE_OPTIMALE** ProductionThermique; // index is pdtHebdo
+
+    std::vector<::ShortTermStorage::RESULTS>* ShortTermStorage = nullptr;
```
sylvlecl: Is it necessary to use a pointer instead of plain vector?

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -153,6 +151,37 @@ typedef struct
...
+struct AREA_INPUT
+{
+    std::vector<SINGLE_OBJECT_INPUT> storages; // index is local
+};
+// problemeHebdo->ShortTermStorage[area_idx].storages[ststor_idx].capacity;
+
+struct RESULTS
+{
+    // Index is the number of the STS in the area
+    std::vector<double> level;      // MWh
```
sylvlecl, line 181: What about having a small struct for each STS, like in the inputs ?

src/libs/antares/study/parts/short-term-storage/unit.h
```diff
@@ -0,0 +1,45 @@
...
+namespace Antares::Data::ShortTermStorage
+{
+class Unit
+{
+public:
+    bool validate() const;
+    bool loadFromFolder(const std::string& folder);
+
+private:
+    Series series;
```
sylvlecl: I guess those fields will need to have some public access

src/libs/antares/study/parts/short-term-storage/container.h
```diff
@@ -0,0 +1,50 @@
...
+{
+public:
+    bool validate() const;
+    // 1. Read list.ini
+    bool createUnitsFromIniFile(const std::string& path);
+    // 2. Read ALL series
+    bool loadSeriesFromFolder(const std::string& folder);
+
+    std::vector<Unit> storagesByIndex;
+    // TODO[FOM] : fill this map with the address from storagesByIndex
```
sylvlecl: It would make sense to have those fields not accessible for modification by outside world, to avoid that kind of consistency issue.

I am not sure of the workflow, but we could have a constructor with only the vector as parameter, which will populate the map, and only expose the vector and the map as const references (or not exposing the map at all and expose a `const Unit* getStorage(const std::string& id)` or something alike).

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: It turns out : no. Not here at least.

It depends on how the superseding structure is allocated. `PROBLEME_HEBDO` objects are allocated using `new`, so all all data members are properly initialized.

On the other hand, `malloc` only allocates memory, it doesn't call constructors for data members. So for example this will fail
```cpp
#include <cassert>
#include <cstdlib>

struct S
{
  int a;
  S() : a(5) {}
};

int main() {
  auto s = static_cast<S*>(malloc(sizeof(S)));
  assert(s->a == 5); // Failure
  free(s);
}
```
Objects of type `RESULTATS_HORAIRES` are allocated using `malloc`, so I got a segfault ;)

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: For this one, yes (see above)

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: That would be nice, will see if it works with the rest of the code. I essentially re-used the same logic as thermal clusters (multiple clusters per area, input & output data)

src/libs/antares/study/parts/short-term-storage/unit.h
flomnes: @payetvin please consider this

src/libs/antares/study/parts/short-term-storage/container.h
flomnes: @payetvin please consider this

src/solver/simulation/sim_structure_probleme_economique.h
sylvlecl: Understood! this will have to wait for deeper changes :)

src/solver/simulation/sim_structure_probleme_economique.h
sylvlecl: Ok, no strong opinion here: it can be better to keep consistent with existing code, waiting for a larger refactoring in the future

flomnes: It's no big deal, just swap 2 indices

src/libs/antares/study/area/list.cpp
```diff
@@ -1279,6 +1287,35 @@ bool AreaList::loadFromFolder(const StudyLoadOptions& options)
         }
     }
 
+    // Short term storage data, specific to areas
+    {
+        logs.info() << "Loading short term storage clusters...";
+        buffer.clear() << pStudy.folderInput << SEP << "st-storage" << SEP << "areas.ini";
+        struct stat sb;
+        if (!stat(buffer.c_str(), &sb))
```
flomnes: `IO::Directory::Exists`

flomnes: https://en.cppreference.com/w/cpp/filesystem/exists
https://en.cppreference.com/w/cpp/filesystem/directory_entry/is_directory

src/solver/variable/economy/shortTermStorageLevels.h
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    void hourForClusters(State& state, unsigned int numSpace)
+    {
+        for (uint stsIndex = 0; stsIndex < state.area->shortTermStorage.storagesByIndex.size(); stsIndex++)
+        {
+            const auto cluster = state.area->shortTermStorage.storagesByIndex[stsIndex];
+            const uint group = Antares::Data::ShortTermStorage::groupIndex(cluster->properties.group);
+            pValuesForTheCurrentYear[numSpace][group][state.hourInTheYear] += (*state.hourlyResults->ShortTermStorage)[state.hourInTheWeek].level[stsIndex];
+        }
```
flomnes: This is where the extraction takes place

src/libs/antares/study/parts/short-term-storage/cluster.h
```diff
@@ -0,0 +1,50 @@
...
+
+#pragma once
+#include <string>
+#include <antares/inifile.h>
+#include "properties.h"
+#include "series.h"
+
+namespace Antares::Data::ShortTermStorage
+{
+class STstorageCluster
```
sylvlecl: capital S if we want to respect `CamelCase`

src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -0,0 +1,77 @@
...
+    Other5
+};
+
+unsigned int groupIndex(Group group);
+
+class Properties
+{
+public:
+    bool validate();
+    bool loadKey(const IniFile::Property* p);
```
sylvlecl, line 56: From the implementation it seems the property is expected to be not-null, we must use a reference instead of a pointer

src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -0,0 +1,77 @@
...
+    // Efficiency factor between 0 and 1
+    double efficiencyFactor;
+    // Cycle duration, 1 <= cycleDuration <= 168
+    unsigned int cycleDuration;
+    // Used to sort outputs
+    Group group = Group::Other1;
+    // cluster name
+    std::string name;
+
+    static const std::map<std::string, enum Group> stStoragePropertyGroupEnum;
```
sylvlecl: We have not defined naming conventions, but I would suggest to name it with UPPER_CASE to make clear it's a constant.

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -0,0 +1,147 @@
...
+    case Group::PSP_closed: return 1;
+    case Group::Pondage: return 2;
+    case Group::Battery: return 3;
+    case Group::Other1: return 4;
+    case Group::Other2: return 5;
+    case Group::Other3: return 6;
+    case Group::Other4: return 7;
+    case Group::Other5: return 8;
+    default:
+      return 0;
```
sylvlecl: Shouldn't we throw an error here ?
If we add new values some day and forget to update this, it may cause silent bugs.

src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -0,0 +1,77 @@
...
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+
+#include <optional>
+#pragma once
```
sylvlecl, line 27: Should be above include?

src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -0,0 +1,77 @@
...
+    double injectionCapacity;
+    // Withdrawal nominal capacity, >= 0
+    double withdrawalCapacity;
+    // Reservoir capacity in MWh, >= 0
+    double capacity;
+    // Initial level, <= capacity
+    std::optional<double> initialLevel;
+    // Efficiency factor between 0 and 1
+    double efficiencyFactor;
+    // Cycle duration, 1 <= cycleDuration <= 168
```
sylvlecl, line 67: This is not checked at validation time

src/solver/variable/state.h
```diff
@@ -169,24 +162,20 @@ class State
...
     VALEURS_DE_NTC_ET_RESISTANCES* ntc;
-    //! Thermal production for the current thermal cluster for the current hour in the year
-    double thermalClusterProduction;
-    //! Thermal available production for the current thermal cluster for the current hour in the
-    //! year
-    double thermalClusterAvailableProduction;
-    //! The operating cost for the current cluster of the current hour (production level*production
+
+    //! Thermal production for thermal clusters for the current hour in the year
+    std::vector<std::vector<double>> thermalClustersProductions;
```
sylvlecl: This refactoring really does not clarify the class from my point of view:
1. why do we have data for all clusters and areas, whereas we have a "current cluster" and a "current area" ?
2. what is the indexing of those vectors ? In general it  is better to create strucs/classes instead of having nested collections, or at least expose getters so that users of the class don't HAVE to know that implementation detail

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -73,6 +70,95 @@ void exportPaliers(const PROBLEME_HEBDO& problemeHebdo,
...
+    const auto& VarOptim_current = CorrespondanceVarNativesVarOptim[pdt];
+    // Cycle over the simulation period
+    const int pdt1 = (pdt + 1) % nombreDePasDeTempsPourUneOptimisation;
+    const auto& VarOptim_next = CorrespondanceVarNativesVarOptim[pdt1];
+    for (auto& storage : shortTermStorageInput)
+    {
+        int nombreDeTermes = 0;
+        const int globalIndex = storage.globalIndex;
+        // L[h+1] - L[h] - efficiency * injection[h] + withdrawal[h] = inflows[h]
+        if (const int varLevel_next = VarOptim_next->ShortTermStorage.LevelVariable[globalIndex];
```
sylvlecl: For the future, I think we should think of syntactic improvements to make this kind of repetitive code more concise, readable and maintainable (always the same operations: get var number, fill vectors, increment count ... ).
For ex:
```cpp
// Helper class to build constraints
class ConstraintFiller {
public:
    ConstraintFiller(double *pi, int *colonne, int &nombreDeTermes)
            : Pi(pi), Colonne(colonne), nombreDeTermes(nombreDeTermes) {}

    void addTerm(int numVar, double value) {
        if (numVar >= 0)
        {
            Pi[nombreDeTermes] = value;
            Colonne[nombreDeTermes] = numVar;
            nombreDeTermes++;
        }
    }

private:
    double* Pi;
    int* Colonne;
    int& nombreDeTermes;
};


// Then in the method only one-liners
// L[h+1] - L[h] - efficiency * injection[h] + withdrawal[h] = inflows[h]
auto& nextHourStorage = VarOptim_next->ShortTermStorage;
auto& currentStorage = VarOptim_current->ShortTermStorage;
ConstraintFiller filler(Pi, Colonne, nombreDeTermes);
filler.addTerm(nextHourStorage.LevelVariable[globalIndex], 1.0);
filler.addTerm(currentStorage.LevelVariable[globalIndex], -1.0);
filler.addTerm(currentStorage.InjectionVariable[globalIndex], -1.0 * storage.efficiency);
filler.addTerm(currentStorage.WithdrawalVariable[globalIndex], 1.0);
```

sylvlecl: (or even better with method chaining, to avoid naming and repeating the "filler" variable)

src/libs/antares/study/parts/short-term-storage/properties.cpp
payetvin: We return an unsigned int so I don't know what value to return for error

flomnes: [`throw`](https://en.cppreference.com/w/cpp/language/throw)

src/libs/antares/study/parts/short-term-storage/properties.h
payetvin: if (!section.firstProperty)
        return false;
We check if first property isn't null and then iterate till null so no need to check in that function

src/libs/antares/study/parts/short-term-storage/properties.h
payetvin: TODO
Need to get Simplex optimization range global parameter to validate cycle duration

flomnes:
Careful: name != ID.

IDs are "sanitized" using function `TransformNameIntoID`, making sure all characters are "simple" and can be contained in a filename.

src/libs/antares/study/parts/short-term-storage/properties.h
sylvlecl: Yes, I am not saying that there is an error in the current code, but the point is to make clear to the caller of this method, who could possibly be anyone in the future since it's public, that he must not pass a null argument.

Using a reference is a semantic way in c++ to document that an argument is not nullable.

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
flomnes: See #1234

src/solver/variable/state.h
flomnes: See #1230

payetvin:
Move `bool loadFile(const std::string& path, std::vector<double>& vect)` in series.cpp to an I/O file

src/libs/antares/study/area/list.cpp
```diff
@@ -1279,6 +1290,28 @@ bool AreaList::loadFromFolder(const StudyLoadOptions& options)
...
+        logs.info() << "Loading short term storage clusters...";
+        buffer.clear() << pStudy.folderInput << SEP << "st-storage";
+        if (IO::Directory::Exists(buffer))
+        {
+            CString<30, false> stStoragePlant;
+            stStoragePlant << SEP << "st-storage" << SEP << "clusters" << SEP;
+
+            for (const auto& [id, area] : areas)
+            {
+                buffer.clear() << pStudy.folderInput << stStoragePlant << area->id;
```
guilpier-code: About : 
```C++
buffer.clear() << pStudy.folderInput << stStoragePlant << area->id;
```
We don't need to clear the buffer, it already contains a useful piece of address from line : 
```C++
buffer.clear() << pStudy.folderInput << SEP << "st-storage";
```
above.
Just add the "clusters" part.

guilpier-code (changes requested):

src/libs/antares/study/area/list.cpp
payetvin: done in #1246

src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -0,0 +1,76 @@
...
+
+unsigned int groupIndex(Group group);
+
+class Properties
+{
+public:
+    bool validate();
+    bool loadKey(const IniFile::Property* p);
+    // Not optional   Injection nominal capacity, >= 0
+    std::optional<double> injectionCapacity;
```
guilpier-code: Names do not fit the specifications :  
- **injectionCapacity** should be renamed **injectionNominalCapacity**, the **nominal** missing piece being quite meaningful.
- Same for withdrawal
- **capacity** should renamed **reservoirCapacity**

guilpier-code (changes requested):

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -153,6 +159,32 @@ typedef struct
...
+{
+    double capacity;
+    double injectionCapacity;
+    double withdrawalCapacity;
+    double efficiency;
+    std::optional<double> initialLevel;
+
+    std::shared_ptr<Antares::Data::ShortTermStorage::Series> series;
+
+    int globalIndex;
```
guilpier-code: `int globalIndex` :
I wondered for a short moment what exactly this index stood for.
If it was called **clusterGlobalIndex** (for instance), I would not have.

guilpier-code (changes requested):

src/libs/antares/study/area/list.cpp
```diff
@@ -1079,12 +1079,23 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
+        ret = area.shortTermStorage.validate() && ret;
+    }
+
     // Renewable cluster list
     if (study.header.version >= 810)
     {
         buffer.clear() << study.folderInput << SEP << "renewables" << SEP << "series";
         ret = area.renewable.list.loadDataSeriesFromFolder(study, options, buffer) && ret;
     }
+    if (!ret)
```
guilpier-code, line 955:
```C++
if (!ret)
   logs.error() << "Error while loading sts series";
```
Why not right after the ST storage load ? **ret == false** here could mean there were a problem when loading renewable clusters, not when loading ST storage time series.

guilpier-code (changes requested):

src/libs/antares/study/runtime/runtime.h
```diff
@@ -171,6 +171,8 @@ class StudyRuntimeInfos
     uint thermalPlantTotalCount;
     uint thermalPlantTotalCountMustRun;
 
+    uint shortTermStorageCount;
```
guilpier-code: Why not : 
```C++
unsigned int shortTermStorageCount = 0;
```
?

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -409,6 +411,7 @@ StudyRuntimeInfos::StudyRuntimeInfos(uint nbYearsParallel) :
  bindingConstraint(nullptr),
  thermalPlantTotalCount(0),
  thermalPlantTotalCountMustRun(0),
+ shortTermStorageCount(0),
```
guilpier-code: If `shortTermStorageCount` already initialized in **runtime.h**, we don't need to initialize it here.

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -446,7 +449,7 @@ bool StudyRuntimeInfos::loadFromStudy(Study& study)
     mode = gd.mode;
     thermalPlantTotalCount = 0;
     thermalPlantTotalCountMustRun = 0;
-
+    shortTermStorageCount = 0;
```
guilpier-code: Do we need this initialization (as `shortTermStorageCount` is supposed to be already initialized to **0**) ?

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -39,6 +39,34 @@ using namespace Antares;
 using namespace Antares::Data;
 using namespace Yuni;
 
+static void importShortTermStorages(
```
guilpier-code, line 42: We could take advantage of this function to compute the total number of ST storage clusters and store it into `problem.NumberOfShortTermStorages`
We would have : 
```C++
static void importShortTermStorages(const AreaList& areas, PROBLEME_HEBDO& problem)
{
    ...
}
```
Doing this would allow to remove the computation of this number from **runtime.cpp**, and even remove all references to ST storage from class **StudyRuntimeInfos**, which I don't like because it stores all kinds of things, don't store proper runtime data and is probably useless eventually.

If we decide to make this move, we'll have to call function **importShortTermStorages** before function **SIM_AllocationProblemeHebdo** : 
```C++
void SIM_InitialisationProblemeHebdo(Data::Study& study,
                                     PROBLEME_HEBDO& problem,
                                     int NombreDePasDeTemps,
                                     uint numSpace)
{
    ...
    importShortTermStorages(study.areas, problem);
    SIM_AllocationProblemeHebdo(problem, NombreDePasDeTemps);
    ...
}
```

guilpier-code (changes requested):

src/libs/antares/study/runtime/runtime.h
guilpier-code: Please see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1163#discussion_r1162733128)

It suggests to remove `unsigned int shortTermStorageCount` from **runtime.h**

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -66,6 +65,12 @@ typedef struct
     int* NumeroDeVariableDuNombreDeGroupesQuiSArretentDuPalierThermique;
     int* NumeroDeVariableDuNombreDeGroupesQuiTombentEnPanneDuPalierThermique;
 
+    struct
+    {
+        int* InjectionVariable;
+        int* WithdrawalVariable;
+        int* LevelVariable;
+    } ShortTermStorage;
```
guilpier-code: There are too many things called **ShortTermStorage** : a namespace, a **struct**, a vector, ...
This is confusing, please try to do some renaming : **st_storage_names**,  ...

src/solver/optimisation/opt_decompte_variables_et_contraintes.cpp
```diff
@@ -220,6 +220,15 @@ int OPT_DecompteDesVariablesEtDesContraintesDuProblemeAOptimiser(PROBLEME_HEBDO*
         }
     }
 
+    for (int areaIndex = 0; areaIndex < problemeHebdo->NombreDePays; areaIndex++)
```
guilpier-code: Instead of making a loop over areas and pick the number of ST storage clusters in each area, couldn't we use `problemeHebdo->NumberOfShortTermStorages`  ?

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
```diff
@@ -157,6 +157,63 @@ void setBoundsForUnsuppliedEnergy(PROBLEME_HEBDO* problemeHebdo,
...
+                                      .WithdrawalVariable[globalIndex];
+                Xmin[varWithdrawal] = 0.;
+                Xmax[varWithdrawal]
+                  = storage.withdrawalCapacity * storage.series->maxWithdrawalModulation[pdtHebdo];
+                AddressForVars[varWithdrawal] = &STSResult.withdrawal[areaWideIndex];
+
+                // 3. Levels
+                int varLevel
+                  = CorrespondanceVarNativesVarOptim->ShortTermStorage.LevelVariable[globalIndex];
+                if (pdtHebdo == PremierPdtDeLIntervalle && storage.initialLevel.has_value())
```
guilpier-code, line 201: About : 
```C++
if (pdtHebdo == PremierPdtDeLIntervalle && storage.initialLevel.has_value())
```
if the simulation period initial level is fixed by user, why do we need bounds for it ?

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/container.h
```diff
@@ -0,0 +1,49 @@
...
+public:
+    bool validate() const;
+    // 1. Read list.ini
+    bool createSTStorageClustersFromIniFile(const std::string& path);
+    // 2. Read ALL series
+    bool loadSeriesFromFolder(const std::string& folder) const;
+    // Number of ST storages
+    std::size_t count() const;
+
+    std::vector<STStorageCluster*> storagesByIndex;
```
guilpier-code, line 46: Why is it called **storagesByIndex** ?  it's simply a list of clusters. Why not call it **clusters** ?

guilpier-code (dismissed):

src/libs/antares/study/area/list.cpp
payetvin: Changed in #1239

src/solver/simulation/sim_calcul_economique.cpp
flomnes: I agree, but having `shortTermStorageCount` in runtime.h allows to display this log (runtime.cpp)
```cpp
    logs.info() << "     short-term storages: " << shortTermStorageCount;
```

I'm not sure we could display it so easily if it was only in `PROBLEME_HEBDO`.

src/solver/optimisation/opt_decompte_variables_et_contraintes.cpp
flomnes: See 996ae317

src/libs/antares/study/runtime/runtime.h
flomnes: See 4187a85

src/libs/antares/study/parts/short-term-storage/container.h
flomnes: To make it clear that variables should be accessed by index. Same idea as `areas.byIndex`

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
flomnes: We could move it to the RHS, but it would force us to make a special case, remove one variable from the problem, etc.

It's possible, but adds a bit of complexity. The decision is to let the solver handle this complexity.

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: `SIM_ShortTermStorage` ?

src/libs/antares/study/runtime/runtime.cpp
flomnes: See 3910a5e

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: @payetvin Can you perform this renaming ?

src/libs/antares/study/parts/short-term-storage/properties.h
flomnes: @payetvin Can you perform this renaming ?

payetvin: should be done after merging #1239

payetvin (approved):

PR #1172, opened 2023-02-17
Max nb columns in output reports : computing an accurate max number of columns (based on synthesis reports)

guilpier-code:
The current PR follows #1159, which is merely a preparatory cleaning PR.
The current PR adresses the real problem of issue #927 :  computing an accurate max number of columns an output report can contain.

Note that, as synthesis reports always contain more columns than any other report, the max number of columns computation is based on the number of columns an output variable can take in a synthesis report.

**Execution steps** : 
- Creating all output variables print info list, that is filling **variablesPrintInfo** where each variable's print status is **true**
  See **parameters.cpp** : `RetrieveVariableList(collector)`.
  At this point, clusters or BCs are not loaded, we can't give a  max column width to the dynamic or BCs output variables.
- Collecting print info status for each output variable (see reading functions in **parameters.cpp**)
- Collecting the max column width of each output variables when clusters or BCs are loaded.
  This is done inside **solver.hxx** : `variables.initializeFromStudy(study)`
  See the calls to `supplyMaxNumberOfColumns(study)`
- Computing the max number of columns a report can contains.
  This is done when calling **solver.hxx** : parameters.variablesPrintInfo.computeMaxColumnsCountInReports();
  By looping over all the kinds of generated reports, getting the number of columns they contain,  and storing the largest size. 
- Getting the already computed largest size in `SurveyResults::initializeMaxVariables`, just before printing

### DIFF
--- a/src/solver/variable/bindConstraints.hxx
+++ b/src/solver/variable/bindConstraints.hxx
@@ -122,6 +122,15 @@ void BindingConstraints<NextT>::initializeFromStudy(Data::Study& study)
         // Does user want to print output results related to the current binding constraint ?
         bc.getPrintStatusFromStudy(study);
     }
+
+    // This is ugly (it's a work around). We should try to improve this.
+    // Supplying the max number of columns to the variable print info collector 
+    if (pBCcount > 0)
+    {
+        NextType& bc = pBindConstraints[0];
+        bc.setBindConstraintsCount(pBCcount);
+        bc.supplyMaxNumberOfColumns(study);
+    }
 }
 
 template<class NextT>
--- a/src/solver/variable/bindConstraints.hxx
+++ b/src/solver/variable/bindConstraints.hxx
@@ -99,12 +99,12 @@ inline void BindingConstraints<NextT>::provideInformations(I& infos)
 template<class NextT>
 void BindingConstraints<NextT>::initializeFromStudy(Data::Study& study)
 {
-    const std::vector<uint> InequalityBCnumbers
+    const std::vector<uint> InequalityBCindices
       = study.runtime->getIndicesForInequalityBindingConstraints();
 
     // The total number of inequality binding constraints count
     // (we don't count BCs with equality sign)
-    pBCcount = (uint)InequalityBCnumbers.size();
+    pBCcount = (uint)InequalityBCindices.size();
 
     // Reserving the memory
     if (pBCcount > 0)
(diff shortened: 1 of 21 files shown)


flomnes:
General remark: are you sure you want to use the **synthesis** case ? For year-by-year, it overestimates the number of columns by a factor of ~4 (min, max, exp, std).

guilpier-code:
> General remark: are you sure you want to use the **synthesis** case ? For year-by-year, it overestimates the number of columns by a factor of ~4 (min, max, exp, std).

What we just did here is computing accurately was what done before approximately.
We always have had one column size for all reports, whether it is a synthesis report or a year by year report.
If we want to adapt the column size to the report we currently print, we can think about it.

src/solver/variable/bindConstraints.hxx
```diff
@@ -122,6 +122,15 @@ void BindingConstraints<NextT>::initializeFromStudy(Data::Study& study)
         // Does user want to print output results related to the current binding constraint ?
         bc.getPrintStatusFromStudy(study);
     }
+
+    // This is ugly (it's a work around). We should try to improve this.
+    // Supplying the max number of columns to the variable print info collector 
+    if (pBCcount > 0)
+    {
+        NextType& bc = pBindConstraints[0];
+        bc.setBindConstraintsCount(pBCcount);
```
flomnes, line 140: We only want "<" or ">" binding constraints here

src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -509,39 +509,8 @@ void SurveyResults::initializeMaxVariables(const Data::Study& s)
     const auto* runtime = s.runtime;
 
     // Getting the any report's max number of columns
-    maxVariables = s.parameters.variablesPrintInfo.getMaxColumnsCount();
+    maxVariables = s.parameters.variablesPrintInfo.getTotalMaxColumnsCount();
```
flomnes: NICE !

flomnes (commented):
Can you briefly explain the logic of your dev ? What steps are involved, who's responsible for what, etc.

src/solver/variable/bindConstraints.hxx
guilpier-code: Ok, see this PR's top comment

guilpier-code:
> Can you briefly explain the logic of your dev ? What steps are involved, who's responsible for what, etc.

Ok, see this PR's top comment

src/solver/variable/bindConstraints.hxx
guilpier-code: That's we have I think : **pBCcount** is the number of inequality BCs, isn't it ?
Maybe we should rename it into **inequalityBCcount_**

src/solver/variable/bindConstraints.hxx
```diff
@@ -122,6 +122,15 @@ void BindingConstraints<NextT>::initializeFromStudy(Data::Study& study)
         // Does user want to print output results related to the current binding constraint ?
         bc.getPrintStatusFromStudy(study);
     }
+
+    // This is ugly (it's a work around). We should try to improve this.
```
guilpier-code: As commented in the code says, we have a workaround here.
Instead of : 
```C++
if (pBCcount > 0)
{
     NextType& bc = pBindConstraints[0];
     bc.setBindConstraintsCount(pBCcount);
     bc.supplyMaxNumberOfColumns(study);
}
```
we should have : 
```C++
NextType::supplyMaxNumberOfColumns(study);
```
as it is done for **Areas** and **Links**.
This is due to the fact that BCs output variables are not **dynamic** output variables.
I left a comment on this **directly in the code** in the (following) PR #1180

flomnes: I think this workaround is OK for now. At least we no longer have hard-coded constants such as "4 variables for each thermal clusters" and so on.

src/solver/variable/bindConstraints.hxx
flomnes: Yes, please do

flomnes (changes requested):

PR #1180, opened 2023-02-21
max nb columns for print : more cleaning

guilpier-code:
The current PR follows #1172.
This is a cleaning work.

**All PR about max nb columns for print** : 
There are currently 3 PR about  **max nb columns for print**
- #1159 : some preliminary cleaning
- #1172 : adresses the real problem : computing this max number of columns accurately
- The current PR

**What this PR is about** :
The first target was to replace an **std::vector** with a **std::map** in class **AllVariablesPrintInfo** and hope it would simplify the code. As a matter of fact, it does.
What was done actually : 
-  replacing an **std::vector** with a **std::map** in class **AllVariablesPrintInfo**, it simplifies methods **setPrintStatus(...)**, **setMaxColumns(...)** and **isPrinted(...)**
- In this **std::map** (that associates **name of vars** to instance of **VariablesPrintInfo**), avoid inserting **VariablesPrintInfo** as pointers. This allows to make the destructor of  **AllVariablesPrintInfo** empty.
- Introducing some more methods to class **AllVariablesPrintInfo** to simplify the code elsewhere, see : 
     - **src/libs/antares/study/parameters.cpp**
     - 	**src/ui/simulator/toolbox/components/datagrid/renderer/select-variables.cpp**
     - **src/ui/simulator/windows/options/select-output/select-output.cpp**
- some more renaming and comments left in the code

### DIFF
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1874,7 +1874,7 @@ void Parameters::saveToINI(IniFile& ini) const
     // Variable selection
     {
         uint nb_tot_vars = (uint)variablesPrintInfo.size();
-        uint nb_selected_vars = (uint)variablesPrintInfo.namesOfEnabledVariables.size();
+        uint nb_selected_vars = (uint)variablesPrintInfo.namesOfEnabledVariables().size();
 
         if (nb_selected_vars != nb_tot_vars)
         {
@@ -1883,12 +1883,12 @@ void Parameters::saveToINI(IniFile& ini) const
             if (nb_selected_vars <= (nb_tot_vars / 2))
             {
                 section->add("selected_vars_reset", "false");
-                for (auto& name : variablesPrintInfo.namesOfEnabledVariables)
+                for (auto& name : variablesPrintInfo.namesOfEnabledVariables())
                     section->add("select_var +", name);
             }
             else
             {
-                for (auto& name : variablesPrintInfo.namesOfDisabledVariables)
+                for (auto& name : variablesPrintInfo.namesOfDisabledVariables())
                     section->add("select_var -", name);
             }
         }
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1874,7 +1874,7 @@ void Parameters::saveToINI(IniFile& ini) const
     // Variable selection
     {
         uint nb_tot_vars = (uint)variablesPrintInfo.size();
-        uint nb_selected_vars = (uint)variablesPrintInfo.namesPrinted.size();
+        uint nb_selected_vars = (uint)variablesPrintInfo.namesOfEnabledVariables.size();
 
         if (nb_selected_vars != nb_tot_vars)
         {
@@ -1883,12 +1883,12 @@ void Parameters::saveToINI(IniFile& ini) const
             if (nb_selected_vars <= (nb_tot_vars / 2))
             {
                 section->add("selected_vars_reset", "false");
-                for (auto name : variablesPrintInfo.namesPrinted)
+                for (auto& name : variablesPrintInfo.namesOfEnabledVariables)
                     section->add("select_var +", name);
             }
             else
             {
-                for (auto name : variablesPrintInfo.namesNotPrinted)
+                for (auto& name : variablesPrintInfo.namesOfDisabledVariables)
                     section->add("select_var -", name);
             }
         }
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -966,23 +966,20 @@ static bool SGDIntLoadFamily_VariablesSelection(Parameters& d,
 {
     if (key == "selected_vars_reset")
     {
-        bool mode = value.to<bool>();
-        if (mode)
-        {
-            for (uint i = 0; i != d.variablesPrintInfo.size(); ++i)
-                d.variablesPrintInfo[i]->enablePrint(true);
-        }
-        else
-        {
-            for (uint i = 0; i != d.variablesPrintInfo.size(); ++i)
-                d.variablesPrintInfo[i]->enablePrint(false);
-        }
+        bool printAllVariables = value.to<bool>();
+        d.variablesPrintInfo.setAllPrintStatusesTo(printAllVariables);
         return true;
     }
-    if (key == "select_var +")
-        return d.variablesPrintInfo.setPrintStatus(value.to<std::string>(), true);
-    if (key == "select_var -")
-        return d.variablesPrintInfo.setPrintStatus(value.to<std::string>(), false);
+    if (key == "select_var +" || key == "select_var -")
+    {
+        // Check if the read output variable exists 
+        if (not d.variablesPrintInfo.exists(value.to<std::string>()))
+            return false;
+
+        bool is_var_printed = (key == "select_var +");
+        d.variablesPrintInfo.setPrintStatus(value.to<std::string>(), is_var_printed);
+        return true;
+    }
     return false;
 }
 static bool SGDIntLoadFamily_SeedsMersenneTwister(Parameters& d,
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1928,17 +1928,17 @@ bool Parameters::saveToFile(const AnyString& filename) const
 
 void Parameters::RenewableGeneration::addExcludedVariables(std::vector<std::string>& out) const
 {
-    const static std::vector<std::string> ren = {"wind offshore",
-                                                 "wind onshore",
-                                                 "solar concrt.",
-                                                 "solar pv",
-                                                 "solar rooft",
-                                                 "renw. 1",
-                                                 "renw. 2",
-                                                 "renw. 3",
-                                                 "renw. 4"};
+    const static std::vector<std::string> ren = {"WIND OFFSHORE",
+                                                 "WIND ONSHORE",
+                                                 "SOLAR CONCRT.",
+                                                 "SOLAR PV",
+                                                 "SOLAR ROOFT",
+                                                 "RENW. 1",
+                                                 "RENW. 2",
+                                                 "RENW. 3",
+                                                 "RENW. 4"};
 
-    const static std::vector<std::string> agg = {"wind", "solar"};
+    const static std::vector<std::string> agg = {"WIND", "SOLAR"};
 
     switch (rgModelling)
     {
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1889,7 +1889,7 @@ void Parameters::saveToINI(IniFile& ini) const
             else
             {
                 for (auto name : variablesPrintInfo.namesNotPrinted)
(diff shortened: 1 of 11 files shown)


src/libs/antares/study/variable-print-info.cpp
```diff
@@ -83,120 +82,110 @@ void variablePrintInfoCollector::add(const AnyString& name,
...
-    for (VariablePrintInfo* v : allVarsPrintInfo)
-        v->enablePrint(b);
+    for (auto pair : allVarsPrintInfo)
+        pair.second.enablePrint(b);
+}
+
+void AllVariablesPrintInfo::reverseAll()
+{
+    for (auto pair : allVarsPrintInfo)
+        pair.second.reverse();
```
flomnes:
```suggestion
    for (auto& pair : allVarsPrintInfo)
        pair.second.reverse();
```
Otherwise you'd change a copy, not the elements themselves. See [this example](https://gist.github.com/flomnes/008920c58ba607c91d201b42a543507c).

guilpier-code: Thanks, this would have caused bugs surely !
Corrected now.

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -220,15 +219,15 @@ void AllVariablesPrintInfo::computeMaxColumnsCountInReports()
         for (uint CFileLevel = 1; CFileLevel <= Category::maxFileLevel; CFileLevel *= 2)
         {
             uint currentColumnsCount = 0;
-            for (auto it = allVarsPrintInfo.begin(); it != allVarsPrintInfo.end(); it++)
+            for (auto& pair : allVarsPrintInfo)
```
flomnes: C++17 offers the more concise syntax `for(auto& [key, value] : myMap)`. Make sure to use it with expressive names for `key` and `value`.

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -239,18 +238,18 @@ void AllVariablesPrintInfo::computeMaxColumnsCountInReports()
...
 
 void AllVariablesPrintInfo::countSelectedLinkVars()
 {
-    for (VariablePrintInfo* v : allVarsPrintInfo)
+    for (auto& pair : allVarsPrintInfo)
     {
-        if (v->isPrinted() && v->getDataLevel() == Category::link)
+        if (pair.second.isPrinted() && pair.second.getDataLevel() == Category::link)
             numberSelectedLinkVariables++;
     }
```
flomnes: https://en.cppreference.com/w/cpp/algorithm/count

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -83,120 +82,120 @@ void variablePrintInfoCollector::add(const AnyString& name,
...
 
-void AllVariablesPrintInfo::prepareForSimulation(bool userSelection,
+void AllVariablesPrintInfo::prepareForSimulation(bool isThematicTrimmingEnabled,
                                                  const std::vector<std::string>& excluded_vars)
 {
-    assert(!isEmpty() && "The variable print info list must not be empty at this point");
-
     // Initializing output variables status
-    if (!userSelection)
+    if (not isThematicTrimmingEnabled)
```
flomnes:
```suggestion
    if (!isThematicTrimmingEnabled)
```

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -83,120 +82,120 @@ void variablePrintInfoCollector::add(const AnyString& name,
...
-    allvarsinfo->add(new VariablePrintInfo(name, dataLevel, fileLevel));
+    allvarsinfo->add(name.to<std::string>(), VariablePrintInfo(dataLevel, fileLevel));
 }
 
 // ============================================================
 // All variables print information
 // ============================================================
 
-AllVariablesPrintInfo::~AllVariablesPrintInfo()
+static std::string to_uppercase(std::string str)
```
flomnes:
```suggestion
static std::string to_uppercase(const std::string& str)
```

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -83,120 +82,120 @@ void variablePrintInfoCollector::add(const AnyString& name,
...
 {
-    clear();
+    std::transform(str.begin(), str.end(), str.begin(), ::toupper);
+    return str;
 }
 
-void AllVariablesPrintInfo::add(VariablePrintInfo* v)
+void AllVariablesPrintInfo::add(std::string name, VariablePrintInfo v)
```
flomnes, line 98:
```suggestion
void AllVariablesPrintInfo::add(const std::string& name, VariablePrintInfo v)
```

src/libs/antares/study/variable-print-info.cpp
```diff
@@ -83,120 +82,120 @@ void variablePrintInfoCollector::add(const AnyString& name,
...
 {
-    allVarsPrintInfo.push_back(v);
+    std::string upperCaseName = to_uppercase(name);
+    if (not exists(upperCaseName))
+    {
+        index_to_name[(unsigned int)allVarsPrintInfo.size()] = upperCaseName;
+        allVarsPrintInfo.insert(std::pair<std::string, VariablePrintInfo>(upperCaseName, v));
```
flomnes, line 104:
```suggestion
        allVarsPrintInfo[upperCaseName] = v;
```

flomnes (commented):
You need strategy to keep `namesPrinted` & `namesNotPrinted` up-to-date when enabling/disabling an output variable.

Either call it by after every for-loop on adding/removing variables. I think it might be out-of-sync in some cases, for example parameters.cpp:1883.

Or (this looks better) you could use a bool `sync` flag
- At construction, `sync=true` (everything is empty, hence synchronized)
- At every write, `sync=false`
- At every read (including e.g `namesPrinted.size()`),
  - if `sync==false`, call `sortVariablesByPrintStatus` then set `sync=true`
  - if `sync==true`, don't do anything
 
You'd need to encapsulate `namesPrinted.size()` into a dedicated method.

src/libs/antares/study/variable-print-info.cpp
guilpier-code: About that, I first wrote the line : 
```C++
allVarsPrintInfo[upperCaseName] = v;
```
But it does not compile if class **VariablePrintInfo** (type of the values in the **std::map**) does not have an explicitly defined **default constructor**.
Indeed, as you may know, the **std::map [] operator** needs to add a default value in the map in case the key does have already a associated value.
As I didn't want this **default constructor** to be defined (to make code shorter), I opted for the **insert** method.

guilpier-code:
> You need strategy to keep `namesPrinted` & `namesNotPrinted` up-to-date when enabling/disabling an output variable.

This has been taken care of. Tell me what you think.

src/libs/antares/study/variable-print-info.cpp
guilpier-code: More readable indeed !
Suggestion applied.

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Here have to choose between use of std algorithm and readability : 
```C++
numberSelectedAreaVariables = std::count_if(
        allVarsPrintInfo.begin(),
        allVarsPrintInfo.end(),
        [](auto& p) {return p.second.isPrinted() &&
                            p.second.getDataLevel() == Category::area; });
```
or 
```C++
    for (auto& [name, variable] : allVarsPrintInfo)
    {
       if (variable.isPrinted() && variable.getDataLevel() == Category::area)
           numberSelectedAreaVariables++;
    }
```

src/libs/antares/study/variable-print-info.cpp
guilpier-code: Giving a **const string&** to **std::transform** does not compile. But **string&** is ok

src/libs/antares/study/variable-print-info.cpp
guilpier-code: see answer below

src/libs/antares/study/variable-print-info.h
```diff
@@ -104,18 +95,21 @@ class AllVariablesPrintInfo
...
-    ~AllVariablesPrintInfo();
+    ~AllVariablesPrintInfo() = default;
 
-    void add(VariablePrintInfo* v);
+    void add(std::string name, VariablePrintInfo v);
     void clear();
-    VariablePrintInfo* operator[](uint i) const;
+    VariablePrintInfo& operator[](uint i);
     size_t size() const;
-    bool isEmpty() const;
```
guilpier-code, line 113: We don't need **isEmpty()**, since the output variables are built at compile time and their number is fixed

flomnes (commented):
It seems that this PR kind of misses the goal of simplifying the management of printed/hidden variables.

We introduce 2 vectors (`namesOfEnabledVariables` and `namesOfDisabledVariables`) and a map `index_to_name` that need to be kept synchronized when adding/removing/reverting a variable.

Maybe we should do it differently ? Let's discuss this option

guilpier-code:
> It seems that this PR kind of misses the goal of simplifying the management of printed/hidden variables.
> 
> We introduce 2 vectors (`namesOfEnabledVariables` and `namesOfDisabledVariables`) and a map `index_to_name` that need to be kept synchronized when adding/removing/reverting a variable.
> 
> Maybe we should do it differently ? Let's discuss this option

I am close to agree.
That said we have small benefits :

guilpier-code:
> It seems that this PR kind of misses the goal of simplifying the management of printed/hidden variables.
> 
> We introduce 2 vectors (`namesOfEnabledVariables` and `namesOfDisabledVariables`) and a map `index_to_name` that need to be kept synchronized when adding/removing/reverting a variable.
> 
> Maybe we should do it differently ? Let's discuss this option

I am close to agree.
While working on it, I was having the same concern because I saw **variable-print-info.h/.cpp** grow and get more complex.
That said, we have small benefits : 
- on this PR, we reduce globally (a tiny bit) the number of lines.
- yes, the code of **variable-print-info.h/.cpp** has grown because they offer more services to clients of this code, and makes the callers simpler.
- In classes of **variable-print-info.h/.cpp,** old methods (existing before this PR) are globally tinier and more understandable
    - **setPrintStatus**, setMaxColumns**, **clear** and **isPrinted** has shrunk
    - **AllVariablesPrintInfo** has now a default destructor

src/libs/antares/study/parameters.cpp
```diff
@@ -1876,35 +1873,23 @@ void Parameters::saveToINI(IniFile& ini) const
 
     // Variable selection
     {
-        assert(!variablesPrintInfo.isEmpty());
         uint nb_tot_vars = (uint)variablesPrintInfo.size();
-        uint selected_vars = 0;
+        uint nb_selected_vars = (uint)variablesPrintInfo.namesOfEnabledVariables().size();
```
flomnes: You create a `vector` here, but use only its size. I suggest creating a dedicated method with a simple implementation (using e.g `std::count` or `std::count_if`) instead.

flomnes (commented):
That's much better. One last thing and we're good

flomnes (approved):

PR #1184, opened 2023-02-23
Pollutant emissions in thermal clusters

### LINKED ISSUE #1173: Pollutant emissions
### Discussed in https://github.com/AntaresSimulatorTeam/Antares_Simulator/discussions/1132

<div type='discussions-op-text'>

<sup>Originally posted by **flomnes** January 30, 2023</sup>
## General idea

For each thermal cluster, add a factor for
- NH3
- SO2
- NOx
- PM 2.5
- PM 5
- PM 10
- NMVOC
- OP1...OP5

For each **area**, create one output column per pollutant (files values-<period>.txt)

No impact on the cost function, this is only a post-process

`Emission = (Emission factor) * (thermal production)`

## Startup emissions

Add emissions related to unit startups ? The fuel used is specific to startup, which may emit different chemical elements. Data may be an issue.

**No startup emissions**

### Input
Add 12 coefficients to class `ThermalCluster` : one for every pollutant.

Add an `enum` for those **pollutants.**

Add read functions for those coefficients. For each thermal cluster, these coefficients are saved in `input/thermal/clusters/<id of area>/list.ini`, in section \<cluster id\>.

Edit property `ThermalClusterLoadFromProperty` in `src/libs/antares/study/parts/thermal/cluster_list.cpp`. If possible, move this function to class `ThermalCluster`.

### Output

**co2.h** should be replaced  with a new header file **thermalAirPollutantEmissions.h**. 
This file should contain the definition of a class associated with a **multiple** output variable (that is a variable for which **VCard::columnCount** is not **1**) called **ThermalAirPollutantEmissions**.  
 As a model for this class, take **src/solver/variable/economy/renewableGeneration.h** (or even better : **dispatchableGeneration.h**). Its **VCard::columnCount** should be **13**, and it should include **co2**.

https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/ebfb6c7a14bbdbad04014d190b8e0291057bb14c/src/solver/variable/economy/renewableGeneration.h#L99-L128

For each thermal cluster, recover the production at every hour, multiply it by the emission factor. Place this in function `hourForEachThermalCluster`, see **src/solver/variable/economy/dispatchableGeneration.h**.

## Documentation
Update `docs/reference-guide/13-file-format.md` appropriately.

payetvin:
#1173


src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -730,5 +730,253 @@ unsigned int ThermalCluster::precision() const
 {
     return 0;
 }
+
+bool ThermalCluster::LoadFromProperty(ThermalCluster& cluster, const IniFile::Property* p)
```
guilpier-code: As **LoadFromProperty** is a method of **ThermalCluster** class, you don't need the argument **ThermalCluster& cluster**.

But thing is I'm not sure adding this method to **ThermalCluster** class is a good thing : 
**ThermalCluster** is supposed to contain only business data and logic.
Adding a method to read or print data for a cluster goes out of the scope of business logic. 
**ThermalCluster** is already quite big and contains things it shouldn't. 
As examples : methods **markAsModified()** or **memoryUsage()** should not be there.
We should keep **LoadFromProperty** as a free function

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -598,7 +395,7 @@ bool ThermalClusterLoadFromSection(const AnyString& filename,
                                << "`: Invalid key/value";
                 continue;
             }
-            if (not ThermalClusterLoadFromProperty(cluster, property))
+            if (not ThermalCluster::LoadFromProperty(cluster, property))
```
flomnes:
```suggestion
            if (!cluster->loadFromProperty(property))
```

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -814,6 +611,24 @@ bool ThermalClusterList::saveToFolder(const AnyString& folder) const
...
+                s->add("pm5", c.emissionFactors[ThermalCluster::PollutantFactor::PM5]);
+                s->add("pm10", c.emissionFactors[ThermalCluster::PollutantFactor::PM10]);
+                s->add("nmvoc", c.emissionFactors[ThermalCluster::PollutantFactor::NMVOC]);
+                s->add("op1", c.emissionFactors[ThermalCluster::PollutantFactor::OP1]);
+                s->add("op2", c.emissionFactors[ThermalCluster::PollutantFactor::OP2]);
+                s->add("op3", c.emissionFactors[ThermalCluster::PollutantFactor::OP3]);
+                s->add("op4", c.emissionFactors[ThermalCluster::PollutantFactor::OP4]);
+                s->add("op5", c.emissionFactors[ThermalCluster::PollutantFactor::OP5]);
+            }
+
```
flomnes: Move this to cluster.h / cluster.cpp. Or even better, create a dedicated class & .h/.cpp, include it in cluster.h/.cpp

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -730,5 +730,253 @@ unsigned int ThermalCluster::precision() const
...
+{
+    if (p->key.empty())
```
flomnes:
```suggestion
bool ThermalCluster::loadFromProperty(const IniFile::Property* p)
{
    if (p->key.empty())
```

flomnes (changes requested):

guilpier-code (dismissed):

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -730,5 +731,253 @@ unsigned int ThermalCluster::precision() const
 {
     return 0;
 }
+
+bool ThermalCluster::loadFromProperty(ThermalCluster& cluster, const IniFile::Property* p)
```
flomnes: Please introduce a `std::map<std::string, function_ptr>` or something similar.

functionMap[p->property](p->value)

This switch..case is horrible to maintain.

src/libs/antares/study/parts/thermal/pollutant.cpp
```diff
@@ -0,0 +1,57 @@
...
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+
+#include "pollutant.h"
+
+namespace Antares::Data
+{
+
+Pollutant::Pollutant() :
+    emissionFactors()
```
flomnes: No need to call the default ctor, the compiler will do it for us.

src/libs/antares/study/parts/thermal/pollutant.cpp
```diff
@@ -0,0 +1,57 @@
...
+{}
+
+void Pollutant::saveEmissionsFactor(IniFile::Section* s) const
+{
+    if (emissionFactors.empty())
```
flomnes: This array is of fixed size, [so it can't be empty](https://en.cppreference.com/w/cpp/container/array/empty), unless the 2nd template argument is 0.

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,68 @@
...
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#ifndef __ANTARES_LIBS_STUDY_PARTS_THERMAL_POLLUTANT_H__
+#define __ANTARES_LIBS_STUDY_PARTS_THERMAL_POLLUTANT_H__
+
+#include <array>
+
+#include "../../../inifile.h"
```
flomnes: Avoid ../.. in includes, it goes against CMake modules

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Antares::Data
+{
+
+class Pollutant
+{
+public:
+    enum PollutantFactor
```
flomnes: This is not a factor, more of a list of pollutants

src/solver/variable/economy/thermalAirPollutantEmissions.h
```diff
@@ -0,0 +1,326 @@
...
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#ifndef __SOLVER_VARIABLE_ECONOMY_thermalAirPollutantEmissions_H__
+#define __SOLVER_VARIABLE_ECONOMY_thermalAirPollutantEmissions_H__
+
+#include "../variable.h"
+#include "../../../libs/antares/study/parts/thermal/pollutant.h"
```
flomnes, line 31: Move this ?

src/solver/variable/economy/thermalAirPollutantEmissions.h
```diff
@@ -0,0 +1,326 @@
...
+{
+    //! Caption
+    static const char* Caption()
+    {
+        return "Pollu EMIS";
+    }
+    //! Unit
+    static const char* Unit()
+    {
+        return "Tois";
```
flomnes: Tons

src/solver/variable/economy/thermalAirPollutantEmissions.h
```diff
@@ -0,0 +1,326 @@
...
+#include "../../../libs/antares/study/parts/thermal/pollutant.h"
+
+namespace Antares::Solver::Variable::Economy
+{
+struct VCardThermalAirPollutantEmissions
+{
+    //! Caption
+    static const char* Caption()
+    {
+        return "Pollu EMIS";
```
flomnes: The name will be CO2, NH3, etc.

src/solver/variable/economy/thermalAirPollutantEmissions.h
```diff
@@ -0,0 +1,326 @@
...
+        //! File level (provided by the type of the results)
+        categoryFileLevel = ResultsType::categoryFile & (Category::id | Category::va),
+        //! Precision (views)
+        precision = Category::all,
+        //! Indentation (GUI)
+        nodeDepthForGUI = +0,
+        //! Decimal precision
+        decimal = 0,
+        //! Number of columns used by the variable (One ResultsType per column)
+        columnCount = 13,
```
flomnes: POLLUTANT_MAX. So we can add or remove pollutants easily

flomnes (commented):
A very good start !

src/libs/antares/study/parts/thermal/cluster_load.cpp
```diff
@@ -0,0 +1,247 @@
...
+        cb = callbackMap.at(k);
+    } catch(const std::out_of_range& oor) {
+        logs.error() << "Property not found: " << p->key;
+        return false;
+    }
+    return cb(cluster, *p);
+}
+
+
+bool ThermalClusterReader::legacyLoadFromProperty(ThermalCluster& cluster, const IniFile::Property* p)
```
flomnes: I think we'll keep the kind of ugly `if..else` statements. We'll investigate serialization later on, but we don't want to re-invent the wheel.

Please remove the map, place everything in the `legacyLoadFromProperty` function and rename this function.

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -102,6 +102,7 @@ namespace Data
 {
 Data::ThermalCluster::ThermalCluster(Area* parent, uint nbParallelYears) :
  Cluster(parent),
+ emissions(),
```
flomnes: The default constructor is called...by default. So there's no need it to call it manually

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -568,13 +568,18 @@ bool Data::ThermalCluster::integrityCheck()
...
     {
-        logs.error() << "Thermal cluster: " << parentArea->name << '/' << pName
-                     << ": The co2 must be positive or null";
-        co2 = 0;
-        ret = false;
+        if (emissions.emissionFactors[i] < 0)
+        {
+            logs.error() << "Thermal cluster: " << parentArea->name << '/' << pName
+                << ": The " << Pollutant::getPollutantName(i)
+                << " pollutant factor must be positive";
```
flomnes:
```suggestion
                << " pollutant factor must be >= 0";
```

src/libs/antares/study/parts/thermal/cluster_load.h
```diff
@@ -0,0 +1,50 @@
...
+
+ThermalClusterReader();
+
+bool loadFromProperty(ThermalCluster& cluster, const IniFile::Property* p);
+
+std::map<std::string, std::function<bool(ThermalCluster&, const IniFile::Property&)>> callbackMap;
+
+
+//TODO delete
+bool legacyLoadFromProperty(ThermalCluster& cluster, const IniFile::Property* p);
```
flomnes: No, just rename it.

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,73 @@
...
+        OP2,
+        OP3,
+        OP4,
+        OP5,
+        POLLUTANT_MAX
+    };
+
+    static const std::map<enum PollutantList, const std::string> pollutantNames;
+    static const std::string getPollutantName(int index);
+    static const char* getPollutantNameChar(int index);
```
flomnes: Actually, it's not necessary to have 2 functions here. It should work with one single function 
```cpp
static const std::string& getPollutantName(int index);
```

src/libs/antares/study/parts/thermal/pollutant.cpp
```diff
@@ -0,0 +1,96 @@
...
+    {PM2_5, "PM2_5 EMIS."},
+    {PM5, "PM5 EMIS."},
+    {PM10, "PM10 EMIS."},
+    {NMVOC, "NMVOC EMIS."},
+    {OP1, "OP1 EMIS."},
+    {OP2, "OP2 EMIS."},
+    {OP3, "OP3 EMIS."},
+    {OP4, "OP4 EMIS."},
+    {OP5, "OP5 EMIS."},
+    {POLLUTANT_MAX, "POLLUTANT_MAX"}
```
flomnes: Please exclude `POLLUTANT_MAX`, it's not a real element in the enum, just an upper bound.

src/libs/antares/study/parts/thermal/pollutant.cpp
```diff
@@ -0,0 +1,96 @@
...
+    {"op1", OP1},
+    {"op2", OP2},
+    {"op3", OP3},
+    {"op4", OP4},
+    {"op5", OP5}
+};
+
+void Pollutant::saveEmissionsFactor(IniFile::Section* s) const
+{
+    //TODO use for auto : Map to save factors
```
flomnes: Yes, it looks easy to do

src/libs/antares/study/parts/thermal/pollutant.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+const std::string Pollutant::getPollutantName(int index)
+{
+    return pollutantNames.at(PollutantList(index));
+}
+
+const char* Pollutant::getPollutantNameChar(int index)
+{
+    return pollutantNames.at(PollutantList(index)).c_str();
+}
```
flomnes:
```suggestion
const std::string& Pollutant::getPollutantName(int index)
{
    return pollutantNames.at(PollutantList(index));
}
```

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,73 @@
...
+        NMVOC,
+        OP1,
+        OP2,
+        OP3,
+        OP4,
+        OP5,
+        POLLUTANT_MAX
+    };
+
+    static const std::map<enum PollutantList, const std::string> pollutantNames;
```
flomnes:
```suggestion
    static const std::map<enum PollutantList, const std::string> pollutantNamesOutputVariables;
```

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,73 @@
...
+    static const std::string getPollutantName(int index);
+    static const char* getPollutantNameChar(int index);
+
+    static const std::map<std::string, enum PollutantList> namesToEnum;
+
+    std::array<double, POLLUTANT_MAX> emissionFactors;
+
+    void saveEmissionsFactor(IniFile::Section* s) const;
+
```
flomnes: Move this to a different class or function, PollutantIniWriter to separate responsabilities.

src/solver/variable/economy/thermalAirPollutantEmissions.h
```diff
@@ -0,0 +1,296 @@
...
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#ifndef __SOLVER_VARIABLE_ECONOMY_thermalAirPollutantEmissions_H__
+#define __SOLVER_VARIABLE_ECONOMY_thermalAirPollutantEmissions_H__
+
+#include "../variable.h"
+#include "../../../libs/antares/study/parts/thermal/pollutant.h"
```
flomnes:
```suggestion
#include <antares/study.h>
```

flomnes (changes requested):
1. Please remove the callback map, it's only a small improvement. We'll probably use a serialization lib later, so let's not re-invent the wheel.
2. Please keep read/write separate from business classes (Pollutant & ThermalCluster). `ThermalCluster`'s reader/writer will call `Pollutant`'s reader/writer
3. Please remove file **co2.h**

src/libs/antares/study/parts/thermal/cluster.cpp
payetvin: There was a code smell about this so I removed this line and the unfeasible tests failed + 6 shorts tests

flomnes: Care to investigate ?

payetvin:
- [ ] tests
- [ ] doc
- [ ] code smell

src/libs/antares/study/parts/thermal/pollutant.h
payetvin: Needed in caption to avoid c_str() deallocating memory

src/solver/variable/economy/thermalAirPollutantEmissions.h
```diff
@@ -24,25 +24,20 @@
...
-namespace Economy
-{
-struct VCardCO2
-{
-    //! Caption
+    //! Caption not used: several columns
     static const char* Caption()
     {
-        return "CO2 EMIS.";
+        return "";
```
flomnes, line 40: This is used internally, please fill a value 
```suggestion
        return "POLLUTANT EMISS.";
```

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,70 @@
...
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#ifndef __ANTARES_LIBS_STUDY_PARTS_THERMAL_POLLUTANT_H__
+#define __ANTARES_LIBS_STUDY_PARTS_THERMAL_POLLUTANT_H__
+
+#include <array>
+#include <map>
+
+#include <antares/inifile.h>
```
flomnes: Not needed

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,70 @@
...
+        NMVOC,
+        OP1,
+        OP2,
+        OP3,
+        OP4,
+        OP5,
+        POLLUTANT_MAX
+    };
+
+    static const std::map<enum PollutantList, const std::string> pollutantNamesOutputVariables;
```
flomnes: You don't need to declare this object here. Declare it in the .cpp only, as a static object.

src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -153,7 +153,7 @@ static void ExportGridInfosAreas(const Data::Study& study,
             outThermal << cluster.minUpTime << '\t';
             outThermal << cluster.minDownTime << '\t';
             outThermal << cluster.spinning << '\t';
-            outThermal << cluster.co2 << '\t';
+            outThermal << cluster.emissions.emissionFactors[0] << '\t'; //CO2
```
flomnes: Don't use values, use the enum instead
```suggestion
            outThermal << cluster.emissions.emissionFactors[YourEnum::CO2] << '\t'; //CO2
```

flomnes (changes requested):
Rename `emissionFactors` -> `factors` ?

src/libs/antares/study/parts/thermal/pollutant.h
flomnes: Sorry, inifile.h is not needed

src/libs/antares/study/parts/thermal/pollutant.h
payetvin: Can't declare a static object outside of class, can declare it outside but for me it makes more sense to keep it in the class plus it allows to use just CO2 instead of using Pollutant::CO2 inside the map

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -208,7 +209,11 @@ class ThermalCluster final : public Cluster, public std::enable_shared_from_this
 
     bool doWeGenerateTS(bool globalTSgeneration) const;
 
-public:
+    /*!
+    ** \brief Load properties from files
+    */
+    static bool loadFromProperty(ThermalCluster& cluster, const IniFile::Property* p);
+
```
flomnes: Please revert this

src/ui/simulator/toolbox/components/datagrid/renderer/area/thermal.areasummary.cpp
```diff
@@ -107,7 +107,7 @@ wxString ThermalClusterSummarySingleArea::cellValue(int x, int y) const
     case 8:
         return DoubleToWxString(cluster->spinning);
     case 9:
-        return DoubleToWxString(cluster->co2);
+        return DoubleToWxString(cluster->emissions.factors[0]); //emissionFactor[0] = CO2
```
flomnes: Use enum

src/ui/simulator/toolbox/components/datagrid/renderer/area/thermal.areasummary.cpp
```diff
@@ -150,7 +150,7 @@ double ThermalClusterSummarySingleArea::cellNumericValue(int x, int y) const
     case 8:
         return cluster->spinning;
     case 9:
-        return cluster->co2;
+        return cluster->emissions.factors[0]; //emissionFactor[0] = CO2
```
flomnes: Use enum

src/ui/simulator/toolbox/components/datagrid/renderer/area/thermal.areasummary.cpp
```diff
@@ -204,7 +204,7 @@ bool ThermalClusterSummarySingleArea::cellValue(int x, int y, const String& v)
         case 8:
             return Update<double, NoCheck, RefeshInspector>(cluster->spinning, v);
         case 9:
-            return Update<double, NoCheck, RefeshInspector>(cluster->co2, v);
+            return Update<double, NoCheck, RefeshInspector>(cluster->emissions.factors[0], v); //emissionFactor[0] = CO2
```
flomnes: Use enum

src/ui/simulator/windows/inspector/accumulator.hxx
```diff
@@ -768,7 +768,7 @@ struct PClusterCO2
     using Type = double;
     static Type Value(const Data::ThermalCluster* cluster)
     {
-        return cluster->co2;
+        return cluster->emissions.factors[0]; //cluster->co2
```
flomnes: Use enum

src/ui/simulator/windows/inspector/property.update.cpp
```diff
@@ -705,13 +705,13 @@ bool InspectorGrid::onPropertyChanging_ThermalCluster(wxPGProperty*,
         if (d < 0.)
         {
             for (; i != end; ++i)
-                (*i)->co2 = 0.;
+                (*i)->emissions.factors[0] = 0.;
```
flomnes: Use enum

src/ui/simulator/windows/inspector/property.update.cpp
```diff
@@ -705,13 +705,13 @@ bool InspectorGrid::onPropertyChanging_ThermalCluster(wxPGProperty*,
...
             pFrame.delayApply();
         }
         else
         {
             for (; i != end; ++i)
-                (*i)->co2 = d;
+                (*i)->emissions.factors[0] = d;
```
flomnes: Use enum

flomnes (changes requested):

src/libs/antares/study/parts/thermal/pollutant.h
```diff
@@ -0,0 +1,68 @@
...
+    static const std::string& getPollutantName(int index);
+
+    static const std::map<std::string, enum PollutantList> namesToEnum;
```
a-zakir: missing  `#include <string>  `?

flomnes (approved):

PR #1186, opened 2023-02-23
Redirect OR-Tools's logs to antares-solver's logs, add support for `std::string_view` in logs

### LINKED ISSUE #1169: Redirect OR-Tools log messages to our logs instead of writing them to `stderr`
## Idea

OR-Tools uses [GLOG](https://github.com/google/glog). From the doc, it may be possible to setup a [`LogSink`](https://github.com/google/glog/blob/f545ff5e7d7f3df95f6e86c8cb987d9d9d4bd481/src/glog/logging.h.in#L1754) to redirect the output into our own logging system.

Currently, all of OR-Tools log messages are written to `stderr`.

## Possible implementation
```cpp
#include <glog/logging.h>

class AntaresSink : public google::base_logging::LogSink {
  void send(LogSeverity severity, const char* full_filename,
            const char* base_filename, int line,
            const LogMessageTime& logmsgtime, const char* message,
            size_t message_len) override {
    switch(severity) {
    case GLOG_INFO:
      logs.info() << message; // check that message is '\0' terminated, it may not be the case
      break;
      // GLOG_WARNING, etc.
    }
  }
};

// At initialization
AntaresSink sink;
InitGoogleLogging(argv[0]);
AddLogSink(sink);

```

flomnes:


flomnes:
OR-Tools's logging interface changed to Absl log, instead of glog in v9.6. This change does not appear in the changelog.

PR #1187, opened 2023-02-23
Add adequacy patch optimization number to MPS filename to avoid overwrite

### LINKED ISSUE #1181: Don't overwrite MPS files when adq-patch is enabled
We need 4 MPS files per week if the adq patch is enabled. Currently, only 0-2 are produced since some files are overwritten.

flomnes:


src/solver/optimisation/adequacy_patch_local_matching/adq_patch_local_matching.cpp
```diff
@@ -136,7 +136,8 @@ void setNTCbounds(double& Xmax,
     Xmin = -(ValeursDeNTC->ValeurDeNTCExtremiteVersOrigine[Interco]);
 
     // set for adq patch first step
-    if (problemeHebdo->adqPatchParams && problemeHebdo->adqPatchParams->AdequacyFirstStep)
+    if (problemeHebdo->adqPatchParams
+        && problemeHebdo->adequacyPatchRuntimeData->LMR_FirstOptimization)
```
flomnes, line 140:
```suggestion
    if (problemeHebdo->adequacyPatchRuntimeData
        && problemeHebdo->adequacyPatchRuntimeData->LMR_FirstOptimization)
```

src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
```diff
@@ -45,9 +45,9 @@ AdequacyPatchOptimization::AdequacyPatchOptimization(PROBLEME_HEBDO* problemeHeb
 }
 void AdequacyPatchOptimization::solve(uint weekInTheYear, int hourInTheYear)
 {
-    problemeHebdo_->adqPatchParams->AdequacyFirstStep = true;
+    problemeHebdo_->adequacyPatchRuntimeData->LMR_FirstOptimization = true;
```
guilpier-code, line 48: ok.
Should conflict with #1189

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -234,8 +237,17 @@ bool OPT_AppelDuSimplexe(PROBLEME_HEBDO* problemeHebdo,
     {
         solver = ORTOOLS_ConvertIfNeeded(&Probleme, solver);
     }
-    const std::string filename = createMPSfilename(optPeriodStringGenerator, optimizationNumber);
-    mpsWriterFactory mps_writer_factory(problemeHebdo->ExportMPS, problemeHebdo->exportMPSOnError, optimizationNumber, &Probleme, ortoolsUsed, solver);
+
+    auto adqPatchOptNumber
```
guilpier-code, line 241: Following the oral talk we just had, I think that the MPS filename should **not** be computed here, but at a higher in the call stack.
Here this name should be passed as an argument.
See [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1187#discussion_r1121932212)

src/solver/optimisation/opt_optimisation_lineaire.cpp
```diff
@@ -110,16 +111,20 @@ bool OPT_OptimisationLineaire(PROBLEME_HEBDO* problemeHebdo, uint numSpace)
                                     problemeHebdo->year);
 
         if (!OPT_AppelDuSimplexe(
-              problemeHebdo, NumeroDeLIntervalle, optimizationNumber, optPeriodStringGenerator))
+              problemeHebdo, NumeroDeLIntervalle, optimizationNumber, *optPeriodStringGenerator))
```
guilpier-code, line 114: Here we should have (something like) : 
```C++
auto mpsFileNames = MPSfileNames(problemeHebdo->OptimisationAuPasHebdomadaire,
                                 NumeroDeLIntervalle,
                                 problemeHebdo->adequacyPatchRuntimeData
                                 problemeHebdo->weekInTheYear,
                                 problemeHebdo->year)

if (!OPT_AppelDuSimplexe(problemeHebdo, 
                         NumeroDeLIntervalle, 
                         optimizationNumber, 
                         mpsFileNames->mps()))
        return false;

if (problemeHebdo->ExportMPS ......)
{
    ...
    OPT_EcrireResultatFonctionObjectiveAuFormatTXT(optimalSolutionCost,
                                                   mpsFileNames->criterion());
}
```
This would allow : 
- extracting filenames creation from functions that print into these files
- calling  **getAdqPatchOptmizationNumber(...)** only once.

guilpier-code (changes requested):

src/solver/optimisation/opt_period_string_generator_base.h
```diff
@@ -37,4 +38,8 @@ class OptPeriodStringGenerator
 public:
     virtual std::string to_string() const = 0;
     virtual ~OptPeriodStringGenerator() = default;
+    std::string createOptimizationFilename(const std::string& title,
```
guilpier-code, line 41: Base class **OptPeriodStringGenerator** is an abstraction for the optimization period, that is either :  
- \<year\>-\<week\>
- \<year\>-\<week\>-\<day\>

inside the filename.
So why do we have a new **createOptimizationFilename** function that take many arguments about the whole file name ?
I guess **OptPeriodStringGenerator** should concern the optimization period only.

guilpier-code (changes requested):

PR #1190, opened 2023-02-28
DO NOT MERGE - FOR EXECUTABLES ONLY RTEi

Milos-RTEi:
This PR is created to generate executables for the beta-12 version of adequacy patch for Elia.

From Daniel's Email:

What we need now is a beta 12 version as follows
-        We take  Antares v8.5.0 (released)
-        We bypass the isolated simulation
-        We thus do NOT impose the ENS < DENS_0 constraint in the interconnected simulation
-        We just run Antares “without any Adq Patch” in the first place anyway. Then we calculated DENS_new as we do now in beta 11 but just on the results without any such constraint ENS < DENS_0
-        We follow the same LMR+CSR procedure as now in beta11
- 
Hence on paper it seems rather feasible to provide us a “beta 12 version on Antares v8.5.0”  quite fast since the steps to take are just*
-        Deactivate & bypass the isolated step
-        Deactivate the ENS < DENS_0 constraint
-        The rest remains the same

### DIFF
--- a/src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
+++ b/src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
@@ -49,7 +49,7 @@ double calculateQuadraticCost(const PROBLEME_HEBDO* problemeHebdo, int hour, int
     else if (problemeHebdo->adqPatchParams->PriceTakingOrder
              == Data::AdequacyPatch::AdqPatchPTO::isDens)
     {
-        priceTakingOrders = problemeHebdo->ResultatsHoraires[area]->ValeursHorairesDENS[hour];
+        priceTakingOrders = problemeHebdo->ResultatsHoraires[area]->ValeursHorairesDENS[hour]; // densNew value taken! OK!
     }
 
     if (priceTakingOrders <= 0.0)
--- a/src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
+++ b/src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
@@ -45,27 +45,33 @@ AdequacyPatchOptimization::AdequacyPatchOptimization(PROBLEME_HEBDO* problemeHeb
 }
 void AdequacyPatchOptimization::solve(uint weekInTheYear, int hourInTheYear)
 {
-    problemeHebdo_->adqPatchParams->AdequacyFirstStep = true;
-    OPT_OptimisationHebdomadaire(problemeHebdo_, thread_number_);
+    // problemeHebdo_->adqPatchParams->AdequacyFirstStep = true;
+    // OPT_OptimisationHebdomadaire(problemeHebdo_, thread_number_);
+
+    // never set AdequacyFirstStep to True and skip the islanding optimization
     problemeHebdo_->adqPatchParams->AdequacyFirstStep = false;
 
     for (int pays = 0; pays < problemeHebdo_->NombreDePays; ++pays)
     {
-        if (problemeHebdo_->adequacyPatchRuntimeData->areaMode[pays]
-            == Data::AdequacyPatch::physicalAreaInsideAdqPatch)
-            memcpy(problemeHebdo_->ResultatsHoraires[pays]->ValeursHorairesDENS,
-                   problemeHebdo_->ResultatsHoraires[pays]->ValeursHorairesDeDefaillancePositive,
-                   problemeHebdo_->NombreDePasDeTemps * sizeof(double));
-        else
+        // if (problemeHebdo_->adequacyPatchRuntimeData->areaMode[pays]
+        //     == Data::AdequacyPatch::physicalAreaInsideAdqPatch)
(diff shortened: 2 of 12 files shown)


Milos-RTEi:
@hugo-antoine-rtei 
ok. The pipeline is running ok. That is great. 
Since I'm going to do this really fast. Please try to review the logic (I'll do the testing and the dev). 
So at least I have the second pair of eyes take a look at the code logic. 
To try to avoid some really stupid mistakes.

Milos-RTEi:
I'll remove you as the reviewer now. Not to fill up your outlook with stupid mails. Will let you know when I finish.

Milos-RTEi:
Hi, @hugo-antoine-rtei @flomnes 

please take a look (Florian volunteered himself).
I hope there is no major flaw in my thinking.

I did lots of testing. 

- First. If you run the simulation without the adequacy patch and then run it with the adequacy patch but with CSR disabled, 
You get identical results, As expected.
- Second. I logged out again the complete formulation (input data) for the CSR solver. And all looks ok. 
- Third. I do not have any issues when running simulations with beta-12 on Daniel's networks. 

SonarCloud is going crazy cause of the unused variables, functions.. commented-out lines...
But I don't want to worry about code quality now. If we get thumbs up for beta 12 then we do the cleanup later.

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
```diff
@@ -49,7 +49,7 @@ double calculateQuadraticCost(const PROBLEME_HEBDO* problemeHebdo, int hour, int
     else if (problemeHebdo->adqPatchParams->PriceTakingOrder
              == Data::AdequacyPatch::AdqPatchPTO::isDens)
     {
-        priceTakingOrders = problemeHebdo->ResultatsHoraires[area]->ValeursHorairesDENS[hour];
+        priceTakingOrders = problemeHebdo->ResultatsHoraires[area]->ValeursHorairesDENS[hour]; // densNew value taken! OK!
```
flomnes, line 52: Careful, `ValeursHorairesDENS[hour] = 0`, see **src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp**.

```suggestion
        priceTakingOrders = problemeHebdo->ResultatsHoraires[pays]->ValeursHorairesDeDefaillancePositive[hour];
```

src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
```diff
@@ -45,27 +45,33 @@ AdequacyPatchOptimization::AdequacyPatchOptimization(PROBLEME_HEBDO* problemeHeb
 }
 void AdequacyPatchOptimization::solve(uint weekInTheYear, int hourInTheYear)
 {
-    problemeHebdo_->adqPatchParams->AdequacyFirstStep = true;
-    OPT_OptimisationHebdomadaire(problemeHebdo_, thread_number_);
+    // problemeHebdo_->adqPatchParams->AdequacyFirstStep = true;
+    // OPT_OptimisationHebdomadaire(problemeHebdo_, thread_number_);
+
+    // never set AdequacyFirstStep to True and skip the islanding optimization
```
flomnes, line 51: You don't need to change the definition of this class (revert all changes).

We made a nice factory method just for that, see file **src/solver/optimisation/base_weekly_optimization.cpp**. Just return `std::make_unique<DefaultWeeklyOptimization>` in all cases (adq patch enabled/disabled), it's much simpler.

flomnes (changes requested):

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
Milos-RTEi: I'm positive we are good here. According to spec:
image
s0 PTO=DENS not PTO=ENS.
if ValeursHorairesDENS[hour] = 0 function calculateQuadraticCost will just return zero. It will not try to divide with zero.
If you meant that by this point in the code the column DENS is going to be all zeros, It is not the case. Zeros are overwritten for areas inside adq-patch in 
src/solver/optimisation/post_process_commands.cpp line 231

src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
Milos-RTEi: Yes, I saw that. Nice. I agree it's a proper way to do it. 
I just wanted to be sure that AdequacyFirstStep = false. Even though I think I overwritten all the code where AdequacyFirstStep is used.
Also for areas outside adq-patch we need to set zero values for DENS somewhere. 
I agree DENS should be initialized to zero somewhere else if just defaultWeeklyOptimization is required. 
On TODO list.

src/solver/optimisation/adequacy_patch_csr/set_problem_cost_function.cpp
flomnes: Ok, LGTM

flomnes:
We will add a new option to disable the LMR, probably in 8.6.

I'll be closing this temporary solution.

PR #1201, opened 2023-03-07
Optimization for short term storage

flomnes:
TODO
- [X] Build variables
- [X] Bounds & address for variables (results)
- [X] Change balance constraint for each area
- [x] Allocate & de-allocate correspondence arrays
- [x] Add level constraints (matrix & RHS)
- [x] Initial level = input value
- [x] Initial level = final level


src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -83,6 +86,7 @@ typedef struct
 
     int* NumeroDeContrainteDesNiveauxPays;
 
+    int* ShortTermStorageLevelConstraint;
 } CORRESPONDANCES_DES_CONTRAINTES;
 
 typedef struct
```
guilpier-code, line 96: **SINGLE_OBJECT_INPUT** is unclear to me. 
I know it's inside a namespace more meaningful, but still.
Could we rename it **ST_STORAGE_PROPERTIES** by any chance ?

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -83,6 +86,7 @@ typedef struct
 
     int* NumeroDeContrainteDesNiveauxPays;
 
+    int* ShortTermStorageLevelConstraint;
```
guilpier-code, line 93:
```C++
struct AREA_INPUT
{
    std::vector<SINGLE_OBJECT_INPUT> storages; // index is local
};
```
If we don't put something else than a vector in the **struct**, do we need a **struct** ?

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -64,6 +64,9 @@ typedef struct
     int* NumeroDeVariableDuNombreDeGroupesQuiSArretentDuPalierThermique;
     int* NumeroDeVariableDuNombreDeGroupesQuiTombentEnPanneDuPalierThermique;
 
+    int* ShortTermStorageInjectionVariable;
```
guilpier-code: It looks like too many things are named **ShortTermStorage** : a name space, some results, a vector for optimization variables. For accustomed eyes, it may not be much of a problem, but for newcomers it is confusing. 
We should make these names more distinct.
For example (but choose whatever you like) : 
- STstorageNames
- STstorageOutput
- STstorageOptVariables

src/solver/optimisation/opt_construction_variables_optimisees_lineaire.cpp
```diff
@@ -88,6 +88,29 @@ void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeLineaire(PROBLEME_HEBD
                 NombreDeVariables++;
             }
 
+            for (const auto& storage : (*problemeHebdo->ShortTermStorage)[pays].storages)
```
guilpier-code: As you see, function **OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(...)** and the source file it belongs to are growing.
A possible quite simple solution for them to stop growing (to be discussed) would be to have : 
```C++
void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(PROBLEME_HEBDO* problemeHebdo, ...)
{
      problemeHebdo->hydroModeller.setBounds();
      problemeHebdo->thermalModeller.setBounds();
      ...
      // And of course
      problemeHebdo->shortTermStorageModeller.setBounds();
      ...
}
```
Several remarks about this :
- This solution should be discussed to determine if it is suitable or feasible, it may not be perfect at all, but at least it would be better than what we have now : the current **OPT** code forces us to make functions grow and do much more things than they should.
- It intends to extract bound management in a more specialized class.
- For this evolution, only the `problemeHebdo->shortTermStorageModeller.setBounds();` part would be taken care of
- This solution would be applicable to any other functions in the **OPT** part
- It would be applicable to any other production asset taking a part in the optimization 
- Class **ShortTermstorageModeller** would have a constructor taking as argument  a PROBLEME_HEBDO and characteristics on short term storage.
- Doing this for every production asset and every **OPT** function will lead to make the **PROBLEME_HEBDO** useless as an argument of these **OPT** functions and should simplify a lot. But is this solution doable ?

guilpier-code (changes requested):

src/solver/optimisation/opt_construction_variables_optimisees_lineaire.cpp
flomnes: I totally agree with this remark, can you open a refactor issue on this ?

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: good question ! I can remove it

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: Good idea, this is confusing !

src/solver/optimisation/opt_construction_variables_optimisees_lineaire.cpp
guilpier-code: Yes I can. I will.
Honestly, what I have in mind, eventually, is something like (you probably guessed already) : 
 ```C++
 // Only when setting bounds here, but can (hopefully) be applied to any optimization preparation stage  
 for (auto* asset : list_of_assets)
     asset->setBounds();
```

**Comment by guilpier-code** [comment]
About source file **parts/short-term-storage/unit.h** : 
For thermal clusters, a unit represent a plant, right ? And a cluster is a set of units.
For ST storage, a **Unit** represents a cluster of possibly many storage, but this cluster is meant to contain only **one** storage for now (it might change eventually, or not).
So if one have in mind the notions of thermal cluster and thermal unit, transposing them to ST storage does not work and is confusing.
We should stick to the specification and be able to make a parallel between thermal and ST storage worlds, so in the context of ST storage, **Unit** should be renamed **STStorageCluster** (for instance).

File: src/solver/simulation/sim_structure_probleme_economique.h (hunk 2 above)
Comment (flomnes) [reply src/solver/simulation/sim_structure_probleme_economique.h:93]: done, removed

File: src/solver/simulation/sim_structure_probleme_economique.h (hunk 1 above)
Comment (flomnes) [reply src/solver/simulation/sim_structure_probleme_economique.h:96]: This struct is used only a few lines after. I'll rename it PROPERTIES

File: src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp [hunk 5]
```diff
@@ -133,6 +133,15 @@ void OPT_InitialiserLeSecondMembreDuProblemeLineaire(PROBLEME_HEBDO* problemeHeb
             AdresseOuPlacerLaValeurDesCoutsMarginaux[cnt] = nullptr;
         }
 
+        for (int shortTermStorage = 0; shortTermStorage < problemeHebdo->NumberOfShortTermStorage;
+             shortTermStorage++)
+        {
+            if (int cnt = CorrespondanceCntNativesCntOptim
+                            ->ShortTermStorageLevelConstraint[shortTermStorage];
+                cnt >= 0)
+                SecondMembre[cnt] = 0; // TODO[FOM] inflows[h]
```
payetvin, line 142: Remove TODO ?

payetvin (approved):

PR #1202, opened 2023-03-07
Short term storage loading ini files

payetvin:
branched from #1163

### DIFF
--- a/src/libs/antares/study/parts/short-term-storage/container.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/container.cpp
@@ -25,23 +25,54 @@
 ** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
 */
 
+
 #include "container.h"
+
+#define SEP IO::Separator
 
 namespace Antares::Data::ShortTermStorage
 {
+
+
+
 bool Container::validate() const
 {
     // TODO
     return false;
 }
+
 bool Container::createUnitsFromIniFile(const std::string& path)
 {
-    // TODO
-    return false;
+    const std::string pathIni = path +  SEP + "list.ini";
+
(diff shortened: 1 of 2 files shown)


src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -25,23 +25,54 @@
...
-    return false;
+    const std::string pathIni = path +  SEP + "list.ini";
+
+    IniFile ini;
+    if (!ini.open(path))
+        return false;
+
+    for (auto* section = ini.firstSection; section; section = section->next)
+    {
+        if (!ContainerLoadFromSection(pathIni, *section))
```
flomnes, line 54: This method is missing ?

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -25,23 +25,54 @@
...
+
+bool Container::loadFromSection(const std::string& pathIni, const IniFile::Section& section)
+{
+    if (section.name.empty())
+        return false;
+
+    Unit unit;
+    unit.loadFromFolder(pathIni);
+    storages.insert({ section.name, unit });
+    return true;
```
flomnes, line 74: Multiple units may be present, so a loop of some kind will be required here

src/libs/antares/study/parts/short-term-storage/container.h
```diff
@@ -28,6 +28,9 @@
 #pragma once
 #include <string>
 #include <map>
+
+#include <antares/study.h>
+
```
flomnes, line 33: Why do you need this header ?

payetvin:
Bad start, switched to a new branch and PR

PR #1203, opened 2023-03-08
Minimal hydro generation (RTEi's -> CR22)

Milos-RTEi:
This PR is created in order to implement RTEi development on CR22 (Minimal Hydro Generation).
Spec:
https://rteinternational.sharepoint.com/:w:/r/sites/AntaresUser%27sClub-Interne/Documents%20partages/Change%20Requests/Selected%20for%20development/CR22%20-%20Multiple%20Minimum%20Generation%20for%20Hydro%20Reservoirs/Functional_Specifications-Multiple_min_gen_for_hydro_clusters_V0.1-HA.docx?d=web3abdf476d94fc8adca58d741bffb94&csf=1&web=1&e=wfbOie

The objective is to allow the user to specify several hourly minimum generating power time-series for a given reservoir. To build a given Monte-Carlo year, Antares would then select one of these time-series (either randomly or via a proper setting of the scenario builder). This parameter is applicable whether the reservoir is managed via the heuristic or with water values.


src/solver/hydro/management/management.cpp
```diff
@@ -135,6 +137,89 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
...
+                logs.error() << "In Area "<< area.name << " the minimum generation of "
+                << totalYearMingen << " MW of TS-" << tsIndex + 1 << " is incompatible with the inflows of "
+                << totalYearInflows << " MW.";
+            }
+        }
+
+        if (not area.hydro.followLoadModulations)
+        {
+            //CR22: Weekly minimum generation <= Weekly inflows for each week 
+            for (uint week = 0; week < 53; ++week)
```
flomnes: 52 ?

src/solver/hydro/management/management.cpp
```diff
@@ -166,7 +251,7 @@ void HydroManagement::prepareNetDemand(uint numSpace)
             {
                 netdemand = +scratchpad.ts.load[ptchro.Consommation][hour]
                             - scratchpad.ts.wind[ptchro.Eolien][hour] - scratchpad.miscGenSum[hour]
-                            - scratchpad.ts.solar[ptchro.Solar][hour] - ror[hour]
+                            - scratchpad.ts.solar[ptchro.Solar][hour] - ror[hour] //CR22 todo add mingen here or not?
```
flomnes, line 354: Ask PP & JMJ

flomnes (commented):
Check the scenario number is taken into account for all components

flomnes:
Automated tests ?

src/solver/hydro/management/management.cpp
Milos-RTEi: 52 weeks x 168 h/week = 8736 h
we have HOURS_PER_YEAR = 8760 entries.
If we go just up to 52 weeks we "forget" the last 8760-8736=24h
so it's 53 weeks but with the additional condition " if (hour < HOURS_PER_YEAR)" not to go above 8760 and sum some non-existing values.

docs/reference-guide/13-file-format.md
```diff
@@ -1,5 +1,12 @@
 # Study format changes
 This is a list of all recent changes that came with new Antares Simulator features. The main goal of this document is to lower the costs of changing existing interfaces, both GUI and scripts.
+## v8.6.0
+### Input
+For each area, under existing `Hydro` section, new tab added `Minimum Generation`.\
+For each area, under existing `Hydro` section, existing tab `Run-of-river (ROR)` changed name to `Run-of-river (ROR)-HydroFatal`.\
+For each area, under existing `Hydro` section, existing tab `Hydro Storage` changed name to `Hydro Storage-HydroMod`.
```
flomnes: This is a UI change right ? If so, it doesn't belong in this documentation (file format)

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
```diff
@@ -287,6 +287,13 @@ void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(PROBLEME_HEBDO* prob
             if (var >= 0)
             {
                 Xmin[var] = 0.0;
+                bool presenceHydro = problemeHebdo->CaracteristiquesHydrauliques[pays]
+                                       ->PresenceDHydrauliqueModulable;
```
flomnes: Why only in this case ?

flomnes:
> This PR is created in order to implement RTEi development on CR22 (Minimal Hydro Generation). Spec:https://rteinternational.sharepoint.com/:w:/s/AntaresUser'sClub-Interne/EfS9OuvZdshPrcpY10G_-5QBTxdJ2XAoFnN3E1l6E_gj5A?e=bWQb7h
> 
> The objective is to allow the user to specify several hourly minimum generating power time-series for a given reservoir. To build a given Monte-Carlo year, Antares would then select one of these time-series (either randomly or via a proper setting of the scenario builder). This parameter is applicable whether the reservoir is managed via the heuristic or with water values.

Can you please post a public link for the spec ?

flomnes (commented):
Why not use ROR ?

src/solver/hydro/management/management.cpp
```diff
@@ -135,6 +137,89 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
...
+            for (uint d = firstDayOfMonth; d != firstDayOfNextMonth; ++d)
+            {
+                for (uint h = 0; h < 24; ++h)
+                {
+                    totalMonthMingen += srcmingen[d*24 + h];
+                }          
+            }
+
+            //CR22: set montly mingen, used later for h2o_m
+            if (not(area.hydro.reservoirCapacity < 1e-4))
```
guilpier-code: Why not take advantage of the `if (not(area.hydro.reservoirCapacity < 1e-4))` already existing above for inflows ? See line 126.

src/solver/hydro/management/management.cpp
```diff
@@ -135,6 +137,89 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
...
+                }
+            }
+        }
+
+        if (area.hydro.followLoadModulations and area.hydro.reservoirManagement)
+        {
+            //CR22: Yearly minimum generation <= Yearly inflows for each year 
+            double totalYearMingen = 0.0;
+            double totalYearInflows = 0.0;
+            for (uint hour = 0; hour < HOURS_PER_YEAR; ++hour)
```
guilpier-code: The monthly inflows and min generation have already been computed, there must be a way to reuse them and make a loop over the months.
And to use **std::accumulate**.

src/solver/hydro/management/management.cpp
guilpier-code: > 52 weeks x 168 h/week = 8736 h we have HOURS_PER_YEAR = 8760 entries. If we go just up to 52 weeks we "forget" the last 8760-8736=24h so it's 53 weeks but with the additional condition " if (hour < HOURS_PER_YEAR)" not to go above 8760 and sum some non-existing values.

I see your point, but the use of **53** and the `if (hour < HOURS_PER_YEAR)` are not optimal at all.
I'll try to find a better way to do that.

guilpier-code (changes requested):

src/solver/hydro/management/management.cpp
```diff
@@ -135,6 +137,89 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
             {
                 data.inflows[realmonth] = totalMonthInflows;
             }
+
+            //CR22: Monthly minimum generation <= Monthly inflows for each month
+            double totalMonthMingen = 0.0;
```
guilpier-code: **srcmingen** is probably an array of doubles.
```C++
double totalMonthMingen = std::accumulate(srcmingen + firstDayOfMonth * 24, 
                                          srcmingen + firstDayOfNextMonth * 24 - 1, 
                                          0.);
``` 
should work.

src/solver/hydro/management/management.cpp
guilpier-code: I made some digging on this.

_A bit of context on the calendar_ :
The reason why we may have 366 days (and therefore 8760 hours) in a year of the calendar is that we may run a simulation on a leap year (see **GUI > Simulation > Calendar > Leap year**).
So we can have a simulation of many MC leap years.
But the simulation will always run on 52 weeks maximum, this means there is a day for which we don't need the input data (data related to this day in the time series), even if we load the whole TS.

That said, in the hydro heuristic like here, you have to deal with the whole calendar 
(whatever the duration of a simulation and whether it runs on a leap year or not), right ?

- You're right to loop over 53 weeks. But I would recommend to use `calendar.maxWeeksInYear`. For some reason I don't know, `calendar.maxWeeksInYear` = 54. So, instead of 53, you may use `calendar.maxWeeksInYear - 1`
- In the calendar, the last week contains only one day, that is 24 hours, so **hour** (index of hours) will not exceed **HOURS_PER_YEAR** in the last week.

To sum up, I recommend : 
```C++
for(uint hour = calendar.weeks[week].hours.first; hour < calendar.weeks[week].hours.end; ++hour)
 {
     totalWeekMingen += srcmingen[hour];
 }
```
An you could also make use of **std::accumulate** instead
```C++
double totalWeekMingen = 0.0;
totalWeekMingen = std::accumulate(srcmingen + calendar.weeks[week].hours.first,
                                  srcmingen + calendar.weeks[week].hours.end,
                                  totalWeekMingen);
```

guilpier-code (changes requested):

src/solver/hydro/management/management.cpp
guilpier-code: I made some digging on this.

_A bit of context on the calendar_ :
The reason why we may have 366 days (and therefore 8760 hours) in a year of the calendar is that we may run a simulation on a leap year (see **GUI > Simulation > Calendar > Leap year**).
So we can have a simulation of many MC leap years.
But the simulation will always run on 52 weeks maximum, this means there is a day for which we don't need the input data (data related to this day in the time series), even if we load the whole TS.

That said, in the hydro heuristic like here, you have to deal with the whole calendar 
(whatever the duration of a simulation and whether it runs on a leap year or not), right ?

- You're right to loop over 53 weeks. But I would recommend to use `calendar.maxWeeksInYear`. For some reason I don't know, `calendar.maxWeeksInYear` = 54. So, instead of 53, you may use `calendar.maxWeeksInYear - 1`
- In the calendar, the last week contains only one day, that is 24 hours, so **hour** (index of hours) will not exceed **HOURS_PER_YEAR** in the last week.

To sum up, I recommend : 
```C++
for(uint hour = calendar.weeks[week].hours.first; hour < calendar.weeks[week].hours.end; ++hour)
 {
     totalWeekMingen += srcmingen[hour];
 }
```
An you could also make use of **std::accumulate** instead
```C++
double totalWeekMingen = 0.0;
totalWeekMingen = std::accumulate(srcmingen + calendar.weeks[week].hours.first,
                                  srcmingen + calendar.weeks[week].hours.end,
                                  totalWeekMingen);
```

src/solver/hydro/management/management.cpp
Milos-RTEi: Already implemented in commit [25cd0d3]. Please pull. And it is actually up to "srcmingen + firstDayOfNextMonth * 24 " without "- 1". I tested it.

src/solver/hydro/management/management.cpp
Milos-RTEi: I agree with "calendar.maxWeeksInYear - 1". Will be implemented.
For the second part. In theory, this should work. We implemented it like this the first time. 
But during testing this happens:
image

And you end up with dummy errors:
image

So what happens? The hourly iterations continue after 8760. And to make things worse starts to sum up some crazy values for:
srcinflows[hour]
even though "srcinflows" should not have more than 8760 rows allocated. 
Actually, I realized these crazy values are numbers from week-0, hours-0-1-2..... 
It basically makes a full circle and starts summing from the start again and putting that in week 52. 
So:
if (hour < HOURS_PER_YEAR)
is kinda unavoidable here.

src/solver/hydro/management/management.cpp
```diff
@@ -102,17 +103,22 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
         auto& ptchro = *NumeroChroniquesTireesParPays[numSpace][z];
 
         auto& inflowsmatrix = area.hydro.series->storage;
+        auto& mingenmatrix = area.hydro.series->mingen;
```
guilpier-code: The current function is named **prepareInflowsScaling**, so it's supposed to make tasks about inflows only.
But there are also minimum hydro generation tasks done.
We could rename this function so that its body strictly adheres to what it does, but for reasons of readability and isolation, we should split it into (at least) 2 functions.
We could even split the min generation task into **minGenerationScaling** and **checkMinGeneration**, I think that it would be better.

guilpier-code (changes requested):

src/solver/hydro/management/daily.cpp
```diff
@@ -378,6 +381,22 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
             for (uint day = firstDay; day != endDay; ++day)
             {
                 problem.TurbineMax[dayMonth] = maxP[day] * maxE[day];
+                double dailyMingen = 0.0;
```
guilpier-code: This a check on the daily minimum generation.
This check should not be done in a function that builds and runs a optimization problem, because it makes the function grow and be less readable (it's already too big and makes already too many things). 
This check does not need to done in the context of this function, it should be moved somewhere else (see  [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1203#discussion_r1156980396) about that, it gives a possible place to move the check).

Furthermore, we should use **std::accumulate** to compute **dailyMingen**.

I can see that the computation of **dailyMingen** is done twice in this source file, this is code duplication. This computation should be done once in **management.cpp** (follow the link above), and its result stored, so that it can be used here.

You could reply that there are many code duplications in this function, and you would be right.
No need to add some more.

guilpier-code (changes requested):

src/solver/hydro/daily/h2o_j_initialiser_les_bornes_des_variables.c
```diff
@@ -72,12 +74,16 @@ void H2O_J_InitialiserLesBornesdesVariables(DONNEES_MENSUELLES* DonneesMensuelle
...
+          if (TurbineCible[Pdt] > TurbineMin[Pdt])
             Xmin[Var] = TurbineCible[Pdt];
+          else
+            Xmin[Var] = TurbineMin[Pdt];
         }
         else
         {
             Xmin[Var] = TurbineMax[Pdt];
-        }
+        }
```
guilpier-code: Instead of :
```C++
if (TurbineMax[Pdt] > TurbineCible[Pdt])
{
    if (TurbineCible[Pdt] > TurbineMin[Pdt])
        Xmin[Var] = TurbineCible[Pdt];
    else
        Xmin[Var] = TurbineMin[Pdt];
}
else
{
       Xmin[Var] = TurbineMax[Pdt];
}        
```
we could have : 
```C++
Xmin[Var] = std::min(TurbineMax[Pdt], std::max(TurbineCible[Pdt], TurbineMin[Pdt]));
```

guilpier-code (changes requested):

src/solver/hydro/daily/h2o_j_lisser_les_sur_turbines.c
```diff
@@ -55,6 +56,7 @@ void H2O_J_LisserLesSurTurbines(DONNEES_MENSUELLES* DonneesMensuelles, int Numer
       = ProblemeHydraulique->ProblemeLineairePartieVariable[NumeroDeProbleme];
 
     TurbineMax = DonneesMensuelles->TurbineMax;
+    TurbineMin = DonneesMensuelles->TurbineMin;
```
guilpier-code: About **TurbineMin** : 
We declare it, we give it a value but we don't use it in the current function.
If not used, should be removed.

src/solver/hydro/daily/h2o_j_initialiser_les_bornes_des_variables.c
Milos-RTEi: This is *.c file. I think I cannot use c++ functions here, like std::min and std::max. Right?
Not sure about Math from Yuni library also.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -191,18 +192,29 @@ bool DataSeriesHydro::loadFromFolder(Study& study, const AreaName& areaID, const
         if (enabledModeIsChanged)
             JIT::enabled = true; // Back to the previous loading mode.
     }
-
     if (ror.width > count)
         count = ror.width;
 
+    if (study.header.version >= 830)
```
guilpier-code: More **860**, instead of **830**.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -191,18 +192,29 @@ bool DataSeriesHydro::loadFromFolder(Study& study, const AreaName& areaID, const
...
+    }
+
     if (study.usedByTheSolver)
     {
         if (0 == count)
         {
             logs.error() << "Hydro: `" << areaID
                          << "`: empty matrix detected. Fixing it with default values";
-            ror.reset(1, DAYS_PER_YEAR);
+            ror.reset(1, HOURS_PER_YEAR);
```
guilpier-code, line 105: `ror.reset(1, HOURS_PER_YEAR);` : thanks for correction.
Block `if (0 == count)` seems to be dead code.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -245,12 +257,41 @@ bool DataSeriesHydro::loadFromFolder(Study& study, const AreaName& areaID, const
                           << "Impossible to find the area `" << areaID << "` to invalidate it";
                 }
             }
+
+            if(mingen.width != count)
```
guilpier-code, line 151: A code duplication was introduced here. 
Can something be done about it ?

guilpier-code (changes requested):

src/solver/hydro/daily/h2o_j_initialiser_les_bornes_des_variables.c
guilpier-code: > This is *.c file. I think I cannot use c++ functions here, like std::min and std::max. Right? Not sure about Math from Yuni library also.

Then you could turn it into a **.cpp** file and make the change. I tried it and it worked.
Eventually, these **.c** files will be turned into **.cpp** files anyway.

flomnes: We may do this renaming here.

src/solver/hydro/management/management.cpp
```diff
@@ -135,6 +137,89 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
...
+        if (not area.hydro.followLoadModulations)
+        {
+            //CR22: Weekly minimum generation <= Weekly inflows for each week 
+            for (uint week = 0; week < 53; ++week)
+            {
+                double totalWeekMingen = 0.0;
+                double totalWeekInflows = 0.0;
+                for(uint hour = calendar.weeks[week].hours.first; hour < calendar.weeks[week].hours.end; ++hour)
+                {
+                    if (hour < HOURS_PER_YEAR)
```
flomnes:
```suggestion
                for(uint hour = calendar.weeks[week].hours.first; hour < calendar.weeks[week].hours.end && hour < HOURS_PER_YEAR; ++hour)
                {
```

Milos-RTEi:
Transferred to https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1273

PR #1208, opened 2023-03-09
Short term storage : hour for clusters functions

guilpier-code:
This work is a first refactoring before adding new output variables for ST storage.
It intends to simplify or even remove definition and use of :  
- **hourForEachThermalCluster**
- **hourForEachRenewableCluster**

in order to introduce functions **hourForClusters** to handle clusters of all kinds (thermal, renewable and ST storage).
Eventually, the content of these functions should be moved to **hourForEachArea**.

### DIFF
--- a/src/solver/variable/setofareas.h
+++ b/src/solver/variable/setofareas.h
@@ -150,7 +150,7 @@ public:
     void hourBegin(unsigned int hourInTheYear);
     void hourForEachArea(State& state);
     void hourForEachLink(State& state);
-    void hourForEachThermalCluster(State& state);
+    void hourForThermalClusters(State& state);
     void hourEnd(State& state, unsigned int hourInTheYear);
 
     void weekBegin(State&);
--- a/src/solver/variable/setofareas.h
+++ b/src/solver/variable/setofareas.h
@@ -150,7 +150,7 @@ public:
     void hourBegin(unsigned int hourInTheYear);
     void hourForEachArea(State& state);
     void hourForEachLink(State& state);
-    void hourForThermalClusters(State& state);
+    void hourForClusters(State& state);
     void hourEnd(State& state, unsigned int hourInTheYear);
 
     void weekBegin(State&);
--- a/src/solver/variable/state.cpp
+++ b/src/solver/variable/state.cpp
@@ -46,21 +46,15 @@ State::State(Data::Study& s) :
  averageOptimizationTime1(0.),
  averageOptimizationTime2(0.)
 {
-    thermalClustersProductions.resize(s.areas.size());
-    for (int area_index = 0; area_index < s.areas.size(); area_index++)
-        thermalClustersProductions[area_index].resize(s.areas[area_index]->thermal.clusterCount());
+    auto resizeThermal = [&s](auto& container) {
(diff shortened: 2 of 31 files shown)


src/solver/variable/state.h
```diff
@@ -169,24 +162,20 @@ class State
...
     VALEURS_DE_NTC_ET_RESISTANCES* ntc;
-    //! Thermal production for the current thermal cluster for the current hour in the year
-    double thermalClusterProduction;
-    //! Thermal available production for the current thermal cluster for the current hour in the
-    //! year
-    double thermalClusterAvailableProduction;
-    //! The operating cost for the current cluster of the current hour (production level*production
+
+    //! Thermal production for thermal clusters for the current hour in the year
+    std::vector<std::vector<double>> thermalClustersProductions;
```
flomnes, line 167: Please make sure to make it precise what the indices should be

src/solver/variable/state.h
```diff
@@ -169,24 +162,20 @@ class State
...
     //! (accurate) unit commitment mode
-    uint thermalClusterNumberON;
-    //! Minimum power of a group of the cluster for the current hour in the year
-    double thermalClusterPMinOfAGroup;
-    //! Minimum power of the cluster for the current hour in the year
-    double thermalClusterPMinOfTheCluster;
+    std::vector<std::vector<uint>>  numberOfUnitsONbyCluster;
+    
+    //! Minimum power of all clusters for the current hour in the year
+    std::vector<std::vector<double>> PMinOfClusters;
```
flomnes, line 178: Add comment about indices

src/solver/variable/state.cpp
```diff
@@ -46,6 +46,21 @@ State::State(Data::Study& s) :
...
+    for (int area_index = 0; area_index < s.areas.size(); area_index++)
+        PMinOfClusters[area_index].resize(s.areas[area_index]->thermal.clusterCount());
+
+    numberOfUnitsONbyCluster.resize(s.areas.size());
+    for (int area_index = 0; area_index < s.areas.size(); area_index++)
+        numberOfUnitsONbyCluster[area_index].resize(s.areas[area_index]->thermal.clusterCount());
+
+    thermalClustersOperatingCost.resize(s.areas.size());
+    for (int area_index = 0; area_index < s.areas.size(); area_index++)
+        thermalClustersOperatingCost[area_index].resize(s.areas[area_index]->thermal.clusterCount());
```
flomnes: Introduce a lambda function ? It's 4 times the same code here

flomnes: Note that since C++14, `auto` is allowed for lambda function args.

src/solver/variable/setofareas.h
```diff
@@ -150,7 +150,7 @@ class SetsOfAreas
     void hourBegin(unsigned int hourInTheYear);
     void hourForEachArea(State& state);
     void hourForEachLink(State& state);
-    void hourForEachThermalCluster(State& state);
+    void hourForThermalClusters(State& state);
```
flomnes: Why not use `hourForEachArea` ?

flomnes (approved):

Issue #1211, opened 2023-03-13
Replace enum with enum classes


sylvlecl:
We should replace the use of plain enums with enum classes.

Also, enum type should be used everywhere it's relevant instead of plain `int`, see methods such as:
https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/96832a3a7f75534b5cb68ab7f62421a729acfb51/src/solver/variable/surveyresults/surveyresults.h#L86

**Edit**
After first try, one difficulty is that often (not always), those `int` represent a mask (a *set* of enum values) and not a unique value.
We should differentiate places where an actual value of the num is expected, from places where a set of values is expected, by using a different type for that second use case.
For that, we could either use `std::bitset` as suggested by @flomnes or possibly our own class:

```cpp
namespace OldWay {
enum TimeSeries : unsigned int
{
    //! TimeSeries : Load
    timeSeriesLoad = 1,
    //! TimeSeries : Hydro
    timeSeriesHydro = 2,
    //! TimeSeries : Wind
    timeSeriesWind = 4,

    //! The maximum number of time-series that we can encounter
    timeSeriesCount = 3,
}; // enum TimeSeries
}

namespace NewWay {
  enum TimeSeries {
    timeSeriesLoad = 0,
    timeSeriesHydro,
    timeSeriesWind,
    timeSeriesCount
  };
  using TimeSeriesBitset = std::bitset<timeSeriesCount>;
}

int main() {
  {
    using namespace OldWay;
    unsigned int ts; // bof
    ts = timeSeriesLoad | timeSeriesWind;
    assert(ts & timeSeriesLoad);
  }
  {
    using namespace NewWay;
    TimeSeriesBitset ts;
    ts.set(timeSeriesLoad).set(timeSeriesWind);
    assert(ts.test(timeSeriesLoad));
  }
}
```

pet-mit:
ANT-2171

PR #1214, opened 2023-03-13
[WIP] Replace integers by proper enums in methods signatures

sylvlecl:
Prerequisite to #1211 : 
- using enum parameters instead of integers where relevant


flomnes:
This seems to be more than search & replace since for example `timeSeriesThermal | timeSeriesLoad` is not an `enum`.

PR #1217, opened 2023-03-14
ST storage output : adding ST storage output variables

guilpier-code:
This work intends to add output variables related to ST storage.
What should be done : 
- [x] For **values-\<period\>.txt**,  add an output variable class for **injection** and **withdrawal**
- [x] Levels should have their own output variable class (is about output generated files **values-\<period\>.txt** as well). 
Note that levels results cannot be generated inside the same class than **injection** and **withdrawal** : the computation that leads to get daily results from hourly results are not the same (sum in one case, average in the other case)
- [ ] New files **details-STstor-\<period\>.txt** should be generated in output, so we need to create another output variable class for this.
- [ ] Check that everything works as well for **adequacy** mode 
- [x] Check that synthesis and district results we have are the expected ones
- [ ] Class **IVariable** has a member function **hourForEachArea(State& state)** and as well as a new **hourForEachArea(State& state, unsigned int numSpace)** member function. See if the first one is still useful.

### DIFF
--- a/src/solver/variable/economy/shortTermStorage.h
+++ b/src/solver/variable/economy/shortTermStorage.h
@@ -1,0 +1,281 @@
+/*
+** Copyright 2007-2023 RTE
+** Authors: Antares_Simulator Team
+**
+** This file is part of Antares_Simulator.
+**
+** Antares_Simulator is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** There are special exceptions to the terms and conditions of the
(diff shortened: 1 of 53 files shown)


src/solver/variable/economy/shortTermStorage.h
```diff
@@ -0,0 +1,281 @@
...
+    {
+        static const char* Caption(const unsigned int indx)
+        {
+            return st_storage_name(indx).c_str();
+        }
+    };
+}; // class VCard
+
+
+static const std::string& st_storage_name(const unsigned int indx)
```
flomnes:
```suggestion
static std::string st_storage_name(const unsigned int indx)
```

payetvin (approved):

PR #1218, opened 2023-03-14
Add I/O changes for BC RHS scenarization (v8.6)

flomnes:


docs/reference-guide/13-file-format.md
```diff
@@ -1,6 +1,31 @@
...
+A few changes related to the scenarization of RHS for binding constraints.
+
+- For each binding constraint, file **input/bindingconstraints/&lt;id&gt;.txt** is split into 3 files:
+    - **input/bindingconstraints/&lt;id&gt;_lt.txt**
+    - **input/bindingconstraints/&lt;id&gt;_gt.txt**
+    - **input/bindingconstraints/&lt;id&gt;_eq.txt**
+
+    Each of these files must have 8674 rows and N >= 1 columns, the same number for the 3 files.
+
+- In file **input/bindingconstraints/bindingconstraints.ini**, add property `group` to every section
```
JasonMarechal25, line 14: What happens if the property "group" is missing ?
- No scenarisation : pick the first column ?
- Individual Random scenarisation
- All BC with empty group are deemed in the same "default group" and random scenarisation

I would personally prefer 1 or 2.

docs/reference-guide/13-file-format.md
```diff
@@ -1,6 +1,31 @@
...
+```
+bc,<group>,<MC Year> = <TS number>
+```
+This line is not mandatory for every group & MC year. If absent, the TS number will be drawn randomly (usual behavior).
+
+- 0 &lt;= MC Year &lt; generaldata.ini/general.nbyears
+- 1 &lt;=TS number &lt;= number of columns for the group
+
+### Output
+Add directory **bindingconstraints** to output directory **ts-numbers**. For every binding constraint group, add a file **ts-numbers/bindingconstraints/&lt;group&gt;.txt** containing the TS numbers used for that group.
```
JasonMarechal25, line 54: What about anonymous constraints? Can we use the id? Is the id garantee to be unique ? IT may collide with group names

JasonMarechal25 (changes requested):

docs/reference-guide/13-file-format.md
JasonMarechal25: Answer from @PPlessiez:
Each BC in anonymous group, let's call them "orphaned" is considered in its own group. In this case each orphaned BC follow it's own scenario. By default randomised.

docs/reference-guide/13-file-format.md
JasonMarechal25: No anonymous constraints. All constraints belong to a groupe. No issue.

docs/reference-guide/13-file-format.md
```diff
@@ -1,6 +1,31 @@
...
 This is a list of all recent changes that came with new Antares Simulator features. The main goal of this document is to lower the costs of changing existing interfaces, both GUI and scripts.
 
+## v8.6.0
+### Input
+A few changes related to the scenarization of RHS for binding constraints.
+
+- For each binding constraint, file **input/bindingconstraints/&lt;id&gt;.txt** is split into 3 files:
+    - **input/bindingconstraints/&lt;id&gt;_lt.txt**
+    - **input/bindingconstraints/&lt;id&gt;_gt.txt**
+    - **input/bindingconstraints/&lt;id&gt;_eq.txt**
```
JasonMarechal25, line 10: Why lt/gt => for range constraints
Also need to keep files when user change the type of constraint.
@Antarest team: careful not to delete/overwrite old file when user select a new type.

docs/reference-guide/13-file-format.md
```diff
@@ -3,8 +3,26 @@ This is a list of all recent changes that came with new Antares Simulator featur
...
 ## v8.6.0
 ### Input
-A few changes related to the introduction of short-term storage objects.
+#### Scenarized RHS for binding constraints
+- For each binding constraint, file **input/bindingconstraints/&lt;id&gt;.txt** is split into 3 files:
+    - **input/bindingconstraints/&lt;id&gt;_lt.txt**
+    - **input/bindingconstraints/&lt;id&gt;_gt.txt**
+    - **input/bindingconstraints/&lt;id&gt;_eq.txt**
 
+    Each of these files must have 8674 rows and N >= 1 columns, the same number for the 3 files.
```
JasonMarechal25:
```suggestion
    Each of these files can be either empty or have N column of 8674 rows
```

docs/reference-guide/13-file-format.md
```diff
@@ -3,8 +3,26 @@ This is a list of all recent changes that came with new Antares Simulator featur
...
+
+- In file **input/bindingconstraints/bindingconstraints.ini**, add property `group` to every section
+- Binding constraints in the same group must have the same number of RHS columns (3 files described above)
```
JasonMarechal25:
```suggestion
- Binding constraints in the same group must have the same number of RHS columns (3 files described above for their respective types). Exception: a constraint in a group can have empty RHS or only one RHS column
```

JasonMarechal25 (approved):

PR #1228, opened 2023-03-17
Loading Series for short term storage

payetvin:


flomnes (changes requested):
Check vector size in `Series::validate` (8760)

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -55,10 +55,10 @@ bool STstorageInput::createSTstorageClustersFromIniFile(const std::string& path)
 
     for (auto* section = ini.firstSection; section; section = section->next)
     {
-        STstorageCluster cluster;
+        STStorageCluster cluster;
         if (!cluster.loadFromSection(*section))
             return false;
-
+        logs.notice() << section->name.c_str();
```
flomnes: Remove this

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -38,6 +40,44 @@ bool Series::validate() const
...
+    ret = loadFile(folder, "inflow.txt", inflows) && ret;
+    ret = loadFile(folder, "lower-rule-curve.txt", lowerRuleCurve) && ret;
+    ret = loadFile(folder, "upper-rule-curve.txt", upperRuleCurve) && ret;
+
+    return ret;
+}
+
+bool Series::loadFile(const std::string& folder, const std::string& filename, std::vector<double>& vect)
+{
+    std::string path(folder + filename);
```
flomnes: `logs.debug() << "Loading file " << path;`

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -38,6 +40,44 @@ bool Series::validate() const
...
+
+    std::ifstream file;
+    file.open(path);
+
+    if ((file.rdstate() & std::ifstream::failbit ) != 0)
```
```suggestion
    if (!file)
```

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -38,6 +40,44 @@ bool Series::validate() const
...
+    ret = loadFile(folder, "PMAX-injection.txt", maxInjection) && ret;
+    ret = loadFile(folder, "PMAX-withdrawal.txt", maxWithdrawal) && ret;
+    ret = loadFile(folder, "inflow.txt", inflows) && ret;
+    ret = loadFile(folder, "lower-rule-curve.txt", lowerRuleCurve) && ret;
+    ret = loadFile(folder, "upper-rule-curve.txt", upperRuleCurve) && ret;
+
+    return ret;
+}
+
+bool Series::loadFile(const std::string& folder, const std::string& filename, std::vector<double>& vect)
```
flomnes: 1. Move to libs/antares, name this function something "loadSingleColumnFile"
2. Add 8760 as a parameter
3. Clear the vector before writing to it

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -38,6 +40,44 @@ bool Series::validate() const
...
+
+    if ((file.rdstate() & std::ifstream::failbit ) != 0)
+        return false;
+
+    vect.reserve(8760);
+
+    std::string line;
+    try
+    {
+        while (getline(file, line))
```
flomnes: Limit this loop to N=8760 iterations

src/solver/optimisation/opt_structure_probleme_a_resoudre.h
```diff
@@ -111,6 +111,7 @@ typedef struct
     std::vector<std::string> NomDesVariables;
     std::vector<std::string> NomDesContraintes;
 
+//TODO [VP] are those vector really used ?
 public:
```
flomnes: Yes, it is used in e.g **solver/utils/ortools_utils.cpp**

src/libs/antares/study/parts/short-term-storage/series.h
```diff
@@ -27,16 +27,23 @@
 #pragma once
 #include <vector>
 #include <string>
+#include <map>
```
flomnes:
```suggestion
```

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -38,6 +40,44 @@ bool Series::validate() const
...
+bool Series::loadFile(const std::string& folder, const std::string& filename, std::vector<double>& vect)
+{
+    std::string path(folder + filename);
+
+    std::ifstream file;
+    file.open(path);
+
+    if ((file.rdstate() & std::ifstream::failbit ) != 0)
+        return false;
+
```
```suggestion
        logs.error() << "Failed to load file " << path;
        return false;

```

flomnes (changes requested):
I/O can be tricky : there can be many different cases leading to an error : empty file, invalid name, etc.

Make sure to add at least a few error messages to handle this.

Make sure to move generic I/O functions e.g to **libs/antares/io**

After reading values into vectors, make sure to set a default value if the vector is still empty (missing file).

# About missing files
Make missing files non-fatal

All properties should be mandatory since they're used in the case of missing files. We don't want to deal with missing series and property.

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -53,31 +56,32 @@ bool STstorageCluster::loadFromSection(const IniFile::Section& section)
...
 }
 
-bool STstorageCluster::validate()
+bool STStorageCluster::fillDefaultSeries()
 {
-    return properties.validate() && series.validate();
+    if (series.maxInjection.empty())
+        series.maxInjection.resize(VECTOR_SERIES_SIZE, properties.injectionCapacity);
+    if (series.maxWithdrawal.empty())
+        series.maxWithdrawal.resize(VECTOR_SERIES_SIZE, properties.withdrawalCapacity);
```
flomnes: We may use 1 for default value (modulation). To be discussed with JMJ

src/libs/antares/study/parts/short-term-storage/series.h
```diff
@@ -28,19 +28,25 @@
...
-private:
     std::vector<double> maxInjection;
     std::vector<double> maxWithdrawal;
     std::vector<double> inflows;
     std::vector<double> lowerRuleCurve;
     std::vector<double> upperRuleCurve;
-};
+
+    bool loadVector(const std::string& path, std::vector<double>& vect);
+
```
flomnes: `private`

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -78,16 +78,10 @@ bool STstorageInput::loadSeriesFromFolder(const std::string& folder)
 
     for (auto& cluster : storagesByIndex)
     {
-        const std::string buffer = folder + SEP + cluster->properties.name + SEP + "series.txt";
-        /* ret = series->series.loadFromCSVFile(buffer, 1, HOURS_PER_YEAR, &s.dataBuffer) && ret; */
-        cluster->loadSeries(folder);
+        const std::string buffer(folder + SEP + cluster->properties.name + SEP);
```
flomnes: 1. Use id, not the name. Names may contain special characters such as \, /, etc. that can't be used for file names. You need to introduce the id for that, and use function `TransformNameIntoID`
2. Don't pass the last separator to the function, let cluster->loadSeries handle it.
```suggestion
        const std::string buffer(folder + SEP + cluster->id);
```

src/libs/antares/study/parts/short-term-storage/cluster.cpp
flomnes: If so, you may move this method to class `Antares::Data::ShortTermStorage::Series`

flomnes (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -33,12 +36,38 @@ namespace Antares::Data::ShortTermStorage
...
 }
 
 bool Series::loadFromFolder(const std::string& folder)
 {
-    return true;
+    bool ret = true;
+
+    ret = loadVector(folder + "PMAX-injection.txt", maxInjection) && ret;
+    ret = loadVector(folder + "PMAX-withdrawal.txt", maxWithdrawal) && ret;
+    ret = loadVector(folder + "inflow.txt", inflows) && ret;
```
```suggestion
    ret = loadVector(folder + "inflows.txt", inflows) && ret;
```

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -113,6 +142,26 @@ bool Properties::loadKey(const IniFile::Property* p)
...
+        logs.error() << "Property withdrawalnominalcapacity not provided or negative, needed "
+            << "for short term storage";
+        return false;
+    }
+    if (!capacity.has_value() || capacity < 0)
+    {
+        logs.error() << "Property reservoircapacity not provided or negative, needed "
+            << "for short term storage";
+        return false;
+    }
```
flomnes: > Order of evaluation of any part of any expression, including order of evaluation of function arguments is unspecified (with some exceptions listed below). The compiler can evaluate operands and other subexpressions in any order, and may choose another order when the same expression is evaluated again. 

https://en.cppreference.com/w/cpp/language/eval_order

flomnes: In other words, there is no guarantee that `f` will be executed first in expression `f() || g()`.

There is no choice but to have two `if`

PR #1230, opened 2023-03-21
Refactor thermal state

### LINKED ISSUE #1216: Use function existing function `hourForEachArea` instead of introducing `hourForClusters`

flomnes:
* Remove 4 `std::vector` of `std::vector`
* Create dedicated class for thermal state data
* Merge `hourForClusters` into `hourForEachArea`


payetvin (approved):

src/solver/variable/adequacy/overallCost.h
```diff
@@ -251,39 +251,33 @@ class OverallCost : public Variable::IVariable<OverallCost<NextT>, NextT, VCardO
...
-    {
-        // Total OverallCost
-        for (uint clusterIndex = 0; clusterIndex != state.area->thermal.clusterCount();
-             ++clusterIndex)
+        // Thermal costs
+        for (uint clusterIndex = 0; clusterIndex != area->thermal.clusterCount(); ++clusterIndex)
         {
             pValuesForTheCurrentYear[numSpace][state.hourInTheYear]
-              += state.thermalClustersOperatingCost[state.area->index][clusterIndex];
+              += state.thermal[area->index].thermalClustersOperatingCost[clusterIndex];
```
guilpier-code: You define `auto area = state.area;` above, why not defining **thermal** in the same way ?
This remark applies to all output variable classes involving clusters (of all kind).

src/solver/variable/economy/shortTermStorageLevels.h
```diff
@@ -215,18 +215,19 @@ class ShortTermStorageLevels
         NextType::computeSummary(numSpaceToYear, nbYearsForCurrentSummary);
     }
 
-    void hourForClusters(State& state, unsigned int numSpace)
+    void hourForEachArea(State& state, unsigned int numSpace)
```
guilpier-code, line 218: ok. This file will be soon removed

src/solver/variable/state.h
```diff
@@ -37,23 +37,40 @@
...
+
+        //! Number of units turned ON by cluster for the current hour in the year with the ucMILP
+        //! (accurate) unit commitment mode
+        std::vector<uint> numberOfUnitsONbyCluster;
+
+        //! Minimum power of all clusters for the current hour in the year
+        std::vector<double> PMinOfClusters;
+    };
+
+    StateForAnArea& operator[](size_t index);
```
guilpier-code:
```C++
StateForAnArea& operator[](size_t index);
```
I guess the argument is an index for an area, right ?
Could we rename this argument **area_index** ?

src/solver/variable/state.h
```diff
@@ -37,23 +37,40 @@
 
 namespace Antares::Solver::Variable
 {
+class ThermalState
```
guilpier-code, line 40: I can't prevent myself from thinking that the computation we do on thermal clusters in **State** class after the optimization has run is an optimization post-treatment. 
Maybe these vectors should join the **problemeHebdo->ResultatsHoraires** and be computed as a post-treatment. 
If this is suitable and doable, the **State** class should shrink dramatically.

guilpier-code (changes requested):

guilpier-code:
To me, this change has pros and cons : 
- Surely it removes the **hourForClusters** member function.
- But it makes **hourForEachArea** longer and more complex.

I don't really have an opinion on this, maybe it's better that way.

src/solver/variable/state.h
flomnes: Yes, the problem is that it's a **yearly** post-treatment, which is why those vectors are of size 8760. Therefore it wouldn't make sense to place it inside problemeHebdo->ResultatsHoraires, since each variable defined there only has size 168 (see e.g src/solver/simulation/economy.cpp:98)

PR #1234, opened 2023-03-23
Refactor constraint builder

### LINKED ISSUE #1205: Short term storage : refactor the OPT part
As an example, function **OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(...)** and the source file it belongs to are growing.
A possible quite simple solution for them to stop growing would be to have : 
```C++
void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(PROBLEME_HEBDO* problemeHebdo, ...)
{
      problemeHebdo->hydroModeller.giveBoundsToOptVariables();
      problemeHebdo->thermalModeller.giveBoundsToOptVariables();
      ...
      // And of course
      problemeHebdo->shortTermStorageModeller.giveBoundsToOptVariables();
      ...
}
```
Several remarks about this :
- This solution should be better than what we have now : the current **OPT** code forces us to make functions grow and do much more things than they should.
- It intends to extract bound settings in a more specialized class, called for instance **ShortTermstorageModeller**.
- This solution would be applicable to any other functions in the **OPT** part. This means, for example that the **ShortTermstorageModeller** would have a public member function **addConstraintsToProblem()**
- It would be applicable to any other production asset taking a part in the optimization 
- Class **ShortTermstorageModeller** would have a constructor taking as argument  a **PROBLEME_HEBDO** and characteristics on short term storage.
- Doing this for every production asset and every **OPT** function should eventually lead to simplify the code a lot : for instance, we could eventually have : 
```C++
 // Only when setting bounds here, but could be applied to any optimization preparation stage
void OPT_InitialiserLesBornesDesVariablesDuProblemeLineaire(ProductionAssetList & list_of_assets)
{
    for (auto* asset : list_of_assets)
        asset->giveBoundsToOptVariables();
}
```

flomnes:
### TODO
- [ ] `OPT_Export_add_variable` calls need to be moved
- [X] Removed `CORRESPONDANCES_DES_CONTRAINTES`
- [X] Merged RHS & matrix construction
- [x] Hourly binding ct
- [x] Daily binding ct
- [x] Weekly binding ct
- [x] Flow dissociation
- [x] Area balance
- [x] Short-term storage
- [X] Clarify differences between `pdt` & `pdtHebdo`
- [X] Refactor constraint naming
- [ ] Refactor variable creation & naming => Later
- [ ] ~~Refactor objective setting~~
- [ ] Restore `adresseDuResultat` / `AdresseOuPlacerLaValeurDesCoutsMarginaux`
- [ ] Most hydro & thermal-accurate constraints
- [ ] Split Matrix / RHS creation ?


src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
```diff
@@ -31,206 +31,435 @@
...
 
-        CORRESPONDANCES_DES_CONTRAINTES_JOURNALIERES& CorrespondanceCntNativesCntOptimJournalieres
-              = problemeHebdo->CorrespondanceCntNativesCntOptimJournalieres[indexCorrespondanceCnt];
+    // TODO reset selectively
+    ProblemeAResoudre->NombreDeContraintes = 0;
+    ProblemeAResoudre->NombreDeTermesDansLaMatriceDesContraintes = 0;
 
-        for (int cntCouplante = 0; cntCouplante < problemeHebdo->NombreDeContraintesCouplantes;
+    AreaBalance areaBalance(problemeHebdo);
+    FictiveLoad fictiveLoad(problemeHebdo);
```
sylvlecl: I am not sure anymore of the added value of the comon interface `IConstraint`: it does not seem to be used, we only use plain values instead?

sylvlecl: I think "fictitious" is more appropriate than "fictive"

src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
```diff
@@ -31,206 +31,435 @@
...
+    // TODO improve this
     {
-        AdresseOuPlacerLaValeurDesCoutsMarginaux[i] = nullptr;
+      ConstraintNamer namer(problemeHebdo->ProblemeAResoudre->NomDesContraintes,
+                          problemeHebdo->NamedProblems);
+      namer.UpdateTimeStep(problemeHebdo->weekInTheYear * 168 + pdt);
+      namer.UpdateArea(problemeHebdo->NomsDesPays[pays]);
+      namer.AreaBalance(problemeHebdo->ProblemeAResoudre->NombreDeContraintes);
+    }
+    builder.updateHourWithinWeek(pdt);
```
sylvlecl: I think we should have a new `builder` for each constraint, it would be more robust.

Actually the build could throw if we try to re-use it, to make 100% sure our constraints creation do not overlap.

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,206 @@
...
+
+    void build();
+
+private:
+    PROBLEME_ANTARES_A_RESOUDRE& problemeAResoudre;
+    PROBLEME_HEBDO& problemeHebdo;
+    const std::vector<CORRESPONDANCES_DES_VARIABLES>& varNative;
+
+    unsigned int hourInWeek_ = 0;
+    // TODO: vérifier que l'index correspond bien à l'objet concerné (zone, lien, cluster thermique, etc.)
```
sylvlecl: Instead of the updateIndex, we could have specific classes for each kind of objects, for example:
```cpp
builder.area(1)  // Return a class specific to areas
    .include(Variable::HydProd, 1.0)  // checks that the variable is only for areas.
                                     // actually we could even have one method by acceptable variable
```

This would also enable to have nested classes for clusters, for example:
```cpp
builder.area(1)
    .cluster(3)
    .include(...)
```

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,206 @@
...
+    unsigned int index_ = 0;
+    char operator_;
+    double rhs_ = 0;
+    int nombreDeTermes_ = 0;
+};
+
+
+class IConstraint {
+public:
+  IConstraint(PROBLEME_HEBDO* problemeHebdo) : problemeHebdo(problemeHebdo), builder(*problemeHebdo, problemeHebdo->CorrespondanceVarNativesVarOptim)
```
sylvlecl: If we keep the interface, I would suggest to make it pure virtual, to not mix the 2 concerns which are:
- on one side defined the interface contract
- on another side share some code between implementations, which are actually implementation details

See for ex in [CPP core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c121-if-a-base-class-is-used-as-an-interface-make-it-a-pure-abstract-class)

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,206 @@
...
+      rhs_ = rhs;
+      return *this;
+    }
+    ConstraintBuilder& less(double rhs)
+    {
+      operator_ = '<';
+      rhs_ = rhs;
+      return *this;
+    }
+    ConstraintBuilder& greater(double rhs)
```
sylvlecl: greaterThan ?

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,206 @@
...
+      }
+      return *this;
+    }
+    ConstraintBuilder& equal(double rhs)
+    {
+      operator_ = '=';
+      rhs_ = rhs;
+      return *this;
+    }
+    ConstraintBuilder& less(double rhs)
```
sylvlecl: lessThan

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,206 @@
...
+      return *this;
+    }
+    ConstraintBuilder& greater(double rhs)
+    {
+      operator_ = '>';
+      rhs_ = rhs;
+      return *this;
+    }
+
+    void build();
```
sylvlecl, line 372: Maybe this could be called directly when defining operator and rhs, but not sure.

sylvlecl (commented):
Some comments/suggestions

src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
flomnes: Yes, this interface can be safely removed. 

Fictitious seems better suited indeed.

src/solver/optimisation/constraint_builder.h
flomnes: I agree, although for thermal clusters & short-term storages, the index is **global** so we don't need to provide area indices.

src/solver/optimisation/constraint_builder.h
flomnes: Indeed ! Let's this class concrete.

src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
```diff
@@ -138,240 +166,758 @@ void OPT_InitialiserLeSecondMembreDuProblemeLineaire(PROBLEME_HEBDO* problemeHeb
...
-            }
+struct FinalStockEquivalent : public Constraint
+{
+    using Constraint::Constraint;
+    void add(int pays)
+    {
+        const auto& pdt = problemeHebdo->NombreDePasDeTempsPourUneOptimisation - 1;
+        builder.updateHourWithinWeek(pdt)
+          .updateHourWithinWeek(problemeHebdo->NombreDePasDeTempsPourUneOptimisation - 1)
+          .include(Variable::FinalStorage(pays), 1.0)
```
flomnes:
```suggestion
        builder.updateHourWithinWeek(pdt)
          .include(Variable::FinalStorage(pays), 1.0)
```

src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
```diff
@@ -138,240 +166,758 @@ void OPT_InitialiserLeSecondMembreDuProblemeLineaire(PROBLEME_HEBDO* problemeHeb
...
+    }
+};
+
+struct FinalStockExpression : public Constraint
+{
+    using Constraint::Constraint;
+    void add(int pays)
+    {
+        const auto& pdt = problemeHebdo->NombreDePasDeTempsPourUneOptimisation - 1;
+        builder.updateHourWithinWeek(pdt)
```
flomnes:
```suggestion
        const auto pdt = problemeHebdo->NombreDePasDeTempsPourUneOptimisation - 1;
        builder.updateHourWithinWeek(pdt)
```

src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
```diff
@@ -138,240 +166,758 @@ void OPT_InitialiserLeSecondMembreDuProblemeLineaire(PROBLEME_HEBDO* problemeHeb
...
-                      += problemeHebdo->CaracteristiquesHydrauliques[pays].NiveauInitialReservoir;
-                }
-                AdresseOuPlacerLaValeurDesCoutsMarginaux[cnt] = nullptr;
-            }
+struct FinalStockEquivalent : public Constraint
+{
+    using Constraint::Constraint;
+    void add(int pays)
+    {
+        const auto& pdt = problemeHebdo->NombreDePasDeTempsPourUneOptimisation - 1;
```
flomnes, line 756:
```suggestion
        const auto pdt = problemeHebdo->NombreDePasDeTempsPourUneOptimisation - 1;
```
Not really an lvalue because of the "-1".

a-zakir: definitely

PR #1236, opened 2023-03-24
Unit testing for short term storage loading

payetvin:


src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -0,0 +1,279 @@
...
+    BOOST_CHECK(series.validate());
+
+    removeFileSeries();
+}
+
+BOOST_AUTO_TEST_CASE(check_series_folder_loading_negative_value)
+{
+    std::string folder = getFolder();
+
+    createFileSeries(-247.0, 8760);
```
flomnes, line 197: we should check values for time-series, this shouldn't pass

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(check_series_folder_loading_too_big)
+{
+    std::string folder = getFolder();
+
+    createFileSeries(1.0, 9000);
+
+    BOOST_CHECK(series.loadFromFolder(folder));
+    BOOST_CHECK(series.validate());
```
flomnes, line 208: This shouldn't pass, right ?

flomnes (changes requested):

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+#define SEP Yuni::IO::Separator
+
+using namespace std;
+using namespace Antares::Data;
+
+std::string getFolder()
+{
+    std::filesystem::path tmpDir = std::filesystem::temp_directory_path();
+    return tmpDir;
```
flomnes:
```suggestion
    return tmpDir.string();
```

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
payetvin: We give the size to the function that loads the vector, it reads just what is needed to so it pass

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(check_series_folder_loading)
+{
+    std::string folder = getFolder();
+
+    createFileSeries(1.0, 8760);
+
+    BOOST_CHECK(series.loadFromFolder(folder));
+    BOOST_CHECK(series.validate());
```
sylvlecl, line 182: We also need to check that the values we read are correct (1.0 here).
It would be nice to have a test with a vector containing different values, not only constant vectors.

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,23 +44,92 @@ bool Series::validate() const
...
+        return std::count_if(v.begin(), v.end(), [](double d){ return d >= 0.0; });
+    };
+    if (!checkVectPositive(maxInjectionModulation) || !checkVectPositive(maxWithdrawalModulation))
+    {
+        logs.warning() << "Values for PMAX series should be positive";
+        return false;
+    }
+
+    auto checkVectBetween = [](std::vector<double> v) {
+        return std::count_if(v.begin(), v.end(), [](double d){ return (d >= 0.0 && d <= 1.0); });
```
```suggestion
        return std::all_of(v.begin(), v.end(), [](double d){ return (d >= 0.0 && d <= 1.0); });
```
Pass the vector by const ref

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,23 +44,92 @@ bool Series::validate() const
...
+    ret = loadFile(folder + SEP + "inflows.txt", inflows) && ret;
+    ret = loadFile(folder + SEP + "lower-rule-curve.txt", lowerRuleCurve) && ret;
+    ret = loadFile(folder + SEP + "upper-rule-curve.txt", upperRuleCurve) && ret;
 #undef SEP
     return ret;
 }
 
-bool Series::loadVector(const std::string& path, std::vector<double>& vect) const
+bool loadFile(const std::string& path, std::vector<double>& vect)
+{
```
flomnes: This belongs in a generic part of the code, not here (I/O)

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,23 +44,92 @@ bool Series::validate() const
...
+            count++;
+        }
+        if (count < HOURS_PER_YEAR)
+            return false;
+    }
+    catch (const std::exception&)
+    {
+        logs.error() << "Failed reading file: " << path;
+        return false;
+    }
```
flomnes: Catching a generic exception won't be of much help to the user
```suggestion
    catch (const std::ios_base::failure& ex)
    {
        logs.error() << "Failed reading file: " << path << " (I/O error)";
        return false;
    }
    catch (const std::invalid_argument& ex)
    {
        logs.error() << "Failed reading file: " << path << " conversion to double failed at line " << line;
    }
    catch (const std::out_of_range& ex)
    {
        logs.error() << "Failed reading file: " << path << " value is out of bounds at line " << line;
    } 
```

src/libs/antares/study/parts/short-term-storage/series.h
```diff
@@ -44,8 +44,9 @@ class Series
     std::vector<double> inflows;
     std::vector<double> lowerRuleCurve;
     std::vector<double> upperRuleCurve;
-
-private:
-    bool loadVector(const std::string& path, std::vector<double>& vect) const;
 };
+
+bool loadFile(const std::string& folder, std::vector<double>& vect);
+bool loadVector(const std::string& path, std::vector<double>& vect);
```
flomnes: Remove this decl, now unused

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,23 +44,92 @@ bool Series::validate() const
...
+    return true;
+}
+
+
+bool loadVector(const std::string& path, std::vector<double>& vect)
```
flomnes: Remove this def, now unused

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -0,0 +1,317 @@
...
+void createFileSeries(unsigned int size)
+{
+    std::string folder = getFolder();
+
+    createIndividualFileSeries(folder + SEP + "PMAX-injection.txt", size);
+    createIndividualFileSeries(folder + SEP + "PMAX-withdrawal.txt", size);
+    createIndividualFileSeries(folder + SEP + "inflows.txt", size);
+    createIndividualFileSeries(folder + SEP + "lower-rule-curve.txt", size);
+    createIndividualFileSeries(folder + SEP + "upper-rule-curve.txt", size);
+}
```
flomnes: I'd rather not have random generation in unit tests. It can fluctuate `ie` give different results in a different run/platform.

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -0,0 +1,317 @@
...
+void removeFileSeries()
+{
+    std::string folder = getFolder();
+
+    std::filesystem::remove(folder + SEP + "PMAX-injection.txt");
+    std::filesystem::remove(folder + SEP + "PMAX-withdrawal.txt");
+    std::filesystem::remove(folder + SEP + "inflows.txt");
+    std::filesystem::remove(folder + SEP + "lower-rule-curve.txt");
+    std::filesystem::remove(folder + SEP + "upper-rule-curve.txt");
+}
```
flomnes: Use a fixture class ?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,29 +44,75 @@ bool Series::validate() const
...
+    {
+        unsigned int count = 0;
+        while (getline(file, line) && count < HOURS_PER_YEAR)
+        {
+            double d = std::stod(line);
+            vect.push_back(d);
+            count++;
+        }
+        if (count < HOURS_PER_YEAR)
+            return false;
```
flomnes: Please log this error

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,29 +44,75 @@ bool Series::validate() const
...
+        logs.debug() << "File not found: " << path;
         return true;
+    }
 
-    vect.resize(HOURS_PER_YEAR);
-    return Array1DLoadFromFile(path.c_str(), vect.data(), HOURS_PER_YEAR);
+    std::string line;
+    try
+    {
+        unsigned int count = 0;
```
```suggestion
        unsigned int line = 0;
```

flomnes (commented):
In case of error, give the name of the short-term-storage

src/libs/antares/study/parts/short-term-storage/series.cpp
payetvin: Line is already used for std::string name, I'll replace count by lineCount

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
payetvin: It can fluctuate but should always return a value between 0 and 1 no ? This test is used to check if loading a file with different values work, using random seems like a good idea but i can do something like **line(n) = 0.0001 * n** to fill the file

flomnes: Yeah that's better, more predictable. We don't want our tests to start failing because the RNG was updated

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,29 +44,90 @@ bool Series::validate() const
...
+    }
+    catch (const std::ios_base::failure& ex)
+    {
+        logs.error() << "Failed reading file: " << path << " (I/O error)";
+        return false;
+    }
+    catch (const std::invalid_argument& ex)
+    {
+        logs.error() << "Failed reading file: " << path << " conversion to double failed at line "
+            << line;
```
flomnes: lineCount + 1 (line would be the line itself)

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -42,29 +44,90 @@ bool Series::validate() const
...
+        return false;
+    }
+    catch (const std::out_of_range& ex)
+    {
+        logs.error() << "Failed reading file: " << path << " value is out of bounds at line "
+            << line;
```
flomnes: lineCount

src/tests/src/libs/antares/study/short-term-storage-input/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+# ====================================
+set(SRC_SC_BUILDER_READ
+	short-term-storage-input.cpp
+)
+add_executable(short-term-storage-input ${SRC_SC_BUILDER_READ})
+
+target_include_directories(short-term-storage-input
+								PRIVATE
+									"${src_libs_antares_study}"
+									"${src_libs_antares_study}/parts/short-term-storage"
```
flomnes, line 14: I think it would be better to remove this line, and to `#include <antares/study.h>`. It gives you access to most elements I think (no public/private headers)

flomnes (changes requested):

src/tests/src/libs/antares/study/short-term-storage-input/CMakeLists.txt
payetvin: I removed "${src_libs_antares_study}" since we only need container.h for this file

PR #1238, opened 2023-03-27
storage storage output : files details + values

guilpier-code:
See **caution** section below.

- [x] adding a class to handle ST storage injection in details results
- [x] adding a class to handle ST storage withdrawal in details results
- [x] adding a class to handle ST storage levels in details results
- [x] modify class **ShortTermStorageEnergy** to intertwine correctly injections, withdrawals and levels in output files **values-\<period\>.txt**

**Caution** : 
As you can see, there are many files modified in this PR.
This comes from the fact that we merged branch **max-nb-colunms-for-print** into the current branch.


src/solver/variable/economy/STStorageInjectionByCluster.h
```diff
@@ -0,0 +1,346 @@
...
+namespace Antares
+{
+namespace Solver
+{
+namespace Variable
+{
+namespace Economy
+{
+struct VCardSTstorageInjectionByCluster
+{
```
flomnes: Merge namespaces

src/solver/variable/economy/shortTermStorage.h
```diff
@@ -100,37 +100,55 @@ struct VCardShortTermStorageEnergy
             case 1:
                 return "PSP_open_withdrawal";
             case 2:
-                return "PSP_closed_injection";
+                return "PSP_open_level";
```
flomnes: It would be nice if we could avoid having this function, use a loop instead. But I know there's a problem with `const char*` causing segfaults...

We should consider switching to `std::string` eventually.

src/solver/variable/economy/shortTermStorage.h
```diff
@@ -217,8 +235,24 @@ class ShortTermStorageEnergy
...
+        // Here we perform time-aggregations :
+        // ---------------------------------
+        // for a given MC year, from hourly results we compute daily, weekly, monthly and annual 
+        // results by aggregation operations (averages or sums).
+        // Caution : 
+        //  - level results are stored in columns of which indices satisfy : col_index % 3 == 1.
+        //    They are time-aggregated by means of averages
+        //  - injection and withdrawal results are stored in columns of which indices 
+        //    satisfy : col_index % 3 != 1.
+        //    They are time-aggregated by means of sums.
```
flomnes: You could make this "hack" a bit more expressive with booleans
```cpp
bool isLevel = (col_index % 3) == XXX;
bool isInjection = (col_index % 3) == XXX;
bool isWithdrawal = (col_index % 3) == XXX;
```

src/solver/variable/info.h
```diff
@@ -381,7 +381,15 @@ struct VariableAccessor<ResultsT, Category::dynamicColumns>
         assert(results.data.area && "Area is NULL");
         const bool thermal_details = fileLevel & Category::de;
         const bool renewable_details = fileLevel & Category::de_res;
-        if (thermal_details && renewable_details)
+        const bool st_storage_details = fileLevel & Category::de_sts;
+
+        std::array<bool, 3> kind_of_details = { thermal_details, renewable_details , st_storage_details };
```
flomnes, line 386: You could use [`std::bitset`](https://en.cppreference.com/w/cpp/utility/bitset) and [`std::bitset::count`](https://en.cppreference.com/w/cpp/utility/bitset/count) to do exactly this.

src/solver/variable/economy/STStorageInjectionByCluster.h
```diff
@@ -0,0 +1,334 @@
...
+    //! Unit
+    static const char* Unit()
+    {
+        return "P-injection - MWh";
+    }
+
+    //! The short description of the variable
+    static const char* Description()
+    {
+        return "Energy generated by ST storage clusters";
```
flomnes: No energy is generated by ST storages, in fact they lose some if efficiency < 1 !

I'd say something akin to "released".

src/solver/variable/economy/STStorageInjectionByCluster.h
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    void initializeFromArea(Data::Study* study, Data::Area* area)
+    {
+        // Get the number of years in parallel
+        pNbYearsParallel = study->maxNbYearsInParallel;
+        pValuesForTheCurrentYear = new VCardType::IntermediateValuesBaseType[pNbYearsParallel];
+
+        // Get the area
+        nbClusters_ = area->shortTermStorage.storagesByIndex.size();
```
flomnes:
```suggestion
        nbClusters_ = area->shortTermStorage.count();
```

flomnes (commented):
Looks very promising, good job !

Lots of code duplication, but we'll figure it out later when we decide to work on output variables. And we will, someday

src/solver/variable/economy/STStorageInjectionByCluster.h
guilpier-code: Done

src/solver/variable/info.h
guilpier-code: I had a quick look on this.
What do you have in mind ?
From my understanding, boolean values have to be converted into bits before insertion in the **std::bitset** object, and this would make the code more complex. 
Did I miss something ?

flomnes: fileLevel would have to be a `std::bitset`

src/solver/variable/economy/shortTermStorage.h
guilpier-code: I have a problem here.

In the case of the pollutants, the returned **std::string**  has an address that cannot be corrupted : it is an element of a constant standard container (an **std::map**). This container is built at compile time and requires no computation.

In the current case, we also need a constant container to hold the strings returned internally inside `static const char* Caption(/*...*/ index)`, but its content cannot be computed at compile time, as it is **const**.

I leave the code as it is until we find a solution to this.

src/solver/variable/economy/STStorageInjectionByCluster.h
guilpier-code: Done, at all places I could find

src/solver/variable/economy/shortTermStorage.h
guilpier-code: ok, I switched to **std::string** as a return type of all **Caption(...)** static member functions

flomnes:
Can you check that new variables (related to ST storages) appear in the "Output selection" window ? I don't think they do

image

guilpier-code:
> Can you check that new variables (related to ST storages) appear in the "Output selection" window ? I don't think they do

They do : I used them to make some checks on the results (disabling all variables but the ST storage ones), and I did it from the GUI. Please scrawl to the end of the list.

flomnes:
What about filtering for files **details-STstorage-XXX.txt** ? We may not want them for time precisions (hourly, daily, weekly, etc.) ?

To answer my own question, geographic trimming seems to work.

Thematic trimming as well.

Good job !

flomnes:
Please fix conflicts with the base branch

src/solver/variable/economy/STStorageInjectionByCluster.h
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    void hourForEachArea(State& state, unsigned int numSpace)
+    {
+        // Next variable
+        NextType::hourForEachArea(state, numSpace);
+    }
+
+    void hourForClusters(State& state, unsigned int numSpace)
+    {
```
flomnes: It seems like this function no longer exists, please move this code to `hourForEachArea`

flomnes (changes requested):
The units are as follows

- Power (injection/withdrawal) -> MW (currently MWh)
- Levels -> MWh (currently MWh)

src/libs/antares/study/parts/short-term-storage/STStorageOutputCaptions.cpp
```diff
@@ -0,0 +1,36 @@
...
+    "PSP_closed", 
+    "Pondage", 
+    "Battery", 
+    "Other1", 
+    "Other2", 
+    "Other3", 
+    "Other4", 
+    "Other5"};
+
+const std::vector<std::string> variables = {"injection", "withdrawal", "level"};
```
flomnes: Make these structures `static`.

Or possibly re-use structures defined in properties.h/properties.cpp

guilpier-code:
> What about filtering for files **details-STstorage-XXX.txt** ? We may not want them for time precisions (hourly, daily, weekly, etc.) ?

I tested the

guilpier-code:
> The units are as follows
> 
> * Power (injection/withdrawal) -> MW (currently MWh)
> * Levels -> MWh (currently MWh)

For the detail files, I could correct it easily, because classes for output of injections, withdrawals and levels are separated.
But for files **values-\<period\>.txt**, all ST storage output variables are dealt with inside one class.
So currently in these files, units of ST storage output variables are the same.
To keep classes as they are, we could change functions `VCardT::Unit()` into `VCardT::Unit(uint index)` but there would be a bit of work. To be examined.

This issue was addressed in a new PR : see #1248

src/solver/variable/economy/shortTermStorage.h
guilpier-code: Done. 
I corrected an error here (level columns satisfy `col_index % 3 == 2` instead of `col_index % 3 == 1`).

src/solver/variable/economy/STStorageInjectionByCluster.h
guilpier-code: Done. I chose **injected**

src/solver/variable/economy/shortTermStorage.h
```diff
@@ -217,8 +177,30 @@ class ShortTermStorageEnergy
...
+            bool isAnInjectionColumn = (col_index % 3) == 0;
+            bool isAnWithdrawalColumn = (col_index % 3) == 1;
+            bool isALevelColumn = (col_index % 3) == 2;
+
+            if (isALevelColumn)
+                pValuesForTheCurrentYear[numSpace][col_index].computeAveragesForCurrentYearFromHourlyResults();
+            if (isAnInjectionColumn)
+                pValuesForTheCurrentYear[numSpace][col_index].computeStatisticsForTheCurrentYear();
+            if (isAnWithdrawalColumn)
+                pValuesForTheCurrentYear[numSpace][col_index].computeStatisticsForTheCurrentYear();
```
flomnes:
```suggestion
            if (isAnInjectionColumn || isAnWithdrawalColumn)
                pValuesForTheCurrentYear[numSpace][col_index].computeStatisticsForTheCurrentYear();

flomnes (changes requested):

src/solver/variable/economy/STStorageInjectionByCluster.h
guilpier-code: Done

src/libs/antares/study/parts/short-term-storage/STStorageOutputCaptions.cpp
guilpier-code: Done

src/solver/variable/economy/shortTermStorage.h
guilpier-code: Done

flomnes (approved):

PR #1239, opened 2023-03-27
Intra-weekly cycle for short-term storage

flomnes:
This dev is a bit more complicated than I first thought.

Initially, I wanted to take advantage of the "Native problem => optimization problem" variable mapping and cheat a little bit with it.

As it turns out, if the intra-weekly cycle is less than the number of hours (typically 168), we want the initial level to lie inside all rule curves for t=0, t=cycle, t=2*cycle, etc.

We could impose this in the optimization problem, but there might be cases when Xmax < Xmin if the rule curves vary a lot.

In such cases, we don't want to provide inconsistent problems to the LP solver, because then it would be harder to trace why the problem is infeasible.

In fact, we want to detect this inconsistency earlier, for example right after loading the data.

In order to do this, we need a calendar or "study range" (to be investigated), not just abstract 8760 hourly points. For that, we need to introduce a real calendar with weeks, days and hours into a check function. This is unusual for timeseries, except in the hydro allocation heuristic where a similar mechanism is used (thanks @PPlessiez).

The solution will eventually look something like
```cpp
bool checkCyclicityAndRuleCurvesBounds(
const Antares::Data::Calendar& calendar,
const std::vector<double>& lowerRuleCurves,
const std::vector<double>& upperRuleCurves,
// etc.
);
```
We also want to keep those bounds in memory, which begs the question: where exactly ?


src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -167,9 +167,10 @@ struct PROPERTIES
     double injectionCapacity;
     double withdrawalCapacity;
     double efficiency;
-
     std::optional<double> initialLevel;
 
+    int storagecycle;
```
sylvlecl: shouldn't it be optional here too?

src/libs/antares/study/parts/short-term-storage/container.h
```diff
@@ -35,7 +35,7 @@ namespace Antares::Data::ShortTermStorage
 class STStorageInput
 {
 public:
-    bool validate() const;
+    bool validate(int simplexe) const;
```
sylvlecl: I think simplex spells without the final "e"

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -62,9 +65,9 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
     return true;
 }
 
-bool STStorageCluster::validate()
+bool STStorageCluster::validate(int simplexe)
```
sylvlecl: Why not use the enum `SimplexOptimization` instead of plain int ?

flomnes (commented):
* Don't pass `int`, pass `Antares::Data::SimplexOptimization`
* Don't use values (0, 1, 2), use `sorUnknown`, `sorDay`, `sorWeek`

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -191,6 +196,31 @@ bool Properties::validate()
         }
     }
 
+    if (simplexe == 1)
+    {
+        if (storagecycle.value() > 24 || storagecycle.value() < 1)
+        {
+            logs.warning() << "simplexe optimzation range set to day: "
+            "storagecycle for cluster: " << name << " should be <= 24 and >= 1";
+            storagecycle = 24;
```
sylvlecl: we should probably state in the log that we force it to 24

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -191,6 +196,31 @@ bool Properties::validate()
...
+        }
+    }
+    else if (simplexe == 2)
+    {
+
+        if (storagecycle.value() > 168 || storagecycle.value() < 1)
+        {
+            logs.warning() << "simplexe optimzation range set to week: "
+            "storagecycle for cluster: " << name << " should be <= 168 and >= 1";
+            storagecycle = 24;
```
sylvlecl: we should probably state in the log that we force it to 24

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -191,6 +196,31 @@ bool Properties::validate()
...
+        }
+    }
+    else
+    {
+        logs.warning() << "Simplexe optimzation range value not recnognized";
```
sylvlecl: typos in recognized / optimization / simplex :)

sylvlecl (commented):
Some comments/questions on implementation details

src/libs/antares/study/parts/short-term-storage/cluster.cpp
payetvin: we can but we need to include parameter.h

src/solver/simulation/sim_structure_probleme_economique.h
flomnes: yes

src/libs/antares/study/area/list.cpp
```diff
@@ -1085,7 +1085,7 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
             << SEP << area.id;
```
guilpier-code, line 945: Shouldn't we have an `if (study.header.version >= 860)` before loading the ST storage time series ?

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -186,11 +191,38 @@ bool Properties::validate()
         if (initialLevel > capacity)
         {
             logs.warning() << "initiallevel for cluster: " << name
-                           << " should be inferior to reservoir capacity: " << capacity.value();
+                << " should be inferior to reservoir capacity: " << capacity.value();
             initialLevel = capacity;
         }
     }
 
+    if (simplex == Antares::Data::sorDay)
```
guilpier-code: There is a bit of code duplication here, not a massive one though.

src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -52,7 +53,7 @@ unsigned int groupIndex(Group group);
 class Properties
 {
 public:
-    bool validate();
+    bool validate(Antares::Data::SimplexOptimization simplex);
```
guilpier-code: About `bool validate(Antares::Data::SimplexOptimization simplex);` : 
As **simplex** can only take 2 values (Day or Week), maybe we could have instead : 
`bool validate(bool isWeeklySimplex);`
This would avoid a dependency to **parameters.h**.

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -57,7 +57,10 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
     }
 
     if (properties.name.empty())
+    {
```
guilpier-code, line 56: Every cluster property value seems to be checked in **Properties::validate(...)** but **properties.name**, whose value is checked here.
Why is that ?

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/container.h
```diff
@@ -35,7 +35,7 @@ namespace Antares::Data::ShortTermStorage
 class STStorageInput
 {
 public:
-    bool validate() const;
+    bool validate(Antares::Data::SimplexOptimization simplex) const;
```
guilpier-code: Please see [this](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1239#discussion_r1158654204)

src/libs/antares/study/parts/short-term-storage/cluster.h
```diff
@@ -37,6 +37,7 @@ namespace Antares::Data::ShortTermStorage
 class STStorageCluster
 {
 public:
+    bool validate(Antares::Data::SimplexOptimization simplex);
```
guilpier-code: Please see [this](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1239#discussion_r1158654204)

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -36,16 +36,20 @@
...
+    return std::all_of(storagesByIndex.cbegin(), storagesByIndex.cend(), [&simplex](auto& cluster) {
+        return cluster->validate(simplex);
     });
 }
 
 bool STStorageInput::createSTStorageClustersFromIniFile(const std::string& path)
 {
     const std::string pathIni(path + SEP + "list.ini");
+
+    if (!Yuni::IO::File::Exists(pathIni))
```
guilpier-code, line 59: Should we not return **false** here ?

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -97,7 +97,7 @@ bool Properties::loadKey(const IniFile::Property* p)
         return p->value.to<std::string>(this->name);
 
     if (p->key == "storagecycle")
-        return p->value.to<unsigned int>(this->cycleDuration);
+        return valueForOptional(this->storagecycle);
 
     if (p->key == "initiallevel")
```
guilpier-code, line 102: For now, if **initialLevel** has no value (for instance **initialLevel  = optim** in the **list.ini**), the code should crash ?

guilpier-code (changes requested):

src/solver/optimisation/opt_construction_variables_optimisees_lineaire.cpp
```diff
@@ -104,11 +105,22 @@ void OPT_ConstruireLaListeDesVariablesOptimiseesDuProblemeLineaire(PROBLEME_HEBD
                   = VARIABLE_BORNEE_DES_DEUX_COTES;
                 NombreDeVariables++;
                 // 3. Level
-                CorrespondanceVarNativesVarOptim->ShortTermStorage.LevelVariable[globalIndex]
-                  = NombreDeVariables;
-                ProblemeAResoudre->TypeDeVariable[NombreDeVariables]
-                  = VARIABLE_BORNEE_DES_DEUX_COTES;
-                NombreDeVariables++;
+                if (pdt == 0 || (pdt % storage.storagecycle != 0))
```
guilpier-code, line 108: There is something I don't understand here : 
If we assume that we don't yet handle the case **initiallevel = optim** (as suggests the code reading the file **list.ini**), the initial level should not be a variable of the optimization, but should show up in the some constraint RHS.
Why then do we create a new optimization variable when **pdt == 0**, as suggests the block  `if (pdt == 0 || (pdt % storage.storagecycle != 0))` ?

For me, there only should be (still assuming that case **initiallevel = optim** is not handled) : 
```C++
if (pdt % storage.storagecycle != 0)
{
   ...
}
```
and that's it. For me, no **else** required.

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/properties.h
payetvin: Was done previously with an int by we switched to the enum

src/libs/antares/study/parts/short-term-storage/cluster.cpp
payetvin: Used to generate the cluster ID, that will be used after to load series
Validate() will be called after loading series because it need all data to check values

src/libs/antares/study/parts/short-term-storage/container.cpp
payetvin: There's no need for a file list.ini in every area, this solve the problem you encountered

payetvin: Add a comment to describre spec in code

src/solver/optimisation/opt_construction_variables_optimisees_lineaire.cpp
guilpier-code, line 108: Shouldn't we have  
```C++
if (! storage.initialLevel.has_value() || pdt % storage.storagecycle != 0)
{
    // Define a new opt variable
}
```
?
Moreover, I can't see the point of the **else** block : if we're dealing with a level at the beginning / end of a cycle, it should show up in some constraint RHS, but we don't need to define a optimization variable, right ?

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,55 @@ bool Series::validateLowerRuleCurve() const
...
+    if (start < end)
+        end += HOURS_PER_YEAR;
+
+    unsigned int simulationHour = (simplexIsWeek) ? 168 : 24;
+
+    if (level.has_value())
+    {
+        if (!checkLevelValue(level.value(), cycle, start, end) ||
+                !checkLevelValue(level.value(), simulationHour, start, end))
+            return false;
```
flomnes: `return a(...) && b(..)`

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,55 @@ bool Series::validateLowerRuleCurve() const
...
+    }
+    else
+    {
+        if (!checkLevelInterval(cycle, start, end) ||
+                !checkLevelInterval(simulationHour, start, end))
+            return false;
```
flomnes: same

src/solver/optimisation/opt_construction_variables_optimisees_lineaire.cpp
flomnes: > Moreover, I can't see the point of the else block : if we're dealing with a level at the beginning / end of a cycle, it should show up in some constraint RHS, but we don't need to define a optimization variable, right ?

If the level is fixed, we could indeed set it as the RHS. But it would add even more complexity.

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,55 @@ bool Series::validateLowerRuleCurve() const
     return checkVectBetweenZeroOne(maxInjectionModulation, "lower rule curve");
 }
 
+bool Series::validateInitialLevelSimplex(bool simplexIsWeek, std::optional<double> level,
+        unsigned int cycle, unsigned int start, unsigned int end)
+{
+    if (start < end)
+        end += HOURS_PER_YEAR;
```
flomnes: I don't understand this

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,55 @@ bool Series::validateLowerRuleCurve() const
...
+}
+
+bool Series::checkLevelInterval(unsigned int time, unsigned int start, unsigned int end) const
+{
+    double min = lowerRuleCurve[start];
+    double max = upperRuleCurve[start];
+
+    for (unsigned int h = start + time; h < end; h += time)
+    {
+        unsigned int realHour = h % HOURS_PER_YEAR;
```
flomnes: `h` should never be `> HOURS_PER_YEAR`

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,55 @@ bool Series::validateLowerRuleCurve() const
...
+    {
+        if (!checkLevelInterval(cycle, start, end) ||
+                !checkLevelInterval(simulationHour, start, end))
+            return false;
+    }
+
+    return true;
+}
+
+bool Series::checkLevelValue(double level, unsigned int time, unsigned int start, unsigned int end) const
```
flomnes: `time` is unclear, it's also the name of a C function defined in time.h

src/libs/antares/study/area/list.cpp
```diff
@@ -1081,11 +1081,15 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
 
     // Short term storage
     {
+        unsigned int start =  study.calendar.days[study.parameters.simulationDays.first].hours.first;
+        unsigned int end =  study.calendar.days[study.parameters.simulationDays.end].hours.end;
```
flomnes: `start` & `end` are unclear. Are they hours, days, weeks ?

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -63,17 +66,19 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
...
     series->fillDefaultSeriesIfEmpty(); // fill series if no file series
     return ret;
 }
 
+bool STStorageCluster::validate(bool simplexIsWeek, unsigned int start, unsigned int end)
+{
+    logs.debug() << "Validating properties and series for st storage: " << id;
+    return properties.validate(simplexIsWeek) && series->validate() &&
+        series->validateInitialLevelSimplex(simplexIsWeek, properties.initialLevel,
+            properties.storagecycle.value(), start, end);
```
flomnes: storagecycle may not have a value at this stage (in unit tests for example). Pass it (as an std::optional) instead of its value ?

flomnes (commented):
@payetvin There are a few issues with the checks, it seems that they are not sufficient

1. If the initial level is specified, check that it lies in all intervals [lower rule curve[k*cycle], upper rule curve[k*cycle]] for each week (one loop on weeks, one loop on k within this week)
2. If not, check that the intersection of the above intervals is not empty

Then @guilpier-code we'll need to use this intersection in the OPT module.

src/libs/antares/study/parts/short-term-storage/cluster.cpp
payetvin: This property is mandatory in properties.cpp:validate(), thus cannot fail

src/libs/antares/study/parts/short-term-storage/series.cpp
payetvin: Was a misktake it's > to see if the start hour is before the end hour
Antares simulation can go from october to january, if it's the case we just increment the end and the realHour var allows us to not care about this case

src/libs/antares/study/area/list.cpp
```diff
@@ -1081,11 +1081,15 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
 
     // Short term storage
     {
+        unsigned int startHour =  study.calendar.days[study.parameters.simulationDays.first].hours.first;
```
guilpier-code: `startHour` : **simulationFirstHour** would be more clear (or **simuFirstHour**)
`endHour` : **simulationLastHour** would be more clear

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -63,17 +66,19 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
...
-}
-
 bool STStorageCluster::loadSeries(const std::string& folder) const
 {
     bool ret = series->loadFromFolder(folder);
     series->fillDefaultSeriesIfEmpty(); // fill series if no file series
     return ret;
 }
 
+bool STStorageCluster::validate(bool simplexIsWeek, unsigned int startHour, unsigned int endHour)
```
guilpier-code: **Indentation** : 
This would be better : 
```C++
return properties.validate(simplexIsWeek) && 
           series->validate() &&
           series->validateInitialLevelSimplex(simplexIsWeek, 
                                               properties.initialLevel,
                                               properties.storagecycle.value(), 
                                               startHour, 
                                               endHour);
```

guilpier-code: Further more, **validateInitialLevelSimplex** could be more meaningful. 
checkInitialLevelBetweenBounds (for example) sounds better to me.

src/libs/antares/study/parts/short-term-storage/series.h
```diff
@@ -33,13 +33,15 @@ namespace Antares::Data::ShortTermStorage
...
-    // check if series values are valid
-    bool validate() const;
-
     // load all series files with folder path
     bool loadFromFolder(const std::string& folder);
     void fillDefaultSeriesIfEmpty();
 
+    // check if series values are valid
+    bool validate() const;
+    bool validateInitialLevelSimplex(bool simplexIsWeek, std::optional<double> level,
```
guilpier-code: About argument `std::optional<double> level` : why not call it **initialLevel** (it's what we're truly talking about here) ?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,60 @@ bool Series::validateLowerRuleCurve() const
     return checkVectBetweenZeroOne(maxInjectionModulation, "lower rule curve");
 }
 
+bool Series::validateInitialLevelSimplex(bool simplexIsWeek, std::optional<double> level,
+        unsigned int cycle, unsigned int startHour, unsigned int endHour)
+{
+    if (startHour > endHour)
+        endHour += HOURS_PER_YEAR;
+
+    unsigned int simulationHour = (simplexIsWeek) ? 168 : 24;
```
guilpier-code: About **simulationHour** : **optimizationRange** would be better ?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,60 @@ bool Series::validateLowerRuleCurve() const
...
+    else
+        return checkLevelInterval(cycle, startHour, endHour) &&
+                checkLevelInterval(simulationHour, startHour, endHour);
+}
+
+bool Series::checkLevelValue(double level, unsigned int cycleDuration, unsigned int startHour, unsigned int endHour) const
+{
+    for (unsigned int h = startHour; h < endHour; h += cycleDuration)
+    {
+        unsigned int realHour = h % HOURS_PER_YEAR;
```
guilpier-code: About **realHour** : **calendarHour** ?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,60 @@ bool Series::validateLowerRuleCurve() const
...
+        logs.notice() << "h: " << realHour << " lrc: " << lowerRuleCurve[realHour] <<  " level : " << level <<
```
guilpier-code: `logs.notice()` : It's surely here temporarily.
If not, this instruction will generate way too much lines in the logs. We should remove this.

src/libs/antares/study/area/list.cpp
payetvin: Need to be done, this branch is actually in 8.5 so adding this condition will cause problem right now

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,59 @@ bool Series::validateLowerRuleCurve() const
...
+
+    if (level.has_value())
+        return checkLevelValue(level.value(), cycleDuration, startHour, endHour) &&
+                checkLevelValue(level.value(), simulationHour, startHour, endHour);
+    else
+        return checkLevelInterval(cycleDuration, startHour, endHour) &&
+                checkLevelInterval(simulationHour, startHour, endHour);
+}
+
+bool Series::checkLevelValue(double level, unsigned int cycleDuration, unsigned int startHour, unsigned int endHour) const
```
guilpier-code: I'm not sure that this function **checkLevelValue(...)** does what we want : 
- Imagine we have a simulation 2 weeks long (from h = 0 to hour 2 * 168 - 1= 335 hours).
- Imagine it is a weekly optimization (over 168 hours blocks)
- Imagine we have a cycle of 100 hours.

What the code here does is : make a check on initial level at hours 0, 100, 168, 200, 300.
What we want is : make a check at hours : 0, 100, 168 and 268.

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
flomnes: For every week (separably) 
- Take the intersection of [Lower bound[h], Upper Bound[h]] for hours h=0, 100
- Check it is not empty
- Save it for the modeler structure `std::vector<BoundsForInitialLevel> bounds`
- The modeler fetches `bounds[weekIndex]` for every week, see function `SIM_RenseignementProblemeHebdo`

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,59 @@ bool Series::validateLowerRuleCurve() const
...
+        endHour += HOURS_PER_YEAR;
+
+    unsigned int simulationHour = simplexIsWeek ? 168 : 24;
+
+    if (level.has_value())
+        return checkLevelValue(level.value(), cycleDuration, startHour, endHour) &&
+                checkLevelValue(level.value(), simulationHour, startHour, endHour);
+    else
+        return checkLevelInterval(cycleDuration, startHour, endHour) &&
+                checkLevelInterval(simulationHour, startHour, endHour);
```
flomnes: Keep only one of these in each if..else

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,74 @@ bool Series::validateLowerRuleCurve() const
...
+        }
+    }
+    return true;
+}
+
+bool Series::checkLevelInterval(unsigned int cycleDuration, unsigned int simuDuration,
+        unsigned int startHour, unsigned int endHour) const
+{
+    // loop on each week or day depending on simulation mode, then on cycles in it
+    for (unsigned int simuIndex = startHour; simuIndex < endHour; simuIndex += simuDuration)
```
flomnes: `simuIndex` isn't very clear. `firstHourOfTheWeek` ?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,74 @@ bool Series::validateLowerRuleCurve() const
...
+    {
+        for (unsigned cycleHour = simuIndex; cycleHour < simuDuration; cycleHour += cycleDuration)
+        {
+            unsigned int realHour = cycleHour % HOURS_PER_YEAR;
+
+            if (upperRuleCurve[realHour] < level ||
+                    lowerRuleCurve[realHour] > level)
```
flomnes: Here you have to multiply `upperRuleCurve` & `lowerRuleCurve` by the reservoir capacity. Otherwise you'd compare dimensionless quantities with MWh, that makes no sense.

src/libs/antares/study/parts/short-term-storage/series.cpp
payetvin: Applied, I didn't use the word week as the opti duration can either be weekly or daily

src/libs/antares/study/parts/short-term-storage/series.cpp
payetvin: Should i pass all the properties to the function ? There's now 3 args that depends on properties: initial level, cycle duration and reservoir capacity

src/solver/simulation/sim_structure_probleme_economique.h
payetvin: std::optional\<unsigned int\>

flomnes (commented):
Check that the initialLevel is multiplied by the reservoir capacity in OPT as well

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp:202

flomnes (commented):
Normally, all indices remain < HOURS_PER_YEAR

So please consider removing `% HOURS_PER_YEAR`

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,116 @@ bool Series::validateLowerRuleCurve() const
...
+bool Series::checkLevelValue(double initialLevel, unsigned int cycleDuration,
+        unsigned int optimizationRange, unsigned int simuFirstHour, unsigned int simuLastHour) const
+{
+    // loop on each week or day depending on simulation mode, then on cycles in it
+    for (unsigned int firstHourOfTheWeek = simuFirstHour; firstHourOfTheWeek < simuLastHour;
+            firstHourOfTheWeek += optimizationRange)
+    {
+        for (unsigned int cycleHour = 0; cycleHour < optimizationRange; cycleHour += cycleDuration)
+        {
+            unsigned int calendarHour = (firstHourOfTheWeek + cycleHour) % HOURS_PER_YEAR;
```
flomnes: `for (unsigned calendarHour = firstHourOfTheWeek; calendarHour < firstHourOfTheWeek + optimizationRange; calendarHour += cycleDuration)`

flomnes: You don't really need `cycleHour`. This holds also for `checkLevelInterval`

src/solver/simulation/sim_structure_probleme_economique.h
payetvin: It's actually not optional and should be an unsigned int, changed accordingly

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +180,120 @@ bool Series::validateLowerRuleCurve() const
     return checkVectBetweenZeroOne(maxInjectionModulation, "lower rule curve");
 }
 
+bool Series::validateInflowsSums(bool simplexIsWeek, unsigned int cycleDuration,
+        unsigned int simuFirstHour, unsigned int simuLastHour) const
+{
+    // allows to loop on hours even if simulation starts in november and ends in january or later
```
flomnes: Please don't loop inside this function.

Add a function for 1 week here, call it in `Area::checkWeeklyData`

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -296,4 +296,52 @@ bool Series::checkLevelInterval(unsigned int cycleDuration, unsigned int optimiz
     return true;
 }
 
+bool Series::validateCycle(unsigned int firstHourOfTheWeek, std::optional<double> initialLevel,
+        unsigned int cycleDuration) const
+{
+    if (initialLevel.has_value())
```
flomnes: Please add a comment like "Check that the initial level is inside the rule curves at all cycle timesteps"

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -296,4 +296,52 @@ bool Series::checkLevelInterval(unsigned int cycleDuration, unsigned int optimiz
...
+                logs.warning() << "Error at line: " << hour + 1 << " for sts series upper"
+                    << " or lower rule curves, values at the start of the cycle are outside of" <<
+                    " those values";
+
+                return false;
+            }
+
+            // reduce the interval if necessary
+            minBase = std::max(minBase, minCycle);
+            maxBase = std::min(maxBase, maxCycle);
```
flomnes: You could simplify this algorithm a bit
1. Compute `b=`the min of all `upperRuleCurve[hour]` for all hours in the cycle set
2. Compute `a=`the max of all `lowerRuleCurve[hour]` for all hours in the cycle set
3. Check that `a < b` 

You could extract 1. and 2. into a `getBoundsForInitialLevel(uint firstHourOfTheWeek)` method, so that it can be re-used by the modeler.

flomnes: You can initialize the min with [`std::numeric_limits<double>::infinity()`](https://en.cppreference.com/w/cpp/types/numeric_limits/infinity) and the max with `-std::numeric_limits<double>::infinity()`

src/libs/antares/study/area/list.cpp
```diff
@@ -937,12 +937,17 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
     }
 
     // Short term storage
+    if (study.header.version >= 860)
     {
+        unsigned int startHour =  study.calendar.days[study.parameters.simulationDays.first].hours.first;
+        unsigned int endHour =  study.calendar.days[study.parameters.simulationDays.end].hours.end;
+        bool simplexIsWeek = (study.parameters.simplexOptimizationRange == sorWeek);
```
flomnes: You don't need to pass `startHour` and `endHour`. Furthermore, these may not be initialized at this moment.

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -316,32 +238,35 @@ bool Series::validateCycle(unsigned int firstHourOfTheWeek, std::optional<double
...
-            // reduce the interval if necessary
-            minBase = std::max(minBase, minCycle);
-            maxBase = std::min(maxBase, maxCycle);
         }
-
     }
     return true;
 }
 
+std::tuple<double, double> Series::getBoundsForInitialLevel(unsigned int firstHourOfTheWeek,
```
flomnes: Create a class with upper bound & lower bound

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,47 +179,41 @@ bool Series::validateLowerRuleCurve() const
...
+            cycleHour += cycleDuration)
     {
-        for (unsigned int cycleHour = 0; cycleHour < optimizationRange; cycleHour += cycleDuration)
+        double sumInflows = 0.0;
+        double sumInjection = 0.0;
+        double sumWithdrawal = 0.0;
+
+        // sum until end of cycle or end of week
+        for (unsigned int i = 0;
+                i < cycleDuration && i + cycleHour < Antares::Constants::nbHoursInAWeek; i++)
```
```suggestion
        for (unsigned int hourInCycle = 0;
                hourInCycle < cycleDuration && hourInCycle + cycleHour < Antares::Constants::nbHoursInAWeek; hourInCycle++)
```

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,47 +179,41 @@ bool Series::validateLowerRuleCurve() const
...
+            sumWithdrawal += maxWithdrawalModulation[calendarHour];
+        }
 
-                sumInflows += inflows[calendarHour];
-                sumInjection += maxInjectionModulation[calendarHour];
-                sumWithdrawal += maxWithdrawalModulation[calendarHour];
-            }
+        if (sumInjection > sumInflows || sumWithdrawal < sumInflows )
+        {
+            logs.warning() << "Error at week: " << firstHourOfTheWeek + 1 << " for sts series "
```
flomnes: short-term storage

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,47 +179,41 @@ bool Series::validateLowerRuleCurve() const
...
-    if (simuFirstHour > simuLastHour)
-        simuLastHour += HOURS_PER_YEAR;
-
-    unsigned int optimizationRange = simplexIsWeek ? 168 : 24;
-
-    // loop on each week or day depending on simulation mode, then on cycles in it
-    for (unsigned int firstHourOfTheWeek = simuFirstHour; firstHourOfTheWeek < simuLastHour;
-            firstHourOfTheWeek += optimizationRange)
+    for (unsigned int cycleHour = 0; cycleHour < Antares::Constants::nbHoursInAWeek;
+            cycleHour += cycleDuration)
```
```suggestion
    for (unsigned int firstHourOfTheCycle = 0; firstHourOfTheCycle < Antares::Constants::nbHoursInAWeek;
            firstHourOfTheCycle += cycleDuration)
```

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,87 @@ bool Series::validateLowerRuleCurve() const
...
+        unsigned int cycleDuration) const
+{
+    // Check that the initial level is inside the rule curves at all cycle timesteps
+    if (initialLevel.has_value())
+    {
+        for (unsigned int hour = firstHourOfTheWeek;
+                hour < firstHourOfTheWeek + Antares::Constants::nbHoursInAWeek;
+                hour += cycleDuration)
+        {
+            if (upperRuleCurve[hour] < initialLevel || lowerRuleCurve[hour] > initialLevel)
```
flomnes: You could have 2 checks to have more specific logs (error in lower rule curve, error in upper rule curve)

Also, make sure to log "short-term storage", not sts

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,87 @@ bool Series::validateLowerRuleCurve() const
...
+            {
+                logs.warning() << "Error at line: " << hour + 1 << " for sts series upper or " <<
+                    "lower rule curves, initial level is not between those values";
```
```suggestion
                logs.warning() << "Error at hour: " << hour  << " for sts series upper or " <<
                    "lower rule curves, initial level is not between those values";
```

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,87 @@ bool Series::validateLowerRuleCurve() const
...
+
+                return false;
+            }
+        }
+    }
+    else
+    {
+        auto bounds = getBoundsForInitialLevel(firstHourOfTheWeek, cycleDuration);
```
flomnes, line 261: Excellent, @guilpier-code you should be able to use this function in SIM

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,87 @@ bool Series::validateLowerRuleCurve() const
...
+
+std::tuple<double, double> Series::getBoundsForInitialLevel(unsigned int firstHourOfTheWeek,
+        unsigned int cycleDuration) const
+{
+    double minCycle = lowerRuleCurve[firstHourOfTheWeek];
+    double maxCycle = upperRuleCurve[firstHourOfTheWeek];
+
+    for (unsigned int hour = firstHourOfTheWeek + cycleDuration;
+            hour < firstHourOfTheWeek + Antares::Constants::nbHoursInAWeek;
+            hour += cycleDuration)
```
```suggestion
    double minCycle = lowerRuleCurve[firstHourOfTheWeek];
    double maxCycle = upperRuleCurve[firstHourOfTheWeek];

    for (unsigned int hour = firstHourOfTheWeek;
            hour < firstHourOfTheWeek + Antares::Constants::nbHoursInAWeek;
            hour += cycleDuration)
```

You might an out of range error otherwise, if the 1st index is out of bounds (unlikely, but still)

flomnes (commented):
Please create a Bounds struct, with lower & upper bounds.

So you don't need to guess that the 1st element is the lower bound and the 2nd one is the upper bound.

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -201,8 +204,8 @@ bool Series::validateInflowsSums(unsigned int firstHourOfTheWeek, unsigned int c
 
         if (sumInjection > sumInflows || sumWithdrawal < sumInflows )
         {
-            logs.warning() << "Error at week: " << firstHourOfTheWeek + 1 << " for sts series "
-                "inflows, sums at cycle timesteps are wrong";
+            logs.warning() << "Error at week: " << firstHourOfTheWeek + 1 << " for short term "
+                "storage inflows, sums at cycle timesteps are wrong";
```
flomnes: The week number is wrong here (it's an hour number). You need to log an hour instead

src/libs/antares/study/parts/short-term-storage/container.h
```diff
@@ -35,11 +35,14 @@ namespace Antares::Data::ShortTermStorage
 class STStorageInput
```
guilpier-code, line 35: This file contains the definition of one class only. So the source file name and the class name should be close.
Besides, there are many files named **container.h/cpp** in the base code.
I suggest we name this file differently, more specific than **container**.

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,98 @@ bool Series::validateLowerRuleCurve() const
...
+}
+
+bool Series::validateCycle(unsigned int firstHourOfTheWeek, std::optional<double> initialLevel,
+        unsigned int cycleDuration) const
+{
+    // Check that the initial level is inside the rule curves at all cycle timesteps
+    if (initialLevel.has_value())
+    {
+        for (unsigned int hour = firstHourOfTheWeek;
+                hour < firstHourOfTheWeek + Antares::Constants::nbHoursInAWeek;
```
guilpier-code: `Antares::Constants::nbHoursInAWeek` : `Constants::nbHoursInAWeek` should be enough.
Here and elsewhere.

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,98 @@ bool Series::validateLowerRuleCurve() const
...
+                return false;
+        }
+    }
+    return true;
+}
+
+Bounds Series::getBoundsForInitialLevel(unsigned int firstHourOfTheWeek,
+        unsigned int cycleDuration) const
+{
+    double minCycle = lowerRuleCurve[firstHourOfTheWeek];
```
guilpier-code: **minCycle** : 
Does not mean anything to me in this context.
Should renamed **maxLowerBound**.

**maxCycle** ==> **minUpperBound**

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -274,4 +274,87 @@ BOOST_AUTO_TEST_CASE(check_container_properties_empty_file)
...
+BOOST_AUTO_TEST_CASE(check_series_interval_lower)
+{
+    createFileSeries(0.5, 8760);
+
+    BOOST_CHECK(series.loadFromFolder(folder));
+    BOOST_CHECK(series.validate());
+
+    series.lowerRuleCurve[120] = 0.6;
+
+    BOOST_CHECK(!series.validateCycle(100, std::optional<double>(), 20));
```
flomnes:
```suggestion
    BOOST_CHECK(!series.validateCycle(100, std::nullopt, 20));
```
see https://en.cppreference.com/w/cpp/utility/optional/nullopt

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -274,4 +274,87 @@ BOOST_AUTO_TEST_CASE(check_container_properties_empty_file)
...
+BOOST_AUTO_TEST_CASE(check_series_upper_curve)
+{
+    createFileSeries(0.5, 8760);
+
+    BOOST_CHECK(series.loadFromFolder(folder));
+    BOOST_CHECK(series.validate());
+    //624 is third week first cycle (100 + 168*3 + 20)
+    series.upperRuleCurve[190] = 0.2;
+
+    BOOST_CHECK(!series.validateCycle(100, std::optional<double>(0.5), 45));
```
flomnes:
```suggestion
    BOOST_CHECK(!series.validateCycle(100, 0.5, 45));
```
`std::optional<T>` has a conversion operator from `T`, so you don't need to explicitly specify it.

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
```diff
@@ -274,4 +274,87 @@ BOOST_AUTO_TEST_CASE(check_container_properties_empty_file)
...
+BOOST_AUTO_TEST_CASE(check_series_lower_curve)
+{
+    createFileSeries(0.5, 8760);
+
+    BOOST_CHECK(series.loadFromFolder(folder));
+    BOOST_CHECK(series.validate());
+
+    series.lowerRuleCurve[120] = 0.6;
+
+    BOOST_CHECK(!series.validateCycle(100, std::optional<double>(0.5), 20));
```
flomnes:
```suggestion
    BOOST_CHECK(!series.validateCycle(100, 0.5, 20));
```

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -63,17 +66,26 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
...
     return ret;
 }
 
+bool STStorageCluster::validate(bool simplexIsWeek)
+{
+    logs.debug() << "Validating properties and series for st storage: " << id;
+    return properties.validate(simplexIsWeek) && series->validate();
+}
+
+bool STStorageCluster::validateCycle(unsigned int firstHourOfTheWeek) const
```
flomnes:
```suggestion
bool STStorageCluster::validateWeeklyTimeseries(unsigned int firstHourOfTheWeek) const
```

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,98 @@ bool Series::validateLowerRuleCurve() const
...
+                hourInCycle++)
+        {
+            calendarHour = firstHourOfTheWeek + firstHourOfTheCycle + hourInCycle;
+
+            sumInflows += inflows[calendarHour];
+            sumInjection += maxInjectionModulation[calendarHour];
+            sumWithdrawal += maxWithdrawalModulation[calendarHour];
+        }
+
+        if (sumInjection > sumInflows || sumWithdrawal < sumInflows )
```
guilpier-code: We have a unity problem here : **inflows** in MW and **injections / withdrawals** in % (inside [0, 1]).

src/libs/antares/study/area/area.cpp
```diff
@@ -136,6 +136,7 @@ Area::~Area()
 bool Area::checkWeeklyData(uint firstHourOfTheWeek) const
```
guilpier-code, line 136: This function **Area::checkWeeklyData(...)** is called from function **postCalendarLoadChecks(...)**, in file **load.cpp**. 
**Area::checkWeeklyData(...)** makes some checks on ST storage cycles and inflows.

At the same code level (**load.cpp**), we call `areas.loadFromFolder(options); `and make other checks on ST storage time series.

**Question** : why are these checks about cycles and inflows so different from other checks on time series that we need to call them separately ?

src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -52,20 +52,20 @@ unsigned int groupIndex(Group group);
...
+    // Initial level, between 0 and 1
     std::optional<double> initialLevel;
     // Efficiency factor between 0 and 1
     double efficiencyFactor = 1;
-    // Cycle duration, 1 <= cycleDuration <= 168
-    unsigned int cycleDuration;
+    // Not optional   Cycle duration, 1 <= storagecycle <= 168
+    std::optional<unsigned int> storagecycle;
     // Used to sort outputs
     Group group = Group::Other1;
```
guilpier-code, line 70: map **ST_STORAGE_PROPERTY_GROUP_ENUM** is used only inside **Properties** class, we should make it **private**.

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -70,25 +70,40 @@ unsigned int groupIndex(Group group)
...
-    };
+    }
+    return true;
+}
 
+bool Properties::loadKey(const IniFile::Property* p)
+{
     if (p->key == "injectionnominalcapacity")
-        return valueForOptional(this->injectionCapacity);
+        return valueForOptional(p, this->injectionCapacity);
```
guilpier-code: Take the example of **injectionCapacity** : 
At some point we have : 
```C++
if (p->key == "injectionnominalcapacity")
    return valueForOptional(p, this->injectionCapacity);
```
and then : 
```C+
if (!checkMandatory(name, injectionCapacity, "injectionnominalcapacity"))
    return false;
```
**Question** : isn't that too complicated ?
Couldn't we have : 
```C++
if (p->key == "injectionnominalcapacity")
    return p->value.to<std::optional<double>>(this->injectionCapacity);
...
if (!injectionCapacity.has_value())
{
    // Some logs here
    return false;
}
```
? 
In other words, **if I'm not missing something**, we could remove functions **valueForOptional(...)** and **checkMandatory(...)**

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -183,14 +193,25 @@ bool Properties::validate()
...
         {
-            logs.warning() << "initiallevel for cluster: " << name
-                           << " should be inferior to reservoir capacity: " << capacity.value();
-            initialLevel = capacity;
+            logs.warning() << "initiallevel for cluster: " << name << " should be inferior to 1";
+            initialLevel = 1;
         }
     }
 
+    auto checkSimplexAndCycle = [&](unsigned int maxValue) {
```
guilpier-code: renaming **maxValue** ==>  **nbHoursInOptim** ?

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/container.h
payetvin: The name container is used for other types of productions in src/libs/antares/study/parts/

Changing the name would be appropriate but would break consistency 
@flomnes what's the best option ?

src/libs/antares/study/parts/short-term-storage/properties.cpp
payetvin: `return p->value.to<std::optional<double>>(this->injectionCapacity);`
Applied this

Should we simply log instead of using checkMandatory() ? @flomnes

src/libs/antares/study/area/area.cpp
payetvin: @flomnes should wee do all checks on sts series after calendar loading ?

src/libs/antares/study/parts/short-term-storage/properties.cpp
guilpier-code: @payetvin : you told me that 
```C++
if (p->key == "injectionnominalcapacity")
    return p->value.to<std::optional<double>>(this->injectionCapacity);
...
```
make unit tests fail at execution.
Could you try this : 
```C++
if (p->key == "injectionnominalcapacity")
    return p->value.to<double>(*this->injectionCapacity); // Mind the use of * dereference operator
...
```
?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,97 @@ bool Series::validateLowerRuleCurve() const
...
+            logs.warning() << "Error at end of cycle: " << calendarHour + 1 << " for short term "
+                "storage inflows, sums at cycle timesteps are wrong";
+
+            return false;
+        }
+    }
+    return true;
+}
+
+bool Series::validateCycle(unsigned int firstHourOfTheWeek, std::optional<double> initialLevel,
```
guilpier-code: validateCycle(...) : could it be renamed into **validateCycleForWeek(...)** ?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,97 @@ bool Series::validateLowerRuleCurve() const
     return checkVectBetweenZeroOne(maxInjectionModulation, "lower rule curve");
 }
 
+bool Series::validateInflowsSums(unsigned int firstHourOfTheWeek, unsigned int cycleDuration,
```
guilpier-code: validateInflowsSums(...) : rename it into **validateInflowsForWeek(...)** ?

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/properties.cpp
payetvin: Same results, injectionCapacity.has_value() is false

src/libs/antares/study/area/list.cpp
```diff
@@ -937,12 +937,15 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
     }
 
     // Short term storage
+    if (study.header.version >= 860)
```
guilpier-code, line 940: We load input data for ST storage clusters at 2 different places : 
- we create the cluster in one place
- we load the time series in another place

Why is that ?
Maybe we want to do as we do for loading **thermal clusters** for example.
But I'm not sure it is done properly regarding **thermal clusters**.

We should load ST storage clusters and time series in one place, for clarity.

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,97 @@ bool Series::validateLowerRuleCurve() const
...
+                hourInCycle < cycleDuration
+                    && hourInCycle + firstHourOfTheCycle < Constants::nbHoursInAWeek;
+                hourInCycle++)
+        {
+            calendarHour = firstHourOfTheWeek + firstHourOfTheCycle + hourInCycle;
+
+            sumInflows += inflows[calendarHour];
+            // multiply by capacity to have the same unit as inflows: MWh
+            sumInjection += maxInjectionModulation[calendarHour] * capacity;
+            sumWithdrawal += maxWithdrawalModulation[calendarHour] * capacity;
```
flomnes: Careful, these modulations must be multiplied by the injection/withdrawal capacity respectively.

Multiplying by the reservoir capacity makes no sense.

src/libs/antares/study/parts/short-term-storage/series.h
```diff
@@ -27,33 +27,53 @@
 #pragma once
 #include <vector>
 #include <string>
+#include <tuple>
```
flomnes:
```suggestion
```

src/libs/antares/study/parts/short-term-storage/properties.cpp
guilpier-code: ok, the unit tests fail because : 
- when using `p->value.to<std::optional<double>>(this->injectionCapacity);` : there is no template specialization of the `to<T>` for **std::optional<double>**, so a default specialization is used, which does nothing.
- when using `p->value.to<double>(*this->injectionCapacity);`, the internal value of the **std::optional** is changed correctly, but not the internal status bool **_has_value_** which remains to **false**

I tried to find a quick and clean solution to this, in file **src/ext/yuni/src/yuni/core/string/traits/into.h**, but I can't see any.
So I give up.

src/libs/antares/study/area/area.cpp
flomnes: > Question : why are these checks about cycles and inflows so different from other checks on time series that we need to call them separately ?

Because when loading other timeseries, the calendar is not yet loaded. Hence, we do the following

1. Load all timeseries
2. Perform most checks
3. Load the calendar
4. Perform the checks that require the calendar

src/libs/antares/study/area/list.cpp
guilpier-code: Eventually this problem was solved.
@flomnes : thx for your precious technical help !

src/libs/antares/study/parts/short-term-storage/properties.cpp
guilpier-code: Eventually this problem was solved.
@flomnes : thx for your precious technical help !

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -180,4 +179,99 @@ bool Series::validateLowerRuleCurve() const
     return checkVectBetweenZeroOne(maxInjectionModulation, "lower rule curve");
```
guilpier-code:
```cpp
bool Series::validateLowerRuleCurve() const
{
    return checkVectBetweenZeroOne(maxInjectionModulation, "lower rule curve");
}
```
Careful here : we validate **maxInjectionModulation**, not lower rule curve

guilpier-code (changes requested):

src/libs/antares/study/area/list.cpp
payetvin: We could but this will indeed break consistency with other clusters like thermal, where series and prepro are loaded in a different function

PR #1242, opened 2023-03-29
Code smells for short term storage

payetvin:


src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -58,7 +58,7 @@ bool Series::loadFromFolder(const std::string& folder)
     return ret;
 }
 
-bool Series::loadVector(const std::string& path, std::vector<double>& vect)
+bool Series::loadVector(const std::string& path, std::vector<double>& vect) const
```
flomnes, line 61: This function could be a free function

src/solver/variable/commons/links/links.h.inc.hxx
```diff
@@ -165,7 +165,7 @@ public:
     void hourBegin(uint hourInTheYear);
     void hourForEachArea(State& state, uint numSpace);
     void hourForEachLink(State& state, uint numSpace);
-    void hourForClusters(State& state, uint numSpace) const;
+    void hourForClusters(const State& state, uint numSpace) const;
```
flomnes, line 168: This function will be removed soon

src/libs/antares/study/parts/short-term-storage/series.cpp
payetvin: will be done in feature/sts-tests

PR #1243, opened 2023-03-30
Fix/remove old versions

### LINKED ISSUE #1229: Remove version-specific code for version < 700
Also
- `internalLoadFormatBefore33`
- `TransformNameIntoNameBefore390`

payetvin:
close #1229 

- [x] Need to remove old properties loading from inifile


src/libs/antares/array/correlation.cpp
```diff
@@ -403,7 +403,7 @@ void Correlation::internalSaveToINI(Study& study, IO::File::Stream& file) const
         logs.error() << correlationName << ": the montlhy correlation coefficients are missing";
 }
 
-bool Correlation::internalLoadFromINIPost32(Study& study, const IniFile& ini, bool warnings)
+bool Correlation::internalLoadFromINITry(Study& study, const IniFile& ini, bool warnings)
```
flomnes, line 406: Is this used ? Why is there also `Correlation::internalLoadFromINI` ?

flomnes (commented):
Check that the application returns an error if header.version < 700

flomnes (changes requested):
- Fix segfault occurring on older versions
- Add the following error message "Studies in version <7.0 are no longer supported. Please upgrade it first"

src/libs/antares/array/correlation.cpp
payetvin: internalLoadFromINI calls internalLoadFromINITry and fallback if the loading failed

src/libs/antares/study/version.cpp
```diff
@@ -171,12 +171,9 @@ Version VersionIntToVersion(uint version)
     case versionUnknown:
         return versionUnknown;
     }
-#ifndef NDEBUG
-    if (version <= (uint)versionLatest)
-    {
-        assert(false and "missing switch entry");
-    }
-#endif
+    logs.error() << "Version not found";
```
flomnes: `default:`

src/libs/antares/study/header.cpp
```diff
@@ -165,7 +165,7 @@ bool StudyHeader::internalLoadFromINIFile(const IniFile& ini, bool warnings)
             logs.error() << "The main section has not been found. The study seems invalid.";
     }
 
-    if (version >= 200 || version == 2)
+    if (version >= 700 || version == 7)
```
flomnes: `version` should never be equal to 7, please remove the 2nd condition

src/libs/antares/study/version.cpp
```diff
@@ -296,74 +167,18 @@ Version VersionIntToVersion(uint version)
...
         return versionUnknown;
-    }
-#ifndef NDEBUG
-    if (version > 100 and version <= (uint)versionLatest)
-    {
-        assert(false and "missing switch entry");
-    }
-#endif
+    default:
+    logs.error() << "Version not found";
```
```suggestion
    logs.error() << "Study version " << version << " is not supported by this version of antares-solver";
    
```

flomnes (changes requested):
Study loads OK with v=700
The error message is a bit laconic for v=600, I propose a change

flomnes (approved):

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -132,235 +132,61 @@ bool PartHydro::LoadFromFolder(Study& study, const AnyString& folder)
...
-        }
-        else
-        {
-            area.hydro.inflowPattern.reset(1, DAYS_PER_YEAR, true);
-            area.hydro.inflowPattern.fillColumn(0, 1.0);
-            area.hydro.inflowPattern.markAsModified();
-        }
+        buffer.clear() << folder << SEP << "common" << SEP << "capacity" << SEP << "reservoir_"
+                       << area.id << '.' << study.inputExtension;
+        ret = area.hydro.reservoirLevel.loadFromCSVFile(
```
guilpier-code, line 174: For reservoir levels loading, block : 
```C++
if (JIT::enabled)
{
    ...
}
...
if (enabledModeIsChanged)
     JIT::enabled = true;
```
disappeared. 
Is it normal ?

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -132,235 +132,61 @@ bool PartHydro::LoadFromFolder(Study& study, const AnyString& folder)
...
-                           << area.id << '.' << study.inputExtension;
-            ret = area.hydro.reservoirLevel.loadFromCSVFile(
-                    buffer, 3, 12, Matrix<>::optFixedSize, &study.dataBuffer)
-                  && ret;
-
-            double temp[3][DAYS_PER_YEAR];
-            uint firstDayMonth[13];
-            uint daysPerMonthDecals[12];
 
-            DataSeriesHydro::AdjustMonth(study, firstDayMonth, daysPerMonthDecals);
```
guilpier-code, line 288: If we don't use **AdjustMonth(...)** anymore, its declaration / definition should be removed from code

src/libs/antares/study/version.cpp
```diff
@@ -296,74 +167,19 @@ Version VersionIntToVersion(uint version)
...
         return versionUnknown;
-    }
-#ifndef NDEBUG
-    if (version > 100 and version <= (uint)versionLatest)
-    {
-        assert(false and "missing switch entry");
-    }
-#endif
+    default:
+    logs.error() << "Study version " << version << " is not supported by this version of antares-solver";
```
guilpier-code, line 174: Please indent code inside **default** case

src/libs/antares/study/area/links.h
```diff
@@ -129,8 +129,6 @@ class AreaLink final : public Yuni::NonCopyable<AreaLink>
     void overrideTransmissionCapacityAccordingToGlobalParameter(GlobalTransmissionCapacities tc);
 
 private:
-    bool linkLoadTimeSeries_for_version_under_320(const AnyString& folder, Study& study);
-    bool linkLoadTimeSeries_for_version_from_320_to_630(const AnyString& folder);
     bool linkLoadTimeSeries_for_version_from_630_to_810(const AnyString& folder);
```
guilpier-code, line 132: As study version below 700 are not accepted anymore :  
**linkLoadTimeSeries_for_version_from_630_to_810**(...)
should be renamed into : 
**linkLoadTimeSeries_for_version_below_810**(...)
don't you think ? (no reference to **630**, we don't care about anymore, and which is now confusing)

src/libs/antares/study/area/list.cpp
```diff
@@ -1090,110 +947,71 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
     // Nodal Optimization
-    if (study.header.version >= 330)
-    {
-        buffer.clear() << study.folderInput << SEP << "areas" << SEP << area.id << SEP
-                       << "optimization.ini";
-        IniFile ini;
-        if (ini.open(buffer))
+    buffer.clear() << study.folderInput << SEP << "areas" << SEP << area.id << SEP
+                   << "optimization.ini";
+    IniFile ini;
```
guilpier-code, line 952: It should be : 
```C++
if (!ini.open(buffer))
    return false;
```
or : 
```C++
if (!ini.open(buffer))
    ret = false
```
if we return **ret**, it could be **true**. This is not what we want, right ?

src/libs/antares/study/cleaner/cleaner-v20.cpp
```diff
@@ -238,7 +238,6 @@ static void PreflightVersion20_interco(PathList& p, const Area* area, StringT& b
 bool PreflightVersion20(StudyCleaningInfos* infos)
 {
     auto* study = new Study();
-    study->inputExtension = "txt";
```
guilpier-code, line 241: `study->inputExtension = "txt";`
Why remove this line ?
Is it useless ? Or is it done eslewhere ?

src/libs/antares/study/parts/hydro/container.cpp
flomnes:
```cpp
if (study.header.version >= 620)
{
    // ...
}
else if (study.header.version >= 390)
{
    bool enabledModeIsChanged = false;
    if (JIT::enabled)
    {
        JIT::enabled = false; // Allowing to read the area's daily max power
        enabledModeIsChanged = true;
    }
    buffer.clear() << folder << SEP << "common" << SEP << "capacity" << SEP << "reservoir_"
                   << area.id << '.' << study.inputExtension;
    ret = area.hydro.reservoirLevel.loadFromCSVFile(
            buffer, 3, 12, Matrix<>::optFixedSize, &study.dataBuffer)
          && ret;
    // ...
}
```
I think that this code block was only executed in versions 390-610, so it was removed with the rest.

src/libs/antares/study/parts/hydro/container.cpp
flomnes: I agree !

src/libs/antares/study/cleaner/cleaner.cpp
```diff
@@ -84,7 +79,7 @@ bool StudyCleaningInfos::analyze()
     }
     default:
     {
-        if ((int)version >= (int)version200 && (int)version <= (int)versionLatest)
+        if ((int)version <= (int)versionLatest)
         {
             if (not PreflightVersion20(this))
```
guilpier-code, line 84: **PreflightVersion20** does not mean anything : 
- **Version20** is not taken into account
- **Preflight** does not mean anything to me.

A bit of context : 
We're in the study cleaner : when run on a study directory, it removes all files or folders that should not be in.

This function makes the list of what should be kept, that is the list of what should be excluded from the remove operation.

So we should now name it (for instance) : **listOfFilesAnDirectoriesToKeep(...)**
Note that there are "preflight" functions call inside **PreflightVersion20** : one for an area, one for a link.
This function should be renamed too.

src/libs/antares/study/parameters.cpp
```diff
@@ -973,13 +929,10 @@ static bool SGDIntLoadFamily_Legacy(Parameters& d,
...
-    if (key == "thresholdmin")
-        return true; // value.to<int>(d.thresholdMinimum);
-    if (key == "thresholdmax")
-        return true; // value.to<int>(d.thresholdMaximum);
+    // deprecated but needed for testing old studies
     if (key == "include-split-exported-mps")
         return true;
+
     return false;
 }
```
guilpier-code, line 938: About block : 
```C++
if (not handleAllKeysInSection(*this, p->key, value, p->value, version))
{
    if (not SGDIntLoadFamily_Legacy(*this, p->key, value, p->value, version))
    {
          // Continue on error
          logs.warning() << ini.filename() << ": '" << p->key << "': Unknown property";
    }
}
```
See below, line 1002.
**handleAllKeysInSection** is a pointer to a function, it points on possibly many functions, see line see line 959.
Now we removed occurrences of **version** from this source, these functions don't need their last argument.
Only **SGDIntLoadFamily_Legacy**  should keep this argument.

src/libs/antares/study/area/list.cpp
payetvin: I kept the previous behavior, now changed it to false to follow your comment

src/libs/antares/study/cleaner/cleaner-v20.cpp
payetvin: done elsewhere in src/libs/antares/study/study.h:700

PR #1246, opened 2023-04-05
Validation of inputs for sts

payetvin:
More detailed validation for series values

### DIFF
--- a/src/libs/antares/study/parts/short-term-storage/cluster.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/cluster.cpp
@@ -65,6 +65,7 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
 
 bool STStorageCluster::validate()
 {
+    logs.debug() << "Validating properties and series for st storage: " << id;
     return properties.validate() && series->validate();
 }
 
--- a/src/libs/antares/study/parts/short-term-storage/cluster.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/cluster.cpp
@@ -75,4 +75,9 @@ bool STStorageCluster::loadSeries(const std::string& folder) const
     return ret;
 }
 
+bool STStorageCluster::validateInitialLevel()
+{
+    return true;
+}
+
 } // namespace Antares::Data::ShortTermStorage
--- a/src/libs/antares/study/parts/short-term-storage/cluster.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/cluster.cpp
@@ -66,7 +66,7 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
 bool STStorageCluster::validate()
 {
     logs.debug() << "Validating properties and series for st storage: " << id;
-    return properties.validate() && series->validate() && series->validateRuleCurve();
+    return properties.validate() && series->validate();
 }
 
 bool STStorageCluster::loadSeries(const std::string& folder) const
@@ -74,11 +74,6 @@ bool STStorageCluster::loadSeries(const std::string& folder) const
     bool ret = series->loadFromFolder(folder);
     series->fillDefaultSeriesIfEmpty(); // fill series if no file series
     return ret;
-}
-
-bool STStorageCluster::validateInitialLevel()
-{
-    return true;
 }
 
 } // namespace Antares::Data::ShortTermStorage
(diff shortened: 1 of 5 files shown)


src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -144,4 +115,46 @@ void Series::fillDefaultSeriesIfEmpty()
...
+{
+    if (maxInjectionModulation.size() != HOURS_PER_YEAR
+        || maxWithdrawalModulation.size() != HOURS_PER_YEAR || inflows.size() != HOURS_PER_YEAR
+        || lowerRuleCurve.size() != HOURS_PER_YEAR || upperRuleCurve.size() != HOURS_PER_YEAR)
+    {
+        logs.warning() << "Size of series for short term storage is wrong";
+        return false;
+    }
+
+    auto checkVectBetween = [](const std::vector<double>& v) {
```
guilpier-code: About lambda function **checkVectBetween** : 
It should be renamed **checkTimeSeriesBetweenBounds**, more clear to me, and the bounds should appear as arguments : 
```C++
auto checkTimeSeriesBetweenBounds = [](const std::vector<double>& ts, double min, double max) {
    return std::all_of(v.begin(), v.end(), [](double d){ return (d >= min && d <= max); });
    };
```
Instead of being defined inside **validate()**, it could be a free **static** function of the current source file **series.cpp**, so it could be called easily in class **Series**'s private member function (see further remarks [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1246#discussion_r1159454369))

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -65,7 +65,8 @@ bool STStorageCluster::loadFromSection(const IniFile::Section& section)
 
 bool STStorageCluster::validate()
 {
-    return properties.validate() && series->validate();
+    logs.debug() << "Validating properties and series for st storage: " << id;
```
guilpier-code, line 68: As rule curves are time series, the validation added here (the one done in member function **validateRuleCurve()**), should be done in **series->validate()**.
So, to me, there is no need for a new **public** member function **validateRuleCurve()**

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -75,4 +76,9 @@ bool STStorageCluster::loadSeries(const std::string& folder) const
     return ret;
 }
 
+bool STStorageCluster::validateInitialLevel()
```
guilpier-code: About `STStorageCluster::validateInitialLevel()` member function :
As you know, the initial level is a cluster property.
A lot of cluster properties are checked in **Properties::validate()**, and so is the initial level, though it remains some checks to do on the  initial level.
So I'm wondering about this **STStorageCluster::validateInitialLevel()** function : what will it do exactly ?

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -144,4 +115,46 @@ void Series::fillDefaultSeriesIfEmpty()
...
+    if (!checkVectBetween(lowerRuleCurve) || !checkVectBetween(upperRuleCurve))
+    {
+        logs.warning() << "Values for rule curve series should be between 0 and 1";
+        return false;
+    }
+
+    return true;
+}
+
+bool Series::validateRuleCurve() const
```
guilpier-code: As you see, there are already some checks on rule curves done in **Series::validate()**.
And now we make new checks on rule curves in another function **Series::validateRuleCurve()**.
What I recommend : 
Keep a public **Series::validate()** member function and add a new private validate functions per kind of  time series
```C++
bool Series::validate() const
{
    // General
    if (maxInjectionModulation.size() != HOURS_PER_YEAR  ||  ....)
    {
        ...
    } 
    if (! validInjectionsModulationTimeSeries())
        return false;
    if (! validWithdrawalsModulationTimeSeries())
        return false;
    if (! validRulesCurvesTimeSeries())
        return false;
    ...
    // Or we could even have : 
    return validInjectionsModulationTimeSeries() &&
           validWithdrawalsModulationTimeSeries() &&
           validRulesCurvesTimeSeries();
}
```

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -144,4 +115,72 @@ void Series::fillDefaultSeriesIfEmpty()
     fillIfEmpty(upperRuleCurve, 1.0);
 }
 
+bool Series::validate() const
```
guilpier-code, line 118: Why not : 
```C++
bool Series::validate() const
{
    return validateTimeSeriesSizes() && 
           validateMaxInjectionTimeSeries() && 
           validateMaxWithdrawalTimeSeries() && 
           validateRuleCurvesTimeSeries();
}
```
?

guilpier-code (changes requested):

guilpier-code (approved):

PR #1247, opened 2023-04-13
Make LMR optional with new option

flomnes:
LMR sometimes produces inconsistent results, hence the user may want to disable it.
### TODO
- [x] Add new exception
- [x] Document file format changes (docs/reference-guide/13-file-format.md)
- [x] Check consistency, logic
- [x] Fix compilation errors
- [x] Add tests


src/solver/application.h
```diff
@@ -40,6 +40,13 @@ class Application : public Yuni::IEventObserver<Application, Yuni::Policy::Singl
     */
     void prepare(int argc, char* argv[]);
 
+    /*!
+    ** \brief Check validity of adequacy patch parameters
+    **
+    ** \return False if the operation failed.
+    */
+    void prepareAdqPatchParams();
+
```
flomnes: 1. Replace `prepare` with `check`
2. Notice how most parameters are in `adqPatchParams`. This strongly suggest that we should create a member function in class `AdequacyPatch::AdqPatchParams`. The only exception is `pParameters->include.hurdleCosts`
3. No need for a new method `prepareAdqPatchParams`, call this new function in `Application::prepare`

src/solver/application.cpp
```diff
@@ -293,12 +257,10 @@ void Application::prepare(int argc, char* argv[])
...
     checkSimplexRangeHydroHeuristic(pParameters->simplexOptimizationRange, pStudy->areas);
 
-    checkAdqPatchStudyModeEconomyOnly(pParameters->adqPatchParams.enabled, pParameters->mode);
-
-    checkAdqPatchContainsAdqPatchArea(pParameters->adqPatchParams.enabled, pStudy->areas);
-    checkAdqPatchIncludeHurdleCost(pParameters->adqPatchParams.enabled,
-                                   pParameters->include.hurdleCosts,
-                                   pParameters->adqPatchParams.curtailmentSharing.includeHurdleCost);
+    if (pParameters->adqPatchParams.enabled)
+        pParameters->adqPatchParams.checkAdqPatchParams(pStudy->parameters.mode,
```
```suggestion
        pParameters->adqPatchParams.checkAdqPatchParams(pParameters.mode,
```

flomnes (commented):
Consider using `BOOST_CHECK_THROW` and `BOOST_CHECK_NO_THROW` to increase readability. They seem to do pretty much the same thing as your code.

https://www.boost.org/doc/libs/1_80_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level_no_throw.html

Milos-RTEi (approved):
I searched in vain for the UI button only to realize that it is not there :)

PR #1248, opened 2023-04-17
St storage output : adapt Unit() to Multiple output variables

guilpier-code:
Adapt **Unit()** static functions to Multiple output  variables.
This PR is an answer [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1238#issuecomment-1508478199) for PR #1238.
Multiple output variables contain several single output variables  that have a different unit, like **ShortTermStorageByGroup** with units  **MWh** or **MW** depending on they represent levels or injections / withdrawals.
**Unit()** static functions should be adapted accordingly.


src/solver/variable/economy/shortTermStorage.h
```diff
@@ -96,6 +96,14 @@ struct VCardShortTermStorage
         {
             return Antares::Data::ShortTermStorage::getVariableCaptionFromColumnIndex(indx);
         }
+
+        static std::string Unit(const unsigned int indx)
+        {
+            if (indx % 3 == 2)
+                return VCardShortTermStorage::Unit();
+            else
+                return "MW";
```
flomnes: This looks like a hack. It would be better to provide `indx` to `VCardShortTermStorage::Unit()`, right ?

flomnes (changes requested):

flomnes:
A few conflicts appeared, can you please resolve them ?

flomnes (approved):

PR #1265, opened 2023-04-26
[CLEAN] Remove unused var from ThermalCluster

### LINKED ISSUE #1260: Remove dependency to nbParallelYears & numSpace in antares/study/parts/thermal
Remove 3 unused members in antares/study/parts/thermal/cluster.h/.cpp
- `unitCountLastHour`
- `productionLastHour`
- `pminOfAGroup`

Remove `nbParallelYears` from `Data::ThermalCluster::ThermalCluster`

payetvin:
close #1260


flomnes (commented):
You removed the allocation `array = new XXX`, but not the de-referencing `array[index]`. No wonder you're getting segfaults.

It this a WIP ?

flomnes (commented):
Remove `ThermalCluster::pminOfAGroup` ?

src/solver/variable/state.cpp
```diff
@@ -55,6 +55,9 @@ void ThermalState::StateForAnArea::initializeFromArea(const Data::Area& area)
     numberOfUnitsONbyCluster.resize(count);
     thermalClustersOperatingCost.resize(count);
     PMinOfClusters.resize(count);
+    unitCountLastHour.resize(count);
+    productionLastHour.resize(count);
```
guilpier-code: **productionLastHour** : this variable is never initialized, it used to be though [(see here)](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1265#discussion_r1180360839)

Further more, it seems like it is used only in `State::initFromThermalClusterIndex`.
Could we not make it a local variable (a simple **double** should suffice) and remove it as a data member of class `ThermalState::StateForAnArea` ?

guilpier-code: And I have the same remark  for **unitCountLastHour**.
**pminOfAGroup** is not initialized but is used in 2 different `State` methods, so I guess it should remain a `ThermalState::StateForAnArea` data member.

guilpier-code (changes requested):

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -100,49 +100,6 @@ namespace Antares
...
-    assert(parent and "A parent for a thermal dispatchable cluster can not be null");
-
-    unitCountLastHour = new uint[nbParallelYears];
-    productionLastHour = new double[nbParallelYears];
-    pminOfAGroup = new double[nbParallelYears];
-    for (uint numSpace = 0; numSpace < nbParallelYears; ++numSpace)
-    {
-        unitCountLastHour[numSpace] = 0;
-        productionLastHour[numSpace] = 0.;
-        pminOfAGroup[numSpace] = 0.;
```
guilpier-code, line 142: Here **unitCountLastHour**, **productionLastHour** and **pminOfAGroup** used to be initialized

guilpier-code (changes requested):

src/libs/antares/study/parts/thermal/cluster.cpp
payetvin: Those var were double/uint arrays before, like `double* productionLastHour;`
We replaced those by vectors and resize allows to set them to the right size and filling them with zeros

src/solver/variable/state.cpp
```diff
@@ -147,12 +168,12 @@ void State::initFromThermalClusterIndex(const uint clusterAreaWideIndex, uint nu
...
+        uint previousUnitCount = thermal[area->index].unitCountLastHour[clusterAreaWideIndex];
 
         // Looking for the new number of units which have been started
         uint newUnitCount;
 
-        if (p > thermalCluster->productionLastHour[numSpace])
+        if (p > thermal[area->index].productionLastHour[clusterAreaWideIndex])
         {
             newUnitCount
               = static_cast<uint>(Math::Ceil(p / thermalCluster->nominalCapacityWithSpinning));
```
guilpier-code, line 177: **thermalClustersOperatingCost** : still the same remark : 
this vector seems to be initialized before we use it, but what's the point to have it as a data member when it could be a simple **double** local variable ?

src/solver/variable/state.cpp
```diff
@@ -73,14 +76,14 @@ State::State(Data::Study& s) :
 {
 }
 
-void State::initFromThermalClusterIndex(const uint clusterAreaWideIndex, uint numSpace)
+void State::initFromThermalClusterIndex(const uint clusterAreaWideIndex)
```
guilpier-code, line 79: variable **thermalClustersProductions** is initialized in a **if... else...** statement, but it's not clean : 
```C++
if (/*condition 1*/)
{
   ... 
   // giving a value to thermalClustersProductions
}
else
{
    // giving another value to thermalClustersProductions
}
```
We should initialize it to zero : `thermalClustersProductions.resize(count, 0.);`

src/solver/variable/state.cpp
```diff
@@ -73,14 +76,14 @@ State::State(Data::Study& s) :
...
 {
```
guilpier-code, line 80: **numberOfUnitsONbyCluster** : same thing. We should make it a current function local **double** variable, and initialize to zero before using it.

src/solver/variable/state.cpp
```diff
@@ -55,6 +55,9 @@ void ThermalState::StateForAnArea::initializeFromArea(const Data::Area& area)
     numberOfUnitsONbyCluster.resize(count);
     thermalClustersOperatingCost.resize(count);
     PMinOfClusters.resize(count);
```
guilpier-code, line 57: **PMinOfClusters** : we should make it a simple **double** local variable

guilpier-code (changes requested):

src/solver/variable/state.cpp
```diff
@@ -353,117 +352,149 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
     }
 
     if (dur > 0)
+        yearEndBuildFromThermalClusterIndexDurPositive(dur, ON_min, ON_max, ON_opt);
+
+    // Calculation of non linear and startup costs
+    yearEndBuildThermalClusterCalculateStartupCosts(dur, ON_min, ON_opt, currentCluster);
+}
+
+void State::yearEndBuildThermalClusterCalculateStartupCosts(uint dur,
```
guilpier-code: we should pass **std::array**s by reference to **const** : 
`const std::array<uint, Variable::maxHoursInAYear>& ON_min`
it would be faster (no copy) and safer.

src/solver/variable/state.cpp
```diff
@@ -353,117 +352,149 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
...
+            = thermalClusterStartupCostForYear + thermalClusterFixedCostForYear;
+        thermalClusterOperatingCostForYear[i] += thermalClusterNonProportionalCostForYear[i];
+
+        // Other variables for output
+        //\todo get from the cluster
+        thermalClusterDispatchedUnitsCountForYear[i] = optimalCount;
+    }
+}
+
+void State::yearEndBuildFromThermalClusterIndexDurPositive(uint dur,
```
guilpier-code: we should pass **std::array**s by reference to **const**, with an exception for **ON_opt** which is modifiable in the function and should be passed as a reference only

guilpier-code (changes requested):

src/solver/variable/state.cpp
```diff
@@ -353,117 +352,149 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
     }
 
     if (dur > 0)
+        yearEndBuildFromThermalClusterIndexDurPositive(dur, ON_min, ON_max, ON_opt);
```
guilpier-code: About function **yearEndBuildFromThermalClusterIndexDurPositive** : 
I don't understand the details of what it does, but what I see is that its mission is to populate **ON_opt** (that is the reason why tests were failing in CI when passing it by value : we populate a copy of **ON_opt**, not the actual one).
As mentioned when declaring  **ON_opt**, it's content is : 
_Nombre de groupes économiquement optimal en fonctionnement à l'heure h_
So we should : 
- rename this function, for instance, into : **computeEconomicallyOptimalNbClustersONforEachHour**
- make this function return the filled **std::array** : 
```C++
if (dur > 0)
    ON_opt = computeEconomicallyOptimalNbClusterForEachHour(dur, ON_min, ON_max);
```
I think this should not be costly at execution, as the move semantics will be triggered automatically, right ?.

guilpier-code: By the way , according to comments coming with its declaration, local variable **dur** should be renamed into **maxDurationON** (for instance).

src/solver/variable/state.cpp
```diff
@@ -353,117 +352,149 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
...
+
+    for (uint i = startHourForCurrentYear; i < endHourForCurrentYear; ++i)
     {
-        ON_opt[startHourForCurrentYear] = ON_min[startHourForCurrentYear];
-        h = startHourForCurrentYear + 1;
+        double thermalClusterStartupCostForYear = 0;
+        double thermalClusterFixedCostForYear = 0;
 
-        while (h < endHourForCurrentYear)
+        uint optimalCount;
```
guilpier-code: About function ` State::yearEndBuildThermalClusterCalculateStartupCosts` : 
Instead of : 
```cpp
uint optimalCount;
(dur == 0) ? (optimalCount = ON_min[i]) : (optimalCount = ON_opt[i]);
```
we should have : 
```cpp
uint optimalCount = (! dur) ? ON_min[i] : ON_opt[i];
```
Same thing for local variable **delta**
And we should replace **i** with **hour**.

guilpier-code (changes requested):

src/solver/variable/state.cpp
```diff
@@ -353,117 +352,149 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
     }
 
     if (dur > 0)
+        yearEndBuildFromThermalClusterIndexDurPositive(dur, ON_min, ON_max, ON_opt);
+
+    // Calculation of non linear and startup costs
+    yearEndBuildThermalClusterCalculateStartupCosts(dur, ON_min, ON_opt, currentCluster);
+}
+
+void State::yearEndBuildThermalClusterCalculateStartupCosts(uint& dur,
```
guilpier-code: Still about function ` State::yearEndBuildThermalClusterCalculateStartupCosts` : 
- the name of this function is misleading : actually, it computes start up costs locally to update more global variables : a unit count ON per hour,  non proportional costs per hour, operating costs per hour. The local start up costs are never stored.
- According to the previous point, this function does 3 things, and could easily be split.

So instead of : 
```c++
yearEndBuildThermalClusterCalculateStartupCosts(dur, ON_min, ON_opt, currentCluster);
```
which is misleading, we could have : 
```c++
thermalClusterDispatchedUnitsCountForYear = computeUnitsON(currentCluster, dur, ON_min, ON_opt);
thermalClusterNonProportionalCostForYear = computeClusterNonProportionalCosts(currentCluster, ON_min, ON_opt, dur);
thermalClusterOperatingCostForYear = computeClusterOperatingCostsCosts();
```
If we want to do this split, it is preferable to turn the type of arrays returned by these new functions into **std::array**, to use move semantics and the confort and speed it provides.

guilpier-code (dismissed):

src/solver/variable/state.cpp
flomnes: No, it's used in `solver/variable/economy/productionByDispatchablePlant.h:352`

src/solver/variable/state.cpp
payetvin: The for loop should be in the calling function `yearEndBuildFromThermalClusterIndex`, this would duplicate code

src/solver/variable/state.cpp
```diff
@@ -352,118 +349,152 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
             ON_max[h] = ON_min[h];
     }
 
-    if (dur > 0)
+    if (maxDurationON > 0)
+        ON_opt = computeEconomicallyOptimalNbClustersONforEachHour(maxDurationON, ON_min, ON_max);
```
flomnes, line 353: `ON_opt` is not initialized if this if block is not executed.

flomnes (changes requested):
Unitialized variable

src/solver/variable/state.cpp
```diff
@@ -242,6 +242,7 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex)
     std::array<uint, Variable::maxHoursInAYear> ON_max;
     // Nombre de groupes économiquement optimal en fonctionnement à l'heure h
     std::array<uint, Variable::maxHoursInAYear> ON_opt;
+    ON_opt.fill(0);
```
flomnes:
```suggestion
    std::array<uint, Variable::maxHoursInAYear> ON_opt{};
```

src/solver/variable/state.h
```diff
@@ -97,7 +111,17 @@ class State
     **
     ** \param areaWideIndex Index of the thermal cluster for the current area
     */
-    void yearEndBuildFromThermalClusterIndex(const unsigned int areaWideIndex, uint numSpace);
+    void yearEndBuildFromThermalClusterIndex(const unsigned int areaWideIndex);
+
+    void yearEndBuildThermalClusterCalculateStartupCosts(const uint& maxDurationON,
```
flomnes: You should split state.h into 2 headers
- state.h for all public functions
- state_priv.h for all private functions, which we don't really want to expose outside of state.cpp

flomnes (changes requested):
Please split state.h

src/solver/variable/state.h
```diff
@@ -60,6 +61,10 @@ class ThermalState
 
         //! Minimum power of all clusters for the current hour in the year
         std::vector<double> PMinOfClusters;
+
+        unsigned int unitCountLastHour = 0;
+        double productionLastHour = 0;
```
flomnes: I think these should be vectors. One coefficient for each thermal cluster within the area.

src/solver/variable/state.cpp
```diff
@@ -147,12 +164,12 @@ void State::initFromThermalClusterIndex(const uint clusterAreaWideIndex, uint nu
         // alias to the production of the current thermal cluster
         double p = thermal[area->index].thermalClustersProductions[clusterAreaWideIndex];
         // alias to the previous number of started units
-        uint previousUnitCount = thermalCluster->unitCountLastHour[numSpace];
+        uint previousUnitCount = thermal[area->index].unitCountLastHour;
```
flomnes:
```suggestion
        uint previousUnitCount = thermal[area->index].unitCountLastHour[clusterAreaWideIndex];
```

flomnes (changes requested):

src/solver/variable/state.h
flomnes: Otherwise you'll only take the last cluster into account, with overwriting, etc.

src/solver/variable/state.cpp
```diff
@@ -352,118 +349,152 @@ void State::yearEndBuildFromThermalClusterIndex(const uint clusterAreaWideIndex,
...
+        // Other variables for output
+        //\todo get from the cluster
+        thermalClusterDispatchedUnitsCountForYear[hour] = optimalCount;
+    }
+}
+
+std::array<uint, Variable::maxHoursInAYear>
+State::computeEconomicallyOptimalNbClustersONforEachHour(const uint& maxDurationON,
+                        std::array<uint, Variable::maxHoursInAYear>& ON_min,
+                        std::array<uint, Variable::maxHoursInAYear>& ON_max) const
```
flomnes:
```suggestion
State::computeEconomicallyOptimalNbClustersONforEachHour(const uint& maxDurationON,
                        const std::array<uint, Variable::maxHoursInAYear>& ON_min,
                        const std::array<uint, Variable::maxHoursInAYear>& ON_max) const
```

flomnes (changes requested):

flomnes (approved):

PR #1267, opened 2023-04-26
Fix indices in modeler

guilpier-code:
This PR for a discussion on how we should handle the **modeler** regarding the **short term storage** functionality.


src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
```diff
@@ -173,9 +173,9 @@ static void setBoundsForShortTermStorage(PROBLEME_HEBDO* problemeHebdo,
           = problemeHebdo->CorrespondanceVarNativesVarOptim[pdtJour];
         for (int areaIndex = 0; areaIndex < problemeHebdo->NombreDePays; areaIndex++)
         {
-            int areaWideIndex = 0;
             for (const auto& storage : (*problemeHebdo->ShortTermStorage)[areaIndex])
             {
+                int storageIndex = 0;
```
flomnes: It seems that storageIndex is always 0, right ?

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -43,14 +43,15 @@ static void importShortTermStorages(
...
   std::vector<::ShortTermStorage::AREA_INPUT>& ShortTermStorageOut)
 {
-    int STindex = 0;
+    int globalIndex = 0;
     for (uint areaIndex = 0; areaIndex != areas.size(); areaIndex++)
     {
         ShortTermStorageOut[areaIndex].resize(areas[areaIndex]->shortTermStorage.count());
         for (auto st : areas[areaIndex]->shortTermStorage.storagesByIndex)
         {
+            int STindex = 0;
```
flomnes: Same remark, STindex is always 0.

flomnes (changes requested):

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
guilpier-code: oops, corrected.

src/solver/simulation/sim_calcul_economique.cpp
guilpier-code: oops, corrected.

src/solver/optimisation/opt_gestion_des_bornes_cas_lineaire.cpp
flomnes, line 1: Only renaming here

src/solver/simulation/sim_calcul_economique.cpp
flomnes, line 1: In this function, you set `toInsert.globalIndex` to the global index, not the local index.

I agree with this change. Strange that we never found it, it shows that our tests are insufficient.

flomnes (approved):

src/solver/simulation/sim_calcul_economique.cpp
flomnes: Also, consider renaming `STindex` to `storageIndex` for consistency.

payetvin (approved):

PR #1269, opened 2023-04-27
Remove pointer on `std::vector`s

flomnes:
### Source of all changes
Please see file [src/solver/simulation/sim_structure_probleme_economique.h](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1269/files#diff-c60d123e7a9e897e3a0aed9a2b8dd3b34cb442f9e8bfbaa64cf749195b2ac2f7) for the source of all changes.

### Remark on dynamic allocation
`PROBLEME_HEBDO`'s C++ members are initialized with `new`, so all constructors are called on data members.
```cpp
solver/simulation/economy.cpp:87:27:        pProblemesHebdo = new PROBLEME_HEBDO*[pNbMaxPerformedYearsInParallel];
solver/simulation/economy.cpp:93:41:            pProblemesHebdo[numSpace] = new PROBLEME_HEBDO();
```

Hopefully, this change allows for easier reading in the rest of the code : we no longer need to de-reference vectors, which was a bit unnatural to say the least.


src/solver/variable/state.hxx
```diff
@@ -79,12 +79,12 @@ inline void State::initFromAreaIndex(const unsigned int areaIndex, uint numSpace
     {
     case Data::stdmEconomy:
     {
-        hourlyResults = problemeHebdo->ResultatsHoraires[areaIndex];
+        hourlyResults = &problemeHebdo->ResultatsHoraires[areaIndex];
```
sylvlecl, line 82: Not just `problemeHebdo.ResultatsHoraires` ?

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -136,7 +136,7 @@ void SIM_AllocationProblemeHebdo(PROBLEME_HEBDO& problem, int NombreDePasDeTemps
     problem.ReserveJMoins1 = (RESERVE_JMOINS1**)MemAlloc(nbPays * sizeof(void*));
     // using new here since RESULTATS_HORAIRES contains C++ structures such as std::vector
```
sylvlecl: we can remove the comment

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -400,7 +400,7 @@ void SIM_AllocationProblemeHebdo(PROBLEME_HEBDO& problem, int NombreDePasDeTemps
         problem.ReserveJMoins1[k] = (RESERVE_JMOINS1*)MemAlloc(sizeof(RESERVE_JMOINS1));
         // using new here since RESULTATS_HORAIRES contains C++ structures such as std::vector
         // we must call their constructors, which MemAlloc doesn't do
-        problem.ResultatsHoraires[k] = new RESULTATS_HORAIRES();
+        // problem.ResultatsHoraires[k] = new RESULTATS_HORAIRES();
```
sylvlecl: we can remove the comment before merging

sylvlecl (commented):
Nice improvement indeed

src/solver/variable/state.hxx
flomnes: Because `problemeHebdo` is a pointer to a `PROBLEME_HEBDO`.

In C/C++, 

- `&a->b` (or `&(a->b)`) is the address of b contained in a structure pointed by `a`
- `a.b` is the member `b` in structure `a`, which is different

PR #1273, opened 2023-04-28
Minimal hydro generation (RTEi's -> CR22)

Milos-RTEi:
This PR is Identical to closed PR: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1203
The transfer was done in order to use sonar-cloud.

===
This PR is created in order to implement RTEi development on CR22 (Minimal Hydro Generation).
Spec:
https://rteinternational.sharepoint.com/:w:/r/sites/AntaresUser%27sClub-Interne/Documents%20partages/Change%20Requests/Selected%20for%20development/CR22%20-%20Multiple%20Minimum%20Generation%20for%20Hydro%20Reservoirs/Functional_Specifications-Multiple_min_gen_for_hydro_clusters_V0.1-HA.docx?d=web3abdf476d94fc8adca58d741bffb94&csf=1&web=1&e=wfbOie

The objective is to allow the user to specify several hourly minimum generating power time-series for a given reservoir. To build a given Monte-Carlo year, Antares would then select one of these time-series (either randomly or via a proper setting of the scenario builder). This parameter is applicable whether the reservoir is managed via the heuristic or with water values.


flomnes:
Hi @Milos-RTEi, thanks for transferring the PR here.

It seems that logs are missing in the 1st tests that is run. This usually means that simulations don't go up to the end. Can you try to reproduce/fix ?

Milos-RTEi:
Hi @flomnes 
I saw that we are failing the first test in this CR. 
If I understand correctly this is like some end-to-end test. Where can I find the Antares model for this test?
Or I'm completely wrong and this is some kind of unit test. In that case in which file is it generated?

flomnes:
I think it's a specific test. Please check .github/workflows for more details.

src/solver/hydro/management/management.cpp
```diff
@@ -96,47 +97,247 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
...
+          }
+          data.totalYearInflows = totalYearInflows;
+      });
+}
 
-        auto& ptchro = *NumeroChroniquesTireesParPays[numSpace][z];
+void HydroManagement::minGenerationScaling(uint numSpace)
+{
+    if (study.header.version < 860)
+        return;
```
flomnes: You may remove this. As much as possible, we try to check version numbers only for I/O

src/solver/hydro/management/management.cpp
```diff
@@ -96,47 +97,247 @@ void HydroManagement::prepareInflowsScaling(uint numSpace)
...
             }
         }
-    });
+    }
+}
+
+void HydroManagement::checkMinGeneration(uint numSpace)
+{
+    if (study.header.version < 860)
+        return;
```
flomnes: You may remove this

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -665,6 +665,18 @@ void SIM_RenseignementProblemeHebdo(PROBLEME_HEBDO& problem,
...
+                if (study.header.version >= 860)
+                {
+                    auto const& srcmingen
+                      = mingenmatrix[tsIndex < mingenmatrix.width ? tsIndex : 0];
+                    for (uint j = 0; j < problem.NombreDePasDeTemps; ++j)
+                    {
+                        problem.CaracteristiquesHydrauliques[k]->MingenHoraire[j]
+                          = srcmingen[PasDeTempsDebut + j];
+                    }
+                }
```
flomnes:
```suggestion
                auto& mingenmatrix = area.hydro.series->mingen;
                {
                    auto const& srcmingen
                      = mingenmatrix[tsIndex < mingenmatrix.width ? tsIndex : 0];
                    for (uint j = 0; j < problem.NombreDePasDeTemps; ++j)
                    {
                        problem.CaracteristiquesHydrauliques[k]->MingenHoraire[j]
                          = srcmingen[PasDeTempsDebut + j];
                    }
```

As much as possible, we try to avoid making distinctions based on the version for anything else than I/O

Previously, Xmin[Var] was 0. I suggest setting `mingenmatrix` to 0 coefficients by default.

Another problem with the current implementation is that if version < 860, `MingenHoraire` is unitialized, which can lead to [UB](https://en.wikipedia.org/wiki/Undefined_behavior).

@guilpier-code what's your take on this ?

flomnes (changes requested):

src/solver/hydro/management/management.cpp
Milos-RTEi: If this is removed unfeasible-problem test will fail and break the workflow.

src/solver/hydro/management/management.cpp
Milos-RTEi: If this is removed unfeasible-problem test will fail and break the workflow.

src/solver/simulation/sim_calcul_economique.cpp
Milos-RTEi: same as above

flomnes (approved):

PR #1283, opened 2023-05-03
Add output variables for short-term storage in adequacy mode

flomnes:
- Variables by group (PSP_Open, etc.)
- Variables by cluster
- Variables by district

Also format `all.h` using clang-format.


src/solver/variable/economy/all.h
```diff
@@ -125,60 +124,57 @@ class Links;
...
+               <DispatchableGeneration // All dispatchable generation
+                <RenewableGeneration   // All renewable generation
+                 <HydroStorage         // Hydro Storage Generation
+                  <Pumping             // Pumping generation
+                   <ReservoirLevel     // Reservoir levels
+                    <Inflows           // Hydraulic inflows
+                     <Overflows        // Hydraulic overflows
+                      <WaterValue      // Water values
+                       <HydroCost      // Hydro costs
+                        <ShortTermStorageByGroup<STstorageInjectionByCluster<
```
guilpier-code, line 156: **Indentation** : 
why folding the ST storage variable here (**economy/all.h**) and not in **adequacy/all.h** ?
To me, unfolding them is more clear.

guilpier-code (changes requested):

src/solver/variable/economy/all.h
flomnes: I only used clang-format, nothing else.

flomnes: I could format by hand, but I don't really want to

PR #1288, opened 2023-05-04
Fix version check for mingen

flomnes:
Hello @Milos-RTEi, following [this discussion](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1273#discussion_r1184988018), I removed a few checks on version in the "core".The problem was that mingen was empty (width=0, height=8760), see call to function `reset`.

The idea is that if no Pmin files are present, we set one column of 0. This way
- The existing behavior is kept (Pmin=0)
- We don't need to worry about version numbers in the computation _ie_ after loading the data

### DIFF
--- a/src/solver/ts-generator/hydro.cpp
+++ b/src/solver/ts-generator/hydro.cpp
@@ -62,6 +62,14 @@ static void PreproHydroInitMatrices(Data::Study& study, uint tsCount)
     });
 }
 
+static void PreproHydroFixMingenToZero(Data::Study& study)
+{
+    study.areas.each([&](Data::Area& area) {
+        auto& hydroseries = *(area.hydro.series);
+        hydroseries.mingen.reset(1, HOURS_PER_YEAR);
+    });
+}
+
 static void PreproRoundAllEntriesPlusDerated(Data::Study& study)
 {
     bool derated = study.parameters.derated;
@@ -321,6 +329,8 @@ bool GenerateHydroTimeSeries(Data::Study& study, uint currentYear, IResultWriter
     }
 
     delete[] NORM;
+
+    PreproHydroFixMingenToZero(study);
 
     return true;
 }
--- a/src/solver/ts-generator/hydro.cpp
+++ b/src/solver/ts-generator/hydro.cpp
@@ -58,7 +58,6 @@ static void PreproHydroInitMatrices(Data::Study& study, uint tsCount)
 
         hydroseries.ror.resize(tsCount, HOURS_PER_YEAR);
         hydroseries.storage.resize(tsCount, DAYS_PER_YEAR);
-        hydroseries.mingen.resize(tsCount, HOURS_PER_YEAR);
         hydroseries.count = tsCount;
     });
 }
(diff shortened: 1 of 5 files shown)


Milos-RTEi:
Ah, you're right. This is so much better solution. Cleaner & shorter. 
I didn't see it.   
However, I see 3 end-to-end tests have failed. 018-040-055
Maybe mingen is not set to zero cause I see some strange numbers for mingen:
image

Milos-RTEi:
I finally realized why we have strange non-zero values for mingen. 
The mingen.reset() method sets properly all the values to zero. 
However, those values are not used at all. 
In these 3 tests, the ts-generator is turned on for Hydro. 
So in addition to ror and storage, random mingen values are generated and that breaks the tests.
Any suggestions!?

Milos-RTEi:
We can remove our changes in src/solver/ts-generator/hydro.cpp?
And that will leave all the values at zero.

flomnes:
> We can remove our changes in src/solver/ts-generator/hydro.cpp? And that will leave all the values at zero.

I think that's the proper way to do it.

src/solver/ts-generator/hydro.cpp
```diff
@@ -58,7 +58,6 @@ static void PreproHydroInitMatrices(Data::Study& study, uint tsCount)
 
         hydroseries.ror.resize(tsCount, HOURS_PER_YEAR);
         hydroseries.storage.resize(tsCount, DAYS_PER_YEAR);
-        hydroseries.mingen.resize(tsCount, HOURS_PER_YEAR);
```
flomnes, line 61: The culprit !

flomnes:
@Milos-RTEi Altough your solution works, it souldn't be necessary to include mingen in TS generation, since mingen TS are not to be generated.

Milos-RTEi:
> @Milos-RTEi Altough your solution works, it souldn't be necessary to include mingen in TS generation, since mingen TS are not to be generated.

Indeed first I tried just to remove mingen from hydro-ts-generator. [6a1e476]
But broke a lot of tests. Because mingen.width = 0 comes to minGenerationScaling method. (if ts-gen is used for hydro)
So it is either checking mingen.width==0 -> mingen.reset(1,8760) in minGenerationScaling.
or do something similar in ts-generator. As I did.

Milos-RTEi (approved):

flomnes:
> > @Milos-RTEi Altough your solution works, it souldn't be necessary to include mingen in TS generation, since mingen TS are not to be generated.
> 
> Indeed first I tried just to remove mingen from hydro-ts-generator. [[6a1e476](https://github.com/AntaresSimulatorTeam/Antares_Simulator/commit/6a1e47660ba4547bc1908d51a8962def2b7cc840)] But broke a lot of tests. Because mingen.width = 0 comes to minGenerationScaling method. (if ts-gen is used for hydro) So it is either checking mingen.width==0 -> mingen.reset(1,8760) in minGenerationScaling. or do something similar in ts-generator. As I did.

I may have found a better solution e4a1162. The problem was that if ts-generation was enabled for hydro, `DataSeriesHydro::loadFromFolder` was not executed, only `PreproHydro::loadFromFolder`. The solution is to set mingen to one column of 0s at construction of class `DataSeriesHydro`.

I reverted 1358d36.

PR #1294, opened 2023-05-09
Name constraints & variables in MPS files

### LINKED ISSUE #1051: MPS : name constraints & variables
### Variables
1. Name
2. Location (if applicable)
3. Time identifier (week/day/hour)

Examples
```
Flow::link<id>::hour<h>
UnsupEnrg::area<id>::hour<h>
DispatchableProduction::area<area-id>::ThermalCluster<id>::hour<h>

```
### Constraints
Same idea as with variables

Examples
```
balance::area<id>::hour<h>
flowDissociation::link<id>::hour<h>
my-bc::weekly_bc::week<w>
HydroEnergySmoothingUsingVariationMaxUp::area<id>::hour<h>
HydroFinalLevelSumOfWaterValueSlices::area<id>::weekly
```

### TODO
- [ ] Handle variables and constraints that have not been named
- [ ] Remove unused file variable.txt
- [ ] Harmonize naming for variables and constraints
- [ ] Hamonize the code for variables and constraints
- [ ] Remove spaces & special characters in variable/constraint names
- [ ] ~~Handle empty constraint/variable names : "C%07d" or "R%07d" as before~~
- [ ] Name thermal clusters in **SIM_**

It may be necessary to convert `std::vector<std::string>` to `const char**`. Possible implementation
```cpp
#include <iostream>
#include <vector>
#include <string>

class Converter {
public:
  std::pair<const char**, size_t> run(const std::vector<std::string>& v)
  {
    p.resize(v.size());
    for (size_t idx = 0; idx < v.size(); ++idx) {
        p[idx] = v[idx].c_str();
    }
    return {p.data(), p.size()};
  }
private:
  std::vector<const char*> p;
};

int main() {
  std::vector<std::string> H = {"Hello", "world", " !"};
  Converter converter;
  auto [array, size] = converter.run(H);
  for (size_t idx = 0; idx < size; idx++)
  {
    std::cout << array[idx] << std::endl;
  }
}
```

a-zakir:
Naming variables and constraints of mps


src/solver/optimisation/opt_construction_contraintes_couts_demarrage.cpp
```diff
@@ -58,11 +61,12 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaireCoutsDeDemarrage(
     int* Colonne = ProblemeAResoudre->Colonne;
 
     int nbTermesContraintesPourLesCoutsDeDemarrage = 0;
-
+    auto study = Study::Current::Get();
     for (int pays = 0; pays < problemeHebdo->NombreDePays; pays++)
     {
         const PALIERS_THERMIQUES* PaliersThermiquesDuPays
           = problemeHebdo->PaliersThermiquesDuPays[pays];
+        const auto& zone = study->areas[pays]->name.c_str();
```
flomnes: Better to use `problemeHebdo->NomsDesPays` here

src/solver/optimisation/opt_alloc_probleme_a_optimiser.cpp
```diff
@@ -130,8 +147,18 @@ void OPT_FreeOptimizationData(PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre)
     MemFree(ProblemeAResoudre->Pi);
     MemFree(ProblemeAResoudre->Colonne);
 
-    ProblemeAResoudre->NomDesVariables.clear();
-    ProblemeAResoudre->NomDesContraintes.clear();
+    // ProblemeAResoudre->NomDesVariables.clear();
+    for (int i(0); i < ProblemeAResoudre->NombreDeVariables; i++)
```
flomnes:
```suggestion
    for (uint i = 0; i < ProblemeAResoudre->NombreDeVariables; i++)
```

src/solver/optimisation/opt_alloc_probleme_a_optimiser.cpp
```diff
@@ -51,6 +51,9 @@ using namespace Antares;
 #else
 #define SNPRINTF snprintf
 #endif
+// #TODO check mps var max size
+#define MAX_VAR_NAME 200
+#define MAX_CNT_NAME 200
```
flomnes: It would be better to allocate only what's needed

flomnes (changes requested):

flomnes:
Please resolve conflicts with the base branch

src/solver/optimisation/opt_alloc_probleme_a_optimiser.cpp
a-zakir: ok then freeing (NomsDesContraintes[i]) has to be moved to the new allocation place.

src/libs/antares/study/study.cpp
```diff
@@ -935,7 +935,11 @@ bool Study::areaRename(Area* area, AreaName newName)
     // A name must not be empty
     if (not area or newName.empty())
         return false;
-
+    if (CheckForbiddenCharacterInAreaName(newName))
```
guilpier-code: Do we need to add this block : 
```C++
if (CheckForbiddenCharacterInAreaName(newName))
{
    logs.error() << "character '*' is forbidden in area name: `" << newName << "`";
    return false;
}
```
?
Indeed, further in the current function, we have the instruction : 
```C++
ret = areas.renameArea(oldid, newid, newName);
```
and renameArea(...) already contains a call to **CheckForbiddenCharacterInAreaName**.

src/libs/antares/study/area/list.cpp
```diff
@@ -497,10 +502,14 @@ Area* AreaListAddFromNames(AreaList& list,
 {
     if (!name || !lname)
         return nullptr;
+    if (CheckForbiddenCharacterInAreaName(name))
```
guilpier-code: Here we're in **AreaListAddFromNames(...)** function.
Due to its this name, we're supposed to only add an area to a list of areas.
Adding a check on the area name here leads to make the function do 2 things : a check on the area name and an addition.
We should move this check where it's more likely to belong : in the function that calls  **AreaListAddFromNames(...)**, where there are other checks on names : 
```C++
if (lname.empty())
{
    logs.warning() << "ignoring invalid area name: `" << name << "`, " << filename
                   << ": line " << line;
    continue;
}
```
And maybe group these checks into one free function names **checkAreaName(areaName)**.

src/libs/antares/study/area/list.cpp
```diff
@@ -1401,6 +1434,11 @@ bool AreaList::renameArea(const AreaName& oldid, const AreaName& newid, const Ar
     if (!oldid || !newName || !newid || areas.empty())
         return false;
 
+    if (CheckForbiddenCharacterInAreaName(newName))
```
guilpier-code, line 1412: Maybe call **checkAreaName(areaName);**, see comment above.

flomnes (changes requested):
If possible, please revert all changes related to indentation.

flomnes (changes requested):
1. Some lines are empty on study [008 Thermal fleet - Accurate unit commitment](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR/tree/main/short-tests/008%20Thermal%20fleet%20-%20Accurate%20unit%20commitment). See [problem.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/11651943/problem.zip).
2. There are still some `R0001344  1.0000000000` in the MPS file, to be investigated.
3. Please branch also the OR-Tools interface (antares-8.6-solver --use-ortools --ortools-solver xpress). Currently when using OR-Tools, MPS files still have C00045 & R00030 and so on.

src/solver/optimisation/opt_structure_probleme_a_resoudre.h
```diff
@@ -109,7 +109,9 @@ typedef struct
 
     /* Nommage des variables & contraintes */
     std::vector<std::string> NomDesVariables;
+    // char** NomDesVariables;
```
guilpier-code: Please remove unnecessary comments

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,70 @@
...
+        const auto location = origin + ZONE_SEPARATOR + destination;
+        auto fullName = BuildName(
+          Antares::Data::Enum::toString(structDict),
+          LocationIdentifier(location, Antares::Data::Enum::ExportStructLocationDict::link),
+          TimeIdentifier(ts, timeStepType));
+        problem->NomDesVariables[var] = fullName;
+    }
+}
+
+void RenameZoneVariable(PROBLEME_ANTARES_A_RESOUDRE* problem,
```
guilpier-code: **Renaming** : 
**RenameZoneVariable(...)** --> **RenameAreaVariable(...)**

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -231,16 +239,19 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* pro
...
+                    RenameLinkVariable(ProblemeAResoudre,
+                                       var,
+                                       Enum::ExportStructDict::ValeurDeNTCOrigineVersExtremite,
+                                       timeStepInYear,
+                                       Enum::ExportStructTimeStepDict::hour,
+                                       origin,
+                                       destination);
                 }
                 interco = problemeHebdo->IndexSuivantIntercoExtremite[interco];
             }
```
guilpier-code, line 218: **Not strictly in the scope of this PR** (but should still be treated)
**exportPaliers(...)** is called twice in this source file. 
Shouldn't it be called only once ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -94,20 +94,34 @@ class TimeMeasurement
     clock::time_point start_;
     clock::time_point end_;
 };
-
+char** VectorOfStringToCharPP(std::vector<std::string>& in, std::vector<char*>& pointerVec)
+{
+    std::transform(in.begin(),
+                   in.end(),
+                   std::back_inserter(pointerVec),
+                   [](std::string& str) { return str.data(); });
```
```suggestion
                   [](std::string& str) { return str.empty() ? nullptr : str.data(); });
```

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -402,14 +420,14 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* pro
...
 
-                    if (exportStructure)
-                    {
-                        OPT_Export_add_variable(varname,
-                                                var,
-                                                Enum::ExportStructDict::DefaillanceNegative,
-                                                timeStepInYear, // TODO[FOM] remove
-                                                pays);
-                    }
+                    // if (exportStructure)
```
guilpier-code: Please remove the commented code.

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -41,8 +42,7 @@ void exportPaliers(const PROBLEME_HEBDO& problemeHebdo,
                    int& nombreDeTermes,
                    double* Pi,
                    int* Colonne,
-                   int timeStepInYear,
-                   std::vector<std::string>& varname)
+                   int timeStepInYear)
 {
     const PALIERS_THERMIQUES* PaliersThermiquesDuPays = problemeHebdo.PaliersThermiquesDuPays[pays];
```
guilpier-code: Not strictly related to this PR : 
**exportPaliers(...)** does 2 things and not only some exporting (as its name suggests) : 
- it builds some terms (related to thermal clusters) of a constraint
- it renames variables and stores them in a **PROBLEME_ANTARES_A_RESOUDRE** 

So either it should be renamed, or it should be split (which I recommend).

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -57,15 +57,15 @@ void exportPaliers(const PROBLEME_HEBDO& problemeHebdo,
...
-            if (problemeHebdo.ExportStructure)
-            {
-                OPT_Export_add_variable(varname,
-                                        var,
-                                        Enum::ExportStructDict::PalierThermique,
-                                        timeStepInYear, // TODO[FOM] remove
-                                        pays,
-                                        palier);
-            }
+            const auto& zone = problemeHebdo.NomsDesPays[pays];
```
guilpier-code: We shouldn't **zone** instead of **area**, which sticks better to the business terminology

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -57,15 +57,15 @@ void exportPaliers(const PROBLEME_HEBDO& problemeHebdo,
...
+            const auto& namePalier = PaliersThermiquesDuPays->NomsDesPaliersThermiques[index];
```
guilpier-code: **namePalier** is half french and half english. We should rename it **thermalClusterName**

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,112 @@
...
+    {
+        auto fullName
+          = BuildName(Antares::Data::Enum::toString(structDict),
+                      LocationIdentifier(zone, Antares::Data::Enum::ExportStructLocationDict::area)
+                        + SEPARATOR + "Layer<" + std::to_string(layerIndex) + ">",
+                      TimeIdentifier(ts, timeStepType));
+        problem->NomDesVariables[var] = fullName;
+    }
+}
+void RenameThermalClusterVariable(PROBLEME_ANTARES_A_RESOUDRE* problem,
```
guilpier-code: **Remark more general than the current file scope**.

RenameThermalClusterVariable(...) and **RenameShortTermStorageVariable(...)** are just about the same.
So only one function is enough.
Moreover, to me, some arguments are useless  for these 2 functions: 
- For these 2 functions, **timeStepType** is always hourly.
- **structDict** can probably be avoided (see a bit further in this comment)
- Not sure **problem** is useful, as this function could return the **fullname**, removing the need for this argument. And the function would be more clear. 
  
Instead of a bunch of free functions, we could create a class, this would surely simplify file **opt_construction_matrice_des_contraintes_cas_lineaire.cpp**
- The starting idea is to reduce the number of arguments each function takes.
- Another idea is that we have loops over time steps, areas and links, so we could update the time step, the area, ... of an object of this class at the beginning of the loops, so that we would not have to pass them as argument.

In **opt_construction_matrice_des_contraintes_cas_lineaire.cpp**, it could be : 
```C++
void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(...)
{
    // The ProblemeAResoudre now does not need to be passed as argument in any 
    // function or method dealing with variable names    
    VariablesNamer varNamer(problemeHebdo->ProblemeAResoudre); 
   
    for (int pdt = 0; ....)
    {
          varNamer.updateTimeStep(pdt);
          for (int pays = 0; ....)
          {
              varNamer.updateAreaName(problemeHebdo->NomsDesPays[pays]);
              ...
              // Number of arguments reduced because addHydroGeneration already knows things
              // and the complexity is move inside the object / class
              varNamer.addHydroGeneration(varNumber);
              ...
              varNamer.addUnsuppliedEnergy(varNumber);
              ...
          }
    }    
    ....
}
```

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -30,6 +30,7 @@
 #include "../simulation/simulation.h"
```
guilpier-code, line 30: I can't see why we rename variables in a file that is about building constraints. But for lack of a better solution, that's the way it is for now.

guilpier-code (changes requested):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -94,20 +94,34 @@ class TimeMeasurement
...
     PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre = problemeHebdo->ProblemeAResoudre;
-    Optimization::PROBLEME_SIMPLEXE_NOMME Probleme(ProblemeAResoudre->NomDesVariables,
-                                                   ProblemeAResoudre->NomDesContraintes,
+    std::vector<char*> pointerVecVar;
+    std::vector<char*> pointerVecCnt;
+    auto NomsDesVariablesCharPP
+      = VectorOfStringToCharPP(ProblemeAResoudre->NomDesVariables, pointerVecVar);
+    auto NomDesContraintesCharPP
+      = VectorOfStringToCharPP(ProblemeAResoudre->NomDesContraintes, pointerVecCnt);
+    Optimization::PROBLEME_SIMPLEXE_NOMME Probleme(NomsDesVariablesCharPP,
```
flomnes: In order to remain valid, this instance of `PROBLEME_SIMPLEXE_NOMME` implicitly needs `pointerVecVar` and `pointerVecCnt`. If one of these vectors is destroyed, `NomsDesVariablesCharPP` or `NomDesContraintesCharPP` will become dangling pointers, leading to UB.

For this reason, I suggest including `pointerVecCnt` & `pointerVecVar` as private members of class `Optimization::PROBLEME_SIMPLEXE_NOMME`. This will tie the lifetime of these objects so the outside world doesn't need to worry about this.

So you can just write something like
```cpp
    Optimization::PROBLEME_SIMPLEXE_NOMME Probleme(ProblemeAResoudre->NomDesVariables,
ProblemeAResoudre->NomDesContraintes);
```
and hide the implementation details in the constructor of `Optimization::PROBLEME_SIMPLEXE_NOMME`.

---

There's another layer of hidden dependency, which is `ProblemeAResoudre->NomDesVariables` & `ProblemeAResoudre->NomDesContraintes`, for which we have exactly the same problem. But since this instance of `PROBLEME_ANTARES_A_RESOUDRE` has a "long life" (created at the start, destroyed at the end), it's not really an issue for now.

flomnes (changes requested):

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -299,27 +360,32 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* pro
...
                                     Pi,
-                                    Colonne);
+                                    Colonne,
+                                    zone,
+                                    timeStepInYear);
 
             CorrespondanceCntNativesCntOptim->NumeroDeContrainteDesBilansPays[pays]
               = ProblemeAResoudre->NombreDeContraintes;
 
+            std::string constraintFullName
```
guilpier-code: To follow on from [this remark](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1294/files?show-viewed-files=true&file-filters%5B%5D=#r1218284014), we could also hide the complexity of renaming constraints : 
```C++
    // Context : building constraint OPT function
    ConstraintNamer constrantNamer(ProblemeAResoudre);
    for (int pdt = 0; ...)
    {
       constrantNamer.update(pdt);
       for (int pays = 0; ...)
       {
            constrantNamer.update(areaName);
            ...
            constrantNamer.addAreaBalance(); // No argument required here
            OPT_ChargerLaContrainteDansLaMatriceDesContraintes(...) // No more constraintFullName arg
            ...
       }
    }
```
This would require to simplify **OPT_ChargerLaContrainteDansLaMatriceDesContraintes**, which would no longer be passes **constraintFullName**, and internally would not store the constraint name into **ProblemeAResoudre**. Indeed, due to its name, this function should not perform that storage.

src/solver/optimisation/opt_construction_contraintes_couts_demarrage.cpp
```diff
@@ -424,16 +566,19 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaireCoutsDeDemarrage(
...
             const int palier
               = PaliersThermiquesDuPays->NumeroDuPalierDansLEnsembleDesPaliersThermiques[index];
+            const auto& clusterName = PaliersThermiquesDuPays->NomsDesPaliersThermiques[index];
 
             for (int pdt = 0; pdt < nombreDePasDeTempsPourUneOptimisation; pdt++)
             {
+                int timeStepInYear = problemeHebdo->weekInTheYear * 168 + pdt;
                 CorrespondanceCntNativesCntOptim
                   = problemeHebdo->CorrespondanceCntNativesCntOptim[pdt];
                 CorrespondanceCntNativesCntOptim
```
guilpier-code, line 463: Not exactly related to this PR : 
Block : 
```C++
if (!Simulation)
{
     CorrespondanceCntNativesCntOptim
          ->NumeroDeContrainteDesContraintesDeDureeMinDArret[palier]
              = -1;
}
```
really seems useless, as the same code is called few lines above.

src/solver/optimisation/opt_construction_contraintes_couts_demarrage.cpp
```diff
@@ -58,11 +61,11 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaireCoutsDeDemarrage(
     int* Colonne = ProblemeAResoudre->Colonne;
 
     int nbTermesContraintesPourLesCoutsDeDemarrage = 0;
-
     for (int pays = 0; pays < problemeHebdo->NombreDePays; pays++)
     {
         const PALIERS_THERMIQUES* PaliersThermiquesDuPays
           = problemeHebdo->PaliersThermiquesDuPays[pays];
+        const auto& zone = problemeHebdo->NomsDesPays[pays];
```
guilpier-code: **zone** should be renamed into **areaName**, sticking better to business naming and to what it really is.

src/libs/antares/study/study.cpp
a-zakir: yes indeed, but if this check is removed then an id is generated from the newName, and it's added (
    // Updating all hydro allocation
    areas.each([&](Data::Area& areait) { areait.hydro.allocation.rename(oldid, newid); });)
    before the call to renameArea

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -281,6 +280,7 @@ struct PALIERS_THERMIQUES
     double* PmaxDUnGroupeDuPalierThermique;
     int* DureeMinimaleDeMarcheDUnGroupeDuPalierThermique;
     int* DureeMinimaleDArretDUnGroupeDuPalierThermique;
+    const char** NomsDesPaliersThermiques;
```
guilpier-code: `const char** NomsDesPaliersThermiques `: why not a **std::vector\<std::string\>** ?

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -417,6 +417,8 @@ void SIM_AllocationProblemeHebdo(PROBLEME_HEBDO& problem, uint NombreDePasDeTemp
           = (int*)MemAlloc(nbPaliers * sizeof(int));
         problem.PaliersThermiquesDuPays[k]->DureeMinimaleDArretDUnGroupeDuPalierThermique
           = (int*)MemAlloc(nbPaliers * sizeof(int));
+        problem.PaliersThermiquesDuPays[k]->NomsDesPaliersThermiques
```
guilpier-code: if **NomsDesPaliersThermiques** was a **std::vector\<std::string\>**, we wouldn't need to delete the related memory

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -293,6 +294,7 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
               = (pbPalier.PmaxDUnGroupeDuPalierThermique[l] < cluster.minStablePower)
                   ? pbPalier.PmaxDUnGroupeDuPalierThermique[l]
                   : cluster.minStablePower;
+            pbPalier.NomsDesPaliersThermiques[l] = cluster.name().c_str();
```
guilpier-code: We could replace index **l** with **clusterIndex**, much more meaningful

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.cpp
guilpier-code: Moreover, as renaming variables has nothing to do with building constraints, we should try to move this renaming to another place, that is where the variables are built : 
- **opt_construction_variables_optimisees_lineaire.cpp**
- **opt_construction_variables_couts_demarrages.cpp**

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -1,12 +1,6 @@
 #include "opt_rename_problem.h"
 #include <sstream>
-void RenameLinkVariable(PROBLEME_ANTARES_A_RESOUDRE* problem,
-                        int var,
-                        Antares::Data::Enum::ExportStructDict structDict,
-                        int ts,
-                        Antares::Data::Enum::ExportStructTimeStepDict timeStepType,
-                        const std::string& origin,
-                        const std::string& destination)
+void VariableNamer::SetLinkVariableName(int var, Antares::Data::Enum::ExportStructDict structDict)
```
guilpier-code: About : **SetLinkVariableName(...)** (and other methods of class **VariableNamer**)
Why do we need an **ExportStructDict** to be passes as arg ?
Could we not use it internally and thus avoid passing this parameter ?

guilpier-code (changes requested):

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -297,28 +293,23 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* pro
...
             CorrespondanceCntNativesCntOptim->NumeroDeContrainteDesBilansPays[pays]
               = ProblemeAResoudre->NombreDeContraintes;
 
             OPT_ChargerLaContrainteDansLaMatriceDesContraintes(
               ProblemeAResoudre, Pi, Colonne, nombreDeTermes, '=');
-
+            constraintNamer.AreaBalance();
             nombreDeTermes = 0;
 
             exportPaliers(*problemeHebdo,
```
guilpier-code: Not related to PR : 
**exportPaliers** is an ill named function : 
- Its name is half-french half-english
- it tells us that we are exporting thermal clusters to somewhere we don't know (could be a file on disk, for instance).

Should be renamed into **insertThermalProductionsOfAreaIntoBC** (for example)

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -57,25 +57,20 @@ void exportPaliers(const PROBLEME_HEBDO& problemeHebdo,
...
-            {
-                OPT_Export_add_variable(varname,
-                                        var,
-                                        Enum::ExportStructDict::PalierThermique,
-                                        timeStepInYear, // TODO[FOM] remove
-                                        pays,
-                                        palier);
-            }
+            const auto& clusterName = PaliersThermiquesDuPays->NomsDesPaliersThermiques[index];
+            variableNamer.DispatchableProduction(var, clusterName);
```
guilpier-code: I can see that **variableNamer** was not moved to another place, is there a difficulty to do it ?

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,115 @@
+#pragma once
+#include "opt_structure_probleme_a_resoudre.h"
+#include "opt_export_structure.h"
+
+const std::string SEPARATOR = "::";
+const std::string ZONE_SEPARATOR = "$$";
```
guilpier-code: **ZONE_SEPARATOR** ==> **AREA_SEP**

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,115 @@
...
+struct ProblemElementInfo
+{
+    PROBLEME_ANTARES_A_RESOUDRE* problem;
+    int timeStep = 0;
+    std::string origin;
+    std::string destination;
+    std::string area;
+};
+class VariableNamer
```
guilpier-code: To me, **Antares::Data::Enum** is cumbersome and hard to read.
Couldn't we have a : 
```C++
using namespace Antares::Data::Enum;
```
at the beginning of the current file ?

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -179,16 +184,17 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* pro
 
     ProblemeAResoudre->NombreDeContraintes = 0;
     ProblemeAResoudre->NombreDeTermesDansLaMatriceDesContraintes = 0;
+    ProblemElementInfo problemElementInfo;
```
guilpier-code: Class **ProblemElementInfo** is useless to me.
If the goal is to be the least intrusive in the current function as possible, we don't need it.
A way to reach this goal would be to change class **VariableNamer** a little bit : 
```C++
class VariableNamer
{
public:
    updateArea(std::string AreaName) { currentAreaName_ = AreaName; }
    updateTimeStep(int timeStep) { currentTimeStep_ = timeStep; };
    
    // And here some renaming functions, as usual.
private:
     std::string currentAreaName_;
     unsigned int currentTimeStep_ = 0; 
};
```
About **origin** and **destination**, as we need both to rename a link variable, we shouldn't hesitate to have : 
```C++
variableNamer.NTCValueOriginToDestination(var, origin, destination);
```

If you feel you need **ProblemElementInfo**, for example because it could be used as a storage of current assets (area, time step, ...) in  **constraintNamer** and **variableNamer**, use it internally. We don't need to expose it. 
In this case, we should rename **ProblemElementInfo** into `class CurrentAssetsStorage` for instance

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,115 @@
+#pragma once
+#include "opt_structure_probleme_a_resoudre.h"
+#include "opt_export_structure.h"
+
+const std::string SEPARATOR = "::";
+const std::string ZONE_SEPARATOR = "$$";
+struct ProblemElementInfo
```
guilpier-code: There is no empty line in this source file (used to separate classes, structs, functions).
Maybe a few empty lines would lighten all this

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
guilpier-code: Still about avoiding the **ProblemElementInfo**, and as I see that **VariableNamer** and **ConstraintNamer** both need to store current assets (area, time step, ...) and the **PROBLEME_ANTARES_A_RESOUDRE**, these current data and the **update** functions could be defined in a parent class named **Namer**.
**VariableNamer** and **ConstraintNamer** could inherit from this class.

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,399 @@
...
+        auto fullName = BuildName(
+          Antares::Data::Enum::toString(structDict),
+          LocationIdentifier(location, Antares::Data::Enum::ExportStructLocationDict::link),
+          TimeIdentifier(problemElementInfo.timeStep,
+                         Antares::Data::Enum::ExportStructTimeStepDict::hour));
+        problemElementInfo.problem->NomDesVariables[var] = fullName;
+    }
+}
+
+void VariableNamer::SetAreaVariableName(int var, Antares::Data::Enum::ExportStructDict structDict)
```
guilpier-code: I can see that, in several functions, we have : 
```C++    
   auto nvars = problemElementInfo.problem->NombreDeVariables;
   if (nvars > var)
   {
       ...
   }
```
I don't know where this check is taken from, but I'm not sure it's useful, as it should have been checked before, no ?

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
a-zakir: I preferred composition with ProblemElementInfo rather than inheritance to avoid duplication of data and line codes: as variable and constraint namer would do things like update area, time step, etc...

src/solver/optimisation/opt_export_structure.h
```diff
@@ -48,13 +48,68 @@ enum class ExportStructDict : unsigned char
 {
```
guilpier-code: **ExportStructDict** : if it's a dictionary, as the name suggest, let's make it a **std::map** instead, rather than this complicated structure. And the name **ExportStructDict** does not help, by the way, we should change it.

src/solver/optimisation/opt_export_structure.h
```diff
@@ -48,13 +48,68 @@ enum class ExportStructDict : unsigned char
...
+{
+    area,
+    link
+};
+enum class ExportStructBindingConstraintType : unsigned char
+{
+    hourly,
+    daily,
+    weekly
 };
```
guilpier-code: It seems like function **OPT_Export_add_variable** is not used anymore. If it's true, let's remove it.

src/solver/optimisation/opt_rename_problem.h
a-zakir: done in .cpp

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,399 @@
+#include "opt_rename_problem.h"
+#include <sstream>
+using namespace Antares::Data;
+void VariableNamer::SetLinkVariableName(int var, Antares::Data::Enum::ExportStructDict structDict)
```
guilpier-code: I don't question the usage of **ExportStructDict** here, but, to me, this structure / Enum should not exist, at least under this form.
Used here, it makes things more complex than it should be.
Do we really need it here ?
For instance, here we could have : 
```C++
void VariableNamer::SetLinkVariableName(int var, std::string variableName)
{
   ...
}
....
void VariableNamer::NTCValueOriginToDestination(int var)
{
    SetLinkVariableName(var, "NTCValueOriginToDestination");
}
```
Instead of that, code in the body of **SetLinkVariableName** is more complex than it should, for example we have `Antares::Data::Enum::toString(structDict)`

And we have the same problem with **ExportStructTimeStepDict**, **ExportStructLocationDict**, **ExportStructTimeStepDict**, ...

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,396 @@
...
+        auto fullName = BuildName(
+          toString(structDict),
+          LocationIdentifier(currentAssetsStorage.area, ExportStructLocationDict::area) + SEPARATOR
+            + "Layer<" + std::to_string(layerIndex) + ">",
+          TimeIdentifier(currentAssetsStorage.timeStep, ExportStructTimeStepDict::hour));
+        currentAssetsStorage.problem->NomDesVariables[var] = fullName;
+    }
+}
+
+void VariableNamer::SetThermalClusterVariableName(int var,
```
guilpier-code: About **SetThermalClusterVariableName** and **SetShortTermStorageVariableName** : these 2 functions seems very similar. What is the difference ? could we not have only one function instead of 2 ?

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,123 @@
...
+    void NegativeUnsuppliedEnergy(int var);
+    void AreaBalance(int var);
+};
+
+class ConstraintNamer
+{
+private:
+    CurrentAssetsStorage& currentAssetsStorage;
+
+    void BindingConstraint(const std::string& name,
```
guilpier-code: About **BindingConstraint** private function, we should rename it **nameWithTimeGranularity**

guilpier-code (changes requested):

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -381,8 +383,10 @@ void SIM_AllocationProblemeHebdo(PROBLEME_HEBDO& problem, uint NombreDePasDeTemp
     {
         const uint nbPaliers = study.areas.byIndex[k]->thermal.list.size();
 
-        problem.PaliersThermiquesDuPays[k]
-          = (PALIERS_THERMIQUES*)MemAlloc(sizeof(PALIERS_THERMIQUES));
+        /* problem.PaliersThermiquesDuPays[k]
```
guilpier-code: Please remove commented code

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,396 @@
...
+void ConstraintNamer::FinalStockExpression()
+{
+    currentAssetsStorage.problem
+      ->NomDesContraintes[currentAssetsStorage.problem->NombreDeContraintes - 1]
+      = BuildName(toString(ExportStructConstraintsDict::FinalStockExpression),
+                  LocationIdentifier(currentAssetsStorage.area, ExportStructLocationDict::area),
+                  TimeIdentifier(currentAssetsStorage.timeStep, ExportStructTimeStepDict::hour));
+}
+
+void ConstraintNamer::BindingConstraint(const std::string& name, ExportStructTimeStepDict type)
```
guilpier-code:
```C++
auto timeStepType
      = (type == ExportStructTimeStepDict::hour)  ? ExportStructBindingConstraintType::hourly
        : (type == ExportStructTimeStepDict::day) ? ExportStructBindingConstraintType::daily
                                                  : ExportStructBindingConstraintType::weekly;
```
Why is it so complicated ?
(Without use of **ExportStructTimeStepDict**, it would be much more simple by the way)

guilpier-code (changes requested):

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
a-zakir: to do in another pr

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -229,16 +237,15 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* pro
...
-                    }
+                    const auto origin
+                      = problemeHebdo
+                          ->NomsDesPays[problemeHebdo->PaysOrigineDeLInterconnexion[interco]];
+                    const auto destination
+                      = problemeHebdo
+                          ->NomsDesPays[problemeHebdo->PaysExtremiteDeLInterconnexion[interco]];
+                    currentAssetsStorage.origin = origin;
+                    currentAssetsStorage.destination = destination;
+                    variableNamer.NTCValueOriginToDestination(var);
```
guilpier-code: About : `variableNamer.NTCValueOriginToDestination(var)`
Method should be renamed into **NTCdirect** and maybe called as follows : 
```C++
variableNamer.NTCdirect(var, origin, destination);
```

guilpier-code (changes requested):

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -108,42 +98,48 @@ static void shortTermStorageLevels(
   double* Pi,
   int* Colonne,
   int nombreDePasDeTempsPourUneOptimisation,
-  int pdt)
+  int pdt,
+  SPConstraintsNamer constraintNamer)
```
flomnes:
```suggestion
  SPConstraintsNamer& constraintNamer)
```

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,372 @@
...
+void VariableNamer::AreaBalance(int var)
+{
+    SetAreaVariableName(var, VariableType::BilansPays);
+}
+
+void ConstraintNamer::FlowDissociation(const std::string& origin, const std::string& destination)
+{
+    problem_->NomDesContraintes[problem_->NombreDeContraintes - 1]
+      = BuildName(toString(ConstraintType::FlowDissociation),
+                  LocationIdentifier(origin + AREA_SEP + destination, LocationType::link),
```
flomnes:
```suggestion
      = BuildName("FlowDissociation",
                  LocationIdentifier(origin + AREA_SEP + destination, LocationType::link),
```

flomnes (changes requested):
You could remove all enums, conversion functions, etc. As @guilpier-code said, it would simplify your code by eliminating the need for boilerplate code.

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,373 @@
...
+    void LayerStorage(int var, int layerIndex) override;
+    void PositiveUnsuppliedEnergy(int var) override;
+    void NegativeUnsuppliedEnergy(int var) override;
+    void AreaBalance(int var) override;
+};
+class EmptyVariableNamer : public IVariableNamer
+{
+public:
+    using IVariableNamer::IVariableNamer;
+    void DispatchableProduction(int var, const std::string& clusterName) override
```
guilpier-code: All these empty functions look like a design problem.
Let's give us time to think about it.

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.h
guilpier-code: A possible way to remove these numerous empty functions would be to isolate and encapsulate the behavior that fills the target vector (**NomDesContraintes** or **NomDesVariables**) with new names, in a very small polymorphic class. 

```c++
// opt_rename_problem.h
...
class BaseNameSender
{
 public:
     void BaseNameSender(std::vector<std::string>& target) : target_(target) { };
     virtual void putNameAtIndex(int index) = 0;
protected:
     std::vector<std::string>& target_; // Reference to either pb->NomDesVariables or pb->NomDesContraintes
}

class DefaultNameSender : public BaseNameSender
{
using BaseNameSender::BaseNameSender;
public:
    void putNameAtIndex(int index, std::string name) override { target_[index] = name; }
}

class NullNameSender : public BaseNameSender
{
using BaseNameSender::BaseNameSender;
public:
    void putNameAtIndex(int index, std::string name) override { }
}
....

class VariableNamer // No more inheritance
{
public:
    VariableNamer(PROBLEME_ANTARES_A_RESOUDRE* problem, 
                  std::shared_ptr<BaseNameSender> nameSender) 
    : nameSender_(nameSender), CurrentAssetsStorage(problem)
    {}
    ...
private:
    ...
    std::shared_ptr<BaseNameSender> nameSender_;
}
```
```c++
// opt_rename_problem.cpp
....
void VariableNamer::SetThermalClusterVariableName(int var,
                                                   const std::string& variableType,
                                                   const std::string& clusterName)
{
    const auto location = LocationIdentifier(area_, AREA) + SEPARATOR + "ThermalCluster" + "<" + clusterName + ">";
    nameSender_->putNameAtIndex(var, BuildName(variableType, location, TimeIdentifier(timeStep_, HOUR));)
}
...

// Here defining a factory for name senders creation.
...
```
```C++
// opt_construction_variables_optimisees_lineaire.cpp
    ...
    auto nameSender = NameSenderFactory(problemeHebdo->NamedProblems);
   std::make_shared<VariableNamer> variableNamer = std::make_shared<VariableNamer>(ProblemeAResoudre, nameSender);
   ...
  ```

File: src/solver/optimisation/opt_rename_problem.h [hunk 43]
```diff
@@ -0,0 +1,155 @@
+#pragma once
+#include "opt_structure_probleme_a_resoudre.h"
+#include "opt_export_structure.h"
+
+const std::string SEPARATOR = "::";
+const std::string AREA_SEP = "$$";
+
+std::string BuildName(const std::string& name,
```
guilpier-code: We don't need the declaration of **BuildName(...)** here.

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,374 @@
+#include "opt_rename_problem.h"
+#include <sstream>
+
+const std::string HOUR("hour");
+const std::string DAY("day");
+const std::string WEEK("week");
+const std::map<std::string, std::string> BindingConstraintTimeGranularity
+  = {{HOUR, "hourly"}, {DAY, "daily"}, {WEEK, "weekly"}};
+const std::string LINK("link");
+const std::string AREA("currentAssetsStorage.area");
```
guilpier-code: About : `const std::string AREA("currentAssetsStorage.area");`
You probably mean : `const std::string AREA("area");`

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+void VariableNamer::SetLinkVariableName(int var, const std::string& variableType)
+{
+    if (name_updater.StringAtIndex(var).empty())
```
guilpier-code: `if (name_updater.StringAtIndex(var).empty())` : why do we have this **if** ?

In this case, we don't need it, we should remove **StringAtIndex(...)** function.

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,155 @@
...
+
+private:
+    std::vector<std::string>& target_;
+    void UpdateTargetAtIndexImpl(const std::string& full_name, unsigned index)
+    {
+        target_[index] = full_name;
+    }
+};
+
+struct CurrentAssetsStorage
```
guilpier-code, line 36: class **CurrentAssetsStorage** reveals to be cumbersome ! 
We should move its content to class **Namer** : 
```C++
class Namer 
{
...
   void UpdateTimeStep(int timeStep)
    {
        current_timeStep_  = timeStep;
    }

    void UpdateArea(const std::string& area)
    {
        current_area_ = area;
    }
protected :        
     int current_timeStep_ = 0;
     std::string current_origin_;
     std::string current_destination_;
     std::string current_area_;
 ...
}
```

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,155 @@
...
+    {
+        currentAssetsStorage.timeStep = timeStep;
+    }
+
+    void UpdateArea(const std::string& area)
+    {
+        currentAssetsStorage.area = area;
+    }
+
+    TargetVectorUpdater name_updater;
```
guilpier-code: `TargetVectorUpdater name_updater;` : should be renamed into : 
`TargetVectorUpdater targetUpdater_;

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.cpp
a-zakir: fixed

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -0,0 +1,141 @@
...
+    {
+        if (isRenamingProcessed)
+        {
+            UpdateTargetAtIndex = std::bind(
+              &TargetVectorUpdater::UpdateTargetAtIndexImpl, this, std::placeholders::_1, std::placeholders::_2);
+        }
+    }
+
+    std::function<void(const std::string&, unsigned index)> UpdateTargetAtIndex
+      = [](const std::string&, unsigned index) {};
```
guilpier-code: `[](const std::string&, unsigned index) {};` : **unsigned** is a type ? Do you mean **unsigned int** ?

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.h
a-zakir: unsigned = unsigned int

guilpier-code (approved):

src/solver/misc/options.cpp
```diff
@@ -169,6 +169,12 @@ std::unique_ptr<GetOpt::Parser> CreateParser(Settings& settings,
                     "mps-export",
                     "Export in the mps format the optimization problems.");
 
+    // --named-problems
+    parser->addFlag(options.namedProblems,
+                    ' ',
+                    "named-problems",
```
```suggestion
                    "named-mps-problems",
```

flomnes (changes requested):
--named-mps-problems

flomnes:
Please replace signed ints by unsigned ints for variable & constraint numbers.

flomnes:
See #1406 for int / unsigned int mix-up complaints by Sonar.

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,364 @@
...
+void VariableNamer::NumberBreakingDownDispatchableUnits(unsigned int var, const std::string& clusterName)
+{
+    SetThermalClusterVariableName(var, "NumberBreakingDownDispatchableUnits", clusterName);
+}
+
+void VariableNamer::NTCDirect(unsigned int var, const std::string& origin, const std::string& destination)
+{
+    origin_ = origin;
+    destination_ = destination;
+    SetLinkVariableName(var, "NTCDirect");
```
flomnes, line 85: For consistency with the rest, you should pass `origin` and `destination` to `SetLinkVariableName`, here and in `VariableNamer::IntercoDirectCost`, `VariableNamer::IntercoIndirectCost`.

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -0,0 +1,364 @@
...
+
+void ConstraintNamer::FlowDissociation(unsigned int numConstraint,
+                                       const std::string& origin,
+                                       const std::string& destination)
+{
+    targetUpdater_.UpdateTargetAtIndex(
+      BuildName("FlowDissociation",
+                LocationIdentifier(origin + AREA_SEP + destination, LINK),
+                TimeIdentifier(timeStep_, HOUR)),
+      numConstraint);
```
flomnes, line 194: Since this constraint seems to be on the link, why not use similar mechanisms as in `VariableNamer` ? Maybe you could factor some stuff since `SetAreaVariableName` & `SetAreaConstraintName` would share a lot

flomnes:
See #1407 for possible refactor

flomnes (approved):

PR #1299, opened 2023-05-11
Remove parameters.hxx, fix Parameters::memoryUsage

flomnes:
.hxx files should be used only for templated classes/functions.


src/libs/antares/study/parameters.cpp
```diff
@@ -1141,6 +1156,11 @@ void Parameters::fixBadValues()
         nbTimeSeriesSolar = 1;
 }
 
+Yuni::uint64 Parameters::memoryUsage() const
```
payetvin, line 1159: only used in src/libs/antares/study/study.cpp:216

Should be removed if it only returns 0

flomnes: We could return something like `sizof(Parameters)`, add the size of pointed members, vectors, arrays, etc.

flomnes: done !

payetvin (approved):

PR #1302, opened 2023-05-12
Fix memory errors detected by valfgrind

payetvin:
- Use of uninitialised value of size 8
- Conditional jump or move depends on uninitialised value(s)


src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -463,7 +463,7 @@ void SIM_AllocationProblemeHebdo(PROBLEME_HEBDO& problem, int NombreDePasDeTemps
         problem.ResultatsHoraires[k].ValeursHorairesDeDefaillancePositive
           = (double*)MemAlloc(NombreDePasDeTemps * sizeof(double));
         problem.ResultatsHoraires[k].ValeursHorairesDENS
-          = (double*)MemAlloc(NombreDePasDeTemps * sizeof(double)); // adq patch
```
JasonMarechal25, line 466: Would it be possible to go one step further and not use malloc but new/delete or even self manage structures (array, vector, etc.) ?

flomnes (approved):

PR #1303, opened 2023-05-12
 Use ISO8601 for date format in the logs

### LINKED ISSUE #1300: Use ISO8601 for date format in the logs
%Y-%m-%d %H:%M:%S

See **src/ext/yuni/src/yuni/core/logs/decorators/time.cpp**

payetvin:
close #1300

### DIFF
--- a/src/ext/yuni/src/yuni/core/logs/decorators/time.cpp
+++ b/src/ext/yuni/src/yuni/core/logs/decorators/time.cpp
@@ -15,8 +15,6 @@
 #endif
 #include "time.h"
 
-#include <iostream>
-
 namespace Yuni
 {
 namespace Private
@@ -27,7 +25,7 @@ namespace LogsDecorator
 
 char* WriteCurrentTimestampToBufferMinGW()
 {
-    char[20] buff;
+    char[32] buff;
     time_t rawtime;
     ::time(&rawtime);
     ::strftime(buff, 32, "%F %T",  ::localtime(&rawtime));
--- a/src/ext/yuni/src/yuni/core/logs/decorators/time.cpp
+++ b/src/ext/yuni/src/yuni/core/logs/decorators/time.cpp
@@ -15,6 +15,8 @@
 #endif
 #include "time.h"
 
+#include <iostream>
+
 namespace Yuni
 {
 namespace Private
@@ -25,15 +27,18 @@ namespace LogsDecorator
 
 char* WriteCurrentTimestampToBufferMinGW()
 {
+    char[20] buff;
     time_t rawtime;
     ::time(&rawtime);
+    ::strftime(buff, 32, "%F %T",  ::localtime(&rawtime));
+
 
     /* MinGW */
     // Note that unlike on (all?) POSIX systems, in the Microsoft
     // C library locatime() and gmtime() are multi-thread-safe, as the
     // returned pointer points to a thread-local variable. So there is no
     // need for localtime_r() and gmtime_r().
-    return ::asctime(::localtime(&rawtime));
+    return ::buff);
 }
 
 #else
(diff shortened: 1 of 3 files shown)


src/ext/yuni/src/yuni/core/logs/decorators/time.cpp
```diff
@@ -25,15 +25,18 @@ namespace LogsDecorator
...
+    ::strftime(buff, 32, "%F %T",  ::localtime(&rawtime));
+
 
     /* MinGW */
     // Note that unlike on (all?) POSIX systems, in the Microsoft
     // C library locatime() and gmtime() are multi-thread-safe, as the
     // returned pointer points to a thread-local variable. So there is no
     // need for localtime_r() and gmtime_r().
-    return ::asctime(::localtime(&rawtime));
+    return buff;
```
flomnes: Dangling reference to a variable allocated on the stack. I suggest passing buff to `WriteCurrentTimestampToBufferMinGW` instead.

flomnes (changes requested):

src/ext/yuni/src/yuni/core/logs/decorators/time.h
```diff
@@ -44,9 +44,9 @@ class YUNI_DECL Time : public LeftType
 #ifndef YUNI_OS_MINGW
         char asc[32]; // MSDN specifies that the buffer length value must be >= 26 for validity
         Private::LogsDecorator::WriteCurrentTimestampToBuffer(asc);
-        out.write(asc, 24);
+        out.write(asc, 19);
 #else
-        out.write(Private::LogsDecorator::WriteCurrentTimestampToBufferMinGW(), 24);
+        out.write(Private::LogsDecorator::WriteCurrentTimestampToBufferMinGW(), 20);
```
guilpier-code: **2023-05-17 14:03:18** contains 19 characters, not 20. 
It was the reason why CI failed on Windows.
Don't know why it does not fail on Linux, probably because **nul** chars are treated differently on Linux.

guilpier-code (dismissed):

src/ext/yuni/src/yuni/core/logs/decorators/time.cpp
```diff
@@ -46,22 +30,20 @@ void WriteCurrentTimestampToBuffer(char buffer[32])
...
 
-    struct tm timeinfo;
 
 #if defined(YUNI_OS_MSVC)
+    struct tm timeinfo;
     // Microsoft Visual Studio
     _localtime64_s(&timeinfo, &rawtime);
     // MSDN specifies that the buffer length value must be >= 26 for validity.
-    asctime_s(buffer, 32, &timeinfo);
+    strftime(buffer, 32, "%F %T", &timeinfo);
```
flomnes, line 39:
```suggestion
    strftime(buffer, 32, "%FT%T", &timeinfo);
```

src/ext/yuni/src/yuni/core/logs/decorators/time.cpp
```diff
@@ -46,22 +30,20 @@ void WriteCurrentTimestampToBuffer(char buffer[32])
...
+
 #else
     // Unixes
-    ::localtime_r(&rawtime, &timeinfo);
-    ::asctime_r(&timeinfo, buffer);
+    ::strftime(buffer, 32, "%F %T",  ::localtime(&rawtime));
```
flomnes, line 43:
```suggestion
    ::strftime(buffer, 32, "%FT%T",  ::localtime(&rawtime));
```

flomnes (commented):
The timezone specifier is missing. If we used UTC, we could simply add "Z" at the end of the string.

However, we use the **local time**, so we'd have to append "+1" or "+2" depending on the period. The problem is that there seems to be no portable way to determine this in C/C++.

https://stackoverflow.com/questions/13804095/get-the-time-zone-gmt-offset-in-c

See https://fr.wikipedia.org/wiki/ISO_8601

flomnes (approved):

PR #1306, opened 2023-05-15
Performance metrics

payetvin:
Add different performance metrics in order to monitor changes throughout versions
- Memory consumption / profiling
- Execution speed

Goal is to integrate this work into CI and store results to compare them.


src/tests/performance/performance.py
```diff
@@ -0,0 +1,43 @@
...
+    study_list[1].run_valgrind_massif(solver_path)
+
+def valgrind_long(solver_path):
+    study_list[2].run_valgrind_base(solver_path)
+    study_list[2].run_valgrind_massif(solver_path)
+
+
+valgrind_short("/home/payetvin/Antares_Simulator/_build_debug/solver/antares-8.6-solver")
+valgrind_medium("/home/payetvin/Antares_Simulator/_build_debug/solver/antares-8.6-solver")
+valgrind_long("/home/payetvin/Antares_Simulator/_build_debug/solver/antares-8.6-solver")
```
flomnes: Needs to be changed

src/tests/performance/performance.py
```diff
@@ -0,0 +1,43 @@
...
+
+from pathlib import Path
+
+class StudyList(object):
+
+    def __init__(self, path):
+        self.path = Path(path).resolve()
+
+    def run_valgrind_base(self, exe):
+        valgrindOutFile = "--log-file=valgrind.out"
```
flomnes: This will be overwritten in case of multiple executions

flomnes (commented):
What do you intend to do with the produced files ?

src/tests/performance/performance.py
```diff
@@ -0,0 +1,59 @@
...
+        assert (res.returncode == 0), "The exec failed for study: " + str(self.path)
+
+    def get_memory(self):
+        ps = subprocess.run(["sed -n -e 's/^.*Maximum.*: //p' " + self.outFile], shell=True, capture_output=True)
+        return int(ps.stdout)
+
+    def get_time(self):
+        ps = subprocess.run(["sed -n -e 's/^.*User time.*: //p' " + self.outFile], shell=True, capture_output=True)
+        ps2 = subprocess.run(["sed -n -e 's/^.*System time.*: //p' " + self.outFile], shell=True, capture_output=True)
+        return float(ps.stdout) + float(ps2.stdout)
```
flomnes: You could run only once. You just need the "User time". Also, do the parsing in Python, don't use sed.

src/tests/performance/performance.py
```diff
@@ -0,0 +1,96 @@
...
+    f.close()
+    m = re.search(pattern, s)
+    if m:
+        return m.groups(1)[1]
+    else:
+        print("Pattern not found, aborting")
+        exit(1)
+
+def get_git_revision_short_hash() -> str:
+    return subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD']).decode('ascii').strip()
```
flomnes:
```suggestion
    return subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('ascii').strip()
```
Let us keep the whole commit-id, to avoid possible collisions.

src/tests/performance/performance.py
```diff
@@ -0,0 +1,96 @@
...
+def get_latest_output_folder(path):
+    output_path_study = os.path.join(path, "output")
+    files = os.listdir(output_path_study)
+    paths = [os.path.join(output_path_study, basename) for basename in files]
+    return max(paths, key=os.path.getctime)
+
+def search_patern_in_file(file_name, pattern):
+    f = open(file_name)
+    s = f.read()
+    f.close()
```
flomnes: `with open(file_name) as f:`

https://docs.python.org/fr/3/reference/compound_stmts.html#the-with-statement

src/tests/performance/performance.py
```diff
@@ -0,0 +1,96 @@
...
+    m = re.search(pattern, s)
+    if m:
+        return m.groups(1)[1]
+    else:
+        print("Pattern not found, aborting")
+        exit(1)
```
flomnes: Better to `throw` an exception and let the caller decide what to do in this case

src/tests/performance/performance.py
```diff
@@ -0,0 +1,96 @@
...
+    # Execution and results for each studies
+    for studies in study_list:
+        studies.run_metrics(solver_path)
+        results[studies.name] = studies.create_json()
+        results[studies.name]['execution_info.ini'] = studies.read_execution_ini()
+
+    # Writing the JSON
+    f = open("results.json", "w")
+    f.write(json.dumps(results))
+    f.close()
```
flomnes: Use `with`

src/tests/performance/performance.py
```diff
@@ -0,0 +1,96 @@
...
+
+class StudyList(object):
+
+    def __init__(self, name, path):
+        self.name = name
+        self.path = Path(path).resolve()
+        self.out_file = name + "-metrics.txt"
+
+    def run_metrics(self, exe):
+        res = subprocess.run(["/bin/time" , "-o", self.out_file, "-v", exe, self.path])
```
flomnes: This won't work in Windows. But it should be OK, it seems like there is no portable way to get peak memory usage

src/tests/performance/performance.py
```diff
@@ -0,0 +1,96 @@
...
+        assert (res.returncode == 0), "The exec failed for study: " + str(self.path)
+
+    def get_memory(self):
+        return int(search_patern_in_file(self.out_file, "Maximum(.*): (.*)"))
+
+    def get_time(self):
+        user_time = float(search_patern_in_file(self.out_file, "User time(.*): (.*)"))
+        syst_time = float(search_patern_in_file(self.out_file, "System time(.*): (.*)"))
+        return user_time + syst_time
```
flomnes: I think that `Elapsed` (wall/real time) would be a better measurement.

src/tests/performance/performance.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def create_json(self):
+        memory = self.get_memory() / 1024 # from Kb to Mb
+        time = self.get_time()
+
+        data = {}
+        data['peak_memory_mb'] = memory
+        data['time_s'] = time
```
flomnes:
```suggestion
        return {'peak_memory_mb': memory, 'time_s' : time}
```

flomnes (changes requested):
Can you post an example of a generated JSON ? Or even add it into a small documentation with instructions on how to start this utility.

payetvin:
JSON output:
```
{"commit-id": "076126aa", "date": "05/25/23 15:45:49", "short": {"peak_memory_mb": 150.58984375, "time_s": 1.58, "execution_info.ini (time in ms)": {"hydro_ventilation": "4", "mc_years": "1086", "post_processing": "0", "study_loading": "12", "synthesis_export": "434", "total": "1567", "tsgen_hydro": "0", "tsgen_thermal": "0", "tsgen_wind": "0"}}, "medium": {"peak_memory_mb": 718.04296875, "time_s": 9.79, "execution_info.ini (time in ms)": {"hydro_ventilation": "35", "mc_years": "6666", "post_processing": "2784", "study_loading": "172", "synthesis_export": "0", "total": "9758", "tsgen_hydro": "109", "tsgen_load": "703", "tsgen_solar": "2577", "tsgen_thermal": "87", "tsgen_wind": "1123"}}}
```

src/tests/performance/performance.py
payetvin: Complex to get with regex as the ":" in the hour interfere

flomnes (approved):

PR #1307, opened 2023-05-16
Add precisions for mingen

flomnes:


MartinBelthle (approved):
Perhaps specify that this new file is created empty by default when a new area is created.

PR #1315, opened 2023-05-24
ST storage : fix a bug on time series

guilpier-code:
Before this fix, the first week of time series were used for all simulation weeks.
For each week, we find a way of using the range of data related to the week, in all ST storage time series.
Time series are :  
- [x] maxInjectionModulation
- [x] maxWithdrawalModulation
- [x] inflows
- [x] lowerRuleCurve
- [x] upperRuleCurve


flomnes (changes requested):
Create a function to compute `weekInTheYear * 168 + pdtHebdo` and use it here & elsewhere ? (look for  `weekInTheYear` in src/solver)

flomnes (approved):

PR #1332, opened 2023-06-05
Add spread on ST storage objects

flomnes:
Sample MPS file
```
    DispatchableProduction::area<area*4>::ThermalCluster<cluster-costly>::hour<3>  FictiveLoads::area<area*4>::hour<3>  -1.0000000000
    ShortTermStorageInjection::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  OBJECTIF  0.0008399490
    ShortTermStorageInjection::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  AreaBalance::area<area*4>::hour<3>  1.0000000000
    ShortTermStorageInjection::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  ShortTermStorageLevel::area<area*4>::hour<3>  -0.9000000000
    ShortTermStorageWithdrawal::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  OBJECTIF  0.0007559541
    ShortTermStorageWithdrawal::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  AreaBalance::area<area*4>::hour<3>  -1.0000000000
    ShortTermStorageWithdrawal::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  ShortTermStorageLevel::area<area*4>::hour<3>  1.0000000000
    ShortTermStorageLevel::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  ShortTermStorageLevel::area<area*4>::hour<2>  1.0000000000
    ShortTermStorageLevel::area<area*4>::ShortTermStorage<my-sts-4>::hour<3>  
```


src/solver/simulation/sim_st_storage_draw_spread.cpp
```diff
@@ -0,0 +1,42 @@
...
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+
+#include <antares/mersenne-twister/mersenne-twister.h>
+#include <antares/constants.h>
+
+#include "sim_st_storage_draw_spread.h"
+
+#define EPSILON_ST_STORAGE 1e-3
```
guilpier-code: **EPSILON_ST_STORAGE** : why not a **const** instead of a pre-processing value ?

src/solver/optimisation/opt_gestion_des_couts_cas_lineaire.cpp
```diff
@@ -59,14 +60,14 @@ static void shortTermStorageCost(
             = CorrespondanceVarNativesVarOptim->SIM_ShortTermStorage.InjectionVariable[clusterGlobalIndex];
             varInjection >= 0)
         {
-            linearCost[varInjection] = 0;
+            linearCost[varInjection] = storage.spreadOnInjectionWithdrawal[hourInTheYear];
```
guilpier-code: About : 
`linearCost[varInjection] = storage.spreadOnInjectionWithdrawal[hourInTheYear];`

We know that we add a a spread on the injection here, so, to me, the following is enough : 
```C++
linearCost[varInjection] = storage.spread[hourInTheYear];
```
By the way, when you generated spreads, you called function **SIM_drawSpreadOnShortTermStorage**, without knowing if they apply to injection or withdrawal.

guilpier-code: We could even have : 
```C++
linearCost[varInjection] = shortTermStorageInput.spread.generate();
```
**shortTermStorageInput.spread** would be an object of class : 
```c++
const double EPSILON_ST_STORAGE = 1e+3;

class spread
{
public:
    spread() // Default constr
    {
        mt_.reset(seed_);
    }
    double generate() { return mt_.next() * EPSILON_ST_STORAGE;} 
private:
    int seed_ = ... ; // Whatever value
    MersenneTwister mt_;
}
```
The idea would be to have one spread object for all ST storage.

This would : 
- Avoid storing all spreads in a vector, and make RAM grow
- Simplify client code (we would not need **hourInTheYear** anymore, so an argument vanishes)
- We probably would need one seed only
- The spread class could maybe reused in another context
- In SIM, the call to **SIM_drawSpreadOnShortTermStorage** would disappear.

guilpier-code (changes requested):

src/solver/optimisation/opt_gestion_des_couts_cas_lineaire.cpp
flomnes: Good idea, will do

src/solver/simulation/sim_st_storage_draw_spread.cpp
flomnes: used `constexpr`

src/solver/optimisation/opt_gestion_des_couts_cas_lineaire.cpp
```diff
@@ -41,32 +42,50 @@
...
-        if (const int varLevel
-            = CorrespondanceVarNativesVarOptim->SIM_ShortTermStorage.LevelVariable[clusterGlobalIndex];
-            varLevel >= 0)
+        for (const auto& storage : shortTermStorageInput[pays])
         {
-            linearCost[varLevel] = 0;
-        }
+            // We use the global index as a seed to avoid identical draws for multiple storages
+            // which may lead to equivalent solutions if 2 storages are identical
+            spreadGenerator.reset(storage.clusterGlobalIndex);
```
guilpier-code: About : `spreadGenerator.reset(storage.clusterGlobalIndex);`
Really sorry but I still do not understand why we need to reset the seed for each ST storage.
I may miss something but if we use the same **spreadGenerator** for each ST storage and we just increment the spread without resetting the seed, we'll have different hourly spreads for each ST storage, won't we ?

guilpier-code (dismissed):

src/solver/optimisation/opt_gestion_des_couts_cas_lineaire.cpp
flomnes: Yes, I agree. Let's drop this !

payetvin (approved):

PR #1347, opened 2023-06-09
STS don't save series if they're default values

### LINKED ISSUE #1337: ST storage : save default time series on disk
When loading a study with at least a ST storage : 
For time series that does not exist, some default time series are built internally.

In the same spirit, when saving a study with ST storage, if a ST storage has time series with the default values,
it's all saved on disk, taking up unnecessary space.

Further more, reloading the study takes unnecessary load time.

Should we not avoid saving on disk default TS ?

payetvin:
close #1137

### DIFF
--- a/src/libs/antares/study/parts/short-term-storage/series.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/series.cpp
@@ -127,15 +127,21 @@ bool Series::saveToFolder(const std::string& folder) const
 
     bool ret = true;
 
-    auto checkWrite = [&ret, &folder](const std::string& name, const std::vector<double>& content) {
-        ret = writeVectorToFile(folder + SEP + name, content) && ret;
+    auto checkWrite = [&ret, &folder]
+        (const std::string& name, const std::vector<double>& vect, double defValue)
+    {
+        // if vector is only default values we don't save it
+        if (all_of(vect.begin(), vect.end(), [&defValue] (const double& i) { return i == defValue; }))
+            return;
+
+        ret = writeVectorToFile(folder + SEP + name, vect) && ret;
     };
 
-    checkWrite("PMAX-injection.txt", maxInjectionModulation);
-    checkWrite("PMAX-withdrawal.txt", maxWithdrawalModulation);
-    checkWrite("inflows.txt", inflows);
-    checkWrite("lower-rule-curve.txt", lowerRuleCurve);
-    checkWrite("upper-rule-curve.txt", upperRuleCurve);
+    checkWrite("PMAX-injection.txt", maxInjectionModulation, 1.0);
+    checkWrite("PMAX-withdrawal.txt", maxWithdrawalModulation, 1.0);
+    checkWrite("inflows.txt", inflows, 0.0);
+    checkWrite("lower-rule-curve.txt", lowerRuleCurve, 0.0);
+    checkWrite("upper-rule-curve.txt", upperRuleCurve, 1.0);
 
     return ret;
 }
@@ -195,7 +201,6 @@ bool Series::validateMaxInjection() const
 bool Series::validateMaxWithdrawal() const
 {
     return checkVectBetweenZeroOne(maxWithdrawalModulation, "PMAX withdrawal");
-
 }
 
 bool Series::validateRuleCurves() const
--- a/src/libs/antares/study/parts/short-term-storage/series.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/series.cpp
@@ -138,7 +138,7 @@ bool Series::saveToFolder(const std::string& folder) const
     {
         // if vector is only default values we don't save it
         if (!hasOnlyDefaultValues(timeSeries, defValue))
-            ret = writeVectorToFile(folder + SEP + filename, timeSeries) && ret;
+            ret = writeTimeSeriesToFile(folder + SEP + filename, timeSeries) && ret;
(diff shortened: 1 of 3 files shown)


src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -127,15 +127,21 @@ bool Series::saveToFolder(const std::string& folder) const
 
     bool ret = true;
 
-    auto checkWrite = [&ret, &folder](const std::string& name, const std::vector<double>& content) {
-        ret = writeVectorToFile(folder + SEP + name, content) && ret;
+    auto checkWrite = [&ret, &folder]
```
guilpier-code, line 136: About **checkWrite** : 
This function does 2 things : it checks and it saves to folder.
Wouldn't it be more clear if we had : 
```C++
if(! hasOnlyDefaultValues(maxInjectionModulation, 1.0))
    ret = saveTimeSeriesToFile(maxInjectionModulation, folder + SEP + "PMAX-injection.txt") && ret;

// Here : same for other time series 
``` 
?
Note that each functions has less arguments (=2), instead of 5 for **checkWrite** (if we count lambda capture)

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -127,15 +127,21 @@ bool Series::saveToFolder(const std::string& folder) const
...
+        (const std::string& name, const std::vector<double>& vect, double defValue)
```
guilpier-code: Renaming : 
`const std::string& name` should be renamed into **filename**
`const std::vector<double>& vect` should be renamed into **timeSeries**

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -127,15 +127,22 @@ bool Series::saveToFolder(const std::string& folder) const
...
-        ret = writeVectorToFile(folder + SEP + name, content) && ret;
+    auto checkWrite = [&ret, &folder]
+        (const std::string& filename, const std::vector<double>& timeSeries, double defValue)
+    {
+        // if vector is only default values we don't save it
+        if (all_of(timeSeries.begin(), timeSeries.end(), [&defValue] (const double& i)
+                    { return i == defValue; }))
+            return;
+
+        ret = writeVectorToFile(folder + SEP + filename, timeSeries) && ret;
```
flomnes: Please use std::find to slightly simplify this. Also, you don't need `return`, just add `!` in the `if` block

flomnes (changes requested):

src/libs/antares/study/parts/short-term-storage/series.cpp
payetvin: remove the lambda means having a one line function with those args and needing:
`saveTimeSeriesToFile(const std::string& folder, const std::string& filename, const std::vector<double>& timeSeries, double defValue)`
and will need to handle ret on the other function like this:
`ret = saveTimeSeriesToFile() && ret`
This was why we used a lambda in the first place

payetvin: Or we can duplicate code and have 5*2 lines doing the same thing with different var

flomnes:
This change will break Antares's convention (empty file = all 0).

To be discussed

src/libs/antares/study/parts/short-term-storage/series.cpp
guilpier-code: I guess you're concerned about duplication and concision.
I'm more concerned about clarity, short function making one thing with small number of arguments.
**checkWrite** makes 2 things (and it's in the name **checkAndWrite**), it's usually considered as a code smell.
Let's discuss it.

src/libs/antares/study/parts/short-term-storage/series.cpp
```diff
@@ -127,15 +133,19 @@ bool Series::saveToFolder(const std::string& folder) const
...
     bool ret = true;
 
-    auto checkWrite = [&ret, &folder](const std::string& name, const std::vector<double>& content) {
-        ret = writeVectorToFile(folder + SEP + name, content) && ret;
+    auto checkWrite = [&ret, &folder]
+        (const std::string& filename, const std::vector<double>& timeSeries, double defValue)
+    {
+        // if vector is only default values we don't save it
+        if (!hasOnlyDefaultValues(timeSeries, defValue))
+            ret = writeVectorToFile(folder + SEP + filename, timeSeries) && ret;
```
guilpier-code: Function **writeVectorToFile** should be renamed into **writeTimeSeriesToFile**.

guilpier-code (changes requested):

flomnes:
To match the default behavior :
- All files are mandatory
- An empty file means only "0"

Both for input & output.

src/libs/antares/study/parts/short-term-storage/series.cpp
JasonMarechal25: To go further we should extract the behaviour to read/write in file in dedicated objects. To go furtherer, have these new implementations be a "file" implementation of interfaces and either
- inject instances to "series" constructor and call them when saving/loading
- Have "series" be built from the new objects.

PR #1354, opened 2023-06-12
GUI : adapt logs reading from GUI

guilpier-code:
Because we recently change the date format in logs, we currently face problems in the GUI : 
- [x]  Fixing the logs loader (button on left of inspector). The logs loader is currently empty whereas log file is not.
- [x] Fixing the errors/warnings catch when simulation ends. The errors/warnings are not currently caught.

### DIFF
--- a/src/ui/simulator/toolbox/jobs/job.cpp
+++ b/src/ui/simulator/toolbox/jobs/job.cpp
@@ -619,7 +619,7 @@ Job::Job(const wxString& title, const wxString& subTitle, const char* icon) :
  pThread(nullptr),
  pAnim(nullptr),
  pLblErrors(nullptr),
- pLogRegex(wxT("\\[([a-zA-Z0-9 :]+)\\]\\[([a-zA-Z0-9 "
+ pLogRegex(wxT("\\[([0-9 :-]+)\\]\\[([a-zA-Z0-9 "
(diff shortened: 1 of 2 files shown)


flomnes (changes requested):
This fix works well for logs produced by v8.6.0+, but fails to display older logs.

Could we keep compatibility at an acceptable development cost ?

Conceptually, you did the following
```diff
- bool LogIsValidLegacy(const std::string& line)
+ bool LogIsValid_860(const std::string& line)
```
You could keep both functions, and then do something like
```cpp
if (LogIsValid_860(line)) 
{
   parseLog_860(line);
} else if (LogIsValidLegacy(line)) 
{
   parseLogLegacy(line);
}
```
in other words, keep the current regex/parser as a fallback alternative if parsing the new format fails.

guilpier-code:
Answering the previous comment.
If we open a study whose version is, let's say **v8.3.0** with the current version IHM (v8.6.0) and run the study, we expect to have logs in the new format. 
My question : wouldn't it be useless to keep the reading of the old format ?

About the logs loader : I can understand we would like to load logs in the old format, when loading an old study with v8.6.0.
But do we really want to do that, given the facts : 
- This use case may not happen every day 
- that historic IHM is about to disappear.

?

flomnes:
> and run the study, we expect to have logs in the new format.

No, we could have run the study with a previous version of antares.

I agree that it's not a very strict requirement, more of a "nice to have".

flomnes (approved):

PR #1358, opened 2023-06-14
Remove binding constraintsRTI

### LINKED ISSUE #1323: BC scenarisation : remove class `Data::BindingConstraintRTI`
It is a mirror of `Data::BindingConstraint`.

Previously 1 ranged binding constraint = 2 BindingConstraintRTI

See **src/libs/antares/study/runtime/runtime.cpp**.
```cpp
        case BindingConstraint::opBoth:
        {
            CopyBCData<BindingConstraint::columnInferior>(rti, constraint);
            ++index;
            bindingConstraint[index].type = constraint.type();
            CopyBCData<BindingConstraint::columnSuperior>(bindingConstraint[index], constraint);
            break;
        }
```
This is now dead code. Two `Data::BindingConstraint` are created upon loading (before, they were created in runtime.cpp).

JasonMarechal25:


src/libs/antares/benchmarking/info_collectors.cpp
```diff
@@ -72,12 +72,15 @@ void StudyInfoCollector::enabledThermalClustersCountToFileContent(FileContent& f
 
 void StudyInfoCollector::enabledBindingConstraintsCountToFileContent(FileContent& file_content)
 {
-    unsigned int nbEnabledBC = study_.runtime->bindingConstraints.size();
-    unsigned int nbEnabledHourlyBC(0), nbEnabledDailyBC(0), nbEnabledWeeklyBC(0);
+    auto enabledBindingConstraints = study_.bindingConstraints.enabled();
+    auto nbEnabledBC = enabledBindingConstraints.size();
+    auto nbEnabledHourlyBC(0);
```
flomnes: Unsigned int

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -483,7 +352,6 @@ bool StudyRuntimeInfos::loadFromStudy(Study& study)
     logs.info() << "     thermal clusters: " << thermalPlantTotalCount;
     logs.info() << "     thermal clusters (must-run): " << thermalPlantTotalCountMustRun;
     logs.info() << "     short-term storages: " << shortTermStorageCount;
-    logs.info() << "     binding constraints: " << bindingConstraints.size();
```
flomnes, line 486: This info is actually quite useful, please keep it if possible

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -244,23 +245,23 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
...
+            PtMat->OffsetTemporelSurLInterco[j] = bc->linkOffset()[j];
         }
 
-        for (uint j = 0; j < bc.clusterCount; ++j)
+        for (uint j = 0; j < bc->clusterCount(); ++j)
         {
-            PtMat->NumeroDuPalierDispatch[j] = bc.clusterIndex[j];
-            PtMat->PaysDuPalierDispatch[j] = bc.clustersAreaIndex[j];
-            PtMat->PoidsDuPalierDispatch[j] = bc.clusterWeight[j];
+            PtMat->NumeroDuPalierDispatch[j] = bc->clusterIndex()[j];
```
flomnes: clusterIndex(j) ?

src/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
```diff
@@ -154,7 +154,7 @@ class BindingConstMarginCost
             pValuesForTheCurrentYear[numSpace].initializeFromStudy(study);
 
         // Set the associated binding constraint
-        associatedBC_ = &(study.runtime->bindingConstraints[bindConstraintGlobalIndex_]);
+        associatedBC_ = study.bindingConstraints.enabled()[bindConstraintGlobalIndex_];
```
flomnes, line 157: Matbe you could compute the list of enabled BCs once, for example after loading them.

flomnes (dismissed):

src/solver/simulation/sim_calcul_economique.cpp
JasonMarechal25: clusterIndex_ is private. Could be public if deemed better

src/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
JasonMarechal25: Yes. I had it in mind but forgot. Thank you

src/libs/antares/study/runtime/runtime.cpp
JasonMarechal25: Done

src/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
JasonMarechal25: The value is now cached

src/solver/simulation/sim_calcul_economique.cpp
flomnes: Yes, clusterIndex could be a getter

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.h
```diff
@@ -179,6 +184,8 @@ class BindingConstraintsRepository final : public Yuni::NonCopyable<BindingConst
     [[nodiscard]] Yuni::uint64 timeSeriesNumberMemoryUsage() const;
 
     [[nodiscard]] bool checkTimeSeriesWidthConsistency() const;
+
+    mutable std::optional<std::vector<std::shared_ptr<BindingConstraint>>> enabledConstraints_;
```
guilpier-code, line 193: About private member **enabledConstraints_** : 
I don't understand the point of this data member.
It's used only in **enabled()**, but even in this function, we could skip it.
And it makes the currant class more complicated than it should.

src/libs/antares/study/binding_constraint/BindingConstraint.h
```diff
@@ -393,17 +393,18 @@ class BindingConstraint final : public Yuni::NonCopyable<BindingConstraint>
     */
     void buildFormula(YString& s) const;
```
guilpier-code, line 380: About type `Data::BindingConstraint::Vector` : 
It is defined in the current source file, but it more surely belongs to : **src/libs/antares/study/binding_constraint/BindingConstraintsRepository.h**

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -219,16 +220,16 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
         problem.PaysExtremiteDeLInterconnexion[i] = link.with->index;
     }
 
-    for (uint i = 0; i < study.runtime->bindingConstraints.size(); ++i)
+    for (uint i = 0; i < enabledBindingConstraints.size(); ++i)
```
guilpier-code, line 223: About `for (uint i = 0; i < enabledBindingConstraints.size(); ++i)` : 
why not `for(auto& bc : enabledBindingConstraints)` instead ?
We would avoid the below `auto bc = enabledBindingConstraints[i];`

src/libs/antares/study/binding_constraint/BindingConstraint.cpp
```diff
@@ -569,40 +564,51 @@ int BindingConstraint::offset(const ThermalCluster* lnk) const
...
 }
 
-void BindingConstraint::initLinkArrays(std::vector<double>& w,
-                                       std::vector<double>& cW,
-                                       std::vector<int>& o,
-                                       std::vector<int>& cO,
-                                       std::vector<long>& linkIndex,
-                                       std::vector<long>& clusterIndex,
-                                       std::vector<long>& clustersAreaIndex) const
+void BindingConstraint::initLinkArrays() const
```
guilpier-code: About **initLinkArrays()** : 
This function does not do what its name announces.
There is **Link** in the name but it's not about links only, but also clusters.
Besides, the name means nothing to me : which **Arrays** ? Is it really about init ? (not sure)
If we want to know what the function does, we must have a look at its body.

I recommend to split it into 2 parts, one for links, one for clusters, and rename this function  **rearrangeProperties** (for instance), because it's it takes existing BC properties and store them in another way.

So a subsequent question arises :  why do we still need **private** member data **pClusterWeights** and **pLinkWeights** if they haven't the right shape in the first place ?
When we populate them, let's directly populate the new private member data **linkIndex_**, **linkWeight_**, ..., and remove **initLinkArrays()** function.

src/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
```diff
@@ -392,7 +393,7 @@ class BindingConstMarginCost
     //! Intermediate values for each year
     typename VCardType::IntermediateValuesType pValuesForTheCurrentYear = nullptr;
     unsigned int pNbYearsParallel = 0;
-    Data::BindingConstraintRTI* associatedBC_ = nullptr;
+    std::shared_ptr<Data::BindingConstraint> associatedBC_ = nullptr;
```
guilpier-code, line 396: About `associatedBC_ = nullptr;` : 
Do we need to initialize a shared pointer to **nullptr** ? 
It's its default value anyway, isn't it ?

guilpier-code (changes requested):

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.h
JasonMarechal25: It's a cached value. Prevent recalculating the list every time.

src/solver/simulation/sim_calcul_economique.cpp
JasonMarechal25: Probably a relicate of previous development where bindingConstraintsCount was used

src/libs/antares/study/binding_constraint/BindingConstraint.h
JasonMarechal25: 8162758304216832d2d72671778d745e79caca79

src/solver/simulation/sim_calcul_economique.cpp
JasonMarechal25: 1. Unforntunatly we need the index for `PtMat = problem.MatriceDesContraintesCouplantes[i];`

src/libs/antares/study/binding_constraint/BindingConstraint.cpp
JasonMarechal25: I've somewhat changed the behaviour of the method in 9e98678ecdb1cfd7b90cafbaa8f962c6bdb42fd7
I however agree that the name could be better.
I'm not even sure it should be defined in BindingConstraint. It probably belong more to SIM_InitialisationProblemeHebdo that it belongs to BindingConstraints

src/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
JasonMarechal25: Yes, it's nullptr by default. It's not required to initialze it, not an issue either. Can be edited on the fly at some point

flomnes: See #1382

src/libs/antares/study/binding_constraint/BindingConstraint.cpp
flomnes: Is taken into account in issue #1382

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.h
guilpier-code: What if we make it a variable local to **enabled()** method ?
As it is modified only inside **enabled()**, we then would not need to reset it in 3 places in the code, right ?

guilpier-code (approved):

PR #1365, opened 2023-06-15
BC scenarisation : Refactor end to end tests

guilpier-code:
- [x] Reduce log lines for each test
- [ ] Remove duplications whenever they exist
- [x] Introduce a fixture with set up and tear down
- [x] Make tests more clear
- [x] Move projects **test_utils** and **test_constraint**  inside  project **Unit-tests**
- [x] Study : make sure one week only is run for each year.
- [x] We should not narrow the checks on results to the first hour only
- [x] Avoid testing BCs but rather focus testing BCs' rhs scenarisation (but maybe we want also tests on BCs only)
- [x] Find a workaround to retrieve results of each year (currently, we use synthesis to make checks on simulation results)

This PR will probably trigger another refactoring on file **src\tests\end-to-end\simple_study\simple-study.cpp**


src/tests/end-to-end/simple_study/simple-study.cpp
```diff
@@ -174,8 +174,8 @@ BOOST_AUTO_TEST_CASE(one_mc_year_one_ts)
 	//Load must be load
 	checkVariable<Solver::Variable::Economy::VCardTimeSeriesValuesLoad>(simulation, pArea, load);
 
-	//Clean simulation
-	cleanSimulation(pStudy, simulation);
+	cleanSimulation(simulation);
+	cleanStudy(pStudy);
```
JasonMarechal25, line 178: Why not have "cleanSimulation" call "cleanStudy" ?

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -11,390 +11,488 @@ namespace utf = boost::unit_test;
...
-    typename Antares::Solver::Variable::Storage<Solver::Variable::Economy::VCardFlowLinear>::ResultsType *result = nullptr;
-    simulation->variables.retrieveResultsForLink<Solver::Variable::Economy::VCardFlowLinear>(&result, link);
-    BOOST_TEST(result->avgdata.weekly[0] == rhs * cost * 7, tt::tolerance(0.001));
-
-    //Clean simulation
-    cleanSimulation(pStudy, simulation);
-}
 
-BOOST_AUTO_TEST_CASE(two_mc_year_two_ts__Binding_Constraints_Hourly)
+BOOST_AUTO_TEST_CASE(daily_BC_restricts_link_direct_capacity_to_60)
```
JasonMarechal25, line 396: There is already Hourly_BC_restricts_link_direct_capacity_to_90, did you meant this one to be yearly ?

flomnes:
Conflicts **src/libs/antares/study/binding_constraint/BindingConstraintsList.cpp**

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
guilpier-code: As you can, this test is on a daily BC (not hourly), and the link capacity is restricted to 60 (not 90).
Did I correctly answer to your question ?

src/tests/end-to-end/simple_study/simple-study.cpp
guilpier-code: To me, the study and the simulation are 2 different things : 
- The study holds (or is supposed to hold) a lot of input data used to create a simulation. Having a look to  the study class,  we can see that it mostly contains functions loading / saving input data, as well as renaming / removing input assets of the study (clusters, areas, links, ...).  By the way, these latter functions are used by the old GUI, and therefore will be removed eventually.
- The simulation is based on the study input data for creation, and it processes all the computations as well as it generates the output results.

JasonMarechal25 (approved):

flomnes (approved):

PR #1175, opened 2023-02-20
FINAL RESERVOIR LEVEL (RTEi's -> CR25)

Milos-RTEi:
This PR is created in order to implement RTEi development on CR25 (Final Reservoir Levels).
Spec/Discussion: https://github.com/AntaresSimulatorTeam/Antares_Simulator/discussions/1145

~~### Two cases~~
- Full year => change monthly inflows only in the monthly heuristic
~~- Partial year => change rule curves~~

PR #1370, opened 2023-06-19
Final Reservoir Level (RTEi's -> CR25)

Milos-RTEi:
This PR is created in order to implement RTEi development on CR25 (Final Reservoir Levels).
Spec/Discussion: https://github.com/AntaresSimulatorTeam/Antares_Simulator/discussions/1145

This CR is identical to https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1175
Moved here to activate tests and sonar-cloud


src/libs/antares/study/parts/hydro/container.h
```diff
@@ -34,6 +34,98 @@
...
 namespace Antares::Data
 {
+/*!
+ ** \brief Final Reservoir Level data for a single area
+ */
+class FinalReservoirLevelRuntimeData
+{
+public:
+    // FInal reservoir level runtime data
+    std::vector<bool> includeFinalReservoirLevel;
```
flomnes: Here it's not clear what the index should be.

Please note that some structures are shared across threads (see numSpace).

src/libs/antares/study/parts/hydro/container.h
```diff
@@ -34,6 +34,98 @@
...
+    }
+
+    void assignEndLevelAndDelta()
+    {
+        includeFinalReservoirLevel.at(yearIndex) = true;
+        endLevel.at(yearIndex) = finalReservoirLevel;
+        deltaLevel.at(yearIndex) = deltaReservoirLevel;
+    }
+
+    double calculateTotalInflows(Antares::Memory::Stored<double>::Type& srcinflows) const
```
flomnes: This function already exists. Re-use it ?

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -64,7 +64,8 @@ void Rules::saveToINIFile(Yuni::IO::File::Stream& file) const
             linksNTC[i].saveToINIFile(study_, file);
         }
         // hydro levels
-        hydroLevels.saveToINIFile(study_, file);
+        hydroInitialLevels.saveToINIFileHydroLevel(study_, file, "hl,");
```
flomnes: Good job for factorization. Could you give the prefix to the constructor of hydroLevelsData ?

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
```diff
@@ -0,0 +1,130 @@
+/*
+** Copyright 2007-2018 RTE
+** Authors: Antares_Simulator Team
```
flomnes: Date, authors

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+static void initializeGeneralData(Data::FinalReservoirLevelRuntimeData& finLevData,
+                           const Data::Parameters& parameters,
+                           uint year)
+{
+    finLevData.simEndDay = parameters.simulationDays.end;
+    finLevData.yearIndex = year;
+}
+
+static void initializePerAreaData(Data::FinalReservoirLevelRuntimeData& finLevData,
```
flomnes: Make this a member function of FinalReservoirLevelRuntimeData ?

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
```diff
@@ -0,0 +1,130 @@
...
+                           const Matrix<double>& scenarioFinalHydroLevels,
+                           const Data::Area& area)
+{
+    finLevData.initialReservoirLevel = scenarioInitialHydroLevels[area.index][finLevData.yearIndex];
+    finLevData.finalReservoirLevel = scenarioFinalHydroLevels[area.index][finLevData.yearIndex];
+    finLevData.deltaReservoirLevel
+      = finLevData.initialReservoirLevel - finLevData.finalReservoirLevel;
+}
+
+void initializePreCheckData(Data::FinalReservoirLevelRuntimeData& finLevData,
```
flomnes: Member function?

flomnes (changes requested):
Please make sure to include unit tests for check and I/O functions.

Milos-RTEi:
We will add some unit tests to cover the functions we wrote. 
By next week.

src/libs/antares/study/parts/hydro/container.h
Milos-RTEi: added explanation for vector indexes.
Tested for parallel work. Works fine.
The pre-checks are completed before parallelization, I checked.
The point is not to break the simulation when it is already running (parallel or not). 
The simulation is aborted before the simulation starts if something is wrong with the final reservoir data.

src/libs/antares/study/parts/hydro/container.h
Milos-RTEi: actually, as I can see it does not exist. 
Total inflows are calculated in method prepareInflowsScaling in management.cpp. 
It is a big method doing a lot of additional stuff, not good for re-using.

src/libs/antares/study/scenario-builder/rules.cpp
Milos-RTEi: done [85e2ae2](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1370/commits/85e2ae29b7d6c0b4caa1eb2bc796c36e85de066d)

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
Milos-RTEi: done

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
Milos-RTEi: Ah, remember this now. Tried to do it like that originally.
However, cannot pass "area" as a function parameter if the method is defined in FinalReservoirLevelRuntimeData.
Always get error: invalid use of incomplete type ‘const class Antares::Data::Area’
Tried with the forward declaration in the container.h but did not help.

src/libs/antares/study/parts/hydro/container.h
```diff
@@ -34,6 +34,104 @@
 
 namespace Antares::Data
 {
+/*!
+ ** \brief Final Reservoir Level data for a single area
+ */
+class FinalReservoirLevelRuntimeData
```
guilpier-code: This class has no constructor, and each member data is public.
So the **information hiding** principle is not respected (see more details further).

src/libs/antares/study/scenario-builder/hydroLevelsData.h
```diff
@@ -45,7 +45,9 @@ class hydroLevelsData final : public dataInterface
     using MatrixType = Matrix<double>;
 
 public:
-    // We use default constructor and destructor
+    // Constructor
+    
+    hydroLevelsData(std::string& saveToINIFilePrefix);
```
guilpier-code: `std::string& saveToINIFilePrefix` as argument :  **iniFilePrefix** is enough, as this prefix can also be read, and not only saved.

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -361,7 +379,8 @@ bool Rules::apply()
             returned_status = renewable[i].apply(study_) && returned_status;
             returned_status = linksNTC[i].apply(study_) && returned_status;
         }
-        returned_status = hydroLevels.apply(study_) && returned_status;
+        returned_status = hydroInitialLevels.applyHydroLevels(study_.scenarioInitialHydroLevels) && returned_status;
```
guilpier-code: About : 
```C++
    returned_status = hydroInitialLevels.applyHydroLevels(study_.scenarioInitialHydroLevels) && returned_status;
    returned_status = hydroFinalLevels.applyHydroLevels(study_.scenarioFinalHydroLevels) && returned_status;
```
We're aware that the TS numbers data classes hierarchy have shortcomings, and maybe that's the reason why you chose **not** to implement a proper **apply(study)** for hydro levels (which is a pure virtual method).
One of the shortcomings is that TS numbers for hydro levels should be stored in each area (in the hydro part) instead of the study. This is how it's done for other TS numbers matrices, why not for hydro levels as well ?
Until we address these shortcomings, we should stick to the plan, implement the **apply(...)** method and call it here : 
```C++
    returned_status = hydroInitialLevels.apply(study_) && returned_status;
    returned_status = hydroFinalLevels.apply(study_) && returned_status;
```

A possible way to do this would be to use dependency injection under the form of an **std::function** : 
```C++
// hydroLevelsData.h
class hydroLevelsData final : public dataInterface
{
    hydroLevelsData(std::string& iniFilePrefix, 
                    std::function<void(Study&, MatrixType&)> applyToTarget);
    ....
 private:
    std::function<void(Study&, MatrixType&)> applyToTarget_;
    ....
}; // class hydroLevelsData

void initLevelApply(Study& study, MatrixType& matrix)
{
     study.scenarioInitialHydroLevels.copyFrom(matrix);
}

void finalLevelApply(Study& study, MatrixType& matrix)
{
     study.scenarioFinalHydroLevels.copyFrom(matrix);
}
...
```
```C++
// hydroLevelsData.cpp
...
hydroLevelsData::hydroLevelsData(std::string& iniFilePrefix, 
                                 std::function<void(Study&, MatrixType&)> applyToTarget) :
    addToPrefix_(iniFilePrefix),
    applyToTarget_(applyToTarget)
{}
...

bool hydroLevelsData::apply(Study& study)
{
    applyToTarget_(study, pHydroLevelsRules);
    return true;
}
```
```C++
// rules.cpp
...

Rules::Rules(Study& study) : study_(study), pAreaCount(0)
{
    std::string prefixInitial = "hl,";
    hydroInitialLevels = hydroLevelsData(prefixInitial, initLevelApply);

    std::string prefixFinal = "hfl,";
    hydroFinalLevels = hydroLevelsData(prefixFinal, finalLevelApply);
}
```

guilpier-code (changes requested):

guilpier-code:
Comments to be continued... (not finished yet)

src/solver/simulation/apply-scenario.cpp
```diff
@@ -55,6 +56,12 @@ void ApplyCustomScenario(Data::Study& study)
 
     study.scenarioRulesDestroy();
     logs.info();
+
+    // final reservoir level functions and pre-checks
+    FinalReservoirLevel(study.scenarioInitialHydroLevels,
```
guilpier-code: Here we're in function **ApplyCustomScenario**.
 And **FinalReservoirLevel** is mainly about checks (and simple intermediate variables computations), which has nothing to see with "applying custom scenarii".
So I would move this call / treatment to hydro management (**management.cpp**), and under a different form (see review of file **hydro-final-reservoir-level-functions.cpp**)

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
```diff
@@ -0,0 +1,130 @@
...
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+
+#include "hydro-final-reservoir-level-functions.h"
+#include <antares/emergency.h>
+
+namespace Antares::Solver
+{
+void initializeGeneralData(Data::FinalReservoirLevelRuntimeData& finLevData,
```
guilpier-code: These functions have so much to do with class **FinalReservoirLevelRuntimeData** that this piece of code seems to beg for being integrated in this class, for cohesion reason !
It would allow to simplify this code : 
- We would have less function arguments and less temporary variables
- We wouldn't have to loop over areas because it would be done in `HydroManagement::prepareFinalReservoirLevelData` (this loop already exists in this function)
- What else ?

src/libs/antares/study/parts/hydro/container.h
guilpier-code: Renaming (minor) : why is there **RuntimeData** is the name ? It is not meaningful to me.

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
guilpier-code: After integration, all the **initializeSomething**(...) functions would disappear and their content could be moved in the class **FinalReservoirLevelRuntimeData** constructor.

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+void initializePreCheckData(Data::FinalReservoirLevelRuntimeData& finLevData,
+                            const Data::Area& area)
+{
+    finLevData.initReservoirLvlMonth = area.hydro.initializeReservoirLevelDate; // month [0-11]
+    finLevData.reservoirCapacity = area.hydro.reservoirCapacity;
+}
+
+void ruleCurveForSimEndReal(Data::FinalReservoirLevelRuntimeData& finLevData, Data::Area& area)
```
guilpier-code: Same for the **ruleCurveForSimEndReal** : its content can be moved to class constructor

src/libs/antares/study/parts/hydro/container.h
```diff
@@ -34,6 +34,104 @@
 
 namespace Antares::Data
 {
+/*!
```
guilpier-code: According to **hydro-final-reservoir-level-functions.cpp**, this class could be about to welcome some more code.
- Feel free to split this class into several collaborating classes if many responsibilities emerge
- Maybe we should create new **.h/.cpp** source files for the subject of final levels

src/libs/antares/study/parts/hydro/container.h
guilpier-code: We should create a constructor such as : 
- we shouldn't hesitate storing (as member data) the **Area** containing an instance of the current class, so we could easily retrieve hydro or even area information we need from the area (area name, inflows, ...).
- further member function `fillEmpty() ` disappears and its content belong to the constructor, as it is default member variable initialization, right ? 
- Probably same thing for **assignEndLevelAndDelta** (after **initializePreCheckData(...)** is integrated)
- what else ?

src/libs/antares/study/parts/hydro/container.h
```diff
@@ -34,6 +34,104 @@
...
+    double finalReservoirLevel;
+    double deltaReservoirLevel;
+    double reservoirCapacity;
+    double lowLevelLastDay;
+    double highLevelLastDay;
+    int initReservoirLvlMonth;
+
+    // methods:
+
+    void fillEmpty()
```
guilpier-code: Once all code from outside is here (from **hydro-final-reservoir-level-functions.cpp** I mean), create as many private methods as possible, because we don't want to expose too much details.
Indeed, as this class is only used in **management.cpp**, we could (roughly) have something like  (to be discussed) : 
```C++
void HydroManagement::prepareFinalReservoirLevelData(uint numSpace, uint year)
{
    study.areas.each([this, &numSpace, &year](Data::Area& area)
    {
          area.hydro.finalLevelInflowsModifyer.makeChecks()

          if (not area.hydro.finalLevelInflowsModifyer.isActive())
              return;

          area.finalLevelInflowsModifyer.updateInflows(year);
    });
}
```
What do you think ?

src/solver/hydro/management/management.cpp
```diff
@@ -351,6 +351,27 @@ bool HydroManagement::checkMinGeneration(uint numSpace)
     return ret;
 }
 
+void HydroManagement::prepareFinalReservoirLevelData(uint numSpace, uint year)
```
guilpier-code: What we mostly do here is modify the inflows due to forced final level, right ?
So why not call this function **changeInflowsDueToFinalLevels** ?

guilpier-code (changes requested):

src/solver/simulation/apply-scenario.cpp
Milos-RTEi: While we agree with other comments and started implementing them (hope nothing out of the ordinary will pop out) I have a comment on this one.
I agree FinalReservoirLevel does not really belong to ApplyCustomScenario. 
However, if we put it inside management.cpp we are already in the parallel run. 
First, structure FinalReservoirLevelRuntimeData is not made for parallel run. 
Second, we actually want to break the simulation before it starts running MC-years, see https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1370#discussion_r1238354629
not to have simulation running e.g. a couple of hours and then break it on MCy e.g. 800 once it realizes something is wrong.
Can we put it somewhere else, before parallelization starts?

src/libs/antares/study/area/list.cpp
```diff
@@ -1324,6 +1325,17 @@ void AreaListEnsureDataHydroTimeSeries(AreaList* l)
...
 
+void AreaListEnsureFinalLevelInflowsModifyerAreaPtr(AreaList* l)
+{
+    /* Asserts */
+    assert(l);
+
+    l->each([&](Data::Area& area) {
+        if (!area.hydro.finalLevelInflowsModifier->areaPtr)
+            area.hydro.finalLevelInflowsModifier->areaPtr = &area;
+    });
```
flomnes: I suggest you pass `area` to the constructor of `finalLevelInflowsModifier`

src/libs/antares/study/parts/hydro/container.cpp
flomnes, line 1: Pass `area` to the constructor of `PartHydro`, then to finalLevelInflowsModifier.

There is no need for a `std::shared_ptr` in my opinion, just a bare structure would probably do the job.

src/libs/antares/study/parts/hydro/container.h
```diff
@@ -172,6 +179,10 @@ class PartHydro
     //! Data for time-series
     DataSeriesHydro* series;
 
+    //! Data for final reservoir level
+    
+    std::shared_ptr<FinalLevelInflowsModifier> finalLevelInflowsModifier;
```
flomnes:
```suggestion
    FinalLevelInflowsModifier finalLevelInflowsModifier;
```

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.h
```diff
@@ -0,0 +1,110 @@
...
+    double initialReservoirLevel;
+    double finalReservoirLevel;
+    double deltaReservoirLevel;
+    double reservoirCapacity;
+    double lowLevelLastDay;
+    double highLevelLastDay;
+    int initReservoirLvlMonth;
+
+public:
+    Area* areaPtr;
```
flomnes:
```suggestion
    Area* area;
```

Consider making area `private`

flomnes (commented):
A lot of progress on your side, this PR is a much easier to understand than earlier versions. Thank you !

I made some minor remarks

src/solver/hydro/management/management.cpp
flomnes: `changeInflowsToAccomodateFinalLevels` ?

src/libs/antares/study/parts/hydro/container.h
Milos-RTEi: Pointer to Area added to the class.
fillEmpty() and all other functions moved from src/solver/simulation/hydro-final-reservoir-level-functions.cpp to the class FinalReservoirLevelRuntimeData (renamed to FinalLevelInflowsModifier). 
Couldn't be done in the constructor. 
Hiding principle was implemented.

src/libs/antares/study/scenario-builder/hydroLevelsData.h
Milos-RTEi: done

src/libs/antares/study/scenario-builder/rules.cpp
Milos-RTEi: done

src/libs/antares/study/parts/hydro/container.h
Milos-RTEi: done

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
Milos-RTEi: done. not to the constructor. Moved as a method of the class.

src/libs/antares/study/parts/hydro/container.h
Milos-RTEi: Splited into separate *.cpp and *.h file

src/libs/antares/study/parts/hydro/container.h
Milos-RTEi: Done very similarly to the proposed one. 
The data is first calculated and stored and then checked. So that is reversed.
return will break the area-loop so not used.

src/solver/simulation/apply-scenario.cpp
Milos-RTEi: Maybe we can put it below ApplyCustomScenario in src/solver/simulation/solver.hxx. 
To be executed only if parameters.useCustomScenario = true.

Milos-RTEi:
Transferred to https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1521

PR #1372, opened 2023-06-20
Disable store in input for Stochastic time series when not at latest version

JasonMarechal25:


src/libs/antares/study/load.cpp
```diff
@@ -117,11 +118,13 @@ void Study::parameterFiller(const StudyLoadOptions& options)
     if (usedByTheSolver)
     {
         // We have time-series to import
-        if (parameters.timeSeriesToImport && (uint) header.version != (uint)versionLatest)
+        StudyVersion studyVersion;
+        if (parameters.exportTimeSeriesInInput && studyVersion.isStudyLatestVersion(folder.c_str()))
```
flomnes, line 122: Why not use existing `header.version` ?

JasonMarechal25: To prevent dependency to study

flomnes (approved):

PR #1381, opened 2023-06-26
Refactor hydro pmin checks

guilpier-code:
This PR follows a question from Kamel on hydro **Pmin** checks.
Regarding hydro **Pmin**, there are many checks done before applying the **Pmin**.
I found it difficult to understand in which case exactly a specific check is done.
This is the intention of this PR : clarify the conditions for a check, by extracting these conditions outside the check functions.
This also sticks to the SRP (single responsibility principle) : the function are only testing something, they don't have to evaluate conditions before doing their real job.


flomnes:
@guilpier-code Can you describe the goal of this work or give some context ?

flomnes (commented):
We need to make sure that those checks occur right after the input data has been loaded, not during the simulation.

.github/workflows/ubuntu.yml
```diff
@@ -141,7 +141,7 @@ jobs:
         simtest-tag: ${{steps.simtest-version.outputs.prop}}
         batch-name: valid-v830
         os: ${{ matrix.test-platform }}
-
+
```
flomnes:
```suggestion

```

src/solver/hydro/management/management.cpp
```diff
@@ -339,14 +332,29 @@ bool HydroManagement::checkMinGeneration(uint numSpace)
...
+
+        if (useHeuristicTarget && followLoadModulations && reservoirManagement)
+        {
             ret = checkYearlyMinGeneration(numSpace, tsIndex, area) && ret;
+        }
+
+        if (useHeuristicTarget && !followLoadModulations)
+        {
             ret = checkWeeklyMinGeneration(tsIndex, area) && ret;
         }
```
flomnes: Factor `useHeuristicTarget`

flomnes:
`checkWeeklyMinGeneration` : I see a difference

# Before
```cpp
    if (!area.hydro.followLoadModulations)
```
# After
```cpp
        if (useHeuristicTarget && !followLoadModulations)
        {
```
Is this intended ?

flomnes:
For function `checkHourlyMinGeneration`, condition `!area.hydro.reservoirManagement` is now checked twice. Once inside the function, once by the caller `HydroManagement::checkMinGeneration`. Consider removing the check inside `checkHourlyMinGeneration`.

flomnes:
Functions `checkYearlyMinGeneration` adds condition `useHeuristicTarget`. Is this intended ?

`checkMonthlyMinGeneration` is the only function where I saw no difference.

guilpier-code:
> @guilpier-code Can you describe the goal of this work or give some context ?

Please see this current PR presentation comment

guilpier-code:
> `checkWeeklyMinGeneration` : I see a difference
> 
> # Before
> ```c++
>     if (!area.hydro.followLoadModulations)
> ```
> 
> # After
> ```c++
>         if (useHeuristicTarget && !followLoadModulations)
>         {
> ```
> 
> Is this intended ?

Yes, it is intended : 
**Before** : **useHeuristicTarget** outside the call, and **!followLoadModulations** inside the call
**After**  : **useHeuristicTarget** && **!followLoadModulations**
You just made the confusion because the previous code was unclear.

guilpier-code:
> For function `checkHourlyMinGeneration`, condition `!area.hydro.reservoirManagement` is now checked twice. Once inside the function, once by the caller `HydroManagement::checkMinGeneration`. Consider removing the check inside `checkHourlyMinGeneration`.

You're right. Corrected.

guilpier-code:
> Functions `checkYearlyMinGeneration` adds condition `useHeuristicTarget`. Is this intended ?
> 
> `checkMonthlyMinGeneration` is the only function where I saw no difference.

Same as above : 
before : the conditions are spread in different places : inside the function and outside (where function is called)
now : all conditions are extracted and tested outside the call.

flomnes (approved):

PR #1388, opened 2023-06-30
Group of constraints

### LINKED ISSUE #1321: BC scenarisation : create container for groups
Iterating over groups, not BCs should be easier & more readable.

JasonMarechal25:


src/libs/antares/CMakeLists.txt
```diff
@@ -529,7 +529,7 @@ set(SRC
     # TODO : should be in a target_include_directories : but need to remove all .h from all SRC
     Enum.hpp
     Enum.hxx
-)
+        study/binding_constraint/BindingConstraintGroup.cpp study/binding_constraint/BindingConstraintGroup.h study/binding_constraint/BindingConstraintGroupRepository.cpp study/binding_constraint/BindingConstraintGroupRepository.h)
```
flomnes: Create a specific CMake target for this sub-library ?

src/libs/antares/study/binding_constraint/BindingConstraintGroup.cpp
```diff
@@ -0,0 +1,39 @@
...
+            return constraint->RHSTimeSeries().width == 1;
+        })) {
+            timeSeriesNumbers_.timeseriesNumbers.fillColumn(0, 0);
+        }
+    }
+
+    unsigned BindingConstraintGroup::numberOfTimeseries() const {
+        //Assume all BC in a group have the same width
+        if (constraints_.empty()) return 0;
+        return (*constraints_.begin())->RHSTimeSeries().width;
```
flomnes, line 36: You de-reference twice, yet check only that one element is not nullptr

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
```diff
@@ -0,0 +1,42 @@
...
+    [[nodiscard]] std::vector<std::shared_ptr<BindingConstraintGroup>>::iterator begin();
+    [[nodiscard]] std::vector<std::shared_ptr<BindingConstraintGroup>>::const_iterator begin() const;
+
+    [[nodiscard]] std::vector<std::shared_ptr<BindingConstraintGroup>>::iterator end();
+    [[nodiscard]] std::vector<std::shared_ptr<BindingConstraintGroup>>::const_iterator end() const;
+
+    std::shared_ptr<BindingConstraintGroup> add(const std::string& name);
+
+private:
+    std::vector<std::shared_ptr<BindingConstraintGroup>> groups_;
```
flomnes: Why do we need a shared_ptr here ?

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,102 @@
...
+            std::shared_ptr<BindingConstraintGroup> group;
+            if (group_found == groups_.end()) {
+                group = groups_.emplace_back(std::make_shared<BindingConstraintGroup>(constraint->group()));
+            } else {
+                group = *group_found;
+            }
+            group->add(constraint);
+        }
+
+        bool tsAreConsistent = timeSeriesWidthConsistentInGroups();
```
flomnes: No need to declare this variable, right ?

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    void BindingConstraintGroupRepository::fixTSNumbersWhenWidthIsOne() {
+        std::for_each(groups_.begin(), groups_.end(), [](auto group) {
+            group->fixTSNumbersWhenWidthIsOne();
+        });
+    }
+
+    std::shared_ptr<BindingConstraintGroup> BindingConstraintGroupRepository::operator[](std::string name) {
+        if (auto group = std::find_if(groups_.begin(), groups_.end(), [&name](auto group) {
+            return group->name() == name;
```
flomnes: Why not use a map ?

src/libs/antares/CMakeLists.txt
JasonMarechal25: Tried to but too difficult to introduce in this PR, too many impact

https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/1394

src/libs/antares/study/binding_constraint/BindingConstraintGroup.cpp
JasonMarechal25: I'm not sure to understand properly, I'm not null checking
`if (constraints_.empty()) return 0;` => If no constraint exist for this group, number of TS is 0. Should not happen in practice but happen in unit tests.

`return (*constraints_.begin())->RHSTimeSeries().width;` 
=> `constraints_.begin()` is iterator. Always valid since from the previous check we know the set is not empty.
=> `(*constraints_.begin())` is the shared_ptr to the constraint

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
JasonMarechal25: The repository is owner of the BindingConstraintGroup and manage there lifetime

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
JasonMarechal25: Why use a map ?

We are more likely to go through the collection than to access specific elements of it, that's why I didn't use a map.

src/libs/antares/study/binding_constraint/BindingConstraintGroup.cpp
flomnes: ok

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
```
flomnes: `group` shadows `group`

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
```diff
@@ -63,36 +63,12 @@ bool compareConstraints(const std::shared_ptr<BindingConstraint>& s1, const std:
 std::shared_ptr<BindingConstraint> BindingConstraintsRepository::add(const AnyString &name) {
     auto bc = std::make_shared<BindingConstraint>();
     bc->name(name);
-    pList.push_back(bc);
-    std::sort(pList.begin(), pList.end(), compareConstraints);
+    constraints_.push_back(bc);
+    std::sort(constraints_.begin(), constraints_.end(), compareConstraints);
+    auto count = bc.use_count();
```
guilpier-code: About `auto count = bc.use_count();` : 
Why do we define a **count** we don't use ?

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        bool tsAreConsistent = timeSeriesWidthConsistentInGroups();
+
+        return tsAreConsistent;
+    }
+
+    bool BindingConstraintGroupRepository::timeSeriesWidthConsistentInGroups() const {
+        bool hasError = false;
+        for(const auto& group: this->groups_) {
```
guilpier-code: `this->groups_`  : `group_ `is enough. We know from the trailing **"_"** that we deal with a private member data

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+            unsigned count = 0;
+            for (const auto& bc: group->constraints()) {
```
guilpier-code: Couldn't we use **std::equal** or an algorithm here ? (It's an open question, maybe it's not possible)
```C++
auto constraints = group->constraints();
if (not std::equal(constraints.begin(), constraints.end(), ..., [](auto& bc) { /*get number of TS */ })
{
    // Log an error
}
```

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+            std::shared_ptr<BindingConstraintGroup> group;
+            if (group_found == groups_.end()) {
+                group = groups_.emplace_back(std::make_shared<BindingConstraintGroup>(constraint->group()));
+            } else {
+                group = *group_found;
+            }
+            group->add(constraint);
+        }
+
+        bool tsAreConsistent = timeSeriesWidthConsistentInGroups();
```
guilpier-code: `timeSeriesWidthConsistentInGroups()` : 
This function is only called inside the class. So strictly speaking, it should be **private**.
But a question bothers me : how do we ensure that all contraints of a group have the same numbers of RHS TS after we read them ?
Does the code checks this somewhere ?
**Answer** : yes, it is done when calling function **buildFrom(...)** from **Study::internalLoadBindingConstraints(...)**

guilpier-code (changes requested):

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
guilpier-code: **Answer** : yes, it is done when calling function **buildFrom(...)** from **Study::internalLoadBindingConstraints(...)**

src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -627,7 +627,11 @@ bool BindingConstraintsTSNumberData::apply(Study& study)
     uint errors = 0;
```
guilpier-code: About `bool BindingConstraintsTSNumberData::apply(Study& study)` : 
Shouldn't this method definition be moved with other methods of class  **BindingConstraintsTSNumberData**, in file **BindingConstraintsTSNumbersData.cpp** ?

src/libs/antares/study/scenario-builder/TSnumberData.cpp
```diff
@@ -627,7 +627,11 @@ bool BindingConstraintsTSNumberData::apply(Study& study)
...
     CString<512, false> logprefix;
     for (const auto& [group_name, ts_numbers]: rules_) {
-        ApplyToMatrix(errors, logprefix, study.bindingConstraints.groupToTimeSeriesNumbers[group_name], ts_numbers[0],
+        auto group = study.bindingConstraintsGroups[group_name];
+        if (group == nullptr) {
```
guilpier-code, line 630: If we can't find a group by its name, why don't we return **false**  (as the current function returns a boolean)?

src/solver/simulation/ITimeSeriesNumbersWriter.h
```diff
@@ -3,13 +3,13 @@
 //
 
 #pragma once
-#include "antares/study/binding_constraint/BindingConstraintsRepository.h"
```
guilpier-code, line 6: This class is useless to me (see [this remark](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1219#discussion_r1229820466))

guilpier-code (changes requested):

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
JasonMarechal25: Probably leftover

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
JasonMarechal25: Indeed.
std::equal was not the right candidate but I've used std::all_of et std::any_of.
```hasError = std::any_of(group, where not std::all_of(bc have the same width) )```

File: src/libs/antares/study/scenario-builder/TSnumberData.cpp (hunk 11 above)
Comment (JasonMarechal25) [reply src/libs/antares/study/scenario-builder/TSnumberData.cpp]: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1398

File: src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h [hunk 14]
```diff
@@ -0,0 +1,44 @@
...
+#include "antares/study/binding_constraint/BindingConstraintGroup.h"
+#include "antares/study/fwd.h"
+#include "BindingConstraint.h"
+#include "BindingConstraintSaver.h"
+
+namespace Antares::Data {
+
+class BindingConstraintGroupRepository {
+public:
+    [[nodiscard]] unsigned size() const;
```
flomnes, line 19: `using iterator = std::vector<std::shared_ptr<BindingConstraintGroup>>::iterator;`

It would simplify the code below, and allow the caller to use `BindingConstraintGroupRepository::iterator` if they don't want to use `auto`.

src/tests/src/libs/antares/study/constraint/test_group.cpp
flomnes, line 1: You don't have to use files. 

Checking that files are read properly is useful.

But you could also create constraints & groups in-memory.

flomnes (changes requested):

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -240,8 +241,8 @@ void BCgroupScenarioBuilder::yearGetsTSnumber(std::string groupName, unsigned in
 class SimulationHandler
 {
 public:
-    SimulationHandler(std::shared_ptr<Study> study)
```
sylvlecl: Not directly related to that PR but:

Instead of using a shared_ptr, the best thing would be to clarify if this handler should have ownership of the study (I don't think so).
If not, we can just pass a reference here.
shared_ptr should be used only when shared ownership is actually required.
Then we wouldn't need the move :)

src/libs/antares/study/binding_constraint/BindingConstraintGroup.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <memory>
+#include "BindingConstraint.h"
+namespace Antares::Data {
+
+class BindingConstraintGroup {
+public:
+    explicit BindingConstraintGroup(std::string name);
+
+    [[nodiscard]] std::string name() { return name_; }
```
sylvlecl, line 15: Should return const reference

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -257,20 +258,24 @@ class SimulationHandler
 
 void SimulationHandler::create()
 {
+    BOOST_CHECK(study_);
     BOOST_CHECK(study_->initializeRuntimeInfos());
     addScratchpadToEachArea(study_);
+    BOOST_CHECK(true);
```
sylvlecl: Are those checks "true" useful?  (I don't know boost test, but usually such assertions in unit tests are suspicious :)  )

src/tests/end-to-end/utils/utils.h
```diff
@@ -11,7 +11,7 @@ Antares::Data::Area* addArea(Antares::Data::Study::Ptr pStudy, const std::string
 
 std::shared_ptr<Antares::Data::ThermalCluster> addCluster(Antares::Data::Area* pArea, const std::string& clusterName, double maximumPower, double cost, int nbTS, int unitCount = 1);
 
-std::shared_ptr<Antares::Data::BindingConstraint> addBindingConstraints(Antares::Data::Study::Ptr study, std::string name, std::string group);
+std::shared_ptr<Antares::Data::BindingConstraint> addBindingConstraints(const Antares::Data::Study::Ptr& study, const std::string& name, const std::string& group);
```
sylvlecl: Note that we should pass the study by reference here, since the function will clearly not take ownership of the study.

src/tests/resources/Antares_Simulator_Tests
sylvlecl, line 1: Note that this commit of tests repository does not exist on the public repository (recursing checkout fails):
we must not merge that change.

src/libs/antares/study/study.h
```diff
@@ -647,6 +648,7 @@ class Study final : public Yuni::NonCopyable<Study>, public IObject, public Laye
     //@{
     //! Binding constraints
     BindingConstraintsRepository bindingConstraints;
+    BindingConstraintGroupRepository bindingConstraintsGroups;
```
sylvlecl, line 651: I don't think it's a good idea to leave the 2 fields public, because they are now linked by an invariant (a BC must be present in both repos).
It's too easy to break the consistency of the study object if we don't link them more strongly at the class level (I know it's already the case in many other places, but let's not add one).

I would advise to make the fields private and have add/remove methods which update the 2 objects at once.

Another option would be to have only one repository class for both BC and their groups, that handles this consistency concern.

src/tests/end-to-end/utils/utils.cpp
```diff
@@ -24,9 +26,14 @@ void prepareStudy(Study::Ptr pStudy, int nbYears)
     Data::Study::Current::Set(pStudy);
 }
 
-std::shared_ptr<BindingConstraint> addBindingConstraints(Study::Ptr study, std::string name, std::string group) {
+std::shared_ptr<BindingConstraint> addBindingConstraints(const Study::Ptr& study, const std::string& name, const std::string& group) {
```
sylvlecl: Would not be needed if consistency is implemented at the study level when we add the BC.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
sylvlecl: Then the best option is to have unique_ptr.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
```diff
@@ -0,0 +1,47 @@
...
+    using iterator = std::vector<std::shared_ptr<BindingConstraintGroup>>::iterator;
+    using const_iterator = std::vector<std::shared_ptr<BindingConstraintGroup>>::const_iterator;
+
+    [[nodiscard]] iterator begin();
+    [[nodiscard]] const_iterator begin() const;
+
+    [[nodiscard]] iterator end();
+    [[nodiscard]] const_iterator end() const;
+
+    std::shared_ptr<BindingConstraintGroup> add(const std::string& name);
```
sylvlecl: The best option is to return a reference, we don't want to transfer ownership, even shared, to the user.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
```diff
@@ -0,0 +1,47 @@
...
+public:
+    [[nodiscard]] unsigned size() const;
+
+    [[nodiscard]] bool buildFrom(BindingConstraintsRepository& repository);
+
+    void resizeAllTimeseriesNumbers(unsigned nb_years);
+
+    void fixTSNumbersWhenWidthIsOne();
+
+    std::shared_ptr<BindingConstraintGroup> operator[](std::string name);
```
sylvlecl: The best option is to return a reference or raw pointer (if nullable): the user should no have ownership of this.
See for example [this article](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f7-for-general-use-take-t-or-t-arguments-rather-than-smart-pointers) of cpp core guidelines for some background about when to (not) use smart pointers.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
```diff
@@ -0,0 +1,47 @@
...
+
+    using iterator = std::vector<std::shared_ptr<BindingConstraintGroup>>::iterator;
```
sylvlecl: Here again, I don't think we want to share ownsership with the caller: the best option would be to have iterators over the value type, not shared_ptr,

The std lib is not very well geared for this (providing such an iterator over an underlying vector of pointers), but we could for example use boost::indirect_iterator, or else implement our own.

We could also use boost ranges, which is great for range adaptation.

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.hxx
```diff
@@ -10,66 +10,35 @@
 namespace Antares::Data {
 
 inline uint BindingConstraintsRepository::size() const {
```
sylvlecl, line 12: It seems to me this should be in .cpp

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.hxx
```diff
@@ -10,66 +10,35 @@
...
-    return (uint) pList.size();
+    return (uint) constraints_.size();
 }
 
 inline bool BindingConstraintsRepository::empty() const {
```
sylvlecl, line 16: It seems to me this should be in .cpp

src/libs/antares/study/scenario-builder/TSnumberData.cpp
guilpier-code: Still unresolved it seems

guilpier-code (dismissed):

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
```diff
@@ -0,0 +1,47 @@
...
+#include "BindingConstraint.h"
+#include "BindingConstraintSaver.h"
+
+namespace Antares::Data {
+
+class BindingConstraintGroupRepository {
+public:
+    [[nodiscard]] unsigned size() const;
+
+    [[nodiscard]] bool buildFrom(BindingConstraintsRepository& repository);
```
sylvlecl: The argument should be const ?

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                        return group_of_constraint->name() == name;
+                                    });
+            group != groups_.end())
+        {
+            return *group;
+        }
+        return nullptr;
+    }
+
+    std::vector<std::shared_ptr<BindingConstraintGroup>>::iterator BindingConstraintGroupRepository::begin() {
```
sylvlecl: In all the signatures, I think we'd better use `BindingConstraintGroupRepository::iterator`, for easier evolutivity and readability (if we change the underlying type, no need to change this).

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.h
sylvlecl: Also the argument should be const reference.

src/libs/antares/study/scenario-builder/BindingConstraintsTSNumbersData.cpp
```diff
@@ -38,13 +38,8 @@ namespace Antares::Data::ScenarioBuilder
...
 {
     const uint nbYears = study.parameters.nbYears;
-    std::set<std::string> group_names;
-    std::for_each(study.bindingConstraints.begin(), study.bindingConstraints.end(), [&group_names](const auto& bc) {
-        group_names.insert(bc->group());
-    });
-    rules_.clear();
-    std::for_each(group_names.begin(), group_names.end(), [&](const auto& group_name) {
-        MatrixType& ts_numbers = rules_[group_name];
+    std::for_each(study.bindingConstraintsGroups.begin(), study.bindingConstraintsGroups.end(), [&](const auto& group) {
```
sylvlecl, line 41: A plain for loop instead of std::for_each + lambda, would be much more readable.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                                              }
+                                                              return sameWidth;
+                                                          });
+                                            return !isConsistent;
+        });
+        return allConsistent;
+    }
+
+    void BindingConstraintGroupRepository::resizeAllTimeseriesNumbers(unsigned int nb_years) {
+        std::for_each(groups_.begin(), groups_.end(), [&](auto &group) {
```
sylvlecl, line 54: Range based for loop would be much more readable.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+            group->timeSeriesNumbers().timeseriesNumbers.clear();
+            group->timeSeriesNumbers().timeseriesNumbers.resize(1, nb_years);
+        });
+    }
+
+    void BindingConstraintGroupRepository::fixTSNumbersWhenWidthIsOne() {
+        std::for_each(groups_.begin(), groups_.end(), [](auto group) {
```
sylvlecl: Range based for loop would be much more readable.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+            group->fixTSNumbersWhenWidthIsOne();
+        });
+    }
+
+    std::shared_ptr<BindingConstraintGroup> BindingConstraintGroupRepository::operator[](std::string name) {
+        if (auto group = std::find_if(groups_.begin(), groups_.end(), [&name](auto group_of_constraint) {
+                                        return group_of_constraint->name() == name;
```
sylvlecl, line 68: Note that such linear searches are not efficient: it does not seem that much used so probably not a big issue.

src/libs/antares/study/binding_constraint/BindingConstraintGroup.h
```diff
@@ -0,0 +1,28 @@
...
+    void add(const std::shared_ptr<BindingConstraint>& constraint);
+    [[nodiscard]] std::set<std::shared_ptr<BindingConstraint>> constraints() const;
```
sylvlecl, line 17: First: should return a const reference, not a copy of the set, but actually there is another design issue: we don't want to shared ownership here, so we should not return shared_ptrs (provide iterators over BCs instead, for example).

src/libs/antares/study/binding_constraint/BindingConstraintGroup.h
```diff
@@ -0,0 +1,28 @@
...
+    [[nodiscard]] BindingConstraintTimeSeriesNumbers& timeSeriesNumbers() { return timeSeriesNumbers_; };
+    void fixTSNumbersWhenWidthIsOne();
+    [[nodiscard]] unsigned numberOfTimeseries() const;
+private:
+    std::set<std::shared_ptr<BindingConstraint>> constraints_;
```
sylvlecl, line 22: Not important: a flat_set is more efficient except for insertions (cf. boost::flat_set and futute std::flat_set) --> for our use case it would probably be interesting since we read much more than we insert.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+namespace Antares::Data {
+
+    unsigned BindingConstraintGroupRepository::size() const {
+        return groups_.size();
+    }
+
+    bool BindingConstraintGroupRepository::buildFrom(BindingConstraintsRepository &repository) {
+        for (auto constraint: repository) {
+            auto group_found = std::find_if(groups_.begin(), groups_.end(), [constraint](auto group) {
```
sylvlecl: Can't we use the `[]` operator instead ?

src/solver/simulation/BindingConstraintsTimeSeriesNumbersWriter.h
```diff
@@ -15,7 +15,7 @@ class BindingConstraintsTimeSeriesNumbersWriter: public ITimeSeriesNumbersWriter
 public:
     explicit BindingConstraintsTimeSeriesNumbersWriter(std::shared_ptr<Antares::Solver::IResultWriter> resultWriter);
     BindingConstraintsTimeSeriesNumbersWriter() = default;
-    void write(const Data::BindingConstraintsRepository &list) override;
+    void write(const Data::BindingConstraintGroupRepository &list) override;
```
sylvlecl: Would be best to rename also the arguments

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -185,7 +186,7 @@ class BCrhsConfig
 {
 public:
     BCrhsConfig() = delete;
-    BCrhsConfig(std::shared_ptr<BindingConstraint> BC, unsigned int nbTimeSeries);
```
sylvlecl, line 188: This helper class is clearly not an owner of the BC: we should provide it a reference instead of shared_ptr.
(possibly for later work since it has not been done in this PR)

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
sylvlecl: See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-smart

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-save.cpp
```diff
@@ -176,6 +176,11 @@ struct commonFixture
         study->bindingConstraints.add("BC_2")->group("group2");
         study->bindingConstraints.add("BC_2")->group("group3");
 
+        study->bindingConstraintsGroups.add("group1");
```
sylvlecl, line 179: As described previously: we should not need to do this, it should be handled at the study class level when we add constraints.

sylvlecl (dismissed):
I think we should improve 2 main design choices:

1. **BCs and groups ownership:**
we need to define clearly who owns them, and restrict the use of smart pointers in there, not spreading shared_ptr everywhere outside.
For example, we could have unique_ptr in the BC repository, and references or raw pointers elsewhere.

2. **Consistency between BC repo and groups repo:**
This should be guaranteed at least at the study class level at construction time, instead of having to call another method after the fact. Otherwise it makes construction hard, and it's too easy to have an inconsistent object.

src/libs/antares/study/scenario-builder/TSnumberData.cpp
JasonMarechal25: `return std::all_of`

We always return true or false now

src/libs/antares/study/binding_constraint/BindingConstraintGroup.h
JasonMarechal25: I disagree, we should return by value and count on copy elision
Theoretically [even for containers](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#discussion-return-containers-by-value-relying-on-move-or-copy-elision-for-efficiency) but surprise may happen on Windows [](url)

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
JasonMarechal25: Probably left over from debugging

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
JasonMarechal25: New issue #1476

src/libs/antares/study/binding_constraint/BindingConstraintGroup.h
JasonMarechal25: Too soon
Standard library header <flat_set> **(C++23)**

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
JasonMarechal25: I will need a bit more details. Where and to what purpose please ?

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
JasonMarechal25: New issue #1477

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-save.cpp
JasonMarechal25: I agree.
The thing is that "adding" a bc to the study is not a use case per se. The use case is to construct a batch of bc from a study (on disk).
"Add" methods are probably useful for legacy UI but for the solver they are just useful (dangerous) helpers

src/libs/antares/study/binding_constraint/BindingConstraintGroup.h
sylvlecl: At least the method should be const, but for the return type: I am really not sure compilers will apply copy elision here ?
The returned value is not an rvalue, so it's really not obvious to me that the compiler can perform that optimization (maybe it's the case but I did not find clear guarantee so far).
I wouldn't rely on it, and rely on the usual syntax instead (but I will not fight for it anyway :) )

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
sylvlecl: I mean that this search is already implemented in the operator [] of the repository, I think we can just call it here. But actually maybe it's only defined in .h but not implemented ? (cannot find it)

src/libs/antares/study/binding_constraint/BindingConstraintGroup.h
sylvlecl: I tend to believe this analysis: https://stackoverflow.com/a/2182819.

If so, RVO will not happen here :(

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-save.cpp
flomnes: So it confirms the usefulness of a proper API with private data members.

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
flomnes: I agree with @sylvlecl

flomnes (approved):

PR #1392, opened 2023-07-03
Fixing test valid-v860/tests-ST-storage

payetvin:


simtest.json
```diff
@@ -1,3 +1,3 @@
 {
-    "version": "v8.6.1"
+    "version": "v8.6.1a"
```
flomnes, line 2: Remove property `storagecycle` from reference study.

flomnes (approved):

JasonMarechal25 (approved):

PR #1393, opened 2023-07-03
Memory allocation, use std::vector 

### LINKED ISSUE #1390: Memory allocation, use std::vector
MemAlloc is still used in those files, replace it with std::vector
- [ ] simulation/sim_allocation_tableaux.cpp
- [ ] optimisation/opt_alloc_probleme_a_optimiser.cpp
- [ ] optimisation/opt_nombre_min_groupes_demarres_couts_demarrage.cpp

payetvin:
- [x] simulation/sim_allocation_tableaux.cpp
- [ ] optimisation/opt_alloc_probleme_a_optimiser.cpp
- [ ] optimisation/opt_nombre_min_groupes_demarres_couts_demarrage.cpp

### DIFF
--- a/src/solver/hydro/management/daily.cpp
+++ b/src/solver/hydro/management/daily.cpp
@@ -237,7 +237,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
     auto const& maxP = maxPower[Data::PartHydro::genMaxP];
     auto const& maxE = maxPower[Data::PartHydro::genMaxE];
 
-    auto& valgen = *ValeursGenereesParPays[numSpace][z];
+    auto& valgen = ValeursGenereesParPays[numSpace][z];
 
     std::shared_ptr<DebugData> debugData(nullptr);
 
--- a/src/solver/hydro/management/daily.cpp
+++ b/src/solver/hydro/management/daily.cpp
@@ -237,7 +237,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
     auto const& maxP = maxPower[Data::PartHydro::genMaxP];
     auto const& maxE = maxPower[Data::PartHydro::genMaxE];
 
-    auto const& valgen = *ValeursGenereesParPays[numSpace][z];
+    auto& valgen = *ValeursGenereesParPays[numSpace][z];
 
     std::shared_ptr<DebugData> debugData(nullptr);
 
--- a/src/solver/hydro/management/daily.cpp
+++ b/src/solver/hydro/management/daily.cpp
@@ -211,7 +211,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
     uint z = area.index;
     assert(z < study.areas.size());
 
-    auto& ptchro = *NumeroChroniquesTireesParPays[numSpace][z];
+    auto& ptchro = NumeroChroniquesTireesParPays[numSpace][z];
 
     auto& inflowsmatrix = area.hydro.series->storage;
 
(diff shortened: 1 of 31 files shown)


flomnes (dismissed):
`TransitMoyenRecalculQuadratique` is no longer initialized

src/solver/hydro/management/daily.cpp
```diff
@@ -237,7 +237,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
     auto const& maxP = maxPower[Data::PartHydro::genMaxP];
     auto const& maxE = maxPower[Data::PartHydro::genMaxE];
 
-    auto const& valgen = *ValeursGenereesParPays[numSpace][z];
+    auto& valgen = ValeursGenereesParPays[numSpace][z];
```
JasonMarechal25, line 240: _Const_ vanished

src/solver/simulation/sim_allocation_tableaux.cpp
```diff
@@ -38,163 +38,52 @@ using namespace Antares;
...
-        rpNtc->VariablesDualesMoyennes = (double*)MemAlloc(sizeOfDoubleHours);
-        rpNtc->RenteHoraire = (double*)MemAlloc(sizeOfDoubleHours);
-    }
-}
-
-static void DeallocateResultsForEconomicMode(void)
-{
-    auto& study = *Data::Study::Current::Get();
-    RESULTATS_PAR_INTERCONNEXION* rpNtc;
+    ResultatsParInterconnexion.resize(1 + study.runtime->interconnectionsCount());
```
flomnes: I don't think the +1 is required here, you should try to remove it

src/solver/simulation/sim_structure_donnees.h
```diff
@@ -70,16 +61,7 @@ typedef struct
...
-    int* TransitMinimumNo;  // UNUSED
-    double* TransitMaximum; // UNUSED
-    int* TransitMaximumNo;  // UNUSED
-    double* TransitStdDev;  // UNUSED
-    double* TransitMoyenRecalculQuadratique;
-    double* VariablesDualesMoyennes; // UNUSED
-    double* TransitAnnuel;           // UNUSED
-    double* RenteHoraire;            // UNUSED
+    std::vector<double> TransitMoyenRecalculQuadratique;
 } RESULTATS_PAR_INTERCONNEXION;
```
JasonMarechal25: Is the struct leftover necessary ?
Can't we just have `std::vector<double> RESULTATS_PAR_INTERCONNEXION` ou `std::vector<double> TransitMoyenRecalculQuadratique`

JasonMarechal25 (commented):
Some comments

src/solver/hydro/management/daily.cpp
payetvin: `valgen.HydrauliqueModulableQuotidien[day] = problem.Turbine[dayMonth];`
It's modified later in the same file, I don't know why it was declared const in the first place

src/solver/simulation/sim_structure_donnees.h
payetvin: We could have a simple std::vector<std::vector<double>> but I think keep the struct name is clearer, @flomnes what do you think ?

flomnes: No, you could keep only TransitMoyen**Interconnexions**RecalculQuadratique

src/solver/hydro/management/daily.cpp
JasonMarechal25: Maybe the pointer was const, not the value

JasonMarechal25 (approved):

PR #1397, opened 2023-07-04
Remove data panels for binding constraints, display message instead

### LINKED ISSUE #1386: BC scenarization : what about historical GUI ?
Currently, we don't know about the behavior of historical GUI regarding binding constraints.
Introducing multiple scenarii for BC RHS probably may have broken something.
As you know, binding constraints part of Antares is daily heavily used. 
- Will users of Antares Simulator v8.7 be able to edit, load and save binding constraints as they used to do ?
- What if a user loads an old study  (version < v8.7) with an Antares v8.7 ? And tries to update this study ?
- In case we broke something, should we change the UI code base in such a way that UI does not crash ?

flomnes:
## Before
image

## After
image

### DIFF
--- a/src/ui/simulator/windows/bindingconstraint/bindingconstraint.cpp
+++ b/src/ui/simulator/windows/bindingconstraint/bindingconstraint.cpp
@@ -51,137 +51,6 @@ namespace Antares
 {
 namespace Window
 {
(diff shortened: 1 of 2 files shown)


JasonMarechal25:
Closes #1386

JasonMarechal25 (approved):

PR #1399, opened 2023-07-05
Avoid leaking char** from Sirius for variable & constraint names

flomnes:
Also update the **antares-deps** submodule.

### DIFF
--- a/src/solver/utils/named_problem.h
+++ b/src/solver/utils/named_problem.h
@@ -17,13 +17,13 @@ private:
     std::vector<char*> Contraintes;
 
 public:
-    PROBLEME_SIMPLEXE_NOMME(std::vector<std::string>& VectNomDesVariables,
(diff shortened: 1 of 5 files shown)


src/solver/utils/named_problem.h
```diff
@@ -17,13 +17,13 @@ struct PROBLEME_SIMPLEXE_NOMME : public PROBLEME_SIMPLEXE
     std::vector<char*> Contraintes;
```
a-zakir, line 17:
```suggestion
    std::vector<char*> Contraintes;
```

a-zakir:
```suggestion
```

a-zakir: same std::vector<char*>  Variables

a-zakir (changes requested):

a-zakir (approved):

PR #1404, opened 2023-07-05
implement --mps-export option

### LINKED ISSUE #1400: CLI option --mps-export has no effect
According to the --help page,
```
      --mps-export           Export in the mps format the optimization
                             problems.
```
However, this does nothing.

Looking at the sources, we only get the following results
```
ag mpsToExport .
File: solver/misc/options.cpp
167:29:    parser->addFlag(options.mpsToExport,

File: libs/antares/study/load-options.cpp
46:2: mpsToExport(false),

File: libs/antares/study/load-options.h
71:10:    bool mpsToExport;
3 matches
```

Meaning that we define a `bool` and set it to true if the option is present. But then it's not used.

a-zakir:
close #1400
With this Pr, ini parameter "include-exportmps" will work with Cli option "--mps-export" according to the following rules:
INI=true, CLI=true => export (both optim)
INI=false,CLI=true => export (both optim)
INI=true, CLI=false => export
INI=false, CLI=false => no export

### DIFF
--- a/src/libs/antares/study/parameters.cpp
+++ b/src/libs/antares/study/parameters.cpp
@@ -1074,9 +1074,7 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
     if (options.usedByTheSolver)
         prepareForSimulation(options);
 
-    if (options.mpsToExport
-        && (this->include.exportMPS == mpsExportStatus::NO_EXPORT
-            || this->include.exportMPS == mpsExportStatus::UNKNOWN_EXPORT))
+    if (options.mpsToExport)
     {
         this->include.exportMPS = mpsExportStatus::EXPORT_BOTH_OPTIMS;
     }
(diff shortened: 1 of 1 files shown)


src/libs/antares/study/parameters.cpp
```diff
@@ -1074,6 +1074,13 @@ bool Parameters::loadFromINI(const IniFile& ini, uint version, const StudyLoadOp
     if (options.usedByTheSolver)
         prepareForSimulation(options);
 
+    if (options.mpsToExport
+        && (this->include.exportMPS == mpsExportStatus::NO_EXPORT
+            || this->include.exportMPS == mpsExportStatus::UNKNOWN_EXPORT))
+    {
+        this->include.exportMPS = mpsExportStatus::EXPORT_BOTH_OPTIMS;
+    }
```
```suggestion
    if (options.mpsToExport)
    {
        this->include.exportMPS = mpsExportStatus::EXPORT_BOTH_OPTIMS;
    }
```

flomnes: I think that we want to force the export of all MPS files if option `--export-mps` is provided by the user.

flomnes (approved):

PR #1422, opened 2023-07-11
Improve unit tests for CR20

flomnes:
- Use RAII for INI (list.ini) & time-series files (CO2Cost.txt, fuelCost.txt)
- Move what can be moved into class `Fixture`
- Clean up


guilpier-code:
As the original PR from @payetvin seems to be merged, I write remarks about it here.
- **src/libs/antares/checks/checkApplication.h** : 
   - function **checkSimplexRangeHydroPricing** is declared twice
   - **checkApplication.h** does not mean anything to me, **checkLoadedInputData.h** could be better for instance.
   - function **checkFuelAndCo2ColumnNumber** : the **And** in its name is a sign that this function should be split.
- **src/libs/antares/study/area/list.hxx** : this source file now contains only one **none template** function. We should move this single function elsewhere and remove the **list.hxx** source file

flomnes:
@guilpier-code done (before your edit) !

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -138,20 +134,16 @@ BOOST_FIXTURE_TEST_SUITE(s, Fixture)
...
-
 }
 
 BOOST_AUTO_TEST_CASE(EconomicInputData_loadFromFolder_basic)
 {
-    createIniFile();
-    createFuelCostFile(8760);
+    FuelCostFile f(8760);
 
     clusterList.loadFromFolder(*study, folder, area);
```
guilpier-code: About` clusterList.loadFromFolder(*study, folder, area);` : 
Are we sure this is useful ? We don't use it in the test, so I ask the question.

I removed this instruction in the following tests and it does not have any effect (they still pass).
- **EconomicInputData_loadFromFolder_basic**
- **EconomicInputData_loadFromFolder_failing_not_enough_value**
- **EconomicInputData_loadFromFolder_working_with_too_much_value**

flomnes:
> function checkFuelAndCo2ColumnNumber : the And in its name is a sign that this function should be split.

You're right, this function can be reduced & split.

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -14,122 +14,118 @@
...
-
-    std::ofstream outfile;
-    outfile.open(folder / "fuelCost.txt", std::ofstream::out | std::ofstream::trunc);
+    explicit ThermalIniFile()
+    {
+        auto folder = getFolder();
+
+        std::ofstream outfile(folder / "list.ini", std::ofstream::out | std::ofstream::trunc);
+
+        outfile << "[area]" << std::endl;
```
guilpier-code: **list.ini** is meant to contain the clusters of a given area.
Here we have a thermal cluster named... **area**, which is confusing.
We should rename this cluster.

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -185,15 +175,14 @@ BOOST_AUTO_TEST_CASE(EconomicInputData_loadFromFolder_working_with_too_much_valu
 
 BOOST_AUTO_TEST_CASE(checkFuelAndCo2_basic_working)
 {
-    createIniFile();
     area->thermal.list.loadFromFolder(*study, folder, area);
     area->thermal.list.mapping["area"]->series = new DataSeriesCommon;
```
guilpier-code: About : 
```c++
area->thermal.list.mapping["area"]->series = new DataSeriesCommon;
```
Causes a memory leak when test ends ?  Or will be reallocated when study is destroyed ?

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -14,122 +14,118 @@
...
-    Fixture(const Fixture && f) = delete;
-    Fixture & operator= (const Fixture & f) = delete;
-    Fixture& operator= (const Fixture && f) = delete;
+    Fixture(const Fixture& f) = delete;
+    Fixture(const Fixture&& f) = delete;
+    Fixture& operator=(const Fixture& f) = delete;
+    Fixture& operator=(const Fixture&& f) = delete;
     Fixture()
     {
         area = study->areaAdd("area");
```
guilpier-code, line 102: To me, a major Antares shortcoming (not bad to recall this from time to time) :

About `area = study->areaAdd("area");` : 
The **Study** is supposed to contain input data only.
here we wee that an area is built and store within the study, that is along with the input data.
This makes input data and simulation objects intertwined. That's why **Study** is a **god** class.
We should have the areas, built from the input data, stored in a list **out** of **Study**.

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
guilpier-code: Besides, about the tests pointed above (lightened of the useless line), are not concerned by the fixture.
They can be moved before  `BOOST_FIXTURE_TEST_SUITE(s, Fixture)`, in a separate test suite

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -185,15 +175,14 @@ BOOST_AUTO_TEST_CASE(EconomicInputData_loadFromFolder_working_with_too_much_valu
 
 BOOST_AUTO_TEST_CASE(checkFuelAndCo2_basic_working)
 {
-    createIniFile();
     area->thermal.list.loadFromFolder(*study, folder, area);
```
guilpier-code, line 165: At first glance, i can see that in this test and in the 2 tests that follow, we have many duplicated lines.
No more to say for now, but worth pointing it.

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -185,15 +175,14 @@ BOOST_AUTO_TEST_CASE(EconomicInputData_loadFromFolder_working_with_too_much_valu
...
     area->thermal.list.mapping["area"]->series->timeSeries.reset(1, 8760);
 
     area->thermal.prepareAreaWideIndexes();
 
-    createFuelCostFile(8760);
-    createCo2CostFile(8760);
+    FuelCostFile fuel(8760);
+    CO2CostFile co2(8760);
 
     BOOST_CHECK(area->thermal.list.mapping["area"]->ecoInput.loadFromFolder(*study, folder));
```
guilpier-code: About `l.areas.erase("area");` : 
Do we need this ? Should normally be destroyed at end of test.

guilpier-code: About `AreaList l(*study);` : 
- Name too short and not meaningful
- But mostly, we shouldn't need it, as we already have `study->areas`, and this list of areas already contain **area**

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -185,15 +175,14 @@ BOOST_AUTO_TEST_CASE(EconomicInputData_loadFromFolder_working_with_too_much_valu
 
 BOOST_AUTO_TEST_CASE(checkFuelAndCo2_basic_working)
```
guilpier-code: Apparently, this test is about the same number of TS between its economic and technical characteristics, and this is the mission of function of **checkFuelAndCo2ColumnNumber(...)**.
But the test's name does not reflect that (and the name of the function does not reflect it either, by the way).

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
guilpier-code, line 165: About `area->thermal.list.loadFromFolder(*study, folder, area);` : 

A test should have one purpose. We suppose here that the current test's purpose is to compare the number of TS of economical and technical sides. 

Why do we load a cluster from a file if we want to test the consistency of technical and economical TS sizes of the cluster ? We can simply create a cluster with the characteristics that we want.

```C++
   ThermalCluster cluster(area);
   
   // Technical characteristics of the cluster
   cluster.series = new DataSeriesCommon;
   cluster.series ->timeSeries.reset(1, 8760);

   // Economical characteristics of the cluster
   cluster.ecoInput.fuelcost.width = 1;
   cluster.ecoInput.co2cost.width = 1;

   area->thermal.list.mapping["my cluster"] = &cluster;
   area->thermal.prepareAreaWideIndexes();
 
   BOOST_CHECK_NO_THROW(Antares::Check::checkFuelAndCo2ColumnNumber(study->areas));
```
Note that all loading functions disappear, including use of class **FuelCostFile** and **CO2CostFile**

guilpier-code (changes requested):

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
```diff
@@ -174,104 +166,73 @@ BOOST_AUTO_TEST_CASE(EconomicInputData_loadFromFolder_failing_not_enough_value)
...
+    CO2CostFile f(10000);
 
     clusterList.loadFromFolder(*study, folder, area);
 
     EconomicInputData eco;
     BOOST_CHECK(eco.loadFromFolder(*study, folder));
 }
 
-BOOST_AUTO_TEST_CASE(checkFuelAndCo2_basic_working)
+BOOST_AUTO_TEST_CASE(checkCo2_basic_working)
```
guilpier-code: For this test and the next 2, the fixture is triggered but unused.
Maybe we should have a different fixture for these 3 tests.

guilpier-code (changes requested):

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
flomnes: Yes, I'm now using `study->areas` now. Not doing so leads to memory leaks/segfaults/etc.

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
flomnes: Yes; I did something extremely close. I used `Matrix<>::reset` instead of setting the `width` manually.

guilpier-code:
The last commit [Simplify tests, fix exception throwing/handling (**a9e807a**)] causes a regression on Windows.
I couldn't reproduce this problem on my Windows machine : locally, everything works fine.

Furthermore, I asked **gitHub** to run the job again, and as you can see, everything's fine now.
Strange !

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
flomnes: The tests above still need `*study`

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
flomnes: valgrind indicates

```
All heap blocks were freed -- no leaks are possible
```

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
flomnes: Done, created a fixture with just `study`, `folder` objects.

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
flomnes: `createIniFile` was removed.

src/tests/src/libs/antares/study/thermal-price-definition/thermal-price-definition.cpp
flomnes: done

guilpier-code (approved):

PR #1437, opened 2023-07-19
tests  end-to-end "in memory" : add BC on clusters

guilpier-code:
The initial purpose of this PR was to make the end-to-end "in memory" on BCs more readable.
Some other improvements came into play in the mean time.
Haphazardly : 
- [x] pave the way to build a practical set of helper classes and functions to build any end-to-end in memory test, as easily as possible
- [x] small helper classes and functions
- [x] make helper functions / classes interfaces independent from implementation details so that they can be easily internally changed without hanging the way they are called.
- [x] change production code, when it's easy, to make easier the end-to-end in memory tests simplifications
- [x] Ajouter une BC sur de la production thermique 

Some possible next step to take on :
- extract the helper classes and functions into several sources files, so that they can be included / joint to any end-to-end in memory test
- Do the same work with the already existing simple-tests, but this by using the helper classes and functions previously defined. For this latter task, see [this PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1457)


src/tests/end-to-end/utils/utils.cpp
```diff
@@ -4,6 +4,208 @@
...
+    unsigned int nbYears = study->parameters.nbYears;
+    for (unsigned int y = 0; y < nbYears; y++)
+        study->parameters.setYearWeight(y, 0.);
+
+    // Set one year's weight to 1
+    study->parameters.setYearWeight(year, 1.);
+
+    // Activate playlist, otherwise previous sets won't have any effect
+    study->parameters.userPlaylist = true;
+}
```
flomnes: This means that all years will be executed, but only the results from 1 year will be used for the computation of synthesis.

Why not enable a single year in the playlist instead ?

src/tests/end-to-end/utils/utils.h
```diff
@@ -4,6 +4,131 @@
...
+
+using namespace Antares::Solver;
+using namespace Antares::Solver::Simulation;
+
+
+void initializeStudy(Study::Ptr study);
+void setNumberMCyears(Study::Ptr study, unsigned int nbYears);
+void configureLinkCapacities(AreaLink* link);
+std::shared_ptr<ThermalCluster> addClusterToArea(Area* area, const std::string& clusterName);
+void addScratchpadToEachArea(Study::Ptr study);
```
flomnes:
```suggestion
void addScratchpadToEachArea(AreaList& areas);
```

src/tests/end-to-end/utils/utils.h
```diff
@@ -4,6 +4,131 @@
...
+// -----------------------
+// BC rhs configuration
+// -----------------------
+class BCrhsConfig
+{
+public:
+    BCrhsConfig() = delete;
+    BCrhsConfig(std::shared_ptr<BindingConstraint> BC, unsigned int nbTimeSeries);
+    void fillTimeSeriesWith(unsigned int TSnumber, double rhsValue);
+
```
flomnes:
```suggestion
    void fillRHSTimeSeriesWith(unsigned int TSindex, double value);

```

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -10,330 +10,110 @@
...
-using namespace Antares::Solver;
-using namespace Antares::Solver::Simulation;
-using namespace Benchmarking;
 
-// ====================
-// Helper functions
-// ====================
+// ===============================
+// Helper functions for BC tests
+// ===============================
```
flomnes: Move these functions to utils.h/utils.cpp ?

src/tests/end-to-end/utils/utils.cpp
```diff
@@ -4,6 +4,208 @@
...
+
+void configureLinkCapacities(AreaLink* link)
+{
+    const double linkCapacityInfinite = +std::numeric_limits<double>::infinity();
+    link->directCapacities.resize(1, 8760);
+    link->directCapacities.fill(linkCapacityInfinite);
+
+    link->indirectCapacities.resize(1, 8760);
+    link->indirectCapacities.fill(linkCapacityInfinite);
+}
```
flomnes, line 20: I don't think this is the right way to set infinite capacities. It will mess with Sirius.

You should use set `link->transmissionCapacities = LocalTransmissionCapacities::infinite` instead.

src/tests/end-to-end/utils/utils.cpp
```diff
@@ -4,6 +4,208 @@
...
+    cluster->setName(clusterName);
+    cluster->reset();
+
+    auto added = area->thermal.list.add(cluster);
+
+    area->thermal.list.mapping[cluster->id()] = added;
+    area->thermal.prepareAreaWideIndexes();
+
+    return cluster;
+}
```
flomnes: Move to main code ? There seems to be code duplication here

src/tests/end-to-end/utils/utils.cpp
```diff
@@ -4,6 +4,208 @@
...
+
+void addScratchpadToEachArea(Study::Ptr study)
+{
+    for (auto [_, area] : study->areas) {
+        for (unsigned int i = 0; i < study->maxNbYearsInParallel; ++i) {
+            area->scratchpad.push_back(AreaScratchpad(*study->runtime, *area));
```
flomnes:
```suggestion
            area->scratchpad.emplace_back(*study->runtime, *area);
```

flomnes (changes requested):

src/tests/end-to-end/utils/utils.cpp
guilpier-code: You're right. It would be faster at execution and code to do that would be simpler.
The original problem is that we can only retrieve synthesis results, a year by year results being erased by results of next year. A way to get a year results  is give a weight to this year only.
But we can activate this year only.

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
guilpier-code: Is done in the following PR on simple-study (see [here)](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1457)

src/tests/end-to-end/utils/utils.h
guilpier-code: It would be more complicated to pass `AreaList& areas`, as we need the **study**'s other data member inside this function

src/tests/end-to-end/utils/utils.h
guilpier-code: Done. 
But it could be redondant : method **fillRHSTimeSeriesWith(...)** belongs to class **BCrhsConfig**, so we know we're dealing we RHS

src/tests/end-to-end/utils/utils.cpp
guilpier-code: Sorry I miss your point, could you be a bit more specific ? which code duplication ?

src/tests/end-to-end/utils/utils.cpp
guilpier-code: Great ! Done

src/tests/end-to-end/utils/utils.cpp
flomnes: This piece of code is a copy-paste of `ThermalClusterList::loadFromFolder`. Maybe we could write a function in libs/antares/study to factorize this.

flomnes (approved):

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -820,6 +823,23 @@ double ThermalCluster::getMarketBidCost(uint serieIndex, uint hourInTheYear) con
...
+            productionCost[h] = marginalCost * modulationCost[h];
+    }
+    else
+    {
+        const auto& marginalCostPerHour = thermalEconomicTimeSeries[0].marginalCostPerHourTs;
+        for (uint h = 0; h != modulation.height; ++h)
+            productionCost[h] = marginalCostPerHour[h] * modulationCost[h];
+    }
+}
+
```
flomnes: here

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -858,6 +862,23 @@ double ThermalCluster::getMarketBidCost(uint serieIndex, uint hourInTheYear) con
...
+            productionCost[h] = marginalCost * modulationCost[h];
+    }
+    else
+    {
+        const auto& marginalCostPerHour = costsTimeSeries[0].marginalCostTS;
+        for (uint h = 0; h != modulation.height; ++h)
+            productionCost[h] = marginalCostPerHour[h] * modulationCost[h];
+    }
+}
+
```
flomnes, line 881: Not a huge fan of this method

`productionCosts` could be a `std::vector` instead of a raw pointer (the `delete[]` is missing I think)

Also production costs are scenarized, so selecting the 1st column is arbitrary.

flomnes (changes requested):
Please remove productionCosts, it seems to be used only in your tests and is quite confusing.

resources/examples
flomnes, line 1: Please revert this

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -559,6 +561,8 @@ void Data::ThermalCluster::reset()
     modulation.fill(1.);
     modulation.fillColumn(thermalMinGenModulation, 0.);
 
+    this->setProductionCost();
+
```
flomnes, line 565: Conflict with #1517

src/libs/antares/study/parts/thermal/cluster.cpp
flomnes: See #1517

flomnes:
@guilpier-code Please resolve conflicts

flomnes:
As requested by @guilpier-code

PR #1452, opened 2023-07-24
Improve clarity for constraints that should be used

JasonMarechal25:
Replace `enabled && !skipped` by `isActive`


flomnes (changes requested):
~~`activeContraints` => `activeBindingConstraints` (at least for the public interface)~~

Sorry, I didn't notice it was already a member of class `BindingConstraintSomething`. Please ignore.

flomnes (approved):

PR #1457, opened 2023-07-25
End-to-end "in memory" tests-simple-study : simplify / clarify

guilpier-code:


src/tests/end-to-end/utils/utils.h
```diff
@@ -8,11 +8,49 @@
 
 using namespace Antares::Solver;
 using namespace Antares::Solver::Simulation;
+using namespace Antares::Data::ScenarioBuilder;
 
 
 void initializeStudy(Study::Ptr study);
-void setNumberMCyears(Study::Ptr study, unsigned int nbYears);
 void configureLinkCapacities(AreaLink* link);
+void addLoadToArea(Area* area, double loadInArea);
```
sylvlecl: I think we should take the habit of using references where argument must not be null, instead of pointer.

src/tests/end-to-end/utils/utils.h
```diff
@@ -8,11 +8,49 @@
...
+
+
+template<class MatixType>
```
sylvlecl: typo in Mat**r**ixType

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -63,18 +30,25 @@ StudyForBCTest::StudyForBCTest()
     Area* area1 = addAreaToStudy("Area 1");
     Area* area2 = addAreaToStudy("Area 2");
 
-    double loadInAreaOne = 0.;
-    addLoadToArea(area1, loadInAreaOne);
+    auto loadTSconfig = std::make_shared<TimeSeriesConfig<Matrix<double, int32_t>>>(area1->load.series->timeSeries);
```
sylvlecl: There is no reason for doing heap allocation and using a shared_ptr here, let's use a plain value on the stack.

src/tests/end-to-end/utils/utils.h
```diff
@@ -8,11 +8,49 @@
...
+class TimeSeriesConfig
+{
+public:
+    TimeSeriesConfig(MatixType& matrix) : ts_(&matrix) {}
+    void setTargetTSmatrix(MatixType& matrix) { ts_ = &matrix; }
```
sylvlecl: I would propose to remove that possibility: the matrix is already provided in the constructor, if we want to change the matrix we can just use a new configurer.

src/tests/end-to-end/utils/utils.h
```diff
@@ -8,11 +8,49 @@
...
 
 
 void initializeStudy(Study::Ptr study);
-void setNumberMCyears(Study::Ptr study, unsigned int nbYears);
 void configureLinkCapacities(AreaLink* link);
+void addLoadToArea(Area* area, double loadInArea);
+
+
+template<class MatixType>
+class TimeSeriesConfig
```
sylvlecl: Maybe "configur**er**" ? The object is not a configuration, it's something that is used to configure the matrix.

src/tests/end-to-end/utils/utils.h
```diff
@@ -8,11 +8,49 @@
...
+{
+public:
+    TimeSeriesConfig(MatixType& matrix) : ts_(&matrix) {}
+    void setTargetTSmatrix(MatixType& matrix) { ts_ = &matrix; }
+    void setNumberColumns(unsigned int nbColumns) { ts_->resize(nbColumns, HOURS_PER_YEAR);}
```
sylvlecl: I think "columns count" would be a better name.

src/tests/end-to-end/utils/utils.h
```diff
@@ -8,11 +8,49 @@
...
+    void fillColumnWith(unsigned int column, double value) { ts_->fillColumn(column, value); }
+private:
+    MatixType* ts_;
+};
+
+class ThermalClusterConfig
+{
+public:
+    ThermalClusterConfig(std::shared_ptr<ThermalCluster> cluster) : cluster_(cluster)
```
sylvlecl: No need for shared_ptr here since the configurer will not own the cluster.

src/tests/end-to-end/utils/utils.h
```diff
@@ -8,11 +8,49 @@
...
+        cluster_->marginalCost = cost;
+        cluster_->marketBidCost = cost; // Must define market bid cost otherwise all production is used
+        cluster_->setProductionCost();
+    }
+    void setAvailablePowerNumberOfTS(unsigned int nbColumns) { tsAvailablePowerConfig_->setNumberColumns(nbColumns); };
+    void setAvailablePower(unsigned int column, double value) { tsAvailablePowerConfig_->fillColumnWith(column, value); }
+
+private:
+    std::shared_ptr<ThermalCluster> cluster_;
+    std::unique_ptr<TimeSeriesConfig<Matrix<double>>> tsAvailablePowerConfig_;
```
sylvlecl: Here also I think there is no point in doing heap allocation for this object, we can have a plain value.

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -117,136 +91,97 @@ BOOST_FIXTURE_TEST_SUITE(TESTS_BINDING_CONSTRAINTS_ON_A_LINK, StudyWithBConLink)
 
 BOOST_AUTO_TEST_CASE(Hourly_BC_restricts_link_direct_capacity_to_90)
```
sylvlecl, line 92: Not part of this PR, but I think we should use test **case** fixtures, not test **suite** fixtures, to better isolate unit tests.
Study should be create from scratch for each unit test, and deleted at the end of the unit test.

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
sylvlecl: For this one and other classes, we can use the classic builder pattern for chaining calls and not repeat the config each time:
```cpp
    TimeSeriesConfig(area1->load.series->timeSeries)
        .setNumberColumns(1)
        .fillColumnWith(0, 0);

    TimeSeriesConfig(area2->load.series->timeSeries)
        .setNumberColumns(1)
        .fillColumnWith(0, 100);
```

sylvlecl (commented):
Some suggestions for improvement.

src/tests/end-to-end/utils/utils.h
guilpier-code: Unused function, was removed

src/tests/end-to-end/utils/utils.h
guilpier-code: Done (as soon as next push is made)

src/tests/end-to-end/utils/utils.h
```diff
@@ -8,13 +8,64 @@
...
+
+template<class MatrixType>
+class TimeSeriesConfig
+{
+public:
+    TimeSeriesConfig(MatrixType& matrix) : ts_(&matrix) {}
+    TimeSeriesConfig& setNumberColumns(unsigned int nbColumns);
+    TimeSeriesConfig& fillColumnWith(unsigned int column, double value);
+private:
+    MatrixType* ts_;
```
flomnes: Maybe it would be nice to have a common `IMatrix` interface, so we don't need to templatize everything.

Something like
```cpp
class IMatrix {
    virtual std::size_t width() const;
    virtual std::size_t height() const;
    virtual void resize(std::size_t width, std::size_t height);
};

template<class T, class U>
class Matrix : public IMatrix {
 // ...
};
```

src/tests/end-to-end/utils/utils.h
guilpier-code: Done

src/tests/end-to-end/utils/utils.h
guilpier-code: Done

src/tests/end-to-end/utils/utils.h
guilpier-code: Done

src/tests/end-to-end/utils/utils.h
guilpier-code: Done

src/tests/end-to-end/utils/utils.h
guilpier-code: Done. For both the cluster and the TS configurer.

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
guilpier-code: Done, as much as I could.

src/tests/end-to-end/utils/utils.h
guilpier-code: If I understand you properly, we would inject the **Matrix** in the **TimeSeriesConfigurer** at construction, meaning that the **TimeSeriesConfigurer**'s constructor would expect an **IMatrix** as parameter,  right ?

flomnes: You'd inject it's `IMatrix` part through [object slicing](https://en.wikipedia.org/wiki/Object_slicing)

sylvlecl (approved):
Nice, I'll just remove template arguments that can be deduced before merging in the other branch.

flomnes (approved):

PR #1461, opened 2023-07-27
Fix incomplete check on constraint group, apply clang-format

flomnes:
# Previous behavior
If the name, id, operator or type of a binding constraint is incorrect, then check the group

# New behavior
Always check the group

Reported by @PPlessiez

### DIFF
--- a/src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
+++ b/src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
@@ -115,32 +115,34 @@ BindingConstraintLoader::load(EnvForLoading env) {
     }
 
     // Checking for validity
-    if (!bc->pName || !bc->pID || bc->pOperator == BindingConstraint::opUnknown ||
-        bc->pType == BindingConstraint::typeUnknown) {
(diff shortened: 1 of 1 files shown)


src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
```diff
@@ -115,48 +128,60 @@ BindingConstraintLoader::load(EnvForLoading env) {
     }
 
     // Checking for validity
-    if (!bc->pName || !bc->pID || bc->pOperator == BindingConstraint::opUnknown ||
-        bc->pType == BindingConstraint::typeUnknown) {
```
flomnes, line 119: This piece of code was removed. That's pretty much it.

JasonMarechal25 (approved):

src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
guilpier-code: This removal was already recommended in the PR of the first version of the developments

guilpier-code (approved):

PR #1465, opened 2023-07-28
Objects in CmakeFiles

JasonMarechal25:


flomnes (changes requested):
${SRC_OBJECT} is now empty in file src/libs/antares/CMakeLists.txt. Please remove line 541. (add_library(antares-core))

flomnes (approved):

PR #1480, opened 2023-08-02
Binding constraint RHS chronicles : fixes asked by Paul P

guilpier-code:
This PR fixes problems raised by Paul P regarding binding constraint RHS chronicles :
- [x] Handle case when a group in **scenariobuilder.dat** does not match any group of study


src/libs/antares/study/scenario-builder/rules.h
```diff
@@ -148,6 +148,7 @@ class Rules final : private Yuni::NonCopyable<Rules>
     Data::AreaLink* getLink(const AreaName& fromAreaName,
                             const AreaName& toAreaName,
                             bool updaterMode);
+    bool checkGroupExists(std::string& group_name);
```
sylvlecl: Constness police ;) : argument should be const ref (and method const, probably).

flomnes: done

PR #1484, opened 2023-08-02
Breakup some lib of libsolver

JasonMarechal25:


sylvlecl:
I am wondering about the granularity we should chose for the new libraries:
I am not sure it's worth it to have a lib for every subdirectory, in particular in the `study` folder.
I think it would make sense to have a single lib for the whole study model, since i really constitutes one single data model (so, it would mean CMakeLists.txt at the study directory level).

PR #1272, opened 2023-04-28
Thermal price definition (RTEi's -> CR20)

Milos-RTEi:
This PR is Identical to closed PR: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1215
The transfer was done in order to use sonar-cloud.

===
This PR is created in order to implement RTEi development on CR20 (Thermal Price Definition).
Spec:
https://rteinternational.sharepoint.com/:w:/r/sites/AntaresUser%27sClub-Interne/Documents%20partages/Change%20Requests/Selected%20for%20development/CR20%20-%20Thermal%20Price%20Definition/Functional_Specifications-Thermal_Price_Definition_V0.1-HA.docx?d=w414939fbdb784b6f957f0bbc32e64a2d&csf=1&web=1&e=TYSOTg

PR #1485, opened 2023-08-02
Thermal price definition (RTEi's -> CR20) - UI remarks from Paul

Milos-RTEi:
This PR is open to clear the issues/remarks Paul has with the UI of https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1272

"In general, the feature works well, but I experienced two issues with the GUI, which would need to be fixed in my view:

1. Units in € are not well displayed in the new tabs “Fuel Cost” and “CO2 cost”.
2. In general, I found that the interface was quite confusing, in particular regarding which parameters are taken into account or not depending on the value of the “Cost generation” parameter. It would be pretty valuable to grey out the parameters which won’t be taken into account depending on the value of the parameter (e.g. grey out “Marginal cost” and “Market bid” in case Cost generation is set to “Use cost timeseries”)."

### DIFF
--- a/src/ui/simulator/windows/inspector/accumulator.hxx
+++ b/src/ui/simulator/windows/inspector/accumulator.hxx
@@ -114,6 +114,37 @@ public:
                                                            PredicateT::TextColor(*i));
         }
     }
+
+    template<class ListT>
+    static void ApplyGreyColor(wxPGProperty* property_one,
+                               wxPGProperty* property_two,
+                               const ListT& list)
+    {
+        assert(list.size() != 0);
+        assert(property_one != NULL);
+        assert(property_two != NULL);
+        if (list.size() == 1)
+        {
+            auto study = *list.begin();
+            property_one->GetGrid()->EnableProperty(property_one->GetBaseName(),
+                                                    PredicateT::Enable(study));
+            property_two->GetGrid()->EnableProperty(property_two->GetBaseName(),
+                                                    PredicateT::Enable(study));
+        }
+        else
+        {
+            auto i = list.cbegin();
+            const auto end = list.cend();
+            ++i;
+            for (; i != end; ++i)
+            {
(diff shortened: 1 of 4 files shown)


PPlessiez:
Thanks for the quick update!

Regarding the first bug (display of the "€" character), everything now seems solved

Regarding the clarity issues, things are definitely better as well! Sorry for being a bit picky but I still have two (minor) comments:
- Could the grey color that you used to grey out elements be a bit lighter (e.g. like the grey on the "min gen modulation" column when zeroes are filled in)? Right now, it's a bit hard to see that these elements are greyed out (at least on my screen, maybe this is also related to the setting of contrast).
- When Cost generation is set to "Set manually", would it be possible to also grey out the fuel cost and CO2 cost time-series in the corresponding tab?

Except these two points, everything looks great!

Milos-RTEi:
Hi Paul. 
Glad you like it. 

As for the additional comments:
1. Unfortunately function "bool DisableProperty( wxPGPropArg id )" used to grey the input fields is an external function from some library "wx-3.0" and does not provide the option to change the color. And since it is external I cannot modify it. 
I even tried to change the color of the field first with the function "SetPropertyTextColour" from the same library. And then disable it with "DisableProperty". But the color is then just overwritten. 
 
2. This one seemed easy to me. At least in the beginning. But after an hour+ of trying I gave up. The TS-column is completely another object type, so I cannot just add one line of code and re-call disable function one more time for TS-columns. Even more TS-Column objects are not even accessible at the point where I did this "disable+gray" modification. I tried some hybrid solutions but they did not work. Maybe if I get a hint from the guys who know GUI better we can re-try to do it later.

PPlessiez (approved):
Thanks for the update Milos!
If it cannot be handled easily, it's okay, what you did is already an improvement compared to the first version. We'll see if this needs to be fixed at a later stage depending on user feedback!

flomnes (approved):

PR #1494, opened 2023-08-03
Remove Study::Get from SIM, pass as function argument instead

### LINKED ISSUE #1119: Remove global variables
**Description**

The presence of global variables and singletons has major drawbacks:
- it makes more or less impossible the use of the solver as a lib, because the user will get bugs when running multiple simulations, at least in parallel, and maybe also in sequence
- it hurts understandability of the code, because functions don't tell correctly their inputs and outputs
- it makes very hard unit testing, for the same reason

Therefore, removing those globale variables is a must-do.
The culprits are:
- `gGlobalCurrentStudy` in `current.cpp`
- all variables related to Monte Carlo draws in `sim_extern_variables_globales.h`

**Implementation**

All those variables must be explicitly provided to functions using them (relevance of including them in existing classes needs to be assessed).

flomnes:


sylvlecl (approved):
(We'll have some minor conflicts with the "optimisation" side)

PR #1505, opened 2023-08-04
Replace remaining vector of pointers by plain vector

sylvlecl:


flomnes:
You should replace
```cpp
CORRESPONDANCE* corresp = &problemeHebdo->Correspondance;
```
by
```cpp
CORRESPONDANCE& corresp = problemeHebdo->Correspondance;
```

payetvin (approved):

PR #1506, opened 2023-08-04
Remove tmpnam from tests

### LINKED ISSUE #1493: the use of `tmpnam' is dangerous, better use `mkstemp'
Fix this annoying warning for good ?

payetvin:


flomnes (commented):
The behavior of `std::filesystem::temp_directory_path` is not strictly equivalent.

You just removed a call to `tmpnam`, which is guaranteed to create a **unique** directory, avoiding possible side-effects (a file that wasn't removed by the previous test makes this one succeed/fail).

flomnes (changes requested):
Also, /tmp already exists, no need to create it

src/tests/src/libs/antares/study/constraint/test_constraint.cpp
```diff
@@ -15,6 +15,15 @@
...
+fs::path generateAndCreateDirName()
+{
+    std::string dir = boost::unit_test::framework::current_test_case().p_name;
+    fs::path working_dir = fs::temp_directory_path() / dir;
+    fs::remove(working_dir);
+    fs::create_directories(working_dir);
+    return working_dir;
+}
+
 BOOST_AUTO_TEST_SUITE(BindingConstraintTests)
```
flomnes: You may need to remove unwanted characters, not all characters are valid in a path

Also, please use a single declaration/definition

payetvin: I don't see which unwanted characters could be found, we're only using alpha numeric chars and underscores in test names.

This function is used in different places, should it be moved in a header at the root of src/tests ?

flomnes: src/tests/end-to-end/utils/utils.h

src/tests/src/solver/simulation/test-time_series.cpp
```diff
@@ -43,9 +44,8 @@ struct Fixture {
     Fixture() {
         study = std::make_shared<Study>();
         study->header.version = version870;
-        auto tmp_dir = fs::temp_directory_path();
-        working_tmp_dir = tmp_dir / std::tmpnam(nullptr);
-        fs::create_directories(working_tmp_dir);
+        std::string testName = boost::unit_test::framework::current_test_case().p_name;
+        working_tmp_dir = generateAndCreateDirName(testName);
```
flomnes: You could create this macro in utils.h / utils.cpp

src/tests/src/libs/antares/study/constraint/test_constraint.cpp
```diff
@@ -5,12 +5,15 @@
...
 #define BOOST_TEST_MODULE binding_constraints
 
-#include <boost/test/included/unit_test.hpp>
+#include <boost/test/unit_test.hpp>
 #include <fstream>
 #include "antares/study/constraint.h"
 #include "antares/study/area/area.h"
 #include "antares/study.h"
 #include <filesystem>
+#include "../../../../utils/utils.h"
```
flomnes: You should not need to use "../../../"

src/tests/src/libs/antares/study/constraint/test_constraint.cpp
```diff
@@ -5,12 +5,15 @@
...
+
+#define testName boost::unit_test::framework::current_test_case().p_name
```
flomnes: Define in utils.h ? Or even better
```cpp
CREATE_TMP_DIR_BASED_ON_BOOST_TEST_NAME();
```

flomnes (changes requested):
I believe that utils.h/.cpp already exist in `develop`, please take a look

src/tests/src/libs/antares/study/constraint/test_constraint.cpp
payetvin: Fixed

src/tests/src/libs/antares/study/constraint/test_constraint.cpp
```diff
@@ -5,12 +5,13 @@
 #define WIN32_LEAN_AND_MEAN
 #define BOOST_TEST_MODULE binding_constraints
 
-#include <boost/test/included/unit_test.hpp>
+#include <boost/test/unit_test.hpp>
```
flomnes, line 10: Please add comment here 
// Why this header ? See here https://www.boost.org/doc/libs/1_82_0/libs/test/doc/html/boost_test/adv_scenarios/single_header_customizations/multiple_translation_units.html

src/tests/src/utils/utils.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <filesystem>
+
+#define CREATE_TMP_DIR_BASED_ON_TEST_NAME() generateAndCreateDirName(boost::unit_test::framework::current_test_case().p_name);
+
+std::filesystem::path generateAndCreateDirName(std::string);
```
flomnes:
```suggestion
std::filesystem::path generateAndCreateDirName(const std::string&);
```

src/tests/src/utils/utils.cpp
```diff
@@ -0,0 +1,11 @@
+#include "utils.h"
+
+namespace fs = std::filesystem;
+
+fs::path generateAndCreateDirName(std::string dirName)
```
flomnes:
```suggestion
fs::path generateAndCreateDirName(const std::string& dirName)
```

src/tests/src/utils/utils.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <filesystem>
+
+#define CREATE_TMP_DIR_BASED_ON_TEST_NAME() generateAndCreateDirName(boost::unit_test::framework::current_test_case().p_name);
```
flomnes, line 6: Add a comment here to explain why we need a macro (linking issues with boost)

flomnes (changes requested):

flomnes (changes requested):
Merge **src/tests/src/utils** (new) with **src/tests/end-to-end/utils/** (existing) ?

flomnes (changes requested):
As requested by @sylvlecl, please create a single **testing_utils** CMake target.

src/tests/src/utils/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+add_library(test_utils_unit
+        utils.cpp
+        utils.h
+        )
+
+target_include_directories(
+        test_utils_unit
+        PUBLIC
+        .
```
flomnes: ${CMAKE_CURRENT_SOURCE_DIR} ?

flomnes (changes requested):

flomnes (approved):

PR #1507, opened 2023-08-04
Split area

JasonMarechal25:


src/libs/antares/CMakeLists.txt
```diff
@@ -533,6 +509,7 @@ add_library(libantares-core
 	${SRC}
 	${SRC_LOGS}
 	${SRC_MEMORY}
+	${SRC_OBJECT}
```
flomnes: Please remove this

src/ui/simulator/toolbox/input/renewable-cluster.h
```diff
@@ -42,6 +42,7 @@
 #include "../components/htmllistbox/item/renewable-cluster-item.h"
 #include "../components/htmllistbox/datasource/renewable-cluster-order.h"
 #include "../components/htmllistbox/item/renewable-cluster-item.h"
+#include "area.h"
```
flomnes, line 45: Unnecessary ?

flomnes (changes requested):

src/ui/simulator/toolbox/input/renewable-cluster.h
JasonMarechal25: It's area.h of toolbox, not of study

src/libs/antares/study/area/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+source_group("study\\area" FILES ${SRC_AREAS})
+
+add_library(area
+        ${SRC_AREAS})
+
+target_include_directories(area
+        PRIVATE
+        ${CMAKE_SOURCE_DIR}/ext/yuni/src
+        ${CMAKE_SOURCE_DIR}/solver/writer
+        ${CMAKE_SOURCE_DIR}/libs
```
sylvlecl, line 25: These are cyclic dependencies:
the new area lib depends on includes from downstream libs (for example study.h) which themselves depend on the area lib.

We must avoid this, it defeats the purpose of having smaller libraries with clear dependencies.
(in particular, changing study.h will still trigger the rebuild of the area library).

So, we should either:
- not separate this lib
- separate it properly, with removal of reverse dependencies

We must take care of libraries depending **only** on headers from upstream libraries.
We need to find what's the best way to ensure this with cmake. In the best case, headers should come simply through the definition of linked libraries. cmake 3.23 introduced the feature of "files set" which allows to define properly public headers, it seems it was not so clear before that version. To be investigated !

sylvlecl (changes requested):
See comment: I don't think we should merge this in the current state, this introduces cyclic dependencies between libraries.
We need to be consistent between linked libraries and included headers. We need to identify best way to achieve this with cmake.

JasonMarechal25:
Closing to prevent conflict with existing PRs and redo following new guidelines

PR #1509, opened 2023-08-04
Remove more usage of study singleton

sylvlecl:
**Related issue**
#1119 

**Description**

Removes last usages of the study singleton in the solver part, **except in technical parts** (signal handling and emergency shutdown...).
Those parts must be limited to the application scope, in a future PR.


flomnes (approved):

flomnes:
> except in technical parts

Yes, this is the hard part. For signal handlers, I'm afraid we don't have much of a choice.

PR #1513, opened 2023-08-07
Remove study singleton in application signal handlers

sylvlecl:
**Description**

Removal of `Study::Current` usage in signal handlers.
Instead, the application uses its own singleton for that particular use.
This way, it's really the responsibility of the application (the executable) to maintain its study, the library does not manage it at all.

**Implementation note:**
we use a `weak_ptr` so that the study is not kept alive only for the use of those signal handlers.

### DIFF
--- a/src/solver/signal-handling/common.cpp
+++ b/src/solver/signal-handling/common.cpp
@@ -6,8 +6,6 @@ using namespace Antares;
 
 namespace Antares::Solver {
 
-using Antares::Solver::IResultWriter;
-
 // weak_ptr because the lifetime must not be managed by this feature.
 // If the study lifetime ends before we receive a signal, we shoud just do
 // nothing when receiving the signal.
@@ -21,10 +19,8 @@ void setApplicationResultWriter(std::weak_ptr<IResultWriter> writer)
 }
 
 namespace {
-
-static void finalizeWrite() {
-    auto writer = Antares::Solver::APPLICATION_WRITER.lock();
-    if (writer) {
+void finalizeWrite() {
+    if (auto writer = Antares::Solver::APPLICATION_WRITER.lock()) {
         writer->finalize(true);
     } else {
         logs.warning() << "Could not finalize write: invalid writer";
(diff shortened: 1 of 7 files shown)


src/solver/signal-handling/common.cpp
```diff
@@ -4,29 +4,38 @@
...
+    APPLICATION_WRITER = writer;
+}
+
+}
+
+namespace {
+
+static void finalizeWrite() {
+    auto writer = Antares::Solver::APPLICATION_WRITER.lock();
+    if (writer) {
```
flomnes:
```suggestion
    if (auto writer = Antares::Solver::APPLICATION_WRITER.lock()) {
```
See [here](https://en.cppreference.com/w/cpp/memory/weak_ptr), where a similar syntax is used (C++17)

src/solver/signal-handling/common.cpp
```diff
@@ -4,29 +4,38 @@
 
 using namespace Antares;
 
-static void finalizeWrite()
+namespace Antares::Solver {
+
+using Antares::Solver::IResultWriter;
+
```
flomnes:
```suggestion
```

We're already in this namespace

src/solver/signal-handling/common.cpp
```diff
@@ -4,29 +4,38 @@
...
-    {
-        logs.warning() << "Could not finalize write: invalid study";
+    APPLICATION_WRITER = writer;
+}
+
+}
+
+namespace {
+
+static void finalizeWrite() {
```
flomnes:
```suggestion
void finalizeWrite() {
```

Everything is `static` in the anonymous namespace.

src/solver/signal-handling/public.h
```diff
@@ -1,6 +1,9 @@
 #pragma once
 
+#include <memory>
+#include <antares/study.h>
```
flomnes:
```suggestion
#include <writer/i_writer.h>
```

src/solver/signal-handling/common.h
```diff
@@ -1,4 +1,15 @@
 #pragma once
 
+#include <memory>
+#include "antares/study.h"
+
```
flomnes:
```suggestion
#include <writer/i_writer.h>

```

flomnes (changes requested):

flomnes (approved):

PR #1516, opened 2023-08-07
Tests end-to-end in memory : simple tests (part 2)

guilpier-code:
- [x] merge from **develop**
- [x] fix a bug in end-to-end in memory tests after merge


src/libs/antares/study/study.cpp
```diff
@@ -72,9 +72,9 @@ static inline void FreeAndNil(T*& pointer)
 Study::Study(bool forTheSolver) :
  LayerData(0, true),
  simulationComments(*this),
- maxNbYearsInParallel(0),
+ maxNbYearsInParallel(1),
  maxNbYearsInParallel_save(0),
- nbYearsParallelRaw(0),
+ nbYearsParallelRaw(1),
  minNbYearsInParallel(0),
  minNbYearsInParallel_save(0),
```
flomnes: Constants can (and should) be defined in study.h

src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp
```diff
@@ -2,548 +2,276 @@
...
-    unsigned int hour = 100;
-    BOOST_TEST(getLinkFlowAthour(simulation->get(), link, hour) >= rhsValue, tt::tolerance(0.001));
+    BOOST_TEST(output->thermalGeneration(cluster.get()).hour(10) == rhsValue, tt::tolerance(0.001));
 }
 
 BOOST_AUTO_TEST_SUITE_END()
 
 
-BOOST_FIXTURE_TEST_SUITE(TESTS_ON_BC_RHS_SCENARIZATION, Fixture)
+BOOST_FIXTURE_TEST_SUITE(TESTING_BC_RHS_SCENARIZATION_WHEN_BC_ON_A_LINK, StudyWithBConLink)
```
flomnes, line 220: Please keep using `BOOST_FIXTURE_TEST_SUITE`, so that `Fixture::Fixture` and `Fixture::~Fixture` will be called for each test.

flomnes (changes requested):
You may have some conflicts after the removal of `ThermalCluster::productionCosts` (#1517)

src/libs/antares/study/area/area.cpp
```diff
@@ -46,23 +46,22 @@ void Area::internalInitialize()
...
+     enabled(true),
+     reserves(fhrMax, HOURS_PER_YEAR),
+     miscGen(fhhMax, HOURS_PER_YEAR),
+     nodalOptimization(anoAll),
+     spreadUnsuppliedEnergyCost(0.),
+     spreadSpilledEnergyCost(0.),
+     filterSynthesis(filterAll),
+     filterYearByYear(filterAll),
+     ui(nullptr),
+     invalidateJIT(false)
```
flomnes: Set constants in area.h to avoid code duplication / mistakes. Leave `reserves` & `miscGen` here.

src/tests/end-to-end/utils/utils.h
```diff
@@ -4,24 +4,189 @@
...
+    typename Variable::Storage<VCard>::ResultsType* retrieveAreaResults(Area* area);
+
+    template<class VCard>
+    typename Variable::Storage<VCard>::ResultsType* retrieveLinkResults(AreaLink* link);
+
+    template<class VCard>
+    typename Variable::Storage<VCard>::ResultsType* retrieveResultsForThermalCluster(ThermalCluster* cluster);
 
-float defineYearsWeight(Study::Ptr pStudy, const std::vector<float>& yearsWeight);
+    std::shared_ptr<ISimulation<Economy>>& simulation_;
```
sylvlecl: Weird to have a reference to smart pointer:
we should have either one or the other.

src/tests/end-to-end/utils/utils.h
```diff
@@ -4,24 +4,189 @@
...
+private:
+    Variable::R::AllYears::AverageData& averageResults_;
+};
+
+
+class OutputRetriever
+{
+public:
+    OutputRetriever(std::shared_ptr<ISimulation<Economy>>& simulation) : simulation_(simulation) {}
+    averageResults overallCost(Area* area);
```
sylvlecl, line 86: Not very important as long as we are in tests, but it would be better if we ask arguments as reference since it should not be null, and it should probably be const because I don't expect it to modifiy the objects (Area and so on).

sylvlecl (commented):
A few not blocking comments about the API

sylvlecl (approved):

flomnes (approved):

PR #1518, opened 2023-08-07
Write full command-line instead of solver location in logs

flomnes:


src/solver/misc/write-command-line.h
```diff
@@ -0,0 +1,3 @@
+#pragma once
+
+void WriteCommandLineIntoLogs(int argc, char** argv);
```
sylvlecl, line 5: We should move it to `Antares::Solver::` namespace

src/solver/misc/write-command-line.cpp
```diff
@@ -0,0 +1,13 @@
+#include <antares/logs.h>
+#include <sstream>
+
+void WriteCommandLineIntoLogs(int argc, char** argv)
```
sylvlecl, line 6: Note that we could also have a more "structured" output if we print `StudyLoadOptions` instead of raw command line arguments.

sylvlecl (changes requested):
namespace to be added

src/solver/misc/write-command-line.cpp
sylvlecl: (Indeed actually it's not enough, there is also `Settings`)

sylvlecl (approved):

PR #1519, opened 2023-08-08
Remove use of current study singleton in emergency exit

sylvlecl:
**Description**

The "current study" singleton is still used in the "emergency shutdown", which calls exit.
This evolution removes those legacy calls to exit, replacing them by an exception throw.

- Exceptions are now caught and logged at the end of the executable.
- the writer finalization and the log writing is done when the study object goes out of scope (RAII).
  In particular, when an exception is thrown and caught at the end of the executable, they will still be correctly finalized.

Additional information:
- removed the `releaseAllLocks` method which is a no-op
- use of `std::set_new_handler` is removed, we rely on std::bad_alloc instead


src/libs/antares/fatal-error.h
```diff
@@ -24,28 +24,22 @@
...
-    ** This routine will call exit() and will never return.
-    */
-    void AntaresSolverEmergencyShutdown(int code = 1) YUNI_NORETURN;
+/*!
+** \brief A generic exception for errors that should end the program.
+*/
+class FatalError : public std::runtime_error
+{
+public:
+    explicit FatalError(const std::string& message);
```
flomnes:
```suggestion
    using std::runtime_error::runtime_error;
```

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -242,12 +240,11 @@ void StudyRuntimeInfos::initializeRangeLimits(const Study& study, StudyRangeLimi
     // weeks, this value must be greater than or equal to 168
     if (limits.hour[rangeCount] < 168)
     {
-        logs.info();
         logs.fatal() << "At least one week is required to run a simulation.";
```
flomnes: Remove this log line ? It should be printed by the catcher anyway

src/solver/hydro/management/monthly.cpp
```diff
@@ -214,13 +214,14 @@ void HydroManagement::prepareMonthlyOptimalGenerations(double* random_reservoir_
...
+            {
+                std::ostringstream msg;
+                msg << "Year : " << y + 1 << " - hydro: " << area.name
+                    << " [month] no solution found";
+                throw FatalError(msg.str());
+            }
             case EMERGENCY_SHUT_DOWN:
-                AntaresSolverEmergencyShutdown();
-                break;
+                throw FatalError("msg.str()");
```
flomnes:
```suggestion
                throw FatalError(msg.str());
```

src/solver/main.cpp
```diff
@@ -112,53 +97,64 @@ const char* const ANTARES_LOGO = "\n\n"
...
-        return EXIT_FAILURE;
+        logs.info(ANTARES_LOGO);
+        logs.info(GPL_ANNOUNCEMENT);
+        // Name of the running application for the logger
+        logs.applicationName("solver");
 
-    // locale
-    InitializeDefaultLocale();
+        if (not memory.initializeTemporaryFolder())
+            return EXIT_FAILURE;
```
flomnes:
```suggestion
            throw FatalError("Could not initialize temporary folder");
```

src/solver/main.cpp
```diff
@@ -112,53 +97,64 @@ const char* const ANTARES_LOGO = "\n\n"
...
+        application.execute();
+        application.writeExectutionInfo();
 
-    FreeUTF8Arguments(argc, argv);
+        FreeUTF8Arguments(argc, argv);
 
-    // to avoid a bug from wxExecute, we should wait a little before returning
-    SuspendMilliSeconds(200 /*ms*/);
+        // to avoid a bug from wxExecute, we should wait a little before returning
+        SuspendMilliSeconds(200 /*ms*/);
```
flomnes, line 146: TODO: remove

src/solver/application.cpp
```diff
@@ -285,8 +271,7 @@ void Application::resetLogFilename() const
 
     if (!logs.logfileIsOpened())
     {
-        logs.error() << "Impossible to create " << logfile;
-        AntaresSolverEmergencyShutdown(); // will never return
+        throw FatalError(std::string("Impossible to create ") + logfile.c_str());
```
```suggestion
        throw FatalError(std::string("Impossible to create the log file at ") + logfile.c_str());
```

flomnes (changes requested):

src/libs/antares/fatal-error.h
sylvlecl: Done, I didn't know/remember about that syntax, thank you !

src/libs/antares/study/runtime/runtime.cpp
sylvlecl: Indeed, done.

src/solver/hydro/management/monthly.cpp
sylvlecl: Done.

src/solver/main.cpp
sylvlecl: Done.

src/solver/application.cpp
sylvlecl: Done.

src/solver/main.cpp
sylvlecl: I left it for that PR: do we create an issue for this, to be done once GUI is actually removed ?

flomnes: Yes ! To be added to the long list of things to do once the GUI is removed

src/solver/hydro/management/daily.cpp
```diff
@@ -48,6 +48,25 @@ using namespace Yuni;
...
 
+namespace
+{
+FatalError fatalError(const std::string& areaName, int year)
+{
+    std::ostringstream msg;
+    msg << "Year : " << year + 1 << " - hydro: " << areaName
+        << " [daily] fatal error";
+    return FatalError(msg.str());}
+
```
flomnes:
```suggestion
    return FatalError(msg.str());
}

```

src/solver/main.cpp
```diff
@@ -112,53 +97,64 @@ const char* const ANTARES_LOGO = "\n\n"
...
+    catch (const std::bad_alloc& exc) {
+        logs.fatal() << exc.what();
+        logAbortion();
+        return ALLOCATION_FAILURE_EXIT_CODE;
+    }
+    catch (const std::exception& exc) {
+        logs.fatal() << exc.what();
+        logAbortion();
+        return EXIT_FAILURE;
+    }
```
flomnes, line 159: `catch (...)` to catch other exceptions here ?

sylvlecl: Indeed it seems best to add it:
- it seems that stack unwinding is not required by the standard when an exception is completely uncaught (https://stackoverflow.com/a/2820407). So if we want to ensure all destructors are called, we need to catch everything
- also, I am not sure what is the exit code on exception, it's better to make it explicit here

Note that this is an evolution compared to the current state, where exceptions are indeed not caught.

flomnes (approved):

PR #1521, opened 2023-08-08
Hydro final lvl (CR 25) to merge in develop [ANT-1084]

guilpier-code:
This PR attempts some code enhancements on branch **feature/final-reservoir-level-cr25**.
Therefore, this PR's associated branch (**fix/hydro-final-levels-cr25--try-fixes**) is based on **feature/final-reservoir-level-cr25**.
- [x]  in class **FinalLevelInflowsModifier**, 
    - try to reduce the number of data members, 
    - making data members **private** whenever it's possible,
    - define a clear interface with a smallest number of public functions
    - and so make the more methods private as possible
- [x] **Unit tests** : clarify their intention and reduce their size
- [ ] **caution** : could not remove yet the **initialize** method. This method should be merged with the constructor, but when we call the constructor, all input pieces of data are not available. that's why **initialize** exists. We should try to make these pieces of data available, for instance by changing the order of data loading

For any purpose, here are doc resources : 
- specs [here](https://rteinternational.sharepoint.com/:w:/r/sites/AntaresUser'sClub-Interne/_layouts/15/Doc.aspx?sourcedoc=%7B7774488E-7E4C-4AEC-B11F-1C9728B96AC0%7D&file=Functional_Specifications-Final_reservoir_level_V0.1-MA.docx&action=default&mobileredirect=true)
- a discussion on that matter [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/discussions/1145)

---
### CAUTION :
**Some cleaning and corrections were introduced in a new pull request** (see [this PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2040)).
This new PR was merged into this one.
**This PR contains a test under the form of a study.
This study should be downloaded and added to the right test repo.**

---
### Left to to

- [ ] Merge from branch develop and remove conflicts
- [ ] Check that all tests exist (unit test for raising failures and study case for fine working)
- [ ] what else ?


Milos-RTEi (approved):

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
```diff
@@ -28,64 +28,54 @@
...
 #include "container.h"
 
-namespace Antares
+namespace Antares::Data
 {
-namespace Data
+
+bool isValidLevel(double level)
 {
+    return (level < 0. || isnan(level)) ? false : true;
```
flomnes:
```suggestion
    return level >= 0. && !isnan(level);
```

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
```diff
@@ -94,90 +84,92 @@ double FinalLevelInflowsModifier::calculateTotalInflows() const
...
+                                           const Matrix<double>& scenarioFinalHydroLevels,
+                                           const uint lastSimulationDay,
+                                           const uint nbYears)
 {
-    fillEmpty();
-    initializeGeneralData(parameters, year);
-    initializePerAreaData(scenarioInitialHydroLevels, scenarioFinalHydroLevels);
-    initializePreCheckData();
-    ruleCurveForSimEndReal();
+    isApplicable_ = std::move(std::vector<bool>(nbYears, false));
```
flomnes:
```suggestion
    isApplicable_.assign(nbYears, false);
```

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
```diff
@@ -94,90 +84,92 @@ double FinalLevelInflowsModifier::calculateTotalInflows() const
...
+    deltaLevel = std::move(std::vector<double>(nbYears, 0.));
```
flomnes: Same

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
```diff
@@ -31,34 +31,38 @@
...
 
-          finalInflows.initializeData(
-            scenarioInitialHydroLevels, scenarioFinalHydroLevels, parameters, year);
+void CheckFinalReservoirLevelsForYear(Data::Study& study, uint year)
+{
+    study.areas.each([&study, &year](Data::Area& area)
+    {
+        if (!area.hydro.finalLevelInflowsModifier.CheckInfeasibility(year))
+        {
+            AntaresSolverEmergencyShutdown();
```
flomnes: This function was removed by @sylvlecl, please `throw` instead.

src/solver/simulation/hydro-final-reservoir-level-functions.h
```diff
@@ -31,8 +31,7 @@
 
 namespace Antares::Solver
 {
-void prepareFinalReservoirLevelDataPerMcY(Data::Study& study, uint year);
-void prepareFinalReservoirLevelData(Data::Study& study);
+void CheckFinalReservoirLevelsConfiguration(Data::Study& study);
```
flomnes: Any chance `study` might be `const` ?

src/solver/simulation/solver.hxx
```diff
@@ -344,7 +345,10 @@ void ISimulation<Impl>::run()
         }
 
         if (parameters.useCustomScenario)
+        {
             ApplyCustomScenario(study);
+            CheckFinalReservoirLevelsConfiguration(study);
```
flomnes: Does this also apply changes ? If so, rename the function accordingly.

flomnes (changes requested):

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
Milos-RTEi: FatalError is not declared here. 
I guess we can implement "throw FatalError" once merging with [feature/final-reservoir-level-cr25].
[feature/final-reservoir-level-cr25] is up to date with the develop branch.

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
guilpier-code: Done

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
guilpier-code: Done

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
guilpier-code: Done

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
guilpier-code: Done

src/solver/simulation/hydro-final-reservoir-level-functions.h
guilpier-code: Done (by @Milos-RTEi)

src/solver/simulation/solver.hxx
guilpier-code: What do you mean ?
This new function does what its name explains

Milos-RTEi:
The PR: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1370 is closed and transferred to this branch/PR
Visit closed PR for links for specifications and test-example.

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
```diff
@@ -0,0 +1,175 @@
...
+    isApplicable_.at(year) = checksOk;
+
+    return checksOk;
+}
+
+bool FinalLevelInflowsModifier::isApplicable(uint year)
+{
+    // If isApplicable_.size() == 0, then instance was not properly initialized
+    // and is not applicable.
+    return isActive() && isApplicable_.size() && isApplicable_.at(year);
```
Milos-RTEi: Calling method isActive() works quite ok in CheckInfeasibility.
However, calling isActive() in method isApplicable() (that is only called in HydroManagement method changeInflowsToAccommodateFinalLevels) is redundant and even more dangerous.
Reason -  isActive() is called with stale data for finalReservoirLevel_ & initialReservoirLevel_ ( last one - from the last year). 
So, if we have 10 MCy. Let's assume that the logic should be implemented for the first 9 years. And for the last (10th) year logic should not be implemented because of the invalid init/final reservoir levels. What will actually happen. The logic implementation will be skipped for all 10 years.

Milos-RTEi (dismissed):
see comment above and PR https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1564.

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
guilpier-code: I totally agree on the error introduced in this PR.
As you suggested, removing the call to **isActive()** inside **isApplicable()** is the shortest way to correct this error.
If the error was not caught sooner, it's because a unit test is missing, right ?
So I will suggest an additional test in the following PR (https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1564.)

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
```diff
@@ -0,0 +1,68 @@
...
+                                                        study.parameters.simulationDays.end,
+                                                        study.parameters.nbYears);
+    });
+}
+
+void CheckFinalReservoirLevelsForYear(const Data::Study& study, uint year)
+{
+    study.areas.each([&study, &year](Data::Area& area)
+    {
+        if (!area.hydro.finalLevelInflowsModifier.CheckInfeasibility(year))
```
guilpier-code: I see a problem here : what if a year is not played (see play list) ?
Do we still want to make checks for this year ?

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
```diff
@@ -0,0 +1,175 @@
...
+                                                     const AreaName& areaName) :
+    hydro_(hydro), areaIndex_(areaIndex), areaName_(areaName)
+{
+}
+
+bool FinalLevelInflowsModifier::CheckInfeasibility(uint year)
+{
+    ComputeDelta(year);
+
+    if (!isActive())
```
guilpier-code: **Question** : 
Here we have : 
```C++
if (!isActive())
    return true;
```
We could expect to return **false**, not **true**. For instance if the initial level is a **nan**, we cannot exploit it, and so we should prevent any final level computation for this area and this hour.
So why do we return **true** ?

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.h
```diff
@@ -0,0 +1,90 @@
...
+    bool CheckInfeasibility(uint year);
+    bool isApplicable(uint year);
+
+    // vectors containing data necessary for final reservoir level calculation
+    // for one area and all MC years
+    // vector indexes correspond to the MC years
+    std::vector<double> deltaLevel;
+
+
+private:
```
guilpier-code: **Idea for a possible refactoring** 
I really hesitated to do it, but eventually resisted, partially for lack of time, and because some questions remain.
Class **FinalLevelInflowsModifier** does checks and allows computing a final level for a given area.
But we can see that a lot of public and private methods have the year as argument.
I see it as a sign that this class should either : 
- become mainly a container for a list  of yearly final level inflow modifier (let's call it class **YearlyLvlModfier** for instance).
- replaced by a list of **YearlyLvlModfier**.

Maybe there would a benefit to keep **FinalLevelInflowsModifier** as a container, for example that we would not have to change the way it's used in the production code or unit tests. Rather we would change the code of **FinalLevelInflowsModifier** internally.

Anyway, the purpose of this refactoring would be to have classes with less responsibilities, and possibly only one. As usual, it would make code : 
- more clear (it maybe would introduce more classes, but smaller and so easier to understand)
- more changeable, as small classes usually facilitates this
- more testable (we could test **YearlyLvlModfier** in isolation, and, as it is supposed to be smaller than **FinalLevelInflowsModifier**, would be unit tested more easily). Same thing for **FinalLevelInflowsModifier**, possibly much smaller.

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
Milos-RTEi: Good point!
Do we still want to make checks for this year ? -> I do not know. I guess not. We should skip for this year, right?

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.cpp
Milos-RTEi: Similar to https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1564#discussion_r1317866196
We return true just to skip:
image
but not to stop the simulation. 
We return false if pre-checks fail and then we stop the simulation.

src/libs/antares/study/parts/hydro/finallevelinflowsmodifyer.h
Milos-RTEi: I have to read this more carefully tomorrow. If we keep the class just as the container or switch to a list. 
Where do we move out all FinalLevelInflowsModifier - methods?

src/solver/simulation/hydro-final-reservoir-level-functions.cpp
guilpier-code: > I guess not. We should skip for this year, right?

Yes. Done by @Milos-RTEi

flomnes:
**Final reservoir level not applicable** should be a warning, not an info

src/libs/antares/study/include/antares/study/scenario-builder/rules.h
```diff
@@ -119,8 +119,10 @@ class Rules final: private Yuni::NonCopyable<Rules>
     //! Renewable (array [0..pAreaCount - 1])
     std::vector<renewableTSNumberData> renewable;
 
-    //! hydro levels
-    hydroLevelsData hydroLevels;
+    //! hydro initial levels
+    hydroLevelsData hydroInitialLevels = hydroLevelsData("hl,", initLevelApply);
+    //! hydro final levels
+    hydroLevelsData hydroFinalLevels = hydroLevelsData("hfl,", finalLevelApply);
```
flomnes:
```suggestion
    hydroLevelsData hydroInitialLevels{"hl,", initLevelApply};
    //! hydro final levels
    hydroLevelsData hydroFinalLevels{"hfl,", finalLevelApply};
```

src/antares-deps
flomnes, line 1: revert this change

src/libs/antares/study/parts/hydro/finalLevelValidator.cpp
```diff
@@ -0,0 +1,157 @@
...
+{
+    return ! isnan(finalLevel_);
+}
+
+bool FinalLevelValidator::compatibleWithReservoirProperties()
+{
+    if (hydro_.reservoirManagement && !hydro_.useWaterValue)
+        return true;
+
+    logs.info() << "Final reservoir level not applicable! Year:" << year_ + 1
```
flomnes:
```suggestion
    logs.warning() << "Final reservoir level not applicable! Year:" << year_ + 1
```

flomnes (changes requested):

src/libs/antares/study/include/antares/study/scenario-builder/rules.h
guilpier-code: Done

src/libs/antares/study/parts/hydro/finalLevelValidator.cpp
guilpier-code: Done

flomnes (approved):

PR #1522, opened 2023-08-08
Remove `goto` number 1

flomnes:


src/solver/optimisation/opt_optimisation_lineaire.cpp
```diff
@@ -66,35 +65,15 @@ double OPT_ObjectiveFunctionResult(const PROBLEME_HEBDO* Probleme,
         return Probleme->coutOptimalSolution2[NumeroDeLIntervalle];
 }
 
-bool OPT_OptimisationLineaire(const OptimizationOptions& options, PROBLEME_HEBDO* problemeHebdo, AdqPatchParams& adqPatchParams,
-                              Solver::IResultWriter& writer)
+static bool runWeeklyOptimization(const OptimizationOptions& options,
```
sylvlecl: We could move the new methods together in the anonymous namespace above

sylvlecl (approved):

src/solver/optimisation/opt_optimisation_lineaire.cpp
flomnes: done

PR #1527, opened 2023-08-09
Fix unfeasible problem analyzer

### LINKED ISSUE #1472: Infeasible problem analyzer is KO since v8.7.0-*
## What happened
The infeasible problem analyzer is a small library inside Antares Simulator that is used to point at inconsistent user input when an infeasible problem is encountered.

It works by a system of regex on constraint names.

Before named MPS, only a handful of constraints were named. Now, all of them are (or none, depending on option `--use-named-mps`).

This is a problem in both cases
- When constraints are named, we can no longer filter them so easily. Furthermore, their name has changed so `Constraint::getType()` (src/solver/infeasible-problem-analysis/constraint.cpp) is not expected to work as it did previously
- When constraints are not named, no constraint is added to the analyzer so it does nothing

### Current behavior with named MPS
```
[2023-07-31 15:11:47][solver][error] The following constraints are suspicious (first = most suspicious)
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] <unknown>
[2023-07-31 15:11:47][solver][error] Possible causes of infeasibility:
[2023-07-31 15:11:47][solver][error] * Negative hurdle costs on lines with infinite capacity (rare).
[2023-07-31 15:11:47][solver][infos] Solver MPS File: `problem-1-1--optim-nb-1.mps'
[2023-07-31 15:11:47][solver][error] Linear optimization failed
[2023-07-31 15:11:47][solver][infos] 
[2023-07-31 15:11:47][solver][fatal] Year 1 failed at week 1.
[2023-07-31 15:11:47][solver][fatal] Year 0 has failed in the previous set of parallel year.
[2023-07-31 15:11:47][solver][error] Aborting now. See logs for more details
```
All constraints are named. The analyzer finds some, but due to the change in naming convention is unable to parse constraint names.

### Current behavior without named MPS
```
[2023-07-31 15:45:18][solver][error] Cannot generate infeasibility report: no constraints have been selected
[2023-07-31 15:45:18][solver][infos] Solver MPS File: `problem-1-1--optim-nb-1.mps'
[2023-07-31 15:45:18][solver][error] Linear optimization failed
[2023-07-31 15:45:18][solver][infos] 
[2023-07-31 15:45:18][solver][fatal] Year 1 failed at week 1.
[2023-07-31 15:45:18][solver][fatal] Year 0 has failed in the previous set of parallel year.
[2023-07-31 15:45:18][solver][error] Aborting now. See logs for more details
omnesflo@gm0winl370:~/Téléchargements/StudyTest_28072023_150722$ 
```
No constraints are named, so no slack variable is added to the auxilliary LP problem.

## What could be done
- Name constraints all the time, provide constraint&variable names only when the user requests named MPS (currently : variables&constraints are named only when named MPS are requested) 
- Adapt `Constraint::getType()` to the new naming convention
- Check that all works

a-zakir:
fix infeasible problems analyzer 
close #1472


src/solver/utils/ortools_utils.cpp
```diff
@@ -9,6 +9,63 @@
...
+    return 'c';
+}
+
+void CheckName(std::string& var_name, unsigned idxVar, MATRIX_ELEMENT_TYPE type)
+{
+    if (var_name.empty())
+    {
+        var_name = FromMatrixElementTypeToChar(type) + std::to_string(idxVar);
+    }
+}
```
flomnes: Use C++ classes instead of `enum` and `if..else`

src/solver/utils/ortools_utils.cpp
```diff
@@ -9,6 +9,63 @@
...
+                              const double* bMax,
+                              const double* costs,
+                              int nbVar)
+{
+    MPObjective* const objective = solver->MutableObjective();
+    for (int idxVar = 0; idxVar < nbVar; ++idxVar)
+    {
+        std::string varName;
+        UpdateCoefficients(
+          solver, objective, bMin, bMax, costs, idxVar, varName, MATRIX_ELEMENT_TYPE::VARIABLE);
```
flomnes: You may wrap this behavior into a namer, taking an `std::vector` as ctor argument.

```cpp
class VariableNamer {
VariableNamer(const std::vector<string>& names);
std::string getName(unsigned index);
};

static void transferVariables(MPSolver* solver,
                              const double* bMin,
                              const double* bMax,
                              const double* costs,
                              int nbVar
                              VariableNamer& namer);

```

Then you only need one version for `transferVariables` and `transferRows`, not 2.

src/solver/utils/ortools_utils.cpp
```diff
@@ -48,31 +116,17 @@ static void transferRows(MPSolver* solver,
...
-        }
-        else if (sens[idxRow] == '<')
-        {
-            bMax = rhs[idxRow];
-        }
-        else if (sens[idxRow] == '>')
-        {
-            bMin = rhs[idxRow];
-        }
+        std::string constraintName = NomDesContraintes[idxRow];
```
```suggestion
        const auto& constraintName = NomDesContraintes[idxRow];
```

src/solver/utils/ortools_utils.cpp
```diff
@@ -137,29 +191,45 @@ namespace Antares
...
-                      problemeSimplexe->NombreDeVariables,
-                      problemeSimplexe->NomDesVariables);
-
-    // Create constraints and set coefs
-    transferRows(solver,
-                 problemeSimplexe->SecondMembre,
-                 problemeSimplexe->Sens,
-                 problemeSimplexe->NombreDeContraintes,
-                 problemeSimplexe->NomDesContraintes);
+    if (problemeSimplexe->UseNamedProblems())
```
```suggestion
    Namer variableNamer(problemeSimplexe->VariableNames, problemeSimplexe->UseNamedProblems());
    Namer constraintNamer(problemeSimplexe->ConstraintNames, problemeSimplexe->UseNamedProblems());
    // Pass these namers to transferVariables and transferRows
```

flomnes (changes requested):

src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -38,6 +39,21 @@ double Constraint::getSlackValue() const
...
+   while (std::getline(tokenStream, token, delimiter))
+   {
+      tokens.push_back(token);
+   }
+   return tokens;
+}
+
+std::string StringBetweenAngleBrackets(const std::string& str){
+  return split(split(str, '<')[1], '>')[0];
+}
```
flomnes: There are multiple issues with this approach
1. This function is not testes
2. There is no error management (what if there are no "<" and ">" ?)
3. You could use iterators instead ; find "<", find ">" and return the string between these 2 iterators

a-zakir: 1. ok
2. ok
3. why not

src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -39,20 +41,45 @@ double Constraint::getSlackValue() const
...
+    {
+    }
+};
 
-std::string StringBetweenAngleBrackets(const std::string& str){
-  return split(split(str, '<')[1], '>')[0];
+std::string StringBetweenAngleBrackets(const std::string& str)
+{
+    const auto& str_begin_iterator = str.begin();
+    const auto& str_end_iterator = str.end();
```
flomnes: `str.begin` is actually shorter than `str_begin_iterator`, consider removing this alias reference, or using just `begin` & `end`

flomnes (changes requested):

flomnes (approved):

PR #1529, opened 2023-08-09
Move writer into libs

JasonMarechal25:


flomnes:
Please resolve conflicts

src/libs/antares/benchmarking/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+        PRIVATE
+        ${CMAKE_SOURCE_DIR}/ext/yuni/src
+        ${CMAKE_SOURCE_DIR}/libs
+        )
+
+target_link_libraries(benchmarking
+        PRIVATE
+        yuni-static-core
+        inifile
+        result_writer
```
sylvlecl: I cannot find where the writer is used in the source files of the lib ?

flomnes: To measure the time spent on writing zip files (formerly, to measure the time waiting for the mutex to be unlocked)

sylvlecl: I mean that in the benchmarking lib source files, we don't include the writer headers --> it seems to me that it's not used inside this lib ?

src/libs/antares/benchmarking/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+set(SRC_BENCHMARKING
```
sylvlecl, line 1: Note: we can probably also use this PR to remove the unused source file `file_writer.cpp` in the benchmarking directory.

src/libs/antares/benchmarking/CMakeLists.txt
sylvlecl: It builds without the dependency, on my side.

flomnes:
Note : there are many conflicts, please resolve them

src/libs/antares/benchmarking/CMakeLists.txt
flomnes: That's surprising !

JasonMarechal25:
> Note : there are many conflicts, please resolve them

No more

src/libs/antares/benchmarking/CMakeLists.txt
JasonMarechal25: Yep

src/libs/antares/benchmarking/CMakeLists.txt
JasonMarechal25: Dependency is more of a "header" deps through study.h

sylvlecl:
Closing this one in favor of https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1548, which is an improved version of this.

Also, this was not just anymore about moving the writer, it makes more sense to have a more global PR about build improvements.

PR #1536, opened 2023-08-10
Final removal of study singleton from the solver

sylvlecl:
**Description**

The PR moves the singleton from the antares core library to the UI, which is the only component that needs it.

This will discourage to reintroduce the use of the singleton outside of the UI.

**Note**

I stumbled upon a possible design issue:
if a `ISimulation` outlives its `Study`, it will crash at destruction time, because `HydroManagement` references the study in its destructor.


flomnes (commented):
Remove also `Study::Ptr` ?

sylvlecl:
> Remove also `Study::Ptr` ?

I would prefer to do such ownership management change in another PR, it's a slightly different concern than the singleton.

In the UI part, I would prefer not to change the current use of `shared_ptr` anyway, since we don't want to maintain it.

flomnes (approved):

sylvlecl:
**Analysis of sonarcloud feedback**
the duplication is only in UI parts, and has not been introduced by this PR which has only renamed a few functions in that part
--> we can ignore it

PR #1543, opened 2023-08-10
Use int32_t and uint32_t in PROBLEME_HEBDO

### LINKED ISSUE #1515: Use type `std::size_t` for `PROBLEME_HEBDO::NombreDe*` ?
Mixing signed and unsigned integers is generally not advised

Sonar sometimes complains in code such as this one
```cpp
struct PROBLEME_HEBDO {
    int NombreDePays = 0;
};

for (int areaIndex = 0; areaIndex < problemeHebdo->NombreDePays; areaIndex++) {
    var = Correspondance.Injection[areaIndex]; // implicit cast int -> unsigned long
}
```
I think we should do the following instead
```cpp
struct PROBLEME_HEBDO {
    std::size_t NombreDePays = 0;
};

for (std::size_t areaIndex = 0; areaIndex < problemeHebdo->NombreDePays; areaIndex++) {
    var = Correspondance.Injection[areaIndex]; // no cast
}
```

payetvin:
close #1515


flomnes (changes requested):
Please uint32_t for Nombre*

flomnes (approved):

PR #1545, opened 2023-08-11
Remove global variables

### LINKED ISSUE #1119: Remove global variables

payetvin:


flomnes (commented):
You may remove file **sim_structure_probleme_adequation.h**

JasonMarechal25 (approved):

src/solver/simulation/solver.h
```diff
@@ -183,6 +192,8 @@ class ISimulation : public Impl
     std::shared_ptr<Yuni::Job::QueueService> pQueueService = nullptr;
     //! Result writer
     Antares::Solver::IResultWriter::Ptr pResultWriter = nullptr;
+
+    RANDOM_ARRAYS randomArrays;
```
JasonMarechal25: I see this was introduced but it any usage in this PR. Did I miss something ?

flomnes: From my understanding, this is work in progress

payetvin: Needs to be used instead of the extern var, it's a WIP

src/solver/aleatoire/alea_fonctions.h
```diff
@@ -27,9 +27,11 @@
 #ifndef __SOLVER_ALEA_H__
 #define __SOLVER_ALEA_H__
 
+#include "../simulation/sim_structure_donnees.h"
```
sylvlecl, line 30: Let's not use relative includes

src/solver/aleatoire/alea_fonctions.h
```diff
@@ -27,9 +27,11 @@
...
 #include "antares/study/study.h"
 
-void ALEA_TirageAuSortChroniques(const Antares::Data::Study& study, double** thermalNoisesByArea, uint numSpace);
+void ALEA_TirageAuSortChroniques(const Antares::Data::Study& study, double** thermalNoisesByArea,
```
sylvlecl: it would be more clear to return the `VAL_GEN_PAR_PAYS`, I don't think we need to pass it by argument.
(in particular, let's not worry about performances thanks to copy ellision and move semantics)

Actually, the best thing would be to return only the values for the required "numspace", so only a std::vector<VALEURS_GENEREES_PAR_PAYS>.

sylvlecl: Also while we are here, we could make clear that the noises are a not modified input, by making them const (like `double const* const*` instead of `double**`)

sylvlecl: Ok, now I understood why it's not possible to simply return the result.
That method only modifies **part** of the underlying `VALEURS_GENEREES_PAR_PAYS`.

This should probably be improved, but probably not in the scope of that PR, so let's leave it this way.

In any case, the comments about passing only the data for the numspace, and about the noises are still valid!

src/solver/hydro/management/management.h
```diff
@@ -100,11 +102,12 @@ class HydroManagement final
     // \return The total inflow for the whole year
     double prepareMonthlyTargetGenerations(Data::Area& area, PerArea& data);
 
-    void prepareDailyOptimalGenerations(Solver::Variable::State& state, uint y, uint numSpace);
+    void prepareDailyOptimalGenerations(Solver::Variable::State& state, uint y, uint numSpace, VAL_GEN_PAR_PAYS& valeursGenereesParPays);
```
sylvlecl: We should pass only the data which correspond to this numspace

src/solver/simulation/simulation.h
```diff
@@ -55,7 +56,8 @@ void SIM_RenseignementProblemeHebdo(const Antares::Data::Study& study,
                                     PROBLEME_HEBDO& problem,
                                     uint weekInTheYear,
                                     uint numSpace,
-                                    const int);
+                                    const int,
+                                    VAL_GEN_PAR_PAYS&);
```
sylvlecl: At this stage, we don't modify anymore the generated values: it should be const here and everywhere "below".

src/solver/simulation/adequacy.h
```diff
@@ -93,7 +94,8 @@ class Adequacy
     void initializeState(Variable::State& state, uint numSpace);
 
 private:
-    bool simplexIsRequired(uint hourInTheYear, uint numSpace) const;
+    bool simplexIsRequired(uint hourInTheYear, uint numSpace,
+            VAL_GEN_PAR_PAYS& valeursGenereesParPays) const;
```
sylvlecl: Should be const

src/solver/simulation/adequacy.h
```diff
@@ -72,14 +72,15 @@ class Adequacy
...
+    bool simulationBegin(VAL_GEN_PAR_PAYS& valeursGenereesParPays);
 
     bool year(Progression::Task& progression,
               Variable::State& state,
               uint numSpace,
               yearRandomNumbers& randomForYear,
               std::list<uint>& failedWeekList,
-              bool isFirstPerformedYearOfSimulation);
+              bool isFirstPerformedYearOfSimulation,
+              VAL_GEN_PAR_PAYS& valeursGenereesParPays);
```
sylvlecl: Should be const

src/solver/simulation/economy.cpp
```diff
@@ -79,7 +79,7 @@ void Economy::initializeState(Variable::State& state, uint numSpace)
     state.numSpace = numSpace;
 }
 
-bool Economy::simulationBegin()
+bool Economy::simulationBegin(VAL_GEN_PAR_PAYS& valeursGenereesParPays)
```
sylvlecl: Should be const

src/solver/simulation/solver.h
```diff
@@ -183,6 +193,10 @@ class ISimulation : public Impl
     std::shared_ptr<Yuni::Job::QueueService> pQueueService = nullptr;
     //! Result writer
     Antares::Solver::IResultWriter::Ptr pResultWriter = nullptr;
+
+    VAL_GEN_PAR_PAYS valeursGenereesParPays;
+
+    RANDOM_ARRAYS randomArrays;
```
sylvlecl: If we don't use it in that PR, it's probably better to remove it for now

src/solver/simulation/solver.hxx
```diff
@@ -1609,6 +1613,34 @@ void ISimulation<Impl>::loopThroughYears(uint firstYear,
     }
 }
 
+template<class Impl>
+void ISimulation<Impl>::allocateValeursGenereesParPays()
```
sylvlecl: I would propose that this method returns the objet instead of initializing it in place, so that it's clear that it's created at this point.

src/solver/simulation/sim_structure_donnees.h
```diff
@@ -64,6 +64,8 @@ typedef struct
     double* ParLigne;
 } MATRICE_2D;
 
+typedef std::vector<std::vector<VALEURS_GENEREES_PAR_PAYS>> VAL_GEN_PAR_PAYS;
```
sylvlecl: Possibly for a future PR but:
the best thing would be to have a small class to make the indices more clear. Here we don't really know what is the index in the first vector, and in the second vector.

We could define something like
```cpp
// All generated values for one space
class SpaceGeneratedValues {
public:
    VALEURS_GENEREES_PAR_PAYS& areaValues(int areaIndex) {
        return values[areaIndex];
    }
    // const overload

private:
   std::vector<VALEURS_GENEREES_PAR_PAYS> values;
};

// All generated values for all spaces, replaces the typedef
class GeneratedValues {
public:
    SpaceGeneratedValues& spaceValues(int spaceIndex) {
        return values[areaIndex];
    }
    // const overload
private:
   std::vector<SpaceGeneratedValues> values;
};

```

sylvlecl (changes requested):
Thanks:
- we need to improve the const management (basically, the values must not be modified after creation, they must be passed as const down to the computation)
- I suggest to improve a little the struct which contains generated values

src/solver/aleatoire/alea_fonctions.h
payetvin:
```
target_include_directories(libantares-solver-aleatoire
		PRIVATE
		${CMAKE_SOURCE_DIR}/solver)
```
Tried this without results

src/solver/simulation/sim_structure_donnees.h
flomnes:
```suggestion
using VAL_GEN_PAR_PAYS = std::vector<std::vector<VALEURS_GENEREES_PAR_PAYS>>;
```

src/solver/simulation/solver.h
```diff
@@ -156,6 +157,8 @@ class ISimulation : public Impl
     */
     void loopThroughYears(uint firstYear, uint endYear, std::vector<Variable::State>& state);
 
+    VAL_GEN_PAR_PAYS allocateValeursGenereesParPays();
+
```
flomnes: This definition is in a templated class, so it will be templated as well. Does it really need to be ?

Also, I'd rather see a function taking `valeursGenereesParPays` as an argument rather than returning it.

src/solver/simulation/solver.hxx
```diff
@@ -1609,6 +1613,36 @@ void ISimulation<Impl>::loopThroughYears(uint firstYear,
     }
 }
 
+template<class Impl>
+VAL_GEN_PAR_PAYS ISimulation<Impl>::allocateValeursGenereesParPays()
+{
+    VAL_GEN_PAR_PAYS val;
```
flomnes:
```suggestion
void ISimulation<Impl>::allocateValeursGenereesParPays(VAL_GEN_PAR_PAYS& val)
{
```

src/solver/hydro/management/management.h
payetvin: numspace is also used for other arrays like pAreas so we can't reduce number of args.
It makes more sense to keep the same reference through the calling stack and use the part we need in the final function to keep the same behavior. It's a reference so it doesn't cost anything to pass

src/solver/simulation/sim_structure_donnees.h
flomnes: I agree, with vectors of vectors it's not clear what index is for what : area ? numSpace ?

flomnes (changes requested):

src/solver/aleatoire/alea_fonctions.h
sylvlecl: Ok, it seems it's because it's included from file src/libs/antares/study/estimate.memory-footprint.cpp:35, which has a (cyclic) dependency on this, and which does not have this include directory.

This will be removed anyway by another PR, let's not worry.

src/solver/hydro/management/management.h
sylvlecl: We could still go in the right direction by passing only the necessary data, although numspace is needed for other data.

But ok, let's tackle it later with a more targeted evolution about spaces (https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/1571).

sylvlecl (approved):

src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
```diff
@@ -42,8 +42,10 @@ AdequacyPatchOptimization::AdequacyPatchOptimization(const Antares::Data::Study&
                                                      PROBLEME_HEBDO* problemeHebdo,
                                                      AdqPatchParams& adqPatchParams,
                                                      uint thread_number,
-                                                     IResultWriter& writer) :
-  WeeklyOptimization(options, problemeHebdo, adqPatchParams, thread_number, writer), study_(study)
+                                                     IResultWriter& writer,
+                                                     const VAL_GEN_PAR_PAYS& valeursGenereesParPays) :
+  WeeklyOptimization(options, problemeHebdo, adqPatchParams, thread_number, writer), study_(study),
+    valeursGenereesParPays_(&valeursGenereesParPays)
```
flomnes: Use a reference instead

flomnes (changes requested):
Final remark, I hope

flomnes (approved):

PR #1546, opened 2023-08-11
First step for yuni string removal: tools and application to cluster name

sylvlecl:
**Description**

The first goal is to identify what it takes to incrementally replace yuni strings in the solver.

As a proof of concept, in this PR we try to use `std::string` for **cluster names**.

The APIs of yuni and std strings differ by a few things:
 - yuni has case conversion methods
 - yuni has implicit conversion to boolean
 - yuni has a UTF8 iterator, to iterate not on `char` but on actual code point
 - yuni has insertion operator `<<`
 - yuni has some string algorithms as class methods `trim`, `replace` ...

The UI seems to make quite a lot of string manipulation using those.
However our goal is not to replace YStrings there, so a simple conversion will do, when necessary.

**Proposed elements to make the migration easier**
- implement implicit conversion to and from `std::string` in yuni, so that we can replace yuni strings bit by bit without having to make explicit conversions each time
- wrap the very few methods that use the UTF8 iterator with `std::string` overloads. In the final step we will get rid of those methods in one way or another
 - add some unit tests to make the expected behaviour of those methods explicit


sylvlecl:
As suggested by @flomnes , since string manipulations are mainly in the UI part which we don't want to maintain, we need to convert from `std::string` to `YString` in the UI part to avoid changing the code there.

sylvlecl:
It seems to work quite well without too much effort on that example (cluster names), once the prerequisite of having the implicit conversion is here.

I think we could go on with this approach by replacing with std::string part by part, in order to avoid conflicts and too large PRs.

sylvlecl:
So finally, what is implemented in the PR is:

1. Implicit conversion between `std::string` and `Yuni::CString`
2. Overloads of some string functions (`BeautifyName` and `TransformNameToId`) to work with both kinds of strings
3. As a first set of application: changing cluster names and file extensions types to std::string

We can choose to have a separate PR for 1 and 2 if we consider better to perform all actual changes in later PRs.
Otherwise we can keep 3 as a first step, and go on with the rest of the data model in subsequent PRs.

src/libs/antares/utils.cpp
```diff
@@ -68,4 +72,41 @@ void BeautifyName(YString& out, AnyString oldname)
...
+}
+
+void BeautifyName(std::string& out, const std::string& oldname)
+{
+    YString yuniOut;
+    BeautifyName(yuniOut, oldname);
+    out = yuniOut.c_str();
+}
+
+char charToLower(char in)
```
flomnes: I may be wrong but it seems that this function should be in the anon namespace

flomnes (commented):
Please place comments when you notice a function or variable is only used in the GUI, so we can remove it later

src/tests/src/libs/antares/test_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <antares/utils.h>
+
+
+namespace {
+
+std::string beautifyStd(const std::string& in) {
+    std::string out;
+    Antares::BeautifyName(out, in);
+    return out;
+}
```
flomnes:
```suggestion
const auto beautifyStd = Antares::BeautifyName<std::string>;
```

src/tests/src/libs/antares/test_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+Yuni::String beautifyYuni(const Yuni::String in) {
+    Yuni::String out;
+    Antares::BeautifyName(out, in);
+    return out;
+}
```
flomnes:
```suggestion
const auto beautifyYuni = Antares::BeautifyName<Yuni::String>;
```

src/ext/yuni/src/yuni/cmake/core/check-includes.cmake
```diff
@@ -47,6 +47,9 @@ check_include_file_cxx(iostream YUNI_HAS_IOSTREAM)
 # cassert
 check_include_file_cxx(cassert YUNI_HAS_CASSERT)
 
+# cassert
+check_include_file_cxx(string YUNI_HAS_STD_STRING)
+
```
flomnes: One can always assume that header <string> exists, please remove this.

flomnes (changes requested):

src/libs/antares/utils.cpp
sylvlecl: Indeed: it has now been removed in the "boost version" anyway.

src/tests/src/libs/antares/test_utils.cpp
sylvlecl: Those 2 methods served mainly to return the result instead of passing it as parameter: however I indeed templatized them to avoid the duplication, thanks!

src/ext/yuni/src/yuni/cmake/core/check-includes.cmake
sylvlecl: Indeed, I removed this unnecessary complexity.

sylvlecl:
Slightly improved version after review + using boost header library for string algorithms.

flomnes (approved):

flomnes:
Please resolve conflicts with  src/libs/antares/study/parts/thermal/cluster.cpp

a-zakir:
> So finally, what is implemented in the PR is:
> 
> 1. Implicit conversion between `std::string` and `Yuni::CString`
> 2. Overloads of some string functions (`BeautifyName` and `TransformNameToId`) to work with both kinds of strings
> 3. As a first set of application: changing cluster names and file extensions types to std::string
> 
> We can choose to have a separate PR for 1 and 2 if we consider better to perform all actual changes in later PRs. Otherwise we can keep 3 as a first step, and go on with the rest of the data model in subsequent PRs.

This is ugly

a-zakir:
Very ugly

PR #1547, opened 2023-08-14
set deps 2.0.7

JasonMarechal25:


a-zakir (approved):

flomnes (changes requested):
See #1550

Build is successful here since minizip-ng from antares-deps is just ignored by Antares Simulator.

We don't want that. The problem is that there is a symbol conflict between zlib1.dll (defined somewhere else) and zlibstatic.lib from minizip-ng.

PR #1548, opened 2023-08-14
Continue cmake improvement

JasonMarechal25:
- Add "antares" include level when missing
- Fix missing files in "source group" for winfows
- Split some directory in libs


sylvlecl:
**Additional information**

This work creates the following libs:
- writer lib
- benchmarking lib

Adds separate CMakeLists.txt for existing targets:
- date
- exceptions
- checks

Renames some targets to less verbose targets: libantares-core-calendar --> date, etc

Defines PUBLIC include directory for most libraries, so that it comes automatically when using `target_link_libraries`, among which:
- array
- io
- inifile 
- ...

For those libraries, a dedicated public headers directory has been created (prefixed by subdirectory `antares/<lib>`).

src/libs/antares/benchmarking/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+        )
+source_group("misc\\benchmarking" FILES ${SRC_BENCHMARKING})
+
+add_library(benchmarking
+        ${SRC_BENCHMARKING}
+        )
+
+target_include_directories(benchmarking
+        PUBLIC
+        ${CMAKE_SOURCE_DIR}/libs #study
```
sylvlecl, line 17: Same comment about cyclic dependency to core, which would avoid all the includes below.

src/libs/antares/correlation/CMakeLists.txt
```diff
@@ -10,11 +10,16 @@ add_library(correlation ${SRC_CORRELATION})
 target_link_libraries(correlation
         PRIVATE
         yuni-static-core
-        result_writer
+        inifile
+        array
+        PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}
```
sylvlecl: to be removed, its place is in includes

src/libs/antares/correlation/CMakeLists.txt
```diff
@@ -10,11 +10,16 @@ add_library(correlation ${SRC_CORRELATION})
...
         )
 
 target_include_directories(correlation
         PRIVATE
-        ${CMAKE_SOURCE_DIR}/libs
         ${CMAKE_SOURCE_DIR}/libs/solver
```
sylvlecl, line 19: This does not exist

src/libs/antares/checks/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+)
+source_group("checks" FILES ${SRC_CHECKS})
+
+add_library(checks
+        ${SRC_CHECKS}
+)
+
+target_include_directories(checks
+        PRIVATE
+        ${CMAKE_SOURCE_DIR}/libs/
```
sylvlecl, line 13: Basically, we need all those include directories as soon as we include study.h, from what I understand.

This is actually a cyclic dependency on "core" lib. Because we don't make it explicit for cmake, we need to include "by hand" all the necessary include directories, instead of letting cmake do it.

I would propose to make it explicit since it's the current situation, by including libantares-core in target_link_libraries.

This will avoid all the include directories that are added after, and make explicit that cyclic dependency.

In the future, we should probably target to remove those cyclic dependencies.

sylvlecl: Probably for a future PR though ...

src/libs/antares/writer/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+project(result-writer)
+add_library(result_writer
+        # Interface class (public API)
+        antares/writer/i_writer.h
+        antares/writer/result_format.h
+
+        # Helper class
+        private/ensure_queue_started.h
```
sylvlecl, line 8: We don't need the "private" directory, because the public headers are anyway already separated in their own directory.

src/tests/src/libs/antares/CMakeLists.txt
```diff
@@ -44,6 +48,9 @@ add_executable(tests-matrix-save ${SRC_TEST_MATRIX_SAVE})
 target_include_directories(tests-matrix-save
 								PRIVATE
 									"${src_libs_antares}/array"
+									"${src_libs_antares}/io"
```
sylvlecl, line 51: Can we not just link with those libs instead?

src/libs/antares/correlation/CMakeLists.txt
sylvlecl: Fixed it

sylvlecl (approved):

sylvlecl:
@JasonMarechal25

I will leave the comments for future work in other PRs, as discussed I think it's better to first merge this large work and then refine in smaller, more targetted PRs.

src/libs/antares/writer/CMakeLists.txt
JasonMarechal25: Public headers are in their own directory but target_include_directories specifies the top "writer" directory as public, making all others headers (and sources ) visible. I added "private" to prevent unwanted inclusion.

Will be solved in the future by making a "include" intermediate directory

PR #1553, opened 2023-08-16
Clarify string -> enum convesion for thermal cluster groups

flomnes:


src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -235,6 +234,29 @@ void Data::ThermalCluster::copyFrom(const ThermalCluster& cluster)
         parentArea->forceReload();
 }
 
+static Data::ThermalCluster::ThermalDispatchableGroup stringToGroup(const Data::ClusterName& newgrp)
+{
+    using namespace Antares::Data;
+    const static std::map<ClusterName, ThermalCluster::ThermalDispatchableGroup> mapping
```
JasonMarechal25, line 240: Maybe use static constexpr if possible

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -235,6 +234,29 @@ void Data::ThermalCluster::copyFrom(const ThermalCluster& cluster)
...
+         {"hard coal", ThermalCluster::thermalDispatchGrpHardCoal},
+         {"gas", ThermalCluster::thermalDispatchGrpGas},
+         {"oil", ThermalCluster::thermalDispatchGrpOil},
+         {"mixed fuel", ThermalCluster::thermalDispatchGrpMixedFuel},
+         {"other", ThermalCluster::thermalDispatchGrpOther1},
+         {"other 1", ThermalCluster::thermalDispatchGrpOther1},
+         {"other 2", ThermalCluster::thermalDispatchGrpOther2},
+         {"other 3", ThermalCluster::thermalDispatchGrpOther3},
+         {"other 4", ThermalCluster::thermalDispatchGrpOther4}};
+    if (mapping.count(newgrp) > 0)
```
sylvlecl: Of course not a bottleneck as a whole, but this will go over all the collection to count, and we will perform another search below: it would be fore efficient to just search once with a `find`.

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -245,86 +267,7 @@ void Data::ThermalCluster::setGroup(Data::ClusterName newgrp)
     }
     pGroup = newgrp;
     newgrp.toLower();
```
sylvlecl: This would probably find a better place inside the `stringToGroup` method since it seems to be used only for this search purpose.

sylvlecl (commented):
Minor suggestions for more improvement

src/libs/antares/study/parts/thermal/cluster.cpp
JasonMarechal25: Disregard comment, std::map is not a litteral type and can't be constxpre

sylvlecl (approved):

Issue #1552, opened 2023-08-16
Class Area : remove data member "enabled"


guilpier-code:
Class **Area** data member **enabled** does not seem of great use.
By the way it is initialized only in one of three **Area**'s constructor, which is strange and could lead to undefined behavior.

flomnes:
And will lead to undefined behavior ;)

PR #1556, opened 2023-08-16
Remove useless Area's data member : enabled

guilpier-code:
Fixes issue #1552


sylvlecl (approved):

flomnes:
@guilpier-code told me that this attribute was used in the TS generator, how about it ?

guilpier-code:
It's not the case actually

PR #1558, opened 2023-08-17
CMake Memory and Sys

JasonMarechal25:


src/libs/antares/array/CMakeLists.txt
```diff
@@ -14,11 +14,13 @@ target_link_libraries(array
         PUBLIC
         io #matrix.hxx
         jit #jit.hxx require logs
+        Antares::memory
         )
 
 target_include_directories(array
         PRIVATE
         ${CMAKE_SOURCE_DIR}/libs #memory usage
+        ${CMAKE_SOURCE_DIR}/libs/memory/include #memory usage -> memory
```
sylvlecl: This dir does not exist actually :)
But indeed it's not needed, it comes with the target_link_libraries above

src/libs/antares/array/matrix.cpp
```diff
@@ -27,6 +27,7 @@
 
 #include <cstdlib>
 #include "antares/study/memory-usage.h"
+#include "antares/jit/antares/jit/jit.h"
```
sylvlecl: we should use the simple form "antares/jit/jit.h" and target_link_library with jit, which already defines its include dir.

src/libs/antares/memory/memory.cpp
```diff
@@ -25,29 +25,27 @@
...
-#include <signal.h>
+#include <csignal>
 #endif
 #ifdef YUNI_OS_WINDOWS
 #include <WinIoCtl.h>
 #endif
 #include "../study/memory-usage.h"
-#include "../sys/policy.h"
+#include <antares/sys/policy.h>
+#include "antares/memory/include/antares/memory/memory.h"
```
sylvlecl: let's remove the first part

src/libs/antares/memory/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+        PRIVATE
+        yuni-static-core
+        logs
+        Antares::sys
+)
+
+target_include_directories(${PROJ}
+        PRIVATE
+        ${CMAKE_SOURCE_DIR}/libs
+        ${CMAKE_SOURCE_DIR}/libs/antares/jit #memory-usage -> matrix -> jit
```
sylvlecl: I think we should completely break the dependency to study by moving the method 
`Memory::EstimateMemoryUsage` to memory-usage.h in study folder.

I think this will remove the cyclic dependency to study, which causes all those issues.

Otherwise again we can wonder if it's worth it to separate this lib :(

src/libs/antares/study/memory-usage.h
```diff
@@ -29,7 +29,7 @@
 
 #include <yuni/yuni.h>
 #include "fwd.h"
-#include <antares/array/matrix.h>
+#include <antares/array/antares/array/matrix.h> //TODO
```
sylvlecl: I guess it should be OK to remove the first part now ?

src/libs/antares/sys/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+        PRIVATE
+        yuni-static-core
+        Antares::logs
+        Antares::inifile
+)
+
+target_include_directories(${PROJ}
+        PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}/include
+        ${CMAKE_SOURCE_DIR}/libs
```
sylvlecl, line 29: this should not be needed: the public headers of this library do not include headers from other libs

sylvlecl (changes requested):

src/libs/antares/study/memory-usage.h
JasonMarechal25: Deprecated with memory usage removed

src/libs/antares/sys/CMakeLists.txt
JasonMarechal25: Indeed it only privately require fatal error

src/libs/antares/correlation/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ target_link_libraries(correlation
         yuni-static-core
         inifile
         array
+        PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}
```
sylvlecl: to be removed

src/libs/antares/memory/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+        memory.cpp)
+source_group("memory" FILES ${SRC_MEMORY})
+
+add_library(${PROJ} ${SRC_MEMORY})
+add_library(Antares::${PROJ} ALIAS ${PROJ})
+
+target_link_libraries(${PROJ}
+        PRIVATE
+        yuni-static-core
+        logs
```
sylvlecl, line 18: Headers from yuni and logs appear in public headers of this lib
--> it means that we must make those dependencies PUBLIC, not PRIVATE, otherwise transitive dependency inclusion will not work.

src/libs/antares/sys/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+        appdata.cpp
+)
+source_group("memory" FILES ${SRC_PROJ})
+
+add_library(${PROJ} ${SRC_PROJ})
+add_library(Antares::${PROJ} ALIAS ${PROJ})
+
+target_link_libraries(${PROJ}
+        PRIVATE
+        yuni-static-core
```
sylvlecl, line 19: yuni must be PUBLIC because the public headers of sys include some headers of yuni

sylvlecl (changes requested):
I think we need to make a few dependencies PUBLIC because of headers inclusions

src/libs/antares/correlation/CMakeLists.txt
JasonMarechal25: It is necessary to make headers public.
I will update with a "include" folder

JasonMarechal25: My bad, mixed up include and link

sylvlecl (approved):

PR #1560, opened 2023-08-17
Fix boost tests

sylvlecl:
**Description**

On oracle linux 8 container, Boost.Test comes only as a shared library.
Our current code relies on the presence of the static library instead, to generate the test executables.

To make it work with this environment, we are left with 2 options:
 - use the header-only variant to generate the executable
 - use the shared-library to run the executable

Here, I am trying the second option, in order to decrease a little build time and binaries disk usage.

### DIFF
--- a/.github/workflows/oracle8.yml
+++ b/.github/workflows/oracle8.yml
@@ -6,7 +6,6 @@ on:
   push:
     branches:
       - develop
-      - fix/*
   schedule:
     - cron: '21 2 * * *'
 
--- a/.github/workflows/oracle8.yml
+++ b/.github/workflows/oracle8.yml
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - develop
+      - fix/*
   schedule:
     - cron: '21 2 * * *'
 
--- a/.github/workflows/oracle8.yml
+++ b/.github/workflows/oracle8.yml
@@ -46,6 +46,11 @@ jobs:
            git config --global safe.directory '*'
            git submodule update --init --remote src/antares-deps src/tests/resources/Antares_Simulator_Tests
 
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        key: 'oraclelinux-8'
+
     - name: Install dependencies
       run: |
           pip3 install -r src/tests/examples/requirements.txt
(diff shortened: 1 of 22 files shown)


flomnes:
Executables indeed become dynamically linked with boost
```
$ ldd ./tests/src/solver/simulation/test-time_series
	linux-vdso.so.1 (0x00007ffd3515f000)
	libboost_unit_test_framework.so.1.71.0 => /lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.71.0 (0x00007fba8d40d000)
	libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1 (0x00007fba8d404000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fba8d3e8000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fba8d3c5000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fba8d1e3000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fba8d094000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fba8d077000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fba8ce85000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fba8d77b000)
```

flomnes:
Please investigate the following compiler warnings
```
/home/omnesflo/Antares_Simulator/src/tests/end-to-end/binding_constraints/test_binding_constraints.cpp:2: warning: "BOOST_TEST_DYN_LINK" redefined
    2 | #define BOOST_TEST_DYN_LINK
      | 
<command-line>: note: this is the location of the previous definition
/home/omnesflo/Antares_Simulator/src/tests/end-to-end/simple_study/simple-study.cpp:2: warning: "BOOST_TEST_DYN_LINK" redefined
    2 | #define BOOST_TEST_DYN_LINK
      | 
<command-line>: note: this is the location of the previous definition
```

sylvlecl:
> Please investigate the following compiler warnings

Indeed, it was due to a prior definition of those symbols in cmake files (which was present before that PR).
I removed them to use the same mechanism everywhere, with definition of the symbol in source files.

flomnes (approved):

PR #1561, opened 2023-08-17
Remove sys as direct dependency of antares-core

JasonMarechal25:


flomnes (commented):
While I agree with the general idea, I have 2 minor remarks
- Why define aliases ? Why not just name the original library antares::${PROJ} ?
- I prefer lower-case for CMake targets, so `antares::sys`, `antares::date`, etc.

Also, do we comply with the recent CMake ACR, having separate public & private headers ?

sylvlecl (approved):

flomnes (approved):

PR #1565, opened 2023-08-18
Cleanup date dependency

JasonMarechal25:


a-zakir (approved):

src/libs/antares/study/study.h
```diff
@@ -40,7 +40,7 @@
 #include "fwd.h"
 
 #include "simulation.h"
-#include "parameters.h"
```
sylvlecl: Parameters class is used in the study definition:
I don't think we should remove it, otherwise this header is not self contained anymore (does not compile by itself).

src/libs/antares/CMakeLists.txt
```diff
@@ -518,6 +517,8 @@ target_link_libraries(libantares-core
 		correlation
 		object
         benchmarking
+	PRIVATE
+		Antares::date
```
sylvlecl, line 517: I don't think it should be private:
some date classes are used in Study class, so any user of the lib will also need the date headers.

sylvlecl (changes requested):
Too much cleaning I think !

flomnes (approved):

sylvlecl (approved):

PR #1567, opened 2023-08-18
Prevent segfault during simulation, check bounds of scenario builder …

### LINKED ISSUE #1566: antares-solver crashes if the scenario-builder mentions out-of-bound columns for BC
## Problem
If a the scenario-builder (settings/scenariobuilder.dat) specifies too many columns for a group of binding constraints, antares-solver (8.7.0-rc2) crashes with an "out-of-bounds" error.

## Example
[Example study](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/12377407/BC-rhs-scenario--simple-BCs.zip)

```ini
[Default Ruleset]
bc,limit-ntc,0 = 1
bc,limit-ntc,1 = 3
```
A binding constraint in group **limit-ntc** has only 2 columns, the simulation crashes at year 2 with the following message (debug only)

```
[2023-08-18 10:40:11][solver][infos] parallel batch size : 1 (1 perfomed)
[2023-08-18 10:40:11][solver][infos] Year(s) 2 
antares: running "BC-rhs-scenario--simple-BCs": /home/omnesflo/Antares_Simulator/src/libs/antares/study/binding_constraint/../area/../../array/matrix.hxx:1665: Antares::Matrix<T, ReadWriteT>::ColumnType& Antares::Matrix<T, ReadWriteT>::operator[](uint) [with T = double; ReadWriteT = double; Antares::Matrix<T, ReadWriteT>::ColumnType = double*; uint = unsigned int]: Assertion `column < width' failed.
Abandon (core dumped)
```

## Correction
Scenario builder data should be checked against the number of columns. The existing behavior for the load (and probably thermal, wind, solar, etc.) is best described by this warning message
```
[2023-08-18 10:47:28][solver][warns] scenario-builder: Load: Area 'Area': value out of bounds for the year 2
```
meaning that out-of-bounds values are ignored, random draws are used instead, just like if no value was provided by the user. Here is a simplified excerpt of the code that issues this warning.

```cpp
if (!CheckValidity(tsNum, data, tsGenMax))
{
      logs.warning() << "scenario-builder: " << logprefix
                     << "value out of bounds for the year " << (y + 1);

    ret = false;
    continue;
}
// Ok, assign. The value provided by the interface is user-friendly
// and starts from 1.
target[y] = tsNum;
```

JasonMarechal25:
…for BC early

### DIFF
--- a/src/libs/antares/study/scenario-builder/applyToMatrix.hxx
+++ b/src/libs/antares/study/scenario-builder/applyToMatrix.hxx
@@ -40,11 +40,9 @@ inline bool CheckValidity<Data::AreaLink>(uint value,
 }
 
 template<>
-inline bool CheckValidity<BindingConstraintGroup>(uint, const BindingConstraintGroup&, uint)
+inline bool CheckValidity<BindingConstraintGroup>(uint value, const BindingConstraintGroup& group, uint)
 {
-    //TS-Generator never used
-    //Should check for time-series width, but we are missing information at this point
-    return true;
+    return value < group.numberOfTimeseries();
 }
 
 template<class StringT, class D>
--- a/src/libs/antares/study/scenario-builder/BindingConstraintsTSNumbersData.cpp
+++ b/src/libs/antares/study/scenario-builder/BindingConstraintsTSNumbersData.cpp
@@ -38,17 +38,18 @@ namespace Antares::Data::ScenarioBuilder
 bool BindingConstraintsTSNumberData::apply(Study& study)
 {
     return std::all_of(rules_.begin(), rules_.end(), [&study, this](const std::pair<std::string, MatrixType>& args){
-        const auto& [group_name, ts_numbers] = args;
-        auto group = study.bindingConstraintsGroups[group_name];
+        const auto& [groupName, tsNumbers] = args;
+        auto group = study.bindingConstraintsGroups[groupName];
         if (group == nullptr) {
-            logs.error("Group with name" + group_name + " does not exists");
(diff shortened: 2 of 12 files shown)


src/libs/antares/study/scenario-builder/applyToMatrix.hxx
```diff
@@ -40,11 +40,9 @@ inline bool CheckValidity<Data::AreaLink>(uint value,
 }
 
 template<>
-inline bool CheckValidity<BindingConstraintGroup>(uint, const BindingConstraintGroup&, uint)
+inline bool CheckValidity<BindingConstraintGroup>(uint value, const BindingConstraintGroup& group, uint)
```
flomnes, line 43: Looks good. Please make sure to include a warning like below

A polite way to tell the user they provided inconsistent TS numbers, and therefore their input is ignored.

```
                        logs.warning() << "scenario-builder: " << logprefix
<< "value out of bounds for the year " << (y + 1);
```

In both these warnings (new & existing), you could make it clear that the values have been discarded since it's not very clear currently.

flomnes (changes requested):
Please add a log message

src/libs/antares/study/scenario-builder/applyToMatrix.hxx
flomnes: I beg your pardon, this log message is displayed by the caller, templated function `ApplyToMatrix`

flomnes (approved):

flomnes (approved):

PR #1582, opened 2023-08-23
Split benchmarking

JasonMarechal25:
Ease dependency to study by having two modules.
One for benchmarking core functionnalities
One for study/simulation benchmarking


src/libs/antares/benchmarking/DurationCollector.cpp
```diff
@@ -0,0 +1,24 @@
...
+    const std::lock_guard<std::mutex> lock(mutex_);
+    duration_items_[name].push_back(duration);
+}
+
+void DurationCollector::toFileContent(FileContent& file_content)
+{
+    for (const auto& element : duration_items_)
+    {
+        const std::string& name = element.first;
+        const auto& durations = element.second;
```
flomnes: Not really the goal of this PR, but
```suggestion
    for (const auto& [name, durations] : duration_items_)
    {
```

src/libs/antares/InfoCollection/StudyInfoCollector.cpp
```diff
@@ -1,13 +1,15 @@
...
-#include <antares/Enum.hpp>
-#include <antares/study/area.h>
+//
+// Created by marechaljas on 22/08/23.
+//
+
+#include "antares/infoCollection/StudyInfoCollector.h"
+#include "antares/benchmarking/DurationCollector.h"
+#include "antares/benchmarking/file_content.h"
+#include "../../../config.h"
```
flomnes, line 8: Relative include

flomnes (changes requested):

flomnes (changes requested):
Please also resolve conflicts

src/libs/antares/InfoCollection/StudyInfoCollector.cpp
JasonMarechal25: Yes. config.h is not part of any library so we can't use chevron (<,>) to include it.
We could do so by adding a traget_include_directory instruction but it would open the way to include anything else.

flomnes (approved):

PR #1583, opened 2023-08-23
Split study in CMakefile

JasonMarechal25:
-  Move study in a CMakeFile
- Remove dependency to Study in date


.github/workflows/windows-vcpkg.yml
```diff
@@ -118,7 +118,7 @@ jobs:
     - name: Build
       shell: bash
       run: |
-           cmake --build _build --config release -j2
+           cmake --build _build --config release
```
flomnes: Why remove `-j2` ?

src/libs/antares/checks/CMakeLists.txt
```diff
@@ -7,18 +7,19 @@ source_group("checks" FILES ${SRC_CHECKS})
...
-        ${CMAKE_SOURCE_DIR}/libs/antares/memory/include #area -> matrix -> memory
+#        ${CMAKE_SOURCE_DIR}/libs/
+#        ${CMAKE_SOURCE_DIR}/libs/antares/study
+#        ${CMAKE_SOURCE_DIR}/libs/antares/inifile #adq-patch-params
+#        ${CMAKE_SOURCE_DIR}/libs/antares/array #area
+#        ${CMAKE_SOURCE_DIR}/libs/antares/io #area -> matrix -> io
+#        ${CMAKE_SOURCE_DIR}/libs/antares/date #area -> parameter -> date
+#        ${CMAKE_SOURCE_DIR}/libs/antares/writer #area -> antares/parts/thermal/cluster -> writer
+#        ${CMAKE_SOURCE_DIR}/libs/antares/jit #area -> matrix -> jit
+#        ${CMAKE_SOURCE_DIR}/libs/antares/memory/include #area -> matrix -> memory
```
flomnes: Please remove those comments

src/libs/antares/date/date.cpp
```diff
@@ -397,34 +395,16 @@ bool Calendar::saveToCSVFile(const AnyString& filename) const
...
-    // We do not retrieve directly the `leapyear` parameters
-    // A simulation should be made in ignoring this parameter (aka false)
-    // but the outputs should rely on it (for printing).
-    // It goes the same for the GUI : since it is _merely_ printing,
-    // it should be taken into consideration.
-    // Consequently, we will let the calling code specifying this value
-    settings.leapYear = leapyear;
-
-    // re-initialize the calendar with the new settings
+    settings_ = settings;
```
flomnes: - Pass `settings` by rvalue-reference ?
- Class names come with a capital, so that would be `Settings`

src/libs/antares/study/CMakeLists.txt
```diff
@@ -0,0 +1,447 @@
...
+        Antares::mersenne
+        Antares::result_writer #study.h
+        Antares::object
+    PRIVATE
+        Antares::exception
+)
+
+target_include_directories(study
+    PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}/../..
```
flomnes: IMO this should avoided just like `#include "../../header.hpp"`

src/libs/antares/study/load.cpp
```diff
@@ -143,11 +143,11 @@ void Study::parameterFiller(const StudyLoadOptions& options)
...
     // calendar update
     if (usedByTheSolver)
-        calendar.reset(parameters, /*force leapyear:*/ false);
+        calendar.reset({parameters.dayOfThe1stJanuary, parameters.firstWeekday, parameters.firstMonthInYear, false});
     else
-        calendar.reset(parameters);
+        calendar.reset({parameters.dayOfThe1stJanuary, parameters.firstWeekday, parameters.firstMonthInYear, parameters.leapYear});
 
-    calendarOutput.reset(parameters);
+    calendarOutput.reset({parameters.dayOfThe1stJanuary, parameters.firstWeekday, parameters.firstMonthInYear, parameters.leapYear});
```
flomnes, line 150: Could indeed be passed by `&&`. But then make sure to mark the copy constructor as `delete`d

src/libs/antares/study/parameters/adq-patch-params.h
flomnes, line 1: Good catch ! `namespace Yuni` was injected through this header, should be avoided.

src/libs/antares/writer/CMakeLists.txt
```diff
@@ -22,12 +22,16 @@ add_library(result_writer
...
+add_library(Antares::result_writer ALIAS result_writer)
+
 target_include_directories(result_writer
     PUBLIC
     ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
 target_link_libraries(result_writer
+        PUBLIC
+        Antares::benchmarking
```
flomnes: This target is linked privately and publicly. One of them must go.

src/solver/CMakeLists.txt
```diff
@@ -73,8 +73,13 @@ add_executable(antares-solver
...
 
-set(ANTARES_SOLVER_LIBS		
-	date
+set(ANTARES_SOLVER_LIBS
+	Antares::date
+	Antares::benchmarking
+	Antares::result_writer
+	Antares::sys
+	Antares::infoCollection
+	Antares::checks
```
flomnes, line 82: We should avoid heterogeneous naming in CMake targets

- infoCollection
- result_writer
- tests-ts-numbers

In other words, we should pick a convention and stick to it.

src/tools/cleaner/main.cpp
```diff
@@ -29,7 +29,7 @@
 #include <antares/study/finder.h>
 #include <yuni/core/getopt.h>
 #include "../../ui/common/winmain.hxx"
```
flomnes, line 31: what an horrible dependency

flomnes (changes requested):
Add `libantares-version` ?

Does the build time decrease a bit after this change when editing a deep header, e.g study.h ?

.github/workflows/windows-vcpkg.yml
JasonMarechal25: It was a tentative to speed up windows build by removing limitation but it didn't seem to change anything

src/libs/antares/date/date.cpp
JasonMarechal25: I don't think r-value ref is proper here. It's a lot of supposition to ask for ownership of a value whereas with pass by value we're signifying that we're doing something to the value without impacting the client.

src/libs/antares/study/CMakeLists.txt
JasonMarechal25: I've added some comment
```
#Dirty trick to allow inclusion via <antares/study/X>
# Make more than just study visible but it's the lesser evil for now
```

The proper way would be to have a proper include folder with all study includes (or split study in several library/objects) but it is in itself a task

src/libs/antares/study/load.cpp
JasonMarechal25: I don't know if this usage should direct the implementation choice of the interface.
For all intent and purpose, parameters could have a Settings struct instead of several values and e would be passing a copy

flomnes: Yes, that would be nice

src/solver/CMakeLists.txt
JasonMarechal25: See #1606 for discussion

.github/workflows/windows-release.yml
```diff
@@ -103,7 +103,7 @@ jobs:
 
     - name: Build
       run: |
-           cmake --build _build --config release -j2
+           cmake --build _build --config release
```
flomnes: Please revert this

JasonMarechal25:
```suggestion
           cmake --build _build --config release -j2
```

flomnes (approved):

PR #1586, opened 2023-08-24
Remove deps to core

JasonMarechal25:
- Remove dependencies in libantarescore, now in study
- Remove several dependencies to libantarescore
- Don't remove dependency to libantarescore in UI


flomnes:
There seems to be a missing dependency to Boost. Build fails on Windows

```
D:\a\Antares_Simulator\Antares_Simulator\src\libs\antares\study\parts\common\cluster_list.cpp(1,10): fatal  error C1083: Cannot open include file: 'boost/algorithm/string/case_conv.hpp': No such file or directory [D:\a\Antares_Simulator\Antares_Simulator\_build\libs\antares\study\study.vcxproj]
```

flomnes (approved):

PR #1594, opened 2023-08-28
Move thermal noises to class Simulation

guilpier-code:
This PR intends to remove some global variables, as they are a bad practice.
In this particular PR, we 
- Remove **AleaCoutDeProductionParPalier** from struct **VALEURS_GENEREES_PAR_PAYS**. This global variable is more or less a copy of **pThermalNoisesByArea**, so it's not needed.
- Make simplifications and clarifications along the way, among which : 
   - removal of **valeursGenereesParPays** from weekly problem classes
   - we don't call **SIM_RenseignementProblemeHebdo** from **AdequacyPatchOptimization :: solve(...)** anymore, as a consequence of previous bullet 

**Remaining** : 
Some more work will have to follow if we want to move global variables down to more specialized part of the code : 
- Other data members in struct **VALEURS_GENEREES_PAR_PAYS** (now these members are only about hydro)
- **NumeroChroniquesTireesParPays** (vector of type **NUMERO_CHRONIQUES_TIREES_PAR_PAYS**)


src/solver/ts-generator/thermal.cpp
```diff
@@ -90,9 +90,8 @@ class GeneratorTempData final
                                const T& duration);
 
 private:
-    uint nbHoursPerYear;
-
-    const uint daysPerYear;
+    const uint nbHoursPerYear = 8760;
+    const uint daysPerYear = 365;
```
flomnes: Use a specific **constant.h** header ?

sylvlecl: Agreed, it will be better to have a named `constexpr` instead of the current situation where the obscure harcoded value 8760 is repeated everywhere.
(note that the constexpr can be more explicit than just "8760", for ex "DAYS_IN_YEAR * 24")

guilpier-code: Why not ? I have currently no idea of what to do with these constants "ever" (no risk they change), but we'll think about it

src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
```diff
@@ -24,42 +24,30 @@
...
-  double const* const* thermalNoisesByArea,
-  uint numSpace,
-  VAL_GEN_PAR_PAYS& valeursGenereesParPays)
+void ApplyRandomTSnumbers(const Study& study,
+                          double const* const* thermalNoisesByArea,
+                          uint numSpace)
 {
     auto& runtime = *study.runtime;
 
     uint year = runtime.timeseriesNumberYear[numSpace];
```
flomnes: Pass `year` directly ? It seems that `runtime` is only required for this.

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -301,6 +301,44 @@ void PrepareRandomNumbers(Data::Study& study,
...
+        //-----------------------------
+        auto end = area.thermal.list.mapping.end();
+        for (auto it = area.thermal.list.mapping.begin(); it != end; ++it)
+        {
+            auto cluster = it->second;
+            uint clusterIndex = cluster->areaWideIndex;
+            double& rnd = randomForYear.pThermalNoisesByArea[indexArea][clusterIndex];
+            double randomClusterProdCost(0.);
+            if (!cluster->enabled)
+                continue;
```
flomnes: Can be placed at the top of the loop ?

src/solver/aleatoire/alea_tirage_au_sort_chroniques.cpp
guilpier-code: `runtime` is also required for transmission capacities

src/solver/simulation/common-eco-adq.cpp
guilpier-code: Done

src/solver/ts-generator/thermal.cpp
guilpier-code: I think we should create a ticket for this, and not handle this in this PR, because it's a subject on its own.

src/solver/simulation/economy.cpp
```diff
@@ -143,9 +142,11 @@ bool Economy::year(Progression::Task& progression,
         pProblemesHebdo[numSpace].weekInTheYear = state.weekInTheYear = w;
         pProblemesHebdo[numSpace].HeureDansLAnnee = hourInTheYear;
 
-        ::SIM_RenseignementProblemeHebdo(state.study,
-          pProblemesHebdo[numSpace], state.weekInTheYear, numSpace, hourInTheYear,
-          valeursGenereesParPays);
+        ::SIM_RenseignementProblemeHebdo(study, pProblemesHebdo[numSpace], state.weekInTheYear,
```
guilpier-code, line 145: Note that we add a call to a new function **BuildThermalPartOfWeeklyProblem**. It's an extraction from **SIM_RenseignementProblemeHebdo** (which is gigantic). We first needed to extract it because of adequacy patch mode calling **SIM_RenseignementProblemeHebdo** a second time and erasing thermal production costs (no further details, sorry). We don't especially need this extraction anymore, but we chose to leave it here, as a remainder that we should split the gigantic **SIM_RenseignementProblemeHebdo** into several simpler ones.

src/libs/antares/study/area/scratchpad.cpp
```diff
@@ -123,7 +124,7 @@ AreaScratchpad::AreaScratchpad(const StudyRuntimeInfos& rinfos, Area& area) : ts
         auto& colPowerOverWater = m[PreproHydro::powerOverWater];
         auto& colMaxEnergy = m[PreproHydro::maximumEnergy];
 
-        for (uint m = 0; m < rinfos.nbMonthsPerYear; ++m)
+        for (uint m = 0; m < nbMonthsPerYear; ++m)
             valueCol += colMaxEnergy[m] * (1. - colPowerOverWater[m]);
```
a-zakir:
```suggestion
        for (uint month = 0; month < nbMonthsPerYear; ++month)
            valueCol += colMaxEnergy[month] * (1. - colPowerOverWater[month]);
```

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -301,6 +301,44 @@ void PrepareRandomNumbers(Data::Study& study,
         }
         problem.CoutDeDefaillanceNegative[area.index] = area.thermal.spilledEnergyCost + alea;
 
+
+        //-----------------------------
+        // Thermal noises
+        //-----------------------------
+        auto end = area.thermal.list.mapping.end();
+        for (auto it = area.thermal.list.mapping.begin(); it != end; ++it)
+        {
```
flomnes, line 309: Add function for this block.

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -190,7 +190,7 @@ bool ShouldUseQuadraticOptimisation(const Data::Study& study)
     if (!flowQuadEnabled)
         return false;
 
-    uint maxHours = study.runtime->nbHoursPerYear;
+    uint maxHours = 8760;
```
payetvin: We should not use hardcoded values for hoursPerYear, still need to figure out why 8760 and sometimes 8784
macro HOURS_PER_YEAR can be used instead of runtime

src/solver/simulation/sim_allocation_tableaux.cpp
```diff
@@ -39,8 +39,7 @@ static void AllocateResultsForEconomicMode(const Data::Study& study)
     transitMoyenInterconnexionsRecalculQuadratique.resize(study.runtime->interconnectionsCount());
 
     for (uint i = 0; i != study.runtime->interconnectionsCount(); i++)
-        transitMoyenInterconnexionsRecalculQuadratique[i]
-            .assign(study.runtime->nbHoursPerYear , 0.);
+        transitMoyenInterconnexionsRecalculQuadratique[i].assign(8760, 0.);
```
payetvin: same

flomnes:
@guilpier-code Can you resolve conflicts ?

flomnes:
@guilpier-code Please resolve conflicts

flomnes (approved):

flomnes (approved):

PR #1607, opened 2023-09-03
Refactor constraint building

a-zakir:
this follows on from the work on #1234 #1605


src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -30,129 +30,28 @@
...
-
-        constraintNamer.ShortTermStorageLevel(ProblemeAResoudre->NombreDeContraintes, storage.name);
-        OPT_ChargerLaContrainteDansLaMatriceDesContraintes(
-          ProblemeAResoudre, Pi, Colonne, nombreDeTermes, '=');
-    }
-}
-
 void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* problemeHebdo, Solver::IResultWriter& writer)
 {
     int var;
```
guilpier-code:
```C++
const bool exportStructure = problemeHebdo->ExportStructure;	    
const bool firstWeekOfSimulation = problemeHebdo->firstWeekOfSimulation;
```
We could move this where it's useful : at end of file. We don't code in C anymore

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,347 @@
...
+
+struct LayerStorage
+{
+    LayerStorage(unsigned area, unsigned layer) : area(area), layer(layer)
+    {
+    }
+    unsigned area, layer;
+};
+
+using Variant = std::variant<DispatchableProduction,
```
guilpier-code: _Temporary comment_ 
**Variant** : this name means nothing to me : I can see that it's is a **std::variant**, no need to recall it in the type name, this type name does not bring more information. When crossing it, we don't know what it is. We have to search if  we want to know

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,347 @@
...
+                             HydProdUp,
+                             Pumping,
+                             HydroLevel,
+                             Overflow,
+                             FinalStorage,
+                             LayerStorage,
+                             PositiveUnsuppliedEnergy,
+                             NegativeUnsuppliedEnergy>;
+
+class Visitor
```
guilpier-code: _Temporary comment_ 
**class Visitor** : renaming : what do we visit ?

src/solver/optimisation/constraint_builder.cpp
```diff
@@ -0,0 +1,68 @@
...
+    // {
+    //     pdt %= nbTimeSteps;
+    //     if (pdt < 0)
+    //         pdt += problemeHebdo.NombreDePasDeTemps;
+    // }
+    // if (pdt < 0 || pdt >= nbTimeSteps)
+    // {
+    //     return -1;
+    // }
+
```
guilpier-code: Remove comment

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,347 @@
...
+
+// TODO remove relative include
+#include "../simulation/sim_structure_probleme_economique.h"
+
+#include <variant>
+#include <utility>
+
+namespace Variable
+{
+struct SingleIndex
```
guilpier-code: _Temporary comment_
Without yet knowing what these **struct** are for, I have a strange feeling of code duplication

guilpier-code (changes requested):

src/solver/optimisation/constraint_builder.h
a-zakir: i couldn't agree more, Since Variant could be one of the Antares Simulator variables, it can be renamed AntaresVariables?
LinearProblemVariables?

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,347 @@
...
+
+    unsigned int hourInWeek_ = 0;
+
+    char operator_;
+    double rhs_ = 0;
+    int nombreDeTermes_ = 0;
+    // ConstraintNamer ConstraintNameManager;
+};
+
+class Constraint
```
guilpier-code: _Design concern_

class **Constraint** obviously represents a general constraint.
It's a **very good sign** that a business notion (here an linear optimization concept) is represented by a class : People find in the code what they know in the business field. It's a great step towards clarity and correct modularity.

The problem here is that this class does not have the expected attributes and behaviors of a constraint : 
- **Attributes** : 
    - list of variables and related weights
    - sign
    - RHS
    - possibly an index (to insert the constraint at the right place in the opt problem)
- **Behaviors** : add a variable with a coefficient, give a sign, add an RHS. Maybe a setter and getter of index.

This class should not know about anything else : it's an agnostic element of a much bigger picture.

**Why does it favor modularity ?** Because if anything changes in the optimization problem formulation, the class constraint won't be impacted : we still need the constraint (with its ever lasting same attributes and behaviors) as a brick.

Instead, this class currently contains a constraint builder (adding the constraint to the optimization problem) and the problem itself. If the problem changes, we have to make changes in the constraint. If the way we build the problem changes, then the class **Constraint** is recompiled.

Unfortunately, should this comment be taken into account, this would require many changes.

src/solver/optimisation/constraint_builder.h
a-zakir: file constraint_builder.cpp:
 const Variable::Visitor visitor(varNative[pdt],
                                    problemeHebdo.NumeroDeVariableStockFinal,
                                    problemeHebdo.NumeroDeVariableDeTrancheDeStock);
    return std::visit(visitor, variable);

the visitor "visitor" is applied on "variable" which is a "variant", std::visit will invoke the right overload of Visitor::operator()(T)
😜

src/solver/optimisation/constraint_builder.h
a-zakir: Each of these struct represents a variable, and stores the variable's index, so feel free to suggest an alternative.

guilpier-code (changes requested):

src/solver/optimisation/constraint_builder.h
a-zakir: Thanks, we had a fruitful discussion on it (with @JasonMarechal25 ). in my sight this Pr is enough large and may certainly be enhanced.
The question of Rhs is crucial. For the moment, they are built twice (1st and 2nd optimization), the only difference being that the contribution of "ReserveJMoins1" is taken into account. there's obviously food for thought

src/solver/optimisation/AreaBalance.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include "constraint_builder.h"
+
+class AreaBalance : private Constraint
```
guilpier-code: This constraint add itself to the optimization problem. This makes the constraint dependent to the opt problem. Change the problem in anyway, this class recompiles.
On this, please read [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1607#discussion_r1318290963)

a-zakir:
Constraint is just a "store" struct, it does nothing else so no need to privatize ProblemHebdo, right?
However the derived classes such AreaBalance need to encapsulate some stuff like getting the index of the variable or the constraint builder. With a public inheritance, Areabalance would have public and private member data(Sonar doesn't like that nor protected member). this is not a concern with Private inheritance (all data member are private)

src/solver/optimisation/constraint_builder.h
guilpier-code: To be more accurate : if a 80 lines of code are a repetition of 4 lines blocks, we should wonder why.
At the end, one index (that is one line of code somewhere) is enough.
**OK**, it's because of the **visitor** thing, that associates a variable index to a place where to store this index.
So on one side we have 21 struct containing each only an index value, and on the other side, we have 21 `operator()` to store the indices. 
Could we not reduce this ?
For instance with a std::map ? index --> std::function ?
As already said [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1607#discussion_r1318520730), it could be interesting **to keep these classes containing just an index**. And these classes instances could be stored in a **Constraint**, like it's recommended [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1607#discussion_r1318290963)

guilpier-code (changes requested):

src/solver/optimisation/constraint_builder.h
guilpier-code: A related things would be the variables.
From what I can see in this PR, it's only an index (+ possibly a name as a std::string for purpose of MPS printing).
And that's the reason why you embedded an index in a **struct** ! ok... I just realize it.

src/solver/optimisation/constraint_builder.h
guilpier-code: ok, this visitor associates a variable index and a place to store it. Not a bad idea to separate : 
- the variable index
- the place to store this index

The index is a business concept (every variable has an index, in any opt problem of all kind), whereas the place where to store depends on the implementation.

src/solver/optimisation/constraint_builder.h
a-zakir: i don't see repetition, they are different struct with single base struct. but ok,
for ex something like std::map<string, uint> will be more accurate? ( with 80 calls to std::map::insert)?

src/solver/optimisation/AreaBalance.h
a-zakir: Yes, getting rid of problemHebdo to send only modifiable or read-only variables is one of the ultimate goals.

src/solver/optimisation/constraint_builder.h
a-zakir: Dear @guilpier-code if you change your comment after my reply, the disccusion becomes mad.

src/solver/optimisation/opt_construction_matrice_des_contraintes_cas_lineaire.cpp
```diff
@@ -182,522 +97,52 @@ void OPT_ConstruireLaMatriceDesContraintesDuProblemeLineaire(PROBLEME_HEBDO* pro
         CORRESPONDANCES_DES_CONTRAINTES& CorrespondanceCntNativesCntOptim
             = problemeHebdo->CorrespondanceCntNativesCntOptim[pdt];
 
+        AreaBalance areaBalance(problemeHebdo);
```
guilpier-code: We can see that we are hampered by the order of constraints insertion in the problem.
- It makes the current source file more complex than it should because we have to be careful with the loop over time step, areas, ... the constraint is built in.
- We mix insertion order and constraint building

Could we not separate constraint insertion order and building ?
So we could create and use constraints with more flexibility.
Like pointed in [this ticket](https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/1612),

guilpier-code (dismissed):

guilpier-code:
**To sum up a possible way of doing things**

This is a very simple piece of code, without knowing about actual feasibility or development cost it would require (I mean changes in Variables creation, changes due to constraint creation with for instance an RHS, ...)

```C++
class Variable
{
public:
    variable(int index, ...);
    ...
private:
    int index_; 
    double boundInf_; // TODO : for later
    double boundSup_; // TODO : for later
}

class Constraint
{
public:
    Constraint(char sign, double rhs, int index);
    addVariable(Variable* var);
    setRhs(double rhs);

private:
    int index_;
    std::vector<Variable*> listVar_;
    char sign;
    double rhs_;
}

class ContraintBuilder
{
public:
    ContraintBuilder(problemAResoudre* pbmToBeSolved);
    addConstraint(Constraint* c); 
private:
    problemAResoudre pbmToBeSolved_;
}

ContraintBuilder::addConstraint(Constraint* c)
{
    // Calling OPT_ChargerLaContrainteDansLaMatriceDesContraintes
    // with what is found inside constraint c
}
```
Classes **Variable** and **Constraint** should not change in the future (designed to stick to the business concern).
 
Class **ContraintBuilder** is an "implementation" class, meaning that its role is to insert constraint in the optimization problem, but with the current modelling of the opt problem (knowing that this modelling could change soon).
This class aims at being changed or replaced.

Then (for instance) : 
```C++
void buildConstraintMatrix(....)
{
    buildBalance();
    buildFictitiousLoad();
    ....
}

buildBalance(ContraintBuilder& constrBuilder)
{
    for (area; ...)
    {
        for (int pdt)
        {             
             Constraint balance(sign, rhs, index);

             Variable var(...);
             balance.add(var);
            
            Variable otherVar(...);
            balance.add(otherVar);
            
            // Finally : 
            constrBuilder.addConstraint(balance);
        }
    }
		
}
```

src/solver/optimisation/constraint_builder.h
guilpier-code: > Dear @guilpier-code if you change your comment after my reply, the disccusion becomes mad.

Sorry dear @a-zakir, when I updated my comment, I didn't refresh this page in the first place, so I didn't see your reply to it.

src/solver/optimisation/BindingConstraintWeek.cpp
```diff
@@ -0,0 +1,89 @@
...
+            int pdt1;
+            if (offset >= 0)
+            {
+                pdt1 = (pdt + offset) % problemeHebdo->NombreDePasDeTempsPourUneOptimisation;
+            }
+            else
+            {
+                pdt1 = (pdt + offset + problemeHebdo->NombreDePasDeTemps)
+                       % problemeHebdo->NombreDePasDeTempsPourUneOptimisation;
+            }
```
flomnes: Use argument `offset` from `include` ?

src/solver/optimisation/constraint_builder.h
flomnes: `Constraints`

src/solver/optimisation/constraint_builder.h
flomnes: `ConstraintVisitor`

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+private:
+    int getVariableIndex(const Variable::Variant& variable, int shift, bool wrap, int delta) const;
+
+public:
+    ConstraintBuilder& include(Variable::Variant var,
+                               double coeff,
+                               int shift = 0,
+                               bool wrap = false,
```
flomnes:
```suggestion
                               int offset = 0,
```

src/solver/optimisation/BindingConstraintWeek.cpp
a-zakir: yes!

src/solver/optimisation/constraint_builder.h
a-zakir: Sure?

flomnes (approved):

flomnes:
@guilpier-code Can you please update your review ?

flomnes:
@guilpier-code Can you update your review ?

src/solver/optimisation/constraint_builder.h
a-zakir: will be done in #1628

src/solver/optimisation/AreaBalance.h
a-zakir: see #1628

src/solver/optimisation/AreaBalance.cpp
```diff
@@ -0,0 +1,74 @@
+#include "AreaBalance.h"
```
sylvlecl: For clarity sake, I would propose to move all constraint implementation files into a dedicated sub-folder (`src/solver/optimisation/constraints`)

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,376 @@
+#pragma once
```
sylvlecl: We should rename the file accordingly with the convention used in other new files (CamelCase), since we seem to be adopting that convention.

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,376 @@
...
+    {
+    }
+
+    ConstraintBuilder& updateHourWithinWeek(unsigned hour)
+    {
+        hourInWeek_ = hour;
+        return *this;
+    }
+
+public:
```
sylvlecl: redundant

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,376 @@
...
+    ConstraintBuilder& DispatchableProduction(unsigned int index,
```
sylvlecl: I think we should add some documentation, either at the method or class level, to explain the reader what's the role of the class/method, and in particular what are the non obvious things:
- what is the difference between shift and delta ?
- what means wrapping ?
- ...

(although already the refactoring documents the code better than before :) )

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,376 @@
...
+    double rhs_ = 0;
+    int nombreDeTermes_ = 0;
+    // ConstraintNamer ConstraintNameManager;
+
+    int GetShiftedTimeStep(int shift, bool wrap, int delta) const;
+    void AddVariable(int index, double coeff);
+    Variable::VariableManager GetVariableManager( int shift = 0, bool wrap = false, int delta = 0);
+};
+
+class Constraint
```
sylvlecl: The name seems misleading: 
those classes actually don't represent individual constraints, but more constraint "types", or constraint factories.

For example, for the reader it's hard to understand why a constraint uses a "constraint builder":  one would expect a constraint builder to **create** constraints instead.
Renaming it would make more sense: a constraint type or factory uses a builder to actually create constraints.

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,376 @@
...
+    ConstraintBuilder(
+      PROBLEME_HEBDO& problemeHebdo,
+      const std::vector<CORRESPONDANCES_DES_VARIABLES>& CorrespondanceVarNativesVarOptim) :
+     problemeHebdo(problemeHebdo),
+     problemeAResoudre(*problemeHebdo.ProblemeAResoudre),
+     varNative(CorrespondanceVarNativesVarOptim)
+    {
+    }
+
+    ConstraintBuilder& updateHourWithinWeek(unsigned hour)
```
sylvlecl: This method seems dangerous and seems to overlap with the role of `shift` argument in the other methods.

To clarify that, I would propose that the "base hour" for a given constraint is **never** changed while building a constraint. Instead, we should use shift arguments in those places. Otherwise, one could make a mistake by changing involuntarily the base hour for subsequent calls.

To enforce this, the best design would be that a constraint builder is used only for creating one constraint, and its base hour is defined in its constructor and not changeable afterwards.
Note that otherwise, another risk is that we could re-use the builder and forget to change its base hour.

src/solver/optimisation/AreaHydroLevel.cpp
```diff
@@ -0,0 +1,35 @@
...
+    CORRESPONDANCES_DES_CONTRAINTES& CorrespondanceCntNativesCntOptim
+      = problemeHebdo->CorrespondanceCntNativesCntOptim[pdt];
+    CorrespondanceCntNativesCntOptim.NumeroDeContrainteDesNiveauxPays[pays]
+      = problemeHebdo->ProblemeAResoudre->NombreDeContraintes;
+    if (problemeHebdo->CaracteristiquesHydrauliques[pays].SuiviNiveauHoraire)
+    {
+        builder.updateHourWithinWeek(pdt).HydroLevel(pays, 1.0);
+        if (pdt > 0)
+        {
+            builder.updateHourWithinWeek(pdt - 1).HydroLevel(pays, -1.0);
```
sylvlecl: For example here, why not use a `shift` instead of changing the "base" hour ?
It seems dangerous. Fortunately we restore the timestep later, but we should not need to do this.

src/solver/optimisation/AreaBalance.cpp
```diff
@@ -0,0 +1,74 @@
...
+static void shortTermStorageBalance(
+  const ::ShortTermStorage::AREA_INPUT& shortTermStorageInput,
+  ConstraintBuilder& constraintBuilder,
+  const CORRESPONDANCES_DES_VARIABLES& CorrespondanceVarNativesVarOptim)
+{
+    for (const auto& storage : shortTermStorageInput)
+    {
+        unsigned index = storage.clusterGlobalIndex;
+        if (const int varInjection
+            = CorrespondanceVarNativesVarOptim.SIM_ShortTermStorage.InjectionVariable[index];
```
sylvlecl: I think this is redundant with the check done inside the builder.

Then if we remove it, we should not need anymore the `CORRESPONDANCES_DES_VARIABLES`.

src/solver/optimisation/ConsistenceNODU.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include "constraint_builder.h"
+
+class ConsistenceNODU : private Constraint
```
sylvlecl: I don't know if everyone would agree, but I would prefer the verbose name instead of `NODU`

src/solver/optimisation/ConsistenceNODU.cpp
```diff
@@ -0,0 +1,47 @@
...
+          = problemeHebdo->PaliersThermiquesDuPays[pays];
+
+        int NombreDePasDeTempsPourUneOptimisation
+          = problemeHebdo->NombreDePasDeTempsPourUneOptimisation;
+
+        int Pdtmoins1 = pdt - 1;
+        if (Pdtmoins1 < 0)
+            Pdtmoins1 = NombreDePasDeTempsPourUneOptimisation + Pdtmoins1;
+
+        CORRESPONDANCES_DES_VARIABLES& CorrespondanceVarNativesVarOptimTmoins1
```
sylvlecl: Not used.

sylvlecl (commented):
Very nice work overall:
I suggest a few naming/organization/design improvements, for discussion.

I hope it's not redundant with previous discussions and with the subsequent draft PR.

src/solver/optimisation/constraint_builder.h
a-zakir: definitely!

src/solver/optimisation/AreaHydroLevel.cpp
a-zakir: right, we should use *shift*

src/solver/optimisation/ConsistenceNODU.h
a-zakir: @flomnes @guilpier-code

flomnes: I agree with @sylvlecl !

flomnes:
Please add a description for new classes : what do they do, how they work, etc.

src/solver/optimisation/constraint_builder.h
a-zakir: done

src/solver/optimisation/AreaBalance.cpp
a-zakir: we are ok now

src/solver/optimisation/constraint_builder.h
a-zakir: k

src/solver/optimisation/constraint_builder.h
a-zakir: i'm agree with you on building one constraint at time, unfortunately we have some weird timestep stuff like in [src/solver/optimisation/constraints/MinDownTime.cpp](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/31b39eface118aa51846ea816c80f44ffec43973/src/solver/optimisation/constraints/MinDownTime.cpp)
and it has to be addressed in dedicated Pr

src/solver/optimisation/AreaHydroLevel.cpp
a-zakir: please [see ](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1607#discussion_r1333000184)

src/solver/optimisation/constraints/AreaBalance.cpp
```diff
@@ -0,0 +1,60 @@
+#include "AreaBalance.h"
+
+static void shortTermStorageBalance(const ::ShortTermStorage::AREA_INPUT& shortTermStorageInput,
+                                    ConstraintBuilder& constraintBuilder)
+{
+    for (const auto& storage : shortTermStorageInput)
+    {
+        unsigned index = storage.clusterGlobalIndex;
+        constraintBuilder.ShortTermStorageInjection(index, 1.0);
+        constraintBuilder.ShortTermStorageWithdrawal(index, -1.0);
```
flomnes:
```suggestion
        constraintBuilder.ShortTermStorageInjection(index, 1.0)
        .ShortTermStorageWithdrawal(index, -1.0);
```

flomnes (commented):
`nbTermesContraintesPourLesCoutsDeDemarrage` seems to be duplicated (ie you have multiple copies if this variable).

It is required for `problemeHebdo->NbTermesContraintesPourLesCoutsDeDemarrage`.

PR #1609, opened 2023-09-04
Remove liblib* CMake targets

### LINKED ISSUE #1602: A few liblib* CMake targets
Example
```
[ 89%] Linking CXX static library liblibantares-solver-main-adequacy.a
[ 89%] Linking CXX static library liblibantares-solver-constraints-builder.a
```

CMake automatically adds the "lib" prefix.

We can either
- Leave it as it is now
- Remove this prefix from CMake [CMAKE_STATIC_LIBRARY_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_STATIC_LIBRARY_PREFIX.html)
- Remove it from our CMakeLists.txt

flomnes:
This is mostly replacing `libantares*` by `antares*` in CMake targets.


flomnes:
Somehow, it seems that the SonarCloud tries to checkout a non-existent commit. Strange, but probably not related to this change.

JasonMarechal25 (approved):

PR #1614, opened 2023-09-08
Fix/global var hydro modulable

guilpier-code:
Before the current pull request, type **VAL_GEN_PAR_PAYS** contained only data members related to hydro, and even hydro made available by hydro ventilation.

- Moving `VAL_GEN_PAR_PAYS valeursGenereesParPays` from class **ISimulation** into class **HydroManagement**, and renaming it `ALL_HYDRO_VENTILATION_RESULTS ventilationResults`.
- Cleaning the class **HydroManagement** : 
    - Removing dependency to **Study**, replaced with dependencies to **AreaList**, **Parameters**, **Calendar**, ...
    - Renaming **operator () (...)** into **makeVentilation(...)**, much more clear

### DIFF
--- a/src/solver/hydro/management/management.h
+++ b/src/solver/hydro/management/management.h
@@ -42,39 +42,71 @@ class State;
 }
 } // namespace Solver
 
+enum
+{
+    //! The maximum number of days in a year
+    dayYearCount = 366
+};
+
+//! Temporary data
+struct TmpDataByArea
+{
+    //! Monthly local effective demand
+    double MLE[12];
+    //! Monthly optimal generation
+    double MOG[12];
+    //! Monthly optimal level
+    double MOL[12];
+    //! Monthly target generations
+    double MTG[12];
+    //! inflows
+    double inflows[12];
+    //! monthly minimal generation
+    std::array<double, 12> mingens;
+
+    //! Net demand, for each day of the year, for each area
+    double DLN[dayYearCount];
+    //! Daily local effective load
+    double DLE[dayYearCount];
+    //! Daily optimized Generation
+    double DOG[dayYearCount];
+    //! daily minimal generation
+    std::array<double, dayYearCount> dailyMinGen;
+
+    // Data for minGen<->inflows preChecks
+    //! monthly total mingen
+    std::array<double, 12> totalMonthMingen;
+    //! monthly total inflows
+    std::array<double, 12> totalMonthInflows;
+    //! yearly total mingen
+    double totalYearMingen;
+    //! yearly total inflows
+    double totalYearInflows;
+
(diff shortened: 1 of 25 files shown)


payetvin (approved):

src/solver/hydro/management/management.h
```diff
@@ -42,39 +42,84 @@ class State;
...
-                    uint numSpace,
-                    VAL_GEN_PAR_PAYS& valeursGenereesParPays);
-
-public:
-    //! Random number generator
-    MersenneTwister random;
+    void makeVentilation(double* randomReservoirLevel,
+                        Solver::Variable::State& state,
+                        uint y,
+                        uint numSpace);
```
flomnes, line 118:
```suggestion
    ALL_HYDRO_VENTILATION_RESULTS makeVentilation(double* randomReservoirLevel,
                        Solver::Variable::State& state,
                        uint y,
                        uint numSpace);
```
One single method instead of two.

src/solver/simulation/economy.h
```diff
@@ -81,7 +81,7 @@ class Economy
               yearRandomNumbers& randomForYear,
               std::list<uint>& failedWeekList,
               bool isFirstPerformedYearOfSimulation,
-              VAL_GEN_PAR_PAYS& valeursGenereesParPays);
+              ALL_HYDRO_VENTILATION_RESULTS&);
```
flomnes:
```suggestion
              const ALL_HYDRO_VENTILATION_RESULTS&);
```

flomnes (changes requested):

src/solver/simulation/economy.h
guilpier-code: Done

src/solver/hydro/management/management.h
guilpier-code: ok. So this means that we could simplify the caller too : 
```C++
// solver.hxx
yearFailed[y] = !simulation_->year(...,
                                   ...,
                                   simulation_->hydroManagement.makeVentilation(/* some args */));
```
As a consequence, we would skip calls the timer / add duration, unless we put these calls in the **HydroManagement :: makeVentilation**.
What do you recommend ?

flomnes (approved):

src/solver/hydro/management/management.h
flomnes: I think we should keep the timer, but have a single function.
```cpp
// Timer ON
auto hydroVentilation = simulation_->hydroManagement.makeVentilation(/* args */);
// Timer OFF
yearFailed[y] = !simulation_->year(...,
                                   ...,
                                   hydroVentilation);
```

PR #1617, opened 2023-09-11
remove ProblemHebdo from Constraint Builder lib

a-zakir:
follows #1607


flomnes:
You may want to change the base branch.

Also, if we want to keep a uniform interface (`add` without one or `std::vector`s), we could provide these as parameters to class constructors (`AreaBalance` for example)

a-zakir:
> You may want to change the base branch.
> 
> Also, if we want to keep a uniform interface (`add` without one or `std::vector`s), we could provide these as parameters to classes (`AreaBalance` for example)

Thanks, I'd do so

Issue #1575, opened 2023-08-21
Remove malloc in src/solver/hydro/


payetvin:
Files are .c, upgrade this to .cpp and remove manual memory allocation

Code is duplicated between 
```
daily/h2o_j_resoudre_le_probleme_lineaire.cpp
daily2/h2o2_j_resoudre_le_probleme_lineaire.cpp

```

PR #1626, opened 2023-09-15
Remove malloc in src/solver/hydro/

payetvin:
issue #1575
previous PR: #1620


src/solver/hydro/daily/h2o_j_free.cpp
```diff
@@ -51,7 +51,8 @@ void H2O_J_Free(DONNEES_MENSUELLES* DonneesMensuelles)
             SPX_LibererProbleme(ProbSpx);
     }
 
-    free(ProblemeHydraulique.Probleme);
+    PROBLEME_SIMPLEXE* pb = (PROBLEME_SIMPLEXE*)ProblemeHydraulique.Probleme;
+    delete pb;
```
flomnes: Why do we need to cast here ?

If you really need to cast, please use `static_cast` instead, it offers more guarantees.

src/solver/hydro/daily/h2o_j_initialiser_les_bornes_des_variables.cpp
```diff
@@ -32,18 +32,14 @@
...
-
     const std::vector<double>& TurbineMax = DonneesMensuelles->TurbineMax;
     const std::vector<double>& TurbineMin = DonneesMensuelles->TurbineMin;
     const std::vector<double>& TurbineCible = DonneesMensuelles->TurbineCible;
     std::vector<double>& Turbine = DonneesMensuelles->Turbine;
 
     PROBLEME_HYDRAULIQUE& ProblemeHydraulique = DonneesMensuelles->ProblemeHydraulique;
 
-    NbPdt = ProblemeHydraulique.NbJoursDUnProbleme[NumeroDeProbleme];
+    int NbPdt = ProblemeHydraulique.NbJoursDUnProbleme[NumeroDeProbleme];
```
flomnes:
```suggestion
    const int NbPdt = ProblemeHydraulique.NbJoursDUnProbleme[NumeroDeProbleme];
```

src/solver/hydro/daily/h2o_j_lisser_les_sur_turbines.cpp
```diff
@@ -32,55 +32,43 @@
...
-    double MargeMin;
-    double SurTurbine;
     PROBLEME_HYDRAULIQUE& ProblemeHydraulique = DonneesMensuelles->ProblemeHydraulique;
 
     std::vector<double>& TurbineMax = DonneesMensuelles->TurbineMax;
     std::vector<double>& Turbine = DonneesMensuelles->Turbine;
     std::vector<double>& TurbineCible = DonneesMensuelles->TurbineCible;
 
-    Flag = (char*) ProblemeHydraulique.ProblemeLineairePartieVariable[NumeroDeProbleme].Xmax.data();
+    char* Flag = (char*) ProblemeHydraulique.ProblemeLineairePartieVariable[NumeroDeProbleme].Xmax.data();
```
flomnes: I think this part is very error-prone
- We have a `std::vector<double>`
- We cast it's pointer to `char*`
- We use it as a temporary `Flag` object

The resulting `Xmax` contains essentially garbage. If we try to read it, we'll get inconsistent values such as 1.22e308.

If `Flag` is used only here, I strongly suggest declaring a `std::vector<bool>` here instead.

src/solver/hydro/daily/h2o_j_resoudre_le_probleme_lineaire.cpp
```diff
@@ -66,23 +56,25 @@ void H2O_J_ResoudreLeProblemeLineaire(DONNEES_MENSUELLES* DonneesMensuelles, int
...
     {
-        Probleme = (PROBLEME_SIMPLEXE*)malloc(sizeof(PROBLEME_SIMPLEXE));
-        if (Probleme == NULL)
+        Probleme = new PROBLEME_SIMPLEXE;
+        if (!Probleme)
         {
             DonneesMensuelles->ResultatsValides = EMERGENCY_SHUT_DOWN;
             return;
         }
         ProblemeHydraulique.Probleme = (void*)Probleme;
```
flomnes: What you want to use is
```suggestion
        Probleme = new(std::nothrow) PROBLEME_SIMPLEXE;
        if (!Probleme)
        {
            DonneesMensuelles->ResultatsValides = EMERGENCY_SHUT_DOWN;
            return;
        }
        ProblemeHydraulique.Probleme = (void*)Probleme;
```
Your version throws in case of allocation error. See https://en.cppreference.com/w/cpp/memory/new/nothrow

Don't forget `#include <memory>`.

src/solver/hydro/management/daily.cpp
```diff
@@ -391,7 +391,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
             uint firstDay = study.calendar.months[simulationMonth].daysYear.first;
             uint endDay = firstDay + daysPerMonth;
 
-            DONNEES_MENSUELLES problem = H2O_J_Instanciation();
+            DONNEES_MENSUELLES& problem = *H2O_J_Instanciation();
```
flomnes: This is dangerous. De-referencing a pointer to keep it in a reference could lead to a dangling reference. Example
```
  int& p = *(new int(5));
  delete &p;
  std::cout << p << std::endl; // invalid read / segfault
```
When we have a pointer, we keep in mind that it may be `nullptr`. There is no such thing for references, so you have to be careful.

src/solver/hydro/monthly/h2o_m_fonctions.h
```diff
@@ -27,15 +27,15 @@
...
-void H2O_M_InitialiserBornesEtCoutsDesVariables(DONNEES_ANNUELLES*);
-void H2O_M_InitialiserLeSecondMembre(DONNEES_ANNUELLES*);
-void H2O_M_ResoudreLeProblemeLineaire(DONNEES_ANNUELLES*, int);
-void H2O_M_AjouterBruitAuCout(DONNEES_ANNUELLES*);
+void H2O_M_ConstruireLesContraintes(DONNEES_ANNUELLES&);
+void H2O_M_ConstruireLesVariables(DONNEES_ANNUELLES&);
+void H2O_M_InitialiserBornesEtCoutsDesVariables(DONNEES_ANNUELLES&);
+void H2O_M_InitialiserLeSecondMembre(DONNEES_ANNUELLES&);
+void H2O_M_ResoudreLeProblemeLineaire(DONNEES_ANNUELLES&, int);
+void H2O_M_AjouterBruitAuCout(DONNEES_ANNUELLES&);
```
flomnes, line 39: Make a C++ class with these methods ? Most of them private, probably just a few public.

Instantiation => constructor
Free => destructor

src/solver/hydro/monthly/h2o_m_resoudre_le_probleme_lineaire.cpp
```diff
@@ -170,39 +158,39 @@ void H2O_M_ResoudreLeProblemeLineaire(DONNEES_ANNUELLES* DonneesAnnuelles, int N
...
         }
     }
 
-    if (ProblemeLineairePartieVariable->ExistenceDUneSolution == OUI_SPX)
+    if (ProblemeLineairePartieVariable.ExistenceDUneSolution == OUI_SPX)
     {
-        ProblemeHydraulique->CoutDeLaSolution = 0.0;
-        for (Var = 0; Var < Probleme->NombreDeVariables; Var++)
+        ProblemeHydraulique.CoutDeLaSolution = 0.0;
+        for (int i = 0; i < Probleme->NombreDeVariables; i++)
```
flomnes: `Var` was better than `i`

src/solver/hydro/monthly/h2o_m_resoudre_le_probleme_lineaire.cpp
```diff
@@ -46,43 +46,33 @@ extern "C"
...
-    ProbSpx = (PROBLEME_SPX*)ProblemeHydraulique->ProblemeSpx[NumeroDeReservoir];
-
-    Probleme = (PROBLEME_SIMPLEXE*)ProblemeHydraulique->Probleme;
-    if (Probleme == NULL)
+    if (!Probleme)
     {
-        Probleme = (PROBLEME_SIMPLEXE*)malloc(sizeof(PROBLEME_SIMPLEXE));
-        if (Probleme == NULL)
+        Probleme = new PROBLEME_SIMPLEXE;
+        if (!Probleme)
```
flomnes: Use the `std::nothrow` variant here.

src/solver/hydro/monthly/h2o_m_resoudre_le_probleme_lineaire.cpp
```diff
@@ -46,43 +46,33 @@ extern "C"
...
+    PROBLEME_HYDRAULIQUE& ProblemeHydraulique = DonneesAnnuelles.ProblemeHydraulique;
+    PROBLEME_LINEAIRE_PARTIE_VARIABLE& ProblemeLineairePartieVariable
+        = ProblemeHydraulique.ProblemeLineairePartieVariable;
+    PROBLEME_LINEAIRE_PARTIE_FIXE& ProblemeLineairePartieFixe
+        = ProblemeHydraulique.ProblemeLineairePartieFixe;
 
-    PROBLEME_HYDRAULIQUE* ProblemeHydraulique;
-    PROBLEME_LINEAIRE_PARTIE_VARIABLE* ProblemeLineairePartieVariable;
-    PROBLEME_LINEAIRE_PARTIE_FIXE* ProblemeLineairePartieFixe;
+    char PremierPassage = OUI;
```
flomnes: Use `bool`

flomnes (changes requested):
Some stuff left to do. The original code was quite dirty.

src/solver/hydro/daily/h2o_j_free.cpp
payetvin: can't delete a void*, it creates a warning 
I'll use static cast then

flomnes: Then why do you need to `delete` ?

flomnes (changes requested):
Only change required is now `Flag`

src/solver/hydro/daily/h2o_j_lisser_les_sur_turbines.cpp
```diff
@@ -32,94 +32,83 @@
...
-    NbPdt = ProblemeHydraulique.NbJoursDUnProbleme[NumeroDeProbleme];
+    std::vector<bool> flag(NbPdt);
 
-    SurTurbineARepartir = 0.0;
-    for (Pdt = 0; Pdt < NbPdt; Pdt++)
+    double SurTurbineARepartir = 0.0;
+    for (int Pdt = 0; Pdt < NbPdt; Pdt++)
     {
-        Flag[Pdt] = 0;
+        flag[Pdt] = 0;
```
flomnes:
```suggestion
        flag[Pdt] = false;
```

src/solver/hydro/daily/h2o_j_lisser_les_sur_turbines.cpp
```diff
@@ -32,94 +32,83 @@
...
         if (Turbine[Pdt] - TurbineCible[Pdt] > ZERO)
-            Flag[Pdt] = 1;
+            flag[Pdt] = 1;
```
flomnes:
```suggestion
            flag[Pdt] = true;
```

src/solver/hydro/daily/h2o_j_lisser_les_sur_turbines.cpp
```diff
@@ -32,94 +32,83 @@
...
+        if (flag[Pdt] == 1)
             SurTurbineARepartir += Turbine[Pdt] - TurbineCible[Pdt];
     }
 
-    for (Pdt = 0; Pdt < NbPdt; Pdt++)
-        Flag[Pdt] = 0;
-    for (Pdt = 0; Pdt < NbPdt; Pdt++)
+    for (int Pdt = 0; Pdt < NbPdt; Pdt++)
+        flag[Pdt] = 0;
+    for (int Pdt = 0; Pdt < NbPdt; Pdt++)
```
flomnes:
```suggestion
    for (int Pdt = 0; Pdt < NbPdt; Pdt++)
        flag[Pdt] = (TurbineMax[Pdt] - TurbineCible[Pdt] > ZERO);
```

src/solver/hydro/daily/h2o_j_lisser_les_sur_turbines.cpp
```diff
@@ -32,94 +32,83 @@
...
     }
 
-    NbCycles = 0;
+    int NbCycles = 0;
 BoucleDeRepartition:
 
-    Np = 0;
-    for (Pdt = 0; Pdt < NbPdt; Pdt++)
+    int Np = 0;
+    for (int Pdt = 0; Pdt < NbPdt; Pdt++)
```
flomnes:
```suggestion
    int Np = std::count(flag.begin(), flag.end(), true);
```

flomnes: You could even use `std::none_of`, since Np is only used to determine if one of `flag[i]` is `true`. We don't seem to care how many are true.

flomnes (changes requested):

flomnes:
@payetvin Please resolve conflicts

flomnes (changes requested):
- Use `std::unique_ptr` or `std::shared_ptr` instead of `new` / `delete`
- You forgot this `malloc`
daily2/h2o2_j_resoudre_le_probleme_lineaire.cpp
77:40:        Probleme = (PROBLEME_SIMPLEXE*)malloc(sizeof(PROBLEME_SIMPLEXE));

src/solver/hydro/management/daily.cpp
```diff
@@ -422,7 +422,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
                 break;
             }
 
-            H2O_J_Free(&problem);
+            H2O_J_Free(problem.get());
```
flomnes: Careful not to call `delete` or `free` on `problem.get()`, this would lead to a double-free since `~unique_ptr()` also de-allocates the memory.

src/solver/hydro/management/daily.cpp
```diff
@@ -389,28 +389,28 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
             uint firstDay = calendar_.months[simulationMonth].daysYear.first;
             uint endDay = firstDay + daysPerMonth;
 
-            DONNEES_MENSUELLES problem = H2O_J_Instanciation();
-            H2O_J_AjouterBruitAuCout(problem);
-            problem.NombreDeJoursDuMois = (int)daysPerMonth;
-            problem.TurbineDuMois = data.MOG[realmonth];
+            std::unique_ptr<DONNEES_MENSUELLES> problem = H2O_J_Instanciation();
```
flomnes:
```suggestion
            auto problem = H2O_J_Instanciation();
```

src/solver/hydro/monthly/h2o_m_free.cpp
```diff
@@ -40,69 +40,20 @@ extern "C"
...
-    free(DonneesAnnuelles->TurbineMin);
-    free(DonneesAnnuelles->TurbineCible);
-    free(DonneesAnnuelles->Apport);
-    free(DonneesAnnuelles->VolumeMin);
-    free(DonneesAnnuelles->VolumeMax);
-    free(DonneesAnnuelles->Turbine);
-    free(DonneesAnnuelles->Volume);
-    free(DonneesAnnuelles);
+    PROBLEME_SIMPLEXE* pb = static_cast<PROBLEME_SIMPLEXE*>(ProblemeHydraulique.Probleme);
+    delete pb;
```
flomnes: Please use `std::unique_ptr` here as well.

src/solver/hydro/monthly/h2o_m_resoudre_le_probleme_lineaire.cpp
```diff
@@ -46,43 +46,35 @@ extern "C"
...
+    PROBLEME_SPX* ProbSpx = (PROBLEME_SPX*)ProblemeHydraulique.ProblemeSpx[NumeroDeReservoir];
+    PROBLEME_SIMPLEXE* Probleme = (PROBLEME_SIMPLEXE*)ProblemeHydraulique.Probleme;
 
-    Probleme = (PROBLEME_SIMPLEXE*)ProblemeHydraulique->Probleme;
-    if (Probleme == NULL)
+    if (!Probleme)
     {
-        Probleme = (PROBLEME_SIMPLEXE*)malloc(sizeof(PROBLEME_SIMPLEXE));
-        if (Probleme == NULL)
+        Probleme = new(std::nothrow) PROBLEME_SIMPLEXE;
```
flomnes: Use `std::unique_ptr` here as well.

flomnes (changes requested):

src/solver/hydro/monthly/h2o_m_ajouter_bruit.cpp
```diff
@@ -9,49 +9,39 @@ constexpr double noiseAmplitude = 1e-3;
...
     noiseGenerator.reset(Constants::seed); // Arbitrary seed, hard-coded since we don't really want
                                            // the user to change it
-    const std::vector<const int*> monthlyVariables
-      = {CorrespondanceDesVariables->NumeroDeVariableVolume,
-         CorrespondanceDesVariables->NumeroDeVariableTurbine,
-         CorrespondanceDesVariables->NumeroDeVariableDepassementVolumeMax,
-         CorrespondanceDesVariables->NumeroDeVariableDepassementVolumeMin,
-         CorrespondanceDesVariables->NumeroDeVariableDEcartPositifAuTurbineCible,
-         CorrespondanceDesVariables->NumeroDeVariableDEcartNegatifAuTurbineCible};
+    const std::vector<std::vector<int>> monthlyVariables
```
flomnes: Careful, you're making a copy of all `std::vector<int>`. You may try `std::reference_wrapper` or just raw pointers.

src/solver/hydro/monthly/h2o_m_sys.h
```diff
@@ -35,5 +35,6 @@
 #include <string.h>
 #include <memory.h>
 #include <time.h>
+#include <vector>
```
flomnes: Try to include only necessary headers. You may include <vector> in the .cpp implementation file instead.

src/solver/hydro/monthly/h2o_m_ajouter_bruit.cpp
```diff
@@ -9,49 +9,39 @@ constexpr double noiseAmplitude = 1e-3;
...
-         CorrespondanceDesVariables->NumeroDeVariableDepassementVolumeMin,
-         CorrespondanceDesVariables->NumeroDeVariableDEcartPositifAuTurbineCible,
-         CorrespondanceDesVariables->NumeroDeVariableDEcartNegatifAuTurbineCible};
+    const std::vector<int*>& monthlyVariables
+      = {CorrespondanceDesVariables.NumeroDeVariableVolume.data(),
+         CorrespondanceDesVariables.NumeroDeVariableTurbine.data(),
+         CorrespondanceDesVariables.NumeroDeVariableDepassementVolumeMax.data(),
+         CorrespondanceDesVariables.NumeroDeVariableDepassementVolumeMin.data(),
+         CorrespondanceDesVariables.NumeroDeVariableDEcartPositifAuTurbineCible.data(),
+         CorrespondanceDesVariables.NumeroDeVariableDEcartNegatifAuTurbineCible.data()};
```
flomnes:
```suggestion
    const std::vector<std::vector<int>*> monthlyVariables
      = {&CorrespondanceDesVariables.NumeroDeVariableVolume,
         &CorrespondanceDesVariables.NumeroDeVariableTurbine,
         &CorrespondanceDesVariables.NumeroDeVariableDepassementVolumeMax,
         &CorrespondanceDesVariables.NumeroDeVariableDepassementVolumeMin,
         &CorrespondanceDesVariables.NumeroDeVariableDEcartPositifAuTurbineCible,
         &CorrespondanceDesVariables.NumeroDeVariableDEcartNegatifAuTurbineCible};
```

flomnes (approved):

PR #1627, opened 2023-09-16
refactor constraint construction

a-zakir:
follows #1617 #1607  #1234  #1605

this pr introduces the following new features:
- Remove ProblemHebdo and ProblemAREsoudre from the core lib of constraint construction i.e ConstraintBuilder to focus on the essentials 
- for each constraint we only provide the necessary data, example [ExchangeBalanceData](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/8bbd2e9b03a7fa56ff4dfa3cae6a7fc431d3ea57/src/solver/optimisation/constraints/ExchangeBalance.h#L5)
- gather constraints in groups (PminMaxDispatchableGenerationGroup is made up of PminDispatchable and PmaxDis...)
- the groups forms the Simple linear Problem Matrix (see [LinearProblemMatrix](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/8bbd2e9b03a7fa56ff4dfa3cae6a7fc431d3ea57/src/solver/optimisation/LinearProblemMatrix.h)) + the [LinearProblemMatrixStartUpCosts](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/8bbd2e9b03a7fa56ff4dfa3cae6a7fc431d3ea57/src/solver/optimisation/LinearProblemMatrixStartUpCosts.h)
- The [QuadraticProblemMatrix](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/8bbd2e9b03a7fa56ff4dfa3cae6a7fc431d3ea57/src/solver/optimisation/QuadraticProblemMatrix.h) is also refactored as well as [csr adq quadratic problem](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/8bbd2e9b03a7fa56ff4dfa3cae6a7fc431d3ea57/src/solver/optimisation/adequacy_patch_csr/csr_quadratic_problem.h)


src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,359 @@
...
+    {
+        return nativeOptimVar.NumeroDeVariableDefaillancePositive[v.index];
+    }
+    int operator()(const NegativeUnsuppliedEnergy& v) const
+    {
+        return nativeOptimVar.NumeroDeVariableDefaillanceNegative[v.index];
+    }
+
+private:
+    const CORRESPONDANCES_DES_VARIABLES& nativeOptimVar;
```
guilpier-code: When data member are **private**, we should mark them with a special trace, like a trailing underscore.
For example : **nativeOptimVar** -> **nativeOptimVar_**

src/solver/optimisation/constraint_builder.h
```diff
@@ -0,0 +1,359 @@
...
+                               HydProdUp,
+                               Pumping,
+                               HydroLevel,
+                               Overflow,
+                               FinalStorage,
+                               LayerStorage,
+                               PositiveUnsuppliedEnergy,
+                               NegativeUnsuppliedEnergy>;
+
+class ConstraintVisitor
```
guilpier-code: _Temporary comment, can change_

**ConstraintVisitor** : A visitor has a meaning for dev community : it's a design pattern. 
Do we have a real visitor here ?

src/solver/optimisation/new_constraint_builder.h
```diff
@@ -0,0 +1,400 @@
...
+#include <utility>
+
+namespace NewVariable
+{
+struct SingleIndex
+{
+    explicit SingleIndex(unsigned index) : index(index)
+    {
+    }
+    unsigned index;
```
sylvlecl: shouldn't we use a more specific and platform independent type such as `uint32_t` ?

src/solver/optimisation/new_constraint_builder.h
```diff
@@ -0,0 +1,400 @@
...
+{
+    using SingleIndex::SingleIndex;
+};
+
+struct LayerStorage
+{
+    LayerStorage(unsigned area, unsigned layer) : area(area), layer(layer)
+    {
+    }
+    unsigned area, layer;
```
sylvlecl: We still need to have a style guide, but I think usually it's preferred to have each member on its own line.

src/solver/optimisation/AreaBalance.h
```diff
@@ -0,0 +1,25 @@
...
+    const PALIERS_THERMIQUES& PaliersThermiquesDuPays;
+    const std::vector<::ShortTermStorage::AREA_INPUT>& ShortTermStorage;
+};
+
+class AreaBalance : public NewConstraint
+{
+public:
+    using NewConstraint::NewConstraint;
+
+    void add(int pdt, int pays, std::shared_ptr<AreaBalanceData> data);
```
sylvlecl: We must refrain, when not necessary, to use heap allocation (in particular for perf reason), and even more shared ownership (little perf but mostly for clarity sake).

Here the method will clearly not take ownership of the data, and not modify it --> it must be passed by const reference.
```cpp
void add(int pdt, int pays, const AreaBalanceData& data);
```
Same thing everywhere in other constraint types.

src/solver/optimisation/Group1.h
```diff
@@ -0,0 +1,24 @@
...
+class Group1 : public ConstraintGroup
+{
+public:
+    using ConstraintGroup::ConstraintGroup;
+
+    /*TODO Rename this*/
+    void Build() override;
+
+private:
+    std::shared_ptr<AreaBalanceData> GetAreaBalanceData(int pdt, uint32_t pays);
```
sylvlecl: For all those factory methods of plain data structures, it would be better to not use heap allocation, and use the plain structures instead: we should return a plain `AreaBalanceData` here, not a pointer.

src/solver/optimisation/ConstraintGroup.h
```diff
@@ -0,0 +1,20 @@
...
+    explicit ConstraintGroup(PROBLEME_HEBDO* problemeHebdo) :
+     problemeHebdo_(problemeHebdo),
+     builder_(NewGetConstraintBuilderFromProblemHebdo(problemeHebdo_))
+    {
+    }
+
+    /*TODO Rename this*/
+    virtual void Build() = 0;
+    PROBLEME_HEBDO* problemeHebdo_;
+    std::shared_ptr<NewConstraintBuilder> builder_;
```
sylvlecl: - why shared ownership, since it's clearly owned by this object ?
- why heap allocation in the first place ? ^^

sylvlecl (dismissed):
Nice design to add the plain data structures as input for building constraints.

On the technical side, the "shared_ptr police" has passed by :) we must not use so much heap allocation, and even less shared ownership.

src/solver/optimisation/new_constraint_builder.h
a-zakir: Thank you, I note your comment, but this code will soon be obsolete. #1607

src/solver/optimisation/AreaBalance.h
a-zakir: i'v tried refs data & faced huge amount of segfault-ish bug #1617

a-zakir: I'll try again

src/solver/optimisation/ConstraintGroup.h
a-zakir: you are right, in the first version of the code the "builder" wasn't a member object and was built for every constraint group.
Since they share the same base i couldn't agree more with you. To do

src/solver/optimisation/new_constraint_builder.h
a-zakir: deprecated following #1607

src/solver/optimisation/AreaBalance.h
a-zakir: modify just a little bit : `data->NumeroDeContrainteDesBilansPays[pays]=.....`

a-zakir:
For reviewers: all "data" objects consist of references or const references

src/solver/optimisation/AreaBalance.h
sylvlecl: Ok! maybe a non const ref is needed.

Ok for the segaults issues: well, I think if we come over that again, we can find out another solution than doing heap allocation, because this should not really change the lifetime of our objects here (not completely true, they might be moved and possibly copied in a few places, but I don't think it should be an issue). Easier said than done though :)

src/solver/optimisation/adequacy_patch_csr/csr_quadratic_problem.cpp
```diff
@@ -104,161 +74,47 @@ void CsrQuadraticProblem::setNodeBalanceConstraints(std::vector<double>& Pi, std
...
+                            hour,
+                            problemeHebdo_->IndexDebutIntercoOrigine,
+                            problemeHebdo_->IndexSuivantIntercoOrigine,
+                            problemeHebdo_->IndexDebutIntercoExtremite,
+                            problemeHebdo_->IndexSuivantIntercoExtremite,
+                            problemeHebdo_->adequacyPatchRuntimeData->originAreaMode,
+                            problemeHebdo_->adequacyPatchRuntimeData->extremityAreaMode,
+                            problemeHebdo_->PaysOrigineDeLInterconnexion,
+                            problemeHebdo_->PaysExtremiteDeLInterconnexion,
+                            hourlyCsrProblem_.numberOfConstraintCsrAreaBalance,
```
flomnes: This is dangerous. Two members could easily be swapped mistakenly. This kind of error is very hard to spot

Use the `{.a=a, .b=b}` syntax ?

src/solver/optimisation/constraints/PMaxDispatchableGeneration.cpp
```diff
@@ -1,38 +1,32 @@
 #include "PMaxDispatchableGeneration.h"
 
-void PMaxDispatchableGeneration::add(int pays,
-                                     int cluster,
-                                     int clusterIndex,
-                                     int pdt,
-                                     bool Simulation)
+void PMaxDispatchableGeneration::add(int pays, std::shared_ptr<StartUpCostsData> data)
```
flomnes:
```suggestion
void PMaxDispatchableGeneration::add(int pays, StartUpCostsData& data)
```

flomnes (commented):
Use references where possible, rather than `shared_ptr`

src/solver/optimisation/adequacy_patch_csr/csr_quadratic_problem.cpp
a-zakir: good

src/solver/optimisation/constraints/PMaxDispatchableGeneration.cpp
a-zakir: can't commit the suggestion as it, the call and data construction must be changed. I'll take care of that right now

src/solver/optimisation/ProblemMatrixEssential.h
```diff
@@ -0,0 +1,59 @@
...
+
+    void InitiliazeProblemAResoudreCounters()
+    {
+        auto& ProblemeAResoudre = problemeHebdo_->ProblemeAResoudre;
+        ProblemeAResoudre->NombreDeContraintes = 0;
+        ProblemeAResoudre->NombreDeTermesDansLaMatriceDesContraintes = 0;
+    }
+
+    PROBLEME_HEBDO* problemeHebdo_;
+    std::vector<ConstraintGroup*> constraintgroups_;
```
guilpier-code, line 41: This data members should not be **public**, because it can be changed from outside the class too easily.
Why not make them **protected** at least ?

guilpier-code:
**Question** for myself (as a remainder, to be explored) : this PR is supposed to be a refactoring, but we added approximately 1200 lines. 
Why is that ?

guilpier-code:
**Questions** : 
- What is the goal of this PR ?
- Why is it so big ? Shouldn't it be split into smaller PR, so that the required changed would have been asked earlier, and thus less numerous, and the review less painful ?

src/solver/optimisation/LinearProblemMatrix.h
```diff
@@ -0,0 +1,64 @@
...
+#include "constraints/HydraulicSmoothingGroup.h"
+#include "constraints/MinMaxHydroPowerGroup.h"
+#include "constraints/MaxPumpingGroup.h"
+#include "constraints/AreaHydroLevelGroup.h"
+#include "constraints/FinalStockGroup.h"
+
+#include <antares/study/study.h>
+
+using namespace Antares::Data;
+class LinearProblemMatrix : public ProblemMatrixEssential
```
guilpier-code, line 45: We should call this class **LP_ConstraintGroupList** (see review of **ProblemMatrixEssential.h**)

src/solver/optimisation/LinearProblemMatrix.h
```diff
@@ -0,0 +1,64 @@
...
+{
+public:
+    explicit LinearProblemMatrix(PROBLEME_HEBDO* problemeHebdo, Solver::IResultWriter& writer);
+
+    void Run() override;
```
guilpier-code, line 50: As a consequence of **ProblemMatrixEssential.h**'s review, we don't need **run()** here.

src/solver/optimisation/LinearProblemMatrix.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (problemeHebdo_->ExportStructure && problemeHebdo_->firstWeekOfSimulation)
+    {
+        OPT_ExportInterco(writer_, problemeHebdo_);
+        OPT_ExportAreaName(writer_, problemeHebdo_->NomsDesPays);
+    }
+}
+
+void LinearProblemMatrix::Run()
+{
+    InitiliazeProblemAResoudreCounters();
```
guilpier-code: About `InitiliazeProblemAResoudreCounters();` : 
- We should not mix french and english. We should choose english instead.
-  Initiliaze (does not exist) --> Initialize.
- We are initialize the constraints matrix, and we are (or should be) in the class **ConstraintsMatrix**, so so it's enough if we rename this function **initialize()**
- **But ultimately** we should do this initialization in the construction, otherwise what are the constructor for, right ?

src/solver/optimisation/ProblemMatrixEssential.h
```diff
@@ -0,0 +1,59 @@
...
+*/
+#pragma once
+#include "opt_structure_probleme_a_resoudre.h"
+#include "sim_structure_probleme_economique.h"
+#include "constraints/ConstraintGroup.h"
+
+#include <antares/study/study.h>
+
+using namespace Antares::Data;
+class ProblemMatrixEssential
```
guilpier-code, line 32: Eventually, what does **ProblemMatrixEssential** do ?

It holds a collections of constraint groups.
It triggers the build of constraints contained in these groups.
And that's it.
So **ProblemMatrixEssential** should be renamed into **ConstraintGroupList**.

Any child class should only fill the collection of constraint groups, and should do it in its constructor.
And that's it. No need to have any other methods or data member, normally.
For an instance of such child, calling **run()** would mean calling the parent's **run()**.

src/solver/optimisation/LinearProblemMatrix.cpp
```diff
@@ -0,0 +1,81 @@
...
+void LinearProblemMatrix::ExportStructures()
+{
+    if (problemeHebdo_->ExportStructure && problemeHebdo_->firstWeekOfSimulation)
+    {
+        OPT_ExportInterco(writer_, problemeHebdo_);
+        OPT_ExportAreaName(writer_, problemeHebdo_->NomsDesPays);
+    }
+}
+
+void LinearProblemMatrix::Run()
```
guilpier-code, line 61: **Run()** should be removed, as already said in parent class's review.

src/solver/optimisation/ProblemMatrixEssential.h
guilpier-code, line 32: Maybe associated **.cpp** file ? (but not a priority)
We would have the interface separated from the implementation, more clear.

src/solver/optimisation/AreaBalance.h
a-zakir: @sylvlecl shared what? :relaxed:

src/solver/optimisation/ProblemMatrixEssential.h
```diff
@@ -0,0 +1,59 @@
...
+{
+public:
+    explicit ProblemMatrixEssential(PROBLEME_HEBDO* problemeHebdo) : problemeHebdo_(problemeHebdo)
+    {
+    }
+
+    virtual void Run()
```
guilpier-code: **Run()** : should have a clearer name.
It builds constraints in groups, but some common constraints (common to linear and quadratic problem it seems). So how would we call these common constraints ?
For example, this function could be renamed  into **buildConstrainytsByGroup**.

About the notion of **group** : 
We should pay attention that **binding constraints** can already belong to a **group**, and in that case, the BCs of the same **group** share the same TS number (for their RHS).
But here, the meaning of a **group** is different. 
So, to avoid confusion, maybe we should replace the notion of **group** with another one.
If I understand correctly, **Current groups** gather constraints that cannot be separated because they are tight by the order we fill the constraints matrix with, not because they have something fundamental in common.
Maybe **filling order chunk** should be better.
We should talk about that altogether.

src/solver/optimisation/QuadraticProblemMatrix.h
```diff
@@ -0,0 +1,25 @@
...
+#include "opt_structure_probleme_a_resoudre.h"
+
+#include "../simulation/simulation.h"
+#include "../simulation/sim_structure_donnees.h"
+#include "../simulation/sim_extern_variables_globales.h"
+#include "opt_fonctions.h"
+#include "constraints/ExchangeBalanceGroup.h"
+#include "ProblemMatrixEssential.h"
+
+class QuadraticProblemMatrix : public ProblemMatrixEssential
```
guilpier-code, line 12: **exchangeBalanceGroup_** is used only as a transition state during the conversion of the weekly problem into a constraint group, right ? It hasn't any other function.
We should find a way to remove it as a data member, the real data member is the collection of constraint groups. 

**A possible way to do it** : 
Eventually, what does **ProblemMatrixEssential** do ?
- It holds a collections of constraint groups
- It triggers the build of constraints contained in these groups.

And that's it.
So **ProblemMatrixEssential** should be renamed into **ConstraintGroupList**.
Please see comment for class **ProblemMatrixEssential**.

From there,  the responsibility of **QuadraticProblemMatrix** is to add some constraint groups to  this list held by its parent (here only one). And that's all. No more things to do.

So, here : 
- in the constructor, add a constraint group the the list held by the parent class
- The run method is useless, it's the parent class's **run()** method that will be called.
- Any data member is useless.

src/solver/optimisation/LinearProblemMatrix.cpp
```diff
@@ -0,0 +1,83 @@
...
+                         &bindingConstraintDayGroup_,
+                         &bindingConstraintWeekGroup_,
+                         &hydroPowerGroup_,
+                         &hydraulicSmoothingGroup_,
+                         &minMaxHydroPowerGroup_,
+                         &maxPumpingGroup_,
+                         &areaHydroLevelGroup_,
+                         &finalStockGroup_};
+}
+void LinearProblemMatrix::ExportStructures()
```
guilpier-code: About **LinearProblemMatrix::ExportStructures()** : 
A constraint matrix class should **not**  export (to any target [txt or zip files]) the data that is used to build it (the weekly problem) for 2 reasons :
- A **business** (**métier** in french) class should never write or read the data it needs, some other class should do it. Doing so, we separate the concerns.
- And even more : a class shouldn't have to handle the data that is processed to retrieve the class internal data. No way :-).

We should remove this function from this class, and call this call the Export free functions in **opt_optimisation_lineaire.cpp**

src/solver/optimisation/LinearProblemMatrix.h
```diff
@@ -0,0 +1,67 @@
...
+public:
+    explicit LinearProblemMatrix(PROBLEME_HEBDO* problemeHebdo,
+                                 Solver::IResultWriter& writer,
+                                 ConstraintBuilder& builder);
+
+    void Run() override;
+    void ExportStructures();
+
+private:
+    Solver::IResultWriter& writer_;
```
guilpier-code: as already said earlier in the review of the associated .cpp file, **writer_** should be removed

src/solver/optimisation/LinearProblemMatrix.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+using namespace Antares::Data;
+
+LinearProblemMatrix::LinearProblemMatrix(PROBLEME_HEBDO* problemeHebdo,
+                                         Solver::IResultWriter& writer,
+                                         ConstraintBuilder& builder) :
+ ProblemMatrixEssential(problemeHebdo),
+ builder_(builder),
+ writer_(writer),
+ group1_(problemeHebdo, builder),
```
guilpier-code, line 40: What are all the **groups** as **data member** for ?
Except in the constructor, they are not used anywhere. 
So they seem useless.
In the constructor, we could instead do :
```C++
constraintgroups_.push_back(std::make_unique<Group1>(problemeHebdo, builder));
constraintgroups_.push_back(std::make_unique<BindingConstraintDayGroup>(problemeHebdo, builder));
...
```
and we could remove both the **associated data members** and the **initialization list**.
To ensure polymorphism, **constraintgroups_** as a vector of unique pointers (see above) seems the best choice.
The class would be significantly shorter.

src/solver/optimisation/LinearProblemMatrix.h
```diff
@@ -0,0 +1,67 @@
...
+    ConstraintBuilder& builder_;
+    Group1 group1_;
```
guilpier-code, line 54: As said (in more details) in **.cpp** review, we should remove all the groups as data members.
As said above, we should remove the writer as well.

About the **builder_**, I'm still not sure, but what does the builder do ?
- first, for me, it should be called **constraintBuilder_**, more precise about what it does.
- then, does it have to be a data member of a high level class that deals with constraint group, and not directly with constraint ? We could possibly need it when building a constraint, but here ?
- Another question : even if we build constraint with the **builder**  (as argument of the constructor) and make it an internal data member, we will probably use it to actually build the constraint. Isn't the constraint able to build itself ? I mean, could the piece of code that build the constraint be moved to the constraint itself ? And is it suitable ?
  The final results would be to remove totally the class  **ConstraintBuilder**, and spread it across the **constraint classes**.

src/solver/optimisation/LinearProblemMatrix.cpp
```diff
@@ -0,0 +1,83 @@
...
+        OPT_ExportInterco(writer_, problemeHebdo_);
+        OPT_ExportAreaName(writer_, problemeHebdo_->NomsDesPays);
+    }
+}
+
+void LinearProblemMatrix::Run()
+{
+    InitiliazeProblemAResoudreCounters();
+
+    ProblemMatrixEssential::Run();
```
guilpier-code, line 65: Please read the review of **ProblemMatrixEssential.h** first.

About the piece of code : 
```C++
1. ProblemMatrixEssential::Run();
2. if (problemeHebdo_->OptimisationAvecCoutsDeDemarrage)
3. {
4.    LinearProblemMatrixStartUpCosts(problemeHebdo_, false, builder_).Run();
5. }
```
This part is unclear to me : when we think about what this code do, after a while, we understand that it triggers the building of constraints, through groups of constraints.
Eventually, it's only a list of constraint groups that are run, we should not have a need for 2 classes and 2 **run()** methods.
Class **LinearProblemMatrix** should be more simple, it could be (according to earlier comments, I remove writer and the builder as data member) : 
```C++
LinearProblemMatrix::LinearProblemMatrix(PROBLEME_HEBDO* problemeHebdo,
                                         ConstraintBuilder& builder) :
     ProblemMatrixEssential(problemeHebdo),
{
    constraintgroups_.push_back(/*group 1 creation here*/);
    constraintgroups_.push_back(/*group 2 creation here */);
    ....
    if (problemeHebdo->OptimisationAvecCoutsDeDemarrage)
    {
         constraintgroups_.push_back(/*group n + 1 creation here*/);
         constraintgroups_.push_back(/*group n + 2 creation here*/);
         ....
    }
}
```
And that's it. Nothing more should be needed.
It's also what is suggested in the review of **ProblemMatrixEssential.h**.

src/solver/optimisation/ProblemMatrixEssential.h
```diff
@@ -0,0 +1,59 @@
+/*
```
guilpier-code, line 1: **Question** : do we have to add every time this long and cumbersome header ?
We could consider not supplying all files with this header.

src/solver/optimisation/ProblemMatrixEssential.h
```diff
@@ -0,0 +1,59 @@
...
+
+    virtual void Run()
+    {
+        for (auto& group : constraintgroups_)
+        {
+            group->Build();
+        }
+    }
+
+    void InitiliazeProblemAResoudreCounters()
```
guilpier-code: We should remove **InitiliazeProblemAResoudreCounters()** and move its body to the constructor.
Initializing is the constructors job, right ?

src/solver/optimisation/ProblemMatrixEssential.h
```diff
@@ -0,0 +1,59 @@
...
+    {
+        auto& ProblemeAResoudre = problemeHebdo_->ProblemeAResoudre;
+        ProblemeAResoudre->NombreDeContraintes = 0;
+        ProblemeAResoudre->NombreDeTermesDansLaMatriceDesContraintes = 0;
+    }
+
+    PROBLEME_HEBDO* problemeHebdo_;
```
guilpier-code, line 40: Do we need the **problemeHebdo_** as a data member ? Not needed, probably not in child classes as well.

src/solver/optimisation/LinearProblemMatrixStartUpCosts.h
```diff
@@ -0,0 +1,58 @@
...
+#include "constraints/ConsistenceNumberOfDispatchableUnitsGroup.h"
+#include "constraints/NbUnitsOutageLessThanNbUnitsStopGroup.h"
+#include "constraints/NbDispUnitsMinBoundSinceMinUpTimeGroup.h"
+#include "constraints/MinDownTimeGroup.h"
+
+#include <antares/study/study.h>
+
+using namespace Antares::Data;
+
+class LinearProblemMatrixStartUpCosts : public ProblemMatrixEssential
```
guilpier-code, line 44: Please see review of **ProblemMatrixEssential.h** and especially **LinearProblemMatrix.cpp**, and remove this class.

src/solver/optimisation/LinearProblemMatrixStartUpCosts.cpp
```diff
@@ -0,0 +1,47 @@
...
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+#pragma once
+#include "LinearProblemMatrixStartUpCosts.h"
+using namespace Antares::Data;
+
+LinearProblemMatrixStartUpCosts::LinearProblemMatrixStartUpCosts(PROBLEME_HEBDO* problemeHebdo,
```
guilpier-code, line 31: See the review of the associated header file and remove this class.

guilpier-code (changes requested):

src/solver/optimisation/ProblemMatrixEssential.h
guilpier-code: `std::vector<ConstraintGroup*> constraintgroups_;` 
If we remove constraint groups as data member from child classes, I can't see how to keep **raw pointers** in this **vector**. **unique_ptr** seems to be the best.

guilpier-code (changes requested):

src/solver/optimisation/LinearProblemMatrix.cpp
a-zakir: @guilpier-code  to this and  about transforming groups into unique_ptr see @sylvlecl recommendation about heap allocation. constraintgroups_ is just an artifact to loop over groups of constraints

src/solver/optimisation/constraints/Group1.h
```diff
@@ -0,0 +1,22 @@
...
+#include "ShortTermStorageLevel.h"
+#include "FlowDissociation.h"
+#include "BindingConstraintHour.h"
+
+class Group1 : public ConstraintGroup
+{
+public:
+    using ConstraintGroup::ConstraintGroup;
+
+        void Build() override;
```
guilpier-code: **void Build() override;**  : take care of indentation.
We could rename it into **buildConstraints()**, more clear.

src/solver/optimisation/constraints/Group1.h
```diff
@@ -0,0 +1,22 @@
...
+
+private:
```
guilpier-code, line 16: As data members, Instead of data to build **constraints** with, we expect the **constraints**. After all, we are in a group of **constraints**, right ?
So naturally, we would expect a **vector** of **constraints** (probably held in the parent class **ConstraintGroup**). It would be natural, it would look like the business rules we're in : when we talk about this, we say "groups of constraints".

But there is the problem of the **order in which the constraints are created** in the matrix. It prevents us from having a simple vector of constraints here, and it prevents us write a simple loop (as we would expect) : 
```C++
Group1::build()
{
    for (auto& constraint : constraintList_)
    {
          constraintList_->build(); // add(...) was replaced with build() here.
    }
}
```
I guess that dealing with the **constraints order** will be addressed later, we can still try to improve the code as it is, keeping the order issue unsolved for now. 

So I would recommend to : 
- Instead of data for building constraints (as **Group1** data members), as we **cannot** have a vector of constraints, let's have constraints instances :  **AreaBalance**, **FictitiousLoad**, ...
- These constraints should be built in the **Group1**'s constructor, by passing to each constraint constructor the weekly problem (problem hebdo). So constraints should not be built in `Group1::Build()`, as done currently. The way it's done now is too complicated and unnatural.

src/solver/optimisation/constraints/ConstraintGroup.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "sim_structure_probleme_economique.h"
+#include "../opt_structure_probleme_a_resoudre.h"
+#include "constraint_builder_utils.h"
+
+class ConstraintGroup
```
guilpier-code, line 7: Here I continue developing the idea of introducing a **Constraint** class , first mentioned in review of **Group1.h**.

src/solver/optimisation/constraints/ConstraintGroup.h
```diff
@@ -0,0 +1,18 @@
...
+{
+public:
+    explicit ConstraintGroup(PROBLEME_HEBDO* problemeHebdo, ConstraintBuilder& builder) :
+     problemeHebdo_(problemeHebdo), builder_(builder)
+    {
+    }
+
+    virtual void Build() = 0;
```
guilpier-code: `virtual void Build() = 0; ` ==> `virtual void buildConstraints() = 0; ` ?

guilpier-code (changes requested):

src/solver/optimisation/constraints/ConstraintGroup.h
```diff
@@ -0,0 +1,18 @@
...
+    PROBLEME_HEBDO* problemeHebdo_;
+    ConstraintBuilder& builder_;
```
guilpier-code, line 17: `ConstraintBuilder& builder_;` : 
**builder_** stored only used to build constraints, but it should not be here, and for 2 reasons :
- As said in **Group1.h**'s review, it should be immediately passed from the group's constructor (any group: **Group1**, ....) to the constructor of constraints the group contains. It shouldn't need to be stored here.
- But **mostly**, it should probably not exist : class **ConstraintBuilder** is too big, and contains logic that constraints need to build themselves. **ConstraintBuilder** should be spread across constraint classes.

src/solver/optimisation/constraints/ConstraintGroup.h
```diff
@@ -0,0 +1,18 @@
...
+class ConstraintGroup
+{
+public:
+    explicit ConstraintGroup(PROBLEME_HEBDO* problemeHebdo, ConstraintBuilder& builder) :
+     problemeHebdo_(problemeHebdo), builder_(builder)
+    {
+    }
+
+    virtual void Build() = 0;
+    PROBLEME_HEBDO* problemeHebdo_;
```
guilpier-code, line 16: `PROBLEME_HEBDO* problemeHebdo_;`
Eventually we also should remove it from class **ConstraintGroup**, but it's trickier than for **builder_**, because it's used in the constraint ordering loops (see **Group1::Build()** for example). We should try to find a way to extract these loops into classes.
So we'll leave it for now.

src/solver/optimisation/constraints/Group1.h
a-zakir: 1- Ok
2- constructed in Group1 ctor -->ok, but not BUILD as you mentioned we must be careful with order

src/solver/optimisation/constraints/ConstraintGroup.h
a-zakir: ok to rename it : **B**uildConstraints() = 0;

a-zakir: #TODO

src/solver/optimisation/LinearProblemMatrixStartUpCosts.h
a-zakir: This class is also inherited by the QuadraticMatrixProblem, be specific pls

src/solver/optimisation/ProblemMatrixEssential.h
a-zakir: possible if we move the initialization of pbmAResoudre counters...etc on hold

src/solver/optimisation/constraints/Group1.cpp
```diff
@@ -0,0 +1,86 @@
+#include "Group1.h"
+AreaBalanceData Group1::GetAreaBalanceData()
```
guilpier-code, line 2: All these **Get...()** should be removed from class **Group1** (see **Group1.h** 's review).

src/solver/optimisation/constraints/Group1.cpp
```diff
@@ -0,0 +1,86 @@
...
+    FictitiousLoad fictitiousLoad(builder_, fictitiousLoadData);
+
+    auto shortTermStorageLevelData = GetShortTermStorageLevelData();
+    ShortTermStorageLevel shortTermStorageLevel(builder_, shortTermStorageLevelData);
+
+    auto flowDissociationData = GetFlowDissociationData();
+    FlowDissociation flowDissociation(builder_, flowDissociationData);
+
+    auto bindingConstraintHourData = GetBindingConstraintHourData();
+    BindingConstraintHour bindingConstraintHour(builder_, bindingConstraintHourData);
```
guilpier-code, line 60:
```c++
auto areaBalanceData = GetAreaBalanceData();
AreaBalance areaBalance(builder_, areaBalanceData);
...
```
As said in **Group1.h** 's review, this block (constraints construction) should be removed.

src/solver/optimisation/constraints/Group1.h
```diff
@@ -0,0 +1,22 @@
...
+    AreaBalanceData GetAreaBalanceData();
```
guilpier-code, line 17: As a consequence of a previous comment of the current file, **AreaBalanceData** seems useless, as its content should be stored  (as data member) in a constraint class (**AreaBalance**).
Same for other data classes here.

guilpier-code (changes requested):

src/solver/optimisation/constraints/AreaBalance.h
```diff
@@ -1,18 +1,37 @@
 #pragma once
 #include "ConstraintBuilder.h"
 
+struct AreaBalanceData
```
guilpier-code, line 4: Struct **AreaBalanceData** : 
Do we need to encapsulate these pieces of data in a **struct** ?
The consequence is that we make the code a bit more difficult to read : the prefix `data.<something>` is always needed, though it could not.

src/solver/optimisation/constraints/AreaBalance.h
```diff
@@ -1,18 +1,37 @@
...
+     ConstraintFactory(builder), data(data)
+    {
+    }
 
     /*!
      * @brief Add variables to the constraint and update constraints Matrix
      * @param pdt : timestep
      * @param pays : area
      */
+
```
guilpier-code, line 32: `void add(int pdt, int pays);` : 
In the high code level, we use **build()**, but we use **add()**. 
Wouldn't we make the code clearer by renaming this method into **build()** ?

src/solver/optimisation/constraints/AreaBalance.h
```diff
@@ -1,18 +1,37 @@
...
+    const std::vector<PALIERS_THERMIQUES>& PaliersThermiquesDuPays;
+    const std::vector<::ShortTermStorage::AREA_INPUT>& ShortTermStorage;
+};
+
 /*!
  * represent 'Area Balance' constraint type
  */
-class AreaBalance : private ConstraintFactory
+
+class AreaBalance : public ConstraintFactory
```
guilpier-code, line 19: `class AreaBalance : public ConstraintFactory` : 
This inheritance from **ConstraintFactory** seems useless : it only exists so that all constraints classes hold an instance of the builder (of type **ConstraintBuilder**).
But, as already said somewhere else, class **ConstraintBuilder** is too large and does things that should be done by the constraints themselves. **ConstraintBuilder** logic should be spread across constraints classes.

src/solver/optimisation/constraints/ConstraintGroup.h
guilpier-code, line 16: The 2 data members should be **protected**, not **public** : otherwise, we allow them to be modified from outside the class hierarchy, which makes no sense, as these data members are rightly meant to be used internally to the class hierarchy.

src/solver/optimisation/constraints/AreaBalance.h
```diff
@@ -1,18 +1,37 @@
...
 {
 public:
-    using ConstraintFactory::ConstraintFactory;
+    AreaBalance(ConstraintBuilder& builder, AreaBalanceData& data) :
```
guilpier-code, line 22: As a consequence, **AreaBalance**'s constructor should only take the weekly problem as argument.

src/solver/optimisation/constraints/AreaBalance.cpp
```diff
@@ -7,45 +7,43 @@ static void shortTermStorageBalance(const ::ShortTermStorage::AREA_INPUT& shortT
...
     }
 }
 
 void AreaBalance::add(int pdt, int pays)
 {
-    CORRESPONDANCES_DES_CONTRAINTES& CorrespondanceCntNativesCntOptim
-      = problemeHebdo->CorrespondanceCntNativesCntOptim[pdt];
-    CorrespondanceCntNativesCntOptim.NumeroDeContrainteDesBilansPays[pays]
-      = problemeHebdo->ProblemeAResoudre->NombreDeContraintes;
+    data.CorrespondanceCntNativesCntOptim[pdt].NumeroDeContrainteDesBilansPays[pays]
```
guilpier-code, line 16: About **nombreDeContraintes** : 
Problems with this variable  : 
- It belongs to an external object (**builder**) that should disappear (does too many things) and that should be spread across constraints classes
- It is updated (see `nombreDeContraintes++`) in several places : in the **ConstraintBuilder** class itself, but also in some of the constraint classes. Why isn't it updated in one place only ?

To be thought about and discussed.

src/solver/optimisation/constraints/ConstraintBuilder.h
```diff
@@ -331,10 +353,12 @@ class ConstraintBuilder
         return nombreDeTermes_;
     }
 
+    ConstraintBuilderData& data;
+
 private:
-    PROBLEME_HEBDO& problemeHebdo;
-    PROBLEME_ANTARES_A_RESOUDRE& problemeAResoudre;
-    const std::vector<CORRESPONDANCES_DES_VARIABLES>& varNative;
+    void OPT_ChargerLaContrainteDansLaMatriceDesContraintes();
```
guilpier-code, line 324: I talk about it here because I can't see a better place too say that :
**src/solver/optimisation/opt_construction_matrice_des_contraintes_outils.cpp** : this file contains a function which does is not used. We should remove this file.
We should remove the function declaration too.

src/solver/optimisation/constraints/ConstraintBuilder.cpp
```diff
@@ -2,21 +2,17 @@
...
+        OPT_ChargerLaContrainteDansLaMatriceDesContraintes();
     }
     nombreDeTermes_ = 0;
 }
 
 int ConstraintBuilder::GetShiftedTimeStep(int offset, int delta) const
 {
     int pdt = hourInWeek_ + offset;
-    const int nbTimeSteps = problemeHebdo.NombreDePasDeTempsPourUneOptimisation;
+    const int nbTimeSteps = data.NombreDePasDeTempsPourUneOptimisation;
```
guilpier-code, line 15: Sorry, i don't quite understand this **if** statement.
If we wanted to make clearer, couldn't we just remove it and leave only what's inside the **if** ?

src/solver/optimisation/constraints/ConstraintBuilder.cpp
```diff
@@ -34,211 +30,249 @@ int ConstraintBuilder::GetShiftedTimeStep(int offset, int delta) const
...
 {
     AddVariable(GetVariableManager(offset, delta).HydProd(index), coeff);
     return *this;
 }
 
 ConstraintBuilder& ConstraintBuilder::HydProdDown(unsigned int index,
-                                                  double coeff,
-                                                  int offset,
-                                                  int delta)
+                                                        double coeff,
```
guilpier-code: About **ConstraintBuilder::HydProdDown(...)**
Many functions (like **HydProdDown**) are never defined as taking 4 arguments, but are never called with 4 arguments.
For me, there is a design problem here.

guilpier-code (changes requested):

src/solver/optimisation/opt_rename_problem.h
```diff
@@ -126,6 +127,12 @@ class ConstraintNamer : public Namer
     void BindingConstraintHour(unsigned int constraint, const std::string& name);
     void BindingConstraintDay(unsigned int constraint, const std::string& name);
     void BindingConstraintWeek(unsigned int constraint, const std::string& name);
+    void CsrFlowDissociation(unsigned int constraint,
```
guilpier-code, line 130: class **Namer** : 
all data members are **public** (changeable from outside), a though they should be **protected**

guilpier-code (changes requested):

src/solver/optimisation/constraints/AreaBalance.cpp
```diff
@@ -7,45 +7,43 @@ static void shortTermStorageBalance(const ::ShortTermStorage::AREA_INPUT& shortT
...
+      = builder.data.nombreDeContraintes;
 
-        ConstraintNamer namer(problemeHebdo->ProblemeAResoudre->NomDesContraintes);
-        namer.UpdateTimeStep(problemeHebdo->weekInTheYear * 168 + pdt);
-        namer.UpdateArea(problemeHebdo->NomsDesPays[pays]);
-        namer.AreaBalance(problemeHebdo->ProblemeAResoudre->NombreDeContraintes);
+    ConstraintNamer namer(builder.data.NomDesContraintes);
```
guilpier-code, line 19: Here we also build the name of constraints for MPS format.
So the current method does (at least) 2 things : it builds a constraint and it builds a name for MPS format.
Normally, a constraint class should only build a constraint, and extracting from a constraint a line for an MPS file should be done by another class.
I can't see a way to do that for now.

src/solver/optimisation/constraints/AreaBalance.cpp
guilpier-code, line 16: About **AreaBalance::add** in general : 
Looking at this method, I see that it is too long. Indeed : 
- it contains details we don't want at this level. Example : do we want to see details such has the `while (interco >= 0)` here ? Or shouldn't these details be hidden in a deeper level ?
- It contains things that shouldn't be here because not strictly related to the class responsibility. Example : renaming constraints for MPS format.

It seems that the real (and only) mission of this method is : build some constraints.
So **ideally**, we should only have : 
```C++
void AreaBalance::add(int pdt, int pays)
{
    // we suppose that class AreaBalance holds a list of variables : variableList_
    for (auto & variable : variableList_)
    {
        variable->add(pdt, pays);
        // or : add(variable); // Take your pick 
    }
    this->equalTo();
    this->storeInConstraintsMatrix();
}
``` 
I think we should ask ourselves : what prevents us from having only this simple code ?

In the meantime, and assuming we keep the **builder**, we could simply have : 
```C++
void AreaBalance::add(int pdt, int pays)
{
    builder.NTCDirect(pdt, pays);
    builder.thermalProduction(pdt, pays);
    builder.HydProd(pdt, pays)
    builder.Pumping(pdt, pays)
    builder.PositiveUnsuppliedEnergy(pdt, pays)
    builder.NegativeUnsuppliedEnergy(pdt, pays);
    builder.ShortTermStorageBalance(pdt, pays);

    builder.equalTo();
    builder.build(); // Meaning we store the constraint in matrix
}
```
It would be more readable : nothing is unexpected, and we stick to the class / method mission only.

src/solver/optimisation/constraints/AreaBalance.cpp
```diff
@@ -7,45 +7,43 @@ static void shortTermStorageBalance(const ::ShortTermStorage::AREA_INPUT& shortT
...
+    interco = data.IndexDebutIntercoExtremite[pays];
     while (interco >= 0)
     {
         builder.NTCDirect(interco, -1.0);
-        interco = problemeHebdo->IndexSuivantIntercoExtremite[interco];
+        interco = data.IndexSuivantIntercoExtremite[interco];
     }
 
-    exportPaliers(*problemeHebdo, builder, pays);
+    ExportPaliers(data.PaliersThermiquesDuPays[pays], builder);
```
guilpier-code, line 40: **ExportPaliers** : I guess this about adding thermal production variables to the current constraint.
Why not have : `builder.thermalProduction(pdt, pays);` instead ?

src/solver/optimisation/constraints/AreaBalance.cpp
```diff
@@ -7,45 +7,43 @@ static void shortTermStorageBalance(const ::ShortTermStorage::AREA_INPUT& shortT
...
     builder.HydProd(pays, -1.0)
       .Pumping(pays, 1.0)
       .PositiveUnsuppliedEnergy(pays, -1.0)
       .NegativeUnsuppliedEnergy(pays, 1.0);
 
-    shortTermStorageBalance(problemeHebdo->ShortTermStorage[pays], builder);
+    shortTermStorageBalance(data.ShortTermStorage[pays], builder);
```
guilpier-code, line 46: Same remark for short term storage than for thermal cluster : 
`builder.ShortTermStorageBalance(pdt, pays);` ?

guilpier-code (dismissed):

src/solver/optimisation/constraints/AreaBalance.cpp
a-zakir: thank you, that's @flomnes idea! To move the constraintnamer to a suitable place.
#TODO :fire:

src/solver/optimisation/constraints/AreaBalance.cpp
a-zakir: It just slipped my mind 
#TODO

src/solver/optimisation/constraints/AreaBalance.cpp
a-zakir: /2?

src/solver/optimisation/opt_rename_problem.h
a-zakir: it is a hell of review
#TODO :smiling_imp:

src/solver/optimisation/constraints/ConstraintBuilder.cpp
a-zakir: Ah! this is from the old way, old not gold

a-zakir: #TODO

src/solver/optimisation/constraints/ConstraintBuilder.cpp
a-zakir: if we have an offset...................................

src/solver/optimisation/constraints/AreaBalance.cpp
a-zakir: STARTUPCOSTS DOES STRANGE THINGS

src/solver/optimisation/constraints/AreaBalance.h
a-zakir: hein weekly?i don't know what it is :running:

src/solver/optimisation/constraints/ConstraintGroup.h
a-zakir: public or private only

src/solver/optimisation/constraints/AreaBalance.h
a-zakir: feel free to suggest an alternative.

src/solver/optimisation/constraints/AreaBalance.h
a-zakir: Ok

src/solver/optimisation/constraints/AreaBalance.h
a-zakir: please have a cup coffee with @flomnes

src/solver/optimisation/constraints/Group1.h
a-zakir: et tout ce monde dans la methode add?
` std::vector<CORRESPONDANCES_DES_CONTRAINTES>& CorrespondanceCntNativesCntOptim;
    const std::vector<int>& IndexDebutIntercoOrigine;
    const std::vector<int>& IndexSuivantIntercoOrigine;
    const std::vector<int>& IndexDebutIntercoExtremite;
    const std::vector<int>& IndexSuivantIntercoExtremite;
    const std::vector<PALIERS_THERMIQUES>& PaliersThermiquesDuPays;
    const std::vector<::ShortTermStorage::AREA_INPUT>& ShortTermStorage;`

src/solver/optimisation/constraints/Group1.h
a-zakir: done

src/solver/optimisation/constraints/Group1.h
a-zakir: finally your idea is to move Constraints class instantiation inside the ctor body and later you suggest to remove the **GetData** methods, at the end we'll have a big ctor...

src/solver/optimisation/constraints/Group1.cpp
a-zakir: group of constraints are not returned in anyway for any reason, i don't find why we would do that

src/solver/optimisation/constraints/AreaBalance.h
a-zakir: you've got a point! this refactor is about freeing constraints builder from the "optimization" lexicon

src/solver/optimisation/constraints/AreaBalance.cpp
a-zakir: we rely on "nombreDeContraintes" to build the name of a constraint. this Pr is large enough let's do it in another one

src/solver/optimisation/ProblemMatrixEssential.h
a-zakir: you're right group notion in this Pr is different in those defined elsewhere

src/solver/optimisation/constraints/AreaBalance.cpp
a-zakir: ExportPaliers is used twice by 'AreaBalance' and 'FictitiousLoad'
i think it's ok to keep it in a function

flomnes (approved):

PR #1631, opened 2023-09-18
Bug on renewable clusters

### LINKED ISSUE #1630: "Other RES" : wrong output variable

guilpier-code:
The problem is that the **group** of a renewable cluster is not read correctly when this group is an **other RES n**.
Let's look at the **area**'s list.ini : 

**input/renewables/clusters/area/list.ini**
```ini
[clstr-other-11]
name = clstr-other-11
group = Other RES 1
...

[clstr-other-21]
name = clstr-other-21
group = Other RES 2
...
```
Before the current change, Antares code base expected (notice the difference between **list.ini** content and the **expected strings)** : 
```C++
void Data::RenewableCluster::setGroup(Data::ClusterName newgrp)
{
    ... 
    if (newgrp == "other renewable 1")
    {
        groupID = renewableOther1;
        return;
    }
    if (newgrp == "other renewable 2")
    {
        groupID = renewableOther2;
        return;
    }
   ...
    // assigning a default value
    groupID = renewableOther1;
}
```
So a quick fix is made here.

Note that : 
- Nothing in the code base is done to prevent wrong input values : there are no raised exception or exit statement.

### DIFF
--- a/src/libs/antares/study/parts/renewable/cluster.cpp
+++ b/src/libs/antares/study/parts/renewable/cluster.cpp
@@ -133,22 +133,22 @@ void Data::RenewableCluster::setGroup(Data::ClusterName newgrp)
         groupID = windOffShore;
         return;
     }
-    if (newgrp == "other renewable 1")
+    if (newgrp == "other res 1")
     {
         groupID = renewableOther1;
         return;
     }
-    if (newgrp == "other renewable 2")
+    if (newgrp == "other res 2")
     {
         groupID = renewableOther2;
         return;
     }
-    if (newgrp == "other renewable 3")
+    if (newgrp == "other res 3")
     {
         groupID = renewableOther3;
         return;
     }
-    if (newgrp == "other renewable 4")
+    if (newgrp == "other res 4")
     {
         groupID = renewableOther4;
         return;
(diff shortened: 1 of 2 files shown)


flomnes (changes requested):
Instead of code duplication, use `const std::string` ?

```cpp
// Define const values here
const char* Data::RenewableCluster::GroupName(enum RenewableGroup grp)
{
    switch (grp)
    {
    case windOffShore:
        return "Wind offshore"; // Use const value
    }
}

void Data::RenewableCluster::setGroup(Data::ClusterName newgrp)
{
    if (newgrp.empty())
    {
        groupID = renewableOther1;
        pGroup.clear();
        return;
    }
    pGroup = newgrp;
    boost::to_lower(newgrp);
    if (newgrp == "wind onshore") // Use const value
    {
        groupID = windOnShore;
        return;
    }
}
```

guilpier-code:
> Instead of code duplication, use `const std::string` ?

I'm sorry, I don't understand what you mean : I can't see the difference between your excerpt and the current code base.

guilpier-code:
> Instead of code duplication, use `const std::string` ?
I'm sorry, I don't understand what you mean : I can't see the difference between your excerpt and the current code base.

guilpier-code:
Besides, have you noticed that  : 
```C++
const char* Data::RenewableCluster::GroupName(enum RenewableGroup grp);
```
is unused ? 
It has to be re-defined from an abstract method, but stays unused.
Something should be done about it : we should probably remove  **GroupName** from parent abstract class.

flomnes (approved):

PR #1632, opened 2023-09-18
Use 50% as a default value for ST-storage property initiallevel

flomnes:
## Previous behavior
- If `initialleveloptim = true`, then `initiallevel` is ignored and so not required
- If `initialleveloptim = false`, then `initiallevel` is required

There is no default numerical value.

## New behavior
If `initiallevel` is required (`initialleveloptim = false`), it's default value is 0. If it is not provided by the user, then `initiallevel = 0` is used.

## In the code
Replace `std::optional<double>` with `double`, remove `has_value`, `value`, etc.

## To be applied to...
Branches 8.6.x and 8.7.x

### DIFF
--- a/src/libs/antares/study/parts/short-term-storage/properties.h
+++ b/src/libs/antares/study/parts/short-term-storage/properties.h
@@ -63,7 +63,7 @@ public:
     // Not optional   Reservoir capacity in MWh, >= 0
     std::optional<double> reservoirCapacity;
     // Initial level, <= 1
-    std::optional<double> initialLevel;
+    double initialLevel = 0;
     // Bool to optimise or not initial level
     bool initialLevelOptim = false;
     // Efficiency factor between 0 and 1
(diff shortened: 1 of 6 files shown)


src/libs/antares/study/parts/short-term-storage/properties.h
```diff
@@ -63,7 +63,7 @@ class Properties
     // Not optional   Reservoir capacity in MWh, >= 0
     std::optional<double> reservoirCapacity;
     // Initial level, <= 1
-    std::optional<double> initialLevel;
+    double initialLevel = 0;
```
guilpier-code: **Question** :
Default initial level for a STS is **0**. 
Why this value rather than another one ? Like **100%** or **50%** for instance ?
I just crossed @JMJ-rte, and for him, 0% is an extreme value (as much as 100%), he would prefer 50%.

flomnes: Ok, go for 50% !

payetvin (approved):

PR #1637, opened 2023-09-19
Add "Antares::action" CMake library, build only if BUILD_UI=ON

### LINKED ISSUE #1636: Build action/handler/antares-study only if BUILD_UI=ON

flomnes:
Some libraries in libs/antares/study are only used by the GUI executable.

So there is no need to build them if BUILD_UI=OFF.

Reflect this dependency in the CMake build system.

This should speed-up the build a bit when BUILD_UI=OFF.


sylvlecl (approved):
Note that we could also completely move it to the UI directory, but it may involve more work?

flomnes:
> Note that we could also completely move it to the UI directory, but it may involve more work?

That's what I did in a 2nd part. And yes, it took some work fixing headers, making some public, etc.

PR #1641, opened 2023-09-21
Remove unused BindingConstraint::contains

flomnes:


flomnes:
So, it was used
```
/home/runner/work/Antares_Simulator/Antares_Simulator/src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp:299:80:   required from here
/home/runner/work/Antares_Simulator/Antares_Simulator/src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp:266:13: error: no matching function for call to ‘Antares::Data::BindingConstraint::contains(const Antares::Data::BindingConstraint* const&)’
  266 |             if (bc->contains(pItem))
```

PR #1644, opened 2023-09-21
Add getValue for thermal clusters, remove State::timeSeriesIndex

guilpier-code:
- [x] Remove use of `NumeroChroniquesTireesParPays` from class `State`
- [x] Remove `ThermiqueParPalier` from struct `NumeroChroniquesTireesParPays`, use `DataSeriesCommon::getValue` instead
- [x] Fix CI crashes


flomnes (dismissed):

src/libs/antares/study/parts/common/series.h
```diff
@@ -44,6 +42,12 @@ class DataSeriesCommon
 
     void markAsModified() const;
 
+    uint64_t memoryUsage() const;
+
+    double getAvailablePower(unsigned int hour, unsigned int year) const;
+
+    double getSeriesIndex(unsigned int hour, unsigned int year) const;
```
flomnes: Mark as private

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -395,19 +383,17 @@ void BuildThermalPartOfWeeklyProblem(Data::Study& study,
...
             auto& area = *study.areas.byIndex[areaIdx];
-            auto& tsIndex = NumeroChroniquesTireesParPays[numSpace][areaIdx];
             area.thermal.list.each([&](const Data::ThermalCluster& cluster)
             {
                     auto& Pt = problem.PaliersThermiquesDuPays[areaIdx]
                                .PuissanceDisponibleEtCout[cluster.index];
 
                     Pt.CoutHoraireDeProductionDuPalierThermique[hourInWeek] =
-                        cluster.getMarketBidCost(tsIndex.ThermiqueParPalier[cluster.areaWideIndex], hourInYear)
+                        cluster.getMarketBidCost(cluster.series->getSeriesIndex(hourInYear, year), hourInYear)
```
flomnes:
```suggestion
                        cluster.getMarketBidCost(hourInYear, year);
```

Change the definition of `getMarketBidCost` to take (hour, year) instead of (hour, tsIndex).

src/solver/variable/economy/profitByPlant.h
```diff
@@ -277,7 +277,7 @@ class ProfitByPlant : public Variable::IVariable<ProfitByPlant<NextT>, NextT, VC
             auto* cluster = state.area->thermal.clusters[clusterIndex];
             double hourlyClusterProduction
               = thermal[area->index].thermalClustersProductions[clusterIndex];
-            uint tsIndex = state.timeseriesIndex->ThermiqueParPalier[cluster->areaWideIndex];
+            uint tsIndex = cluster->series->timeseriesNumbers[0][state.year];
```
sylvlecl, line 280: Can we add a getter to the series class, so that we don't need to specify the `0` index everytime ?
It clearly looks like an implementation detail that the uses of the class should not need to know.

sylvlecl: Actually it seems that you have added it, but not yet used it everywhere :)

src/libs/antares/study/parts/thermal/cluster.cpp
```diff
@@ -770,8 +770,10 @@ double ThermalCluster::getMarginalCost(uint serieIndex, uint hourInTheYear) cons
      inside -> that is used for all (e.g.10) TS*/
 }
 
-double ThermalCluster::getMarketBidCost(uint serieIndex, uint hourInTheYear) const
+double ThermalCluster::getMarketBidCost(uint hourInTheYear, uint year) const
 {
+    uint serieIndex = (series->timeSeries.width == 1) ? 0 : series->timeseriesNumbers[0][year];
```
sylvlecl, line 775: A small drawback of the new approach is that it makes us compute that indirection at every call.

We could provide a wrapper class to do it only once instead, which would expose a `getMarketBidCost`, something like:
```cpp
// wrapper for the data of one particular year
class ThermalClusterYearData
{
public:
   ThermalClusterYearData(ThermalCluster cluster, uint year);
    double getMarketBidCost(uint hour);
};
//created from the cluster class:
class ThermalCluster
{
public:
    ThermalClusterYearData getYearData(uint year);
}
```

Not a MUST though.

src/solver/optimisation/post_process_commands.cpp
```diff
@@ -30,23 +30,17 @@ void DispatchableMarginPostProcessCmd::execute(const optRuntimeData& opt_runtime
...
-                assert(chro < matrix.width);
-                auto& column = matrix.entry[chro];
-                assert(hourInYear + nbHoursInWeek <= matrix.height && "index out of bounds");
-
                 for (uint h = 0; h != nbHoursInWeek; ++h)
                 {
                     double production = hourlyResults.ProductionThermique[h]
                                           .ProductionThermiqueDuPalier[cluster.index];
-                    dtgmrg[h] += column[h + hourInYear] - production;
+                    double availability = cluster.series->getAvailablePower(h, year);
```
flomnes:
```suggestion
                    double availability = cluster.series->getAvailablePower(h+hourInYear, year);
```

src/solver/variable/economy/profitByPlant.h
flomnes: Yes, the plan is to remove it everywhere and use only `year`.

src/libs/antares/study/parts/thermal/cluster.cpp
flomnes: So the lazy user would call
```
cluster.getYearData(42).getMarketBidCost(51)
```
?

payetvin (approved):

Issue #1615, opened 2023-09-11
Remove global variables related to TS numbers


flomnes:
#1594 - Thermal noises
#1614 - Hydro structures (ventilation)

Global variables related to TS numbers still need to be removed.

PR #1659, opened 2023-09-28
Remove RenouvelableParPalier from global variable

payetvin:
#1615 

https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1644#discussion_r1340097135
https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1644#discussion_r1340087646

- [x] rename getAvailablePower()
- [x] remove scratchpad.ts
- [x] clean alea.cpp

### DIFF
--- a/src/libs/antares/study/parts/hydro/series.h
+++ b/src/libs/antares/study/parts/hydro/series.h
@@ -105,6 +105,8 @@ public:
     */
     void checkMinGenTsNumber(Study& s, const AreaName& areaID);
 
+    unsigned int getIndex(unsigned int year) const;
+
 public:
     /*!
     ** \brief Run-of-the-river - ROR (MW)
--- a/src/libs/antares/study/parts/load/series.cpp
+++ b/src/libs/antares/study/parts/load/series.cpp
@@ -93,7 +93,7 @@ double* DataSeriesLoad::getAvailablePower(unsigned int year) const
 
 unsigned int DataSeriesLoad::getIndex(unsigned int year) const
 {
-    return (timeSeries.width != 1) ? (long)timeseriesNumbers[0][year] : 0;
+    return (timeSeries.width != 1) ? timeseriesNumbers[0][year] : 0;
 }
 
 bool DataSeriesLoad::forceReload(bool reload) const
--- a/src/libs/antares/study/parts/load/series.cpp
+++ b/src/libs/antares/study/parts/load/series.cpp
@@ -91,6 +91,11 @@ double* DataSeriesLoad::getCoefficient(unsigned int year) const
     return timeSeries[getIndex(year)];
 }
 
+double DataSeriesLoad::getCoefficient(const unsigned int year, const unsigned int hour) const
+{
+    return timeSeries[getIndex(year)][hour];
+}
+
 unsigned int DataSeriesLoad::getIndex(unsigned int year) const
 {
     return (timeSeries.width != 1) ? timeseriesNumbers[0][year] : 0;
--- a/src/libs/antares/study/parts/load/series.cpp
+++ b/src/libs/antares/study/parts/load/series.cpp
@@ -86,7 +86,7 @@ int DataSeriesLoadSaveToFolder(DataSeriesLoad* s, const AreaName& areaID, const 
     return 0;
 }
 
-double* DataSeriesLoad::getAvailablePower(unsigned int year) const
+double* DataSeriesLoad::getCoefficient(unsigned int year) const
 {
     return timeSeries[getIndex(year)];
 }
--- a/src/libs/antares/study/parts/load/series.cpp
+++ b/src/libs/antares/study/parts/load/series.cpp
@@ -86,6 +86,16 @@ int DataSeriesLoadSaveToFolder(DataSeriesLoad* s, const AreaName& areaID, const 
     return 0;
 }
 
+double* DataSeriesLoad::getAvailablePower(unsigned int year) const
+{
+    return timeSeries[getIndex(year)];
+}
+
+unsigned int DataSeriesLoad::getIndex(unsigned int year) const
+{
+    return (timeSeries.width != 1) ? (long)timeseriesNumbers[0][year] : 0;
+}
+
 bool DataSeriesLoad::forceReload(bool reload) const
 {
     return timeSeries.forceReload(reload);
--- a/src/libs/antares/study/parts/load/series.cpp
+++ b/src/libs/antares/study/parts/load/series.cpp
@@ -86,7 +86,7 @@ int DataSeriesLoadSaveToFolder(DataSeriesLoad* s, const AreaName& areaID, const 
     return 0;
 }
 
-double* DataSeriesLoad::getCoefficient(unsigned int year) const
+double* DataSeriesLoad::getColumn(unsigned int year) const
 {
     return timeSeries[getIndex(year)];
 }
--- a/src/libs/antares/study/parts/load/series.h
+++ b/src/libs/antares/study/parts/load/series.h
@@ -49,6 +49,9 @@ public:
     */
     void markAsModified() const;
 
+    double* getAvailablePower(unsigned int year) const;
+    unsigned int getIndex(unsigned int year) const;
+
 public:
     /*!
     ** \brief Time series (MW)
--- a/src/libs/antares/study/parts/load/series.h
+++ b/src/libs/antares/study/parts/load/series.h
@@ -49,7 +49,7 @@ public:
     */
     void markAsModified() const;
 
-    double* getAvailablePower(unsigned int year) const;
+    double* getCoefficient(unsigned int year) const;
     unsigned int getIndex(unsigned int year) const;
 
 public:
--- a/src/libs/antares/study/parts/load/series.h
+++ b/src/libs/antares/study/parts/load/series.h
@@ -50,6 +50,7 @@ public:
     void markAsModified() const;
 
     double* getCoefficient(unsigned int year) const;
+    double getCoefficient(const unsigned int year, const unsigned int hour) const;
     unsigned int getIndex(unsigned int year) const;
 
 public:
--- a/src/libs/antares/study/parts/load/series.h
+++ b/src/libs/antares/study/parts/load/series.h
@@ -49,7 +49,7 @@ public:
     */
     void markAsModified() const;
 
-    double* getCoefficient(unsigned int year) const;
+    double* getColumn(unsigned int year) const;
     double getCoefficient(const unsigned int year, const unsigned int hour) const;
     unsigned int getIndex(unsigned int year) const;
 
--- a/src/libs/antares/study/parts/renewable/cluster.cpp
+++ b/src/libs/antares/study/parts/renewable/cluster.cpp
@@ -206,9 +206,9 @@ double RenewableCluster::valueAtTimeStep(uint hourInYear, uint year) const
         return 0.;
     uint timeSeriesIndex = (series->timeSeries.width == 1) ? 0 : series->timeseriesNumbers[0][year];
 
-    assert(timeStepIndex < series->timeSeries.height);
+    assert(hourInYear < series->timeSeries.height);
     assert(timeSeriesIndex < series->timeSeries.width);
-    const double tsValue = series->timeSeries[timeSeriesIndex][timeStepIndex];
+    const double tsValue = series->timeSeries[timeSeriesIndex][hourInYear];
     switch (tsMode)
     {
     case powerGeneration:
--- a/src/libs/antares/study/parts/renewable/cluster.cpp
+++ b/src/libs/antares/study/parts/renewable/cluster.cpp
@@ -200,10 +200,11 @@ YString Data::RenewableCluster::getTimeSeriesModeAsString() const
     return "unknown";
 }
 
-double RenewableCluster::valueAtTimeStep(uint timeSeriesIndex, uint timeStepIndex) const
+double RenewableCluster::valueAtTimeStep(uint year, uint timeStepIndex) const
 {
     if (!enabled)
         return 0.;
+    uint timeSeriesIndex = (series->timeSeries.width == 1) ? 0 : series->timeseriesNumbers[0][year];
 
     assert(timeStepIndex < series->timeSeries.height);
     assert(timeSeriesIndex < series->timeSeries.width);
--- a/src/libs/antares/study/parts/renewable/cluster.cpp
+++ b/src/libs/antares/study/parts/renewable/cluster.cpp
@@ -200,7 +200,7 @@ YString Data::RenewableCluster::getTimeSeriesModeAsString() const
     return "unknown";
 }
 
-double RenewableCluster::valueAtTimeStep(uint timeStepIndex, uint year) const
+double RenewableCluster::valueAtTimeStep(uint hourInYear, uint year) const
 {
     if (!enabled)
         return 0.;
--- a/src/libs/antares/study/parts/renewable/cluster.cpp
+++ b/src/libs/antares/study/parts/renewable/cluster.cpp
@@ -200,7 +200,7 @@ YString Data::RenewableCluster::getTimeSeriesModeAsString() const
     return "unknown";
 }
 
-double RenewableCluster::valueAtTimeStep(uint year, uint timeStepIndex) const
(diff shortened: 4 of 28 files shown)


src/solver/hydro/management/management.cpp
```diff
@@ -417,8 +417,7 @@ void HydroManagement::prepareNetDemand(uint numSpace)
 
                 area.renewable.list.each([&](const Antares::Data::RenewableCluster& cluster) {
                     assert(cluster.series->timeSeries.jit == NULL && "No JIT data from the solver");
-                    netdemand -= cluster.valueAtTimeStep(
-                      ptchro.RenouvelableParPalier[cluster.areaWideIndex], hour);
+                    netdemand -= cluster.valueAtTimeStep(year, hour);
```
flomnes: For consistency with thermal cluster, I suggest `(hour, year)` instead of `(year, hour)` (to be discussed)

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -200,10 +200,11 @@ YString Data::RenewableCluster::getTimeSeriesModeAsString() const
     return "unknown";
 }
 
-double RenewableCluster::valueAtTimeStep(uint timeSeriesIndex, uint timeStepIndex) const
+double RenewableCluster::valueAtTimeStep(uint timeStepIndex, uint year) const
```
flomnes:
```suggestion
double RenewableCluster::valueAtTimeStep(uint hourInYear, uint year) const
```

flomnes (changes requested):
I think the precise naming for "timeStep" is **hourInYear**

So the caller know that it's an hour between 0 (included) and 8760 (excluded).

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -200,14 +200,15 @@ YString Data::RenewableCluster::getTimeSeriesModeAsString() const
     return "unknown";
 }
 
-double RenewableCluster::valueAtTimeStep(uint timeSeriesIndex, uint timeStepIndex) const
+double RenewableCluster::valueAtTimeStep(uint hourInYear, uint year) const
```
guilpier-code, line 203: One could expect that arguments are switched (as it was before) : first **year**, then **hourInYear**. Another way to see this, is : you cannot have an hour in the year if you don't have a year first, right ?

src/libs/antares/study/parts/renewable/cluster.cpp
```diff
@@ -200,14 +200,15 @@ YString Data::RenewableCluster::getTimeSeriesModeAsString() const
...
 {
     if (!enabled)
         return 0.;
+    uint timeSeriesIndex = (series->timeSeries.width == 1) ? 0 : series->timeseriesNumbers[0][year];
```
guilpier-code, line 207: About `uint timeSeriesIndex = (series->timeSeries.width == 1) ? 0 : series->timeseriesNumbers[0][year];` : 
There is already a method that does the same thing : 
```C++
uint DataSeriesCommon::getSeriesIndex(unsigned int year)  const
```
So, if we don't use it, we have a small code duplication.
I made the same remark in a review of #1644 I think. 
So there is at least another place where this duplication is made. 
This method is private, we should make it **public** and use it.

src/libs/antares/study/parts/wind/series.cpp
```diff
@@ -72,6 +72,16 @@ int DataSeriesWindSaveToFolder(DataSeriesWind* d, const AreaName& areaID, const
...
+double* DataSeriesWind::getAvailablePower(unsigned int year) const
+{
+    return timeSeries[getIndex(year)];
+}
+
+unsigned int DataSeriesWind::getIndex(unsigned int year) const
+{
+    return (timeSeries.width != 1) ? (long)timeseriesNumbers[0][year] : 0;
+}
+
```
guilpier-code: We start to have some code duplication with **clusters** (thermal or renewable) : classes for time series look very much the same : same interface, and even same body.
Furthermore, we expect more code duplication to come : when changes will reach load, solar, ...
We should think about a way to gather what's common in a single class.

guilpier-code (changes requested):

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -570,6 +570,7 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
             auto& area = *(study.areas.byIndex[k]);
             auto& scratchpad = area.scratchpad[numSpace];
             auto& ror = area.hydro.series->ror;
+            auto& windSeries = area.wind.series;
```
guilpier-code: As we use `problem.year` several times, **maybe** we could define `unsigned int year =  problem.year;` once for all at the beginning of function **SIM_RenseignementProblemeHebdo**
Up to you.

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -578,15 +579,15 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
             {
                 assert((uint)hourInYear < scratchpad.ts.solar.height);
                 assert((uint)hourInYear < scratchpad.ts.wind.height);
-                assert((uint)tsIndex.Eolien < scratchpad.ts.wind.width);
+                assert(windSeries->getIndex(problem.year) < scratchpad.ts.wind.width);
```
guilpier-code: About line `assert(windSeries->getIndex(problem.year) < scratchpad.ts.wind.width);` : 
- The **assert** raises a question : it is really needed ? I mean : at this point, is it possible that the **assert** raises a problem ? (I feel like removing this **assert**)
- The **scratchpad** : whatever the reason why the `scratchpad.ts` exists, keeping **ts** inside the **scratchpad** seems useless. We should remove it. 

More generally, what is the **scratchpad** for ?
It contains very different and heterogeneous things, and some of them are useless (as above). 
We should try to split it : its pieces could be moved in other places, where they belong.
At least we could make this class shrink and see what's left.

By the way, I feel the same thing for **StudyRuntimeInfos** (as some of you know).

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -96,7 +96,7 @@ static bool Remix(const Data::AreaList& areas, PROBLEME_HEBDO& problem, uint num
             double bottom = std::numeric_limits<double>::max();
             double top = 0;
 
-            uint loadTS = NumeroChroniquesTireesParPays[numSpace][index].Consommation;
+            uint loadTS = area.load.series->getIndex(problem.year);;
```
guilpier-code: One ";" is enough (end of line)

src/solver/variable/commons/wind.h
```diff
@@ -196,10 +196,10 @@ class TimeSeriesValuesWind
     {
         if (isRenewableGenerationAggregrated)
         {
+            /* pValuesForTheCurrentYear[numSpace].hour = pArea->wind.series->getAvailablePower(year); */
```
guilpier-code: Please remove any useless comment

guilpier-code (dismissed):

flomnes:
Rename getAvailablePower into something else

src/libs/antares/study/parts/load/series.h
```diff
@@ -49,6 +49,10 @@ class DataSeriesLoad
     */
     void markAsModified() const;
 
+    double* getCoefficient(unsigned int year) const;
```
flomnes:
```suggestion
    double* getColumn(unsigned int year) const;
```

src/solver/hydro/management/management.cpp
```diff
@@ -400,25 +394,24 @@ void HydroManagement::prepareNetDemand(uint numSpace)
             // Aggregated renewable production: wind & solar
             if (parameters_.renewableGeneration.isAggregated())
             {
-                netdemand = +scratchpad.ts.load[ptchro.Consommation][hour]
-                            - scratchpad.ts.wind[ptchro.Eolien][hour] - scratchpad.miscGenSum[hour]
-                            - scratchpad.ts.solar[ptchro.Solar][hour] - ror[hour]
+                netdemand = +area.load.series->getCoefficient(year, hour)
+                            - area.wind.series->getCoefficient(year, hour) - scratchpad.miscGenSum[hour]
+                            - area.solar.series->getCoefficient(year, hour)
```
flomnes: Here you could use the version that returns a `double*`, so you'd only need 1 function call instead of 8760

src/solver/hydro/management/management.cpp
```diff
@@ -400,25 +394,24 @@ void HydroManagement::prepareNetDemand(uint numSpace)
...
                             - ((ModeT != Data::stdmAdequacy) ? scratchpad.mustrunSum[hour]
                                                              : scratchpad.originalMustrunSum[hour]);
             }
 
             // Renewable clusters, if enabled
             else if (parameters_.renewableGeneration.isClusters())
             {
-                netdemand = scratchpad.ts.load[ptchro.Consommation][hour]
+                netdemand = area.load.series->getCoefficient(year, hour)
```
flomnes: same

src/libs/antares/study/parts/load/series.cpp
```diff
@@ -86,6 +86,21 @@ int DataSeriesLoadSaveToFolder(DataSeriesLoad* s, const AreaName& areaID, const
     return 0;
 }
 
+double* DataSeriesLoad::getCoefficient(unsigned int year) const
+{
```
flomnes: getColumn

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -566,28 +567,21 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
...
-            if (parameters.renewableGeneration.isAggregated())
-            {
-                assert((uint)hourInYear < scratchpad.ts.solar.height);
-                assert((uint)hourInYear < scratchpad.ts.wind.height);
-                assert((uint)tsIndex.Eolien < scratchpad.ts.wind.width);
-                assert((uint)tsIndex.Solar < scratchpad.ts.solar.width);
-            }
 
-            uint tsFatalIndex = (uint)tsIndex.Hydraulique < ror.width ? tsIndex.Hydraulique : 0;
+            uint tsFatalIndex = hydroSeriesIndex < ror.width ? hydroSeriesIndex : 0;
```
sylvlecl, line 580: It seems that the `hydroSeriesIndex` is used only to get this ROR index:
we could probably have a more direct `getRorIndex` in the series class ?

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -566,28 +567,21 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
...
             double& mustRunGen = problem.AllMustRunGeneration[hourInWeek].AllMustRunGenerationOfArea[k];
             if (parameters.renewableGeneration.isAggregated())
             {
-                mustRunGen = scratchpad.ts.wind[tsIndex.Eolien][hourInYear]
-                             + scratchpad.ts.solar[tsIndex.Solar][hourInYear]
+                mustRunGen = area.wind.series->getCoefficient(year, hourInYear)
```
sylvlecl: For wind we get the coefficient here, but for load and solar we get them above:
it's a little misleading, we can probably perform all those operations in the same place

src/solver/variable/commons/hydro.h
```diff
@@ -200,7 +200,7 @@ class TimeSeriesValuesHydro
         // The current time-series
         auto& ror = pArea->hydro.series->ror;
         const unsigned int nbchro
-          = NumeroChroniquesTireesParPays[numSpace][pArea->index].Hydraulique;
+          = pArea->hydro.series->getIndex(year);
         pFatalValues[numSpace] = &(ror.entry[(nbchro < ror.width ? nbchro : 0)]);
```
sylvlecl, line 204: Could be simplified too with a `getRorIndex`

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -566,28 +567,21 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
 
         for (uint k = 0; k < nbPays; ++k)
         {
-            auto& tsIndex = NumeroChroniquesTireesParPays[numSpace][k];
             auto& area = *(study.areas.byIndex[k]);
             auto& scratchpad = area.scratchpad[numSpace];
             auto& ror = area.hydro.series->ror;
+            auto loadSeries = area.load.series->getCoefficient(year, hourInYear);
```
sylvlecl, line 573: As pointed by @flomnes , we can retrieve the full columns only once et the start and use them (it avoids to repeat the small logic on the index)

src/libs/antares/study/parts/hydro/series.h
```diff
@@ -105,6 +105,8 @@ class DataSeriesHydro
     */
     void checkMinGenTsNumber(Study& s, const AreaName& areaID);
 
+    unsigned int getIndex(unsigned int year) const;
```
sylvlecl, line 108: I think that every time we get this index, it is only to use it later to get one of the 3 corresponding columns in ROR / storage / mingen.

I would be in favor, instead, of returning straight from here what is actually useful, which are those 3 columns, in a small struct:
```cpp
struct YearlyHydroSeries {
    Matrix<double, int32_t>::ColumnType ror;
    Matrix<double, int32_t>::ColumnType storage;
    Matrix<double, int32_t>::ColumnType mingen;
};

class DataSeriesHydro {
...
    YearlyHydroSeries getSeries(unsigned int year);
...
};
```
or possibly 3 getters to get the columns.

But maybe for a later PR, since future refactorings are planned ?

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -638,10 +631,12 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
...
-                uint tsIndex = (NumeroChroniquesTireesParPays[numSpace][k]).Hydraulique;
-                auto& inflowsmatrix = area.hydro.series->storage;
+                auto& hydroSeries = area.hydro.series;
+                uint tsIndex = hydroSeries->getIndex(year);
+
+                auto& inflowsmatrix = hydroSeries->storage;
                 auto const& srcinflows = inflowsmatrix[tsIndex < inflowsmatrix.width ? tsIndex : 0];
-                auto& mingenmatrix = area.hydro.series->mingen;
+                auto& mingenmatrix = hydroSeries->mingen;
                 auto const& srcmingen = mingenmatrix[tsIndex < mingenmatrix.width ? tsIndex : 0];
```
sylvlecl, line 641: We could have a getter for this column in the HydroSeries class, instead of going through 1. retrieve the index then 2. retrieve the column.

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -638,10 +631,12 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
             if (problem.CaracteristiquesHydrauliques[k].PresenceDHydrauliqueModulable > 0)
             {
                 auto& area = *study.areas.byIndex[k];
-                uint tsIndex = (NumeroChroniquesTireesParPays[numSpace][k]).Hydraulique;
-                auto& inflowsmatrix = area.hydro.series->storage;
+                auto& hydroSeries = area.hydro.series;
+                uint tsIndex = hydroSeries->getIndex(year);
+
+                auto& inflowsmatrix = hydroSeries->storage;
                 auto const& srcinflows = inflowsmatrix[tsIndex < inflowsmatrix.width ? tsIndex : 0];
```
sylvlecl, line 639: Same here, we could have a getter for the column itself instead of the index

sylvlecl (commented):
Nice:
I think we could improve again the design by returning more often column objects (double* in practice) instead of getting the year index then only retrieving the column using that index.

However, maybe that's for a future work on series classes?

src/libs/antares/study/parts/hydro/series.h
flomnes: See #1669 ;)

flomnes (approved):

src/libs/antares/study/parts/renewable/cluster.cpp
flomnes: @guilpier-code to be done in the next (factorization) work

PR #1661, opened 2023-09-29
Check value for spinning, emit error if > 99.999%

### LINKED ISSUE #1662: Spinning value 100 should not be accepted, it causes silent errors
It produces `nan` values in generated time-series for thermal clusters

```
-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan
-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan	-nan
```

flomnes:


sylvlecl (approved):
Note for later linked with https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/1662#issuecomment-1740777103 :
we should not change in-place the timeseries in the input data (`calculationOfSpinning`), and try to revert it afterwards. Instead we should leave those input data untouched, and only take into account the spinning at computation time.

flomnes:
According to AB, spinning=100% should still be admissible.

We should instead avoid divisions by 0.

PR #1667, opened 2023-10-02
Remove code duplication in src/libs/antares/study/parameters.cpp

flomnes:


src/libs/antares/study/parameters.cpp
```diff
@@ -1350,12 +1350,6 @@ void Parameters::prepareForSimulation(const StudyLoadOptions& options)
     switch (mode)
     {
     case stdmEconomy:
-    {
-        // The year-by-year mode might have been requested from the command line
-        if (options.forceYearByYear)
-            yearByYear = true;
-        break;
-    }
```
flomnes, line 1358: `case stdmAdequacy` (below) has the exact same code block.

JasonMarechal25 (approved):

PR #1674, opened 2023-10-04
[Win] Increase file size limit when reading file

JasonMarechal25:


a-zakir (approved):

flomnes (commented):
I'm afraid I don't understand this change. Is the size of `enum` limited to that of an `int` ?

a-zakir:
> I'm afraid I don't understand this change. Is the size of `enum` limited to that of an `int` ?

in this particular case, in msvc anonymous enum with one component has zero as value. e.g `IO::Error LoadFromFile(std::string& out,
                       const AnyString& filename,
                       uint64_t hardlimit = sizeHardLimit);` 
**`uint64_t hardlimit = 0`**

PR #1676, opened 2023-10-04
Use `std::unique_ptr::operator->` rather than `std::unique_ptr::get` & raw pointers

flomnes:
Using `std::unique_ptr::operator->` offers a bit more safety against the possible scenario

1. A raw pointer is obtained through `T* ptr = unique.get()`
2. Object `unique` is deleted, freeing the underlying memory
3. The caller de-references `ptr`, resulting in a segfault

### DIFF
--- a/src/analyzer/atsp/correlations.cpp
+++ b/src/analyzer/atsp/correlations.cpp
@@ -629,7 +629,7 @@ bool ATSP::computeMonthlyCorrelations()
     }
 
     // removing the mapping list
-    delete mapping;
+    delete[] mapping;
 
     return true;
 }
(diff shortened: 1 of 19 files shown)


src/analyzer/atsp/correlations.cpp
```diff
@@ -629,7 +629,7 @@ bool ATSP::computeMonthlyCorrelations()
     }
 
     // removing the mapping list
-    delete mapping;
+    delete[] mapping;
```
flomnes, line 632:
```cpp
uint* mapping = new uint[pArea.size()];
```

g++12 gives a warning, g++-9 doesn't

JasonMarechal25 (approved):

PR #1677, opened 2023-10-04
Refactor time series

### LINKED ISSUE #1664: Global variables : remarks on merged PRs
If relevant, these remarks should be the subject of a new PR.

**Remarks on PR** #1644 : 
[1] **src/libs/antares/study/parts/thermal/cluster.cpp** : function **getMarketBidCost** : 
  1st line `uint serieIndex = (series->timeSeries.width == 1) ? 0 : series->timeseriesNumbers[0][year];` seems to be exactly what **DataSeriesCommon::getSeriesIndex** does. So we have a small code duplication. We could make **getSeriesIndex** public and call it from **getMarketBidCost**.

**Remarks on PR** #1660 : 
[2] I can see (in this PR), that we replace some calls to **getAvailablePower** with calls to **getAvailablePowerYearly**. I haven't checked if all calls are replaced, but if that's the case, we should consider removing  **getAvailablePower**

### LINKED ISSUE #1669: De-duplicate code associated to time-series & time-series numbers
- [ ] Binding constraints
- [x] Links

Main PR:  #1677

## Idea
One shared `TimeSeriesNumbers` object, used by one or many `TimeSeries` objects.

`TimeSeries` can only be constructed from an existing `TimeSeriesNumbers`.

```cpp
#include <vector>

class TimeSeriesNumbers {
public:
  int getTSNumber(int year) const {
      tsNumbers_[year];
  }
private:
  std::vector<int> tsNumbers_;
};

class TimeSeries {
public:
  TimeSeries(const TimeSeriesNumbers& tsNumbers) : tsNumbers(tsNumbers) {}
  double getCoefficient(int year, int hourInYear) {
      return coefficients[tsNumbers.getTSNumber(year)][hourInTheYear];
  }
private:
  Matrix<> coefficients;
  const TimeSeriesNumbers& tsNumbers;
};

// Thermique pour un cluster thermique
class ThermalCluster {
public:
  ThermalCluster() : availability(tsNumbers), marketBidCost(tsNumbers) {}

  TimeSeries availability;
  TimeSeries marketBidCost;
private:
  TimeSeriesNumbers tsNumbers;
};

// Hydro pour une zone
class Hydro {
public:
  Hydro() : ror(tsNumbers), ruleCurves(tsNumbers) {}

  TimeSeries ror;
  TimeSeries ruleCurves;
private:
  TimeSeriesNumbers tsNumbers;
};

void usage() {
  ThermalCluster cluster;
  cluster.availability.getCoefficient(year, hourInYear);
  cluster.availability.getColumn(year)[hourInTheYear];
  auto& avail = cluster.availability.getColumn(year);
  a = avail[hourInYear];
  Hydro hydro;
}
```

payetvin:
#1669

# TODO
- [X] Wind
- [X] Load #1686
- [X] Solar #1688 
- [x] Thermal #1693 
- [x] Renewable clusters #1693 
- [x] Hydro #1705
- [x] `(year, hour)` or `(hour, year)` ?
- [x] File name for hydro/thermal
- [x] tests
- [ ] Binding constraints ?
- [ ] Links ?
- [ ] use `std::vector` instead of `Matrix<>`
- [ ] ihm


src/libs/antares/study/parts/common/series.cpp
```diff
@@ -38,6 +38,75 @@ using namespace Yuni;
...
+    if (s.usedByTheSolver && s.parameters.derated)
+        timeSeries.averageTimeseries();
+
+    timeseriesNumbers.clear();
+
+    return ret;
+}
+
+int TimeSeries::timeSeriesSaveToFolder(const AreaName& areaID, const std::string& folder,
+                                       const std::string& filename) const
```
flomnes: ~~filename~~ prefix

src/libs/antares/study/parts/common/series.cpp
```diff
@@ -38,6 +38,75 @@ using namespace Yuni;
...
+{
+    Clob buffer;
+    int ret = 1;
+    buffer.clear() << folder << SEP << filename << areaID << ".txt";
+    ret = timeSeries.saveToCSVFile(buffer, 0) && ret;
+
+    return ret;
```
flomnes:
```suggestion
    buffer.clear() << folder << SEP << filename << areaID << ".txt";
    return timeSeries.saveToCSVFile(buffer, 0);
```

src/libs/antares/study/parts/common/series.h
```diff
@@ -33,6 +33,51 @@
 
 namespace Antares::Data
 {
+
+class TimeSeries
+{
+public:
+    typedef Matrix<uint32_t> TSNumbers;
```
flomnes:
```suggestion
    using TSNumbers =Matrix<uint32_t>;
```

flomnes: Also add a typedef for `timeSeries`

src/libs/antares/study/parts/common/series.h
```diff
@@ -33,6 +33,54 @@
...
+    uint32_t getSeriesIndex(uint32_t year) const;
+
+    void reset();
+
+    bool forceReload(bool reload = false) const;
+    void markAsModified() const;
+    uint64_t memoryUsage() const;
+
+    TScoefficients timeSeries;
+    TSNumbers timeseriesNumbers;
```
flomnes: In some cases (thermal, hydro), we have multiple `TScoefficients` objects sharing a single instance of TSNumbers. For example availability, CO2 cost, fuel cost, etc.

```suggestion
    TScoefficients timeSeries;
    const TSNumbers& timeseriesNumbers;
```

flomnes (dismissed):

src/libs/antares/study/parts/common/series.h
```diff
@@ -33,6 +33,54 @@
 
 namespace Antares::Data
 {
+
+class TimeSeries
```
sylvlecl: We should take the opportunity of this refactoring to have some unit tests on those classes, in particular this one.

We need to test the indexing logic and the serialization methods.

src/libs/antares/study/CMakeLists.txt
```diff
@@ -1,3 +1,21 @@
...
+)
+source_group("study\\part\\series" FILES ${SRC_STUDY_PART_SERIES})
+
+target_sources(series PUBLIC ${SRC_STUDY_PART_SERIES})
+
+target_link_libraries(series PUBLIC Antares::array)
+
+target_include_directories(series
+        PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}/parts/series
```
sylvlecl: We should move this library to the same level as other libraries: array, study, etc. This CMakeLists.txt should be only dedicated to the study lib, to make things clear.
Here it's really not clear: all other subfolders are part of the "study" library, but only this subfolder is a separate library.

Also we must not expose this directory as a public header directory: instead only the new "include" directory must be defined as the public header directory.

src/libs/antares/study/parts/series/include/antares/series/series.h
```diff
@@ -0,0 +1,90 @@
...
+#include <antares/array/matrix.h>
+
+namespace Antares::Data
+{
+
+class TimeSeries
+{
+public:
+    using TSNumbers = Matrix<uint32_t>;
+    using TS = Matrix<double>;
```
sylvlecl: It seems a little dangerous to remove the fact that some classes were using integer types as the read/write type: what are the impacts, in particular on the UI ?

I'll try to think a little more about this.

src/libs/antares/study/parts/hydro/series.h
```diff
@@ -27,6 +27,7 @@
 #ifndef __ANTARES_LIBS_STUDY_PARTS_HYDRO_TIMESERIES_H__
 #define __ANTARES_LIBS_STUDY_PARTS_HYDRO_TIMESERIES_H__
 
+#include <include/antares/series/series.h>
```
sylvlecl: We should not have the "include" folder, here, see comment on the définition of the include dir in CMakeLists.txt

sylvlecl (changes requested):
First comments just about the new lib definition:
we need to move it outside the study folder, and to fix its include directory.

src/libs/antares/study/parts/thermal/ecoInput.h
```diff
@@ -86,8 +87,8 @@ class EconomicInputData
 
     //! All {FO,PO}{Duration,Rate} annual values
     // max x DAYS_PER_YEAR
-    Matrix<double> fuelcost;
```
sylvlecl, line 89: Not sure it's relevant to use that type here since it was not really involved in this work.

src/libs/antares/study/parts/series/include/antares/series/series.h
```diff
@@ -0,0 +1,90 @@
...
+    /*!
+     ** \brief Load series from a file
+     **
+     ** \param d Data series
+     ** \param areaID The ID of the area associated to the data series
+     ** \param folder The source folder
+     ** \param filename The filename of the series
+     ** \return A non-zero value if the operation succeeded, 0 otherwise
+     */
+    bool timeSeriesLoadFromFolder(const std::string& path,
```
sylvlecl: Just `loadFromFolder` to avoid name redundancy

src/libs/antares/study/parts/series/include/antares/series/series.h
```diff
@@ -0,0 +1,90 @@
...
+    /*!
+     ** \brief Save time series to a file
+     ** \ingroup windseries
+     **
+     ** \param areaID The ID of the area associated to the data series
+     ** \param folder The target folder
+     ** \param filename The filename of the series
+     ** \return A non-zero value if the operation succeeded, 0 otherwise
+     */
+    int timeSeriesSaveToFolder(const AreaName& areaID, const std::string& folder,
```
sylvlecl: Just `saveToFolder` to avoid name redundancy

src/libs/antares/study/parts/series/include/antares/series/series.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include <antares/array/matrix.h>
+
+namespace Antares::Data
+{
+
+class TimeSeries
+{
+public:
+    using TSNumbers = Matrix<uint32_t>;
```
sylvlecl: Since it's already in `TimeSeries` prefix, we could rename it just `Numbers`.

src/libs/antares/study/parts/series/include/antares/series/series.h
sylvlecl: Here again there's a redundancy with encompassing `TimeSeries`, maybe use `Values` ?

src/libs/antares/study/parts/series/include/antares/series/series.h
```diff
@@ -0,0 +1,90 @@
...
+*/
+#ifndef __ANTARES_LIBS_STUDY_PARTS_COMMON_TIMESERIES_H__
+#define __ANTARES_LIBS_STUDY_PARTS_COMMON_TIMESERIES_H__
+
+#include <antares/array/matrix.h>
+
+namespace Antares::Data
+{
+
+class TimeSeries
```
sylvlecl: We should add some documentation comment to clarify the goal of this class (encapsulating the redirection to the correct timeseries for a given year, basically)

src/libs/antares/study/parts/series/include/antares/series/series.h
```diff
@@ -0,0 +1,90 @@
...
+    double getCoefficient(uint32_t year, uint32_t hourInYear) const;
+    const double* getColumn(uint32_t year) const;
+    uint32_t getSeriesIndex(uint32_t year) const;
+
+    double* operator[](uint32_t index);
+
+    void reset();
+    void unloadFromMemory() const;
+    void roundAllEntries();
+    void resize(uint32_t year, uint32_t hour);
```
sylvlecl: Plural `years` and `hours` would be clearer.

But actually "year" is misleading here: usually what we call years are the monte carlo years. But here, we are not changing the number of MC years, we are changing the number of timeseries. So we could call the parameter "timeSeriesCount" ?

src/libs/antares/study/parts/series/include/antares/series/series.h
```diff
@@ -0,0 +1,90 @@
...
+     */
+    int timeSeriesSaveToFolder(const AreaName& areaID, const std::string& folder,
+                               const std::string& prefix) const;
+
+    bool setCoefficient(double value, uint32_t year, uint32_t hourInYear);
+    double getCoefficient(uint32_t year, uint32_t hourInYear) const;
+    const double* getColumn(uint32_t year) const;
+    uint32_t getSeriesIndex(uint32_t year) const;
+
+    double* operator[](uint32_t index);
```
sylvlecl: Thanks for renaming it to index, just a thought:
it makes me think that it was probably not a great idea to mix in the same class the access by index and the access by years.
We will really need further refactoring after this PR to separate better those 2 concerns.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -78,84 +94,87 @@ bool DataSeriesHydro::loadFromFolder(Study& study, const AreaName& areaID, const
...
 
-    if (study.usedByTheSolver)
+    if (!study.usedByTheSolver)
     {
-        if (0 == count)
-        {
-            logs.error() << "Hydro: `" << areaID
-                         << "`: empty matrix detected. Fixing it with default values";
-            ror.reset(1, HOURS_PER_YEAR);
-            storage.reset(1, DAYS_PER_YEAR);
```
sylvlecl, line 105: Here we have changed the behaviour: we had days per year.

It's an issue, it means that the index in `TimeSeries` class is not always an hour, so we cannot name it hour in that class.

src/libs/antares/study/study.h
```diff
@@ -610,7 +610,7 @@ class Study: public Yuni::NonCopyable<Study>, public IObject, public LayerData
     ScenarioBuilder::Sets* scenarioRules = nullptr;
     //@}
 
-    Matrix<double> scenarioHydroLevels;
```
sylvlecl, line 613: Here too, not sure it's relevant to change the type for now.

src/solver/hydro/management/management.cpp
```diff
@@ -328,7 +320,7 @@ bool HydroManagement::checkHourlyMinGeneration(uint tsIndex, Data::Area& area) c
                     logs.error()
                         << "In area: " << area.name << " [hourly] minimum generation of "
                         << srcmingen[day * 24 + h] << " MW in timestep " << day * 24 + h + 1
-                        << " of TS-" << tsIndex + 1
+                        << " of TS-" << year + 1
```
sylvlecl: We either need to change the log, or to retrieve the corresponding TS index here (unfortunately). Better to get the TS index, so the user understands what data it is.
Otherwise it's misleading: when reading the log, we will understand it's the TS number, whereas it's actually the MC year.

src/solver/hydro/management/management.cpp
```diff
@@ -245,34 +239,32 @@ bool HydroManagement::checkMonthlyMinGeneration(uint numSpace, uint tsIndex, con
...
-bool HydroManagement::checkYearlyMinGeneration(uint numSpace, uint tsIndex, const Data::Area& area) const
+bool HydroManagement::checkYearlyMinGeneration(uint numSpace, uint year, const Data::Area& area) const
 {
     const auto& data = tmpDataByArea_[numSpace][area.index];
     if (data.totalYearMingen > data.totalYearInflows)
     {
         // Yearly minimum generation <= Yearly inflows
         logs.error() << "In Area " << area.name << " the minimum generation of "
-                     << data.totalYearMingen << " MW of TS-" << tsIndex + 1
+                     << data.totalYearMingen << " MW of TS-" << year + 1
```
sylvlecl: Same, TS index

src/solver/hydro/management/management.cpp
```diff
@@ -245,34 +239,32 @@ bool HydroManagement::checkMonthlyMinGeneration(uint numSpace, uint tsIndex, con
         {
             logs.error() << "In Area " << area.name << " the minimum generation of "
                          << data.totalMonthMingen[realmonth] << " MW in month " << month + 1
-                         << " of TS-" << tsIndex + 1 << " is incompatible with the inflows of "
+                         << " of TS-" << year + 1 << " is incompatible with the inflows of "
```
sylvlecl: Same, TS index

src/solver/hydro/management/management.cpp
```diff
@@ -234,7 +228,7 @@ void HydroManagement::minGenerationScaling(uint numSpace, uint year) const
       });
 }
 
-bool HydroManagement::checkMonthlyMinGeneration(uint numSpace, uint tsIndex, const Data::Area& area) const
+bool HydroManagement::checkMonthlyMinGeneration(uint numSpace, uint year, const Data::Area& area) const
```
sylvlecl, line 231: we need to change the parameter name in headers too, same for all methods

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -96,8 +96,8 @@ static bool Remix(const Data::AreaList& areas, PROBLEME_HEBDO& problem, uint num
             double bottom = std::numeric_limits<double>::max();
             double top = 0;
 
-            uint loadTS = area.load.series->getIndex(problem.year);
-            auto& load = area.load.series->timeSeries;
+            uint loadTS = area.load.series.getSeriesIndex(problem.year);
+            auto& load = area.load.series.timeSeries;
             assert(load.width > 0);
 
             auto& L = (loadTS < load.width) ? load[loadTS] : load[0];
```
sylvlecl: should be replaced with call to getCOlumn, and remove the getSeriesIndex above

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -574,32 +574,31 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
...
             auto& scratchpad = area.scratchpad[numSpace];
             auto& ror = area.hydro.series->ror;
-            auto loadSeries = area.load.series->getCoefficient(year, hourInYear);
-            auto windSeries = area.wind.series->getCoefficient(year, hourInYear);
-            auto solarSeries = area.solar.series->getCoefficient(year, hourInYear);
-            auto hydroSeriesIndex = area.hydro.series->getIndex(year);
+            auto loadSeries = area.load.series.getCoefficient(year, hourInYear);
+            auto windSeries = area.wind.series.getCoefficient(year, hourInYear);
+            auto solarSeries = area.solar.series.getCoefficient(year, hourInYear);
+            auto rorSeriesIndex = area.hydro.series->ror.getSeriesIndex(year);
```
sylvlecl: I think we can just get the rorSeries like for others, not the index

sylvlecl (dismissed):
Some more comments.

A quite important one is that not all timeseries are actually hourly, at least one is a daily one.
So we need to make a choice:
- either use the same class and rename the "hour" parameters in a more generic "timestep"
- or separate between 2 classes

The benefit of the second one is to document by the type what kind of data is actually expected, which would be nice.
In the "cons", it would probably imply some code duplication. Or just some template parameter, but if we do this we lose the benefit of being able to name correctly the arguments (`hour` and `day`).

guilpier-code (approved):

Issue #1658, opened 2023-09-28
Remove dependency to UI


flomnes:
See solver/main.cpp

```
#include "../ui/common/winmain.hxx"  //TODO: remove that reverse dependency to UI
```

Create a CMake target out of the UI directory.

We could also create a small C++ wrapper to avoid having to manually call FreeUTF8Arguments (`~Wrapper() {FreeUTF8Arguments();}`)

PR #1678, opened 2023-10-04
Remove dependency to UI, use RAII to handle resources

guilpier-code:
Fix issue #1658 

- [x] extract **main(...)** arguments processing from UI
- [x] Call **main(...)** arguments processing from the physical new location for each call
- [x] Make the arguments processing a class (or a wrapper) and use the RAII to avoid calling a destruction function
- [x] Correct the location of the arguments processing header (.h) to stick to what is done elsewhere (**args/antares/args**)


src/libs/antares/args/args_to_utf8.h
flomnes, line 1: Please create directory **antares/args** here and put this header file inside. 

See [this ADR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/develop/docs/Architecture_Decision_Records/CMake_design_choice.md)

See for example **src/libs/antares/date/antares/date/date.h**.

guilpier-code: Done

src/libs/antares/args/args_to_utf8.cpp
```diff
@@ -0,0 +1,51 @@
...
+{
+    wchar_t** wargv = CommandLineToArgvW(GetCommandLineW(), &argc);
+    char** argvUTF8 = (char**)malloc(argc * sizeof(char*));
+    for (int i = 0; i != argc; ++i)
+    {
+        const uint len = (uint)wcslen(wargv[i]);
+        const uint newLen = WideCharToMultiByte(CP_UTF8, 0, wargv[i], len, NULL, 0, NULL, NULL);
+        argvUTF8[i] = (char*)malloc((newLen + 1) * sizeof(char));
+        memset(argvUTF8[i], 0, (newLen + 1) * sizeof(char));
+        WideCharToMultiByte(CP_UTF8, 0, wargv[i], len, argvUTF8[i], newLen, NULL, NULL);
```
flomnes: I think you could simplify a bit
```cpp
char** AntaresGetUTF8Arguments(int argc, char**)
{
#ifdef YUNI_OS_WINDOWS
// windows implementation here
#else
// non-windows implementation here
#endif
}
```

flomnes:
For the RAII topic, you have (at least) two options

- Make a C++ class that owns a `char**`
- Make this function operate on and return `std::vector<std::string>` or `std::vector<std::wstring>` and let `std::vector` manage it's own memory

guilpier-code:
> For the RAII topic, you have (at least) two options
> 
> * Make a C++ class that owns a `char**`
> * Make this function operate on and return `std::vector<std::string>` or `std::vector<std::wstring>` and let `std::vector` manage it's own memory

Ok, knowing that the second change is more invasive than the first.

src/libs/antares/args/args_to_utf8.cpp
guilpier-code: Done, even if a class now replaces the 2 functions

src/libs/antares/args/args_to_utf8.cpp
```diff
@@ -0,0 +1,42 @@
...
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN 1
+#endif
+#include <windows.h>
+#include <shellapi.h>
+#include <stdlib.h>
+#include <stdio.h>
+#endif
+
+IntoUTF8ArgsTranslator::IntoUTF8ArgsTranslator(int argc, char**& argv)
```
flomnes:
```suggestion
IntoUTF8ArgsTranslator::IntoUTF8ArgsTranslator(int argc, char** argv)
```

src/libs/antares/args/include/antares/args/args_to_utf8.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+class IntoUTF8ArgsTranslator
+{
+public:
+	IntoUTF8ArgsTranslator(int argc, char**& argv);
```
flomnes:
```suggestion
	IntoUTF8ArgsTranslator(int argc, char** argv);
```

src/libs/antares/args/include/antares/args/args_to_utf8.h
```diff
@@ -0,0 +1,11 @@
...
+	~IntoUTF8ArgsTranslator();
+private:
+	int argc_ = 0;
+	char**& argv_;
```
flomnes:
```suggestion
	char** argv_;
```

src/libs/antares/args/args_to_utf8.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#ifdef YUNI_OS_WINDOWS
+#include <string.h>
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN 1
+#endif
+#include <windows.h>
+#include <shellapi.h>
+#include <stdlib.h>
+#include <stdio.h>
```
flomnes:
```suggestion
```

Use `new` & `delete`

src/libs/antares/args/args_to_utf8.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "antares/args/args_to_utf8.h"
+
+
+#ifdef YUNI_OS_WINDOWS
+#include <string.h>
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN 1
+#endif
+#include <windows.h>
+#include <shellapi.h>
```
flomnes, line 12: It seems that these headers only make sense for Windows, right ?

flomnes (changes requested):
I suggest adding a convert method that returns a new `argv` as a `char**`

flomnes (approved):

flomnes (approved):

PR #1680, opened 2023-10-05
Fix/test 50 windows

a-zakir:
this Pr is intended to fix `long-tests-2/050 Export MPS File/
`


flomnes (changes requested):
As discussed, this deserves a bit of refactoring because the current code is error-prone.

flomnes (approved):
Revert windows-vcpkg.yml, check that long-tests-2 are passing and we're good to merge

PR #1684, opened 2023-10-05
Add & use IntoUTF8ArgsTranslator::convert

flomnes:


src/libs/antares/args/include/antares/args/args_to_utf8.h
```diff
@@ -1,11 +1,14 @@
 #pragma once
 
+#include <utility>
+
 class IntoUTF8ArgsTranslator
 {
 public:
-	IntoUTF8ArgsTranslator(int argc, char**& argv);
+	IntoUTF8ArgsTranslator(int argc, char** argv);
```
guilpier-code: It seems we have an indentation inconsistency here

src/libs/antares/args/include/antares/args/args_to_utf8.h
```diff
@@ -1,11 +1,14 @@
...
+    std::pair<int, char**> convert();
 	~IntoUTF8ArgsTranslator();
 private:
-	int argc_ = 0;
-	char**& argv_;
+	int argc_;
```
guilpier-code: I guess that, because these **private** member variables are set by the constructor (and because this constructor is the only constructor the class has), we don't need to give them a default value in the current class declaration, right ?

guilpier-code (changes requested):

src/libs/antares/args/include/antares/args/args_to_utf8.h
flomnes: The default is indeed no longer given

src/libs/antares/args/include/antares/args/args_to_utf8.h
flomnes: done (applied clang-format)

guilpier-code (approved):

Issue #1681, opened 2023-10-05
Exceptions handling in computation threads (QueueService)


sylvlecl:
**Description**

Since release 8.7, calls to `exit` have been replaced by exceptions.
Those exceptions are not correctly handled in computation threads which are managed by the `QueueService` class:
that class "swallows" silently all exceptions.

**How to Reproduce**

TODO

**Expected behavior**

Instead, exceptions should be caught and then thrown in the main computation thread.

In order to achieve this, we can use [std::future](https://en.cppreference.com/w/cpp/thread/future) to transfer that information between the computation thread and the calling thread.
For example:
```cpp
class QueueService{
    ...
    std::future<void> add(const IJob::Ptr& job);
    ...
};
```

To be assessed: is use of `std::async` relevant ?

PR #1685, opened 2023-10-06
Fixes swallowed exceptions in computation thread

sylvlecl:
See #1681 

Exceptions are now transferred from the computation thread to the main one using std::future mechanisms.
Only the first encountered exception is thrown in the main thread.

**Note**
As a side addition, the errors occurring during the writing of results to zip now cause exceptions that are also "transferred" to the main thread, when calling `IResultWriter::flush()`.


src/tests/src/libs/antares/concurrency/test_concurrency.cpp
```diff
@@ -0,0 +1,76 @@
...
+    FutureSet futures;
+    for (int i = 0; i < 10; i++) {
+        futures.add(AddTask(*threadPool, incrementCounter));
+    }
+    futures.join();
+    BOOST_CHECK(counter == 10);
+}
+
+class TestException1 {};
+class TestException2 {};
```
flomnes:
```suggestion
template<int n>
class TestException {};
```

src/tests/src/libs/antares/concurrency/test_concurrency.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+BOOST_AUTO_TEST_CASE(test_future_set_rethrows_first)
+{
+    auto threadPool = createThreadPool(2);
+    FutureSet futures;
+    futures.add(AddTask(*threadPool, failingTask<TestException1>()));
+    futures.add(AddTask(*threadPool, failingTask<TestException2>()));
+    BOOST_CHECK_THROW(futures.join(), TestException1);
```
flomnes: Is the task created at l. 73 guaranteed to be executed before the one at l. 74 ?

If there is no such guarantee, we can only expect that an exception among `TestException1` and `TestException2` will be thrown & caught, but not necessarily the 1st one.

src/libs/antares/concurrency/include/antares/concurrency/concurrency.h
```diff
@@ -0,0 +1,49 @@
...
+#define ANTARES_CONCURRENCY_H
+
+#include <future>
+#include "yuni/job/queue/service.h"
+
+namespace Antares::Concurrency
+{
+
+typedef std::function<void()> Task;
+typedef std::future<void> TaskFuture;
```
flomnes:
```suggestion
using Task = std::function<void()>;
using TaskFuture = std::future<void>;
```

src/libs/antares/concurrency/concurrency.cpp
```diff
@@ -0,0 +1,53 @@
...
+        task_();
+    }
+
+private:
+    std::packaged_task<void()> task_;
+};
+
+std::unique_ptr<Yuni::Job::IJob> MakeJob(const Task& task) {
+    return std::make_unique<JobImpl>(task);
+}
```
flomnes: unused

src/libs/antares/concurrency/concurrency.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+std::unique_ptr<JobImpl> MakePackagedJob(const Task& task) {
+    return std::make_unique<JobImpl>(task);
+}
```
flomnes: Only used internally, make static or place in anonymous NS

Note that in C++20 `auto` [is a valid type](https://stackoverflow.com/a/60355539) for methods.

src/libs/antares/concurrency/concurrency.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+std::future<void> AddTask(Yuni::Job::QueueService& threadPool, const Task& task) {
+    auto job = MakePackagedJob(task);
+    auto future = job->getFuture();
+    threadPool.add(Yuni::Job::IJob::Ptr(job.release()));
```
flomnes: Does this compile ?
```suggestion
    threadPool.add(job.release());
```

src/libs/antares/concurrency/concurrency.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+void FutureSet::add(TaskFuture&& f) {
+    futures_.push_back(std::move(f));
+}
+
+void FutureSet::join() {
+    for (auto& f: futures_) {
+        f.get();
+    }
```
flomnes: This might require a `mutex` to prevent concurrent access

src/tests/src/libs/antares/concurrency/test_concurrency.cpp
sylvlecl: The execution order is clearly not guaranteed, but the exception thrown will be the one corresponding to the first **submitted** task that fails, not the first exception that occurs.

src/libs/antares/concurrency/concurrency.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Antares::Concurrency
+{
+
+std::future<void> AddTask(Yuni::Job::QueueService& threadPool,
+                          const Task& task,
+                          Yuni::Job::Priority priority) {
+    auto job = std::make_unique<PackagedJob>(task);
+    auto future = job->getFuture();
+    threadPool.add(Yuni::Job::IJob::Ptr(job.release()), priority);
```
flomnes: Being the devil's advocate, why bother using a `std::unique_ptr` if you `release` it 2 lines after ?

```suggestion
    PackageJob* job = new PackagedJob(task);
    auto future = job->getFuture();
    threadPool.add(Yuni::Job::IJob::Ptr(job), priority);
```

src/solver/simulation/solver.hxx
```diff
@@ -45,11 +45,15 @@
 #include <yuni/core/system/suspend.h>
 #include <yuni/job/job.h>
 
+#include "antares/concurrency/concurrency.h"
+
 namespace Antares::Solver::Simulation
 {
 
+using namespace Antares::Concurrency;
```
flomnes: You could be a bit more specific and name what classes/function you'll be using

src/libs/antares/writer/zip_writer.cpp
```diff
@@ -46,7 +46,7 @@ static std::unique_ptr<mz_zip_file> createInfo(const std::string& entryPath)
 }
 
 template<class ContentT>
-void ZipWriteJob<ContentT>::onExecute()
+void ZipWriteJob<ContentT>::writeEntry()
```
flomnes, line 62: I see a few `logs.error()` below. Could we also `throw` exceptions now ?

sylvlecl: Indeed: it will change a little the behaviour from what it is now, but I think it's better to get an error instead of having an incomplete/corrupt result without knowing it (who looks at the log? :) ).

src/libs/antares/writer/private/zip_writer.hxx
```diff
@@ -11,8 +12,9 @@ void ZipWriter::addEntryFromBufferHelper(const std::string& entryPath, ContentTy
         return;
 
     EnsureQueueStartedIfNeeded ensureQueue(this, pQueueService);
-    pQueueService->add(
-      new ZipWriteJob<ContentType>(*this, entryPath, entryContent, pDurationCollector),
-      Yuni::Job::priorityLow);
+    Antares::Concurrency::AddTask(*pQueueService,
+                                  ZipWriteJob<ContentType>(*this, entryPath, entryContent, pDurationCollector),
+                                  Yuni::Job::priorityLow);
```
flomnes: Unfortunately, the result of this is not stored in a `FutureSet`, so we won't be able to catch exceptions.

Maybe we could wrep `QueueService` and `FutureSet` together and expose a simpler `add` method ?

```cpp
class Wrapper {
Yuni::QueueService* queue;
FutureSet results;
public:
add(...);
};
```
Note that `Wrapper` may be returned by `createThreadPool` that is defined in the tests, instead of a mere QueueService.

src/libs/antares/concurrency/concurrency.cpp
sylvlecl: After more thoughts, I actually added a mutex to make all methods thread safe, in particular the addition of new futures, since vector are not thread safe.

src/libs/antares/concurrency/concurrency.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::lock_guard lock(mutex_);
+    futures_.push_back(std::move(f));
+}
+
+void FutureSet::join() {
+    std::vector<TaskFuture> moved;
+    {
+        std::lock_guard lock(mutex_);
+        moved = std::move(futures_);
+        futures_ = std::vector<TaskFuture>{};
```
flomnes:
```suggestion
        std::swap(moved, futures_);
```

src/libs/antares/writer/private/zip_writer.hxx
sylvlecl: Indeed the simple call to `AddTask` is not useful by itself.

What I did to cope with it is:
- Addition of [[nodiscard]] to the method to be warned when we ignore the future
- Addition of a `flush()` method to the writer, in order to be able to synchronize a thread with underlying I/O tasks, and get possible exception from here

The writer does not use the same thread pool as computations any more, but its own one, which is never started/stopped.
The only synchonization mechanism used for I/O now the flush method.

src/libs/antares/writer/zip_writer.cpp
```diff
@@ -82,12 +96,14 @@ ZipWriter::ZipWriter(std::shared_ptr<Yuni::Job::QueueService> qs,
  pQueueService(qs),
  pState(ZipState::can_receive_data),
  pArchivePath(std::string(archivePath) + ".zip"),
- pDurationCollector(duration_collector)
+ pDurationCollector(duration_collector),
+ pendingTasks_()
```
flomnes: Calling the default constructor is not necessary, it will be called implicitly

src/libs/antares/writer/zip_writer.cpp
```diff
@@ -82,12 +96,14 @@ ZipWriter::ZipWriter(std::shared_ptr<Yuni::Job::QueueService> qs,
...
 {
     mz_zip_writer_create(&pZipHandle);
     if (int32_t ret = mz_zip_writer_open_file(pZipHandle, pArchivePath.c_str(), 0, 0); ret != MZ_OK)
     {
         logs.error() << "Error opening zip file " << pArchivePath << " (" << ret << ")";
+        logErrorAndThrow("Error opening zip file " + pArchivePath + " (" + std::to_string(ret) + ")");
```
flomnes, line 104: Please remove the previous line (redundant)

src/tests/src/libs/antares/writer/test_zip_writer.cpp
flomnes, line 1: A nice addition !

sylvlecl: It's a little clumsy though, with the minizip API !
If we use more zip in the future, it would be nice to add a small c++ wrapper around that C API :)

Also I did not really find a way to cause a failure in the writing of entries ... if you have any idea about this it's welcome.

src/tests/src/libs/antares/writer/test_zip_writer.cpp
```diff
@@ -0,0 +1,106 @@
...
+using Benchmarking::IDurationCollector;
+using Antares::Solver::IResultWriter;
+
+// Handles lifetime of necessary objects
+struct TestContext
+{
+    std::shared_ptr<QueueService> threadPool;
+    std::unique_ptr<IDurationCollector> durationCollector;
+    std::shared_ptr<IResultWriter> writer;
+};
```
flomnes, line 57: Fixture ?

src/tests/src/libs/antares/writer/test_zip_writer.cpp
flomnes: Providing negative `len` maybe ? Or an invalid path ?

src/tests/src/libs/antares/writer/test_zip_writer.cpp
sylvlecl: I did not investigate it so much, but I cannot easily find a good use here: 
to create those objects I need to pass some parameters (path, threads). Fixtures are limited to only one constructor argument. It could be a small struct but it becomes weird...

Those arguments are not strictly necessary, but would be useful for additional tests.

src/libs/antares/writer/zip_writer.cpp
sylvlecl: Removed

src/libs/antares/writer/zip_writer.cpp
sylvlecl: Removed

src/libs/antares/concurrency/concurrency.cpp
sylvlecl: Done indeed!

src/tests/src/libs/antares/concurrency/test_concurrency.cpp
sylvlecl: Done

src/solver/application.cpp
```diff
@@ -281,8 +281,11 @@ void Application::processCaption(const Yuni::String& caption)
 void Application::prepareWriter(Antares::Data::Study& study,
                                 Benchmarking::IDurationCollector& duration_collector)
 {
+    ioQueueService = std::make_shared<Yuni::Job::QueueService>();
```
sylvlecl, line 284: @flomnes : Note that the zip writing is now handled by this separate thread pool which is allocated only 1 thread.

Since I don't do "introspection" on the type of writer used, the thread is created also when using plain files, and not used, which was not the case before.
The runtime negligible cost seems reasonable to me.

flomnes: You could use `IResultWriter::needsTheJobQueue()`, or even make `EnsureQueueStartedIfNeeded` public and use it.

sylvlecl: The writer is not yet created when we need to know this:
we can instead have a factory interface (`ResultWriterFactory` with 2 implementations) which provides that information before creation of the writer, but honestly I think this is overkill, having a sleeping thread is really not an issue.

It would make responsibilities clearer to leave it as is: the application provides a usable thread pool, the writer has the choice to use it or not.
Actually, I would also be in favor of removing `EnsureQueueStartedIfNeeded` completely, together with the `needsTheJobQueue()` in the interface, to make things simpler. 
It's the responsibility of the application to provide a working thread pool, the writer should not need to worry about starting anything.

sylvlecl: Or alternatively the writer could maintain its own working thread, but it would make it quite "hidden".

flomnes: Ok, let us keep it like that

flomnes (approved):

src/libs/antares/concurrency/include/antares/concurrency/concurrency.h
```diff
@@ -0,0 +1,88 @@
...
+ * This allows to handle exceptions occuring in the underlying task,
+ * as opposite to Yuni::Job::QueueService::add which swallows them.
+ */
+[[nodiscard]] TaskFuture AddTask(Yuni::Job::QueueService& threadPool,
+                   const Task& task,
+                   Yuni::Job::Priority priority = Yuni::Job::priorityDefault);
+
+/*!
+ * \brief Utility class to gather futures to wait for.
+ */
```
a-zakir, line 51: yeah pretty doc ;)

PR #1686, opened 2023-10-06
Use generic class for load series

payetvin:
#1669


src/libs/antares/study/parts/load/container.cpp
```diff
@@ -34,46 +34,40 @@ using namespace Yuni;
...
-    return sizeof(Container) + ((!series) ? 0 : DataSeriesLoadMemoryUsage(series))
-           + ((!prepro) ? 0 : prepro->memoryUsage());
+    return sizeof(Container) + series.memoryUsage() + ((!prepro) ? 0 : prepro->memoryUsage());
 }
 
 void Container::resetToDefault()
 {
-    if (series)
-        series->timeSeries.reset(1, HOURS_PER_YEAR);
+    series.reset();
```
flomnes, line 70: Keep arguments ?

src/libs/antares/study/parts/load/container.h
```diff
@@ -27,7 +27,7 @@
 #ifndef __ANTARES_LIBS_STUDY_PARTS_LOAD_CONTAINER_H__
 #define __ANTARES_LIBS_STUDY_PARTS_LOAD_CONTAINER_H__
 
-#include "series.h"
+#include <antares/study/parts/common/series.h>
```
flomnes, line 30: How about
```suggestion
#include <antares/study/series.h>
```
?

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -96,8 +96,8 @@ static bool Remix(const Data::AreaList& areas, PROBLEME_HEBDO& problem, uint num
             double bottom = std::numeric_limits<double>::max();
             double top = 0;
 
-            uint loadTS = area.load.series->getIndex(problem.year);
-            auto& load = area.load.series->timeSeries;
+            uint loadTS = area.load.series.getSeriesIndex(problem.year);
+            auto& load = area.load.series.timeSeries;
```
flomnes, line 100: We could use `getColumn` here

src/libs/antares/study/parts/load/container.h
payetvin: doesn't work

PR #1687, opened 2023-10-06
New log msg when solver not found in or-tools

guilpier-code:


src/solver/utils/ortools_utils.cpp
```diff
@@ -364,11 +364,16 @@ MPSolver* MPSolverFactory(const Antares::Optimization::PROBLEME_SIMPLEXE_NOMME*
             solver = MPSolver::CreateSolver((OrtoolsUtils::solverMap.at(solverName)).LPSolverName);
 
         if (!solver)
-            throw Antares::Data::AssertionError("Solver not found: " + solverName);
+        {
+            std::string msg_to_throw = "Solver " + solverName + " not found. \n";
+            msg_to_throw += "Please make sure that your OR-Tools install supports solver " + solverName + ".";
+
+            throw Antares::Data::AssertionError(msg_to_throw);
+        }
```
flomnes, line 372: I don't think anything is displayed in the logs ? Am I wrong ?

guilpier-code: Yes, I checked that it was actually printed

flomnes (approved):

Issue #1682, opened 2023-10-05
Check variable bounds in the infeasible problems analyzer


flomnes:
In the infeasible problem analyzer, check that for every variable `var`,

> var->lb() <= var->ub()

If not, display the name of the variable and the bounds in the logs.

guilpier-code:
Maybe we could have something like : 
```C++
// opt_appel_solveur_lineaire.cpp
...
InfeasibleProblemAnalysis analysis(options.solverName, &Probleme);
try
{
    analysis.run(); // Here we create the modified problem and run it
    analysis.createReport(); // Here we just create a list of reports (all with the same interface)
    analysis.printReport(); // Here we only loop over the list of reports to pretty print them
}
catch (/* ...  */)
{
    ....
}
...
```
@flomnes : What do you think ?

flomnes:
Yes I agree. Only one report is required, it would make no sense to run 2 analyses (constraints & variable bounds).

payetvin:
done

PR #1689, opened 2023-10-06
Unfeasible problem analyzer : add a new check

guilpier-code:
The current PR to fix issue #1682

what's done here : 
- [ ] Make a refactoring before welcoming the change (add a new analysis)
   - [x] Remove exceptions catching a problem in the analyzer itself. Removing theses exceptions simplifies the code.
   - [x] Simplify analyzer's interface to simplify the client calls.
   - [x] split analyzer into : an algorithm that manages a collection of analyses and the slack variables analysis. Note that split and separation makes code more testable (in isolation).
   - [x] Currently, when slack variables analysis cannot detect the unfeasibiliy's reason, it prints an error message. This can be misleading for users because it's not exactly an error but an inability of one analysis to catch the unfeasibility's reason. There may  be another analysis that can catch it. So these messages can be printed as **warnings** or **notice**.
   - [x] Separate the analysis manager and every analysis in different sources files 
   - [x] **add some unit tests** : test each class in isolation (units) and add one or two integration tests to check everything work together
   - [ ] what else ?
- [x] add the new check on consistency of each (original) variable bounds


a-zakir (approved):

src/solver/infeasible-problem-analysis/report.h
```diff
@@ -7,6 +7,8 @@
 #include "constraint.h"
 #include "ortools/linear_solver/linear_solver.h"
 
+using namespace operations_research;
+
```
flomnes: Adding this at the top of header file is considered bad practice

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -11,80 +11,179 @@ namespace Antares
...
+}
+
+void VariablesBoundsAnalysis::storeIncorrectVariable(std::string name, double lowBound, double upBound)
+{
+    incorrectVars_.push_back(VariableBounds(name, lowBound, upBound));
+}
+
+bool VariablesBoundsAnalysis::foundIncorrectVariables()
+{
+    return (incorrectVars_.size() != 0);
```
flomnes:
```suggestion
    return !incorrectVars_.empty();
```

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -11,80 +11,179 @@ namespace Antares
...
 {
-    addSlackVariables();
-    if (mSlackVariables.empty())
+    InfeasibleProblemReport report(slackVariables_);
+    report.prettyPrint();
+}
+
+// ============================
+// Variables bounds analysis
+// ============================
```
flomnes: I strongly suggest splitting this file, and renaming the new files accordingly (constraint slack analysis, variable bound check)

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -11,80 +11,179 @@ namespace Antares
...
+// ===============================
+
+// gp : this class should be renamed into UnfeasibilityAnalyzer
+
+InfeasibleProblemAnalysis::InfeasibleProblemAnalysis(const std::string& solverName, const PROBLEME_SIMPLEXE_NOMME* ProbSpx)
+{
+    // gp : Here we have a dependency on PROBLEME_SIMPLEXE_NOMME and MPSolver.
+    // gp : We should Convert() the PROBLEME_SIMPLEXE_NOMME into a MPSolver outside (and before) this constructor.
+    // gp : And so we should have a MPSolver* passed here.
+    // gp : It would be easier to test this class in isolation.
```
flomnes: I agree, but it's actually quite convenient to have this intermediate layer (probleme_simplexe), since class MPSolver has deleted copy constructor & operator.

If we passed the original problem, we'd need to make a copy. Although it is possible, it can be quite time-consuming.

src/solver/infeasible-problem-analysis/problem.cpp
```diff
@@ -11,80 +11,179 @@ namespace Antares
...
+void InfeasibleProblemAnalysis::printReport()
+{
+    for (auto& analysis : analysisList_)
+    {
+        if (analysis->hasDetectedInfeasibilityCause())
+        {
+            logs.info() << analysis->title() << " : printing report";
+            analysis->printReport();
+            return;
+        }
```
flomnes: As discussed, the code looks much clearer without a list
```suggestion
    VariableBoundsChecker varChecker(problem_);
    if (varChecker.infeasibilityDetected())
    {
         // Handle variable infeasibility, for example
         varChecker.printReport();
    }
    else
    {
        ConstraintCheckerUsingSlackVariables constraintChecker(problem_);
        if (constraintChecker.infeasibilityDected())
        {
            // Handle this infeasibility, for example
            constraintChecker.printReport();
        }
   }
```

src/solver/infeasible-problem-analysis/problem.h
```diff
@@ -11,22 +11,77 @@ namespace Antares
...
+};
+
+
+class SlackVariablesAnalysis : public SingleAnalysis
+{
+    using SingleAnalysis::SingleAnalysis;
+public:
+    void run() override;
+    void printReport() override;
+    std::string title() override { return "Slack variables analysis"; }
```
flomnes: This interface class isn't really needed

src/solver/infeasible-problem-analysis/problem.cpp
sylvlecl: For more flexibility in the use of this class and to break the dependency, we can use a supplier of MPSolver as an argument: `std::function<std::shared_ptr<MPSolver>()>`.

We can then use it either with a raw MPSolver, in unit tests, or with a PROBLEME_SIMPLEXE_NOMME that will be converted as needed (by a lambda for ex):

```cpp
PROBLEME_SIMPLEXE_NOMME* problem = ...;
InfeasibleProblemAnalysis analysis("XPRESS", [problem]() {
    return std::shared_ptr<MPSolver>(ProblemSimplexeNommeConverter(solverName, ProbSpx).Convert()
});
```

src/solver/infeasible-problem-analysis/problem.h
```diff
@@ -11,22 +11,77 @@ namespace Antares
...
+
+
+class SingleAnalysis
 {
 public:
-    InfeasibleProblemAnalysis() = delete;
-    explicit InfeasibleProblemAnalysis(const std::string& solverName, const PROBLEME_SIMPLEXE_NOMME* ProbSpx);
-    InfeasibleProblemReport produceReport();
+    SingleAnalysis(std::shared_ptr<operations_research::MPSolver> problem_);
+    virtual void run() = 0;
```
sylvlecl: Not a must-have, but I would advise a more functional style, this interface does not need to be stateful (does not need to manage data members):
```cpp
virtual bool run(operations_research::MPSolve& problem) = 0;
// no data member
```

src/solver/infeasible-problem-analysis/problem.h
```diff
@@ -11,22 +11,77 @@ namespace Antares
...
-class InfeasibleProblemAnalysis
+
+
+class SingleAnalysis
 {
 public:
-    InfeasibleProblemAnalysis() = delete;
-    explicit InfeasibleProblemAnalysis(const std::string& solverName, const PROBLEME_SIMPLEXE_NOMME* ProbSpx);
-    InfeasibleProblemReport produceReport();
+    SingleAnalysis(std::shared_ptr<operations_research::MPSolver> problem_);
```
sylvlecl: no need for shared ownership here, reference or raw pointer seems more appropriate

src/solver/infeasible-problem-analysis/problem.h
```diff
@@ -11,22 +11,77 @@ namespace Antares
...
+    std::vector<VariableBounds> incorrectVars_;
+};
+
+class InfeasibleProblemAnalysis
+{
+public:
+    InfeasibleProblemAnalysis() = delete;
+    explicit InfeasibleProblemAnalysis(const std::string& solverName, const PROBLEME_SIMPLEXE_NOMME* ProbSpx);
+    void run();
+    void printReport();
```
sylvlecl: Would be better to return a report object from the `run` method, to make the workflow more natural and separate roles:
```cpp
auto report = analysis.run();
report.print();
```

src/solver/infeasible-problem-analysis/problem.h
sylvlecl: Instead of the boolean, it could actually return a richer "report part" object, to be concatenated in the final report.

sylvlecl (commented):
Suggestions for further improvement on the design

src/solver/infeasible-problem-analysis/report.h
guilpier-code: I replaced **operations_research::MPVariable** with **MPVariable**.
As the second is clear enough, and the first cumbersome, I  found this better.
I will change back if you insist.

src/solver/infeasible-problem-analysis/problem.cpp
guilpier-code: Better indeed. Done.

src/solver/infeasible-problem-analysis/problem.cpp
guilpier-code: Yes I will. 
Done.

src/solver/infeasible-problem-analysis/problem.cpp
guilpier-code: @flomnes : I still think that creating a **MPSolver** from a **PROBLEME_SIMPLEXE_NOMME** is not the concern of the unfeasible problem analyzer. This analyzer should only receive an optimization problem and handle several analysis on this problem. So I would recommend to pass to the constructor at most a problem (see comment below for a change on this).

@sylvlecl : We talked about your suggestion. It's useful if we have to regenerate a fresh copy of the original problem in case an analysis modified this problem (as it's the case for slack variables analysis), and another analysis has to do its work after that.
But it's not the case for now. I would reserve the fresh copy creation handling for a possible future change that really requires it. For now we respect **YAGNI**.

src/solver/infeasible-problem-analysis/problem.cpp
guilpier-code: I'm sorry to disagree, for the following reasons : 
- In this code suggested here, this is a duplication : both analysis are called in the same way.
- The analyzer depends on the two analysis, and cannot properly be tested in isolation.
- The analyzer'interface has changed, the analysis part is no more separated from reporting part. I guess it's a question of taste : client of the analyzer could prefer (or not) that the 2 steps are done in one glance.

src/solver/infeasible-problem-analysis/problem.h
guilpier-code: See answer above

src/solver/infeasible-problem-analysis/problem.h
guilpier-code: Please see comments below for this matter.

src/solver/infeasible-problem-analysis/problem.h
guilpier-code: - About the **problem** as an argument of the **run(...)** method  : 
  In the current state of the code, we have 3 concrete classes : an analyzer (which supervises the whole analysis), and the 2 different analyses. Each analysis owns the problem, but the analyzer does not see the problem at all : it just knows if an analysis failed to find the reason of unfeasibility or succeeded, and if it must stop the analyse there or run another one. 
Should the analyzer see the problem ? At least it's a dependency it does not have.
So in this context, passing the problem to an analysis run function is impossible.
- About the **returned report** :  
  Currently the analyzer has the given interface  : **void run()** and **void printReport()**. So a client of the analyser needs to call the 2 public methods. Splitting the the whole analyze in 2 parts can make the code both clear (not too many calls) and informative (we know more precisely what the analyzer do). This can be discussed though.
But if we choose the leave the analyzer as it is, it may be neither easy nor consistent that the run method returns a report.

src/solver/infeasible-problem-analysis/problem.h
guilpier-code: Done

src/solver/infeasible-problem-analysis/report.h
sylvlecl: We indeed need to remove it because this impacts all files which will include this file.
We don't want to promote this namespace in the global namespace for all those user files.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -375,21 +376,16 @@ bool OPT_AppelDuSimplexe(const OptimizationOptions& options,
...
-            logs.error() << ex.what();
-        }
-        catch (const Optimization::ProblemResolutionFailed& ex)
-        {
-            logs.error() << ex.what();
-        }
+
+        auto MPproblem = std::shared_ptr<MPSolver>(ProblemSimplexeNommeConverter(options.solverName, &Probleme).Convert());
+
+        std::vector<std::shared_ptr<UnfeasibilityAnalysis>> analysisList;
```
sylvlecl: I think it's more a responsibility of the unfeasibility analysis library to build the analyzer for a given problem.
We should not need to worry here of which check to be added in which order.
We can add a function for this in the the unfeasibility analysis lib.

src/solver/infeasible-problem-analysis/unfeasibility-analysis.h
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <memory>
+#include "ortools/linear_solver/linear_solver.h"
+
+namespace Antares::Optimization
+{
+class UnfeasibilityAnalysis
```
sylvlecl, line 12: We could add some documentation comments on the few classes we added, to clarify their role

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -375,21 +376,16 @@ bool OPT_AppelDuSimplexe(const OptimizationOptions& options,
...
+        analysisList.push_back(std::make_shared<VariablesBoundsConsistency>(MPproblem.get()));
+        analysisList.push_back(std::make_shared<ConstraintSlackAnalysis>(MPproblem.get()));
+
+        Optimization::UnfeasiblePbAnalyzer analyzer(analysisList);
+        analyzer.run();
```
sylvlecl: I still think we must add the MPSolver as an argument to `run(problem)`.
Actually, the API design is wrong, because we could provide different problems to the different steps of the analysis:
```cpp
analysisList.push_back(std::make_shared<VariablesBoundsConsistency>(problem1));
analysisList.push_back(std::make_shared<ConstraintSlackAnalysis>(problem2));
```
which does not make sense.

Instead, we should only pass the problem once, into the run method:
```cpp
Optimization::UnfeasiblePbAnalyzer analyzer(analysisList);
analyzer.run(problem);
```

src/solver/infeasible-problem-analysis/variables-bounds-consistency.h
```diff
@@ -0,0 +1,33 @@
...
+    VariableBounds(std::string var_name, double low_bound, double up_bound)
+        : name(var_name), lowBound(low_bound), upBound(up_bound)
+    {}
+
+    std::string name;
+    double lowBound;
+    double upBound;
+};
+
+class VariablesBoundsConsistency : public UnfeasibilityAnalysis
```
sylvlecl, line 23: Can we add unit tests to this module, like a few tests for each component, the 2 analysis and the analyzer?

src/solver/infeasible-problem-analysis/unfeasible-pb-analyzer.h
```diff
@@ -0,0 +1,24 @@
...
+class UnfeasiblePbAnalyzer
+{
+public:
+    UnfeasiblePbAnalyzer() = delete;
+    explicit UnfeasiblePbAnalyzer(std::vector<std::shared_ptr<UnfeasibilityAnalysis>> analysisList);
+    void run();
+    void printReport();
+
+private:
+    std::vector<std::shared_ptr<UnfeasibilityAnalysis>> analysisList_;
```
sylvlecl: We could probably have unique_ptr here instead of shared ones, the analyzer really owns the underlying analysis, I guess.

src/solver/infeasible-problem-analysis/constraint-slack-analysis.h
```diff
@@ -0,0 +1,25 @@
...
+    void printReport() override;
+    std::string title() override { return "Slack variables analysis"; }
+
+private:
+    void buildObjective() const;
+    void addSlackVariables();
+    operations_research::MPSolver::ResultStatus Solve() const;
+
+    std::vector<const operations_research::MPVariable*> slackVariables_;
+    const std::string constraint_name_pattern = "^AreaHydroLevel::|::hourly::|::daily::|::weekly::|^FictiveLoads::";
```
sylvlecl, line 27: Should be `constexpr`

src/solver/infeasible-problem-analysis/variables-bounds-consistency.cpp
```diff
@@ -0,0 +1,41 @@
+#include "variables-bounds-consistency.h"
+#include <antares/logs/logs.h>
+
+namespace Antares::Optimization
+{
+    
+void VariablesBoundsConsistency::run()
+{
+    for (auto& var : problem_->variables())
```
sylvlecl: const

src/solver/infeasible-problem-analysis/variables-bounds-consistency.cpp
```diff
@@ -0,0 +1,41 @@
...
+    {
+        double lowBound = var->lb();
+        double upBound = var->ub();
+        std::string name = var->name();
```
sylvlecl, line 15: const ref will avoid a copy

src/solver/infeasible-problem-analysis/variables-bounds-consistency.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+bool VariablesBoundsConsistency::foundIncorrectVariables()
+{
+    return !incorrectVars_.empty();
+}
+
+void VariablesBoundsConsistency::printReport()
+{
+    for (auto& var : incorrectVars_)
```
sylvlecl: const

sylvlecl (changes requested):
a few more design issue + unit tests would be nice :)

src/solver/infeasible-problem-analysis/report.h
guilpier-code: ok, I will.
Done.

src/solver/infeasible-problem-analysis/problem.h
guilpier-code: Done.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp

src/solver/infeasible-problem-analysis/variables-bounds-consistency.h
guilpier-code: Yes we can. It should be the subject of a further issue / PR.

src/solver/infeasible-problem-analysis/constraint-slack-analysis.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Antares::Optimization
+{
+
+class ConstraintSlackAnalysis : public UnfeasibilityAnalysis
+{
+public:
+    ConstraintSlackAnalysis() = default;
+    void run(operations_research::MPSolver* problem) override;
+    void printReport() override;
```
a-zakir: could be const

src/solver/infeasible-problem-analysis/constraint-slack-analysis.h
```diff
@@ -0,0 +1,24 @@
...
+    std::string title() override { return "Slack variables analysis"; }
```
a-zakir: this one also

a-zakir:
Not that i'm against it but the new file's name are not consistent with other lib name convention (we generally don't have hyphens)

src/solver/infeasible-problem-analysis/unfeasible-pb-analyzer.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include <antares/logs/logs.h>
+
+
+using namespace operations_research;
+
+namespace Antares::Optimization
+{
+
+UnfeasiblePbAnalyzer::UnfeasiblePbAnalyzer(std::vector<std::shared_ptr<UnfeasibilityAnalysis>> analysisList)
+        : analysisList_(analysisList)
```
a-zakir: is there a reason to pass `analysisList` by value?

src/solver/infeasible-problem-analysis/unfeasible-pb-analyzer.cpp
```diff
@@ -0,0 +1,49 @@
...
+        logs.info() << analysis->title() << " : running...";
+        analysis->run(problem);
+        if (analysis->hasDetectedInfeasibilityCause())
+            return;
+
+        logs.notice() << analysis->title() << " : nothing detected.";
+    }
+}
+
+void UnfeasiblePbAnalyzer::printReport()
```
a-zakir: if it's just about "printing" things then we can const this method

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
guilpier-code: Done, under the form of a factory function, added to files **unfeasible-pb-analyzer.h/.cpp**

src/solver/infeasible-problem-analysis/unfeasible-pb-analyzer.h
guilpier-code: Done.

src/solver/infeasible-problem-analysis/variables-bounds-consistency.cpp
guilpier-code: Done

src/solver/infeasible-problem-analysis/variables-bounds-consistency.cpp
guilpier-code: Done.

src/solver/infeasible-problem-analysis/variables-bounds-consistency.cpp
guilpier-code: Done.

src/solver/infeasible-problem-analysis/constraint-slack-analysis.h
guilpier-code: Done for **printReport()**, now a **const** method

src/solver/infeasible-problem-analysis/constraint-slack-analysis.h
guilpier-code: Done for **title()** and **printReport()**, now **const** methods

src/solver/infeasible-problem-analysis/unfeasible-pb-analyzer.cpp
guilpier-code: **analysisList** is a **vector** of unique  pointers, copying this vector is not copying the objects it contains but only the pointers. And to be accurate, they are not **copied**, but moved (see call of **std::move(...)**)

src/solver/infeasible-problem-analysis/unfeasible-pb-analyzer.cpp
guilpier-code: You're right. Done.

src/solver/infeasible-problem-analysis/constraint-slack-analysis.h
guilpier-code: Impossible to associate **constexpr** to a string, compilation refuses. Even **static const** does not compile.

a-zakir (approved):

src/solver/infeasible-problem-analysis/variables-bounds-consistency.h
sylvlecl: Will be done in https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1716

sylvlecl (approved):

PR #1693, opened 2023-10-09
Refactor thermal & renewable series

payetvin:
#1669


src/libs/antares/checks/checkLoadedInputData.cpp
```diff
@@ -142,7 +142,7 @@ void checkMinStablePower(bool tsGenThermal, const Antares::Data::AreaList& areas
 // TS
 template<class ExceptionT>
 static void checkThermalColumnNumber(const Antares::Data::AreaList& areas,
-                                     Matrix<double> Antares::Data::EconomicInputData::*matrix)
+                                     Matrix<double, int32_t> Antares::Data::EconomicInputData::*matrix)
```
flomnes: As discussed, all matrices should be `Matrix<double>`

src/libs/antares/study/parts/common/cluster.h
```diff
@@ -118,7 +118,7 @@ class Cluster
     uint areaWideIndex;
 
     //! Series
-    DataSeriesCommon* series;
+    TimeSeries series;
```
flomnes, line 121: - `series` is a confusing name, I'd rather use `availability` or something like that because thermal clusters have multiple `series`
- As discussed, we should have a single `TimeSeriesNumbers`, shared by multiple `TimeSeries` (availability, market bid cost, fuel cost, CO2 cost, etc.). Please make changes accordingly.

flomnes (commented):
Please consider using a typedef for the matrix type, it will allow you to change it in one line of code

flomnes (commented):
Use `TimeSeries::getCoefficient`, remove `RenewableCluster::valueAtTimeStep` completely.

src/libs/antares/study/parts/common/series.h
```diff
@@ -38,7 +38,7 @@ class TimeSeries
 {
 public:
     using TSNumbers = Matrix<uint32_t>;
-    using TScoefficients = Matrix<double, int32_t>;
+    using TS = Matrix<double>;
```
flomnes, line 39: Use this wherever `Matrix<double>` is used

Also, `double` is the default value for this template argument, so `Matrix<>` should be enough.

src/tests/end-to-end/simple_study/simple-study.cpp
```diff
@@ -24,7 +24,7 @@ struct StudyFixture : public StudyBuilder
 	double loadInArea = 0.;
 	double clusterCost = 0.;
 	ThermalClusterConfig clusterConfig;
-	TimeSeriesConfigurer<Matrix<double, int32_t>> loadTSconfig;
+	TimeSeriesConfigurer<Matrix<double>> loadTSconfig;
```
flomnes: Considering all matrices are `Matrix<double>`, I think we could remove this template argument, and so remove this template as well.

PR #1699, opened 2023-10-11
Fix regression on cluster renaming, add unit tests

### LINKED ISSUE #1697: GUI : impossible to rename thermal and renewable clusters
We have a regression : 
When creating a cluster from the GUI, impossible to give it the name we want : it will be saved as **new cluster**.
And if we want to have another cluster (in the same area), it's **new cluster 2**, no choice again.

Observed on : 
- **Windows**.
- **Antares** v8.7.0

sylvlecl:
**Description**
See #1697:
a regression was introduced in #1546, which changed the implementation of cluster names (from yuni strings to std strings), making it impossible to renamed thermal and renewable clusters in the GUI.

The method `Study::clusterRename` was not properly working anymore.

**Solution**
Fixed the renaming logic, and added some basic unit tests for operations on areas and clusters in the study.

Note: I also checked for other usage of `empty` in the PR that caused the regression, and found none.

### DIFF
--- a/src/tests/src/libs/antares/study/CMakeLists.txt
+++ b/src/tests/src/libs/antares/study/CMakeLists.txt
@@ -6,3 +6,13 @@ add_subdirectory(output-folder)
 add_subdirectory(short-term-storage-input)
 add_subdirectory(thermal-price-definition)
 add_subdirectory(constraint)
+
+add_executable(test-study)
+target_sources(test-study PRIVATE test_study.cpp)
+target_link_libraries(test-study
+        PRIVATE
+        Antares::study
+        Boost::unit_test_framework
+)
+
+add_test(NAME test-study COMMAND test-study)
--- a/src/tests/src/libs/antares/study/test_study.cpp
+++ b/src/tests/src/libs/antares/study/test_study.cpp
@@ -1,0 +1,95 @@
+/*
+** Copyright 2007-2023 RTE
+** Authors: Antares_Simulator Team
+**
+** This file is part of Antares_Simulator.
+**
+** Antares_Simulator is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** There are special exceptions to the terms and conditions of the
+** license as they are applied to this software. View the full text of
+** the exceptions in file COPYING.txt in the directory of this software
+** distribution
+**
+** Antares_Simulator is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with Antares_Simulator. If not, see <http://www.gnu.org/licenses/>.
+**
+** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
+*/
+
+#define BOOST_TEST_MODULE study
+#define BOOST_TEST_DYN_LINK
+#define WIN32_LEAN_AND_MEAN
+#include <boost/test/unit_test.hpp>
+
+#include "antares/study/study.h"
+
+using namespace Antares::Data;
+
+BOOST_AUTO_TEST_CASE(basic_areas_operations)
+{
+    Study study;
+    Area* areaA = study.areaAdd("A");
+    BOOST_CHECK(areaA != nullptr);
+    BOOST_CHECK(areaA->name == "A");
+    BOOST_CHECK(areaA->id == "a");
+    BOOST_CHECK(study.areaRename(areaA, "B"));
+    BOOST_CHECK(areaA->name == "B");
+    BOOST_CHECK(areaA->id == "b");
+    BOOST_CHECK(study.areaDelete(areaA));
+    BOOST_CHECK(study.areas.empty());
+}
+
+BOOST_AUTO_TEST_CASE(basic_thermal_clusters_operations)
+{
+    Study study;
+    Area& area = *study.areaAdd("A");
+    auto newCluster = std::make_shared<ThermalCluster>(&area);
+    newCluster->setName("Cluster");
+    BOOST_CHECK(newCluster->name() == "Cluster");
+    BOOST_CHECK(newCluster->id() == "cluster");
+
+    area.thermal.list.add(newCluster);
+    BOOST_CHECK(area.thermal.list.find("cluster") == newCluster.get());
+    BOOST_CHECK(area.thermal.list.find("Cluster") == nullptr);
+
+    //Renaming
+    BOOST_CHECK(study.clusterRename(newCluster.get(), "Renamed"));
+    BOOST_CHECK(newCluster->name() == "Renamed");
+    BOOST_CHECK(newCluster->id() == "renamed");
+
+    area.thermal.list.remove("renamed");
+    BOOST_CHECK(area.thermal.list.find("renamed") == nullptr);
+    BOOST_CHECK(area.thermal.list.empty());
+}
+
+BOOST_AUTO_TEST_CASE(basic_renewable_clusters_operations)
+{
+    Study study;
+    Area& area = *study.areaAdd("A");
+    auto newCluster = std::make_shared<RenewableCluster>(&area);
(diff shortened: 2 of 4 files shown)


a-zakir (approved):

src/tests/src/libs/antares/study/CMakeLists.txt
```diff
@@ -6,3 +6,13 @@ add_subdirectory(output-folder)
 add_subdirectory(short-term-storage-input)
 add_subdirectory(thermal-price-definition)
 add_subdirectory(constraint)
+
```
guilpier-code, line 9: Question : why not make a new test directory for testing basic study functionalities ?

src/tests/src/libs/antares/study/test_study.cpp
```diff
@@ -0,0 +1,95 @@
...
+#define BOOST_TEST_MODULE study
+#define BOOST_TEST_DYN_LINK
+#define WIN32_LEAN_AND_MEAN
+#include <boost/test/unit_test.hpp>
+
+#include "antares/study/study.h"
+
+using namespace Antares::Data;
+
+BOOST_AUTO_TEST_CASE(basic_areas_operations)
```
guilpier-code: It's a good thing to test these functions, but they're only used in the **GUI**, not in the solver or anywhere else.
Do we want to invest on tests about **GUI** code ?

src/tests/src/libs/antares/study/test_study.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
```
guilpier-code, line 54: The name of the  test is : **basic_areas_operations**.
What we do here is test **areaAdd** and **areaRename**.
Maybe we should split this test in 2 (because unit test should be short and have one goal only).
These 2 tests could be very simply called (for instance) : 
- **add_area_to_study**
- **rename_an_area**

If we have to repeat a set up for these tests, we should use a fixture.

src/tests/src/libs/antares/study/test_study.cpp
```diff
@@ -0,0 +1,95 @@
...
+    BOOST_CHECK(areaA->name == "A");
+    BOOST_CHECK(areaA->id == "a");
+    BOOST_CHECK(study.areaRename(areaA, "B"));
+    BOOST_CHECK(areaA->name == "B");
+    BOOST_CHECK(areaA->id == "b");
+    BOOST_CHECK(study.areaDelete(areaA));
+    BOOST_CHECK(study.areas.empty());
+}
+
+BOOST_AUTO_TEST_CASE(basic_thermal_clusters_operations)
```
guilpier-code: Again, what I've read about unit tests is : make them as short as possible with a good name. We're supposed to isolate one piece of functionality only. 
Here we could have 3 tests instead of one : 
- naming_a_thermal _cluster
- renaming_a_thermal_cluster
- removing_a_thermal_cluster

src/tests/src/libs/antares/study/test_study.cpp
```diff
@@ -0,0 +1,95 @@
...
+    BOOST_CHECK(study.clusterRename(newCluster.get(), "Renamed"));
+    BOOST_CHECK(newCluster->name() == "Renamed");
+    BOOST_CHECK(newCluster->id() == "renamed");
+
+    area.thermal.list.remove("renamed");
+    BOOST_CHECK(area.thermal.list.find("renamed") == nullptr);
+    BOOST_CHECK(area.thermal.list.empty());
+}
+
+BOOST_AUTO_TEST_CASE(basic_renewable_clusters_operations)
```
guilpier-code: Same for renewable clusters

src/tests/src/libs/antares/test_utils.cpp
```diff
@@ -18,6 +18,13 @@ T beautify(const T& in) {
 constexpr auto beautifyStd = beautify<std::string>;
 constexpr auto beautifyYuni = beautify<Yuni::String>;
 
+std::string transformNameToId(const AnyString& name)
```
guilpier-code, line 21: Is it a problem to test **Antares::TransformNameIntoID** directly without the help of **transformNameToId** ?

guilpier-code (changes requested):

src/tests/src/libs/antares/study/CMakeLists.txt
sylvlecl: I just followed the structure of the libs themselves: the study class is at the root dir of the "study" lib.
I thought it was the implicit rule, but actually I see that it's not the case for STS and thermal prices.

src/tests/src/libs/antares/study/test_study.cpp
sylvlecl: As long as they are present in the lib we need to guarantee they work unfortunately.
I don't know when we will actually completely drop the GUI, this could save us some more regressions :) 

Note also that we may want to provide the possibility to modify studies through other interface than the GUI, later (like python bindings).

src/tests/src/libs/antares/test_utils.cpp
sylvlecl: It's just less readable: this small function allows to have one-liners in the actual test code.

src/tests/src/libs/antares/study/CMakeLists.txt
guilpier-code: Yes, sorry, I wanted to remove this comment/question, but I forgot

src/tests/src/libs/antares/test_utils.cpp
guilpier-code: Then, as this new helper function's calls are more readable, why not change the production function to a more simple signature ?
After all, we have tests, so the risk of introducing a bug is low. And the change is very light, so this risk is even lower.

src/tests/src/libs/antares/test_utils.cpp
```diff
@@ -18,6 +18,13 @@ T beautify(const T& in) {
...
+{
+    std::string res;
+    Antares::TransformNameIntoID(name, res);
+    return res;
+}
+
 }
 
 BOOST_AUTO_TEST_CASE(test_beautify_name_std)
```
guilpier-code, line 30: A comment on tests already there (not introduced in this PR) : 
- on lines : 
```c++
BOOST_TEST(std::string("tÿst").size() == 5);
BOOST_TEST(Yuni::String ("tÿst").size() == 5);
```
One line would be enough, and more clear : 
```C++
 BOOST_TEST(Yuni::String ("tÿst").size() == std::string("tÿst").size());
```
- When I see the tests on **beautifyYuni** (and then when I look to its definition), I understand that it removes special character, and that **beautify** in its name is inaccurate. So we should rename it **removeSpecialChars**, shouldn't we ?
- **BeautifyName** has a strange signature (2 parameters: **in** and **out**), it should have one parameter (**in**) and return the changed string. That's the reason we currently use a helper function **beautify(...)** here : it's because **BeautifyName** should have a more simple signature. We use the helper function to test behavior of **BeautifyName**, which is odd.
  We should change the signature of the production function (**BeautifyName**) and remove the helper function. As we have tests on this function, there are less risk to introduce a bug. And the change is simple, so te risk is even lower.

src/tests/src/libs/antares/study/test_study.cpp
sylvlecl: I separated all 3 tests in 3 unit tests following your advice.
I arranged them in 3 test suites, one for areas, one for thermal clusters, one for renewables.

guilpier-code (changes requested):

src/tests/src/libs/antares/test_utils.cpp
sylvlecl: As discussed : I agree that the function should be changed in actual code, but it's out of the scope of this PR.
We should make a dedicated PR for this change, which will impact many places in the code base!

src/tests/src/libs/antares/test_utils.cpp
sylvlecl: About the 2 separate lines:
it's important to see that the size of those strings is 5, and not 4 as we could believe since there is indeed only 4 letters.

For comments about naming and signature, as discussed I agree, but this should be done in a separate PR!

guilpier-code:
I have created [this issue](https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/1700) for remaining requests.

guilpier-code (approved):

PR #1705, opened 2023-10-16
Refactor hydro series

payetvin:
#1669


src/solver/hydro/management/management.cpp
```diff
@@ -380,7 +380,7 @@ void HydroManagement::prepareNetDemand(uint numSpace, uint year)
 
         auto& rormatrix = area.hydro.series->ror;
         auto tsIndex = area.hydro.series->getIndex(year);
-        auto& ror = rormatrix[tsIndex < rormatrix.width ? tsIndex : 0];
+        const auto* ror = rormatrix.getColumn(tsIndex < rormatrix.timeSeries.width ? tsIndex : 0);
```
sylvlecl: I think the `getColumn` already handles the indirection : we don't need any more to first get the index and use it, we should use the year instead here.
```cpp
const auto* ror = rormatrix.getColumn(year);
```

PR #1707, opened 2023-10-16
Refactor hydro max power TS consistency

guilpier-code:
What was done (mainly) :
- [x] Factorizing the code that make time series collections of same size
- [x] Changing TS counts for : 
    - **ror**, **storage**, **mingen** ,
    - max power **gen** and **pump**
  
  from a **public** data member into a **private** data member. Indeed, one should not be able to change this TS count without resizing the associated TS accordingly.

Some other things done : 
- [x] Remove the exception raising, not necessary : the **gotFatalError** (set to true) will trigger an exception anyway.
- [x] class **NbTsComparer** : is renamed (into **pairOfTSsizes**), generalized to any numbers (not just TS numbers) and moved elsewhere 
- [x] Class **TsActions** removed (useless)


nikolaredstork (commented):
There are no numberComparison .h and .cpp files. Also, maybe I'm wrong, in **src/libs/antares/study/parts/parts.h** file, you probably have to include numberComparison.h file.

src/libs/antares/study/parts/hydro/pair-of-integers.h
```diff
@@ -0,0 +1,17 @@
...
+    bool same() const;
+    bool bothGreaterThanOne() const;
+    unsigned int sup() const { return numberOfTSsup_; }
+
+private:
+    unsigned int numberOfTS_1_;
+    unsigned int numberOfTS_2_;
+    unsigned int numberOfTSsup_{ 0 };
+    unsigned int numberOfTSinf_{ 0 };
+};
```
nikolaredstork, line 18: For the ror, hydro-storage and mingen there is three number comparison, so we should go with class that can receive variable list of integers, so that we can use it for both cases:

-  Maximum Generation/Pumping 2 integers comparison.
- ROR, Hydro-Storage and Minimum Generation 3 integers comparison.
image

src/libs/antares/study/area/list.cpp
```diff
@@ -1641,10 +1641,7 @@ void AreaList::removeLoadTimeseries()
 void AreaList::removeHydroTimeseries()
 {
     each([&](Data::Area& area) {
-        area.hydro.series->ror.reset(1, HOURS_PER_YEAR);
-        area.hydro.series->storage.reset(1, DAYS_PER_YEAR);
-        area.hydro.series->mingen.reset(1, HOURS_PER_YEAR);
-        area.hydro.series->count = 1;
+        area.hydro.series->reset();
```
nikolaredstork, line 1652: **void AreaList::removeHydroTimeseries()** function is used to reset TS's when TS generator is enabled. 
image

Maximum Generation/Pumping can't be generated via TS generator, so do we have to reset them when TS generator is enabled for hydro?

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
```diff
@@ -122,59 +122,15 @@ void HydroMaxTimeSeriesReader::copyDailyMaxPumpingEnergy(Area& area) const
...
-}
-
 bool HydroMaxTimeSeriesReader::operator()(const AnyString& folder, Area& area, bool usedBySolver)
 {
     bool ret = true;
 
     ret = loadDailyMaxPowersAndEnergies(folder, area, usedBySolver) && ret;
     copyDailyMaxEnergy(area);
-    copyDailyMaxPowerAsHourly(area);
+    area.hydro.series->buildMaxPowerFromDailyTS(dailyMaxPumpAndGen[genMaxP], dailyMaxPumpAndGen[pumpMaxP]);
```
nikolaredstork, line 131: The whole idea was to wrap around necessary code which will provide support for old studies, and also if in future we don't need that support anymore, we could just erase `HydroMaxTimeSeriesReader` class and it's usage in **list.cpp** file. With current situation `DataSeriesHydro` class have functions which are going to be used just if heder version variable is less than 870 and with upgrading a study these functions are useless  for `DataSeriesHydro` class, so I don't think that functions:

- `buildMaxPowerFromDailyTS`
- `ConvertDailyTSintoHourlyTS`

should be part of the DataSeriesHydro class.

src/libs/antares/study/parts/hydro/series.h
```diff
@@ -144,108 +151,34 @@ class DataSeriesHydro
...
-    /**
-     * This class provides comparing two integers
-     * that represents width of two corresponding matrices.
-     */
-    class NbTsComparer
-    {
-    public:
-        NbTsComparer(uint32_t nbOfGenPowerTs, uint32_t nbOfPumpPowerTs);
-
-        bool bothZeros() const;
```
nikolaredstork, line 220: Function `bothZeros `is not useless as it seems. When we are loading study via GUI, matrices ror, storage, mingen will not be loaded and their width are 0. In current version we have excluded bothZeros function, so every time on loading study via GUI, code will enter `EqualizeTSsize` function, and because both width are 0, if condition `pairOfTSsizes.same() = true` and function `EqualizeTSsize` will return 0.

image


. Maybe we don't need bothZero function, we just have to use usedBySolver bool variable in `EqualizeGenerationTSsizes` function like this:

image

Current version will not break simulation, whether it is loaded via GUI or solver, but it is not necessary to check matrices if we know that their width are 0 when we loading study via GUI.

**NOTE:** If you are wondering what will happen to the MaxPower TS (maxHourlyGenPower and maxHourlyPumpPower), for them we are checking whether or not study is loaded via GUI or via solver in **list.cpp** file: 

image

src/libs/antares/study/area/list.cpp
```diff
@@ -904,7 +904,11 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
             buffer.clear() << study.folderInput << SEP << "hydro" << SEP << "series";
-            ret = area.hydro.series->loadFromFolder(study, area.id, buffer) && ret;
+            ret = area.hydro.series->loadROR(area.id, buffer) && ret;
+            ret = area.hydro.series->loadINFLOWS(area.id, buffer) && ret;
+            ret = area.hydro.series->loadMINGEN(area.id, buffer, study.header.version) && ret;
+
+            area.hydro.series->EqualizeGenerationTSsizes(area, study.usedByTheSolver, study.gotFatalError);
         }
 
         if (area.hydro.series && study.header.version < 870)
```
nikolaredstork: Maybe we should extract loading of deprecated files from `AreaListLoadFromFolderSingleArea` function and implement specific load function for deprecated files `AreaLoadFromFolderDeprecatedFiles`, in order to avoid confusion for loading hydro files:

Implementation:
image

Function call:
image

src/libs/antares/study/area/list.cpp
```diff
@@ -922,14 +926,14 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
-                ret = area.hydro.series->postProcessMaxPowerTS(area) && ret;
-                area.hydro.series->setMaxPowerTSWhenDeratedMode(study);
+                area.hydro.series->EqualizeMaxPowerTSsizes(area, study.gotFatalError);
             }
             else
-                area.hydro.series->setHydroModulability(study, area.id);
+                area.hydro.series->setHydroModulability(area);
         }
 
+        area.hydro.series->resizeTSinDeratedMode(study.parameters.derated, study.header.version);
```
nikolaredstork: If we are loading via GUI, matrices ror, storage, mingen, maxHourlyGenPower, maxHourlyPumpPower are not actually loaded and in case of derated mod, loading via GUI code will calculating average column on empty matrix. Resize in derated mod should be called if study is used by the solver.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -224,173 +257,78 @@ void DataSeriesHydro::reset()
...
-    timeseriesNumbersHydroMaxPower.clear();
-
-    return ret;
+    ror.resize(width, HOURS_PER_YEAR);
+    storage.resize(width, DAYS_PER_YEAR);
+    generationTScount_ = width;
 }
 
-bool DataSeriesHydro::postProcessMaxPowerTS(Area& area)
+void DataSeriesHydro::resizeGenerationTS(unsigned int w, unsigned int h)
```
nikolaredstork, line 278: You are you using `resizeGenerationTS` and `resizeRORandSTORAGE` for calculating TS in derated mod? In derated mod, we are calculating average matrices for ror, storage, mingen, maxHourlyGenPower, maxHourlyPumpPower with data provided from .txt files. In this case if derated mod is on matrices will have width 1 filled with junk data (reset function will set all matrix value to 0, resize will only set width and height).

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -77,128 +133,105 @@ bool DataSeriesHydro::saveToFolder(const AreaName& areaID, const AnyString& fold
...
-            ror.averageTimeseries();
-            storage.averageTimeseries();
-            mingen.averageTimeseries();
-            count = 1;
-        }
-    }
+    filepath.clear() << folder << SEP << areaID << SEP << "maxHourlyGenPower.txt";
+    ret = maxHourlyGenPower.loadFromCSVFile(filepath, 1, HOURS_PER_YEAR, &fileContent) && ret;
 
-    timeseriesNumbers.clear();
```
nikolaredstork, line 167: We excluded clearing of timeseriesNumbers matrix?

src/libs/antares/study/parts/hydro/pair-of-integers.h
guilpier-code: Not sure it would be useful, and here is why (in my opinion) :
- Let's take an exemple from another field : sort algorithm. From a comparison between 2 elements, sort algorithm manages to sort the elements of a list.
Here, the idea would be the same : use the comparison between 2 elements to compare many of them.
So we should keep this class as it is, and build on it.
- As you said, we have the **ror**, **storage** and **mingen** time series, and we have to "equalize" their sizes. It's currently done in function **DataSeriesHydro::EqualizeGenerationTSsizes**. As you can see in this function's body, there is some code duplication : we have 2 blocks very much alike. This is usually a sign that we should refactor to remove the code duplication. But I spent some time to get to this results on reading the hydro TS, and there's still things to do, apart from removing this code duplication which is, in fact the subject you're talking about here.
To me, we should first try to merge this as it is, create a gitHub issue if we're not happy, and fix this later.

src/libs/antares/study/area/list.cpp
guilpier-code: As you just said, **AreaList::removeHydroTimeseries()** is only by the historic **GUI**. 
This GUI will disappear (the sooner the better), and will be fully replaced with the web version.
So this function will be removed as well.
But you're right, some extra things are done that we don't need. 
**Unless we introduce a bug in the GUI**, the work required to have it clean is not worth it.

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
guilpier-code: Here was my idea on that matter : 
- I saw what you were trying to do : **isolate** and **externalize** the reading of the daily TS to turn them into some hourly TS. The idea of isolating looks really good to me. But reading these daily TS is still reading hydro TS, and so it should be done inside a reading function or class, for logical and readability reasons.
- **Externalize** : what you did was **externalizing** the daily TS reading, that is : reading these TS outside the TS reader itself (**HydroMaxTimeSeriesReader** is totally external  to **DataSeriesHydro**). This is what is called a **feature envy**. A consequence of this is that you're forced to supply your reader (**DataSeriesHydro**) with a **public** setter **setNbTimeSeriesSup(...)**, and that the real TS  sizes (inside the reader) are not synchronized automatically with the **number of TS** (**nbTimeSeriesSup_**) : you can set this **number of TS** without modifying the size of your TSs. And this is a problem to me.
- **isolate** : as I said, isolating **IS** very much suitable. But it has to be done inside the reader : the reader delegates the reading of daily TS to an object of which reading daily TS is its specialty.

The design of the reader is still not currently ideal (we hope it will be improved soon).
Following what I just said, here is a rough idea of what it could be (at least it would be much better than today) : 
```C++
// list.cpp
...
HydroTSfetcher hydroTSfetcher(studyVersion, 
                             pathToInputDir, 
                             area, 
                             parameters.timeSeriesToRefresh, 
                             /* other args ? */);
ret = hydroTSfetcher.go() && ret;
``` 
As you can, there 's only one object / interface to load all TS collections (for a given area). 
Inside the fetcher, we would build and run sub-objects that would **read** or **generate** time series (depending on the value of `parameters.timeSeriesToRefresh`), and that would do it differently depending of the study version. All internal classes would be a lot smaller than the current **DataSeriesHydro**.

But all this is for another time.

src/libs/antares/study/parts/hydro/series.h
guilpier-code: I removed the **bothZero** case because : 
Before calling **EqualizeTSsize** function, we always call a load function (**loadROR**, **loadMINGEN**, ...).
Each of them contain a call to **Matrix::loadFromCSVFile(...)**. And this function always gives to the matrix a width at least equal to one. So, in **EqualizeTSsize(...)**, even the case where one of the 2 matrices has a width = 0 is impossible.
Did I miss something ? Did you have a crash or a bug with GUI while loading hydro TS ?

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
nikolaredstork: Thanks for the explanation.

src/libs/antares/study/area/list.cpp
guilpier-code: The above comment [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1707#discussion_r1371459705) should be an answer.

nikolaredstork: Thanks!

src/libs/antares/study/parts/hydro/series.h
nikolaredstork: You are absolutely right what you have said, but that is the case when a study is loaded via solver. In case when we are opening (loading) a study via GUI, matrices will not be loaded and we are performing checks for empty matrices. How I know that they are empty when opening study via GUI? I've debug it. This is a snapshot of debug mode, where you can see that ror and storage matrices are empty after loading via GUI. 

image

I know that we are deprecating this GUI, but should we leave it this way?

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: Oops, it seems I made a mistake. Thanks for seeing that.
Correction done.

src/libs/antares/study/area/list.cpp
guilpier-code: Easy to correct. Thanks to have such good eyes.
Correction done.

src/libs/antares/study/parts/hydro/series.h
guilpier-code: No, you're right, we don't want any bug / crash in GUI, even if it will be removed some day.
Thanks for having tested this.
Correction done.

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: you're right. Done.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -316,15 +316,24 @@ uint DataSeriesHydro::maxPowerTScount() const
...
-void DataSeriesHydro::resizeTSinDeratedMode(bool derated, unsigned int studyVersion)
+void DataSeriesHydro::resizeTSinDeratedMode(bool derated, 
+                                            unsigned int studyVersion,
+                                            bool usedBySolver)
 {
     if (!derated)
         return;
 
-    resizeGenerationTS(1, HOURS_PER_YEAR);
+    ror.averageTimeseries();
```
nikolaredstork, line 330: I think that ror, storage and mingen should also be averaged if usedBySolver is true and also I think that mingen should be averaged if study header version is greater and equal to 860. Studies below 860 version does not have mingen.txt file, but in constructor of the DataSeriesHydro mingen matrix is set to width 1 with values of 0, meaning that averaging will be performed on the 1x8760 matrix with values of 0.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -62,6 +62,13 @@ unsigned int EqualizeTSsize(Matrix<double, int32_t>& TScollection1,
 {
     PairOfIntegers pairOfTSsizes(TScollection1.width, TScollection2.width);
 
+    if (pairOfTSsizes.bothZero())
+    {
+        TScollection1.reset(1, HOURS_PER_YEAR);
+        TScollection2.reset(1, HOURS_PER_YEAR);
+        return 1;
```
nikolaredstork, line 71: Storage matrix has 365 height (DAYS_PER_YEAR), so we need somehow to parse heights of the matrices.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -153,21 +153,15 @@ bool loadTSfromFile(Matrix<double, int32_t>& ts,
...
-{
-    return loadTSfromFile(storage, areaID, folder, "mod.txt", DAYS_PER_YEAR);
-}
+    timeseriesNumbers.clear();
 
-bool DataSeriesHydro::loadMINGEN(AreaName& areaID, const AnyString& folder, unsigned int studyVersion)
-{
+    bool ret = loadTSfromFile(ror, areaID, folder, "ror.txt", HOURS_PER_YEAR);
+    ret = loadTSfromFile(storage, areaID, folder, "mod.txt", DAYS_PER_YEAR) && ret;
     if (studyVersion < 860)
```
nikolaredstork: `if(studyVersion >= 860)`
We obviously don't have end-to-end test for minimum generation feature, otherwise that end-to-end test would fail.

src/libs/antares/study/parts/hydro/series.h
nikolaredstork: Just to add, this will not crash GUI, but nevertheless it should be considered.

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: ok, done.

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: ok, done.

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: ok, done.

PR #1711, opened 2023-10-19
Python version consistency in CI

sylvlecl:
**Description**

Currently, on windows CI, we get some inconsistencies between python version defined in the `setup-python` action, and the python version found by cmake.

By default, cmake searches for python version based on registry, which is not guaranteed to be the same as selected in `setup-python` action.

To ensure that cmake uses the same python version as selected by `setup-python` in the CI, we specify now the `Python3_EXECUTABLE` hint variable, using the output of the setup-python step.

_This will ensure that we completely control the python version used in the CI._


a-zakir:
I'm ok with your proposition, just for test, did you try in setup-python an older version of python (e.g <3.12) ?

sylvlecl:
> I'm ok with your proposition, just for test, did you try in setup-python an older version of python (e.g <3.12) ?

Indeed, I checked with 3.11, see [this job](https://github.com/AntaresSimulatorTeam/Antares_Simulator/actions/runs/6572675941/job/17854222130).

Cmake correctly picks the 3.11 version defined by setup-python:
```
 -- Found Python3: C:\hostedtoolcache\windows\Python\3.11.6\x64\python.exe (found version "3.11.6") found components: Interpreter 
```

(I cancelled the workflow after that)

a-zakir (approved):

PR #1713, opened 2023-10-20
Unit tests for timeseries

payetvin:
https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1677

### DIFF
--- a/src/libs/antares/study/parts/series/series.cpp
+++ b/src/libs/antares/study/parts/series/series.cpp
@@ -90,11 +90,11 @@ uint32_t TimeSeries::getSeriesIndex(uint32_t year) const
         return timeseriesNumbers[0][year];
 }
 
-double* TimeSeries::operator[](uint32_t year)
+double* TimeSeries::operator[](uint32_t index)
 {
-    if (timeSeries.width <= year)
+    if (timeSeries.width <= index)
         return nullptr;
-    return timeSeries[year];
+    return timeSeries[index];
 }
 
 void TimeSeries::reset()
--- a/src/libs/antares/study/parts/series/series.cpp
+++ b/src/libs/antares/study/parts/series/series.cpp
@@ -92,7 +92,7 @@ uint32_t TimeSeries::getSeriesIndex(uint32_t year) const
 
 double* TimeSeries::operator[](uint32_t year)
 {
-    if (timeSeries.width < year)
+    if (timeSeries.width <= year)
         return nullptr;
     return timeSeries[year];
 }
--- a/src/libs/antares/study/parts/series/series.cpp
+++ b/src/libs/antares/study/parts/series/series.cpp
@@ -92,6 +92,8 @@ uint32_t TimeSeries::getSeriesIndex(uint32_t year) const
 
 double* TimeSeries::operator[](uint32_t year)
 {
+    if (timeSeries.width < year)
+        return nullptr;
     return timeSeries[year];
 }
 
--- a/src/tests/src/libs/antares/study/series/CMakeLists.txt
+++ b/src/tests/src/libs/antares/study/series/CMakeLists.txt
@@ -11,11 +11,6 @@ target_link_libraries(timeseries-tests
                       Boost::unit_test_framework
                       Antares::series
 )
-# Linux
-if(UNIX AND NOT APPLE)
-target_link_libraries(timeseries-tests PRIVATE stdc++fs)
-endif()
-
 
 # Storing timeseries-tests under the folder Unit-tests in the IDE
 set_target_properties(timeseries-tests PROPERTIES FOLDER Unit-tests/timeseries-tests)
--- a/src/tests/src/libs/antares/study/series/CMakeLists.txt
+++ b/src/tests/src/libs/antares/study/series/CMakeLists.txt
@@ -1,5 +1,5 @@
 # ====================================
-# Tests on thermal price definition
+# Tests on TimeSeries class
 # ====================================
 set(SRC_TIMESERIES_TESTS
 	timeseries-tests.cpp
--- a/src/tests/src/libs/antares/study/series/CMakeLists.txt
+++ b/src/tests/src/libs/antares/study/series/CMakeLists.txt
@@ -1,0 +1,25 @@
+# ====================================
+# Tests on thermal price definition
+# ====================================
+set(SRC_TIMESERIES_TESTS
+	timeseries-tests.cpp
+)
+add_executable(timeseries-tests ${SRC_TIMESERIES_TESTS})
+
+target_link_libraries(timeseries-tests
+                      PRIVATE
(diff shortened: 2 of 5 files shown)


src/libs/antares/study/parts/series/series.cpp
```diff
@@ -92,6 +92,8 @@ uint32_t TimeSeries::getSeriesIndex(uint32_t year) const
 
 double* TimeSeries::operator[](uint32_t year)
 {
+    if (timeSeries.width < year)
```
a-zakir:
```suggestion
    if (timeSeries.width <= year)
```

src/tests/src/libs/antares/study/series/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# ====================================
+# Tests on thermal price definition
```
sylvlecl: to be removed

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -0,0 +1,138 @@
...
+BOOST_AUTO_TEST_SUITE(timeseries_tests)
+
+BOOST_FIXTURE_TEST_CASE(getSeriesIndex, Fixture)
+{
+    tsnum.resize(1, 10);
+    for (unsigned int i = 0; i < 10; i++)
+        tsnum[0][i] = i;
+
+    //timeSeries.width == 1 so returns 0
+    BOOST_CHECK_EQUAL(ts.getSeriesIndex(5), 0);
```
sylvlecl, line 69: I'm surprised by this behaviour! Is it not a situation where we should throw ?

We have an index for series, but because we only have one timeseries, we don't take that index into account. it looks like an inconsistency in input data.

Let's not change it in this work, but I note it for later.

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    ts.resize(2, HOURS_PER_YEAR);
+    for (unsigned int i = 0; i < 10; i++)
+        BOOST_CHECK_EQUAL(ts.getSeriesIndex(i), i);
```
sylvlecl, line 73: I think this behaviour is even more problematic:
our timeseries has only size 2, but we get an index greater than 2.

For me we should throw an exception in that case, it's too fragile.

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+    fillColumn(0);
+    BOOST_CHECK_EQUAL(ts.getCoefficient(0, 12), 12);
+    BOOST_CHECK_EQUAL(ts.getCoefficient(0, 8750), 8750);
+}
+
+BOOST_FIXTURE_TEST_CASE(getCoefficientWidth0, Fixture)
+{
+    ts.resize(0, HOURS_PER_YEAR);
+    BOOST_CHECK_EQUAL(ts.getCoefficient(0, 12), 0);
```
sylvlecl, line 86: For me this should throw too. However we need to make sure it's not a regression, I'll look into it too.

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+    for (unsigned int i = 0; i < ts.timeSeries.height; i++)
+        ts.timeSeries[idx][i] = HOURS_PER_YEAR - i;
+}
+
+void Fixture::fillTsnum()
+{
+    tsnum.resize(1, ts.timeSeries.width);
+    for (unsigned int i = 0; i < ts.timeSeries.width; i++)
+        tsnum[0][i] = i;
```
sylvlecl, line 53: This is bad test data:
the index and what it points to are the same, so we don't really test that the indirection works well.
Some explicit hard coded values would be better like `{3,0,2,1}` or whatever.

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -0,0 +1,138 @@
...
+    fillTsnum();
+    BOOST_CHECK_EQUAL(ts.getCoefficient(3, 12), 0);
+    BOOST_CHECK_EQUAL(ts.getCoefficient(3, 8750), 0);
+}
+
+BOOST_FIXTURE_TEST_CASE(getCoefficientWidthMoreThan1, Fixture)
+{
+    ts.resize(5, HOURS_PER_YEAR);
+    fillTsnum();
+    fillColumn(3);
```
sylvlecl: This is bad test data too:
we need to have different values in the different columns, otherwise we don't test that we indeed fetch data from the right column.

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    BOOST_CHECK_EQUAL(ts.getCoefficient(3, 12), 12);
```
sylvlecl, line 104: See here, we don't know if the 12 value comes from the column for the third year.

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -0,0 +1,138 @@
...
+    auto col = ts.getColumn(3); //emptyColumn
+    BOOST_CHECK_EQUAL(col[38], 0);
+    BOOST_CHECK_EQUAL(col[7463], 0);
+
+    ts.resize(4, HOURS_PER_YEAR);
+    fillTsnum();
+    fillColumn(2);
+
+    col = ts.getColumn(2);
+    BOOST_CHECK_EQUAL(col[38], 38);
```
sylvlecl, line 124: Same here, because of the test data identical in all columns, we are not sure that we have the right column.

sylvlecl (changes requested):
We need to improve the test data: they don't allow to validate the behaviour as is!

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
sylvlecl: I misunderstood the code: indeed the data is different in other columns (zeros).

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
sylvlecl: This analysis was wrong too.

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
sylvlecl: Wrong analysis too

sylvlecl (approved):

PR #1716, opened 2023-10-25
Add check in unfeasible analyzer: unit tests

sylvlecl:
**Description**
Adding some unit tests on top of #1689 and of existing code.

### DIFF
--- a/.github/workflows/windows-vcpkg.yml
+++ b/.github/workflows/windows-vcpkg.yml
@@ -123,7 +123,7 @@ jobs:
     - name: Run unfeasibility-related tests
       run: |
            cd _build
-           ctest -C Release --output-on-failure -V -R unfeasible
+           ctest -C Release --output-on-failure -R "^unfeasible$"
 
     - name: Run unit and end-to-end tests
       run: |
--- a/.github/workflows/windows-vcpkg.yml
+++ b/.github/workflows/windows-vcpkg.yml
@@ -123,7 +123,7 @@ jobs:
     - name: Run unfeasibility-related tests
       run: |
            cd _build
-           ctest -C Release --output-on-failure -R unfeasible
+           ctest -C Release --output-on-failure -V -R unfeasible
 
     - name: Run unit and end-to-end tests
       run: |
--- a/src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
+++ b/src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
@@ -64,6 +64,8 @@ public:
             shouldDetectCause_(shouldDetectCause)
     {}
 
+    ~AnalysisMock() override = default;
+
     void run(operations_research::MPSolver *problem) override
     {
         hasRun_ = true;
@@ -125,12 +127,12 @@ BOOST_AUTO_TEST_CASE(test_problem_analyzer)
 
 BOOST_AUTO_TEST_CASE(test_variable_bounds_consistency)
 {
-    VariablesBoundsConsistency analysis;
     std::unique_ptr<MPSolver> problem(MPSolver::CreateSolver("GLOP"));
     problem->MakeNumVar(-1, 1, "ok-var");
     problem->MakeNumVar(1, -1, "not-ok-var");
-    analysis.run(problem.get());
 
+    VariablesBoundsConsistency analysis;
+    analysis.run(problem.get());
     auto incorrectVars = analysis.incorrectVars();
     BOOST_CHECK_EQUAL(incorrectVars.size(), 1);
 
@@ -174,26 +176,32 @@ std::string validConstraintNames[] =
 
 BOOST_DATA_TEST_CASE(test_slack_variables_with_unfeasible_constraint, bdata::make(validConstraintNames), constraintName)
 {
-    ConstraintSlackAnalysis analysis;
     std::unique_ptr<MPSolver> unfeasibleProblem = createProblem(constraintName, false);
-
     BOOST_CHECK(unfeasibleProblem->Solve() == MPSolver::INFEASIBLE);
 
+    ConstraintSlackAnalysis analysis;
     analysis.run(unfeasibleProblem.get());
     BOOST_CHECK(analysis.hasDetectedInfeasibilityCause());
-    analysis.printReport();
+}
+
+BOOST_AUTO_TEST_CASE(test_slack_variables_with_unfeasible_constraint_name_ignored)
+{
+    std::unique_ptr<MPSolver> unfeasibleProblem = createProblem("ignored-name", false);
+    BOOST_CHECK(unfeasibleProblem->Solve() == MPSolver::INFEASIBLE);
+
+    ConstraintSlackAnalysis analysis;
+    analysis.run(unfeasibleProblem.get());
+    BOOST_CHECK(!analysis.hasDetectedInfeasibilityCause());
 }
 
 BOOST_DATA_TEST_CASE(test_slack_variables_no_problematic_constraint, bdata::make(validConstraintNames), constraintName)
 {
-    ConstraintSlackAnalysis analysis;
     std::unique_ptr<MPSolver> feasibleProblem = createProblem(constraintName, true);
     BOOST_CHECK(feasibleProblem->Solve() == MPSolver::OPTIMAL);
 
+    ConstraintSlackAnalysis analysis;
     analysis.run(feasibleProblem.get());
     BOOST_CHECK(analysis.hasDetectedInfeasibilityCause());  // Would expect false here instead?
-
-    analysis.printReport();
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
+++ b/src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
@@ -165,7 +165,7 @@ std::unique_ptr<MPSolver> createProblem(const std::string& constraintName, bool 
     return problem;
 }
 
-std::string validConstraintNames[] =
+static const std::string validConstraintNames[] =
 {
     "BC::hourly::hour<36>",
     "BC::daily::day<67>",
--- a/src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
+++ b/src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
@@ -55,13 +55,15 @@ bool variableEquals(const VariableBounds& lhs, const VariableBounds& rhs)
 }
 
 /*!
- * Analysis mock
+ * Analysis mock, used to assess which step has been run by the analyzer
  */
 class AnalysisMock : public UnfeasibilityAnalysis
 {
 public:
-    explicit AnalysisMock(bool shouldDetectCause):
-            shouldDetectCause_(shouldDetectCause)
+    AnalysisMock(bool shouldDetectCause, bool& hasRun, bool& hasPrinted):
+            shouldDetectCause_(shouldDetectCause),
+            hasRun_(hasRun),
+            hasPrinted_(hasPrinted)
     {}
 
     ~AnalysisMock() override = default;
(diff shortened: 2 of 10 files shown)


.github/workflows/windows-vcpkg.yml
```diff
@@ -123,7 +123,7 @@ jobs:
     - name: Run unfeasibility-related tests
       run: |
            cd _build
-           ctest -C Release --output-on-failure -R unfeasible
+           ctest -C Release --output-on-failure -V -R unfeasible
```
guilpier-code: We'll have to roll it back to what it was

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+    return lhs.name == rhs.name &&
+        lhs.lowBound == rhs.lowBound &&
+        lhs.upBound == rhs.upBound;
+}
+
+/*!
+ * Analysis mock, used to assess which step has been run by the analyzer
+ */
+class AnalysisMock : public UnfeasibilityAnalysis
```
guilpier-code, line 60: with another C++ unit test library (see **GoogleTest**, **catch 2**, **CppUTest**, ...), we wouldn't need this mock class, and we would need variables like **hasRun1**, **hasPrinted1**, .... 
We should consider adopting one of these unit tests framework

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+    bool hasRun1 = false;
+    bool hasPrinted1 = false;
+    bool hasRun2 = false;
+    bool hasPrinted2 = false;
+    std::vector<std::unique_ptr<UnfeasibilityAnalysis>> analysis;
+    analysis.push_back(std::make_unique<AnalysisMock>(false, hasRun1, hasPrinted1));
+    analysis.push_back(std::make_unique<AnalysisMock>(true, hasRun2, hasPrinted2));
+    std::unique_ptr<MPSolver> problem(MPSolver::CreateSolver("GLOP"));
+
+    UnfeasiblePbAnalyzer analyzer(std::move(analysis));
```
guilpier-code, line 107: Do we need the use of **std::move** here, as it's used as well in the constructor initialization list ?

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+private:
+    bool shouldDetectCause_;
+    bool& hasRun_;
+    bool& hasPrinted_;
+};
+
+
+BOOST_AUTO_TEST_SUITE(unfeasible_problem_analyzer)
+
+BOOST_AUTO_TEST_CASE(test_problem_analyzer)
```
guilpier-code: The test name is starts with **test_** : we know the context is unit test, so this prefix is useless, don't you think ?
Here we could call the test : **probe every step of unfeasible analysis** ? (may be there's a better name)
Please see next comment.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+private:
+    bool shouldDetectCause_;
+    bool& hasRun_;
+    bool& hasPrinted_;
+};
+
+
+BOOST_AUTO_TEST_SUITE(unfeasible_problem_analyzer)
```
guilpier-code, line 94: I'm wondering (it's your choice to take this into account) : 
We would expect to test an object's **behavior** (not methods), using it's interface.
We would test simple behaviors, and then more tricky ones.
For example, the names of the tests could be (in the first tests, no need to print a report) : 
- run_analyzer_with_no_analysis (what should we expect in this case ?)
- run_analyzer_with_one_analysis_detecting_a_pb
- run_analyzer_with_one_analysis_unable_to_detect_pb
- run_analyzer_with_2_analyses__none_detects_a_pb
- run_analyzer_with_2_analyses__only_one_detects_pb
- run_analyzer_with_2_analyses_detecting_a_pb__only the first_one_is_run
- what else ?

About printing reports : 
@sylvlecl : we just had a chat about it, run() method should produce a report (for example a vector of **std::string**), either return by **run()**, or retrievable with a **get** method. 
This would allow to remove the dependency to the logger.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+/*!
+ * Creates a problem with 2 variables linked by 1 constraint:
+ *  - Variable 1 must be greater than 1
+ *  - Variable 2 must be lesser than -1
+ *  - but if feasible is false, constraint enforces that variable 2 is greater than variable 1 --> infeasible
+ */
+std::unique_ptr<MPSolver> createProblem(const std::string& constraintName, bool feasible)
+{
+    ConstraintSlackAnalysis analysis;
```
guilpier-code: `ConstraintSlackAnalysis analysis;` : analysis is unused in this function.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+    std::unique_ptr<MPSolver> problem(MPSolver::CreateSolver("GLOP"));
+    const double inf = problem->infinity();
```
guilpier-code: We should not hesitate naming infinity "infinity" instead of "inf", which can be confused with **inferior**

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+    BOOST_CHECK(variableEquals(incorrectVars[0], expected));
+}
+
+/*!
+ * Creates a problem with 2 variables linked by 1 constraint:
+ *  - Variable 1 must be greater than 1
+ *  - Variable 2 must be lesser than -1
+ *  - but if feasible is false, constraint enforces that variable 2 is greater than variable 1 --> infeasible
+ */
+std::unique_ptr<MPSolver> createProblem(const std::string& constraintName, bool feasible)
```
guilpier-code, line 148: Calling **createProblem()** like this `createProblem(constraintName, false);` is a bit unclear.
A solution could be to have 3 functions : 
- **std::unique_ptr\<MPSolver\> createSimpleProblem()**
- **void createFeasibleProblem(MPSolver\* pb)**
- **void createUnfeasibleProblem(MPSolver\* pb)**

The last 2 functions would call the first for for the creation a problem with one constraint of 2 variables.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+static const std::string validConstraintNames[] =
+{
+    "BC::hourly::hour<36>",
+    "BC::daily::day<67>",
+    "BC::weekly::week<12>",
+    "FictiveLoads::hour<25>",
+    "AreaHydroLevel::hour<8>",
+};
+
+BOOST_DATA_TEST_CASE(test_slack_variables_with_unfeasible_constraint, bdata::make(validConstraintNames), constraintName)
```
guilpier-code: **Test_** prefix not needed.
This test name does not really reflects what it does.
What it does is : 
slack variable analysis detects unfeasibility on many well names constraints.
A bit long, but much clearer.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+    std::unique_ptr<MPSolver> unfeasibleProblem = createProblem(constraintName, false);
+    BOOST_CHECK(unfeasibleProblem->Solve() == MPSolver::INFEASIBLE);
+
+    ConstraintSlackAnalysis analysis;
+    analysis.run(unfeasibleProblem.get());
+    BOOST_CHECK(analysis.hasDetectedInfeasibilityCause());
+}
+
+BOOST_AUTO_TEST_CASE(test_slack_variables_with_unfeasible_constraint_name_ignored)
```
guilpier-code: A name could be : 
slack variable analysis does not detect unfeasibilty on ill-named constraint

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+    std::unique_ptr<MPSolver> unfeasibleProblem = createProblem("ignored-name", false);
+    BOOST_CHECK(unfeasibleProblem->Solve() == MPSolver::INFEASIBLE);
+
+    ConstraintSlackAnalysis analysis;
+    analysis.run(unfeasibleProblem.get());
+    BOOST_CHECK(!analysis.hasDetectedInfeasibilityCause());
+}
+
+BOOST_DATA_TEST_CASE(test_slack_variables_no_problematic_constraint, bdata::make(validConstraintNames), constraintName)
```
guilpier-code: Do we need to parameterize a test for an analysis that won't detect anything ?

guilpier-code (changes requested):

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
sylvlecl: Yep, necessary for the compiler, because the vector is move-only (because its content is move-only).
without `std::move` I guess the compiler tries to perform a copy.

.github/workflows/windows-vcpkg.yml
sylvlecl: Done together with narrowing the regex to ^unfeasible$

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
sylvlecl: Done

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
sylvlecl: Done

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
sylvlecl: I left the test as is to test the actual workflow. I agree that for more important piece of codes we should bother to have finer grained tests though.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
sylvlecl: Overall, I changed the namings according to your input, thank you!
In particular, I indeed removed the test_ prefix, and tried to give a small explanation of the expected behavior ("should" do this).

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
sylvlecl: I provided the 2 methods `createFeasible` and `createUnfeasible` which just call the initial one.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
sylvlecl: Not very useful indeed, a little redundant with the other test case, I removed the parameterization.

Issue #1427, opened 2023-07-13
Solver traces from OR-Tools


flomnes:
According to @klorel, it is very useful for debugging.

PR #1717, opened 2023-10-25
print solver logs

### LINKED ISSUE #1709: Access to optimization solver logs
**Description**

It seems that there is no way in the simulator, today, to get the logs of the underlying solver (xpress through ortools, for example).

This would be highly useful to get information about the execution and performances of the optimization problem resolution.

Mentioned by @tbittar. 

Note that antares-xpansion already provides this possibility: if needed, we can get some implementation guidance from there, although it might not be relevant when using it through ortools.

a-zakir:
In this Pr, i'm trying to enable solvers logs:
##TODO
- [x] with ortools interface
- [x] Sirius direct call
added : 
 - CLI option ` --solver-logs`
 - `solver-log` boolean option to be read in `generaldata.ini`
will close #1709 maybe #1427?

### DIFF
--- a/docs/reference-guide/10-command_line.md
+++ b/docs/reference-guide/10-command_line.md
@@ -40,7 +40,7 @@ _In all cases, arguments " –h" or "–help" can be used to get help_
 |--no-ts-import | Do not import timeseries into the input folder. <br/> (This option may be useful for running old studies without upgrade)|
 |-m, --mps-export | Export anonymous mps weekly or daily optimal UC+dispatch linear |
 |-s, --named-mps-problems | Export named mps weekly or daily optimal UC+dispatch linear |
-|-o, --solvers-logs | Print solver logs |
+|--solver-logs | Print solver logs |
 
 - Misc.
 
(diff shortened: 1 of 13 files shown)


payetvin (approved):

docs/reference-guide/10-command_line.md
```diff
@@ -40,6 +40,7 @@ _In all cases, arguments " –h" or "–help" can be used to get help_
 |--no-ts-import | Do not import timeseries into the input folder. <br/> (This option may be useful for running old studies without upgrade)|
 |-m, --mps-export | Export anonymous mps weekly or daily optimal UC+dispatch linear |
 |-s, --named-mps-problems | Export named mps weekly or daily optimal UC+dispatch linear |
+|-o, --solvers-logs | Print solver logs |
```
sylvlecl: Would prefer "solver-logs" without the plural.

Also, I would remove the short option, I think it's OK to only have the long one.
We can keep the "-o" for a future option that would be more compatible with it.

sylvlecl (changes requested):

docs/reference-guide/10-command_line.md
a-zakir: good

a-zakir: done!

src/libs/antares/study/parameters.cpp
```diff
@@ -626,6 +627,11 @@ static bool SGDIntLoadFamily_Optimization(Parameters& d,
     {
         return stringToGlobalTransmissionCapacities(value, d.transmissionCapacities);
     }
+
+    if (key == "solver-logs")
```
sylvlecl, line 631: We'll need to add it in the documentation too (new section v8.9 in the file format doc).

a-zakir: was thinking about it,  and said to myself that @flomnes  would do it 😛

sylvlecl (approved):

PR #1723, opened 2023-10-28
Multiple Maximum Generation and Pumping for Hydro (Merging with develop branch)

nikolaredstork:
This PR is created due to major changes on the develop branch, that also caused significant changes on the CR23.

Notable changes:

1. Type of the matrices that represents Maximum Generation and Pumping data is changed into newly implemented type `TimeSeries`.
2. Type of the matrices that represents Mean Maximum Generation and Pumping data implemented in AreaScratchpad class is changed into newly implemented type `TimeSeries`.
3. These previous changes affected usage of Maximum Generation and Pumping data in solver, which required few changes in solver code.

To do:
- [x] Discard using raw matrix objects in function class members of the `DataHydroSeries` class and in unit tests. 
- [x] Change type arguments from `Matrix<double>&` to `TimeSeries&` for non-class static functions in **src/libs/antares/study/parts/hydro/series.cpp**


src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.h
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Antares::Data
+{
+/*!
+**  This class provides support for old studies, reading from deprecated files,
+**  fils matrix dailyMaxPumpAndGen and transfers data to the corresponding data
+**  class members of class PartHydro. Just versions below 8.7 will use instance
+**  of this class to be compatible with current implementation.
+*/
+class HydroMaxTimeSeriesReader
```
guilpier-code: This class is a reader for hydro time series, and it's external to **area.hydro** (of type **PartHydro**), meaning that **PartHydro** does not use **HydroMaxTimeSeriesReader** internally.
This externality seems to be a good thing, because class **PartHydro** should not have to read the TS it contains :  in case we decide to read TS differently, **PartHydro** stays unchanged.

(by the way, why are hydro TS like **maxDailyGenEnergy**, and **maxDailyPumpEnergy** not hosted in **area.hydro.series** [that is in class **DataSeriesHydro**] ?).

But this class is strange to me : 
- it has 2 public data members (an **enum** and a **Matrix**), that are **rightfully** used only internally. So why are they not **private** ?  Unit tests are using these data members from outside and this is not a good sign (to me).
- Internally, we read some TS data, and then we copy them to their real destination. Why not get a reference to their real destination, then work with, and avoid copies ?
- This class reads a specific TS (for hydro), but takes the whole area as an argument, though it should take at most a file path and a reference to the TS to be read (and a **usedByTheSolver**, ok).
- A class that needs data from the outside world to do its job should be given this data at construction whenever possible (otherwise, why are constructor for ?). Here, the constructor should get this data (path, TS&, ...). To me, it does not really make sense to have an **operator** here, because operator have no name, defining and calling it is unclear. A **read()** function is better.

guilpier-code (changes requested):

src/libs/antares/study/parts/hydro/container.h
```diff
@@ -172,8 +164,10 @@ class PartHydro
     //! Data for time-series
     DataSeriesHydro* series;
 
-}; // class PartHydro
+    Matrix<double, double> maxDailyGenEnergy;
+    Matrix<double, double> maxDailyPumpEnergy;
```
guilpier-code: **To be done some day** : 
These data members are both hydro time series, but are **not** contained in the `DataSeriesHydro* series;` data member. Is the current place for these **new** TS the right place ?
And if we choose to host them in class **DataSeriesHydro** (where the other hydro TS are hosted), the place where we load them should also change, from **container.cpp** to **list.cpp** (the place where the other TS are loaded).

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.h
nikolaredstork: 1.**maxDailyGenEnergy** and **maxDailyPumpEnergy** are not part of **DataSeriesHydro** class because we can not have multiple TS's like for hydro storage or ROR.
2.We can make them private.
3.We have old file with 4x365 matrix and we are loading that data in the **dailyMaxPumpAndGen** matrix. From **dailyMaxPumpAndGen** matrix we are distributing data to the other 4 destinations (every column of the **dailyMaxPumpAndGen** matrix will be copied to the one of the 4 destinations). That destinations are two matrices from the **PartHydro** class **maxDailyGenEnergy** and **maxDailyPumpEnergy** and other two matrices from the **DataSeriesHydro** class **maxHourlyGenPower** and **maxHourlyPumpPower**.  These four matrices (destinations) are used in solver, and they must be filled with appropriate data and if it is a case of an old study those data are coming from **dailyMaxPumpAndGen**  matrix, which we load with deprecated file (maxpower_areaid.txt). Don't see a way without coping these data, because we don't want to use  **dailyMaxPumpAndGen** directly in solver, because that would complicate code even more. Also we can't use move assignment operator, because we have situation that from one source we are distributing data to 4 destinations, maybe I'm wrong about this?
4.Regarding other two points, that can be implemented also.

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -114,146 +116,121 @@ bool PartHydro::LoadFromFolder(Study& study, const AnyString& folder)
...
+          area.hydro.useHeuristicTarget = true;
+          area.hydro.useLeeway = false;
+          area.hydro.powerToLevel = false;
+          area.hydro.leewayLowerBound = 1.;
+          area.hydro.leewayUpperBound = 1.;
+          area.hydro.initializeReservoirLevelDate = 0;
+          area.hydro.reservoirCapacity = 0.;
+          area.hydro.pumpingEfficiency = 1.;
+
+          if (study.header.version >= 870)
```
guilpier-code: Loading hydro energy TS for **version >= 870** is done here, but is done in **list.cpp** for **version < 870**. It should not be done that way.
We should take care of this some day.

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.h
nikolaredstork: Addition to the third statement. If we had load function for class **Matrix** that can receive as argument range of the columns that we want to load, not to load whole txt file, just columns that we want, into our **Matrix** instance, than this **HydroReader** class would be useless, because than we could load data from the deprecated file directly to our 4 destinations **(maxDailyGenEnergy, maxDailyPumpEnergy, maxHourlyGenPower, maxHourlyPumpPower)**, and we wouldn't need **dailyMaxPumpAndGen** matrix to distribute data.

guilpier-code: - Answer to **1.**  : we don't have to do it now, but it would seem that the natural place for **maxDailyGenEnergy** and **maxDailyPumpEnergy** is **DataSeriesHydro** (at least, I can't see a reason why not).
- Answer to **3.** : you're right, I missed that point.
- Answer to the very last comment : **absolutely**, but unfortunately, it seems impossible. As you said, the only real mission of this class is to load a 4 columns matrix and extract the columns we're interested in. Apart from that, we wouldn't need a class.

guilpier-code (changes requested):

src/libs/antares/study/area/scratchpad.cpp
```diff
@@ -35,9 +35,37 @@ using namespace Yuni;
...
+
+bool doWeHaveOnePositiveMaxDailyEnergy(const Matrix<double, int32_t>& power,
+    const Matrix<double>::ColumnType& energy)
+{
+    for (uint tsNumber = 0; tsNumber < power.width; ++tsNumber)
+    {
+        for (uint day = 0; day < DAYS_PER_YEAR; ++day)
+        {
+            if (power[tsNumber][day] * energy[day] > 0.);
+            return true;
```
nikolaredstork: Indent return statement and remove ; at the end of the if statement.

guilpier-code: Done by @nikolaredstork

src/libs/antares/study/parts/hydro/container.h
guilpier-code: A TODO comment was added in the code, we'll handle this later.

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.h
nikolaredstork: My idea is to remove copy functions from HydroReader class (copy function are used to copy data to the **maxDailyGenEnergy** and **maxDailyPumpEnergy**) and to place them in the PartHydro class, like you did for the **maxHourlyGenPower** and **maxHourlyPumpPower** because of the **feature envy**. Then we will have something like this:
image

What is you opinion?

guilpier-code: Honestly, I don't know yet what to think about it. 
The problem seems more global : we should think at the **hydro time series** scale.
I mean that hydro part has several time series, and several ways to load them (depending on which is the study version, if the loader is UI or Solver, and more [related to TS generation]). 
What should we do to first clarify this first ?
Imagining a more global way to load would help, I think.

And there is the **how** we do it. Do we want external TS loader (and saver) ? This means that we would empty class   **PartHydro** from all loading and saving functions. This could be a way to respect more the **single responsibility principle**, which requires small classes. And **PartHydro** has many lines of code.

I'm sorry I can't be more specific. 

About the change you suggest, the loading part would stay externalized, but the copy would be internalized (done in **PartHydro**), right ? What would be the benefit of this ?

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
```diff
@@ -0,0 +1,136 @@
...
+    dailyNbHoursAtPumpPmax.reset(1U, DAYS_PER_YEAR, true);
+
+    dailyNbHoursAtPumpPmax.pasteToColumn(0, dailyMaxPumpE);
+}
+
+bool HydroMaxTimeSeriesReader::operator()(const AnyString& folder, Area& area, bool usedBySolver)
+{
+    bool ret = true;
+
+    ret = loadDailyMaxPowersAndEnergies(folder, area, usedBySolver) && ret;
```
flomnes:
```suggestion
    bool ret = loadDailyMaxPowersAndEnergies(folder, area, usedBySolver);
```

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
```diff
@@ -0,0 +1,136 @@
...
+    copyDailyMaxEnergy(area);
+    area.hydro.series->buildMaxPowerFromDailyTS(dailyMaxPumpAndGen[genMaxP], dailyMaxPumpAndGen[pumpMaxP]);
```
flomnes:
```suggestion
    area.hydro.series->buildHourlyMaxPowerFromDailyTS(dailyMaxPumpAndGen[genMaxP], dailyMaxPumpAndGen[pumpMaxP]);
```

src/libs/antares/study/area/list.cpp
```diff
@@ -897,12 +897,41 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
+        if (study.header.version < 870)
+        {
+            buffer.clear() << study.folderInput << SEP << "hydro";
+
+            HydroMaxTimeSeriesReader reader;
+            ret = reader(buffer, area, study.usedByTheSolver) && ret;
+        }
+
+        if (study.header.version >= 870)
+        {
```
```suggestion
        else        {
```

src/libs/antares/study/area/list.cpp
```diff
@@ -897,12 +897,41 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
         {
             buffer.clear() << study.folderInput << SEP << "hydro" << SEP << "series";
-            ret = area.hydro.series->loadFromFolder(study, area.id, buffer) && ret;
+            ret = hydroSeries->loadGenerationTS(area.id, buffer, study.header.version) && ret;
+
+            hydroSeries->EqualizeGenerationTSsizes(
+                area, study.usedByTheSolver, study.gotFatalError);
         }
 
+        if (study.header.version < 870)
```
```suggestion
        if (study.header.version < 890)
```

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.h
flomnes:
```cpp
loadForVersionXXX(); // old
loadForVersionYYY(); // new
convertFromXXXtoYYY();

load(int version) {
  if(version <= XXX)
  {
    loadForVersionXXX();
    convertFromXXXtoYYY();
  }
  else
    loadForVersionYYY();
}
```
To be discussed

src/antares-deps
flomnes, line 1: revert this ?

nikolaredstork: Accidentally pushed. I've tried to delete folder, but that is not the solution. How to resolve this?

nikolaredstork: Accidentally pushed.

flomnes:
My compiler (GCC 9) gives the following warning
```
/home/omnesflo/Antares_Simulator/src/libs/antares/study/scenario-builder/HydroMaxPowerTSNumberData.cpp:12:61: warning: unused parameter ‘study’ [-Wunused-parameter]
   12 | uint hydroMaxPowerTSNumberData::get_tsGenCount(const Study& study) const
```
@nikolaredstork can you please fix this ?

flomnes:
Move input TS from
- `input/hydro/common/capacity/maxDailyGenEnergy_area.txt` -> `input/hydro/series/area/maxDailyGenEnergy.txt`
- `input/hydro/common/capacity/maxDailyPumpEnergy_area.txt` -> `input/hydro/series/area/maxDailyPumpEnergy.txt`

It would be more consistent with the Hourly version.

Also, why do we try to read the daily data (for example input/hydro/common/capacity/maxDailyGenEnergy_area.txt) when `study.version >= 870` ? Is this data still useful ?

---

Reply by @nikolaredstork 
This data is useful and we are inserting that data via:

image

Milos-RTEi:
https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1723#issuecomment-1808652601

Hi all. Regarding the comment above.
Can we please in the future (for example CR24 now) concentrate on the provided CR specification.
Especially the "'File Format changes" chapter. 
This (file/folder proposed locations) has been specified in the word document from day one.
Changing it immediately (during the design process) would not be a major problem at all.
Now, so late in the development process it really consumes a lot of time.

flomnes:
> [#1723 (comment)](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1723#issuecomment-1808652601)
> 
> Hi all. Regarding the comment above. Can we please in the future (for example CR24 now) concentrate on the provided CR specification. Especially the "'File Format changes" chapter. This (file/folder proposed locations) has been specified in the word document from day one. Changing it immediately (during the design process) would not be a major problem at all. Now, so late in the development process it really consumes a lot of time.

@Milos-RTEi I agree with you. Let's keep it this way then.

flomnes:
@Milos-RTEi @nikolaredstork 
OK so the logic is the following 
logic

Am I right ?

Yes, you are completely right.

Milos-RTEi:
https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1723#issuecomment-1810333803
Yes!

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
nikolaredstork: Done! Look new PR https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1784

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
nikolaredstork: Done!

src/libs/antares/study/area/list.cpp
nikolaredstork: Done!

guilpier-code (dismissed):
About : **src/antares-deps**
This **git module** was (not intentionally) changed.
It will corrupt use of this module on **develop** branch if we merge.
We should cancel this change before merging.

==> I fixed this, **no need to take care of it**

src/libs/antares/study/area/list.cpp
guilpier-code: Still a lot of **870** in the code base, we should take care of this and turn them to **890**.

src/antares-deps
guilpier-code: I took care of this, it has now disappear

nikolaredstork: Thanks!

Milos-RTEi: Thanks a lot!! It was killing us, how to fix it!

src/libs/antares/study/parameters.cpp
```diff
@@ -1114,6 +1124,22 @@ void Parameters::fixGenRefreshForNTC()
     }
 }
 
+void Parameters::fixGenRefreshForHydroMaxPower()
+{
+    if ((timeSeriesHydroMaxPower & timeSeriesToGenerate) != 0)
+    {
+        timeSeriesToGenerate &= ~timeSeriesHydroMaxPower;
+        logs.error() << "Time-series generation is not available for hydro max power. It "
+                        "will be automatically disabled.";
```
```suggestion
        logs.warning() << "Time-series generation is not available for hydro max power. It "
                        "will be automatically disabled.";
```

src/libs/antares/study/parameters.cpp
```diff
@@ -1114,6 +1124,22 @@ void Parameters::fixGenRefreshForNTC()
...
+    }
+    if ((timeSeriesHydroMaxPower & timeSeriesToRefresh) != 0)
+    {
+        timeSeriesToRefresh &= ~timeSeriesHydroMaxPower;
+        logs.error() << "Time-series refresh is not available for hydro max power. It will "
+                        "be automatically disabled.";
```
```suggestion
        logs.warning() << "Time-series refresh is not available for hydro max power. It will "
                        "be automatically disabled.";
```

docs/reference-guide/13-file-format.md
```diff
@@ -1,5 +1,12 @@
 # Study format changes
 This is a list of all recent changes that came with new Antares Simulator features. The main goal of this document is to lower the costs of changing existing interfaces, both GUI and scripts.
+## v9.1.0
+### (Input) Hydro Maximum Generation/Pumping Power
+* For each area, new files are added **input/hydro/series/&lt;area&gt;/maxHourlyGenPower.txt** and **input/hydro/series/&lt;area&gt;/maxHourlyPumpPower.txt**. These files have one or more columns, and 8760 rows. The number of columns in these two files must be the same if there is more than one column in each file, but if there is just one column for example in maxHourlyGenPower.txt file, maxHourlyPumpPower.txt file can have more than one column and vice versa. For old studies, file **input/hydro/common/capacity/maxpower_&lt;area&gt;** will be deleted after cleaning a study.
```
flomnes:
```suggestion
* For each area, new files are added **input/hydro/series/&lt;area&gt;/maxHourlyGenPower.txt** and **input/hydro/series/&lt;area&gt;/maxHourlyPumpPower.txt**. These files have one or more columns, and 8760 rows. The number of columns in these two files must be the same if there is more than one column in each file, but if there is just one column for example in maxHourlyGenPower.txt file, maxHourlyPumpPower.txt file can have more than one column and vice versa. Starting from v9.0, file **input/hydro/common/capacity/maxpower_&lt;area&gt;** is no longer used.
### How to upgrade capacity time-series ?
Source file = maxpower_&lt;area&gt;, destination files are maxHourlyGenPower.txt and maxHourlyPumpPower.txt. To upgrade time-series, you need to 
1. For generation, multiply 1st and 2nd rows. For pumping, multiply 3rd and 4rd rows.
2. Duplicate values 24 times : source has 365 rows, destination has 365 * 24 = 8760 rows
```

flomnes (approved):

PR #1725, opened 2023-10-30
[CI] Add sirius lib to PATH 

sylvlecl:
**Description**

On windows, running antares requires to have sirius DLL in the search path for libraries.
In particular, unit tests linked with sirius need that. 

The current implemented solution is a somewhat dirty workaround:
for each executable linked (possibly transitively) with sirius, we have a special CMake macro which needs to be used to copy sirius DLL to the same directory as the executable.

This requires developers to modify the build just to solve an installation issue, and this for **each** executable that are linked with sirius.

**Solution**
Instead, the PR proposes to add sirius installation directory to the PATH variable.
Executing unit tests will now rely on normal windows DLL search mechanism, and the developer only has to use the standard `target_link_libraries` in CMake files.


a-zakir:
i personally use `CMAKE_PREFIX_PATH` for sirius and ortools install directory, i think it's all the same 👍🏽

sylvlecl:
> i personally use `CMAKE_PREFIX_PATH` for sirius and ortools install directory, i think it's all the same 👍🏽

Ah, thanks! it sound better, if it works, it does not require the user to alter the environment outside of cmake related commands and files.
In that case, the best thing would be to remove completely DEPS_INSTALL_DIR from expected variables. It has no reason to exist in the first place in my opinion, we should have used CMAKE_PREFIX_PATH for this, it's the standard way to define dependencies directories.

sylvlecl:
Unfortunately, only `CMAKE_PREFIX_PATH` is not enough to make the execution work.
The DLL is still not found when executing test, see that run on a test branch: 
https://github.com/AntaresSimulatorTeam/Antares_Simulator/actions/runs/6737147648

```
The following tests FAILED:
	  5 - end-to-end-simple-study (Exit code 0xc0000135
)
	  6 - end-to-end-binding_constraints (Exit code 0xc0000135
)
```

So it seems that it's necessary to add it in the PATH.

a-zakir:
> Unfortunately, only `CMAKE_PREFIX_PATH` is not enough to make the execution work. The DLL is still not found when executing test, see that run on a test branch: https://github.com/AntaresSimulatorTeam/Antares_Simulator/actions/runs/6737147648
> 
> ```
> The following tests FAILED:
> 	  5 - end-to-end-simple-study (Exit code 0xc0000135
> )
> 	  6 - end-to-end-binding_constraints (Exit code 0xc0000135
> )
> ```
> 
> So it seems that it's necessary to add it in the PATH.

🤐

JasonMarechal25:
> **Solution** Instead, the PR proposes to add sirius installation directory to the PATH variable. Executing unit tests will now rely on normal windows DLL search mechanism, and the developer only has to use the standard `target_link_libraries` in CMake files.


Is the solution described up to date? The code change does not explicitly refer to target_link_library

Also, does it impact local development in any way? Looking only at the changelog I would say yes because dependency is not copied anymore but maybe it supported by other configuration?

sylvlecl:
> Is the solution described up to date? The code change does not explicitly refer to target_link_library

Yes it is: the goal of the change is that using `target_link_libraries` is enough, we don't need anymore to use `copy_dependency`.

> Also, does it impact local development in any way? Looking only at the changelog I would say yes because dependency is not copied anymore but maybe it supported by other configuration?

Indeed it does: the developer must ensure to have sirius in its PATH too (or any other place that windows searches for).
The developer must ensure his setup is correct for the OS he uses.
I will check where we could add some documentation for it.

For me it does not invalidate the solution: it does not make sense to copy libs we want to use in all directories where we want to execute something. The goal of shared libraries is to be shared from the one place where they are installed :)

a-zakir:
I’m takin it

flomnes (approved):

PR #1726, opened 2023-10-30
Reset hydro storage correctly

payetvin:
Fixed a bug in the GUI where storage size after reset was hour per year instead of day per year

### DIFF
--- a/src/libs/antares/series/include/antares/series/series.h
+++ b/src/libs/antares/series/include/antares/series/series.h
@@ -77,6 +77,7 @@ public:
     double* operator[](uint32_t index);
 
     void reset();
+    void reset(uint32_t width, uint32_t height);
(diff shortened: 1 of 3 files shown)


guilpier-code (approved):

nikolaredstork (commented):
It is cleared that reset of the storage TS's is done correctly, and it should serve it purpose. Now we have two reset function, but in fact we should only have one, the one with two arguments. That is something that should be considered for later work

guilpier-code:
> It is cleared that reset of the storage TS's is done correctly, and it should serve it purpose. Now we have two reset function, but in fact we should only have one, the one with two arguments. That is something that should be considered for later work

If we feel that the interface of **TimeSeries** is too big, we'll change it later.
I think we should merge now.

PR #1728, opened 2023-10-30
update form develop and simplifications on CR 23

guilpier-code:
Update CR 23 (branch  **max-gen-pump-cr23-develop-merge**) with **develop**, and make small simplifications


nikolaredstork (commented):
So far, so good, we started to use **TimeSeries** class member functions.

nikolaredstork (approved):

PR #1729, opened 2023-10-31
Simplification on CR23: ResizeTS function

nikolaredstork:
Changing function arguments of EqualizeTSsize fuction.

### DIFF
--- a/src/libs/antares/series/include/antares/series/series.h
+++ b/src/libs/antares/series/include/antares/series/series.h
@@ -86,7 +86,6 @@ public:
     bool forceReload(bool reload = false) const;
     void markAsModified() const;
     uint64_t memoryUsage() const;
-    void resizeTSNoDataLoss(uint width);
 
     TS timeSeries;
     numbers& timeseriesNumbers;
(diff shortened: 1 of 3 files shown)


src/libs/antares/series/include/antares/series/series.h
```diff
@@ -86,6 +86,7 @@ class TimeSeries
     bool forceReload(bool reload = false) const;
     void markAsModified() const;
     uint64_t memoryUsage() const;
+    void resizeTSNoDataLoss(uint width);
```
guilpier-code: I don't see the point of turning the old free function **resizeMatrixNoDataLoss** into a new data member function **resizeTSNoDataLoss**. 
To me, it's not necessary : we apparently need it **only** for hydro TS (so it's an implementation detail particular to hydro), and it makes generic class **TimeSeries** interface more complex.
We should keep **TimeSeries** interface as simple as possible.

guilpier-code (changes requested):

guilpier-code (approved):

Issue #1571, opened 2023-08-21
Remove variable numSpace


flomnes:
Basically, we have structures whose member data is duplicated for multi-threading, and indexed by numSpace, the thread number.

My advice would be to confine this `numSpace` variable to the highest level, so that this implementation detail doesn't leak everywhere. I think this work will mainly concern the code that generates the output, which is where most occurrences of `numSpace` are found. There are also a few in the hydro heuristics code.

Example (**src/solver/variable/economy/inflow.h**)
```cpp
void initializeFromStudy(Data::Study& study)
{
    pNbYearsParallel = study.maxNbYearsInParallel;

    InitializeResultsFromStudy(AncestorType::pResults, study);

    pValuesForTheCurrentYear = new VCardType::IntermediateValuesBaseType[pNbYearsParallel];
    for (unsigned int numSpace = 0; numSpace < pNbYearsParallel; numSpace++)
        pValuesForTheCurrentYear[numSpace].initializeFromStudy(study);

    // Next
    NextType::initializeFromStudy(study);
}
```

payetvin:
`numspace` is named `thread_number` in code related to adequacy patch (mainly in `src/solver/optimisation`)

It should be renamed `numspace` for consistency

flomnes:
As of June 2024, there are 2211 matches for `numSpace`, out of which 2123 matches in **src/solver/variables**. It seems tricky to remove it.

pet-mit:
ANT-2117

PR #1730, opened 2023-10-31
Refactor numspace use in hydro solver

payetvin:
#1571

### DIFF
--- a/src/solver/hydro/management/management.h
+++ b/src/solver/hydro/management/management.h
@@ -107,7 +107,7 @@ public:
                     unsigned int maxNbYearsInParallel,
                     Solver::IResultWriter& resultWriter);
 
-    double randomReservoirLevel(double min, double avg, double max, MersenneTwister& random);
+    static double randomReservoirLevel(double min, double avg, double max, MersenneTwister& random);
 
     //! Perform the hydro ventilation
     void makeVentilation(double* randomReservoirLevel,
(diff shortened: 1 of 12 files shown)


sylvlecl (changes requested):
As just discussed:
unfortunately, this will not work properly anymore in multithread.

We need to check if that case is correctly covered by CI !

src/solver/simulation/solver.hxx
```diff
@@ -717,9 +715,10 @@ void ISimulation<Impl>::computeRandomNumbers(randomNumbers& randomForYears,
             // Previous month's first day in the year
             int firstDayOfMonth = study.calendar.months[initResLevelOnSimMonth].daysYear.first;
 
-            double randomLevel = hydroManagement.randomReservoirLevel(min[firstDayOfMonth],
+            double randomLevel = HydroManagement::randomReservoirLevel(min[firstDayOfMonth],
```
sylvlecl: Can we make it a free function instead of a static function ?
Since it has nothing to do anymore with the `HydroManagement` class, it will make things clearer.

src/solver/hydro/management/management.h
```diff
@@ -107,39 +106,37 @@ class HydroManagement final
...
     void makeVentilation(double* randomReservoirLevel,
-                        Solver::Variable::State& state,
-                        uint y,
-                        uint numSpace);
+                         Solver::Variable::State& state,
+                         uint y,
+                         uint numSpace);
 
-    ALL_HYDRO_VENTILATION_RESULTS& ventilationResults() { return ventilationResults_; }
+    HYDRO_VENTILATION_RESULTS& ventilationResults() { return ventilationResults_; }
```
sylvlecl: While were are here: this should be `const HYDRO_VENTILATION_RESULTS& ventilationResults() const`
to make clear that the result will not be changed outside.

sylvlecl (changes requested):
minor comments

sylvlecl (approved):

PR #1470, opened 2023-07-31
Multiple Maximum Generation and Pumping for Hydro  (RTEi's -> CR23) 

Milos-RTEi:
This PR is created in order to implement RTEi development on CR23 (Multiple Maximum Generation and Pumping for Hydro).
Spec available at:
https://rteinternational.sharepoint.com/:w:/s/AntaresUser'sClub-Interne/EeDOZ0wxTLJHujcT1I0rMhMBsfOLYbf18_p9_fgWyp1dgQ?e=3uptbu
Test example available at:
https://rteinternational.sharepoint.com/:u:/s/AntaresUser'sClub-Interne/EVa5797c569OjLKHX0hGATAB6ofpwN2nLJvBoAKnchPfcA?e=8hDZBb

TODO:
* [x] Fix crashed unit & end-to-end test
* [x] Sonar
* [x] Unit Tests
* [ ] Refactor-review from RTE
* [x] Merge Conflicts

PR #1732, opened 2023-10-31
Small refactoring CR23 - File deletion

nikolaredstork:
Excluding deletion of deprecated files from load functions.
Comment from the main PR https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1470#discussion_r1364148657

### DIFF
--- a/src/libs/antares/study/study.h
+++ b/src/libs/antares/study/study.h
@@ -449,12 +449,6 @@ public:
     void removeTimeseriesIfTSGeneratorEnabled();
     //@}
 
-    /*!
-    ** \brief Delete deprecated files when upgrading a study
-    */
-    void deleteDeprecatedFiles(uint previousVersion);
-    //@}
-
     //! \name
     //@{
     /*!
--- a/src/libs/antares/study/study.h
+++ b/src/libs/antares/study/study.h
@@ -449,6 +449,12 @@ public:
     void removeTimeseriesIfTSGeneratorEnabled();
     //@}
 
+    /*!
+    ** \brief Delete deprecated files when upgrading a study
+    */
+    void deleteDeprecatedFiles(uint previousVersion);
+    //@}
+
     //! \name
     //@{
     /*!
--- a/src/tools/updater/main.cpp
+++ b/src/tools/updater/main.cpp
@@ -93,14 +93,12 @@ public:
                     study->parameters.readonly = true;
 
                 logs.info() << "Saving...";
-                uint tempVersionVar = (uint)study->header.version;
 
-                if (bool isSaved = study->saveToFolder(folder),
-                    versionUpgrade = tempVersionVar != (uint)Antares::Data::versionLatest;
-                    isSaved && versionUpgrade)
-                {
-                    study->deleteDeprecatedFiles(tempVersionVar);
-                }
+                uint tempVersionVar = (uint)study->header.version;
+                bool isSaved = study->saveToFolder(folder);
+
+                if (isSaved && (tempVersionVar != Antares::Data::versionLatest))
+                    study->deleteAllDeprecatedFiles(tempVersionVar);
             }
             else
             {
--- a/src/tools/updater/main.cpp
+++ b/src/tools/updater/main.cpp
@@ -93,7 +93,14 @@ public:
                     study->parameters.readonly = true;
 
                 logs.info() << "Saving...";
-                study->saveToFolder(folder);
+                uint tempVersionVar = (uint)study->header.version;
+
+                if (bool isSaved = study->saveToFolder(folder),
+                    versionUpgrade = tempVersionVar != (uint)Antares::Data::versionLatest;
+                    isSaved && versionUpgrade)
+                {
+                    study->deleteDeprecatedFiles(tempVersionVar);
+                }
             }
             else
             {
--- a/src/tools/updater/main.cpp
+++ b/src/tools/updater/main.cpp
@@ -93,12 +93,7 @@ public:
                     study->parameters.readonly = true;
 
                 logs.info() << "Saving...";
-
-                uint tempVersionVar = (uint)study->header.version;
-                bool isSaved = study->saveToFolder(folder);
-
-                if (isSaved && (tempVersionVar != Antares::Data::versionLatest))
-                    study->deleteDeprecatedFiles(tempVersionVar);
+                study->saveToFolder(folder);
             }
             else
             {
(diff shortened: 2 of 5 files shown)


src/ui/simulator/application/study.cpp
```diff
@@ -393,7 +393,14 @@ class JobSaveStudy final : public Toolbox::Jobs::Job
         }     // save as
```
nikolaredstork: @guilpier-code  This is obviously code duplication. Do you have proposal where potential function should be defined?

src/libs/antares/study/study.h
```diff
@@ -449,6 +449,12 @@ class Study: public Yuni::NonCopyable<Study>, public IObject, public LayerData
     void removeTimeseriesIfTSGeneratorEnabled();
     //@}
 
+    /*!
```
guilpier-code: Class **Study** is large enough, we plan to split it in many pieces, and we don't want it to grow.
- For now, we should make this function a free function. We could add it as a couple of source files file-removal.h/.cpp
- What it actually does is removing deprecated hydro max power input files, right ? Don't hesitate renaming it **removeDeprecatedHydroMaxPowerTSfiles**.
- this function should take 3 arguments : 
   - a list of areas (AreaList&), 
   - a input folder path (as an **std::string** or in a standard format, we don't want dependencies on **Yuni** strings, we want to get rid of it)
   - a current study version (as a unsigned int)
 
It could be something like : 
   
```c++
void removeDeprecatedHydroMaxPowerTSfiles(AreaList& areas, 
                                          std::string inputFolder, 
                                           unsigned int currentStudyVersion)
{
    if (currentStudyVersion >= 870)
        return;
    
    vector<std::string> areasNames;
    areas.each( [](const Area& area) { areasNames.push_back(area.id.to<std::string>(); });
    
    for (const name :  areasNames)
    {
        std::string filePath = folderInput + "/hydro/common/capacity/" + name + ".txt"
        removeFile(filePath); // To be coded in the same source file, in the .cpp file.
    }
}
```
 On the client side, it would be : 
```C++
removeDeprecatedHydroMaxPowerTSfiles(study.areas, study.folderInput.to<std::string>, study.header.version),
```

Note : about `removeFile(filePath)`, it probably already exists in the unit tests code area

src/tools/updater/main.cpp
```diff
@@ -93,7 +93,12 @@ class MyStudyFinder final : public Data::StudyFinder
                     study->parameters.readonly = true;
 
                 logs.info() << "Saving...";
-                study->saveToFolder(folder);
+
+                uint tempVersionVar = (uint)study->header.version;
```
guilpier-code: **tempVersionVar** : we should be more clear. This variable is the current study version, isn't it ?
So let's call it **currentStudyVersion**

src/tools/updater/main.cpp
```diff
@@ -93,7 +93,12 @@ class MyStudyFinder final : public Data::StudyFinder
                     study->parameters.readonly = true;
 
                 logs.info() << "Saving...";
-                study->saveToFolder(folder);
+
```
guilpier-code: Normally, we should add this to the **src/libs/antares/study/cleaner/cleaner.cpp**, which is supposed to clean any study from unsuitable files.

guilpier-code (changes requested):

src/libs/antares/study/study.h
nikolaredstork: The idea is to first read currentStudyVerstion and save it value to temporary variable, after that we are calling `study->saveToFolder`. Why we need temporary variable? We need it because, if we are upgrading a study, variable `study->.header.version` will be changed (after `study->saveToFolder` function call) to a current version of our software, and if we are passing value of the `study->.header.version` variable to the `deleteDeprecatedHydroMaxPowerTSFiles` function we will be passing already updated value of our study version (currently 870). If we are upgrading for example from 8.2 version than we need to pass 820 value to the `deleteDeprecatedHydroMaxPowerTSFiles` function, and we are doing that via temporary variable.
Condition `if (currentStudyVersion >= 870)`  must be changed to `if(previousVersion < 870)` . Why? Because  currentStudyVersion after saving a study will have 870 value, and deletion won't be performed.

src/ui/simulator/application/study.cpp
guilpier-code: > @guilpier-code This is obviously code duplication. Do you have proposal where potential function should be defined?

As you know, we don't want to enhance the UI code, we think it's not worth it.
Besides, it's not exactly code we touched for this CR 23.
But you're right, there is a code duplication.

src/libs/antares/study/study.h
nikolaredstork: Also, we need to delete files just if `study->saveToFolder()` function call return true, than we are sure that our new files are saved and all data from deprecated files are stored in our new files.

src/tools/updater/main.cpp
nikolaredstork: You are right. I've looked up cleaner.cpp file in details, and apparently after upgrading and saving our new files, we will delete deprecated files with clean command, so I don't think that we need automatic deletion after upgrade. With that been said, this PR shouldn't be merged and code for deletion of files should be excluded.

guilpier-code (changes requested):

src/libs/antares/study/study.h
guilpier-code: ok. Your last comments regard more the **updater/main.cpp** than the **src/libs/antares/study/study.h**.
So I answer you in **updater/main.cpp** (below)

src/tools/updater/main.cpp
guilpier-code: An answer to the previous comment from @nikolaredstork was given [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1732#discussion_r1377905709).

guilpier-code: Here is what think, eventually about  **tempVersionVar** and this added piece of code.
I think we should have : 
```C++
  ...
  bool isSaved = study->saveToFolder(folder);
  if (isSaved)
      deleteDeprecatedHydroMaxPowerTSFiles(study->areas, 
                                           study->folderInput.to<std::string>, 
                                           study->header.version);
``` 
Note we don't need the source (what you called **tmp**) study version here.
Why ?
When get here, a study was loaded, and it may have input deprecated files.
Then we upgrade it, meaning that we erase input data with fresh updated one.
Deprecated files (hydro max power TS in this case) can still remain.
Calling **deleteDeprecatedHydroMaxPowerTSFiles** will delete them if they exists and if the new study version requires it.
Note that the small **removeFile** function suggested in [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1732#discussion_r1377785454) removes a file only if it exists of course.

It's more simple like this, don't you think ?

guilpier-code (changes requested):

guilpier-code (approved):

Issue #1741, opened 2023-11-02
-z option not working on oraclelinux8


JasonMarechal25:
**Description**
Unable to use -z option on oracle linux 8

**How to Reproduce**
Execute a simulation with -z option

**Expected behavior**
Simulation is performed properly

**Screenshots and additional files**
`./antares-8.8-solver antares-xpansion-develop/data_test/examples/hurdles-cost-profile-value-over-one/ -z

[2023-11-02 14:23:47][solver][infos]   Output folder : /antares-8.8.0-rc2-OracleServer-8.8/bin/antares-xpansion-develop/data_test/examples/hurdles-cost-profile-value-over-one/output/20231102-1423eco
[2023-11-02 14:23:47][solver][error] Error opening entry about-the-study/comments.txt (-109)
[2023-11-02 14:23:47][solver][infos] 
[2023-11-02 14:23:47][solver][infos] Generating calendar informations
[2023-11-02 14:23:47][solver][error] Error opening entry about-the-study/study.ini (-109)
[2023-11-02 14:23:47][solver][error] Error opening entry about-the-study/parameters.ini (-109)
[2023-11-02 14:23:47][solver][error] Error opening entry info.antares-output (-109)
[2023-11-02 14:23:47][solver][error] Error opening entry about-the-study/areas.txt (-109)
[2023-11-02 14:23:47][solver][error] Error opening entry about-the-study/links.txt (-109)
[2023-11-02 14:23:47][solver][infos] Calendar: hours:1..8736, days:1..364, weeks:1..52, months:1..12, years:1..2
[2023-11-02 14:23:47][solver][infos] Simulation days per month : 31, 31, 30, 31, 30, 31, 31, 28, 31, 30, 31, 29
[2023-11-02 14:23:47][solver][infos] 
[2023-11-02 14:23:47][solver][infos] Removing disabled thermal clusters in from solver computations...
[2023-11-02 14:23:47][solver][infos] No disabled thermal cluster removed before solver computations
[2023-11-02 14:23:47][solver][infos] 
[2023-11-02 14:23:47][solver][infos] Optimizing the thermal clusters in 'must-run' mode...
[2023-11-02 14:23:47][solver][infos] No thermal cluster in 'must-run' mode
[2023-11-02 14:23:47][solver][infos] 
[2023-11-02 14:23:47][solver][infos] 
[2023-11-02 14:23:47][solver][infos] Summary

[...]

PR #1744, opened 2023-11-03
Fix oraclelinux minizip build + actually run zip unit tests

sylvlecl:
**Description**

See #1741: support for zip output is not working on our oracle linux distribution,
because zlib is not installed on the build image.

It should have been identified by zip writer unit test, but it's not run by the workflow, unfortunately.

**Quick fixes**
- Install zlib on oraclelinux build image
- Add missing "unit" label to relevant unit tests


flomnes (approved):
Reproduced runtime error in OL8 with v8.7.0

I was unable to verify if this patch fixes, since OL8 jobs are only run "rarely".

sylvlecl:
> I was unable to verify if this patch fixes, since OL8 jobs are only run "rarely".

I checked the fix once unit tests have been activated, in those workflows:
- this one [fails](https://github.com/AntaresSimulatorTeam/Antares_Simulator/actions/runs/6747263508/job/18343008211), before the fix (before addition of zlib)  
- this one [succeeds](https://github.com/AntaresSimulatorTeam/Antares_Simulator/actions/runs/6747672790), after the fix

PR #1749, opened 2023-11-06
Scratchpad numspace

payetvin:
#1571


src/solver/hydro/management/daily.cpp
```diff
@@ -233,7 +233,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::St
 
     auto& data = tmpDataByArea_[z];
 
-    auto& scratchpad = area.scratchpad[numSpace];
+    auto* scratchpad = scratchmap.at(&area);
```
flomnes: Unhandled exception ?

src/libs/antares/study/area/area.h
```diff
@@ -60,6 +60,7 @@ class Area final : private Yuni::NonCopyable<Area>
     using Vector = std::vector<Area*>;
     using VectorConst = std::vector<const Area*>;
     using List = std::list<Area*>;
+    using ScratchMap = std::map<const Area*, AreaScratchpad*>;
```
flomnes: Using a raw pointer as a map key works, but it seems a bit clumsy. Indeed, `std::map` is sorted. Depending on (arbitrary) pointer values, elements could be sorted differently.

As for the values, why do you use a raw pointer ?

src/solver/simulation/adequacy.cpp
```diff
@@ -73,7 +73,11 @@ bool Adequacy::simulationBegin()
         pProblemesHebdo.resize(pNbMaxPerformedYearsInParallel);
         for (uint numSpace = 0; numSpace < pNbMaxPerformedYearsInParallel; numSpace++)
         {
-            SIM_InitialisationProblemeHebdo(study, pProblemesHebdo[numSpace], 168, numSpace);
+            Antares::Data::Area::ScratchMap scratchmap;
+            study.areas.each([&scratchmap, &numSpace](Antares::Data::Area& a) {
+                    scratchmap.try_emplace(&a, &a.scratchpad[numSpace]); });
+
+            SIM_InitialisationProblemeHebdo(study, pProblemesHebdo[numSpace], 168, scratchmap);
```
flomnes: Why not move l.76-78 into `SIM_InitialisationProblemeHebdo` ? After all, these are temporary variables used solely by `SIM_InitialisationProblemeHebdo`.

The last argument could be removed.

flomnes (commented):
This PR is a step in the right direction.

However, to be complete, I think that area.scratchpad should be removed completely. Thus, only newly-introduced `scratchmap`s would be used.

src/libs/antares/study/area/area.h
payetvin: Fixed, now using a reference for AreaScratchpad

flomnes:
Need test on parallel run

src/solver/simulation/adequacy.cpp
payetvin: My goal was to lock the numspace responsibility to economy/adequacy.cpp
I can move it to `SIM_InitialisationProblemeHebdo`

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -147,9 +147,15 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
         break;
     }
 
+    Antares::Data::Area::ScratchMap scratchmap;
+    study.areas.each([&scratchmap, &numspace](Antares::Data::Area& a) {
+            scratchmap.try_emplace(&a, a.scratchpad[numspace]); });
```
flomnes: You could define a small wrapper around the map that would allow you to just write
```suggestion
    Antares::Data::Area::ScratchMap scratchmap(study.areas);
```
ie you transfer the responsability of the creation to the constructor of `ScratchMap`

flomnes (commented):
Some instances of `area.scratchpad` remain. Would it be possible to eliminate them as well ?

JasonMarechal25 (approved):

PR #1752, opened 2023-11-08
Simplify TS numbers drawings, fix bug related to refresh & local thermal generation

guilpier-code:
We intend to simplify and clarify the TS number management (specially source file **timeseries-numbers.cpp**).
For this we did several things : 
- [x] For generated TS, separate resizing from generation, and call the resizing before drawing TS number, so that we have the definitive TS sizes before we draw the TS numbers : this is the **main leverage** for TS number management simplifications and clarification.
- [x] At construction, the 1-d matrix **timeseriesNumbers** is given zeros. This allows many simplifications and code reductions.
- [ ] For productions (only for **Load**, **Wind** and **Solar**) that generate null TS (due to the configuration of their TS generator [see prepro capacity]), raise a warning somewhere for user to know.


src/solver/ts-generator/generator.hxx
```diff
@@ -29,12 +29,52 @@
...
 #include <antares/logs/logs.h>
 
-namespace Antares
+namespace Antares::Solver::TSGenerator
 {
-namespace Solver
-{
-namespace TSGenerator
+ 
+inline void ResizeGeneratedTimeSeries(Data::AreaList& areas, Data::Parameters& params)
```
sylvlecl: It does not seem templated --> should go in a .cpp

src/solver/ts-generator/generator.hxx
```diff
@@ -29,12 +29,52 @@
...
 {
+    for (auto i = areas.begin(); i != areas.end(); ++i)
```
sylvlecl: We should use range-based for loop, no need for iterator.

src/solver/ts-generator/generator.hxx
```diff
@@ -29,12 +29,52 @@
...
+
+        // Hydro
+        if (params.timeSeriesToGenerate & Data::timeSeriesHydro)
+        {
+            area.hydro.series->resizeRORandSTORAGE(params.nbTimeSeriesHydro);
+        }
+
+        // Thermal
+        auto end = area.thermal.list.mapping.end();
+        for (auto it = area.thermal.list.mapping.begin(); it != end; ++it)
```
sylvlecl: we should use range-based for loop, no need for iterator

sylvlecl (commented):
A few minor comments

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -286,24 +252,17 @@ class IntraModalConsistencyChecker
...
-    {
-        if (not checkTSconsistency())
-            return false;
-    }
-    else
-        nbTimeseries_ = tsCounter_->getGeneratedTimeSeriesNumber();
-
+    
+    if (not checkTSconsistency())
+        return false;
```
flomnes: `return checkTSconsistency();`

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -364,19 +322,20 @@ bool checkIntraModalConsistency(array<uint, timeSeriesCount>& nbTimeseriesByMode
...
-          tsKind, isTSintramodal, isTSgenerated, tsRetriever, study);
-        if (!intraModalchecker.check())
-            return false;
-        nbTimeseriesByMode[indexTS] = intraModalchecker.getTimeSeriesNumber();
+        if (isTSintramodal[indexTS])
+        {
+            IntraModalConsistencyChecker intraModalchecker(tsKind, tsRetriever, study);
+            if (!intraModalchecker.check())
+                return false;
+            nbTimeseriesByMode[indexTS] = intraModalchecker.getTimeSeriesNumber();
```
flomnes: Careful, `nbTimeseriesByMode[indexTS]` isn't filled if `!intraModalchecker.check()` is `false`. Is this intended ?

src/solver/ts-generator/generator.hxx
flomnes: And `inline` can be removed, too

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -236,10 +236,11 @@ void DataSeriesHydro::reset()
...
 }
 
-void DataSeriesHydro::resizeRORandSTORAGE(unsigned int width)
+void DataSeriesHydro::resetGenerationTS(unsigned int width)
 {
-    ror.resize(width, HOURS_PER_YEAR);
-    storage.resize(width, DAYS_PER_YEAR);
+    ror.reset(width, HOURS_PER_YEAR);
+    storage.reset(width, DAYS_PER_YEAR);
+    mingen.reset(width, HOURS_PER_YEAR);
```
flomnes: This will discard user input in some cases.

flomnes (changes requested):

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -716,14 +597,11 @@ void drawAndStoreTSnumbersForNOTintraModal(const array<bool, timeSeriesCount>& i
     for (auto& group: study.bindingConstraintsGroups)
     {
         const auto nbTimeSeries = group->numberOfTimeseries();
-        auto& value = group->timeseriesNumbers[0][year];
-        if (nbTimeSeries == 1)
+        auto& groupTsNumber = group->timeseriesNumbers[0][year];
+        groupTsNumber = 0;
+        if (nbTimeSeries != 1)
```
flomnes:
```suggestion
        if (nbTimeSeries == 1)
                groupTsNumber = 0;
        else
                // ...
```

src/solver/ts-generator/generator.h
```diff
@@ -34,12 +34,11 @@
 #include <antares/writer/i_writer.h>
 #include "xcast/xcast.h"
 
-namespace Antares
-{
-namespace Solver
-{
-namespace TSGenerator
+namespace Antares::Solver::TSGenerator
```
flomnes:
```suggestion
namespace Antares::TSGenerator
```

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: Apparently, the previous comment is not valid anymore : the code changed since then

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: This part hasn't changed, it's still like it used to be.
But some changes should occur here at some point.

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: You're right : 
This function is called only when hydro TS are generated.

In this case, for **ror** and **storage** TS, we don't do anything new : TS must be resized to the number of TS the user gave in the TS management pane (or in the **input** equivalent on disk). The resized matrix is then filled when generating the TS.

**mingen** is not generated though, and the TS given by user (from files) must be kept intact.
So **mingen** TS matrix is resized to the generation number of TS (in TS management pane, hydro part) : 
- if new size > old user size, then extra columns are copies of first column
- if new size < old user size, the resizing is done, but that's all. This means dumping some old TS, but they won't be used anyway, because the TS number are necessarily drawn or chosen in such way they can't be reached.  

Note that for **mingen**, this operation of resizing and filling extra columns is done in the same spirit as it's done right after loading hydro ready made TS.
Eventually (at some point), generating or loading TS should be done at the same place, that is at the place the loading has now, so that : 
- the resizing / filling of mingen can be done once instead of twice today 
-  the  TS mangement is the simplest as possible after that stage of generation/load (no need to manage if TS are generated or loaded, they are just TS)

src/solver/ts-generator/generator.hxx
guilpier-code: Done

src/solver/ts-generator/generator.hxx
guilpier-code: I can't see how to do it. If you see, please show an example.

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: Now 1-d Matrix **timeseriesNumbers** is reset (it's given zeros everywhere) as soon as it is sized, so we don't need to assign it a zero when associated TS matrix is one column sized.

src/solver/ts-generator/generator.h
guilpier-code: You suggest that namespace **TSGenerator** should not be contained in namespace  **Solver**.
That's right.
I moved  **TSGenerator** from **Antares::Solver** to **Antares** (many changes).

src/libs/antares/series/series.cpp
```diff
@@ -74,24 +72,17 @@ int TimeSeries::saveToFile(const std::string& filename, bool saveEvenIfAllZero)
 
 double TimeSeries::getCoefficient(uint32_t year, uint32_t timestep) const
 {
-    if (timeSeries.width == 0)
```
guilpier-code, line 77: **Explanation** : by construction, **timeSeries** can never have a width == 0, so we don't need the previous complexity

src/libs/antares/series/series.cpp
```diff
@@ -74,24 +72,17 @@ int TimeSeries::saveToFile(const std::string& filename, bool saveEvenIfAllZero)
...
 const double* TimeSeries::getColumn(uint32_t year) const
 {
-    if (timeSeries.width == 0)
-        return emptyColumn.data();
     return timeSeries[getSeriesIndex(year)];
 }
 
 uint32_t TimeSeries::getSeriesIndex(uint32_t year) const
 {
-    if (timeSeries.width == 1)
```
guilpier-code, line 91: **Explanation** : now, at construction of **timeseriesNumbers**, default content is **zero** for every year, that's why we don't need the previous complexity.

src/libs/antares/study/area/area.cpp
```diff
@@ -277,14 +277,14 @@ void Area::resizeAllTimeseriesNumbers(uint n)
     }
     else
     {
-        load.series.timeseriesNumbers.resize(1, n);
```
guilpier-code, line 281: Explanation : now we reset (resize + assign **zeros**), which makes TS number **zero** by default. This allows to reduce the amount of code in **timeseries-numbers.cpp** (see changes in **GenerateDeratedMode(...)** for instance)

src/tests/src/solver/simulation/tests-ts-numbers.cpp
```diff
@@ -126,6 +127,7 @@ BOOST_AUTO_TEST_CASE(two_areas_with_5_ready_made_ts_on_load___check_intra_modal_
 	area_2->resizeAllTimeseriesNumbers(1 + study->runtime->rangeLimits.year[rangeEnd]);
 	area_2->load.series.timeSeries.resize(5, 1);
 
+	TSGenerator::ResizeGeneratedTimeSeries(study->areas, study->parameters);
```
guilpier-code, line 129: **Explanations** : Here we generate all TS we use. Now that we separated TS sizing and TS generation, we must call a function for both.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -87,122 +87,56 @@ static bool GenerateDeratedMode(Study& study)
     logs.info() << "  :: using the `derated` mode";
     if (study.parameters.useCustomScenario)
         logs.warning() << "The derated mode is enabled. The custom building mode will be ignored";
-
```
guilpier-code, line 90: **Explanations** : main changes are in this file.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -810,68 +660,6 @@ void applyMatrixDrawsToInterModalModesInArea(Matrix<uint32_t>* tsNumbersMtx,
...
-    if (width == 1)
-    {
-        for (uint year = 0; year < years; year++)
-        {
-            tsNumbers[0][year] = 0;
-        }
-    }
-}
-
-static void fixTSNumbersWhenWidthIsOne(Study& study)
```
guilpier-code, line 812: **Explanations** : now we that we made **zero** a  default value (at construction) for TS numbers, this makes these previous fix functions useless.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -722,14 +576,10 @@ void drawAndStoreTSnumbersForNOTintraModal(const array<bool, timeSeriesCount>& i
     for (auto& group: study.bindingConstraintsGroups)
     {
         const auto nbTimeSeries = group->numberOfTimeseries();
-        auto& value = group->timeseriesNumbers[0][year];
-        if (nbTimeSeries == 1)
-        {
-            value = 0;
```
guilpier-code, line 717: We don't need to assign **zero** to a TS number (in case associated size is **1**) : **zero** is the default value now

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -703,17 +564,10 @@ void drawAndStoreTSnumbersForNOTintraModal(const array<bool, timeSeriesCount>& i
             {
                 auto& link = *(it->second);
                 const uint nbTimeSeries = link.directCapacities.timeSeries.width;
-                if (nbTimeSeries == 1)
```
guilpier-code, line 695: We don't need to assign **zero** to a TS number (in case associated size is **1**) : **zero** is the default value now

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -256,63 +176,35 @@ class areaLinksTransCapaNumberOfTSretriever : public areaNumberOfTSretriever
...
     areaNumberOfTSretriever* tsCounter_;
     Study& study_;
-    uint nbTimeseries_;
+    uint nbTimeseries_ = 0;
     string tsTitle_;
 };
 
-bool IntraModalConsistencyChecker::check()
-{
-    if (isTSintramodal_ && not isTSgenerated_)
```
guilpier-code, line 300: **Explanations** : we don't have to know if TS are generated or not anymore, that's why this function is removed

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -912,28 +700,18 @@ bool TimeSeriesNumbers::Generate(Study& study)
     array<uint32_t, timeSeriesCount> intramodal_draws;
     std::fill(intramodal_draws.begin(), intramodal_draws.end(), 0);
 
-    const array<bool, timeSeriesCount> isTSgenerated
```
guilpier-code, line 897: **Explanations** : same as above : no need to know if TS are generated or not, so no need for this array anymore.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -236,11 +236,22 @@ void DataSeriesHydro::reset()
     count = 1;
 }
 
-void DataSeriesHydro::resizeRORandSTORAGE(unsigned int width)
+void DataSeriesHydro::resize_ROR_STORAGE_MINGEN_whenGeneratedTS(unsigned int newWidth)
 {
-    ror.resize(width, HOURS_PER_YEAR);
-    storage.resize(width, DAYS_PER_YEAR);
-    count = width;
+    ror.resize(newWidth, HOURS_PER_YEAR);
```
guilpier-code, line 250: **Explanations** : in case we generate hydro TS, **ror** and **storage** are generated, but not **mingen**.
So **ror** and **storage** are resized and completely overridden. But we must be careful with **mingen** TS : it is only resized, but we don't want to loose existing **mingen** TS given by the user : 
- If we extend **mingen** TS, we keep TS given by users (and fill the extra ones with a copy of the first TS)
- if we reduce **mingen** TS, we can remove extra TS given by user, but we must keep the left ones intact

src/solver/simulation/solver.hxx
```diff
@@ -312,6 +312,8 @@ void ISimulation<Impl>::run()
 
     ImplementationType::setNbPerformedYearsInParallel(pNbMaxPerformedYearsInParallel);
 
+    TSGenerator::ResizeGeneratedTimeSeries(study.areas, study.parameters);
```
guilpier-code, line 315: **Explanations** : this is the place where we call the resizing for TS to be generated.

src/solver/ts-generator/hydro.cpp
```diff
@@ -40,18 +40,12 @@ using namespace Yuni;
...
 
+using namespace Antares::Solver;
+
 namespace Antares
 {
-namespace Solver
-{
 namespace TSGenerator
 {
-static void PreproHydroInitMatrices(Data::Study& study, uint tsCount)
```
guilpier-code, line 49: **Explanations** : we separated resizing from generation, so we moved this code to another place (see **generator.cpp**)

src/solver/ts-generator/thermal.cpp
```diff
@@ -251,15 +248,11 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
 
     if (0 == cluster.unitCount or 0 == cluster.nominalCapacity)
     {
-        cluster.series.timeSeries.reset(1, nbHoursPerYear);
-
```
guilpier-code, line 255: Explanations : due to separation between TS resizing and generation, this code (**resize**) was moved elsewhere

src/solver/ts-generator/xcast/studydata.hxx
```diff
@@ -57,18 +55,12 @@ void StudyData::loadFromStudy(Data::Study& study,
             logs.info() << "  Added the area '" << area.name << "'";
             localareas.push_back(&area);
         }
-        else
```
guilpier-code, line 60: **Explanations** : code moved elsewhere due to separation between TS resizing and generation

src/solver/ts-generator/generator.hxx
guilpier-code: Done : for this, I used **each()**

src/solver/ts-generator/generator.hxx
guilpier-code: Thanks to @JasonMarechal25 who suggested a correct solution (see code)

src/solver/ts-generator/generator.hxx
```diff
@@ -29,12 +29,52 @@
...
+        {
+            auto& cluster = *(it->second);
+            bool globalThermalTSgeneration = params.timeSeriesToGenerate & Data::timeSeriesThermal;
```
JasonMarechal25: This constant can go outside the loop

JasonMarechal25 (approved):

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -87,122 +87,56 @@ static bool GenerateDeratedMode(Study& study)
...
-    std::vector<uint> getAreaTimeSeriesNumber(const Area& area)
-    {
-        std::vector<uint> to_return = {area.load.series.timeSeries.width};
-        return to_return;
-    }
-    uint getGeneratedTimeSeriesNumber()
+    std::vector<uint> getAreaTimeSeriesNumber(const Area& area) override
     {
-        return study_.parameters.nbTimeSeriesLoad;
+        return { area.load.series.timeSeries.width };
```
payetvin, line 105: Remove braces in return statement

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -87,122 +87,56 @@ static bool GenerateDeratedMode(Study& study)
...
-    std::vector<uint> getAreaTimeSeriesNumber(const Area& area)
-    {
-        std::vector<uint> to_return = {area.hydro.series->TScount()};
-        return to_return;
-    }
-    uint getGeneratedTimeSeriesNumber()
+    std::vector<uint> getAreaTimeSeriesNumber(const Area& area) override
     {
-        return study_.parameters.nbTimeSeriesHydro;
+        return { area.hydro.series->TScount() };
```
payetvin, line 114: Same here

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -87,122 +87,56 @@ static bool GenerateDeratedMode(Study& study)
...
-        {
-            auto& cluster = *(area.renewable.clusters[i]);
-            cluster.series.timeseriesNumbers.zero();
-        }
-    });
-
     return true;
 }
 
 class areaNumberOfTSretriever
```
payetvin: Class names should start with a capital letter by convention

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -87,122 +87,56 @@ static bool GenerateDeratedMode(Study& study)
...
 {
 public:
-    areaNumberOfTSretriever(Study& study) : study_(study)
-    {
-    }
+    virtual ~areaNumberOfTSretriever() = default;
     virtual std::vector<uint> getAreaTimeSeriesNumber(const Area& area) = 0;
```
payetvin, line 97: There's specific cases for hydro and links needing to override the base implementation.

I think the others containers should using the default class and function
The default class should take the series as an argument in constructor.

This would allow to have like
```
class AreaNumberOfTSretriever
{
public:

const TimeSeries& series;

AreaNumberOfTSretriever(const TimeSeries& s) : series(s)
{}

std::vector<uint> getAreaTimeSeriesNumber()
{
    return series.width;
}
};
```

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -87,122 +87,56 @@ static bool GenerateDeratedMode(Study& study)
...
 
 class thermalAreaNumberOfTSretriever : public areaNumberOfTSretriever
 {
 public:
-    virtual ~thermalAreaNumberOfTSretriever() = default;
-    thermalAreaNumberOfTSretriever(Study& study) : areaNumberOfTSretriever(study)
-    {
-    }
-    std::vector<uint> getAreaTimeSeriesNumber(const Area& area)
+    std::vector<uint> getAreaTimeSeriesNumber(const Area& area) override
```
payetvin, line 139: Thermal and renewable behave the same way, using common/cluster_list
We may be able to merge them by passing only clusters

src/solver/ts-generator/xcast/xcast.cpp
```diff
@@ -43,12 +43,13 @@ using namespace Yuni;
...
 namespace Antares
 {
-namespace Solver
-{
 namespace TSGenerator
 {
 namespace XCast
 {
+
+const unsigned int nbHoursADay = 24;
```
payetvin: Maybe define in in constants.h with HOURS_PER_YEAR ?
It's fine declaring it locally so up to you

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -236,11 +236,22 @@ void DataSeriesHydro::reset()
...
+    storage.resize(newWidth, DAYS_PER_YEAR);
+
+    // Resizing mingen : new extra columns are copies of the first column.
+    //    Note that mingen has necessarily one column at least, by construction
```
flomnes: Why do we want to do that ?

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: We want to return a **std::vector**, so we should keep it that way, should we not ?

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: Same answer as above

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: Yes, you're right !
What you recommend is planned for another PR (about re-thinking this file's design).

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: Sure, you're right, but as you know, there are many changes to come from another PR, for example : **clusterCount()** is about to disappear from **thermal/container.h**, and other things.
To remove the code duplication you recommend, we could wait a bit, so that the part about clusters simplification is done.

src/solver/ts-generator/xcast/xcast.cpp
guilpier-code: Done

src/solver/simulation/timeseries-numbers.cpp
guilpier-code: Done

src/solver/ts-generator/generator.hxx
guilpier-code: Done 
**Caution** : due to another remark from @sylvlecl, this loop (and the function it belongs to) was moved to a new source file named **generator.cpp**

src/libs/antares/study/area/area.cpp
flomnes: Can put this as a comment in the code ?

src/libs/antares/study/parts/hydro/series.cpp
flomnes: Can you include this as a comment ?

src/libs/antares/study/area/area.cpp
guilpier-code: What kind of comment do you want to put exactly ?
I just made some more simplification here, based on the fact that n is a number of years, and cannot be **0**.

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: Please see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1752#discussion_r1417403948)

guilpier-code: > Why do we want to do that ?

Please see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1752#discussion_r1417403948)

src/libs/antares/study/area/area.cpp
guilpier-code: > Can put this as a comment in the code ?

What kind of comment do you want to put exactly ?
I also just made some more simplifications here, based on the fact that **n** is a number of years, and can never be **0** (I checked).

src/libs/antares/study/parts/hydro/series.cpp
guilpier-code: > Can you include this as a comment ?

Done

flomnes (approved):

PR #1754, opened 2023-11-08
Refactor numspace use in solver

payetvin:
#1571


flomnes:
What should be done with this PR ?

flomnes:
See #1757

PR #1757, opened 2023-11-09
Allow yearJob to be non-copyable

sylvlecl:
This is a technical preliminary for #1754 :
in order to wrap it into a `std::function` that can be used as a task, the `yearJob` now requires to be copyable.
This should not be a requirement on that class.

To circumvent this, we introduce an overload of `AddTask` which accepts a pointer to an arbitrary callable object.


a-zakir (approved):

src/libs/antares/concurrency/include/antares/concurrency/concurrency.h
```diff
@@ -82,6 +95,44 @@ class FutureSet
...
+namespace Detail { //implementation details
+
+/*!
+ * Utility class to wrap a callable object pointer
+ * into a copyable callable object.
+ *
+ * @tparam T the underlying callable type
+ */
+template<class T>
+class SharedCallable
```
flomnes:
```suggestion
class CopyableCallable
```

src/tests/src/libs/antares/concurrency/test_concurrency.cpp
flomnes, line 1: Add test on `std::function` itself ?

src/libs/antares/concurrency/include/antares/concurrency/concurrency.h
```diff
@@ -82,6 +95,44 @@ class FutureSet
...
+};
+
+}
+
+template <class T>
+TaskFuture AddTask(Yuni::Job::QueueService& threadPool,
+                                 const std::shared_ptr<T>& task,
+                                 Yuni::Job::Priority priority)
+{
+    Task wrappedTask = Detail::SharedCallable<T>(task);
```
flomnes: What is type `Task` ? I don't understand how `Detail::SharedCallable<T>(task)` can be of type `Task`.

sylvlecl: `Task` in this namespace is just an alias for std::function<void()>, so it can be assigned any function object (as long as it's copyable).

src/tests/src/libs/antares/concurrency/test_concurrency.cpp
sylvlecl: The signature with `std::function` is already extensively tested, but maybe I misunderstood.

flomnes: You're right

src/libs/antares/concurrency/include/antares/concurrency/concurrency.h
sylvlecl: Renamed!

flomnes (approved):

PR #1767, opened 2023-11-13
Fix output variable profit

### LINKED ISSUE #1753: Invalid values for output variable PROFIT
## Description
See [this study](https://antares-web-recette.rte-france.com/studies/4bc8e4a7-0070-4e36-ab8e-7b3172a584de), in area **SK00**, thermal cluster **Lignite old 1**.

Most hourly values (including the one for the 1st hour) have wrong values.

## Developer notes
See https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/1cf1b86d734a6e538dd554562be4d0ce1ddafefa/src/solver/variable/economy/profitByPlant.h#L282-L285

- [OUTPUT] `hourlyClusterProduction` production of thermal cluster for given hour
- [INPUT] `PthetaInf` minimal production for a given hour
- [OUTPUT] `areaMarginalCosts` **MRG. PRICE**
- [INPUT] `cluster->getMarginalCost(tsIndex, hourInTheYear)` marginal cost associated to thermal cluster

payetvin:
close #1753

### DIFF
--- a/src/libs/antares/study/parts/thermal/cluster.cpp
+++ b/src/libs/antares/study/parts/thermal/cluster.cpp
@@ -137,6 +137,7 @@ Data::ThermalCluster::ThermalCluster(Area* parent) :
     forcedLaw(thermalLawUniform),
     plannedLaw(thermalLawUniform),
     PthetaInf(HOURS_PER_YEAR, 0),
+    pMin(HOURS_PER_YEAR, 0),
     costsTimeSeries(1, CostsTimeSeries())
 {
     // assert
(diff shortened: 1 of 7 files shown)


flomnes (changes requested):
This won't work in parallel execution since `ThermalCluster::pMin` is shared across threads. More specifically, multiple threads read/write the same memory block, possibly leading to different results sequential vs parallel.

flomnes:
We should probably re-generate some reference results, after checking that the changes are as expected.

flomnes (approved):

PR #1769, opened 2023-11-15
Allow absence of group attribute for binding constraints

### LINKED ISSUE #1765: Upgrading Study with binding constraints does not work
**Description**
Upgrading a study v8.6 to newest version (with GUI) does not upgrade BC groups (it seems).

**How to Reproduce**
Unzip the following study and upgrade it with GUI to newest version.
[047 PSP strategies-4-Opt weekly.zip](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/13333629/047.PSP.strategies-4-Opt.weekly.zip)
You'll see that, in **input/bindingconstraints/bindingconstraints.ini**, the (only) BC has no group (it should be in **default** group) and running it leads to a runtime error.

**OS and version**
 - OS: Windows (issue should be observed on Linux too)
 - Antares Simulator version: newest version

flomnes:


a-zakir (approved):

src/libs/antares/study/binding_constraint/BindingConstraintSaver.cpp
```diff
@@ -21,7 +21,9 @@ bool BindingConstraintSaver::saveToEnv(EnvForSaving& env, const BindingConstrain
     env.section->add("filter-year-by-year", datePrecisionIntoString(bindingConstraint->pFilterYearByYear));
     env.section->add("filter-synthesis", datePrecisionIntoString(bindingConstraint->pFilterSynthesis));
 
-    if (!bindingConstraint->group().empty())
+    if (bindingConstraint->group().empty())
```
guilpier-code: Shouldn't the default group be set when loading ?

flomnes: Yes, although it's rather equivalent

guilpier-code: Well, if we set the **default** group when saving, this probably means that, at least for some cases (old studies), the group name of a constraint is **empty** during the simulation, right ?
Is it a good thing ? I'm not sure it's clean. Any constraint should always belong to a group with a name, shouldn't it ?

flomnes: We will allow "no binding constraint group", see latest change

guilpier-code (approved):

PR #1787, opened 2023-11-23
Install Antares component

JasonMarechal25:
- Fix missing string.h installation in yuni
- Fix extra target_include_dir
- Add install instruction for all required component

Note: Some license headers lost a "*" that is due to applying CLion copyright profiles on commit and them not being able to prefix lines with "**"


src/libs/antares/concurrency/CMakeLists.txt
```diff
@@ -8,4 +8,12 @@ target_include_directories(concurrency PUBLIC
...
 
-target_link_libraries(concurrency yuni-static-core)
\ No newline at end of file
+target_link_libraries(concurrency yuni-static-core)
+
+install(DIRECTORY include/antares
+        DESTINATION "include"
+)
+
+install(DIRECTORY include/antares
```
sylvlecl: Duplicate instruction

sylvlecl: Duplicate instruction

src/solver/constraints-builder/CMakeLists.txt
```diff
@@ -27,6 +27,10 @@ target_link_libraries(antares-solver-constraints-builder PRIVATE antares-core
 )
 
 target_include_directories(antares-solver-constraints-builder
-	PUBLIC
-		$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+        PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
```
sylvlecl, line 31: To be checked:
I think we should add an "INSTALL_INTERFACE" instruction in all the `target_include_directories`, for the case where libs are used from their install directory.

src/solver/hydro/CMakeLists.txt
```diff
@@ -1,78 +1,85 @@
...
-	daily/h2o_j_initialiser_les_bornes_des_variables.cpp
-	daily/h2o_j_initialiser_les_second_membre.cpp
-	daily/h2o_j_instanciation.cpp
-	daily/h2o_j_optimiser_un_mois.cpp
-	daily/h2o_j_resoudre_le_probleme_lineaire.cpp
-	daily/h2o_j_lisser_les_sur_turbines.cpp
-    daily/h2o_j_ajouter_bruit_au_cout.cpp
+        daily/h2o_j_construire_les_contraintes.cpp
+        daily/h2o_j_construire_les_variables.cpp
+        include/antares/solver/hydro/daily/h2o_j_donnees_mensuelles.h
```
sylvlecl, line 6: To be confirmed:
I think we don't need to have those includes in the SRC lists, since they are in the target include directories.

sylvlecl (commented):
Mainly one question about the definition of the "install interface":
not 100% sure but I think that if you want to import the targets from their install directories (in xpansion build), we need to define the `target_include_directories` for the `INSTALL_INTERFACE` as well, like:
```
target_include_directories(mylib PUBLIC
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
  $<INSTALL_INTERFACE:include>
)
```
First line is used when building antares-simulator, and second line tells the user of the installed lib (xpansion) that includes are in `<install dir>/include`.

sylvlecl (approved):
As just discussed:
let's keep in mind the pending question about install interface, but merge as-is for now since in practice the installed target is correctly defined.

src/solver/hydro/CMakeLists.txt
JasonMarechal25: Headers in add_library command is not mandatory but helpful to IDE

src/solver/constraints-builder/CMakeLists.txt
JasonMarechal25: I have the same understanding but it works nonetheless. I choose not to add INSTALL_INTERFACE directives to prevent clutering

PR #1788, opened 2023-11-23
remove unuseful yuni string separator

a-zakir:


src/solver/utils/ortools_utils.cpp
```diff
@@ -214,12 +214,8 @@ static void extract_from_MPSolver(MPSolver* solver,
 
 std::string generateTempPath(const std::string& filename)
 {
-    namespace fs = std::filesystem;
-    std::ostringstream tmpPath;
-    tmpPath << fs::temp_directory_path().string() << Yuni::IO::SeparatorAsString << filename;
-    return tmpPath.str();
+    return (std::filesystem::temp_directory_path() / filename).string().c_str();
```
a-zakir: https://en.cppreference.com/w/cpp/filesystem/path/string

a-zakir: Whenever

a-zakir:
see comments in https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1791

src/solver/utils/ortools_utils.cpp
```diff
@@ -212,14 +215,35 @@ static void extract_from_MPSolver(MPSolver* solver,
...
+        return parentDir;
+    }
+}
 std::string generateTempPath(const std::string& filename)
 {
-    namespace fs = std::filesystem;
-    std::ostringstream tmpPath;
-    tmpPath << fs::temp_directory_path().string() << Yuni::IO::SeparatorAsString << filename;
-    return tmpPath.str();
+    return (CreateRandomSubDir(std::filesystem::temp_directory_path()) / filename).string().c_str();
```
flomnes, line 245: This changes the behavior. How about keeping the same behavior as previously ?

a-zakir: see Pr [reviews](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1791)

a-zakir: and we may remove the last 2 commits

PR #1790, opened 2023-11-24
Add NaN in matrix load test

flomnes:
Check that `loadFromCSV` can load matrices containing the special double value `nan`.

### DIFF
--- a/src/tests/src/libs/antares/array/tests-matrix-load.cpp
+++ b/src/tests/src/libs/antares/array/tests-matrix-load.cpp
@@ -9,7 +9,7 @@
 
 #include <iostream>
 #include <fstream>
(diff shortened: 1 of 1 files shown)


a-zakir:
Serious objections have been raised for this.
Are you willing to listen?

a-zakir:
Thanks @flomnes, I'm not sure of the feasibility or the technical way at this stage, suppose we have in a bindingconstraints.ini file
`
name = Belgique_import
id = belgique_import

…
Deactivated_timeStep = deac_ts.txt`
Then before loading the matrix a « simple loop » on all timestep minus deactivated ones could do the trick.

a-zakir:
sounds good?

PR #1791, opened 2023-11-25
remove unuseful yuni string separator

a-zakir:


JasonMarechal25 (approved):

a-zakir:
Thanks @JasonMarechal25, would you merge it?

sylvlecl:
Note that sonar raises an interesting security issue that I think we should take into account, at least in another PR if not in that one:
the temp file has a too "predictable" name which can cause bugs (if another process handles the same study for example) or security issues.

a-zakir:
> Note that sonar raises an interesting security issue that I think we should take into account, at least in another PR if not in that one: the temp file has a too "predictable" name which can cause bugs (if another process handles the same study for example) or security issues.

Do you mean the « Make sure publicly writable **directories** are used safely here » message?
Otherwise we can:
Rename the function: catTmpPathWithMyFilename?( the filename (name)is decided by the caller)
- add a Rand generated sub dir name

sylvlecl:
Indeed a recommendation is to have a random part in the filename:
https://sonarcloud.io/project/security_hotspots?id=AntaresSimulatorTeam_Antares_Simulator&pullRequest=1788&sinceLeakPeriod=true&tab=how_to_fix

PR #1795, opened 2023-11-27
Runvcpkg + runcmake+presets

a-zakir:
@sylvlecl, There are few things left, feel free to suggest/modify this Pr.

GH Actions —->[here](https://github.com/BAZ-C/Antares_Simulator/pull/1/files#diff-ffa04b13dcfffb59639a56ff70a00f3c8c93b293c3f83bb4c438b86f0c3d86fd)

### DIFF
--- a/.github/workflows/runvcpkg-runcmake.yml
+++ b/.github/workflows/runvcpkg-runcmake.yml
@@ -138,6 +138,7 @@ jobs:
 
       - name: Enable git longpaths
         run:  git config --system core.longpaths true
+        if: matrix.os == 'windows-latest'
 
         
       - name: List $RUNNER_WORKSPACE before build
--- a/.github/workflows/runvcpkg-runcmake.yml
+++ b/.github/workflows/runvcpkg-runcmake.yml
@@ -86,17 +86,21 @@ jobs:
           path: 'antares-deps-version.json'
           prop_path: 'antares_deps_version'
 
-      - name: setpreset
+      - name: setpreset ub
         shell: bash 
         run: |
-          [[ ${{ matrix.os}} == 'ubuntu-latest' ]] 
           echo "configurePreset=ninja-multi-vcpkg" >> $GITHUB_ENV
           echo "configubuildPreset=ninja-multi-vcpkg" >> $GITHUB_ENV
           echo "archive_type=tgz" >> $GITHUB_ENV
-          [[ ${{ matrix.os}}  == 'windows-latest' ]] 
+        if: matrix.os == 'ubuntu-latest'
+        
+        - name: setpreset Win
+          shell: bash 
+          run: |
           echo "configurePreset=msbuild-vcpkg" >> $GITHUB_ENV
           echo "configubuildPreset=msbuild-vcpkg" >> $GITHUB_ENV
           echo "archive_type=zip" >> $GITHUB_ENV
+        if: matrix.os == 'windows-latest'
         
       - name: Download pre-compiled librairies (win)
         if: matrix.os == 'windows-latest'
--- a/.github/workflows/runvcpkg-runcmake.yml
+++ b/.github/workflows/runvcpkg-runcmake.yml
@@ -116,7 +116,8 @@ jobs:
         uses: ./.github/workflows/download-extract-precompiled-libraries-tgz
         with:
           antares-deps-version: ${{steps.antares-deps-version.outputs.prop}}
-          os: ${{matrix.os}}
+          #os: ${{matrix.os}}
+          os: ubuntu-20.04
           ortools-url: ${{env.ORTOOLS_URL}}
           ortools-dir: ${{env.ORTOOLS_DIR}}
           
--- a/.github/workflows/runvcpkg-runcmake.yml
+++ b/.github/workflows/runvcpkg-runcmake.yml
@@ -94,12 +94,12 @@ jobs:
           ortools-url: ${{env.ORTOOLS_URL}}
           ortools-dir: ${{env.ORTOOLS_DIR}}
 
-      - name: Setup Python 3.11
+      - name: Setup Python 3.12
         uses: actions/setup-python@v4
         id: setup-python
         with:
           architecture: 'x64'
-          python-version: '3.11'
+          python-version: '3.12'
 
       - name: Install pip dependencies if necessary
         run: pip install -r src/tests/examples/requirements.txt
--- a/.github/workflows/runvcpkg-runcmake.yml
+++ b/.github/workflows/runvcpkg-runcmake.yml
@@ -98,7 +98,7 @@ jobs:
           echo "configubuildPreset=msbuild-vcpkg" >> $GITHUB_ENV
           echo "archive_type=zip" >> $GITHUB_ENV
         
-      - name: Download pre-compiled librairies
+      - name: Download pre-compiled librairies (win)
         if: matrix.os == 'windows-latest'
         uses: ./.github/workflows/download-extract-precompiled-libraries-zip
         with:
@@ -106,6 +106,8 @@ jobs:
           os: ${{matrix.os}}
           ortools-url: ${{env.ORTOOLS_URL}}
           ortools-dir: ${{env.ORTOOLS_DIR}}
+          
+      - name: Download pre-compiled librairies
         if: matrix.os == 'ubuntu-latest'
         uses: ./.github/workflows/download-extract-precompiled-libraries-tgz
         with:
--- a/.github/workflows/runvcpkg-runcmake.yml
+++ b/.github/workflows/runvcpkg-runcmake.yml
@@ -86,8 +86,20 @@ jobs:
           path: 'antares-deps-version.json'
           prop_path: 'antares_deps_version'
 
+      - name: setpreset
+        shell: bash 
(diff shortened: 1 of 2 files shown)


.github/workflows/runvcpkg-runcmake.yml
```diff
@@ -0,0 +1,308 @@
...
+      - name: Setup vcpkg
+        uses: lukka/run-vcpkg@main
+        id: runvcpkg
+        with:
+          # This one is not needed, as it is the default value anyway.
+          vcpkgDirectory: '${{ github.workspace }}/vcpkg'
+          vcpkgJsonGlob: '**/vcpkg.json'
+          # let run-cmake do it
+          # runVcpkgInstall: true
+          doNotCache: false
```
sylvlecl: Might not be necessary since anyway it's just a download, but let's keep it if it's not too large in the cache!

.github/workflows/runvcpkg-runcmake.yml
```diff
@@ -0,0 +1,308 @@
...
+             mkdir -p vcpkg_installed
+             mkdir -p vcpkg_cache
+
+      - name: Setup vcpkg
+        uses: lukka/run-vcpkg@main
+        id: runvcpkg
+        with:
+          # This one is not needed, as it is the default value anyway.
+          vcpkgDirectory: '${{ github.workspace }}/vcpkg'
+          vcpkgJsonGlob: '**/vcpkg.json'
```
sylvlecl: According to doc, this is not needed when install is not done in this step (to be confirmed)

.github/workflows/runvcpkg-runcmake.yml
```diff
@@ -0,0 +1,308 @@
...
+      - name: Prints output of run-vcpkg's action.
+        run: echo "root='${{ steps.runvcpkg.outputs.RUNVCPKG_VCPKG_ROOT_OUT }}', triplet='${{ steps.runvcpkg.outputs.RUNVCPKG_VCPKG_DEFAULT_TRIPLET_OUT }}' "
+      
+
+      - name: Run CMake-Config-Build (Release configuration)
+        uses: lukka/run-cmake@v10
+        id: runcmake
+        with:
+          cmakeListsTxtPath: '${{ github.workspace }}/src/CMakeLists.txt'
+          # configurePresetAdditionalArgs: "['-DPython3_EXECUTABLE=${{ steps.setup-python.outputs.python-path }}']"
```
sylvlecl: Still needed indeed (at least for windows).

.github/workflows/runvcpkg-runcmake.yml
```diff
@@ -0,0 +1,308 @@
...
+        id: antares-deps-version
+        uses: notiz-dev/github-action-json-property@release
+        with:
+          path: 'antares-deps-version.json'
+          prop_path: 'antares_deps_version'
+
+      - name: setpreset ub
+        shell: bash 
+        run: |
+          echo "configurePreset=ninja-multi-vcpkg" >> $GITHUB_ENV
```
sylvlecl, line 92: I wonder if there is a drawback to using a multi config build ? (since we don't really need it here) maybe installing unnecessary debug versions of dependencies?

.github/workflows/runvcpkg-runcmake.yml
```diff
@@ -0,0 +1,308 @@
...
+          echo "configubuildPreset=ninja-multi-vcpkg" >> $GITHUB_ENV
```
sylvlecl, line 93: typo in the variable name (only the configure step was run in the workflow, I think that's why)

src/vcpkg.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "antares-simulator",
+  "version-string": "8.5.0",
+  "dependencies": [
+    "wxwidgets",
+    "boost-test",
+
+    {
```
sylvlecl: I guess we can remove wxwidgets and boost-test from platform specific stuff.

src/vcpkg.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "antares-simulator",
+  "version-string": "8.5.0",
+  "dependencies": [
+    "wxwidgets",
+    "boost-test",
```
sylvlecl: vcpkg allows specifying versions, either with a minimum, or with a specific version.
For best build stability, I think the best option will be to use specific versions in this file.

src/CMakeLists.txt
```diff
@@ -10,7 +10,9 @@ set(ANTARES_BETA 0)
 set(ANTARES_RC 0)
 
 set(ANTARES_VERSION_YEAR      2023)
-
+if(DEFINED ENV{VCPKG_DEFAULT_TRIPLET} AND NOT DEFINED VCPKG_TARGET_TRIPLET)
```
sylvlecl: Is it needed?

src/CMakeLists.txt
```diff
@@ -37,6 +39,8 @@ include(FetchContent)
 
 set(CMAKE_CXX_STANDARD 17)
 
+
+
 if (DEFINED VCPKG_ROOT)
     include(${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake)
```
sylvlecl: Not in that PR but I wonder if this is not redundant with the use of toolchain file.

I am not 100% sure it's possible, but overall I think it would be nice to not have explicit VCPKG stuff in our cmake file.
VCPKG is "just" the tool we will use for installing dependencies.

src/CMakePresets.json
```diff
@@ -0,0 +1,158 @@
...
+      "displayName": "MSBuild (vcpkg toolchain) Configure Settings",
+      "description": "Configure with VS generators and with vcpkg toolchain",
+      "binaryDir": "${sourceDir}/../_build",
+      "generator": "Visual Studio 17 2022",
+      "architecture": {
+        "strategy": "set",
+        "value": "x64"
+      },
+      "cacheVariables": {
+        "CMAKE_TOOLCHAIN_FILE": {
```
sylvlecl: It seems it is not useful in the build step (the property toolchainFile is only for configure step).

src/CMakePresets.json
```diff
@@ -0,0 +1,158 @@
...
+      "toolchainFile": "$env{TOOLCHAINFILE}"
+    },
+    {
+      "name": "ninja-multi-vcpkg",
+      "displayName": "Ninja Multi-Config Configure Settings",
+      "description": "Configure with vcpkg toolchain",
+      "binaryDir": "${sourceDir}/../_builds/",
+      "generator": "Ninja Multi-Config",
+      "cacheVariables": {
+        "CMAKE_TOOLCHAIN_FILE": {
```
sylvlecl: To be replaced with property toolchainFile

src/CMakePresets.json
```diff
@@ -0,0 +1,158 @@
+{
```
sylvlecl: Nice to introduce the use of this file ! 
In this first step I think we should introduce only presets that are actually used in CI, to avoid any confusion.

Further down it could be nice to introduce also commonly used default for developers, but this probably need to experiment a little.

.github/workflows/runvcpkg-runcmake.yml
```diff
@@ -0,0 +1,308 @@
...
+        shell: bash
+
+      - name: vcpkg initialization
+        shell: bash  
+        run: |
+             mkdir -p vcpkg_installed
+             mkdir -p vcpkg_cache
+
+      - name: Setup vcpkg
+        uses: lukka/run-vcpkg@main
```
sylvlecl: Before merging we'll need to select a version here

sylvlecl (commented):
Mostly minor comments / thoughts

.github/workflows/runvcpkg-runcmake.yml
a-zakir: you are right i'll re-check 👍🏽

.github/workflows/runvcpkg-runcmake.yml
a-zakir: will compare

src/CMakePresets.json
a-zakir: nice

src/CMakeLists.txt
a-zakir: i've inserted a lot of unnecessary lines in this pr, i'll clean

.github/workflows/runvcpkg-runcmake.yml
a-zakir: Merci

.github/workflows/runvcpkg-runcmake.yml
a-zakir: ultimate goal: 
to have in the preset file all config/build for all usage: local/Ci/platform

a-zakir: one would modify the preset file if needed

.github/workflows/runvcpkg-runcmake.yml
a-zakir: TODO

src/vcpkg.json
a-zakir: definetly

PR #1796, opened 2023-11-27
Minor code fixes, cleaning

JasonMarechal25:
- Rename `ExtractFilePath` -> `parentPath`
- Remove unused methods `HasKey` and `DumpToStdOut` from `namespace LocalPolicy`, in header antares/sys/policy.h
- Add a few const qualifiers on methods and argument parameters
- Extract the copy of file **simulation-comments.txt** into dedicated method `Application::writeComment`
- Remove unused class `NoResultWriter`


src/solver/application.cpp
```diff
@@ -407,18 +399,7 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
...
-
-            if (!pSettings.commentFile.empty())
-            {
-                resultWriter->addEntryFromFile(study.buffer.c_str(), pSettings.commentFile.c_str());
-
-                pSettings.commentFile.clear();
-                pSettings.commentFile.shrink();
-            }
-        }
+        WriteComment(study);
```
flomnes: `writeComment`

src/tools/batchrun/main.cpp
```diff
@@ -230,7 +230,7 @@ int main(int argc, char* argv[])
 
         // The folder that contains the solver
         String dirname;
-        IO::ExtractFilePath(dirname, solver);
+        IO::parent_path(dirname, solver);
```
flomnes:
```suggestion
        IO::ParentPath(dirname, solver);
```

flomnes (commented):
Add unit tests on `IO::parent_path` ?

src/tools/batchrun/main.cpp
JasonMarechal25: I used snake case to match std https://en.cppreference.com/w/cpp/filesystem/path/parent_path

src/ext/yuni/src/yuni/io/filename-manipulation.cpp
```diff
@@ -294,18 +294,18 @@ static inline void ExtractFilePathImpl(StringT& out, const AnyString& path, bool
...
-    ExtractFilePathImpl(out, path, systemDependant);
+    parent_path_impl(out, path, systemDependant);
 }
 
-void ExtractFilePath(Clob& out, const AnyString& path, bool systemDependant)
+void parent_path_impl(Clob& out, const AnyString& path, bool systemDependant)
 {
-    ExtractFilePathImpl(out, path, systemDependant);
+    parent_path_impl(out, path, systemDependant);
 }
```
flomnes: I'm not too sure about this. Isn't there a risk of confusion ? The templated function has the same name & number of arguments. We rely on delicate template deduction rules by doing this.

src/tools/batchrun/main.cpp
flomnes: I think we want to use CamelCase in Antares Simulator.

src/ext/yuni/src/yuni/io/filename-manipulation.cpp
flomnes: I chose to remove this version, since template deduction will do the job. This function must be here for historical reasons, since CTAD only seems to exist since C++17.

flomnes (approved):

PR #1797, opened 2023-11-27
Remove Yuni::Atomic

### LINKED ISSUE #1782: Remove uses of Yuni::Atomic
Use `std::atomic` instead

payetvin:
close #1782


flomnes (commented):
I think that most `std::atomic<int32_t>` here could be replaced by `std::atomic<int>`, what do you think ?

JasonMarechal25 (approved):

PR #1798, opened 2023-11-28
Delete "rename" process and useless calls around

JasonMarechal25:
**Yuni::Process::Rename** does not rename the process as the name implies. What it effectively does is that it take a array of char* as parameter, clear it except the first element and change the first element to a new string.
"Rename Process" is in fact the usage intended because it is called with argv and argv[0] being the name of the process. Of course, changing argv[0] value is permitted but does not influence in any way the process name at system level.

Another thing is that the result of Yuni::Process::Rename, that is a new argv with argv[0] edited, is never used.

Rename "IO::parentPath" to "IO::parent_path" to match std signature.


a-zakir:
Why?

JasonMarechal25:
> Why?

Indeed.

I've updated the top comment for clarification

src/ext/yuni/src/yuni/core/process/rename.h
```diff
@@ -1,28 +0,0 @@
-/*
```
guilpier-code, line 1: As I can understand, **Rename** is never used in **Antares Simulator** or in **Yuni** itself, so we remove this function.
But **Yuni** is an external lib (even if its sources are now integrated to Antares), that should stay untouched (except in case of bugs, or absolutely required evolution).
**Question** : why do we bother removing code from **Yuni** ?

src/ext/yuni/src/yuni/cmake/parser/importer.cpp
```diff
@@ -60,7 +60,7 @@ int main(int argc, char* argv[])
     }
 
     String targetFolder;
-    IO::parentPath(targetFolder, targetfile);
```
guilpier-code, line 63: Why do we bother make this change in **Yuni** ?
In the PR description, you say : **Rename "IO::parentPath" to "IO::parent_path" to match std signature.**
Could you be more specific about **std signature** ?

src/solver/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ if(MSVC)
 endif(MSVC)
 
 
+
```
guilpier-code, line 9: Revert this useless change ?

guilpier-code (commented):
In this review, I mainly ask questions rather than require changes

src/ext/yuni/src/yuni/cmake/parser/importer.cpp
JasonMarechal25: I'm talking about https://en.cppreference.com/w/cpp/filesystem/path/parent_path
I would have liked to simply use std parent_path but IO and std don't have the exact same behaviour. Mainly IO::parent_path doesn't discriminate plateforms like std. With std, on Unix C: is not the same as on windows C:, in IO it is supported to manipulate "C:"-style path on unix as if on windows.

Renaming serve two purpose:

-  If I search for "parent_path" I can find IO signature and usage. They're not hidden due to a different naming scheme
- Paving the path for the future. If it is decided we can safely change IO::parent_path for std::file_system::parent_path, the diff will be simpler.

src/ext/yuni/src/yuni/core/process/rename.h
JasonMarechal25: For all intent and purpose Yuni is not external since it is directly in the source code. Removing dead code allow:

- Smaller code base
- No risk to use it since the code is not maintained and we want to remove it
- Removed from static analyser
- Removed from global search (ctrl+f)

payetvin (approved):

flomnes (approved):

PR #1799, opened 2023-11-28
Enable cpp20

JasonMarechal25:


flomnes:
Change the base branch ?

JasonMarechal25:
> Change the base branch ?

Superfluous commit will be removed, as of now, it's more a test to see what happens in CI

a-zakir:
Maybe an antares-deps compiled with std20

a-zakir:
> Maybe an antares-deps compiled with std20
It may impact Xpansion…

JasonMarechal25:
> > Maybe an antares-deps compiled with std20
> > It may impact Xpansion…

Depends if ABI is stable. Should be the case between 11 and 20

.github/workflows/centos7.yml
```diff
@@ -4,8 +4,6 @@ on:
   release:
     types: [created]
   push:
-    branches:
```
sylvlecl, line 7: To be restored

sylvlecl (commented):
Nice :) just a small fix in the centos CI is needed

src/ui/simulator/windows/hydro/management.cpp
```diff
@@ -69,11 +68,8 @@ void Management::createComponents()
 
     const wxSize ourDefaultSize(55, wxDefaultSize.GetHeight());
 
-    enum
-    {
-        right = wxRIGHT | wxALIGN_RIGHT | wxALIGN_CENTER_VERTICAL,
-        left = wxLEFT | wxALIGN_LEFT | wxALIGN_CENTER_VERTICAL,
-    };
+    auto right = (wxRIGHT | wxALIGN_RIGHT | wxALIGN_CENTER_VERTICAL);
+    auto left = (wxLEFT | wxALIGN_LEFT | wxALIGN_CENTER_VERTICAL);
```
flomnes:
```suggestion
    const auto right = (wxRIGHT | wxALIGN_RIGHT | wxALIGN_CENTER_VERTICAL);
    const auto left = (wxLEFT | wxALIGN_LEFT | wxALIGN_CENTER_VERTICAL);
```

flomnes (approved):
LGTM. What about Windows ? No need to install a newer VS compiler ?

flomnes:
Also make sure to document the fact that GCC 10+ is now required to build (C++17)

flomnes (approved):

Issue #1487, opened 2023-08-03
Thermal cluster lists : simplify and clarify


guilpier-code:
There are many cluster lists for an area : 
- **src/libs/antares/study/parts/thermal/container.h** : `ThermalClusterList list->mapping;` 
- **src/libs/antares/study/parts/thermal/container.h** :  `ThermalClusterList mustrunList;`
- **src/libs/antares/study/parts/thermal/container.h** ; `std::vector<ThermalCluster*> clusters;`
- **src/libs/antares/study/parts/common/cluster_list.h** : `Map cluster;`
- other lists ?

It's confusing, and it's pretty sure that the number of cluster lists can be reduced.
We should have a deep look into this in order to simplify and clarify

flomnes:
Note that the logic is similar for renewable clusters, as well as short-term storages. This PR also applies to those.

flomnes:
@guilpier-code

## Proposal
Remove `std::map`, keep the `std::vector`s and `std::sort` them in the loading function, after the loading has been completed.

- For thermal clusters, only keep `std::vector<ThermalCluster> clusters`, remove `ThermalClusterList list` and `ThermalClusterList mustrunList`.
- For short-term storage, only keep `std::vector<STStorageCluster> storagesByIndex`, remove `std::map<std::string, STStorageCluster> storagesById`.
- For renewable clusters, only keep `std::vector<RenewableCluster> clusters`, remove `RenewableClusterList list`.

payetvin:
For renewable clusters, keep `RenewableClusterList list`,  removed `std::vector<RenewableCluster> clusters`

PR #1804, opened 2023-11-30
Remove sts container

payetvin:
This PR brings an answer to issue #1487

### DIFF
--- a/src/libs/antares/study/parts/short-term-storage/cluster.h
+++ b/src/libs/antares/study/parts/short-term-storage/cluster.h
@@ -37,7 +37,7 @@ namespace Antares::Data::ShortTermStorage
 class STStorageCluster
 {
 public:
-    bool validate();
+    bool validate() const;
     bool loadFromSection(const IniFile::Section& section);
 
     bool loadSeries(const std::string& folder) const;
@@ -48,7 +48,7 @@ public:
     std::string id;
 
     std::shared_ptr<Series> series = std::make_shared<Series>();
-    Properties properties;
+    mutable Properties properties;
 
 };
 } // namespace Antares::Data::ShortTermStorage
--- a/src/libs/antares/study/parts/short-term-storage/container.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/container.cpp
@@ -65,9 +65,9 @@ bool STStorageInput::createSTStorageClustersFromIniFile(const std::string& path)
         storagesByIndex.push_back(cluster);
     }
 
-    /* std::ranges::sort(storagesByIndex, [](const MyClass &a, const MyClass &b) { */
-    /*     return a.properties.name < b.properties.name; */
-    /* }); */
+    std::sort(storagesByIndex.begin(), storagesByIndex.end(), [&](const auto& a, const auto& b){
+        return a.properties.name < b.properties.name;
+    });
 
     return true;
 }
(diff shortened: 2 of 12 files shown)


src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -61,12 +62,12 @@ bool STStorageInput::createSTStorageClustersFromIniFile(const std::string& path)
...
-        storagesById.try_emplace(cluster.properties.name, cluster);
+        storagesByIndex.push_back(cluster);
     }
 
-    storagesByIndex.reserve(storagesById.size());
-    for (auto& [id, storage] : storagesById)
-        storagesByIndex.push_back(&storage);
+    /* std::ranges::sort(storagesByIndex, [](const MyClass &a, const MyClass &b) { */
+    /*     return a.properties.name < b.properties.name; */
+    /* }); */
```
flomnes: Uncomment ?

src/libs/antares/study/parts/short-term-storage/cluster.h
```diff
@@ -48,7 +48,7 @@ class STStorageCluster
     std::string id;
 
     std::shared_ptr<Series> series = std::make_shared<Series>();
-    Properties properties;
+    mutable Properties properties;
```
guilpier-code, line 51: Not a fan of adding a **mutable** keyword here.
I guess **mutable** was put here because **Properties::validate()** both is now **const** and though changes some data member values : for some data members, we tolerate wrong values among the input data, but the code corrects these wrong values in order to make to simulation pass.
I talked to @flomnes about that, and we don't want to do that anymore :  
We should forbid these wrong values, and raise an error when we bump into such values.

src/libs/antares/study/parts/short-term-storage/container.h
```diff
@@ -46,8 +46,6 @@ class STStorageInput
     bool saveToFolder(const std::string& folder) const;
     bool saveDataSeriesToFolder(const std::string& folder) const;
 
-
-    std::vector<STStorageCluster*> storagesByIndex;
-    std::map<std::string, STStorageCluster> storagesById;
+    std::vector<STStorageCluster> storagesByIndex;
```
guilpier-code, line 49: Not sure **ByIndex** in **storagesByIndex** is necessary. 
We could call it merely **clusters**, like it's done everywhere else for thermal and renewable

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/cluster.h
payetvin: I will do this work in another pull request, there's some refactoring/fixing to do in sts
#1808

src/libs/antares/study/parts/short-term-storage/container.h
payetvin: Same, will be done in a separate PR

src/libs/antares/study/parts/short-term-storage/cluster.h
guilpier-code: No problem for me.
But in this new issue's presentation (#1808), there is no reference to remarks of the current PR that remain to be taken care of.
So I will add an TODO item in this issue, and approuve the current one.

guilpier-code: Will be taken care of in #1808

src/libs/antares/study/parts/short-term-storage/container.h
guilpier-code: Will be taken care of in #1808

guilpier-code (approved):

PR #1806, opened 2023-12-01
Remove Study::gotFatalError, throw exceptions instead

flomnes:
Remove `Study::gotFatalError`, use exceptions instead.

Side effect : since `throw` is used, it breaks the execution flow. As a consequence, only the 1st error will be displayed, not all errors.

Errors displayed with `Antares::logs` are still non-breaking.


src/libs/antares/exception/LoadingError.cpp
```diff
@@ -99,11 +99,6 @@ IncompatibleDailyOptHeuristicForArea::IncompatibleDailyOptHeuristicForArea(
 {
 }
 
-WritingProgressFile::WritingProgressFile(const Yuni::String& file) :
- LoadingError(std::string("I/O error: impossible to write ") + file.c_str())
-{
-}
-
```
flomnes, line 106: unused

src/solver/application.cpp
```diff
@@ -375,7 +372,7 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
     }
 
     // Errors
-    if (pErrorCount || pWarningCount || study.gotFatalError)
+    if (pErrorCount || pWarningCount)
```
flomnes: In some cases, this will never be run (see l. 352). This is a great drawback IMO

src/libs/antares/study/study.h
```diff
@@ -664,11 +664,6 @@ class Study: public Yuni::NonCopyable<Study>, public IObject, public LayerData
     void* cacheTSGenerator[timeSeriesCount];
     //@}
 
-    /*!
-    ** \brief
-    */
-    bool gotFatalError = false;
```
sylvlecl, line 670: 100% agreed it had nothing to do here: it should have been the result of a loading function, not a class member

src/libs/antares/study/area/links.cpp
```diff
@@ -416,22 +417,21 @@ bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const
     return false;
 }
 
-void logLinkDataCheckError(bool& gotFatalError, const AreaLink& link, const String& msg, int hour)
+void logLinkDataCheckError(const AreaLink& link, const String& msg, int hour)
 {
     logs.error() << "Link (" << link.from->name << "/" << link.with->name << "): Invalid values ("
                  << msg << ") for hour " << hour;
-    gotFatalError = true;
+    throw Antares::Error::ReadingStudy();
```
sylvlecl, line 422: The minor drawback of this approach is that we will not list anymore all errors, only the first one.

However, it seems reasonable to me!

sylvlecl (approved):
See comment about minor behaviour change (stop on first error).

Seems reasonable to me, but I let you confirm this is intended.

src/libs/antares/study/area/links.cpp
flomnes: It is not acceptable from a user's point of view, since they may want to fix as many errors as possible before launching a new simulation, rather than fixing errors one by one.

flomnes: We need to group all these checks in the same part of the code probably

PR #1807, opened 2023-12-04
Add `enabled` property for ST storage objects, fix bug related to saving ST objects

flomnes:
- Add bool property `enabled` for each ST storage object
- Whenever a ST storage is not `enabled`, remove it from computations relatively early (right after loading the data), using a similar mechanism as thermal & renewable clusters
- Also fix list.ini files containing only the last ST storage object because of overwrite, add unit test case
- Cherry-pick of "Remove sts container (#1804)"


src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -53,6 +53,9 @@ static void importShortTermStorages(
         int storageIndex = 0;
         for (auto st : areas[areaIndex]->shortTermStorage.storagesByIndex)
         {
+            if (!st->properties.enabled)
```
flomnes: @payetvin 
```suggestion
            if (!st->enabled())
```

flomnes (commented):
Instead of 
```c
for (...)
    if (!cluster.enabled()) continue;
```
we could expose an iterator to iterate over enabled clusters only. WDYT ?

flomnes:
Check indexes for `pValuesForTheCurrentYear`, possibly elsewhere.

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -76,9 +84,9 @@ bool STStorageCluster::loadSeries(const std::string& folder) const
     return ret;
 }
 
-bool STStorageCluster::saveProperties(const std::string& path) const
+void STStorageCluster::saveProperties(IniFile& ini) const
 {
-    return properties.saveToFolder(path);
+    properties.saveToFolder(ini);
```
flomnes:
```suggestion
    properties.save(ini);
```

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -91,24 +92,44 @@ bool STStorageInput::saveToFolder(const std::string& folder) const
...
+                       storagesByIndex.end(),
+                       [](const STStorageCluster& st) {
+                           return st.properties.enabled;
+                       });
 }
+
+uint STStorageInput::removeDisabledClusters()
+{
+    const auto& it = std::remove_if(storagesByIndex.begin(), storagesByIndex.end(),
+        [](const auto& c) { return c.enabled(); });
```
flomnes:
```suggestion
    const auto& it = std::remove_if(storagesByIndex.begin(), storagesByIndex.end(),
        std::mem_fn(&STStorageCluster::enabled));
```
C++20 makes this so much simpler...

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input.cpp
flomnes, line 1: Rename short-term-storage-input-output.cpp

flomnes (commented):
Add a unit test on the save function

PR #1813, opened 2023-12-08
Refactor common ClusterList class

payetvin:
This PR handles a part of issue #1487 (simplify and clarify cluster lists) : 

- [x] Removing **common** cluster list **byIndex**
- [x] Changing **common** cluster list **cluster** from **std::map** into **std::vector**
- [x] Handling consequences of previous items

### DIFF
--- a/src/libs/antares/InfoCollection/StudyInfoCollector.cpp
+++ b/src/libs/antares/InfoCollection/StudyInfoCollector.cpp
@@ -59,13 +59,9 @@ void StudyInfoCollector::enabledThermalClustersCountToFileContent(FileContent& f
     for (auto i = study_.areas.begin(); i != end; ++i)
     {
         Area& area = *(i->second);
-        auto end = area.thermal.list.end();
-        for (auto i = area.thermal.list.begin(); i != end; ++i)
-        {
-            auto& cluster = i->second;
+        for (const auto& cluster : area.thermal.list)
             if (cluster->enabled)
                 nbEnabledThermalClusters++;
-        }
     }
 
     // Adding an item related to number of enabled thermal clusters to the file content
--- a/src/libs/antares/InfoCollection/StudyInfoCollector.cpp
+++ b/src/libs/antares/InfoCollection/StudyInfoCollector.cpp
@@ -59,9 +59,8 @@ void StudyInfoCollector::enabledThermalClustersCountToFileContent(FileContent& f
     for (auto i = study_.areas.begin(); i != end; ++i)
     {
         Area& area = *(i->second);
-        for (const auto& cluster : area.thermal.list)
-            if (cluster->enabled)
-                nbEnabledThermalClusters++;
+        nbEnabledThermalClusters +=
+            std::ranges::count_if(area.thermal.list, [](const auto& c) { return c->enabled; });
     }
 
     // Adding an item related to number of enabled thermal clusters to the file content
--- a/src/libs/antares/study/cleaner/cleaner-v20.cpp
+++ b/src/libs/antares/study/cleaner/cleaner-v20.cpp
@@ -135,23 +135,19 @@ static void listOfFilesAnDirectoriesToKeepForArea(PathList& e, PathList& p, cons
         buffer.clear() << "input/thermal/clusters/" << id << "/list.ini";
         e.add(buffer);
 
-        auto end = area->thermal.list.end();
-        for (auto i = area->thermal.list.begin(); i != end; ++i)
+        for (const auto& cluster : area->thermal.list)
         {
-            // Reference to the thermal cluster
-            auto& cluster = *(i->second);
-
-            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster.id();
+            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster->id();
             p.add(buffer);
-            buffer.clear() << "input/thermal/series/" << id << '/' << cluster.id();
+            buffer.clear() << "input/thermal/series/" << id << '/' << cluster->id();
             p.add(buffer);
 
-            buffer.clear() << "input/thermal/series/" << id << '/' << cluster.id() << "/series.txt";
+            buffer.clear() << "input/thermal/series/" << id << '/' << cluster->id() << "/series.txt";
             e.add(buffer);
 
-            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster.id() << "/data.txt";
+            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster->id() << "/data.txt";
             e.add(buffer);
-            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster.id()
+            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster->id()
                            << "/modulation.txt";
             e.add(buffer);
         }
@@ -168,16 +164,12 @@ static void listOfFilesAnDirectoriesToKeepForArea(PathList& e, PathList& p, cons
         buffer.clear() << "input/renewables/clusters/" << id << "/list.ini";
         e.add(buffer);
 
-        auto end = area->renewable.list.end();
-        for (auto i = area->renewable.list.begin(); i != end; ++i)
+        for (const auto& cluster : area->renewable.list)
         {
-            // Reference to the thermal cluster
-            auto& cluster = *(i->second);
-
-            buffer.clear() << "input/renewables/series/" << id << '/' << cluster.id();
+            buffer.clear() << "input/renewables/series/" << id << '/' << cluster->id();
             p.add(buffer);
 
-            buffer.clear() << "input/renewables/series/" << id << '/' << cluster.id()
+            buffer.clear() << "input/renewables/series/" << id << '/' << cluster->id()
                            << "/series.txt";
             e.add(buffer);
         }
--- a/src/libs/antares/study/parts/common/cluster.cpp
+++ b/src/libs/antares/study/parts/common/cluster.cpp
@@ -49,45 +49,43 @@ void Cluster::setName(const AnyString& newname)
 }
 
 #define SEP Yuni::IO::Separator
-int Cluster::saveDataSeriesToFolder(const AnyString& folder) const
+bool Cluster::saveDataSeriesToFolder(const AnyString& folder) const
 {
-    if (not folder.empty())
+    if (folder.empty())
+        return true;
+
+    Yuni::Clob buffer;
+
+    buffer.clear() << folder << SEP << parentArea->id << SEP << id();
+    if (Yuni::IO::Directory::Create(buffer))
     {
-        Yuni::Clob buffer;
-
-        buffer.clear() << folder << SEP << parentArea->id << SEP << id();
-        if (Yuni::IO::Directory::Create(buffer))
-        {
-            int ret = 1;
-            buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series.txt";
-            ret = series.timeSeries.saveToCSVFile(buffer, precision()) && ret;
-
-            return ret;
-        }
-        return 0;
-    }
-    return 1;
-}
-
-int Cluster::loadDataSeriesFromFolder(Study& s, const AnyString& folder)
-{
-    if (not folder.empty())
-    {
-        auto& buffer = s.bufferLoadingTS;
-
-        int ret = 1;
-        buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series."
-                       << s.inputExtension;
-        ret = series.timeSeries.loadFromCSVFile(buffer, 1, HOURS_PER_YEAR, &s.dataBuffer) && ret;
-
-        if (s.usedByTheSolver && s.parameters.derated)
-            series.timeSeries.averageTimeseries();
-
-        series.timeseriesNumbers.clear();
+        bool ret = true;
+        buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series.txt";
+        ret = series.timeSeries.saveToCSVFile(buffer, precision()) && ret;
 
         return ret;
     }
-    return 1;
+    return false;
+}
+
+bool Cluster::loadDataSeriesFromFolder(Study& s, const AnyString& folder)
+{
+    if (folder.empty())
+        return true;
+
+    auto& buffer = s.bufferLoadingTS;
+
+    bool ret = 1;
+    buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series."
+        << s.inputExtension;
+    ret = series.timeSeries.loadFromCSVFile(buffer, 1, HOURS_PER_YEAR, &s.dataBuffer) && ret;
+
+    if (s.usedByTheSolver && s.parameters.derated)
+        series.timeSeries.averageTimeseries();
+
+    series.timeseriesNumbers.clear();
+
+    return ret;
 }
 #undef SEP
 
--- a/src/libs/antares/study/parts/common/cluster.cpp
+++ b/src/libs/antares/study/parts/common/cluster.cpp
@@ -55,17 +55,12 @@ bool Cluster::saveDataSeriesToFolder(const AnyString& folder) const
         return true;
 
     Yuni::Clob buffer;
-
     buffer.clear() << folder << SEP << parentArea->id << SEP << id();
-    if (Yuni::IO::Directory::Create(buffer))
-    {
-        bool ret = true;
-        buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series.txt";
-        ret = series.timeSeries.saveToCSVFile(buffer, precision()) && ret;
+    if (!Yuni::IO::Directory::Create(buffer))
+        return true;
 
-        return ret;
-    }
-    return true;
+    buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series.txt";
+    return series.timeSeries.saveToCSVFile(buffer, precision());
 }
 
 bool Cluster::loadDataSeriesFromFolder(Study& s, const AnyString& folder)
--- a/src/libs/antares/study/parts/common/cluster.cpp
+++ b/src/libs/antares/study/parts/common/cluster.cpp
@@ -65,7 +65,7 @@ bool Cluster::saveDataSeriesToFolder(const AnyString& folder) const
 
         return ret;
     }
-    return false;
+    return true;
 }
 
 bool Cluster::loadDataSeriesFromFolder(Study& s, const AnyString& folder)
@@ -75,7 +75,7 @@ bool Cluster::loadDataSeriesFromFolder(Study& s, const AnyString& folder)
 
     auto& buffer = s.bufferLoadingTS;
 
-    bool ret = 1;
+    bool ret = true;
     buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series."
         << s.inputExtension;
     ret = series.timeSeries.loadFromCSVFile(buffer, 1, HOURS_PER_YEAR, &s.dataBuffer) && ret;
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -100,11 +100,7 @@ void ClusterList<ClusterT>::clear()
 template<class ClusterT>
 void ClusterList<ClusterT>::resizeAllTimeseriesNumbers(uint n)
 {
-    assert(n < 200000); // arbitrary number
-    if (n == 0)
-        each([&](Cluster& cluster) { cluster.series.timeseriesNumbers.clear(); });
-    else
-        each([&](Cluster& cluster) { cluster.series.timeseriesNumbers.resize(1, n); });
+    each([&](Cluster& cluster) { cluster.series.timeseriesNumbers.resize(1, n); });
 }
 
 #define SEP IO::Separator
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -256,7 +256,7 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
 template<class ClusterT>
 bool ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder) const
 {
-    return std::all_of(clusters.begin(), clusters.end(), [&folder](const auto& c){
+    return std::ranges::all_of(clusters, [&folder](const auto& c){
         return c->saveDataSeriesToFolder(folder);
     });
 }
@@ -264,17 +264,15 @@ bool ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder) cons
 template<class ClusterT>
 bool ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder, const String& msg) const
 {
-    bool ret = true;
     uint ticks = 0;
 
-    for (const auto& c : clusters)
+    return std::ranges::all_of(clusters, [&](const auto& c)
     {
         logs.info() << msg << "  " << (ticks * 100 / (1 + this->clusters.size()))
             << "% complete";
-        ret = c->saveDataSeriesToFolder(folder) && ret;
         ++ticks;
-    }
-    return ret;
+        return c->saveDataSeriesToFolder(folder);
+    });
 }
 
 template<class ClusterT>
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -57,19 +57,9 @@ typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::end() cons
 }
 
 template<class ClusterT>
-const ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
+ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
 {
     for (const auto& c : cluster)
-        if (c->id() == id)
-            return c.get();
-
-    return nullptr;
-}
-
-template<class ClusterT>
-ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id)
-{
-    for (auto& c : cluster)
         if (c->id() == id)
             return c.get();
 
@@ -122,18 +112,6 @@ const ClusterT* ClusterList<ClusterT>::find(const ClusterT* p) const
     for (const auto& c : cluster)
         if (c.get() == p)
             return c.get();
-
-    return nullptr;
-}
-
-
-template<class ClusterT>
-typename ClusterList<ClusterT>::SharedPtr ClusterList<ClusterT>::find
-  (const ClusterList<ClusterT>::SharedPtr& p)
-{
-    for (auto& c : cluster)
-        if (c == p)
-            return c;
 
     return nullptr;
 }
@@ -192,13 +170,13 @@ void ClusterList<ClusterT>::rebuildIndex()
 
 template<class ClusterT>
 typename ClusterList<ClusterT>::SharedPtr ClusterList<ClusterT>::add(
-  const ClusterList<ClusterT>::SharedPtr& newcluster)
+    const ClusterList<ClusterT>::SharedPtr newcluster)
 {
     if (!newcluster)
         return nullptr;
 
     if (exists(newcluster->id()))
-        return this->find(newcluster);
+        return newcluster;
 
     cluster.push_back(newcluster);
     ++(groupCount[newcluster->groupId()]);
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -204,18 +204,16 @@ template<class ClusterT>
 typename ClusterList<ClusterT>::SharedPtr ClusterList<ClusterT>::add(
   const ClusterList<ClusterT>::SharedPtr& newcluster)
 {
-    if (newcluster)
-    {
-        if (exists(newcluster->id()))
-            return this->find(newcluster);
+    if (!newcluster)
+        return nullptr;
 
-        newcluster->index = (uint)size();
-        cluster.push_back(newcluster);
-        ++(groupCount[newcluster->groupId()]);
-        rebuildIndex();
-        return cluster.back();
-    }
-    return nullptr;
+    if (exists(newcluster->id()))
+        return this->find(newcluster);
+
+    cluster.push_back(newcluster);
+    ++(groupCount[newcluster->groupId()]);
+    rebuildIndex();
+    return newcluster;
 }
 
 template<class ClusterT>
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -136,7 +136,7 @@ void ClusterList<ClusterT>::storeTimeseriesNumbers(Solver::IResultWriter& writer
     std::string ts_content;
 
     each([&](const Cluster& cluster) {
-        path.clear() << "ts-numbers" << SEP << typeID() << SEP << cluster.parentArea->id << SEP
+        path.clear() << "ts-numbers" << SEP << typeID << SEP << cluster.parentArea->id << SEP
                      << cluster.id() << ".txt";
         ts_content.clear(); // We must clear ts_content here, since saveToBuffer does not do it.
         cluster.series.timeseriesNumbers.saveToBuffer(ts_content, 0, true, predicate, true);
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -115,8 +115,7 @@ template<class ClusterT>
 void ClusterList<ClusterT>::clear()
 {
     byIndex.clear();
-    if (not cluster.empty())
-        cluster.clear();
+    cluster.clear();
 }
 
 template<class ClusterT>
@@ -192,7 +191,9 @@ template<class ClusterT>
 void ClusterList<ClusterT>::rebuildIndex()
 {
     byIndex.clear();
-
+    std::sort(cluster.begin(), cluster.end(), [&](const auto& a, const auto& b){
+        return a->id() < b->id();
+    });
     if (not empty())
     {
         uint indx = 0;
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -114,7 +114,6 @@ Data::ClusterList<ClusterT>::~ClusterList()
 template<class ClusterT>
 void ClusterList<ClusterT>::clear()
 {
-    byIndex.clear();
     cluster.clear();
 }
 
@@ -190,20 +189,13 @@ void ClusterList<ClusterT>::storeTimeseriesNumbers(Solver::IResultWriter& writer
 template<class ClusterT>
 void ClusterList<ClusterT>::rebuildIndex()
 {
-    byIndex.clear();
     std::sort(cluster.begin(), cluster.end(), [&](const auto& a, const auto& b){
         return a->id() < b->id();
     });
-    if (not empty())
-    {
-        uint indx = 0;
-        byIndex.resize(size());
-        for (auto& c : cluster)
-        {
-            byIndex[indx] = c.get();
-            c->index = indx++;
-        }
-    }
+
+    uint indx = 0;
+    for (auto& c : cluster)
+        c->index = indx++;
 }
 
 
--- a/src/libs/antares/study/parts/common/cluster_list.cpp
+++ b/src/libs/antares/study/parts/common/cluster_list.cpp
@@ -24,43 +24,43 @@ using namespace Antares;
 template<class ClusterT>
 inline uint ClusterList<ClusterT>::size() const
 {
-    return (uint)cluster.size();
+    return (uint)clusters.size();
 }
 
 template<class ClusterT>
 inline bool ClusterList<ClusterT>::empty() const
 {
-    return cluster.empty();
+    return clusters.empty();
 }
 
 template<class ClusterT>
 typename ClusterList<ClusterT>::iterator ClusterList<ClusterT>::begin()
 {
-    return std::begin(cluster);
+    return std::begin(clusters);
 }
 
 template<class ClusterT>
 typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::begin() const
 {
-    return std::begin(cluster);
+    return std::begin(clusters);
 }
 
 template<class ClusterT>
 typename ClusterList<ClusterT>::iterator ClusterList<ClusterT>::end()
 {
-    return std::end(cluster);
+    return std::end(clusters);
 }
 
 template<class ClusterT>
 typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::end() const
 {
-    return std::end(cluster);
+    return std::end(clusters);
 }
 
 template<class ClusterT>
 ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
 {
-    for (const auto& c : cluster)
+    for (const auto& c : clusters)
         if (c->id() == id)
             return c.get();
 
@@ -70,7 +70,7 @@ ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
 template<class ClusterT>
 const ClusterT* ClusterList<ClusterT>::find(const ClusterT* p) const
 {
-    for (const auto& c : cluster)
+    for (const auto& c : clusters)
         if (c.get() == p)
             return c.get();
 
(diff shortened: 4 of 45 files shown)


src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -25,10 +26,12 @@ class ClusterList
     using SharedPtr = typename std::shared_ptr<ClusterT>;
     // Map container
     using Map = typename std::map<ClusterName, SharedPtr>;
```
guilpier-code: Is `using Map = typename std::map<ClusterName, SharedPtr>;` still required ?
Apparently still required in child class (**ThermalClusterList**)
So, this definition should be moved to class **ThermalClusterList**.

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -219,10 +215,8 @@ class ClusterList
     uint64_t memoryUsage() const;
 
 public:
-    //! All clusters by their index
-    std::vector<ClusterT*> byIndex;
```
guilpier-code, line 223: Here, we remove the **byIndex** and turn  **cluster** from a **std::map** into a **std::vector**.
It reminds me that we have the same situation in class **AreaList**.
Shouldn't we do the same thing for that class ?

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -58,21 +58,27 @@ typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::end() cons
 template<class ClusterT>
 const ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
```
guilpier-code: There are 2 **find(id)** functions, they both do the same thing. One of them is redundant, probably the non **const**. 
We should take advantage of the current PR try to remove it.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -109,32 +112,37 @@ Data::ClusterList<ClusterT>::~ClusterList()
...
 {
-    byIndex.clear();
-    if (not cluster.empty())
-        cluster.clear();
+    cluster.clear();
 }
 
 template<class ClusterT>
 const ClusterT* ClusterList<ClusterT>::find(const ClusterT* p) const
 {
```
guilpier-code: Same remark as above : could we remove one of the 2 find functions ?

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -109,32 +112,37 @@ Data::ClusterList<ClusterT>::~ClusterList()
...
-    }
+    for (auto& c : cluster)
+        if (c.get() == p)
+            return c.get();
+
+    return nullptr;
+}
+
+template<class ClusterT>
+typename ClusterList<ClusterT>::SharedPtr ClusterList<ClusterT>::find
```
guilpier-code: Do we need this `find(SharedPtr&)` function ?
Or can we replace its usages by a usage to `find(const ClusterT* p)` ?

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -179,38 +187,31 @@ void ClusterList<ClusterT>::storeTimeseriesNumbers(Solver::IResultWriter& writer
 template<class ClusterT>
 void ClusterList<ClusterT>::rebuildIndex()
 {
-    byIndex.clear();
+    std::sort(cluster.begin(), cluster.end(), [&](const auto& a, const auto& b){
```
guilpier-code: We don't need any reference to outside variable here, so we should replace **[&]** with **[]**

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -242,12 +243,10 @@ bool ClusterList<ClusterT>::rename(Data::ClusterName idToFind, Data::ClusterName
     Antares::TransformNameIntoID(newName, newID);
 
     // Looking for the renewable cluster in the list
-    auto it = cluster.find(idToFind);
-    if (it == cluster.end())
+    auto* p = this->find(idToFind);
```
guilpier-code: **p** is not meaningful.
**cluster** is already taken by a vector data member. 
We should rename it into **cluster_ptr** (for instance)

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -282,32 +278,32 @@ template<class ClusterT>
...
+    for (const auto& c : cluster)
+        c->markAsModified();
 }
 
 template<class ClusterT>
 bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
 {
-    auto i = cluster.find(id);
-    if (i == cluster.end())
+    const auto& it = find_if(cluster.begin(), cluster.end(),
```
guilpier-code: Here we use/call another find(...) (**std::find_if**) function.
I guess we use this one here because it gives iterator on the found cluster, that we can use when calling the std::erase function few lines further.
We should try to have only 2 **find** functions, one taking an **id**, the other a pointer to a cluster, and code the body of these function with **std::find** functions.
In that case, the caller will have to change a bit.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -320,35 +316,30 @@ template<class ClusterT>
 int ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder) const
 {
     if (empty())
-        return 1;
+        return false;
```
guilpier-code: If we return **false** (which is better that returning **1**, as we once were), then let's return a **bool**

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -320,35 +316,30 @@ template<class ClusterT>
...
 
-    int ret = 1;
+    bool ret = true;
```
guilpier-code: Why not use **std::accumulate** ? We would have to make **saveDataSeriesToFolder** return a **bool** first (it should return a **bool** rather than an **int** anyway).
```C++
bool ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder) const
{
    return std::accumulate(cluster.begin(), cluster.end(), true, [&folder](bool ret, const auto& c){
        return c->saveDataSeriesToFolder(folder) && ret;
    });
}
```

guilpier-code (changes requested):

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -282,32 +278,32 @@ template<class ClusterT>
 bool ClusterList<ClusterT>::forceReload(bool reload) const
```
guilpier-code, line 192: Use **std::accumulate** ?
```C++
bool ClusterList<ClusterT>::forceReload(bool reload) const
{
    return std::accumulate(cluster.begin(), cluster.end(), true, [&reload](bool ret, const auto& c){
        return c->forceReload(folder) && ret;
    });
}
```

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: Here we change **rename(...)** function for a cluster.
This function is called from the **GUI** only.
We should make a manual test at least (renaming a cluster), to see if we didn't break anything, have we ?
A unit test may be too much (= overkill), because we don't want to add code regarding **GUI**, right ?

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -282,32 +278,32 @@ template<class ClusterT>
...
 {
-    auto end = cluster.end();
-    for (auto i = cluster.begin(); i != end; ++i)
-        (i->second)->markAsModified();
+    for (const auto& c : cluster)
+        c->markAsModified();
 }
 
 template<class ClusterT>
 bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
```
guilpier-code, line 208: Same comment as for **rename(...)** : a manual test should be done.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -320,35 +316,30 @@ template<class ClusterT>
...
-    for (auto it = cluster.begin(); it != end; ++it)
-    {
-        auto& cluster = *(it->second);
-        ret = cluster.saveDataSeriesToFolder(folder) and ret;
-    }
     return ret;
 }
 
 template<class ClusterT>
 int ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder, const String& msg) const
```
guilpier-code: We should return a **bool** instead of an **int** : that's the real intention of this function.

guilpier-code: It should work with **std::accumulate**

guilpier-code (changes requested):

src/libs/antares/study/parts/renewable/container.cpp
```diff
@@ -66,11 +66,9 @@ void PartRenewable::prepareAreaWideIndexes()
         return;
     }
 
-    auto end = list.end();
     uint idx = 0;
-    for (auto i = list.begin(); i != end; ++i)
+    for (const auto& t : list)
```
guilpier-code: what is **t** ? 
We should rename t into **cluster**

src/libs/antares/study/parts/renewable/container.cpp
```diff
@@ -86,8 +84,8 @@ uint PartRenewable::removeDisabledClusters()
```
guilpier-code: About **removeDisabledClusters()** : 
we should try to use **std::remove(...)**, in combination with **std::erase()** : 
- **std::remove(...)** used with an lambda function for criteria to select the element to be removed. This algorithm **does not** remove elements from the vector, but rejects them at the vector's tail (or end), and returns a pointer to the first element to be removed.
- **std::erase()** **actually removes** selected element to be removed.

At the end, we would have : 
```C++
uint PartRenewable::removeDisabledClusters()
{
    auto firstClusterToRemove = std::remove(list.begin(), list.end(), [] (auto& cluster) {
        return not cluster->enabled});

    list.erase(firstClusterToRemove , list.end()); // Actually remove the disabled clusters

    if (list.size())
        list.rebuildIndex();

    return list.size();
}
```
This function is safer and shorter (at least 10 lines smaller)

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -485,7 +458,7 @@ bool ThermalClusterList::loadPreproFromFolder(Study& study,
     Clob buffer;
```
guilpier-code, line 449: About **ThermalClusterList::loadPreproFromFolder** (not mandatory) : 
in its body, using an algorithm like **std::accumulate** would allow to remove lines like : 
```C++
...
if (empty())
        return true;
...
assert(c->parentArea and "cluster: invalid parent area"); // Really useful here ?
...
ret = result and ret;
...
return ret;
```

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -518,7 +491,7 @@ bool ThermalClusterList::loadEconomicCosts(Study& study, const AnyString& folder
...
```
guilpier-code: Same as above

src/libs/antares/study/parts/thermal/container.cpp
```diff
@@ -148,8 +144,8 @@ uint PartThermal::removeDisabledClusters()
 
     for (auto& it : list)
```
guilpier-code: About **removeDisabledClusters** : 
Exactly the same code as in **renewable/container.cpp**.
So , we should : 
- Find a way to have one function for both thermal and renewable clusters
- apply remarks about this function in **renewable/container.cpp**.

src/libs/antares/study/parts/thermal/container.cpp
```diff
@@ -99,10 +96,9 @@ uint PartThermal::prepareClustersInMustRunMode()
     do
```
guilpier-code, line 96: This loop do ... while is complicated. Why is it there ? Is it necessary or not ?
We'll to think about it some day.

src/libs/antares/study/parts/thermal/container.cpp
```diff
@@ -99,10 +96,9 @@ uint PartThermal::prepareClustersInMustRunMode()
...
     {
         mustContinue = false;
-        auto end = list.end();
-        for (auto i = list.begin(); i != end; ++i)
+        for (auto i = list.begin(); i != list.end(); ++i)
```
guilpier-code: We have to keep the range loop with an iterator i because of the : 
```C++
std::shared_ptr<ThermalCluster> cluster = list.detach(i);
```
We should try (some day) to get rid of this iterator. For later.

src/ui/action/handler/antares-study/area/create.cpp
```diff
@@ -296,7 +296,7 @@ void Create::createActionsForAStandardAreaCopy(Context& ctx, bool copyPosition)
         auto end = area->thermal.list.end();
```
guilpier-code: Change into a `for(auto& cluster : area->thermal.list)`, quickly done, and more clear

src/ui/simulator/windows/inspector/frame.cpp
```diff
@@ -180,7 +180,7 @@ void Frame::onSelectAllPlants(wxCommandEvent&)
             Data::Area& area = *(*i);
             auto end = area.thermal.list.end();
```
guilpier-code: Change into a `for(auto& cluster : area->thermal.list)`, quickly done, and more clear

src/ui/simulator/windows/inspector/frame.cpp
```diff
@@ -216,7 +216,7 @@ void Frame::onSelectAllPlantsFromArea(wxCommandEvent& evt)
         data->ThClusters.clear();
```
guilpier-code: **onSelectAllPlantsFromArea** : apparently not used (see comment before function), let's remove it

guilpier-code (changes requested):

src/libs/antares/study/parts/common/cluster_list.cpp
payetvin: Both function are used, as we sometimes need to get pass a const object or a non const one

src/libs/antares/study/parts/common/cluster_list.cpp
payetvin: removed the non const function

src/libs/antares/study/parts/common/cluster_list.cpp
payetvin: return type is different and is needed in the add function

src/libs/antares/study/parts/common/cluster_list.cpp
payetvin: I've tested renaming, adding and deleting clusters in the GUI

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -58,21 +59,27 @@ typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::end() cons
 template<class ClusterT>
 const ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
 {
-    auto i = cluster.find(id);
-    return (i != cluster.end()) ? i->second.get() : nullptr;
+    for (const auto& c : cluster)
```
sylvlecl: 1. we could use  `std::ranges::find_if`
2. But this kind of linear search is VERY inefficient: we'd better use a map here.
An alternative, if we are sure the vector is sorted (I think it's the case), is to use a search algo such as `std::range::lower_bound`

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -109,32 +113,28 @@ Data::ClusterList<ClusterT>::~ClusterList()
...
+        if (c.get() == p)
+            return c.get();
+
     return nullptr;
 }
 
+
 template<class ClusterT>
-ClusterT* ClusterList<ClusterT>::find(const ClusterT* p)
+typename ClusterList<ClusterT>::SharedPtr ClusterList<ClusterT>::find
```
sylvlecl: Not the topic of this PR, but the type name `ClusterList<ClusterT>::SharedPtr` here is very misleading: it looks like a shared_ptr of the list, whereas it's actually a shared_ptr of values.
Just using `std::shared_ptr<ClusterT>` is a lot clearer.

src/libs/antares/study/parts/common/cluster_list.cpp
sylvlecl: Those 2 functions (find a pointer) seems a little absurd:
basically it does not search anything, it just checks that the pointer is already in the list. Is it useful ? 
I've been able to remove them and build, so I guess it's not :)
We need to remove them.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -317,52 +306,44 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
 }
 
 template<class ClusterT>
-int ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder) const
+bool ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder) const
 {
     if (empty())
```
sylvlecl: we can remove this, it's redundant with the following "all_of"

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -179,38 +179,31 @@ void ClusterList<ClusterT>::storeTimeseriesNumbers(Solver::IResultWriter& writer
...
-        ++(groupCount[newcluster->groupId()]);
-        rebuildIndex();
-        return cluster[newcluster->id()];
-    }
-    return nullptr;
+    if (!newcluster)
+        return nullptr;
+
+    if (exists(newcluster->id()))
+        return this->find(newcluster);
```
sylvlecl: it does not make sense to perform a search here, we can just return the provided pointer which is equivalent

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -281,33 +269,34 @@ bool ClusterList<ClusterT>::rename(Data::ClusterName idToFind, Data::ClusterName
...
+        c->markAsModified();
 }
 
 template<class ClusterT>
 bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
 {
-    auto i = cluster.find(id);
-    if (i == cluster.end())
+    // using find_if to get an iterator and use erase later
+    const auto& it = find_if(cluster.begin(), cluster.end(),
```
sylvlecl: You may use the range version here to avoid iterator pair.
Actually I think you can replace the 2 class with `std::vector::erase_if`

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -317,52 +306,44 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
...
+    return std::all_of(cluster.begin(), cluster.end(), [&folder](const auto& c){
+        return c->saveDataSeriesToFolder(folder);
+    });
 }
 
 template<class ClusterT>
-int ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder, const String& msg) const
+bool ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder, const String& msg) const
 {
     if (empty())
```
sylvlecl: Not really useful

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -317,52 +306,44 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
...
     return ret;
 }
 
 template<class ClusterT>
-int ClusterList<ClusterT>::loadDataSeriesFromFolder(Study& s,
+bool ClusterList<ClusterT>::loadDataSeriesFromFolder(Study& s,
                                                     const StudyLoadOptions& options,
                                                     const AnyString& folder)
 {
     if (empty())
```
sylvlecl: Not really useful

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -379,22 +360,30 @@ void ClusterList<ClusterT>::retrieveTotalCapacityAndUnitCount(double& total, uin
...
 }
 
+template<class ClusterT>
+uint ClusterList<ClusterT>::removeDisabledClusters()
+{
+    // nothing to do if there is no cluster available
+    if (empty())
+        return 0;
+
+    auto firstClusterToRemove = std::remove_if(cluster.begin(), cluster.end(), [] (auto& c) {
```
sylvlecl: you may use just `std::vector::erase_if` instead of the 2 calls (new in C++ 20)

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -219,10 +207,8 @@ class ClusterList
...
     //! All clusters
-    Map cluster;
+    Vect cluster;
```
sylvlecl: I think we should push further the cleanup of this class:

1. If the point of this PR is to clean up this class, we must make this private.
The class is responsible to guarantee that the list remains sorted. It cannot guarantee this if we leave it public.
We should also make `rebuildIndex` private, it's the responsibility of this class to maintain the internal consistency.

2. **(Optionally)** As a whole, it may not be the best choice to remove the map and not remove the vector:
a lot of operations of the class involve maintain the sort order and searching through it, which is what a map does naturally.

With a map, we can just expose a public getter (for example a range view) for when we want to iterate on clusters:
```cpp
// assuming clusters is a map
<some range type here> getClusters() {
    return clusters | std::views::values;
}
```

src/libs/antares/study/cleaner/cleaner-v20.cpp
```diff
@@ -135,23 +135,19 @@ static void listOfFilesAnDirectoriesToKeepForArea(PathList& e, PathList& p, cons
...
 
-        auto end = area->thermal.list.end();
-        for (auto i = area->thermal.list.begin(); i != end; ++i)
+        for (const auto& cluster : area->thermal.list)
         {
-            // Reference to the thermal cluster
-            auto& cluster = *(i->second);
-
-            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster.id();
+            buffer.clear() << "input/thermal/prepro/" << id << '/' << cluster->id();
```
a-zakir, line 140: please use std::filesystem::path

🔢

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -58,21 +59,27 @@ typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::end() cons
...
+
+    return nullptr;
 }
 
 template<class ClusterT>
 ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id)
 {
-    auto i = cluster.find(id);
-    return (i != cluster.end()) ? i->second.get() : nullptr;
+    for (auto& c : cluster)
```
a-zakir: may i suggest :
for (auto& cluster: clusters)

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -58,21 +59,27 @@ typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::end() cons
...
+            return c.get();
+
+    return nullptr;
 }
 
 template<class ClusterT>
 typename std::shared_ptr<ClusterT> ClusterList<ClusterT>::detach(iterator i)
 {
-    SharedPtr c = i->second;
+    SharedPtr c = *i;
```
a-zakir: cluster

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -109,32 +113,28 @@ Data::ClusterList<ClusterT>::~ClusterList()
...
 {
-    auto end = cluster.end();
-    for (auto i = cluster.begin(); i != end; ++i)
-    {
-        if (p == i->second.get())
-            return i->second.get();
-    }
+    for (const auto& c : cluster)
+        if (c.get() == p)
+            return c.get();
```
a-zakir: don't recall c.get() just return p ;)

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -179,38 +179,31 @@ void ClusterList<ClusterT>::storeTimeseriesNumbers(Solver::IResultWriter& writer
...
-    }
-    return nullptr;
+    if (!newcluster)
+        return nullptr;
+
+    if (exists(newcluster->id()))
+        return this->find(newcluster);
+
+    cluster.push_back(newcluster);
+    ++(groupCount[newcluster->groupId()]);
```
a-zakir: this need a simplification ;)

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -242,36 +235,31 @@ bool ClusterList<ClusterT>::rename(Data::ClusterName idToFind, Data::ClusterName
     Antares::TransformNameIntoID(newName, newID);
 
     // Looking for the renewable cluster in the list
-    auto it = cluster.find(idToFind);
-    if (it == cluster.end())
+    auto* cluster_ptr = this->find(idToFind);
```
a-zakir, line 160: auto it=  was ok

a-zakir:
recommendation: std::shared_ptr

src/libs/antares/InfoCollection/StudyInfoCollector.cpp
```diff
@@ -59,13 +59,9 @@ void StudyInfoCollector::enabledThermalClustersCountToFileContent(FileContent& f
     for (auto i = study_.areas.begin(); i != end; ++i)
     {
         Area& area = *(i->second);
-        auto end = area.thermal.list.end();
-        for (auto i = area.thermal.list.begin(); i != end; ++i)
-        {
-            auto& cluster = i->second;
+        for (const auto& cluster : area.thermal.list)
```
flomnes: You could use std::count_if here

src/libs/antares/study/cleaner/cleaner-v20.cpp
flomnes: Maybe we'll do this later, in a separate PR. Or maybe we will never do it because we will change the file format to JSON/yaml/protobuf/avro/etc.

src/libs/antares/study/parts/common/cluster.cpp
```diff
@@ -49,45 +49,43 @@ void Cluster::setName(const AnyString& newname)
...
-        }
-        return 0;
+        bool ret = true;
+        buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series.txt";
+        ret = series.timeSeries.saveToCSVFile(buffer, precision()) && ret;
+
+        return ret;
     }
-    return 1;
+    return false;
```
flomnes: The caller should change it's behavior a bit, adding a ! to this return value

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -58,21 +59,27 @@ typename ClusterList<ClusterT>::const_iterator ClusterList<ClusterT>::end() cons
 template<class ClusterT>
 const ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id) const
 {
-    auto i = cluster.find(id);
-    return (i != cluster.end()) ? i->second.get() : nullptr;
+    for (const auto& c : cluster)
+        if (c->id() == id)
```
flomnes: You may use std::find or rather std::find_if here

src/libs/antares/study/parts/common/cluster_list.cpp
flomnes: This may be used in the UI. But I agree returning the operation of get is NOT memory-safe and should be avoided

flomnes: As guilpier said, we should consider removing the raw-pointer version of this function

src/libs/antares/study/parts/common/cluster_list.cpp
flomnes: On the contrary, I would remove the raw-pointer version

src/libs/antares/study/parts/common/cluster_list.h
flomnes: Yes open an issue please

src/libs/antares/study/study.cpp
```diff
@@ -1128,12 +1128,8 @@ void Study::destroyAllWindTSGeneratorData()
 void Study::destroyAllThermalTSGeneratorData()
 {
     areas.each([&](Data::Area& area) {
-        auto pend = area.thermal.list.end();
-        for (auto j = area.thermal.list.begin(); j != pend; ++j)
-        {
-            ThermalCluster& cluster = *(j->second);
-            FreeAndNil(cluster.prepro);
-        }
+        for (const auto& cluster : area.thermal.list)
```
flomnes, line 1131: 6 lines -> 2 lines. Nice !

flomnes (changes requested):
Almost good

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -51,25 +52,17 @@ class ClusterList
     template<class PredicateT>
     void each(const PredicateT& predicate)
     {
-        auto end = cluster.cend();
-        for (auto i = cluster.cbegin(); i != end; ++i)
-        {
-            auto& it = *(i->second);
-            predicate(it);
-        }
+        for (auto& c : cluster)
```
sylvlecl: If the point of this PR is to clean up this class, I would propose to remove those iterations methods.

Their use should be replaced with STL algos, here std::for_each.

sylvlecl (changes requested):
I think we should push further the improvements:
 - by encapsulating the list as a private field.
 - We also need to clean up the interface from not useful methods (find pointers...). 
 - And we also need to take care of performances (no linear search).

src/libs/antares/study/cleaner/cleaner-v20.cpp
a-zakir: ok

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: 2 seemingly useless **find** functions were removed

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: ok. There are now 2 **find** "unique" functions (= with a different body)

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: `find(const ClusterList<ClusterT>::SharedPtr& p)` was removed

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: **std::accumulate** replaced with **std::all_of**, more suitable. Done by @payetvin

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: **std::accumulate** ==> **std::all_of**, better. Done by @payetvin

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: **std::all_of** is better, done by @payetvin

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -379,22 +329,30 @@ void ClusterList<ClusterT>::retrieveTotalCapacityAndUnitCount(double& total, uin
...
+        total += c->unitCount * c->nominalCapacity;
     }
 }
 
+template<class ClusterT>
+uint ClusterList<ClusterT>::removeDisabledClusters()
+{
+    // nothing to do if there is no cluster available
+    if (empty())
+        return 0;
```
guilpier-code, line 277:
```C++
if (empty())
    return 0;
```
is useless. 
If list is empty, we don't start any loop or algorithm anyway.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -317,52 +278,41 @@ bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
...
+    return std::all_of(cluster.begin(), cluster.end(), [&folder](const auto& c){
+        return c->saveDataSeriesToFolder(folder);
+    });
 }
 
 template<class ClusterT>
-int ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder, const String& msg) const
+bool ClusterList<ClusterT>::saveDataSeriesToFolder(const AnyString& folder, const String& msg) const
 {
     if (empty())
```
guilpier-code: About : 
```C++
if (empty())
    return true;
```
We should try to remove them : they are often useless.

src/libs/antares/study/parts/renewable/container.cpp
```diff
@@ -66,40 +66,14 @@ void PartRenewable::prepareAreaWideIndexes()
         return;
     }
 
-    auto end = list.end();
     uint idx = 0;
```
guilpier-code, line 69: About block : 
```C++
    // Copy the list with all renewable clusters
    // And init the areaWideIndex (unique index for a given area)
    if (list.empty())
    {
        return;
    }
```
We need neither the comment (which is incorrect, by the way [there is no copy]), nor the `if (list.empty())` block.

src/libs/antares/study/parts/renewable/container.cpp
guilpier-code: Done by @payetvin

src/libs/antares/study/parts/common/cluster.cpp
```diff
@@ -49,45 +49,43 @@ void Cluster::setName(const AnyString& newname)
...
-            ret = series.timeSeries.saveToCSVFile(buffer, precision()) && ret;
-
-            return ret;
-        }
-        return 0;
+        bool ret = true;
+        buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series.txt";
+        ret = series.timeSeries.saveToCSVFile(buffer, precision()) && ret;
+
+        return ret;
```
flomnes:
```suggestion
        buffer.clear() << folder << SEP << parentArea->id << SEP << id() << SEP << "series.txt";
        return series.timeSeries.saveToCSVFile(buffer, precision());
```

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -281,88 +236,69 @@ bool ClusterList<ClusterT>::rename(Data::ClusterName idToFind, Data::ClusterName
...
 template<class ClusterT>
 bool ClusterList<ClusterT>::remove(const Data::ClusterName& id)
 {
-    auto i = cluster.find(id);
-    if (i == cluster.end())
-        return false;
-
-    // Getting the pointer on the cluster
-    SharedPtr c = i->second;
+    auto nbDeletion = std::erase_if(clusters, [&id](const SharedPtr& c) { return c->id() == id; });
```
flomnes, line 210: Careful, `std::erase_if` doesn't remove elements. See
https://en.cppreference.com/w/cpp/container/vector/erase2, https://en.cppreference.com/w/cpp/algorithm/remove and the [Erase-remove idiom](https://en.wikipedia.org/wiki/Erase-remove_idiom)

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -281,88 +236,69 @@ bool ClusterList<ClusterT>::rename(Data::ClusterName idToFind, Data::ClusterName
...
+    bool ret = true;
     uint ticks = 0;
 
-    auto end = cluster.end();
-    for (auto it = cluster.begin(); it != end; ++it)
+    for (const auto& c : clusters)
     {
-        auto& cluster = *(it->second);
-        logs.info() << msg << "  " << (ticks * 100 / (1 + this->cluster.size()))
+        logs.info() << msg << "  " << (ticks * 100 / (1 + this->clusters.size()))
```
flomnes, line 236: Use `std::all_of`

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -14,171 +15,113 @@ namespace Antares
...
-    //! iterator to the begining of the list
     const_iterator begin() const;
 
-    //! iterator to the end of the list
     iterator end();
-    //! iterator to the end of the list
     const_iterator end() const;
 
+    SharedPtr& operator[](std::size_t idx) { return clusters[idx]; }
+    const SharedPtr& operator[](std::size_t idx) const { return clusters[idx]; }
```
flomnes: It would be better to return by copy rather than by reference.

The reference could become invalid if `std::vector` is resized, for example if the resize triggers a memory realloc.

Copying a `std::shared_ptr` isn't costly given it's size.

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -190,76 +133,71 @@ class ClusterList
...
+    */
+    uint removeDisabledClusters();
+
+protected:
+    /// The vector containing the clusters
+    Vect clusters;
+
+private:
+    /// thermal, renewable, etc.
+    std::string typeID = "common";
```
flomnes: This won't work. We need dynamic polymorphism for this, and static member variables can't have that. Please revert to `virtual` member function.

src/libs/antares/study/parts/thermal/container.cpp
flomnes:
```suggestion
        for (auto i : list)
```

src/ui/simulator/toolbox/components/map/manager.cpp
```diff
@@ -768,7 +768,7 @@ void Manager::selectFromBoundingBox(const wxPoint& a, const wxPoint& b, const si
                             for (Data::ThermalClusterList::iterator t = area->thermal.list.begin();
                                  t != tend;
                                  ++t)
-                                clusterlist.push_back(t->second.get());
+                                clusterlist.push_back((*t).get());
```
flomnes, line 771:
```suggestion
                                clusterlist.push_back(t->get());
```

flomnes (changes requested):
- Minor remarks: style
- Major remarks
  - `std::erase`, `std::remove`, etc.
  - `typeID` => revert to `virtual` function

src/libs/antares/study/parts/common/cluster_list.cpp
payetvin: erase_if does both the remove and erase

flomnes: Oh it's a shiny new feature from C++20, sorry for the confusion !

src/libs/antares/study/parts/renewable/container.cpp
guilpier-code: This function was removed, we use the parent class version of this function.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -109,66 +103,30 @@ Data::ClusterList<ClusterT>::~ClusterList()
...
-        if (0 == n)
-        {
-            each([&](Cluster& cluster) { cluster.series.timeseriesNumbers.clear(); });
-        }
-        else
-        {
-            each([&](Cluster& cluster) { cluster.series.timeseriesNumbers.resize(1, n); });
-        }
-    }
+    if (n == 0)
```
guilpier-code: We don't need to separate cases **n ==0** and **n != 0** : 
Indeed, if **n == 0** and we call `resize(1, n)`, then method **clear(...)** is called anyway behind the scene (please see **resize(...)**'s definition in **Matrix**).
So I guess : 
```C++
template<class ClusterT>
void ClusterList<ClusterT>::resizeAllTimeseriesNumbers(uint n)
{
    assert(n < 200000); // arbitrary number
    each([&](Cluster& cluster) { cluster.series.timeseriesNumbers.resize(1, n); });
}
```
is enough.

src/libs/antares/study/parts/common/cluster_list.cpp
```diff
@@ -22,80 +24,63 @@ using namespace Antares;
...
-ClusterT* ClusterList<ClusterT>::find(const Data::ClusterName& id)
-{
-    auto i = cluster.find(id);
-    return (i != cluster.end()) ? i->second.get() : nullptr;
+    return (it != clusters.end()) ? it->get() : nullptr;
 }
 
 template<class ClusterT>
-typename std::shared_ptr<ClusterT> ClusterList<ClusterT>::detach(iterator i)
+const ClusterT* ClusterList<ClusterT>::find(const ClusterT* p) const
```
guilpier-code: Now that we simplified the **find(...)** functions, why wouldn't we have : 
```C++
template<class ClusterT>
const ClusterT* ClusterList<ClusterT>::find(const ClusterT* p) const
{
    return find(p->id());
}
```
?

guilpier-code: I don't remember if I once tested if this function `find(const ClusterT* p)` is called anywhere.
I would test that, but honestly, with the duration and CPU a recompilation would take, I feel lazy.

guilpier-code (changes requested):

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: Made a quick search, `find(const ClusterT* p)` does not seem to called.
We should try to remove it and see what happens at compilation.

src/libs/antares/study/parts/common/cluster_list.cpp
guilpier-code: We can even wonder if **each(...)** is useful, given that **std** contains now all kinds of facilities.
By the way, looking at *each(...)**'s content, it's a simple call to an **std** function.
Don't know if we can remove it easily though.
And it's the same for functions : **begin(...)**, **end(...)**, **empty()**, **size()**, ...

guilpier-code (changes requested):

PR #1815, opened 2023-12-11
reshape constraint construction

a-zakir:
follows #1627

### DIFF
--- a/src/solver/optimisation/ProblemMatrixEssential.h
+++ b/src/solver/optimisation/ProblemMatrixEssential.h
@@ -43,7 +43,7 @@ public:
     {
         for (auto& group : constraintgroups_)
         {
-            group->buildConstraints();
+            group->BuildConstraints();
         }
     }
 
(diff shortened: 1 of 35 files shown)


src/solver/optimisation/constraints/AreaHydroLevel.h
```diff
@@ -1,18 +1,31 @@
 #pragma once
 #include "ConstraintBuilder.h"
 
+struct AreaHydroLevelData
+{
+    std::vector<CORRESPONDANCES_DES_CONTRAINTES>& CorrespondanceCntNativesCntOptim;
```
flomnes:
```suggestion
    const std::vector<CORRESPONDANCES_DES_CONTRAINTES>& CorrespondanceCntNativesCntOptim;
```

a-zakir: can't be const see [usage](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/0af56167402535b0f631656b3221980a3dd43850/src/solver/optimisation/constraints/AreaHydroLevel.cpp#L5C5-L5C14).

PR #1786, opened 2023-11-22
Parallel tests

payetvin:
close #1759

PR #1816, opened 2023-12-11
Parallel tests during scheduled jobs

payetvin:
#1786

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -175,15 +175,6 @@ jobs:
         batch-name: valid-v870
         os: ${{ matrix.test-platform }}
 
-    - name: Run parallel tests
-      if: ${{ env.IS_PUSH == 'true' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ubuntu.yml
```diff
@@ -208,6 +199,15 @@ jobs:
         batch-name: adequacy-patch-CSR
         os: ${{ matrix.test-platform }}
 
+    - name: Run parallel tests
+      if: ${{ env.RUN_EXTENDED_TESTS == 'true' }}
```
JasonMarechal25, line 203: When is "RUN_EXTENDED_TESTS" true (or false) ?

payetvin: During scheduled runs at night, or during releases

payetvin: parallel tests takes the same time as short tests, we decided it's shouldn't be run for each job to save time

JasonMarechal25 (approved):

PR #1825, opened 2023-12-15
Take into account breaking change in OR-Tools's API, enable SCIP & GLPK solvers, bump OR-Tools

flomnes:
- `SetStartingLpBasisInt` and `GetFinalBasisInt` have been removed from recent versions, due to redundancy with their non-Int counterparts.
- Bump OR-Tools v9.5-rte2.0 -> v9.5-rte3.0 
- Add SCIP solver in lookup list (works better for MILP) and GLPK


src/solver/utils/ortools_utils.cpp
```diff
@@ -289,6 +289,17 @@ MPSolver* ORTOOLS_ConvertIfNeeded(const std::string& solverName,
     }
 }
 
+template<class SourceT>
+static void transferBasis(std::vector<operations_research::MPSolver::BasisStatus>& destination,
+                          const SourceT& source)
```
tbittar, line 294: J'aurais trouvé plus logique une signature `tranferBasis(source, destination)` plutôt que `tranferBasis(destination, source)`

flomnes: On utilise la convention `destination = source`, où `destination` est à gauche du signe égal.

payetvin (approved):

PR #1827, opened 2023-12-18
Fix pmax contribution

a-zakir:


flomnes (approved):
I proposed the same fix after investigating for about 1h30min ! Revert YAML and we can merge

PR #1828, opened 2023-12-18
Remove yuni volatile warning

payetvin:
Antares_Simulator/src/ext/yuni/src/yuni/../yuni/core/string/../traits/extension/../../smartptr/policies/../../atomic/int.hxx:99:90: warning: ‘--’ expression of ‘volatile’-qualified type is deprecated [-Wvolatile]
   99 |     return (threadSafe) ? Private::AtomicImpl::Operator<size, TP>::Decrement(*this) : (--pValue);

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -203,7 +203,7 @@ jobs:
         os: ${{ matrix.test-platform }}
 
     - name: Run parallel tests
-      if: ${{ env.IS_PUSH == 'true' }}
+      if: ${{ env.RUN_EXTENDED_TESTS == 'true' }}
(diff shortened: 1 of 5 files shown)


flomnes:
Is policy `SingleThreaded` ever used ? It is the only one for which `threadSafe = false`. For other policies (`ObjectLevelLockable,` `ObjectLevelLockableNotRecursive`, `ClassLevelLockable`), `threadSafe = true`.

What you're doing is assuming that `threadSafe` is always `true`, in other words that `SingleThreaded` is never used.

flomnes (approved):
I added those static_asserts, the code still builds fine. So I assume that `threadSafe` is always `true`
```cpp
    static_assert(!threadSafe);
```

PR #1834, opened 2023-12-19
Fix 2 build warnings

flomnes:
- `#pragma once` in .cpp file
- `-Wreorder` because member initialization is out of order


a-zakir (approved):

flomnes:
Tests are broken on Windows, see #1835 for fix (unrelated to current PR)

PR #1835, opened 2023-12-19
Fix flawed concurrency test (concurrent write on non-atomic int)

flomnes:
Multiple threads writing to a non-atomic object is UB. Use `std::atomic` to guarantee result and fix broken test.

```
D:/a/Antares_Simulator/Antares_Simulator/src/tests/src/libs/antares/concurrency/test_concurrency.cpp(86): error: in "test_future_set": check counter == 10 has failed
```


sylvlecl (approved):
Thanks!

PR #1836, opened 2023-12-20
Fix compile warnings on Linux with g++>=10

payetvin:
- Use `#pragma GCC diagnostic ignored` where needed, as narrow as possible
- Remove useless asserts
- Add a few `[[maybe_unused]]` where needed

### DIFF
--- a/src/ui/simulator/application/study.cpp
+++ b/src/ui/simulator/application/study.cpp
@@ -573,7 +573,8 @@ void MarkTheStudyAsModified(const Data::Study::Ptr& study)
     if (!(!study) and study == GetCurrentStudy())
         MarkTheStudyAsModified();
 }
-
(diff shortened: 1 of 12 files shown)


src/ui/simulator/application/study.cpp
```diff
@@ -573,7 +573,8 @@ void MarkTheStudyAsModified(const Data::Study::Ptr& study)
     if (!(!study) and study == GetCurrentStudy())
         MarkTheStudyAsModified();
 }
-
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wvolatile"
```
flomnes: You could place these around the code that triggers the warning to make it clear what piece of code triggers the warning

flomnes (approved):

PR #1837, opened 2023-12-21
Add optimization options in OR-Tools / XPRESS, refactor

flomnes:
This PR allows the user to specify solver specific parameters from the command line. The work on basis manipulation is deferred to a subsequent PR.

This PR adds a command-line argument :
  - `--solver-parameters="param1 value1 param2 value2"` to specify solver specific parameters

The parameter syntax is solver specfic, and the user should refer to the specific solver documentation to know how to set the wanted parameters. PR #2042 will give some explicit examples.

Allow the user to change the following properties
- [X] Presolve (using solver specific parameters)
- [X] Scaling (using solver specific parameters)
- [ ] ~~Use basis for 1st resolution~~
- [ ] ~~Use basis for 2nd resolution~~
- [ ] ~~Primal tolerance~~
- [ ] ~~Dual tolerance~~
- [X] Any solver specific parameters (with subtle cases of ORTools generic parameters that will override them)
- [X] I/O for generaldata.ini
- [x] Add docs

Use a single structure for all these options.


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -241,7 +241,14 @@ static SimplexResult OPT_TryToCallSimplex(
     if (options.useOrtools)
     {
         const bool keepBasis = (optimizationNumber == PREMIERE_OPTIMISATION);
-        solver = ORTOOLS_Simplexe(&Probleme, solver, keepBasis);
+        solver = ORTOOLS_Simplexe(&Probleme,
```
a-zakir: this is getting long, we could create an appropriate struct for ortools/solver options

a-zakir: OptimizationOptions?

flomnes: I had the same idea, without reading your comment. See these [2 consecutive commits](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1837/files/07074583e5d29d8c8beb869d3b1f8ee070741511..23f67dc1d69f2d65bff801ecd544bfbe9ab47a25)

src/libs/antares/optimization-options/include/antares/optimization-options/options.h
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Antares::Solver::Optimization {
+struct OptimizationOptions
+{
+    bool useOrtools = false;
+    std::string solverName = "sirius";
+    bool presolve = false;
+    bool scaling = false;
+    bool useBasisOptim1 = false;
+    bool useBasisOptim2 = true;
```
flomnes: Check default values ?

src/solver/utils/ortools_utils.cpp
```diff
@@ -98,6 +103,8 @@ void ProblemSimplexeNommeConverter::TuneSolverSpecificOptions(MPSolver* solver)
     default:
         break;
     }
+    // How / When to check if user has input wrong parameters ?
+    solver->SetSolverSpecificParametersAsString(solverParameters);
```
flomnes: This function returns `true` in case of success, `false` in case of error. Also, make sure to pass `THREADS 1`, otherwise XPRESS may to run multiple threads by default. https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/THREADS.html

src/solver/utils/ortools_utils.cpp
```diff
@@ -294,43 +301,52 @@ bool solveAndManageStatus(MPSolver* solver, int& resultStatus, const MPSolverPar
...
+}
 template<class SourceT>
 static void transferBasis(std::vector<operations_research::MPSolver::BasisStatus>& destination,
                           const SourceT& source)
 {
     destination.resize(source.size());
     for (size_t idx = 0; idx < source.size(); idx++)
     {
-        destination[idx] = source[idx]->basis_status();
+        destination[idx] = source[idx]->assibasis_status();
```
flomnes: Google won't accept anything like that. They want conversion of statuses in `MPSolver::BasisStatus`.

tbittar:
Some solvers do not implement `SetSolverSpecificParameters` in or-tools. For these solver, specifying parameters from `solver-parameters` in `generaldata.ini` will result in an exception thrown by the solver.

.github/workflows/centos7.yml
```diff
@@ -5,6 +5,7 @@ on:
     types: [created]
   push:
     branches:
+      - fix/xpress-perf
```
pet-mit: don't forget to revert this

src/libs/antares/checks/checkLoadedInputData.cpp
```diff
@@ -49,6 +51,34 @@ void checkOrtoolsUsage(Antares::Data::UnitCommitmentMode ucMode,
     }
 }
 
+void checkOrtoolsSolverSpecificParameters(
+  const Antares::Data::UnitCommitmentMode& ucMode,
+  const Antares::Solver::Optimization::OptimizationOptions& optOptions)
+{
+    if (optOptions.ortoolsUsed)
+    {
+        MPSolver* solver;
```
pet-mit: creating a disposable MPSolver object has an overhead. Do we really need to check the parameters separately?

src/libs/antares/optimization-options/include/antares/optimization-options/options.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+namespace Antares::Solver::Optimization {
+struct OptimizationOptions
+{
+    //! Force ortools use
+    bool ortoolsUsed = false;
+    //! THe solver name, sirius is the default
```
pet-mit:
```suggestion
    //! The solver name, sirius is the default
```

src/libs/antares/optimization-options/include/antares/optimization-options/options.h
pet-mit, line 1: add copyright header

src/solver/application/application.cpp
```diff
@@ -325,6 +326,10 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
     {
         loadingException = std::current_exception();
     }
+
+    //For solver
```
pet-mit:
```suggestion
    // For solver
```

src/solver/utils/ortools_utils.cpp
```diff
@@ -24,18 +24,40 @@
 #include <antares/exception/AssertionError.hpp>
 #include "antares/antares/Enum.hpp"
 #include <filesystem>
+#include "../optimisation/include/antares/solver/optimisation/opt_constants.h"
```
pet-mit: avoid relative include's

pet-mit (changes requested):

src/libs/antares/checks/checkLoadedInputData.cpp
tbittar: Maybe not, it seems that if solver specific parameters are not successfully applied, ortools issues a warning and do not crash, so we may not need to check that parameters are set with success, what do you think @flomnes ?

tbittar: Howerver, the user should then carrefully check the logs to see if its parameters are really taken into account

pet-mit: It's because Google has a "no crash" policy, but we don't have to have the same policy.
The SetSolverSpecificParameters has a boolean return value. If we want to crash instead of making the user read the logs, I suggest checking this boolean at call time (before solve) and crashing then, instead of creating a disposable MPSolver instance beforehand, which would have a runtime cost.
PS: I, personnally, am in favor of crashing, instead of ignoring the issue

pet-mit:
@tbittar 
Would you mind documenting the new parameter here? (preferably in a PR from a new branch to doc/optim_update)
https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/doc/optim_update/docs/user-guide/solver/04-parameters.md

docs/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ toc_depth: 2
 --------------------
 ### Improvements
 * STS groups are now "dynamic" : group names are no longer fixed by code, user is free to define these groups.
+* Solver specific parameters can be specified in the generaldata.ini (#1837)
```
a-zakir: or-tools solvers? only applicable on Xpress?

docs/reference-guide/13-file-format.md
```diff
@@ -38,18 +38,22 @@ Other5_level
 
 The default value for group is "OTHER1".
 
-## Output
-### Hydro maximum generation/pumping power
+#### Optimization : solver parameters (expert option)
+
+In file **settings/generaldata.ini**, in section `optimization` add property `solver-parameters` [str]. Default value = `""`. Allows to specify solver specific parameters in a string formatted as `param1 value1 param2 value2`
```
a-zakir: same remark ⬆️

pet-mit: actually the OR-Tools doc states the following, I think we should do the same (maybe with an addition of an example for xpress):
(from [here](https://github.com/google/or-tools/blob/5124f4975bd6ab5a863ea55657833ab640730429/ortools/linear_solver/linear_solver.h#L676))
"The format is solver-specific and is the same as the corresponding solver configuration file format. Returns true if the operation was successful."

src/libs/antares/checks/checkLoadedInputData.cpp
```diff
@@ -49,6 +51,34 @@ void checkOrtoolsUsage(Antares::Data::UnitCommitmentMode ucMode,
...
+              (OrtoolsUtils::solverMap.at(optOptions.ortoolsSolver)).LPSolverName);
+        }
+
+        bool status = solver->SetSolverSpecificParametersAsString(optOptions.solverParameters);
+        if (!status)
+        {
+            throw Error::InvalidSolverSpecificParameters(optOptions.ortoolsSolver,
+                                                         optOptions.solverParameters);
+        }
+        ORTOOLS_LibererProbleme(solver);
```
a-zakir: i'm not sure i fully understand, why you instantiate an MPSolver to just check options? and then it's freed!
 why not just checks options here? 
this work is only about Xpress options right?

src/libs/antares/study/parameters.cpp
```diff
@@ -1522,8 +1527,7 @@ void Parameters::prepareForSimulation(const StudyLoadOptions& options)
         logs.info() << "  :: The problems will contain named variables and constraints";
     }
     // indicated whether solver logs will be printed
-    logs.info() << "  :: Printing solver logs : " << (solverLogs ? "True" : "False");
-    
+    logs.info() << "  :: Printing solver logs : " << (optOptions.solverLogs ? "True" : "False");
```
a-zakir, line 1754:
```suggestion
    logs.info() << "  :: Printing solver logs : " <<std::boolalpha << optOptions.solverLogs ;
```

src/libs/antares/study/parameters.cpp
```diff
@@ -1084,12 +1091,7 @@ bool Parameters::loadFromINI(const IniFile& ini, StudyVersion& version, const St
     if (options.forceDerated)
         derated = true;
 
-    // Define ortools parameters from options
-    ortoolsUsed = options.ortoolsUsed;
-    ortoolsSolver = options.ortoolsSolver;
-
     namedProblems = options.namedProblems;
-    solverLogs = options.solverLogs || solverLogs;
```
a-zakir, line 1279: where did you moved it?
it says: file options OR CLI options

a-zakir: how do you manage ""solver-parameters" in the .ini and it's CLI equivalent ?

src/solver/optimisation/opt_liberation_problemes_simplexe.cpp
```diff
@@ -54,11 +55,10 @@ void OPT_LiberationProblemesSimplexe(const OptimizationOptions& options, const P
     {
         for (int numIntervalle = 0; numIntervalle < nbIntervalles; numIntervalle++)
         {
-            auto ProbSpx
-              = (PROBLEME_SPX*)(ProblemeAResoudre->ProblemesSpx[numIntervalle]);
+            auto ProbSpx = (PROBLEME_SPX*)(ProblemeAResoudre->ProblemesSpx[numIntervalle]);
             auto solver = (MPSolver*)(ProblemeAResoudre->ProblemesSpx[numIntervalle]);
 
-            if (options.useOrtools && solver != NULL)
+            if (options.ortoolsUsed && solver != NULL)
```
a-zakir:
```suggestion
            if (options.ortoolsUsed && solver)
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -122,7 +120,7 @@ static SimplexResult OPT_TryToCallSimplex(
     {
         if (problemeHebdo->ReinitOptimisation)
         {
-            if (options.useOrtools && solver != nullptr)
+            if (options.ortoolsUsed && solver != nullptr)
```
a-zakir:
```suggestion
            if (options.ortoolsUsed && solver)
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -256,7 +257,7 @@ static SimplexResult OPT_TryToCallSimplex(
     {
         if (ProblemeAResoudre->ExistenceDUneSolution != SPX_ERREUR_INTERNE)
         {
-            if (options.useOrtools && solver != nullptr)
+            if (options.ortoolsUsed && solver != nullptr)
```
a-zakir:
```suggestion
            if (options.ortoolsUsed && solver)
```

src/libs/antares/checks/checkLoadedInputData.cpp
tbittar: Ok I agree with the remark (that was my first idea), but @flomnes mentioned that if we do this way, we do a lot of things before crashing (reading data, constructing the problem), so the user may wait a while before the crash. This is why I tried to do the check earlier.

docs/CHANGELOG.md
tbittar: Yes only for solvers called with ortools, therefore xpress, coin, glpk, scip. I will add the detail in the docs

docs/reference-guide/13-file-format.md
tbittar: Yes and for instance, with Xpress, you should write `THREADS 1` whereas with scip you would write `parallel/maxnthreads = 1`, so this is an advanced usage and the user must refer to the docs of the specific solver to know how to use it

src/libs/antares/checks/checkLoadedInputData.cpp
tbittar: I agree, this is related to Peter's remark, see the corresponding answer. To check the options we need to know the return code of `solver->SetSolverSpecificParameters`, therefore we need a `MPSolver` object

src/libs/antares/study/parameters.cpp
tbittar: What is CLI ?

a-zakir: command line option

src/libs/antares/checks/checkLoadedInputData.cpp
pet-mit: I understand his point of view, but I think it's better to penalize the user "a lot" once (ie in the one study run with wrong parameters, they would have to wait before crash), than to penalize them "a little" each and every time (ie add a small overhead to check that the parameters are correct in every study, after they have corrected the parameters)

src/libs/antares/study/parameters.cpp
a-zakir: > how do you manage ""solver-parameters" in the .ini and it's CLI equivalent ?

sorry you did not add an option in the command line "solver-parameters", still solver-logs can be requested either by this way or in the .ini

src/libs/antares/study/parameters.cpp
tbittar: I have compilation issues with yuni when trying this, therefore I do not include this suggestion

src/libs/antares/study/parameters.cpp
tbittar: Done, command line is correctly handled now and I have added support to set solver-parameters from the CLI

src/solver/optimisation/opt_liberation_problemes_simplexe.cpp
tbittar: Done

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
tbittar: Done

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
tbittar: Done

src/libs/antares/checks/checkLoadedInputData.cpp
tbittar: I agree, I have moved the check only at solve

docs/reference-guide/10-command_line.md
```diff
@@ -41,6 +41,7 @@ _In all cases, arguments " –h" or "–help" can be used to get help_
 |-m, --mps-export | Export anonymous mps weekly or daily optimal UC+dispatch linear |
 |-s, --named-mps-problems | Export named mps weekly or daily optimal UC+dispatch linear |
 |--solver-logs | Print solver logs |
+|--solver-parameters | Set solver specific parameters: <ul><li>With Xpress : `--solver-parameters="param1 value1 param2 value2"`</li><li>With SCIP : `--solver-parameters="param1 value1, param2 value2"`</li></ul> Command-line parameters override those specified in generaldata.ini within `xpress-parameters` or `scip-parameters` fields |
```
flomnes: This is markdown, not HTML. Please remove markups and use the [markdown syntax](https://www.markdownguide.org/basic-syntax/) instead.

src/CMakeLists.txt
```diff
@@ -464,4 +464,4 @@ endif()
 include(CPack)
 
 #Last. We need all target defined
-add_subdirectory(packaging)
+# add_subdirectory(packaging)
```
flomnes: Revert this

src/libs/antares/study/parameters.cpp
flomnes: Yuni logs don't support iomanip operations from std (`std::setprecision`, `std::boolalpha`, etc.)

src/libs/antares/study/parameters.cpp
```diff
@@ -1633,7 +1661,15 @@ void Parameters::saveToINI(IniFile& ini) const
...
-        section->add("solver-logs", solverLogs);
+        section->add("solver-logs", optOptions.solverLogs);
+        if (optOptions.ortoolsSolver == "xpress")
+        {
+            section->add("xpress-parameters", optOptions.solverParameters);
+        }
+        else if (optOptions.ortoolsSolver == "scip")
+        {
+            section->add("scip-parameters", optOptions.solverParameters);
+        }
```
flomnes: This is very risky. You should have one field for each solver, to avoid losing information when changing the solver.

```suggestion
            section->add("xpress-parameters", optOptions.solverParameters.xpress);
            section->add("scip-parameters", optOptions.solverParameters.scip);
        }
```

flomnes (commented):
Display solver specific options in the logs (ortools solver used xpress **with specific options XXX***)

flomnes (commented):
For consistency, it may be better to use command-line parameters `--xpress-parameters` and `--scip-parameters`, like in generaldata.ini.

a-zakir (approved):

docs/reference-guide/10-command_line.md
tbittar: A bullet list inside a table does not render well if I use the markdown syntax, therefore I switch to html for this particular feature: https://stackoverflow.com/questions/19950648/how-to-write-lists-inside-a-markdown-table It renders well on the readthedocs. 

Does this have an impact elsewhere ? PDF generation maybe ?

pet-mit: I think we should try to avoid html as much as possible too

pet-mit (approved):

src/solver/utils/ortools_utils.cpp
```diff
@@ -38,6 +42,56 @@ static void setGenericParameters(MPSolverParameters& params)
...
+    case MPSolver::XPRESS_MIXED_INTEGER_PROGRAMMING:
+    {
+        specificParams = XPRESS_PARAMS + " " + solverParameters.xpress;
+        status = solver->SetSolverSpecificParametersAsString(specificParams);
+        checkSetSolverSpecificParameters(status, solverName, specificParams);
+        break;
+    }
+    case MPSolver::SCIP_MIXED_INTEGER_PROGRAMMING:
+    {
+        specificParams = SCIP_PARAMS + ", " + solverParameters.scip;
```
flomnes: I think SCIP_PARAMS/XPRESS_PARAMS should be overriden (ie ignored) if the user provides custom parameters.

Otherwise it seems impossible to set a different value for the number of threads

flomnes (commented):
We're close ! A few additional comments

flomnes (commented):
`checkOrtoolsSolverSpecificParameters` is declared twice, but never defined or used. Please remove it.

flomnes (commented):
Also, failure occurs way too late. Solver specific parameters should be check at the start, not in ORTOOLS_Simplexe (during the simulation)

src/solver/utils/ortools_utils.cpp
tbittar: I think it works as is as the `SetSolverSpecificParameters` will set parameters in parsing order so in theory threads can be set twice (one from XPRESS/SCIP PARAMS, and one from user specific choice). But this is really unclear, so I will change according to your suggestion

tbittar:
Yes, this was already discussed in https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1837#discussion_r1565328633

src/solver/utils/ortools_utils.cpp
tbittar: Finally, following our discussion, I will keep the current version of the code as the behavior is ok

PR #1838, opened 2023-12-21
Use `type*` for pointers instead of `auto&`, add `const` qualifier

flomnes:


src/solver/utils/ortools_utils.cpp
```diff
@@ -156,7 +156,7 @@ static void extractSolutionValues(const std::vector<MPVariable*>& variables,
     int nbVar = problemeSimplexe->NombreDeVariables;
     for (int idxVar = 0; idxVar < nbVar; ++idxVar)
     {
-        auto& var = variables[idxVar];
+        auto* var = variables[idxVar];
```
a-zakir: while we're at it: 
const auto* ?

flomnes: Yeah, right !

a-zakir (approved):

PR #1809, opened 2023-12-04
Renewable containers

payetvin:
#1487

PR #1844, opened 2023-12-22
Thermal clusters lists : cleaning

guilpier-code:
This PR follows the PR #1809 (about cleaning common cluster list and renewable cluster list)
It aims at simplifying thermal cluster lists as much as possible.

**Goal** : 
Having only one list of thermal clusters.
This list would be **private**
Code would interact with this list through an interface. This interface would contain functions returning **std::views** that would result in filtering thermal clusters using a certain property of clusters : enabled / disabled, must-run or not, a combination of these. 

**What's done** : 
- [x] Simplifying the `ClusterList<T>::add()` method
- [x] Changing **std::map** `mapping` from class **ThermalClusterList** into a **std::vector** (named `allClusters_`)
- [x] Removing **PartThermal::clusters** (often reached by `area.thermal.clusters`)
- [x] Removing **PartThermal::mustrunList** (often reached by `area.thermal.mustrunList`)
- [x] Removing **PartThermal::list** (often reached by `area.thermal.list.clusters`)

**Before starting a review, one should know several things** : 
- what the old lists of clusters used to contain : 
    -  **area.thermal.list.mapping** : all clusters
    - **area.thermal.mustrunList.clusters** : clusters must-run and enabled
    - **area.thermal.clusters** : enabled clusters
    - **area.thermal.list.clusters** : list used to build the optimization problem (clusters enabled and not must-run). Careful with that list : when loading clusters from input, this list contains all clusters. Then it is progressively cleared from disabled and must-run clusters. So its content depends on where we are in the execution. This list is also used from the **GUI**. In that case, it always contains all clusters : it is **not** cleared from specific clusters.
- the list **area.thermal.list.clusters** is the only list used in the **GUI**, and in that case, it contains all clusters.
- As said previously, all lists of clusters are removed, except **area.thermal.list.mapping**, turned into a private **std::vector** (**area.thermal.list.allClusters_**) containing all clusters, and filtered to access only clusters of interest at any point of the code.
- **cluster indices** : any cluster (thermal / renewable) used to have 2 indices : 
    - **index** : it turns out to be an index regarding only enabled and not must-run clusters. So this index never concerns renewable clusters. It was moved down from parent class **Cluster** to class **ThermalCluster**.
    - **areaWideIndex** : index regarding only enabled clusters. Concerns both thermal and renewable clusters
- the 2 lists  **area.thermal.mustrunList.clusters** and **area.thermal.list.clusters** are closely related (inter-dependent) to the **renewable** cluster list, due to a common inheritance. So this work deeply impacts the **renewable** cluster lists.


src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -242,6 +242,9 @@ class ThermalCluster final : public Cluster, public std::enable_shared_from_this
 
     //! Mustrun
     bool mustrun;
+    bool isMustRun() const { return mustrun; }
+    bool isNotMustRun() const { return not mustrun; }
```
flomnes: You might use https://en.cppreference.com/w/cpp/utility/functional/logical_not

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -84,19 +84,16 @@ static void StudyRuntimeInfosInitializeAllAreas(Study& study, StudyRuntimeInfos&
...
 
         area.scratchpad.reserve(nbYearsInParallel);
         for (uint numSpace = 0; numSpace < nbYearsInParallel; numSpace++)
             area.scratchpad.emplace_back(r, area);
 
         // statistics
-        r.thermalPlantTotalCount += area.thermal.list.size();
-        r.thermalPlantTotalCountMustRun += area.thermal.mustrunList.size();
+        r.thermalPlantTotalCount += area.thermal.list.enabledAndNotMustRunCount();
+        r.thermalPlantTotalCountMustRun += area.thermal.list.mustRunAndEnabledCount();
```
flomnes: `enabled` first everywhere ?

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -260,11 +257,10 @@ void StudyRuntimeInfos::checkThermalTSGeneration(Study& study)
     thermalTSRefresh = globalThermalTSgeneration;
 
     study.areas.each([this, globalThermalTSgeneration](Data::Area& area) {
-        area.thermal.list.each(
-          [this, globalThermalTSgeneration](const Data::ThermalCluster& cluster) {
-              thermalTSRefresh
-                = thermalTSRefresh || cluster.doWeGenerateTS(globalThermalTSgeneration);
-          });
+        for (auto c : area.thermal.list.each_enabled_and_not_mustrun())
```
flomnes, line 254: Some redundancy. In some cases, we want to iterate over a container. In other cases, we want the number of elements.

It would be nicer to have
```cpp
for (auto c : thermal.list.enabledNotMustRun.each())
{
 // ...
}
auto nbClusters = thermal.list.enabledNotMustRun.count();
```
Also, `clear`
```cpp
thermal.list.enabledNotMustRun.clear();
```

src/libs/antares/study/runtime/runtime.h
```diff
@@ -136,8 +136,6 @@ class StudyRuntimeInfos
     void initializeRangeLimits(const Study& study, StudyRangeLimits& limits);
     //! Prepare all thermal clusters in 'must-run' mode
     void initializeThermalClustersInMustRunMode(Study& study) const;
-    void removeDisabledThermalClustersFromSolverComputations(Study& study);
-    void removeDisabledRenewableClustersFromSolverComputations(Study& study);
     void removeDisabledShortTermStorageClustersFromSolverComputations(Study& study);
```
flomnes: TODO

src/libs/antares/study/parts/renewable/cluster_list.cpp
```diff
@@ -156,20 +165,13 @@ bool RenewableClusterList::loadFromFolder(const AnyString& folder, Area* area)
...
-                    // This error should never happen
-                    logs.error() << "Impossible to add the renewable cluster '" << cluster->name()
-                                 << "'";
-                    continue;
-                }
+                addToCompleteList(cluster);
             }
         }
 
+        giveIndicesToClusters();
```
flomnes: `rebuildIndex` ?

src/ui/action/handler/antares-study/area/create.cpp
```diff
@@ -293,7 +293,7 @@ void Create::createActionsForAStandardAreaCopy(Context& ctx, bool copyPosition)
         auto* root = new RootNodePlant(pOriginalAreaName);
 
         // browsing each thermal cluster
-        for (auto& c : area->thermal.list)
```
guilpier-code, line 290: About : 
```C++
for (auto& c : area->thermal.list)
```
Actually, `area->thermal.list.clusters` was traversed here, because of method **begin()** and **end()** (now gone) were used behind the scene.

src/libs/antares/study/parts/common/cluster.h
```diff
@@ -24,37 +24,21 @@ struct CompareClusterName final
 class Cluster
 {
 public:
-    //! Map of renewable clusters
-    using Map = std::map<ClusterName, Cluster*>;
+    using Set = std::set<Cluster*, CompareClusterName>;
```
guilpier-code: `using Set = std::set<Cluster*, CompareClusterName>;` : only moved from the bottom to the top of the class

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -28,41 +27,17 @@ class ClusterList
...
     using SharedPtr = typename std::shared_ptr<ClusterT>;
-    using Vect = typename std::vector<SharedPtr>;
-    using iterator = typename Vect::iterator;
-    using const_iterator = typename Vect::const_iterator;
 
-    /*!
-    ** \brief Iterate through all clusters (const)
-    */
-    template<class PredicateT>
-    void each(const PredicateT& predicate) const
```
guilpier-code: `void each(const PredicateT& predicate) const` : 
**each(...)** was used to traverse list **clusters** (clusters enabled and not must-run) and process a task on each element. We don't need it anymore : it was often replaced with a loop over the filtered complete list.

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -28,41 +27,17 @@ class ClusterList
...
-    */
-    void clear();
-
-    /*!
-    ** \brief Add a cluster in the list
-    **
-    ** \param t The cluster to add
-    ** \return True if the cluster has been added, false otherwise
-    */
-    SharedPtr add(const SharedPtr clusters);
```
guilpier-code: `SharedPtr add(const SharedPtr clusters); ` : 
is replace with : 
`void addToCompleteList(std::shared_ptr<ClusterT> cluster);`

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -28,41 +27,17 @@ class ClusterList
...
+    bool empty() const;
 
     /*!
-    ** \brief Try to find a cluster from its id (const)
+    ** \brief Try to find a cluster from its id (const) in the complete cluster list
     **
     ** \param id ID of the cluster to find
     ** \return A pointer to a cluster. nullptr if not found
     */
-    ClusterT* find(const Data::ClusterName& id) const;
```
guilpier-code: `ClusterT* find(const Data::ClusterName& id) const;` : 
Used to be called only in **GUI**, searches in list **clusters**.
In **GUI**, **clusters** was containing all clusters.
It's been renamed **findInAll** , and now searches in private list **allClusters_**

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -72,6 +47,10 @@ class ClusterList
     */
     bool exists(const Data::ClusterName& id) const;
 
+    auto each_enabled() const { return allClusters_ | std::views::filter(&ClusterT::isEnabled); }
```
guilpier-code: **each_enabled()** : 
This is where we filter the all clusters list to get only enabled one.
There are other filters, more specific to thermal clusters, involving a must-run status.
For these, see file **thermal/cluster_list.h**

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -89,21 +68,12 @@ class ClusterList
...
-    //! Return true if the list is empty
-    bool empty() const;
     //@}
 
-    iterator begin();
-    const_iterator begin() const;
-
-    iterator end();
-    const_iterator end() const;
+    SharedPtr operator[](std::size_t idx) { return allClusters_[idx]; }
```
guilpier-code: This  **[ ]** operator is now used only in **GUI**.
It used to operate on **thermal.list.clusters**, that was containing all clusters when using the **GUI**.
So its new body is consistent with the old one.
In the **solver** part, the calls to this operator were replaced with something else.
For example, in **sim_calcul_economique.cpp** : 
```C++
for (uint clusterIndex = 0; clusterIndex != area.thermal.list.size(); ++clusterIndex)
{
    auto& cluster = *(area.thermal.list[clusterIndex]);  // call to operator []
    ...
}
```
was replaced with : 
```C++
for (auto cluster : area.thermal.list.each_enabled_and_not_mustrun())
{
    ...
}
```

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -89,21 +68,12 @@ class ClusterList
...
+    const SharedPtr operator[](std::size_t idx) const { return allClusters_[idx]; }
 
-    SharedPtr operator[](std::size_t idx) { return clusters[idx]; }
-    const SharedPtr operator[](std::size_t idx) const { return clusters[idx]; }
+    SharedPtr enabledClusterAt(unsigned int index) const;
```
guilpier-code: `SharedPtr enabledClusterAt(unsigned int index) const;` : 
replaces calls to : 
**area.thermal.clusters[index]**. Remind that **area.thermal.clusters** used contains only **enabled** clusters

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -130,7 +100,7 @@ class ClusterList
     ** \brief Get the size (bytes) occupied in memory by a `ClusterList` structure
     ** \return A size (in bytes)
     */
-    uint64_t memoryUsage() const;
+    virtual uint64_t memoryUsage() const = 0;
```
guilpier-code: **memoryUsage()** became pure virtual to correct a error : 
memory print of a thermal cluster list is not the same as the memory usage of a renewable cluster list, because of must-run clusters, which does not exist in renewable cluster list

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -157,28 +127,21 @@ class ClusterList
...
-    ** \brief Removes disabled clusters
-    **
-    ** All clusters with the flag 'enabled' turned to false will be removed from 'list'.
-    ** As a consequence, they will no longer be seen as thermal clusters
-    ** from the solver's point of view.
-    ** \warning This method should only be used from the solver
-    **
-    ** \return The number of disabled clusters found
-    */
-    uint removeDisabledClusters();
```
guilpier-code: `uint removeDisabledClusters(); ` :  this was removed, because (as its name suggests) it was used to remove disabled clusters from a list. As we now filter a full list, we don't need to remove not matching clusters

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -157,28 +127,21 @@ class ClusterList
...
 
-    /// thermal, renewable, etc.
     virtual std::string typeID() const = 0;
 
+    // Give a special index to enabled clusters (thermal / renewable)
+    void giveIndicesToClusters();
+
 private:
-    /// Sort the vector, set index value for each cluster
-    void rebuildIndex();
```
guilpier-code: `void rebuildIndex();` : it turns out that the indices processed in this function (cluster->index) are useless to renewable clusters. So this function was moved to the more specific class **ThermalClusterList**

src/libs/antares/study/parts/common/cluster.h
```diff
@@ -97,25 +81,24 @@ class Cluster
...
-    //! Enabled
-    bool enabled;
+    bool isEnabled() const { return enabled; }
+    bool enabled = true;
 
     //! The associate area (alias)
     Area* parentArea;
 
-    //! The index of the cluster (within a list)
-    uint index;
```
guilpier-code: **index** : only used by thermal cluster. So this index was moved down to  class **ThermalCluster**

src/libs/antares/study/parts/thermal/cluster_list.h
```diff
@@ -14,8 +15,7 @@ class ThermalClusterList : public ClusterList<ThermalCluster>
 {
 public:
     std::string typeID() const override;
-    // Map container
-    using Map = typename std::map<ClusterName, std::shared_ptr<ThermalCluster>>;
+    uint64_t memoryUsage() const override;
```
guilpier-code: `uint64_t memoryUsage()` : this function turns out to be particular to each kind of cluster list

src/libs/antares/study/parts/thermal/cluster_list.h
```diff
@@ -87,6 +98,14 @@ class ThermalClusterList : public ClusterList<ThermalCluster>
     bool saveEconomicCosts(const AnyString& folder) const;
 
     bool saveToFolder(const AnyString& folder) const override;
+
+    unsigned int mustRunAndEnabledCount() const;
+    unsigned int enabledAndNotMustRunCount() const;
+
+private:
+    // Give a special index to enbled and not must-run THERMAL clusters 
+    void rebuildIndex();
```
guilpier-code: `void rebuildIndex();` : is specific to thermal clusters, so it was moved here. See review of **common/cluster_list.h**

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -38,10 +37,34 @@ std::string ThermalClusterList::typeID() const
     return "thermal";
 }
 
+uint64_t ThermalClusterList::memoryUsage() const
```
guilpier-code, line 61: `memoryUsage() const ` : moved from parent class to each child class (thermal or renewable), because this computation depends on which kind of clusters we're considering

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -38,10 +37,34 @@ std::string ThermalClusterList::typeID() const
...
+    uint64_t ret = sizeof(ThermalClusterList) + (2 * sizeof(void*)) * mustRunAndEnabledCount();
+    std::ranges::for_each(each_enabled_and_not_mustrun(), [&ret](const auto c) { ret += c->memoryUsage(); });
+    return ret;
+}
+
 static bool ThermalClusterLoadFromSection(const AnyString& filename,
                                           ThermalCluster& cluster,
                                           const IniFile::Section& section);
 
+void ThermalClusterList::rebuildIndex()
```
guilpier-code: `rebuildIndex()` ; was move down here from parent class.
Indeed, **cluster->index** only concerns thermal clusters

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -134,20 +157,12 @@ bool ThermalClusterList::loadFromFolder(Study& study, const AnyString& folder, A
 
         // Check the data integrity of the cluster
         cluster->integrityCheck();
-
-        // adding the thermal cluster
-        auto added = add(cluster);
```
guilpier-code, line 160: Here we used to add a thermal cluster freshly loaded from input in 2 different lists.
We don't even know how `add(cluster);` can return a null pointer.
Now we add the cluster to only one list, and more simply.

src/libs/antares/study/parts/common/cluster_list.h
```diff
@@ -157,28 +127,21 @@ class ClusterList
...
 protected:
-    /// The vector containing the clusters
-    Vect clusters;
+    std::vector<std::shared_ptr<ClusterT>> allClusters_;
 
-    /// thermal, renewable, etc.
     virtual std::string typeID() const = 0;
 
+    // Give a special index to enabled clusters (thermal / renewable)
+    void giveIndicesToClusters();
```
guilpier-code: `giveIndicesToClusters()` replaces `PartThermal::prepareAreaWideIndexes()` : 
this function gives an index to enabled clusters. See comment in the new function.
Ok : the name is still bad, even worse that it used to be.
We should do something about it.

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -134,20 +157,12 @@ bool ThermalClusterList::loadFromFolder(Study& study, const AnyString& folder, A
...
-            logs.error() << "Impossible to add the thermal cluster '" << cluster->name()
-                         << "'";
-            continue;
-        }
-        // keeping track of the cluster
-        mapping[cluster->id()] = added;
+        addToCompleteList(cluster);
     }
 
+    giveIndicesToClusters();
```
guilpier-code: Here we give each thermal cluster a value for their 2 indices once for all.
Before this work, we were forced to call these 2 functions in several places : 
- when we removed disabled clusters from the list used to build optimization problem (clusters enabled ans not must-run)
- when we removed must-run clusters from that same list

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -274,23 +289,25 @@ bool ThermalClusterLoadFromSection(const AnyString& filename,
 
 void ThermalClusterList::calculationOfSpinning()
 {
-    each([&](ThermalCluster& cluster) { cluster.calculationOfSpinning(); });
+    for (auto cluster : each_enabled_and_not_mustrun())
```
guilpier-code: **each(...)** was removed (from parent class).
Much more clear now, here and below.

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -306,96 +323,96 @@ bool ThermalClusterList::saveToFolder(const AnyString& folder) const
         // Allocate the inifile structure
         IniFile ini;
 
-        // Browse all clusters
-        each([&](const Data::ThermalCluster& c) {
```
guilpier-code, line 331: **saveToFolder(...)** method is only used from **GUI**, so the associated list of clusters is the full list (all clusters).

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -415,14 +432,15 @@ bool ThermalClusterList::savePreproToFolder(const AnyString& folder) const
     Clob buffer;
     bool ret = true;
 
-    each([&](const ThermalCluster& c) {
```
guilpier-code, line 439: Same as above : **savePreproToFolder(...)** only called from GUI, so it processes on all clusters

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -431,11 +449,12 @@ bool ThermalClusterList::saveEconomicCosts(const AnyString& folder) const
...
```
guilpier-code, line 455: Same as above : **saveEconomicCosts** processes on all clusters

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -448,9 +467,7 @@ bool ThermalClusterList::loadPreproFromFolder(Study& study,
 
     Clob buffer;
     auto hasPrepro = [&](auto c)
-    {   
-        ++options.progressTicks;
```
guilpier-code, line 473: About : 
```C++
 ++options.progressTicks;
  options.pushProgressLogs();
```
These instructions (here and in other places) are used to log the progress of the operation of loading clusters.
We thought that it was not useful, and that it was bloating the code, so we removed it all.

src/libs/antares/study/parts/renewable/cluster.h
```diff
@@ -38,15 +38,6 @@ namespace Antares
 {
 namespace Data
 {
-/* gp : useless in v8.1, could be useful in a further version
```
guilpier-code: **YAGNI** (you ain't gonna need it) ==> bye bye

src/libs/antares/study/parts/renewable/cluster_list.cpp
```diff
@@ -29,27 +37,28 @@ bool RenewableClusterList::saveToFolder(const AnyString& folder) const
         IniFile ini;
 
         // Browse all clusters
-        each([&](const Data::RenewableCluster& c) {
```
guilpier-code, line 53: As already said somewhere, **saveToFolder(...)** is only called from the **GUI**, so it processes the full cluster list (unfiltered).

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -232,30 +232,36 @@ class ThermalCluster final : public Cluster, public std::enable_shared_from_this
 
     bool isActive() const;
 
+    //! The index of the cluster (within a list)
+    uint index = 0;
```
guilpier-code: **index** : only useful for thermal clusters, so it was moved from parent class to the current one

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -232,30 +232,36 @@ class ThermalCluster final : public Cluster, public std::enable_shared_from_this
...
     **
     ** This value is computed from the field 'group' in 'group()
     ** \see group()
     */
-    ThermalDispatchableGroup groupID;
+    ThermalDispatchableGroup groupID = thermalDispatchGrpOther1;
 
     //! Mustrun
-    bool mustrun;
+    bool mustrun = false;
```
guilpier-code: We should make `bool mustrun` private

src/libs/antares/study/parts/thermal/cluster.h
```diff
@@ -232,30 +232,36 @@ class ThermalCluster final : public Cluster, public std::enable_shared_from_this
...
+    bool isMustRun() const { return mustrun; }
+    bool isNotMustRun() const { return not mustrun; }
```
guilpier-code: **isMustRun()** and **isNotMustRun()** added to simplify filtering the full cluster list with predicates

src/libs/antares/study/parts/renewable/container.h
```diff
@@ -61,16 +61,6 @@ class PartRenewable
...
 
-
-
-    /*!
-    ** \brief Create and initialize the list of all clusters (with the mustrun flag or not)
-    **
-    ** This method must be called before prepareClustersInMustRunMode()
-    ** to ensure the same order whatever the value of the 'mustrun' flag is.
-    */
-    void prepareAreaWideIndexes();
```
guilpier-code: `void prepareAreaWideIndexes(); ` is now internal to the parent cluster list class **ClusterList**.
Gives an index to enabled clusters.

src/libs/antares/study/parts/thermal/container.h
```diff
@@ -59,28 +59,6 @@ class PartThermal
...
 
-
-
-    /*!
-    ** \brief Create and initialize the list of all clusters (with the mustrun flag or not)
-    **
-    ** This method must be called before prepareClustersInMustRunMode()
-    ** to ensure the same order whatever the value of the 'mustrun' flag is.
-    */
-    void prepareAreaWideIndexes();
```
guilpier-code: `void prepareAreaWideIndexes();`  is now internal to the parent cluster list class **ClusterList**.
Gives an index to enabled clusters.

src/libs/antares/study/parts/thermal/container.h
```diff
@@ -59,28 +59,6 @@ class PartThermal
...
-    ** \brief Prepare all thermal clusters in the 'mustrun' mode
-    **
-    ** All clusters with the flag will be moved to the container 'mustrunList'.
-    ** As a consequence, they will no longer be seen as thermal clusters
-    ** from the solver's point of view.
-    ** \warning This method should only be used from the solver
-    **
-    ** \return The number of clusters found in 'must-run' mode
-    */
-    uint prepareClustersInMustRunMode();
```
guilpier-code: `uint prepareClustersInMustRunMode();` : we don't need this function anymore, as we can filter the full list of clusters : we can avoid must-run cluster when traversing that full list.

src/libs/antares/study/parts/thermal/container.h
```diff
@@ -115,31 +93,10 @@ class PartThermal
 
     //! List of all thermal clusters (enabled and disabled) except must-run clusters
     ThermalClusterList list;
-    //! List of all thermal clusters in 'must-run' mode
-    //! \warning This list must only be used from the solver
-    ThermalClusterList mustrunList;
```
guilpier-code: `ThermalClusterList mustrunList;` : we don't need that list. We filter the full cluster list instead (avoiding the must-run clusters).

src/libs/antares/study/parts/thermal/container.h
```diff
@@ -115,31 +93,10 @@ class PartThermal
...
-    /*!
-    ** \brief All clusters for the area
-    **
-    ** This variable is only available at runtime from the solver.
-    ** It is initialized in the same time that the runtime data.
-    **
-    ** This list is mainly used to ensure the same order of the
-    ** thermal clusters in the outputs.
-    */
-    std::vector<ThermalCluster*> clusters;
```
guilpier-code: `std::vector<ThermalCluster*> clusters;` : used to contain enabled clusters only. We don't need it, for the same reason as above.

src/libs/antares/study/parts/thermal/container.cpp
```diff
@@ -49,114 +49,47 @@ bool PartThermal::forceReload(bool reload) const
...
 {
     list.markAsModified();
-    mustrunList.markAsModified();
 }
 
-PartThermal::~PartThermal()
-{
-}
-
-void PartThermal::prepareAreaWideIndexes()
```
guilpier-code, line 60: See review of the associated **.h** file

src/libs/antares/study/parts/thermal/container.hxx
```diff
@@ -1,43 +0,0 @@
...
-** SPDX-License-Identifier: licenceRef-GPL3_WITH_RTE-Exceptions
-*/
-#ifndef __ANTARES_LIBS_STUDY_PARTS_THERMAL_CONTAINER_HXX__
-#define __ANTARES_LIBS_STUDY_PARTS_THERMAL_CONTAINER_HXX__
-
-namespace Antares
-{
-namespace Data
-{
-inline void PartThermal::resizeAllTimeseriesNumbers(uint n)
```
guilpier-code: this function is called once for each area.
It does have to be **inline**, so it does have to fill a whole source file.
We moved it in the **.cpp**

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -84,19 +84,16 @@ static void StudyRuntimeInfosInitializeAllAreas(Study& study, StudyRuntimeInfos&
         // Spinning - Economic Only - If no prepro
         if (!(timeSeriesThermal & study.parameters.timeSeriesToRefresh))
         {
-            // Calculation of the spinning
             area.thermal.list.calculationOfSpinning();
```
guilpier-code, line 81: Here we have a calculate the spinning for clusters of 2 lists.
The union of these 2 lists contains the enabled clusters. 
That's why the new **calculationOfSpinning()** loops over enables clusters

src/libs/antares/study/load-options.h
```diff
@@ -44,9 +44,6 @@ class StudyLoadOptions
     StudyLoadOptions();
     //@}
 
-    //! Push a new log entry according to the local progress data
```
guilpier-code: All these code removals used to be about logging load progression. We found it was bloating the code unnecessarily.

src/libs/antares/study/load-options.cpp
```diff
@@ -53,16 +53,6 @@ StudyLoadOptions::StudyLoadOptions() :
 {
 }
 
-void StudyLoadOptions::pushProgressLogs() const
```
guilpier-code, line 48: See comment in associated **.h** file

src/libs/antares/study/cleaner/cleaner-v20.cpp
```diff
@@ -135,7 +135,7 @@ static void listOfFilesAnDirectoriesToKeepForArea(PathList& e, PathList& p, cons
         buffer.clear() << "input/thermal/clusters/" << id << "/list.ini";
         e.add(buffer);
 
-        for (const auto& cluster : area->thermal.list)
```
guilpier-code, line 130: Here we're in the cleaner.
The list :` area->thermal.list.clusters` used to be traversed. 
In the cleaner, this list used to contain all clusters (special clusters are not removed as in the **solver**).
That's why we **now** traverse ` area->thermal.list.all()`

src/libs/antares/study/binding_constraint/BindingConstraint.cpp
```diff
@@ -274,7 +274,7 @@ void BindingConstraint::copyWeights(const Study &study,
             if (localParent)
             {
                 const ThermalCluster *localTC
-                        = localParent->thermal.list.find(thermalCluster->id());
```
guilpier-code, line 271: These small changes are located in functions only called from **GUI**, where we used to search in the full cluster list (`area.thermal.list.clusters`).
That's why we use new **findInAll**

src/libs/antares/InfoCollection/StudyInfoCollector.cpp
```diff
@@ -59,8 +59,7 @@ void StudyInfoCollector::enabledThermalClustersCountToFileContent(FileContent& f
     for (auto i = study_.areas.begin(); i != end; ++i)
     {
         Area& area = *(i->second);
-        nbEnabledThermalClusters +=
```
guilpier-code, line 82: At this point if execution, old list **area.thermal.list.clusters** (that was used here) was containing enabled and **not** must-run thermal clusters. 
So, by the way, it was useless to filter the enabled clusters.
Now we use  **mustRunAndEnabledCount()**, which count only suitable cluster (please see its body)

src/libs/antares/checks/checkLoadedInputData.cpp
```diff
@@ -169,16 +169,15 @@ static void checkThermalColumnNumber(const Antares::Data::AreaList& areas,
     for (uint areaIndex = 0; areaIndex < areas.size(); ++areaIndex)
     {
         const auto& area = *(areas.byIndex[areaIndex]);
-        for (uint clusterIndex = 0; clusterIndex != area.thermal.clusterCount(); ++clusterIndex)
```
guilpier-code, line 161: Here **area.thermal.clusterCount()** used to count the number of elements of `area.thermal.clusters`, which contained enabled clusters only.
That's why we now call **area.thermal.list.each_enabled()**

src/libs/antares/study/area/list.cpp
```diff
@@ -774,14 +774,6 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
                                              StringT& buffer,
                                              const StudyLoadOptions& options)
 {
-    // Progression
```
guilpier-code, line 772: Logs loading progression. Unnecessary and bloating the code.

src/libs/antares/study/runtime/runtime.cpp
guilpier-code: ok, you mean in the name of the function.
I will rename **mustRunAndEnabledCount()** into **enabledAndMustRunCount()** as soon as I can.

src/libs/antares/study/runtime/runtime.cpp
guilpier-code: So what you suggest is several lists, right ?
If so, it's goes in the opposite direction of the current work, which is to have only one list that we can filter.

src/libs/antares/study/runtime/runtime.h
guilpier-code: What do you mean ?

src/libs/antares/study/scenario-builder/RenewableTSNumberData.cpp
```diff
@@ -32,17 +32,15 @@ bool renewableTSNumberData::apply(Study& study)
 
     const uint tsGenCountRenewable = get_tsGenCount(study);
 
-    uint clusterIndex = 0;
```
guilpier-code, line 55: For renewable clusters, `area.renewable.list` used to contain (at this point of execution) only enabled clusters.
That's why we now loop over `area.renewable.list.each_enabled()`

src/libs/antares/study/scenario-builder/RenewableTSNumberData.cpp
```diff
@@ -66,25 +64,16 @@ void renewableTSNumberData::saveToINIFile(const Study& /* study */,
     if (!pArea)
         return;
 
-// Foreach year
-#ifndef NDEBUG
```
guilpier-code, line 90:
```C++
#ifndef NDEBUG
..
#endif
```
was remove, unnecessary / too few benefit

src/libs/antares/study/scenario-builder/RenewableTSNumberData.cpp
```diff
@@ -99,7 +88,7 @@ bool renewableTSNumberData::reset(const Study& study)
     //   solver or not.
     // WARNING: At this point in time, the variable pArea->renewable.clusterCount()
     //   might not be valid (because not really initialized yet)
-    const uint clusterCount = pArea->renewable.list.size();
```
guilpier-code, line 122: **reset()** can be called both from **GUI** and **solver**. In GUI, the cluster list used contains contains all clusters. Otherwise, only enabled clusters. Setting **clusterCount** with **allClustersCount()** ensures that the matrix (of TS number) reset with **clusterCount** is large enough, as it is a max.

src/libs/antares/study/scenario-builder/ThermalTSNumberData.cpp
```diff
@@ -15,11 +15,11 @@ bool thermalTSNumberData::reset(const Study& study)
     // If an area is available, it can only be an overlay for thermal timeseries
     // WARNING: The total number of clusters may vary if used from the
     //   solver or not.
-    // WARNING: At this point in time, the variable pArea->thermal.clusterCount()
```
guilpier-code, line 38: In this file, the cluster lists usage used to be and still is particularly tricky.
What I mean : 
the general philosophy in the scenario builder is : 
   - if a function of this file is called from **GUI**, the clusters list it processes contains all clusters
   - if a function of this file is called from **solver** it contains only **enabled** clusters (must-run or not)
 
For example : 
- **reset()** is called from **load** higher level functions, so it can be called from both **solver** and **GUI**. This why we have a ternary operator inside this function. But the code **before changes** is wrong : when **loading**, the list **pArea->thermal.list.clusters** contains all cluster. So size **pArea->thermal.list.size() + pArea->thermal.mustrunList.size()** is too large. New code is both more clear and correct.
- **apply()** : this function is only used by the solver. In the old code, it processes list  **area.thermal.clusters**, that contains **enabled thermal clusters** (so it confirms that the intention is to process **enabled clusters** only in the scenario builder on the solver side). That's why **each_enabled()** is called in the new code.
- **saveToINIFile(...)** : this function is only called from the **GUI**. In the old code, it processes **pArea->thermal.list**, which (when using the GUI) contains all clusters. New code is much more clear about that.

src/solver/simulation/solver.hxx
```diff
@@ -627,7 +627,7 @@ void ISimulation<Impl>::allocateMemoryForRandomNumbers(randomNumbers& randomForP
         {
             // logs.info() << "   area : " << a << " :";
             auto& area = *(study.areas.byIndex[a]);
-            size_t nbClusters = area.thermal.list.mapping.size();
```
guilpier-code: Recall that **area.thermal.list.mapping** used to contain all thermal clusters

src/libs/antares/study/parts/thermal/cluster_list.h
```diff
@@ -59,7 +59,18 @@ class ThermalClusterList : public ClusterList<ThermalCluster>
...
+    auto each_mustrun_and_enabled() const 
+    { 
+        return allClusters_ | std::views::filter(&ThermalCluster::isMustRun)
+                            | std::views::filter(&ThermalCluster::isEnabled);
+    }
+
+    auto each_enabled_and_not_mustrun() const 
+    {   
+        return allClusters_ | std::views::filter(&ThermalCluster::isEnabled)
+                            | std::views::filter(&ThermalCluster::isNotMustRun);
```
flomnes: `for (auto& cluster : list.enabled() | list.notMustRun())`

src/libs/antares/study/parts/thermal/cluster.h
payetvin: used in the ui, need getter and setter otherwise

src/libs/antares/study/runtime/runtime.cpp
payetvin: done

payetvin (approved):

PR #1849, opened 2023-12-26
Disable hourly binding constraints for some hours

flomnes:
Disable binding constraints at some hours, depending on RHS values. If a NaN / +/-inf is detected at the given hour, at for a given scenario, the constraint is not provided to the mathematical solver.

Therefore, if the user has provided options to write MPS files, these constraints won't be there either. 

## Syntax

In RHS timeseries (directory input/bindingconstraints), add

```
nan 1 2
nan 1 2
nan 1 2
...
```

## Known limitations
Only applies to **hourly** binding constraints.

## TODO
- [ ] `BC MARG COST` when a BC is disabled for some hours ?


src/solver/optimisation/constraints/BindingConstraintHour.cpp
```diff
@@ -2,16 +2,25 @@
...
+        // By convention, any value that is < 0 represents a non-existing constraint
+        data.CorrespondanceCntNativesCntOptim[pdt]
+          .NumeroDeContrainteDesContraintesCouplantes[cntCouplante] = -1;
+        return;
+    }
+
+    data.CorrespondanceCntNativesCntOptim[pdt]
+      .NumeroDeContrainteDesContraintesCouplantes[cntCouplante]
+      = builder.data.nombreDeContraintes;
+
```
flomnes: The trick is here. Where `NumeroDeContrainteDesContraintesCouplantes[cntCouplante] = -1` is interpreted as "no constraint". This is a general convention used whenever some constraint may or may not be enabled.

src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
```diff
@@ -5,9 +5,36 @@
...
+                   const std::string& input)
+{
+    output.assign(Antares::Constants::nbHoursInAWeek, false);
+
+    std::istringstream ss(input);
+    while (ss) {
+        std::string token;
+        if (std::getline(ss, token, ',')) { // Split by comma
+            try {
+                unsigned int index = std::stoi(token) - 1;
```
a-zakir: test the return value of stoi in case of nonconvertible str and use the more robust  function [strtol](https://cplusplus.com/reference/cstdlib/strtol/)

src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
```diff
@@ -5,9 +5,36 @@
...
+                if (index < Antares::Constants::nbHoursInAWeek && index >= 0)
+                    output[index] = true;
+                else
+                    return false;
+            } catch (const std::invalid_argument& e) {
+                return false;
+            } catch (const std::out_of_range& e) {
```
a-zakir: from if cond in l25 , this catch (const std::out_of_range& e) is unreachable

src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
flomnes: Why is it more robust ? It's a C function

a-zakir: i don't know, why i read atoi, 👎🏽  for me

a-zakir: probably i was thinking about this cases:

`
#include <iostream>
#include <string>
#include <cstdlib>
#include <ios>
using namespace std;
void test_str(std::string s){
    errno = 0;
    cout<<"convert to int: "<<s<<"\n";
    
    
    char* ptr_end{};
    const char*c_s =s.c_str(); 
    cout<<"strtol\n";
    auto  j = std::strtol(c_s, &ptr_end, 10);
    cout<<"j: "<<j<<"\n";
    cout<<"ptr_end: "<<ptr_end<<"\n";
    cout << "ptr_end == c_s: "<<std::boolalpha<<  (c_s == ptr_end)<<"\n";
    
    
    cout<<"stoi\n";
    auto  i = std::stoi(s);
    cout<<"i: "<<i<<"\n";
    cout<<"**************************\n";
    
    
}
int main()
{
    test_str("123abc");
    test_str("123");
    test_str("");

    return 0;
}`

a-zakir: but irrelevant since it make the code harder to read

flomnes: This part of the code was removed BTW

src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
flomnes, line 1: revert this

src/solver/simulation/sim_structure_probleme_economique.h
flomnes, line 1: partially revert this

payetvin (approved):

PR #1851, opened 2023-12-27
Fail if OR-Tools solver is not found

### LINKED ISSUE #1739: antares-solver should fail when OR-Tools solver is not found
**Description**

In version 8.8, when you use options `--use-ortools` and `--ortools-solver=***`, and when the corresponding solver is not found or invalid, antares silently backs up to using sirius (with a small warning log actually).

This is dangerous, because one can easily not see that we did not use the requested solver.

Instead, we should exit on an error message.

flomnes:
This change follows a discussion with @sylvlecl. According to him, falling back to Sirius without OR-Tools is not a desirable behavior since it can be misleading. For example, the user thinks they're using XPRESS, but since it's not available (missing licence, misplaced env variables, etc.), Antares Solver silently uses Sirius with a warning in the logs instead.

The choice has been made to refuse running the simulation when the user requests an unavailable solver, and to throw a nice error message

> [2023-12-27 10:17:30][solver][fatal] Can't use solver 'spx' : it does not match any available OR-Tools solver. Possible choices are coin, sirius, xpress.

### DIFF
--- a/src/libs/antares/exception/LoadingError.cpp
+++ b/src/libs/antares/exception/LoadingError.cpp
@@ -1,3 +1,5 @@
+#include <sstream>
+#include <iomanip>
 #include "antares/exception/LoadingError.hpp"
 
 namespace Antares
@@ -41,8 +43,19 @@ InvalidSimulationMode::InvalidSimulationMode() :
 {
 }
 
-InvalidSolver::InvalidSolver(const std::string& solver) :
- LoadingError("Can't use solver '" + solver + "' : it does not match any available OR-Tools solver")
+static std::string InvalidSolverHelper(const std::string& solver, const std::list<std::string>& availableSolverList) {
+  std::ostringstream message;
+  message << "Can't use solver '" << solver << "' : it does not match any available OR-Tools solver. Possible choices are ";
+  for (const std::string& avail : availableSolverList)
+  {
+    bool last = &avail == &availableSolverList.back();
+    std::string sep = last ? "." : ", ";
+    message << std::quoted(avail) << sep;
+  }
+  return message.str();
+}
+
+InvalidSolver::InvalidSolver(const std::string& solver, const std::list<std::string>& availableSolverList) : LoadingError(InvalidSolverHelper(solver, availableSolverList))
 {
 }
 
--- a/src/libs/antares/exception/LoadingError.cpp
+++ b/src/libs/antares/exception/LoadingError.cpp
@@ -1,6 +1,5 @@
-#include <sstream>
-#include <iomanip>
 #include "antares/exception/LoadingError.hpp"
+#include <sstream>
 
 namespace Antares
 {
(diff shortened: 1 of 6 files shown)


src/solver/misc/options.cpp
```diff
@@ -268,26 +268,18 @@ void checkAndCorrectSettingsAndOptions(Settings& settings, Data::StudyLoadOption
...
 void checkOrtoolsSolver(Data::StudyLoadOptions& options)
 {
-    std::string baseSolver = "sirius";
     if (options.ortoolsUsed)
     {
         const std::list<std::string> availableSolverList = getAvailableOrtoolsSolverName();
-        if (availableSolverList.empty())
-        {
-            throw Error::InvalidSolver(options.ortoolsSolver);
-        }
```
flomnes, line 278: This case is covered below : if the list is empty, we can't find anything in it.

src/solver/misc/options.cpp
```diff
@@ -68,7 +68,7 @@ static std::string availableOrToolsSolversString()
 }
 
 std::unique_ptr<GetOpt::Parser> CreateParser(Settings& settings,
-                                             Antares::Data::StudyLoadOptions& options)
+                                             StudyLoadOptions& options)
```
flomnes, line 56: See `using namespace Antares::Data` at the top of the file

src/libs/antares/exception/LoadingError.cpp
```diff
@@ -41,8 +43,19 @@ InvalidSimulationMode::InvalidSimulationMode() :
...
+  message << "Can't use solver '" << solver << "' : it does not match any available OR-Tools solver. Possible choices are ";
+  for (const std::string& avail : availableSolverList)
+  {
+    bool last = &avail == &availableSolverList.back();
+    std::string sep = last ? "." : ", ";
+    message << std::quoted(avail) << sep;
+  }
+  return message.str();
+}
+
```
flomnes: We need this helper to construct an error message

src/libs/antares/exception/LoadingError.cpp
```diff
@@ -41,8 +43,19 @@ InvalidSimulationMode::InvalidSimulationMode() :
...
- LoadingError("Can't use solver '" + solver + "' : it does not match any available OR-Tools solver")
+static std::string InvalidSolverHelper(const std::string& solver, const std::list<std::string>& availableSolverList) {
+  std::ostringstream message;
+  message << "Can't use solver '" << solver << "' : it does not match any available OR-Tools solver. Possible choices are ";
+  for (const std::string& avail : availableSolverList)
+  {
+    bool last = &avail == &availableSolverList.back();
+    std::string sep = last ? "." : ", ";
+    message << std::quoted(avail) << sep;
+  }
```
a-zakir: is it necessary to present available solver like this way?
if no:
 for (const std::string& solver :avail){
message <<solver<<"\n";}
if yes:
for (const std::string& solver :avail){
message <<solver<<", ";}
message.subtr....

flomnes: Yes, it's pretty much a user helping feature. It's nicer to present available options in case of failure.

src/solver/misc/options.cpp
```diff
@@ -268,26 +268,18 @@ void checkAndCorrectSettingsAndOptions(Settings& settings, Data::StudyLoadOption
...
           = (std::find(
                availableSolverList.begin(), availableSolverList.end(), options.ortoolsSolver)
              != availableSolverList.end());
-
         if (!found)
         {
-            logs.warning() << "Invalid ortools-solver option. Got '" << options.ortoolsSolver
-                           << "'. reset to " << baseSolver;
-            options.ortoolsSolver = baseSolver;
+            throw Error::InvalidSolver(options.ortoolsSolver, availableSolverList);
```
a-zakir: what happened to throw an exception instead of using the base solver?

src/libs/antares/exception/LoadingError.cpp
a-zakir: see this [function](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/5e462c03737157548caf48c1e039277621acb423/src/solver/misc/options.cpp#L56)

flomnes: Good point, see cc879ebba for correction & broader use of `availableOrToolsSolversString`, esp. in function `printSolvers`

src/solver/misc/options.cpp
flomnes: It follows a discussion with @sylvlecl. According to him, falling back to Sirius without OR-Tools is not a desirable behavior since it can be misleading. The user thinks they're using XPRESS, but since it's not available (missing licence, misplaced env variables, etc.), Antares Solver silently uses Sirius with a warning in the logs.

The choice has been made to straight up refuse unavailable solvers.

a-zakir (approved):

src/solver/misc/options.cpp
a-zakir: ok 👍🏽

PR #1852, opened 2023-12-27
Add possibility to release without running any tests

flomnes:
This is useful for releases where we know results will deviate from reference results.

# Currently
## Workflow when results change
1. Disable tests by commenting them in the YML workflows
2. Create a RC release
4. Generate new reference results
5. Uncomment tests in YML workflows
6. Create a definitive release with tests

## Workflow when results don't change
1. Generate new release with tests

# After
## Workflow when results change
1. Create release without running the tests
2. Create a RC release
3. Generate new reference results
4. Generate definitive release with tests

We save two steps (comment/uncomment).

## Workflow when results don't change
1. Generate new release with tests

# Note
It won't be possible to manually create a new release, all releases will have to be created using the `new_release.yml` workflow. This is because the condition for asset upload changes from `${{ github.event_name == 'release' && github.event.action == 'created' }}` to `${{ github.event_name == 'workflow_call' }}`.

# Reference
https://docs.github.com/en/actions/using-workflows/reusing-workflows 
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype

### DIFF
--- a/.github/workflows/new_release.yml
+++ b/.github/workflows/new_release.yml
@@ -9,8 +9,8 @@ on:
       release_name:
         description: "Release name"
         required: true
-      run_tests:
-        description: "true/false"
+      run-tests:
+        description: "Run all tests (true/false)"
         required: true
 jobs:
     release:
@@ -29,3 +29,10 @@ jobs:
             body: |
               Please complete description
 
+
+    ubuntu:
+      name: Release - Ubuntu
+      needs: release
+      uses: ./github/workflows/ubuntu.yml
+      with:
+        run-tests: ${{ inputs.run_tests }}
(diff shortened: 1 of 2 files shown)


.github/workflows/new_release.yml
```diff
@@ -0,0 +1,38 @@
...
+        required: true
+      release_name:
+        description: "Release name"
+        required: true
+      run-tests:
+        description: "Run all tests (true/false)"
+        required: true
+jobs:
+    release:
+      runs-on: ubuntu-latest
```
a-zakir: ubuntu-20.04?

a-zakir (approved):

PR #1857, opened 2023-12-29
Circumvent runner bug where input type is lost

flomnes:
```
The template is not valid. .github/workflows/new_release.yml (Line: 38, Col: 20): Unexpected value 'false'
```

https://github.com/AntaresSimulatorTeam/Antares_Simulator/actions/runs/7357236644

See https://github.com/actions/runner/issues/2206, and more specifically https://github.com/actions/runner/issues/2206#issuecomment-1532246677


a-zakir (approved):

.github/workflows/ubuntu.yml
```diff
@@ -22,7 +22,7 @@ env:
   GITHUB_TOKEN: ${{ github.token }}
   IS_RELEASE: ${{ github.event_name == 'workflow_call' }}
   RUN_SIMPLE_TESTS: ${{ github.event_name == 'push' || inputs.run-tests }}
```
a-zakir, line 24: et là?

PR #1863, opened 2024-01-04
Remove YUNI_STATIC_ASSERT

### LINKED ISSUE #1853: Use `static_assert` instead of `YUNI_STATIC_ASSERT`
See **src/ext/yuni/src/yuni/core/static/assert.h**
```cpp
#ifdef YUNI_HAS_CPP_STATIC_ASSERT
#define YUNI_STATIC_ASSERT(X, ID) static_assert((X), #ID)
#else
#define YUNI_STATIC_ASSERT(X, ID)                                  \
    struct StaticAssert_##ID                                       \
    {                                                              \
        enum                                                       \
        {                                                          \
            Assert_##ID = Yuni::Static::Assert<(0 != (X))>::result \
        };                                                         \
    };                                                             \
    typedef char invokeStaticAssert_##ID[StaticAssert_##ID::Assert_##ID]
#endif
```

`static_assert` gained official support since C++11, and a message-free version was added in C++17. We no longer need this preprocessor macro.

Then we might as well remove `Yuni::Static::Assert` if it's not used elsewhere.

Reference [`static_assert`@cppreference](https://en.cppreference.com/w/cpp/language/static_assert)

payetvin:
close  #1853

### DIFF
--- a/src/libs/antares/study/sets.hxx
+++ b/src/libs/antares/study/sets.hxx
@@ -404,9 +404,8 @@ template<class T>
 template<class StringT>
 inline bool Sets<T>::hasOutput(const StringT& s) const
 {
-    using namespace Yuni;
     // Assert, if a C* container can not be found at compile time
-    YUNI_STATIC_ASSERT(Traits::CString<StringT>::valid, HasOutput_InvalidTypeForBuffer);
+    static_assert(Yuni::Traits::CString<StringT>::valid);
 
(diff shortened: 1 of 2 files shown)


payetvin:
`Yuni::Static::Assert` is used internally in Yuni in several files, can't remove `assert.h`

a-zakir (approved):

Issue #1671, opened 2023-10-03
Perform hydro checks at the start instead of every year


flomnes:
Currently, hydro checks are done every year.

This can lead to half completed studies if some hydro data is invalid for some (but not all) years.

See `HydroManagement::makeVentilation`, especially `checkMinGeneration`.

This is made possible (or easier) by the removal of global variables for time series numbers.

guilpier-code:
> This is made possible (or easier) by the removal of global variables for time series numbers.

The TS numbers are no longer contained in a global variable (so this applies to **hydro** TS numbers).
Instead, they are contained in each container of generation means (load, wind, solar, **hydro**, thermal, ...), nested inside any area.
For **hydro**, this means TS numbers are contained somewhere inside class **PartHydro**.

From the text of this issue, what I imagine is something like : we should move checks on hydro min generation right after hydro data is read. And a good place to do move these checks to is **Application::prepare(...)** : 
```C++
void Application::prepare(int argc, char* argv[])
{
    ...
    // Loading the study
    readDataForTheStudy(options);
    ...
    // Here some checks are done
    // It would be a good place for checks on hydro min generation
    ...
}
```

That said, as the code is done now (I mean in the current state of **develop** branch), this move may cause several troubles :
- it requires that the hydro TS numbers are drawn, which they are not when we have just loaded the hydro data. We could do without it : by making checks for every **mingen** TS, even if never selected (that s never associated to any year).
- PR **CR 23** (#1723) implies changes on the checks about **mingen**. Waiting for this PR (almost ready) to be merged in **develop**  would avoid solving conflicts.

What I may recommend : 
- We should wait for the **CR 23** to be integrated in **develop**
- We have to choose between 2 options : 
   - moving the TS numbers draws sooner in the execution : as soon as the random number generators are initialized (see **initializeRandomNumberGenerators(...)**)
   - don't wait for the random generator, do the checks for every **mingen** TS

flomnes:
We can wait for CR23 to be merged.

Then the checks have to be performed after TS numbers have been drawn.

payetvin:
already in gopro

PR #1864, opened 2024-01-04
Move checks on mingen

guilpier-code:
This PR intends to fix issue #1671.
At time of this writing, questions are raised, asking for answer.
Please see related issue (1671).


flomnes:
Nothing to see here

PR #1865, opened 2024-01-08
Use dedicated URL path for Doxygen, instead of root

a-zakir:
In order to add benchmark charts in [gh-pages](https://antaressimulatorteam.github.io/Antares_Simulator/) this PR moves the Doxygen doc from the root page to [doxygen-doc](https://antaressimulatorteam.github.io/Antares_Simulator/doxygen).


.github/workflows/doxygen.yml
```diff
@@ -35,11 +35,10 @@ jobs:
         # If you're changing the branch from main,
         # also change the `main` in `refs/heads/main`
         # below accordingly.
-        # /!\ tmp 
         if: github.ref == 'refs/heads/develop'
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./html
-
+          destination_dir: ./doxygen
```
flomnes: Is the `if:` at line 38 really required ? This action is only triggered on a push to branch `develop`

a-zakir: right!

payetvin (approved):

PR #1867, opened 2024-01-09
Fix HTML generation for readthedocs

a-zakir:


readthedocs.yml
```diff
@@ -1,9 +1,12 @@
 version: 2
 
+build:
+  os: "ubuntu-20.04"
+  tools:
+    python: "3.8"
```
flomnes, line 6: According to [this page](https://www.python.org/downloads/release/python-380/), Python 3.8 is outdated

> Note: The release you're looking at is Python 3.8.0, an outdated release. Python 3.11 is now the latest feature release series of Python 3.

Do we really want to use this particular version ?

a-zakir: I don't know, the  '8' symbol is aesthetically more interesting. ✴️

flomnes (approved):

PR #1873, opened 2024-01-11
Bump OR-Tools (fix crash with XPRESS)

flomnes:
Since OR-Tools 9.5-rte3.0 does not support `SetStartingLpBasis` (it supports `SetStartingLpBasisInt`), we must use 9.8-rte1.0-beta2 instead.

Using XPRESS with Antares Simulator 8.8.0 (OR-Tools 9.5-rte3.0) causes a crash with the following error message
```
W0111 15:00:55.315875 169991 environment.cc:282] Optimizer version: 39.01.01 (OR-Tools was compiled with version 41).
[2024-01-11 15:00:55][solver][infos] Solver OR-Tools MPS File: `problem-1-1--optim-nb-2.mps'
F0111 15:00:55.821070 169991 linear_solver.h:1766] Not supported by this solver.
*** Check failure stack trace: ***
Abandon (core dumped)
```

Also, don't convert CMAKE_BUILD_TYPE to lower-case. It causes issues with SCIP and antares-deps.


a-zakir (approved):

a-zakir:
not sure, but a fix could be using "with" context [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/bfdedf65b866d7f407d065d3ec602ef7973c1878/src/tests/run-study-tests/check_on_results/check_logs_content.py#L34)

PR #1887, opened 2024-01-17
Use std::shared_ptr instead of indices for active binding constraints in results

### LINKED ISSUE #1454: Do not use indexes for active binding constraints
Active binding constraints are returned by a function. Using fixed indexes implicitly uses the fact that this function will always return the same vector.

This assumption is not satisfied if e.g binding constraints are added after a first call to this function.

I suggest using `std::shared_ptr<BindingConstraint>` instead. Since those indexes are used to access arrays, I suggest turning those arrays into `std::map<std::shared<BindingConstraint>, double*>` or something similar.

```cpp
pValuesForTheCurrentYear[numSpace].day[dayInTheYear]
  -= state.problemeHebdo
       ->ResultatsContraintesCouplantes[bindConstraintGlobalIndex_]
       .variablesDuales[dayInTheWeek];
```

payetvin:
close #1454


src/libs/antares/study/binding_constraint/BindingConstraintsRepository.h
```diff
@@ -150,9 +150,9 @@ class BindingConstraintsRepository final : public Yuni::NonCopyable<BindingConst
 
     static std::vector<std::shared_ptr<BindingConstraint>> LoadBindingConstraint(EnvForLoading env);
 
-    [[nodiscard]] std::vector<std::shared_ptr<BindingConstraint>> activeContraints() const;
+    [[nodiscard]] std::vector<std::shared_ptr<BindingConstraint>> activeConstraints() const;
 
-    [[nodiscard]] std::vector<uint> getIndicesForInequalityBindingConstraints() const;
+    [[nodiscard]] std::vector<std::string> getNamesForInequalityBindingConstraints() const;
```
flomnes:
```suggestion
    [[nodiscard]] std::vector<std::shared_ptr<BindingConstraint>> getNamesForInequalityBindingConstraints() const;
```

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -540,7 +540,7 @@ struct PROBLEME_HEBDO
 
     uint32_t NombreDeContraintesCouplantes = 0;
     std::vector<CONTRAINTES_COUPLANTES> MatriceDesContraintesCouplantes;
-    std::vector<RESULTATS_CONTRAINTES_COUPLANTES> ResultatsContraintesCouplantes;
+    std::unordered_map<std::string, RESULTATS_CONTRAINTES_COUPLANTES> ResultatsContraintesCouplantes;
```
flomnes:
```suggestion
    std::unordered_map<std::shared_ptr<BindingConstraint>, RESULTATS_CONTRAINTES_COUPLANTES> ResultatsContraintesCouplantes;
```

src/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
```diff
@@ -394,8 +402,8 @@ class BindingConstMarginCost
     typename VCardType::IntermediateValuesType pValuesForTheCurrentYear = nullptr;
     unsigned int pNbYearsParallel = 0;
     std::shared_ptr<Data::BindingConstraint> associatedBC_ = nullptr;
-    int bindConstraintGlobalIndex_ = -1;
-    uint nbCount_ = 0; // Number of inequality BCs 
+    std::string bindConstraintName;
```
flomnes:
```suggestion
    std::shared_ptr<BindingConstraint> bindConstraintName;
```

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -289,37 +289,38 @@ void SIM_AllocationConstraints(PROBLEME_HEBDO& problem,
...
           .assign(NombreDePasDeTemps, 0.);
 
-        problem.MatriceDesContraintesCouplantes[k].NumeroDeLInterconnexion
-          .assign(activeConstraints[k]->linkCount(), 0);
-        problem.MatriceDesContraintesCouplantes[k].PoidsDeLInterconnexion
-          .assign(activeConstraints[k]->linkCount(), 0.);
-        problem.MatriceDesContraintesCouplantes[k].OffsetTemporelSurLInterco
-          .assign(activeConstraints[k]->linkCount(), 0);
+        problem.MatriceDesContraintesCouplantes[constraintIndex].NumeroDeLInterconnexion
+          .assign(bc->linkCount(), 0);
```
flomnes, line 303: Hardest part : you need to use a `map` here as well

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -289,37 +289,38 @@ void SIM_AllocationConstraints(PROBLEME_HEBDO& problem,
...
     {
-        assert(k < activeConstraints.size());
-        assert(activeConstraints[k]->linkCount() < 50000000);
-        assert(activeConstraints[k]->clusterCount() < 50000000);
+        assert(constraintIndex < bindingConstraintCount);
 
-        problem.MatriceDesContraintesCouplantes[k].SecondMembreDeLaContrainteCouplante
+        auto bc = activeConstraints[constraintIndex];
+        assert(bc->linkCount() < 50000000);
+        assert(bc->clusterCount() < 50000000);
```
flomnes: Remove

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
```diff
@@ -329,41 +329,33 @@ void BindingConstraintsRepository::markAsModified() const
...
+    return activeConstraints_;
 }
 
 static bool isBindingConstraintTypeInequality(const Data::BindingConstraint& bc)
 {
     return bc.operatorType() == BindingConstraint::opLess || bc.operatorType() == BindingConstraint::opGreater;
 }
 
-std::vector<uint> BindingConstraintsRepository::getIndicesForInequalityBindingConstraints() const
+std::vector<std::string> BindingConstraintsRepository::getNamesForInequalityBindingConstraints() const
```
flomnes:
```suggestion
std::vector<std::shared_ptr<BindingConstraint>> BindingConstraintsRepository::getNamesForInequalityBindingConstraints() const
```

flomnes (changes requested):

src/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -163,6 +163,8 @@ struct CONTRAINTES_COUPLANTES
     std::vector<int> OffsetTemporelSurLePalierDispatch;
 
     const char* NomDeLaContrainteCouplante;
+
+    std::shared_ptr<Data::BindingConstraint> addressOfTheBindingConstraint;
```
flomnes:
```suggestion
    std::shared_ptr<Data::BindingConstraint> bindingConstraint;
```

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -335,8 +334,10 @@ void SIM_AllocationConstraints(PROBLEME_HEBDO& problem,
             nbTimeSteps = 0;
             break;
         }
+        RESULTATS_CONTRAINTES_COUPLANTES res;
+        res.variablesDuales.assign(nbTimeSteps, 0.);
         if (nbTimeSteps > 0)
-            problem.ResultatsContraintesCouplantes[k].variablesDuales.assign(nbTimeSteps, 0.);
+            problem.ResultatsContraintesCouplantes.emplace(bc, res);
```
flomnes: You could change `RESULTATS_CONTRAINTES_COUPLANTES`'s constructor a bit to write this more compactly
```suggestion
if (nbTimeSteps > 0)
    problem.ResultatsContraintesCouplantes.emplace(bc, nbTimeSteps);
```

Full example
```cpp
#include <iostream>
#include <vector>
#include <map>

struct RESULT {
  RESULT(size_t size) : v_(size) {}
  std::size_t size() const {return v_.size();}
private:
  std::vector<int> v_;
};

int main() {
  std::map<int, RESULT> m;
  m.emplace(3, 4);
  for (auto& [k, v] : m)
    std::cout << k << " " << v.size() << std::endl;
  // 3 4
}
```

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -336,7 +335,7 @@ void SIM_AllocationConstraints(PROBLEME_HEBDO& problem,
             break;
         }
         if (nbTimeSteps > 0)
-            problem.ResultatsContraintesCouplantes[k].variablesDuales.assign(nbTimeSteps, 0.);
+            problem.ResultatsContraintesCouplantes.emplace(bc, std::vector<double>(nbTimeSteps));
```
flomnes: `emplace` is variadic, you may give it the arguments you'd normally pass to the constructor
```cpp
template< class... Args >
std::pair<iterator, bool> emplace( Args&&... args );
```

It seems that the current code constructs a `std::vector`, then moves it. You can save the `std::move` step in this way.

```suggestion
            problem.ResultatsContraintesCouplantes.emplace(bc, nbTimeSteps, 0));
```

Also make sure to provide 0, otherwise the vector is unitialized.

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
```diff
@@ -329,41 +329,34 @@ void BindingConstraintsRepository::markAsModified() const
...
-        std::copy_if(constraints_.begin(), constraints_.end(), std::back_inserter(out),
-                     [](const auto &bc) {
-                         return bc->isActive();
-                     });
-        activeConstraints_ = std::move(out);
-        return activeConstraints_.value();
-    }
+std::vector<std::shared_ptr<BindingConstraint>> BindingConstraintsRepository::activeConstraints() const
+{
+    if (!activeConstraints_.empty())
```
flomnes, line 334: This is (and was) broken. This caching doesn't work in the following scenario
1. Add an enabled constraint A
2. Call `BindingConstraintsRepository::activeConstraints` => return {A} **OK**
3. Add an enabled constraint B
4. Call `BindingConstraintsRepository::activeConstraints` => return {A} **KO**

I will open a separate issue since this is not really in the scope, and you didn't break anything (the original code is broken).

In short, we need to invalidate the cache whenever
- A constraint is added (KO)
- A constraint is removed (OK)
- A constraint becomes enabled (KO)
- A constraint becomes disabled (KO)

Overall, it would be simpler to have no cache, especially for the 2 last items.

flomnes: @payetvin No immediate action is required from you in this PR.

flomnes (approved):

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
```diff
@@ -336,7 +335,7 @@ void SIM_AllocationConstraints(PROBLEME_HEBDO& problem,
             break;
         }
         if (nbTimeSteps > 0)
-            problem.ResultatsContraintesCouplantes[k].variablesDuales.assign(nbTimeSteps, 0.);
+            problem.ResultatsContraintesCouplantes.emplace(bc, nbTimeSteps, 0));
```
flomnes:
```suggestion
            problem.ResultatsContraintesCouplantes.emplace(bc, nbTimeSteps, 0.);
```

PR #1891, opened 2024-01-19
Add STS level constraint to suspect list for infeasible problem analysis

flomnes:
## Add STS level constraint to suspect list for infeasible problem analysis (slack variables)

In some cases, the data associated to STS objects can lead to infeasible problems.

This includes
    - Large inflows (no overflow is possible)
    - Large negative inflows (level can't be < 0)
    - Steep rule curves (energy can't be transferred fast enough because of PMAX constraints)

Most of these cases can be captured from the level equation, since it ties all of the above data.

### DIFF
--- a/src/solver/infeasible-problem-analysis/constraint.cpp
+++ b/src/solver/infeasible-problem-analysis/constraint.cpp
@@ -5,6 +5,10 @@
 #include <iomanip>
 #include <algorithm>
 
+namespace {
+    const std::string kUnknown;
+}
+
 namespace Antares
 {
 namespace Optimization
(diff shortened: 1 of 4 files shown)


src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -104,6 +108,8 @@ std::string Constraint::getTimeStepInYear() const
     case ConstraintType::fictitious_load:
     case ConstraintType::hydro_reservoir_level:
         return StringBetweenAngleBrackets (mItems.at(mItems.size()-2));
+    case ConstraintType::short_term_storage_level:
```
a-zakir, line 108: Same as other cases

src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -5,6 +5,10 @@
 #include <iomanip>
 #include <algorithm>
 
+namespace {
+    const std::string kUnknown;
+}
```
a-zakir, line 10: Uninitialized

flomnes:
@a-zakir I fixed everything in 99d031371

a-zakir (approved):

PR #1906, opened 2024-01-29
Fixes when installing Simulator locally on Windows

guilpier-code:
These changes were necessary to compile and use **Antares Simulator**, locally and on **Windows**, in 2 ways : 
- the current way (we want no regression)
- So that we can make **Antares Simulator** a library (not exactly usable yet, see next PR) by running the install step.
 
That is the reason why the branch associated to the current PR is based on branch **feature/install_antares_lib** (see PR #1773)

Note that : 
- @JasonMarechal bumped into many problems when trying to make **Antares Simulator** a lib without **minizip** as an cmake internal target. Problems vanish when making  **minizip** an external dependency.
That's why we had to make **minizip** an external lib, and then run config command : 
```cmake
cmake -B build -S src -DDEPS_INSTALL_DIR=D:/Users/pierregui/Desktop/deps-compiled/antares-deps-2.0.2b/rte-antares-deps-Release -DVCPKG_ROOT=D:/Users/pierregui_al/softwares/vcpkg -DCMAKE_PREFIX_PATH="D:/Users/pierregui/Desktop/deps-compiled/ortools-v9.8-rte1.0-with-sirius/install;D:/Users/pierregui/Desktop/deps-compiled/minizip/install" -DVCPKG_TARGET_TRIPLET=x64-windows -DCMAKE_BUILD_TYPE=Release -DBUILD_not_system=OFF -DBUILD_TESTING=ON -DBUILD_TOOLS=ON -DBUILD_UI=OFF
```
please look at the use of **CMAKE_PREFIX_PATH** in this command.
- a cmake > 3.24 was necessary to avoid errors at configuration step
- When making Antares a lib : installation step then passed without problem, running command : 
```cmake
cmake --install build --prefix "D:/Users/pierregui/Desktop/antares-installed"
```
(**--prefix** for installation destination)


flomnes:
No longer useful, many conflicts

PR #1909, opened 2024-02-01
Proposal for StudyVersion

flomnes:


JasonMarechal25:
Do we really need hard coded versions ?
A suggestion to explore would be to have the list of supported version in a configuration file. We could add or remove previous version support without needing to release a new binary.

flomnes:
> Do we really need hard coded versions ?
A suggestion to explore would be to have the list of supported version in a configuration file. We could add or remove previous version support without needing to release a new binary.

We would need to ship configuration files as resources, which is not really done at the moment

src/libs/antares/study/version.cpp
```diff
@@ -103,31 +85,51 @@ StudyVersion::StudyVersion(const std::string& s)
...
+    // if the string doesn't contains a dot it's legacy format
+    if (versionStr.find(".") == std::string::npos)
+        *this = parseLegacyVersion(versionStr);
+    else
+        *this = parseCurrentVersion(versionStr);
+
+    if (isSupported())
+        return true;
+
+    return false;
```
```suggestion
    return isSupported();
```

src/libs/antares/study/version.cpp
```diff
@@ -103,31 +85,51 @@ StudyVersion::StudyVersion(const std::string& s)
...
 }
 
 StudyVersion StudyVersion::unknown()
 {
-    return StudyVersion(0, 0);
+    return StudyVersion();
 }
 
-bool StudyVersion::isVersionSupported(const std::string& version)
+bool StudyVersion::isSupported() const
```
```suggestion
bool StudyVersion::isSupported(bool verbose) const
```

src/libs/antares/study/version.cpp
```diff
@@ -103,31 +85,51 @@ StudyVersion::StudyVersion(const std::string& s)
...
-    logs.error() << "Version: " << version << " not supported";
+    logs.error() << "Version: " << toString() << " not supported";
 
-    if (StudyVersion(version) > latest())
+    if (*this > latest())
     {
-        logs.error() << "Maximum study version supported: " << supportedVersions.back();
+        logs.error() << "Maximum study version supported: " << supportedVersions.back().toString();
         logs.error() << "Please upgrade the solver to the latest version";
     }
```
```cpp
if (verbose) {
 /// logs
}
```

src/libs/antares/study/version.cpp
flomnes: OK, you changed in the meantime, for the better

src/libs/antares/study/version.cpp
```diff
@@ -103,31 +85,52 @@ StudyVersion::StudyVersion(const std::string& s)
...
+    return StudyVersion(major, minor);
+}
+
+bool StudyVersion::fromString(const std::string& versionStr)
+{
+    // if the string doesn't contains a dot it's legacy format
+    if (versionStr.find(".") == std::string::npos)
+        *this = parseLegacyVersion(versionStr);
+    else
+        *this = parseCurrentVersion(versionStr);
```
```suggestion
    if (size_t pos = versionStr.find("."); pos == std::string::npos)
        *this = parseLegacyVersion(versionStr);
    else
        *this = parseCurrentVersion(pos, versionStr);
```

So you don't need to call find(".") twice

PR #1920, opened 2024-02-06
Fix build (CMake, src, headers, etc.)

flomnes:


pet-mit:
Suggestions taken into account in a seperate commit

PR #1922, opened 2024-02-07
Fix `RUN_EXTENDED_TESTS` being overriden at job level

flomnes:
`RUN_EXTENDED_TESTS` is defined at the root level, and is overriden at the job level.

Keep only the definition at the root level.


.github/workflows/windows-vcpkg.yml
```diff
@@ -32,7 +32,6 @@ jobs:
       # Indicates the location of the vcpkg as a Git submodule of the project repository.
       VCPKG_ROOT: ${{ github.workspace }}/vcpkg
       ORTOOLS_DIR: ${{ github.workspace }}/or-tools
-      RUN_EXTENDED_TESTS: ${{ github.event_name == 'schedule'}}
       os: windows-latest
```
flomnes, line 35: See l. 25 for first definition

a-zakir (approved):

PR #1924, opened 2024-02-07
Remove typedef `Impl = ImplememtationType`

flomnes:
Having multiple names for the same type makes the code harder to read. Whenever possible, I prefer 1 name = 1 type.

### DIFF
--- a/src/solver/simulation/solver.h
+++ b/src/solver/simulation/solver.h
@@ -47,10 +47,6 @@ class ISimulation : public Impl
     friend class yearJob<Impl>;
 
 public:
-    //! The real implementation for the simulation
-    typedef Impl ImplementationType;
-
-public:
     //! \name Constructor & Destructor
     //@{
     /*!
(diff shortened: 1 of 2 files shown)


src/solver/simulation/solver.h
```diff
@@ -46,10 +46,6 @@ class ISimulation : public Impl
 {
     friend class yearJob<Impl>;
 
-public:
-    //! The real implementation for the simulation
-    typedef Impl ImplementationType;
```
sylvlecl, line 51: Note that this did not only define an alias to `Impl`, it also allows to use `T::ImplementationType` in template code that would use an implementation of ISimulation as T (similar to `iterator::value_type` for example).

However since the removal does not cause any issue, it does not seem useful in our code base :)

sylvlecl (approved):

src/solver/simulation/solver.h
flomnes: Yes, agreed

flomnes: @sylvlecl At the same time, the caller could just use `Impl` rather than `ISimulation<Impl>::ImplementationType`, right ?

PR #1927, opened 2024-02-07
Add "in memory" implementation for `IResultWriter`

flomnes:
Mostly for debugging purposes, in-memory tests

1. Where a IResultWriter is expected, provide an InMemoryWriter
2. Read results in memory
3. Profit

## TODO
- [x] In Windows, separators are '\\'. I suggest to replace them by '/' for platform-independence.


src/libs/antares/writer/private/in_memory_writer.h
```diff
@@ -0,0 +1,55 @@
...
+    void addEntryFromBuffer(const std::string& entryPath, std::string& entryContent) override;
+    void addEntryFromFile(const std::string& entryPath, const std::string& filePath) override;
+    void flush() override;
+    bool needsTheJobQueue() const override;
+    void finalize(bool verbose) override;
+
+private:
+    std::mutex pMapMutex;
+    // minizip-ng requires a void* as a zip handle.
+    MapType pEntries;
```
sylvlecl: Don't we need a way to retrieve the content ?
- either through getters
- or letting the user provide the object where content is written

flomnes: Yes, I agree. I'll provide a getter for the map in this class, so the caller can write this
```cpp
void f(IResultWriter& writer);
InMemoryWriter writer(...);
f(writer);
auto& map = writer->getMap();
```

What do you think ?

JasonMarechal25 (approved):

PR #1945, opened 2024-02-19
Enable $ delimiter for latex expressions

sylvlecl:
Mathjax does not recognize simple $ signs as delimiter,
by default.
We add it to our configuration, to have consistent
interpretation of latex expressions between sphinx 
and mkdocs.


sylvlecl:
Wrong branch name

PR #1947, opened 2024-02-20
Generate time series for specific thermal cluster ANT-1068

payetvin:
Allow specific thermal clusters TS to be generated via a command line argument.


src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -441,6 +441,27 @@ void ISimulation<ImplementationType>::writeResults(bool synthesis, uint year, ui
...
+        {
+            std::string first = token.substr(0, pos);
+            std::string second = token.substr(pos + 1);
+            pairs.push_back(std::make_pair(first, second));
+        }
+    }
+
+    return pairs;
+}
+
```
flomnes: Add tests on this parser

flomnes (commented):
Suggestion
1. Don't pass names to `generateSpecificThermalTimeSeries`, pass clusters instead. Add a function to translate `names` into `std::vector<ThermalCluster*>` for example.
2. One single function instead of two (`GenerateThermalTimeSeries` and `generateSpecificThermalTimeSeries`)

```cpp
std::vector<ThermalCluster*> clusters;
if (settings.clustersToGen.empty()) {
    clusters = allClusters(study.areas);
} else {
   const auto names = splitStringIntoPairs(settings.clustersToGen, ';', '.');
   clusters = selectClusters(study.areas, names);
}
GenerateThermalTimeSeries(clusters, year, pData.haveToRefreshTSThermal, refreshTSonCurrentYear, pResultWriter);
```

Note that `area = cluster->parentArea`.

flomnes:
Also, the syntax should be as follows `--objects thermal:area.cluster;thermal:area2.cluster2`. Other types should be ignored for now.

flomnes:
Make sure to update **docs/reference-guide/10-command_line.md**.

payetvin:
I don't like the `--objects` name for the command line, as it doesn't need to be in generator mode only in order to be called/to work.
`objects` is fine if we do a separate tool but in the context of calling the solver I don't think it's an option name reflecting what it's actually doing.

src/solver/ts-generator/include/antares/solver/ts-generator/generator.h
```diff
@@ -41,11 +41,16 @@ template<enum Data::TimeSeriesType T>
 bool GenerateTimeSeries(Data::Study& study, uint year, IResultWriter& writer);
 
 bool GenerateThermalTimeSeries(Data::Study& study,
+                               const std::string& clustersToGen,
```
flomnes: In my opinion there should be a collection of `ThermalCluster` here. We need to split parsing & generation into separate functions.

src/solver/ts-generator/thermal.cpp
```diff
@@ -45,7 +46,6 @@ class GeneratorTempData final
 {
 public:
     GeneratorTempData(Data::Study& study,
-                      Solver::Progression::Task& progr,
```
flomnes: Can't progression be given in all cases ? Why remove it ?

src/tests/src/solver/simulation/tests-ts-numbers.cpp
```diff
@@ -751,3 +751,39 @@ BOOST_AUTO_TEST_CASE(check_all_drawn_ts_numbers_are_bounded_between_0_and_nb_of_
 	BOOST_CHECK(thermalTsNumber < thermalNumberOfTs);
     BOOST_CHECK_LT(binding_constraints_TS_number, binding_constraints_number_of_TS);
 }
+
+BOOST_AUTO_TEST_CASE(split_string_ts_cluster_gen)
```
flomnes, line 755: You've tested the string -> token mapping. You need to test also the token -> ThermalCluster lookup. Create a study with a few clusters and test that they are found if they exist, not found if they don't, etc.

flomnes (changes requested):

src/solver/ts-generator/thermal.cpp
payetvin: I thought it was only used in the GUI
only add logs like this in the solver
`[solver][progress] task 0 thermal, year: 0, 100`

src/solver/ts-generator/include/antares/solver/ts-generator/generator.h
payetvin: where should the code doing the parsing goes ?
Plus we don't have the same checks/behavior if we generate all clusters or not, thus we need to know which mode we're in

`if (cluster->doWeGenerateTS(globalThermalTSgeneration) && refreshTSonCurrentYear)`

src/tests/src/solver/simulation/tests-ts-numbers.cpp
payetvin: already tested in
- `libs/antares/study/test_study.cpp`
- `libs/antares/study/thermal-price-definition/thermal-price-definition.cpp`

`findInAll()`

flomnes (changes requested):
More like this

```cpp
std::vector<ThermalCluster*> clusters = getClusters(study.areas, settings.clustersToGen, /* ... */); // <= Parse here
GenerateThermalTimeSeries(study, clusters, year, /* ... */);
```

src/solver/ts-generator/thermal.cpp
```diff
@@ -596,7 +597,81 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+    }
+
+    return clusters;
+}
+
 bool GenerateThermalTimeSeries(Data::Study& study,
+                               const std::string& clustersToGen,
                                uint year,
                                bool globalThermalTSgeneration,
                                bool refreshTSonCurrentYear,
```
flomnes: Move `getClustersToGen` up one level

```suggestion
bool GenerateThermalTimeSeries(const std::vector<ThermalCluster*> clusters, uint year);
// Call (solver.hxx)
auto clusters = getClustersToGen(/* ... */);
GenerateThermalTimeSeries(clusters, year);
```

src/solver/ts-generator/thermal.cpp
```diff
@@ -596,11 +597,83 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+            for (auto cluster : area.thermal.list.all())
+                if (cluster->doWeGenerateTS(globalThermalTSgeneration) && refreshTSonCurrentYear)
+                    clusters.push_back(cluster.get());
+        });
+    }
+    else
+    {
+        const auto names = splitStringIntoPairs(clustersToGen, ';', '.');
+
+        for (const auto& [areaName, clusterName] : names)
```
flomnes:
```suggestion
        for (const auto& [areaID, clusterID] : names)
```

These lookup functions use IDs, see **list.cpp** and **cluster_list.cpp**.

src/solver/ts-generator/include/antares/solver/ts-generator/generator.h
```diff
@@ -41,11 +41,18 @@ template<enum Data::TimeSeriesType T>
...
 
 bool GenerateThermalTimeSeries(Data::Study& study,
+                               std::vector<Data::ThermalCluster*> clusters,
                                uint year,
-                               bool globalThermalTSgeneration,
-                               bool refresh,
-                               IResultWriter& writer);
+                               Solver::IResultWriter& writer);
 
+std::vector<std::pair<std::string, std::string>> splitStringIntoPairs (const std::string& s,
```
guilpier-code, line 48: Function **splitStringIntoPairs** is a low level and helper function that should not be part of generator interface. In y opinion, it should be declared and defined in separate **.h/.cpp** pair of source file, and the **.h** should be included in **thermal.cpp**

guilpier-code: **splitStringIntoPairs** returns a vector of pairs. It could return a **std::map** instead, more natural.
There is probably a reason for that.
**Question** : why do we not return a **std::map** ?

src/solver/ts-generator/thermal.cpp
```diff
@@ -596,11 +597,83 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+
+std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
+                                                    const std::string& clustersToGen,
+                                                    bool globalThermalTSgeneration,
+                                                    bool refreshTSonCurrentYear)
+{
+
+    std::vector<Data::ThermalCluster*> clusters;
+
+    if (clustersToGen.empty()) // generate TS for all clusters
```
guilpier-code, line 633: In my opinion, this function (**getClustersToGen**) should be simpler.
Currently, it knows for which reason a cluster should have its TS generated. It can be because : 
- the user set the clusters input parameters so that : 
  - all thermal clusters TS should be generated
  - for some of them, TS generation was forced (see **gen-ts** parameter)
- the user specified a command line option for some thermal cluster.

This  function should be oblivious to the reason why we generate TS.
Eventually, it should look like : 
```C++
std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
                                                    bool refreshTSonCurrentYear)
{
    std::vector<Data::ThermalCluster*> clusters;
    areas.each([&clusters, &refreshTSonCurrentYear](Data::Area& area) 
    {
         for (auto cluster : area.thermal.list.all())
             if (cluster->doWeGenerateTS(refreshTSonCurrentYear)
                 clusters.push_back(cluster.get());
    });
    return clusters;
}
```
So, at some point way before this one in the execution (I mean right after input data was read), we should aggregate thermal clusters data from files and from command line.

flomnes (changes requested):
Create an in-memory study, call the main with `--generator-only --etc` (see src/solver/main.cpp)
```cpp
TEST(thermal_gen) {
int argc = /* */;
const char* argv[] = {"antares-solver", "--generator-only", "--cluster-gen", "area1.cluster1"};
        // Getting real UTF8 arguments
        IntoUTF8ArgsTranslator toUTF8ArgsTranslator(argc, argv);
        std::tie(argc, argv) = toUTF8ArgsTranslator.convert();
        Antares::Solver::Application application;
        application.prepare(argc, argv);
        application.execute();
}
```
}

guilpier-code (changes requested):

src/solver/ts-generator/include/antares/solver/ts-generator/generator.h
flomnes: @guilpier-code What template arguments for the `std::map` ?

src/solver/ts-generator/thermal.cpp
```diff
@@ -596,11 +597,83 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+    std::string token;
+
+    while (std::getline(ss, token, delimiter1))
+    {
+        size_t pos = token.find(delimiter2);
+        if (pos != std::string::npos)
+        {
+            std::string first = token.substr(0, pos);
+            std::string second = token.substr(pos + 1);
+            pairs.push_back(std::make_pair(first, second));
```
flomnes:
```suggestion
            pairs.push_back({first, second});
```

src/solver/ts-generator/thermal.cpp
```diff
@@ -596,11 +597,83 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+    std::stringstream ss(s);
+    std::string token;
+
+    while (std::getline(ss, token, delimiter1))
+    {
+        size_t pos = token.find(delimiter2);
+        if (pos != std::string::npos)
+        {
+            std::string first = token.substr(0, pos);
+            std::string second = token.substr(pos + 1);
```
flomnes: first => begin
second => end

src/solver/ts-generator/thermal.cpp
```diff
@@ -596,11 +597,83 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+            pairs.push_back(std::make_pair(first, second));
+        }
+        else
+        {
+            logs.warning() << "Error with area and cluster: " << token;
+            logs.warning() << "Correct format: \"area1.cluster1;area2.cluster2\"";
```
flomnes: This function splits strings, it has no clue about clusters and areas

src/solver/ts-generator/thermal.cpp
flomnes: It's not so easy because of the spec
- If the user provides a `--clusters-to-gen` argument, we ignore the local thermal cluster parameter
- If the user doesn't provide this argument, we use the local cluster parameter instead.

guilpier-code: ok. So what you say it that we don't do that because it's too much work ?

src/solver/ts-generator/include/antares/solver/ts-generator/generator.h
payetvin: we can have [area1, cluster1] [area1, cluster2], map won't work

src/solver/ts-generator/thermal.cpp
guilpier-code: **Question** on function **getClustersToGen** : 
Something is a bit strange for me : a soon as user gives one cluster as a command line argument, the policy of TS generation of all other clusters (even clusters of other areas) is removed, and we don't generate any TS for these other clusters.
> If the user provides a `--clusters-to-gen` argument, we ignore the local thermal cluster parameter

For sure this was intended, but still it is strange.

guilpier-code: Ultimately, if we do as I recommend, a call to **getClustersToGen** is not necessary in **solver.hxx**.
Filling a temporary vector with clusters we need to generate TS for is not necessary.
We could just select them in **GenerateThermalTimeSeries**.

guilpier-code (changes requested):

src/solver/ts-generator/include/antares/solver/ts-generator/generator.h
guilpier-code: A **std:map** is nothing but a vector of pairs (more or less).
But you're right : with a **std::map**, we can have identical keys (area names), unless we make the cluster names the keys, which can be identical.
The **std::map** does not seem so natural now.

src/solver/ts-generator/thermal.cpp
flomnes: The code looks more readable if we
1. Pick what clusters to run the Generator on
2. Run the generator
We don't want to mix these, in particular we don't want to parse anything in **GenerateThermalTimeSeries**

flomnes: > ok. So what you say it that we don't do that because it's too much work ?

No, this is the intended behavior. Eventually the local flag will disappear (it has no use if we don't do in-memory generation).

src/solver/ts-generator/thermal.cpp
```diff
@@ -596,11 +597,80 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+
+std::vector<std::pair<std::string, std::string>> splitStringIntoPairs (const std::string& s,
+                                                                       char delimiter1,
+                                                                       char delimiter2)
+{
+    std::vector<std::pair<std::string, std::string>> pairs;
+    std::stringstream ss(s);
+    std::string token;
+
+    while (std::getline(ss, token, delimiter1))
```
sylvlecl, line 609: You may iterate on a range to simplify
```cpp
for (auto token: std::views::split(s, delimiter1) {
    ...
}
```

src/solver/misc/include/antares/solver/misc/options.h
```diff
@@ -48,8 +48,11 @@ class Settings
     int ignoreWarningsErrors = 0;
     //! Ignore constraints
     bool ignoreConstraints = false;
-    //!
+
+    //! Run the TS generator only
     bool tsGeneratorsOnly = false;
+    std::string clustersToGen = "";
```
sylvlecl, line 54: It would be best to have the vector of pairs here (actually better vector of small structs), and have the parsing done at options parsing time.
This way:
1. we will catch errors early
2. the business logic will not be cluttered with parsing code

src/solver/ts-generator/thermal.cpp
sylvlecl: The separation between cluster selection and generation is a big improvement !

The best target would be to have:
1. the selection logic based on the executable option in the executable code
2. the selection logic based on the study parameters in the simulation code (but we will remove it at some point)

But it does not seem possible with the current implementation where the executable calls the whole simulation function ... 
However in the other PR with a new tool, I would advise to to this separation.

src/solver/misc/include/antares/solver/misc/options.h
payetvin: The parsing should be done in the separated tool
We need to keep it here for the moment to support the in memory mode

src/solver/ts-generator/thermal.cpp
payetvin: token is not a string in this context and needs to be transformed in order to be used, increasing the complexity

flomnes:
See #1967

PR #1918, opened 2024-02-05
Remove version from exe

a-zakir:
This Pr removes project version from executables name (e.g antares-X.Y-solver.exe ->>antares-solver.exe)
allow to release from a specific branch/commit SHA

PR #1956, opened 2024-02-22
add branch selection in release form

a-zakir:
add branch selection in release form

### DIFF
--- a/.github/workflows/build-userguide.yml
+++ b/.github/workflows/build-userguide.yml
@@ -12,6 +12,9 @@ on:
       run-tests:
         required: true
         type: string
+      target_branch:
+        required: true
+        type: string
 
 env:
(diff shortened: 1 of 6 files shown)


JasonMarechal25:
Overall ok with it. Is it possible to add some documentation about the difference between the combobox "Use workflow from" and the new input "target branch". At least a comment the workflow, maybe a description that can be seen on github when starting a workflow

a-zakir:
> Overall ok with it. Is it possible to add some documentation about the difference between the combobox "Use workflow from" and the new input "target branch". At least a comment the workflow, maybe a description that can be seen on github when starting a workflow

ok but this pr is no more needed (see merged [1918](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1918)

PR #1961, opened 2024-02-23
Add unit tests for legacy workflow

flomnes:


src/solver/optim/impl/LinearProblemImpl.cpp
```diff
@@ -49,7 +49,10 @@ MPConstraint& LinearProblemImpl::addBalanceConstraint(string name, double bound,
 
 MPConstraint& LinearProblemImpl::getConstraint(string name)
 {
-    return *mpSolver->LookupConstraintOrNull(name);
+    MPConstraint* cnt = mpSolver->LookupConstraintOrNull(name);
+    if (!cnt)
+        throw;
```
pet-mit, line 55: you can add a TODO so we don't forget to put a message here

src/solver/optim/impl/LinearProblemImpl.cpp
```diff
@@ -65,6 +68,8 @@ void LinearProblemImpl::setMinimization(bool isMinim)
 MipSolution LinearProblemImpl::solve()
 {
     auto status = mpSolver->Solve();
+    // std::string str;
+    // mpSolver->ExportModelAsLpFormat(false, &str);
```
pet-mit, line 74:
```suggestion
    // TODO: remove this
    // std::string str;
    // mpSolver->ExportModelAsLpFormat(false, &str);
```

pet-mit (approved):

PR #1966, opened 2024-02-26
Fix adq patch regression

guilpier-code:
This PR intends to fix [gopro ticket 1155](https://gopro-tickets.rte-france.com/browse/ANT-1155).

Note that : 
- Note a special branch **branch-v8.8.0** was created so that this PR can be based on it. It seems that basing a PR on a tag is impossible.
- the fix was derived from the first commit where the regression occurs (which is somewhere in between **v8.7.0** and **v8.8.0**). Apparently there are conflict with tag v8.8.0.
- This fix is a **dirty quick hack** : its purpose is to demonstrate the cause of an **adequacy patch** regression, introduced when removing a global variable. This PR mainly consists in restoring a call to function **SIM_RenseignementProblemeHebdo**.
- Carrying data required by this call turns out to be difficult, the resulting code is uglier than it was : optimization abstract method **solve(...)** has now more arguments that are not used in all derived implementation classes, specially the default class. 
- If we decide to integrate these changes to **develop** branch, we need remove the previous identified flaws. That said, we may not need to apply this fix to develop : @JMJ-rte opinion is that adequacy patch isolation first step should be skipped, so we won't need a new call to **SIM_RenseignementProblemeHebdo**
- We validated this fix by running the study supplied in **ticket 1155** while asking mps files printings, and comparing them (a use of **md5sum** is enough).
- This regression was not caught sooner because there seems to be no associated test. We should have a test validating this.

### DIFF
--- a/src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
+++ b/src/solver/optimisation/adequacy_patch_local_matching/adequacy_patch_weekly_optimization.cpp
@@ -30,6 +30,7 @@
 #include "../simulation/simulation.h"
 #include "../simulation/adequacy_patch_runtime_data.h"
 #include "antares/study/fwd.h"
+#include "../simulation/common-eco-adq.h"
 
 using namespace Antares::Data::AdequacyPatch;
 using Antares::Constants::nbHoursInAWeek;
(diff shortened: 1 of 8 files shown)


a-zakir:
temp change for windows ci:
[python setup](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/c2b71e9a28c9e75b8ebeedb10bb1dc4e703e6241/.github/workflows/windows-vcpkg.yml#L88) ->
 
```
    - name: Setup Python 3.12
      uses: actions/setup-python@v4
      id: setup-python
      with:
        architecture: 'x64'
        python-version: '3.12'
```
and 
```cmake
-DPython3_EXECUTABLE="${{ env.Python3_ROOT_DIR }}/python.exe"
```

flomnes:
This feature will be removed.

PR #1968, opened 2024-02-29
Remove direct use of Sirius solver : first removal

guilpier-code:
This PR intends to be a first work in order to remove the direct call of Sirius as a solver.
We want to keep the use of Sirius, but only called through **or-tools** interface.
The associated **gopro** ticket is [this one](https://gopro-tickets.rte-france.com/browse/ANT-1152).
Here we intend to : 
- remove the use of command line option **--use-ortools**
- rename current command line option **--ortools-solver** into **--solver**
- handle direct consequences of these changes.

So this PR does not intend to be thorough : these first changes leave remaining dead code to be removed 
and renaming to be done. This further changes will be done in a coming PR.

**Remaining to be done** : 
- or-tools (even if we use Sirius behind it) does not print MPS for named problems. 
  So we left the code that print these correctly. We'll have to find a solution for this.
  We could leave the responsibility of printing these named MPS to Antares code (as it did when Sirius was run alone) for a moment
- We should remove call to Sirius alone in the legacy GUI (not long).
- What else ?


guilpier-code:
Note that, at this stage, some tests fail.
These tests are the **named mps tests**.
These failure can easily be explained : calling **Sirius** behind **or-tools** dumps different **mps** files : in these **mps** files, the names of the optimization problem's variables/constraints are not updated, like they are when calling directly **Sirius**.
If we want these tests to pass, we should first update these names, and as a first try, we may rebuild the weekly problem every week in a further PR (which one ?).
For now, we **temporarily** remove the run of **named mps tests**, to see if other tests pass.

guilpier-code:
This PR is replaced with #2450

PR #1969, opened 2024-02-29
Correct the ts count for no gen thermal clusters ANT-1303

payetvin:


flomnes (changes requested):
Wrong code. We need something like this
```cpp
for (auto &cluster : clusters) {
  uint tsGenCount = cluster->tsGenBehavior == LocalTSGenerationBehavior::forceNoGen ? cluster->series.timeSeries.width : get_tsGenCount(study);
}
```

Otherwise one cluster may affect the next one, which is an unwanted side-effect.

flomnes (changes requested):
Add unit tests ?

flomnes (approved):

src/tests/src/libs/antares/study/scenario-builder/test-sc-builder-file-read-line.cpp
```diff
@@ -417,4 +417,25 @@ BOOST_FIXTURE_TEST_CASE(binding_constraints_group_groupTest__Load_TS_4_for_year_
...
+    tsdata.setTSnumber(thCluster_12.get(), 5, 32); //out of bounds
+
+    study->parameters.nbTimeSeriesThermal = 1;
+    thCluster_12->tsGenBehavior = LocalTSGenerationBehavior::forceNoGen;
+    thCluster_12->series.timeSeries.resize(30, 8760);
+
+    tsdata.apply(*study);
+
+    BOOST_CHECK(thCluster_12->series.timeseriesNumbers[0][2] == 21);
+    BOOST_CHECK(thCluster_12->series.timeseriesNumbers[0][5] == 0);
```
flomnes:
```suggestion
    BOOST_CHECK_EQUAL(thCluster_12->series.timeseriesNumbers[0][2], 21);
    BOOST_CHECK_EQUAL(thCluster_12->series.timeseriesNumbers[0][5], 0);
```

PR #1972, opened 2024-03-04
Restore windows debug info format, for debug build

sylvlecl:
/Zi flag has been removed from compiler flags so that
ccache works on windows, but this has caused a regression
from the developer point of view.

This pull request restores the flag only for debug builds,
while at the same time simplifying a little the flag definition.

### DIFF
--- a/src/cmake/common-settings.cmake
+++ b/src/cmake/common-settings.cmake
@@ -99,22 +99,20 @@ endif()
 
 
 if(MSVC)
-	#Compile option must be updated if VPCKG static is used (can't mix different compile option)
+	set(CMAKE_C_FLAGS   "${COMMON_MSVC_FLAGS}")
+	set(CMAKE_CXX_FLAGS "${COMMON_MSVC_FLAGS}")
+
+	# If other libraries have been linked statically with C runtime, we need to keep consistent with it
 	if("${VCPKG_TARGET_TRIPLET}" STREQUAL "x64-windows-static" OR "${VCPKG_TARGET_TRIPLET}" STREQUAL "x86-windows-static")
-		set(CMAKE_C_FLAGS_DEBUG   "${COMMON_MSVC_FLAGS} /MTd /GR /Ot /Od /EHsc /RTC1")
-		set(CMAKE_CXX_FLAGS_DEBUG "${COMMON_MSVC_FLAGS} /MTd /GR /Ot /Od /EHsc /RTC1 /fp:except")
+		set(CRT_LIBRARY_DEBUG "/MTd")
+		set(CRT_LIBRARY_RELEASE "/MT")
     else()
-		set(CMAKE_C_FLAGS_DEBUG   "${COMMON_MSVC_FLAGS} /MDd /GR /Ot /Od /EHsc /RTC1")
-		set(CMAKE_CXX_FLAGS_DEBUG "${COMMON_MSVC_FLAGS} /MDd /GR /Ot /Od /EHsc /RTC1 /fp:except")
+		set(CRT_LIBRARY_DEBUG "/MDd")
+		set(CRT_LIBRARY_RELEASE "/MD")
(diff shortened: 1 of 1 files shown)


sylvlecl:
**Additional information about the changes**

- Removed the part which assigned `CMAKE_<LANG>_FLAGS_<CONFIG>` to `CMAKE_<LANG>_FLAGS`:
this is redundant with cmake behaviour which actually already appends them, and actually it defeats the purpose of those variables in multi-config build systems (such as MSVC), since the config is not necessarily known at configure time.

- De-duplicated checks on VCPKG triplet, to remove code duplications

src/cmake/common-settings.cmake
```diff
@@ -99,23 +99,21 @@ endif()
...
+		set(CRT_LIBRARY_DEBUG "/MTd")
+		set(CRT_LIBRARY_RELEASE "/MT")
     else()
-		set(CMAKE_C_FLAGS_DEBUG   "${COMMON_MSVC_FLAGS} /MDd /GR /Ot /Od /EHsc /RTC1")
-		set(CMAKE_CXX_FLAGS_DEBUG "${COMMON_MSVC_FLAGS} /MDd /GR /Ot /Od /EHsc /RTC1 /fp:except")
+		set(CRT_LIBRARY_DEBUG "/MDd")
+		set(CRT_LIBRARY_RELEASE "/MD")
 	endif()
 
+	set(CMAKE_C_FLAGS_DEBUG   "${CRT_LIBRARY_DEBUG} /Zi /GR /Ot /Od /EHsc /RTC1")
```
sylvlecl, line 114: Restored `Zi` flag here and in line below

src/cmake/common-settings.cmake
```diff
@@ -220,14 +213,3 @@ macro(executable_strip TARGET)
 		endif()
 	endif()
 endmacro()
-
-if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
```
sylvlecl, line 224: This part was unnecessary: cmake already appends the config-specific flags to common flags, see [doc](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_CONFIG.html#variable:CMAKE_%3CLANG%3E_FLAGS_%3CCONFIG%3E)

src/cmake/common-settings.cmake
```diff
@@ -99,23 +99,21 @@ endif()
...
 if(MSVC)
-	#Compile option must be updated if VPCKG static is used (can't mix different compile option)
-	if("${VCPKG_TARGET_TRIPLET}" STREQUAL "x64-windows-static" OR "${VCPKG_TARGET_TRIPLET}" STREQUAL "x86-windows-static")
-		set(CMAKE_C_FLAGS_DEBUG   "${COMMON_MSVC_FLAGS} /MTd /GR /Ot /Od /EHsc /RTC1")
-		set(CMAKE_CXX_FLAGS_DEBUG "${COMMON_MSVC_FLAGS} /MTd /GR /Ot /Od /EHsc /RTC1 /fp:except")
-    else()
-		set(CMAKE_C_FLAGS_DEBUG   "${COMMON_MSVC_FLAGS} /MDd /GR /Ot /Od /EHsc /RTC1")
-		set(CMAKE_CXX_FLAGS_DEBUG "${COMMON_MSVC_FLAGS} /MDd /GR /Ot /Od /EHsc /RTC1 /fp:except")
-	endif()
+	set(CMAKE_C_FLAGS   "${COMMON_MSVC_FLAGS}")
```
sylvlecl, line 102: This ensures those flags will be used for all builds, including debug an release.
Reference to `COMMON_MSVC_FLAGS` in config-specific flags are removed hereafter, since they are redundant.

flomnes (approved):

a-zakir (approved):

PR #1975, opened 2024-03-04
Add tests for ts generator tool

payetvin:


src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,29 @@
...
+        self.path = Path(path).resolve()
+        self.args = args
+
+    def run_and_compare(self, exe):
+        res = subprocess.run([exe, "--thermal=" + self.args, self.path])
+        assert (res.returncode == 0), "The exec failed for study: " + str(self.path)
+
+        out_path = self.path / find_dated_output_folder(self.path) / "ts-generator"
+        ref_path = self.path / "output" / "reference" / "ts-generator"
+        res = subprocess.run(["diff", "-bur", out_path, ref_path])
```
a-zakir: sadly 'diff' is not available on windows by default

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+import subprocess
+import sys
+
+from pathlib import Path
+from utils.find_output import find_dated_output_folder
+
+class StudyReference(object):
```
a-zakir: ok

src/tests/run-study-tests/ts-generator/ts-generator.py
a-zakir: do you need to compare folders? what if later, new outputs are added ?

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,29 @@
...
+        res = subprocess.run(["diff", "-bur", out_path, ref_path])
+        assert (res.returncode == 0), "Wrong results for study: " + str(self.path)
+
+
+study_list = []
+study_list.append(StudyReference("../resources/batches/valid-tsgenerator/007 Thermal fleet - Fast unit commitment", "area.base;area.peak"))
+
+@pytest.mark.tsgenerator
+def test1(exe_tsgenerator_path):
+    study_list[0].run_and_compare(exe_tsgenerator_path)
```
a-zakir:
```suggestion
   for (study in study_list):
       study.run_and_compare(exe_tsgenerator_path)
```

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,29 @@
...
+        ref_path = self.path / "output" / "reference" / "ts-generator"
+        res = subprocess.run(["diff", "-bur", out_path, ref_path])
+        assert (res.returncode == 0), "Wrong results for study: " + str(self.path)
+
+
+study_list = []
+study_list.append(StudyReference("../resources/batches/valid-tsgenerator/007 Thermal fleet - Fast unit commitment", "area.base;area.peak"))
+
+@pytest.mark.tsgenerator
+def test1(exe_tsgenerator_path):
```
a-zakir:
```suggestion
def test_tsgenerator(exe_tsgenerator_path):
```

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,29 @@
...
+        assert (res.returncode == 0), "The exec failed for study: " + str(self.path)
+
+        out_path = self.path / find_dated_output_folder(self.path) / "ts-generator"
+        ref_path = self.path / "output" / "reference" / "ts-generator"
+        res = subprocess.run(["diff", "-bur", out_path, ref_path])
+        assert (res.returncode == 0), "Wrong results for study: " + str(self.path)
+
+
+study_list = []
+study_list.append(StudyReference("../resources/batches/valid-tsgenerator/007 Thermal fleet - Fast unit commitment", "area.base;area.peak"))
```
a-zakir: ok the study path could be an arg for the test suite. may be specifying just the root folder where all studies are and loop over subdirectories

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,31 @@
...
+
+from pathlib import Path
+from utils.find_output import find_dated_output_folder
+from parse_studies.look_for_studies import look_for_studies
+
+
+def run_and_compare(exe, path):
+    clusterToGenFile = open(path / "clustersToGen.txt", 'r')
+    line = clusterToGenFile.readline().rstrip() # remove new line char
+    clusterToGenFile.close()
```
flomnes: Use with..as https://docs.python.org/3/reference/compound_stmts.html#with

flomnes (approved):

guilpier-code:
**Questions** :
Why do we choose to reuse the **python scripts** (currently running solver over study directories and comparing results produced by this runs) in order to tests a new side tool executable ?
I guess there are some reuse considerations behind this. 
How far should we use this python code to test side tool executable ?  
If we want to make the python code evolve to test any side tools executable, should we not think of making the code more generic ?

src/tests/CMakeLists.txt
```diff
@@ -84,6 +84,13 @@ if (BUILD_TOOLS)
         COMMAND Python3::Interpreter -m pytest -m kirchhoff --exe_kirchhoff_path=$<TARGET_FILE:antares-kirchhoff-constraints-builder> kirchhoff_pytest.py
         WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/kirchhoff-cbuilder"
         )
+
+      add_test(
+        NAME ts-generator
+        COMMAND Python3::Interpreter -m pytest -m json --solver-path=$<TARGET_FILE:antares-ts-generator> --ts-generator
```
guilpier-code: **--solver-path** should be renamed **--exe-path**, as we want to generalize this script.
Besides, why do we need **--ts-generator** ? 
This information is already contained in : 
**--solver-path=$\<TARGET_FILE:antares-ts-generator\>**
and we want to make python script generic, that is ignorant of the executable given as a command line argument.

payetvin:
`run-study-tests/ts-generator` contains a different way of testing, without integrating to the main test framework

src/tests/run-study-tests/actions_on_study/study_run.py
```diff
@@ -30,6 +33,13 @@ def run(self):
             command.append('--named-mps-problems')
         if self.parallel:
             command.append('--force-parallel=4')
+
```
guilpier-code, line 33: Function **run()** should **not** have the responsibility  to unpack command line options, especially if we want to make this code more generic, that is make it oblivious to the executable it runs.
We should have only : 
```python
class command_runner:
    def __init__(self, cmd_to_run):
       self.cmd_to_run = cmd_to_run
        self.raise_exception_on_failure = True
        self.return_code = 0

    def run(self):  
        process = subprocess.Popen(self.cmd_to_run, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
        process.communicate()

        self.return_code = process.returncode

        if not self.raise_exception_on_failure:
            return

        check(self.return_code == 0, "Solver returned error")
...
``` 
The command to be run should be passed in the current class constructor, and so should be built before instantiating a **study_run** (here renamed into **command_runner**)

src/tests/run-study-tests/check_on_results/compare_ts_files.py
```diff
@@ -0,0 +1,42 @@
...
+
+from check_decorators.print_name import printNameDecorator
+from check_on_results.check_general import check_interface
+from utils.find_reference import find_reference_folder
+from utils.find_output import find_dated_output_folder, find_simulation_folder
+from utils.assertions import check
+import sys
+
+@printNameDecorator
+class compare_ts_files(check_interface):
```
guilpier-code: Here we have many similarities with **compare_mps_files.py**, we could find a way to extract common code.

src/tests/run-study-tests/ts-generator/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
```
guilpier-code: Previous file (**src/tests/run-study-tests/ts-generator/CMakeLists.txt**) is empty. Please remove it from repo.

src/tests/run-study-tests/ts-generator/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
```
guilpier-code: We should remove this file, because its content can already be found in **src/tests/run-study-tests/pytest.ini**, and there's not need to separate this config information from the rest, is it ?

src/tests/run-study-tests/ts-generator/conftest.py
```diff
@@ -0,0 +1,8 @@
...
+
```
guilpier-code: Why do we have 2 **conftest.py** source files ?
As **--exe_tsgenerator_path** should become **--exe_path** and give the path to a generic executable, this information should be moved to **src/tests/run-study-tests/conftest.py**

src/tests/CMakeLists.txt
guilpier-code: **-m json** option means that we intend to execute tests marked with the **json** flag (meaning tests prefixed with **@pytest.mark.json**).
But apparently, you prefixed your new tests with the new marker **@pytest.mark.tsgenerator**.
So I'm wondering : how can it work like this ?

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,31 @@
+import pytest
```
guilpier-code: Do we need a new directory **ts-generator** to contain config information and test running ?
Indeed, as already suggested, config information should moved to already existing config files, and test running moved where other test running top script are.

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,31 @@
...
+import subprocess
+import sys
+
+from pathlib import Path
+from utils.find_output import find_dated_output_folder
+from parse_studies.look_for_studies import look_for_studies
+
+
+def run_and_compare(exe, path):
```
guilpier-code: **path** to be renamed in **study_path**, more specific.

src/tests/run-study-tests/ts-generator/ts-generator.py
```diff
@@ -0,0 +1,31 @@
...
+    clusterToGenFile = open(path / "clustersToGen.txt", 'r')
```
guilpier-code: Here we read clusters we want to generate TS for.
These clusters names are contained in a new text file located in the study.
Why not use the already existing **json** file to hold this piece of information.

src/tests/run-study-tests/fixtures.py
```diff
@@ -40,8 +40,8 @@ def resultsRemover(study_path):
     return results_remover(study_path)
 
 @pytest.fixture
-def simulation(study_path, solver_path, use_ortools, ortools_solver, named_mps_problems, parallel):
-    return study_run(study_path, solver_path, use_ortools, ortools_solver, named_mps_problems, parallel)
+def simulation(study_path, solver_path, use_ortools, ortools_solver, named_mps_problems, parallel, ts_generator):
+    return study_run(study_path, solver_path, use_ortools, ortools_solver, named_mps_problems, parallel, ts_generator)
```
guilpier-code: - class **study_run** replaced by **command_runner** (see **src/tests/run-study-tests/actions_on_study/study_run.py**'s review).
- not sure we need **simulation** fixture anymore. If still needed, we should rename it into **executable_run**
- This is probably the location to unpack the executable command line options by calling a function or a requiring this service to a factory object (see **src/tests/run-study-tests/actions_on_study/study_run.py**)

guilpier-code (dismissed):

PR #1977, opened 2024-03-05
Remove Yuni::Bind from src, excluding src/ui

flomnes:
Remove `Yuni::Bind`, use `std::function`, `std::bind` instead.

Don't do it in src/ui (deprecated component).

### DIFF
--- a/src/ui/simulator/toolbox/components/map/settings.h
+++ b/src/ui/simulator/toolbox/components/map/settings.h
@@ -25,6 +25,7 @@
 #include <wx/bitmap.h>
 #include <wx/image.h>
 #include <list>
+#include <cstdint>
 
 namespace Antares
 {
--- a/src/ui/simulator/windows/cleaner.cpp
+++ b/src/ui/simulator/windows/cleaner.cpp
@@ -57,7 +57,9 @@ public:
         {
             std::lock_guard locker(pParent->pMutex);
             if (pParent->pInfos)
-                pParent->pInfos->onProgress = nullptr;
+                pParent->pInfos->onProgress = []([[maybe_unused]] unsigned int) -> bool {
+                    return false;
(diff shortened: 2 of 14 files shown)


src/ui/simulator/toolbox/components/map/settings.h
```diff
@@ -25,6 +25,7 @@
 #include <wx/bitmap.h>
 #include <wx/image.h>
 #include <list>
+#include <cstdint>
```
flomnes, line 28: See `uint16_t` below

src/ui/simulator/windows/cleaner.cpp
```diff
@@ -57,7 +57,7 @@ class CleaningThread final : public Yuni::Thread::IThread
         {
             std::lock_guard locker(pParent->pMutex);
             if (pParent->pInfos)
-                pParent->pInfos->onProgress.unbind();
+                pParent->pInfos->onProgress = nullptr;
```
flomnes: Need to check what happens if we call an unbinded Yuni::Bind function. Calling a nullptr std::function is UB (?)

flomnes: It seems that calling a non-bound `Yuni::Bind` is a no-op
```cpp
// Bind: unbind
template<class R>
inline void Bind<R(), void>::unbind()
{
    pHolder = new Private::BindImpl::None<R()>(); // unbind
}

template<class R>
inline R Bind<R(), void>::operator()() const
{
    return pHolder->invoke();
}
```

flomnes: Done, will use `[]{}`, see https://stackoverflow.com/a/2982599

JasonMarechal25 (approved):

PR #1981, opened 2024-03-07
Add OR-Tools solver option for `batchrun` tool

flomnes:
Allow the **batchrun** tool to forward OR-Tools options to antares-solver.

### DIFF
--- a/src/tools/batchrun/main.cpp
+++ b/src/tools/batchrun/main.cpp
@@ -33,6 +33,8 @@
 #include <antares/args/args_to_utf8.h>
 #include <antares/antares/version.h>
 #include <antares/locale/locale.h>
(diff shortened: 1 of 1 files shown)


src/tools/batchrun/main.cpp
```diff
@@ -80,6 +82,7 @@ int main(int argc, char* argv[])
 
     // options
     String optInput;
+    String ortoolsSolver;
```
payetvin: use std::string

payetvin (approved):

src/tools/batchrun/main.cpp
flomnes: done

payetvin (approved):

PR #1984, opened 2024-03-12
Add missing parameter to command built by batchrun for OR-Tools

flomnes:


src/tools/batchrun/main.cpp
```diff
@@ -293,7 +293,7 @@ int main(int argc, char* argv[])
             if (optForceParallel)
                 cmd << " --force-parallel=" << *optForceParallel;
 	    if (ortoolsUsed)
- 	        cmd << " --ortools-solver=" << ortoolsSolver;
+ 	        cmd << " --use-ortools --ortools-solver=" << ortoolsSolver;
```
sylvlecl: We should check that ortools solver is not empty, otherwise the command will not be valid

sylvlecl (changes requested):

src/tools/batchrun/main.cpp
```diff
@@ -190,166 +189,179 @@ int main(int argc, char* argv[])
...
-            logs.fatal() << "The solver has not been found. specify --solver=" << solver;
-            return EXIT_FAILURE;
+            const auto availableSolvers = getAvailableOrtoolsSolverName();
+            if (auto it
+                = std::find(availableSolvers.begin(), availableSolvers.end(), ortoolsSolver);
+                it == availableSolvers.end())
+            {
+                logs.error() << "Please specify a solver using --ortools-solver. Available solvers "
+                             << availableOrToolsSolversString() << "";
+                return EXIT_FAILURE;
```
flomnes, line 202: Change 1

src/tools/batchrun/main.cpp
```diff
@@ -190,166 +189,179 @@ int main(int argc, char* argv[])
...
+                if (optNoTSImport)
+                    cmd << " --no-ts-import";
+                if (optIgnoreAllConstraints)
+                    cmd << " --no-constraints";
+                if (optParallel)
+                    cmd << " --parallel";
+                if (optForceParallel)
+                    cmd << " --force-parallel=" << *optForceParallel;
+                if (ortoolsUsed)
+                    cmd << " --use-ortools --ortools-solver=" << ortoolsSolver;
```
flomnes, line 307: Change 2

flomnes (commented):
Most changes are due to auto-formatting. I highlighted "real" changes.

sylvlecl (approved):

PR #1985, opened 2024-03-12
[ANT-1330] yaml parser and Antlr4 integration

a-zakir:
|candidate   | cmake integration  |Linux| Windows (MSVC)|   without extra-dependency  | langage | vcpkg
|---|---|---|---|---|---|---|
| [libfyaml](https://github.com/pantoniou/libfyaml)  | ❌   |  ✔️ | ❌   | ❌ | c| ❌ 
| [LibYAML](https://pyyaml.org/wiki/LibYAML)         | ❌    | ✔️   | ✔️     | ❌| c| ✔️ 
| [libcyaml](https://github.com/tlsa/libcyaml)            | ❌   | ✔️    | ❌     | ❌  | c| ❌ 
| [yaml-cpp](https://github.com/jbeder/yaml-cpp)    |  ✔️  |   ✔️  | ✔️  | ✔️ |c++| ✔️ 

added vcpkg  package [antlr4](https://github.com/antlr/antlr4)  on both (ub & windows)


flomnes (commented):
Very nice summary ! Instead of a `main`, you could add a very simple unit test with not much effort, right ?

a-zakir:
> Very nice summary ! Instead of a `main`, you could add a very simple unit test with not much effort, right ?

yes :)

.github/workflows/ubuntu.yml
```diff
@@ -28,6 +28,10 @@ env:
   RUN_SIMPLE_TESTS: ${{ github.event_name == 'push' || inputs.run-tests == 'true' }}
   RUN_EXTENDED_TESTS: ${{ github.event_name == 'schedule' || inputs.run-tests == 'true' }}
   REF: ${{ inputs.target_branch =='' && github.ref || inputs.target_branch}}
+  VCPKG_ROOT: ${{ github.workspace }}/vcpkg
+  vcpkgPackages: yaml-cpp antlr4 boost-test
+  triplet: x64-linux-dynamic
```
sylvlecl: I think we'd better use static linking, it will be far easier to package.
We already link almost everything statically: ortools, its solvers, minizip (not sirius but I don't see any reason for this).

src/tests/src/libs/antares/antlr4-interface/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+target_link_libraries(${execname}
+                       PRIVATE
+                       antlr-interface
+		               Boost::unit_test_framework
+                     )  
+
+
+target_include_directories(${execname} 
+        PRIVATE
+         ${ANTLR4_INCLUDE_DIR})
```
sylvlecl, line 17: If it's needed by users of the lib, we need to make it public in the CMakeList of the lib itself.
Even better would be to be able to add it to the antlr4-runtime imported target, but I don't know if it's possible.

sylvlecl (commented):
It looks good ! :)
To be confirmed but for me the best option will be to link statically.

.github/workflows/ubuntu.yml
a-zakir: i did this because of boost which seems to be linked dynamically (?) 
will find a way

.github/workflows/ubuntu.yml
```diff
@@ -50,12 +54,50 @@ jobs:
...
            sudo apt-get install g++-10 gcc-10
+          #  sudo apt-get install libyaml-cpp-dev libantlr4-runtime-dev 
+    
+    - name: export wxWidgets script
+      shell: bash
+      run: |
+           echo ${{ env.WX_CONFIG}}
+           cat ${{ env.WX_CONFIG}}
+           export WX_CONFIG=/usr/bin/wx-config
+           cat $WX_CONFIG
```
flomnes: Is this needed ?

src/tests/src/libs/antares/antlr4-interface/test_antlr_interface.cpp
```diff
@@ -0,0 +1,49 @@
...
+    {
+        std::cout << token->toString() << std::endl;
+    }
+    BOOST_CHECK(tokens.getTokens().size() == 6);
+    // ExprParser parser(&tokens);
+    // tree::ParseTree* tree = parser.main();
+
+    // auto s = tree->toStringTree(&parser);
+    // auto input_stream = parser.getInputStream();
+    // BOOST_CHECK(input_stream->size() == my_input.size());
```
flomnes: Clean up commented stuff, don't write to `std::cout`

flomnes (commented):
If ANTLR4 and yaml-cpp become dependencies, we need to install them on our machines as well. Are they easy to setup on Ubuntu & Windows ?

a-zakir:
> If ANTLR4 and yaml-cpp become dependencies, we need to install them on our machines as well. Are they easy to setup on Ubuntu & Windows ?

yes they are, just follow respective workflow instructions

src/tests/src/libs/antares/antlr4-interface/test_antlr_interface.cpp
a-zakir: ok

.github/workflows/ubuntu.yml
a-zakir: the env var WX_CONFIG is expected by [FindwxWidgets.cmake](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/b515cc9fddf7c2378a8aa7a4b598ab763cd64265/src/cmake/wxWidgets/FindwxWidgets.cmake#L820)

flomnes: It is defined above ⬆️

flomnes:
> > If ANTLR4 and yaml-cpp become dependencies, we need to install them on our machines as well. Are they easy to setup on Ubuntu & Windows ?
> 
> yes they are, just follow respective workflow instructions

For Ubuntu, yaml-cpp seems to be available using APT, ANTLR4 as well. We may have to update the BUILD doc accordingly.

For Windows, use vcpkg ?

flomnes:
Please fix conflicts

a-zakir:
> > > If ANTLR4 and yaml-cpp become dependencies, we need to install them on our machines as well. Are they easy to setup on Ubuntu & Windows ?
> > 
> > 
> > yes they are, just follow respective workflow instructions
> 
> For Ubuntu, yaml-cpp seems to be available using APT, ANTLR4 as well. We may have to update the BUILD doc accordingly.
> 
> For Windows, use vcpkg ?

so you don"t want vcpkg in ubuntu?

.github/workflows/ubuntu.yml
a-zakir: Okay, I'll clean it up

flomnes:
> > > > If ANTLR4 and yaml-cpp become dependencies, we need to install them on our machines as well. Are they easy to setup on Ubuntu & Windows ?
> > > 
> > > 
> > > yes they are, just follow respective workflow instructions
> > 
> > 
> > For Ubuntu, yaml-cpp seems to be available using APT, ANTLR4 as well. We may have to update the BUILD doc accordingly.
> > For Windows, use vcpkg ?
> 
> so you don"t want vcpkg in ubuntu?

If it is possible to use APT, I'd rather use APT

a-zakir:
> > > > > If ANTLR4 and yaml-cpp become dependencies, we need to install them on our machines as well. Are they easy to setup on Ubuntu & Windows ?
> > > > 
> > > > 
> > > > yes they are, just follow respective workflow instructions
> > > 
> > > 
> > > For Ubuntu, yaml-cpp seems to be available using APT, ANTLR4 as well. We may have to update the BUILD doc accordingly.
> > > For Windows, use vcpkg ?
> > 
> > 
> > so you don"t want vcpkg in ubuntu?
> 
> If it is possible to use APT, I'd rather use APT

ok

flomnes (commented):
Add CMAKE options for these new dependencies ? We don't always need them for now

src/tests/src/libs/antares/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ if(WITH_YAMLCPP)
 endif()
 
 if(WITH_ANTLR4)
-    add_subdirectory(antlr-interface)
+    add_subdirectory(antlr4-interface)
```
a-zakir, line 11: why are you doing this?

a-zakir: you should rename the lib also and the dir under src/lib/

payetvin: src/tests/src/libs/antares/antlr4-interface/test_antlr_interface.cpp

flomnes (approved):

PR #1986, opened 2024-03-12
Add ts-generation for links [ANT-1084]

payetvin:


src/tools/ts-generator/include/antares/ts-generator/thermal_interface.h
```diff
@@ -0,0 +1,50 @@
...
+{
+
+struct thermalInterface
+{
+    std::array<double, 8760> foDuration;
+    std::array<double, 8760> poDuration;
+    std::array<double, 8760> foRate;
+    std::array<double, 8760> poRate;
+    std::array<double, 8760> npoMin;
+    std::array<double, 8760> npoMax;
```
flomnes, line 37: Pros 
- We know that `foDuration` and its friends must of size 8760.

Cons
-  On my machine, `sizeof(thermalInterface) = 420512`, which is huge. With this, you **will** get stack overflows unless you're lucky.

Maybe you could wrap a `std::vector` inside a class that does not allow resizing ?

```cpp
class YearlyTS {
public:
YearlyTS() : mVec(8760) {}
double& operator[](std::size_t idx) { return mVec[idx];}
// etc.
// Do *NOT* expose resize, clear, push_back, etc.
private:
  std::vector<double> mVec;
};

struct newStructure
{
    YearlyTS foDuration;
    YearlyTS poDuration;
    YearlyTS foRate;
    YearlyTS poRate;
    YearlyTS npoMin;
    YearlyTS npoMax;

    unsigned unitCount;
    float nominalCapacity;

    double forcedVolatility;
    double plannedVolatility;

    Data::ThermalLaw forcedLaw = Data::thermalLawUniform;
    Data::ThermalLaw plannedLaw = Data::thermalLawUniform;
};
```

So you get both the "small object on the stack" and "fixed size" aspects ? Note that on my machine, `sizeof(newStructure) = 176`.

src/tools/ts-generator/include/antares/ts-generator/thermal_interface.h
```diff
@@ -0,0 +1,50 @@
...
+
+    unsigned unitCount;
+    float nominalCapacity;
+
+    double forcedVolatility;
+    double plannedVolatility;
+
+    Data::ThermalLaw forcedLaw = Data::thermalLawUniform;
+    Data::ThermalLaw plannedLaw = Data::thermalLawUniform;
+
```
flomnes: Put the definition of these enums in this file, instead of where it is now.

src/tools/ts-generator/include/antares/ts-generator/thermal_interface.h
```diff
@@ -0,0 +1,50 @@
...
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+# pragma once
+
+#include <array>
+
+#include <antares/study/parts/thermal/cluster.h>
```
flomnes: This should not be necessary

flomnes (commented):
Pros 
- We know that `foDuration` and its friends must of size 8760.

Cons
-  On my machine, `sizeof(thermalInterface) = 420512`, which is huge. With this, I could reproduce a segfault by just declaring `thermalStructure t[24];` on the stack.

Maybe you could wrap a `std::vector` inside a class that does not allow resizing ?

```cpp
class YearlyTS {
public:
YearlyTS() : mVec(8760) {}
double& operator[](std::size_t idx) { return mVec[idx];}
// etc.
// Do *NOT* expose resize, clear, push_back, etc.
private:
  std::vector<double> mVec;
};

struct newStructure
{
    YearlyTS foDuration;
    YearlyTS poDuration;
    YearlyTS foRate;
    YearlyTS poRate;
    YearlyTS npoMin;
    YearlyTS npoMax;

    unsigned unitCount;
    float nominalCapacity;

    double forcedVolatility;
    double plannedVolatility;

    Data::ThermalLaw forcedLaw = Data::thermalLawUniform;
    Data::ThermalLaw plannedLaw = Data::thermalLawUniform;
};
```

So you get both the "small object on the stack" and "fixed size" aspects ? Note that on my machine, `sizeof(newStructure) = 176`.

src/tools/ts-generator/include/antares/ts-generator/thermal_interface.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Antares::TsGenerator
+{
+
+class preproVector
+{
+public:
+    preproVector() : mVec(8760) {}
```
flomnes:
```suggestion
    preproVector() : mVec(HOURS_PER_YEAR) {}
```

src/tools/ts-generator/thermal_interface.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+template<class T>
+std::vector<thermalInterface> convertToInterface(T list)
+{
+    std::vector<thermalInterface> clusters;
+    clusters.resize(list.size());
+
+    return clusters;
+}
+
```
flomnes: This won't work, we need to place this in the header, or in a .hxx file.

src/solver/ts-generator/include/antares/solver/ts-generator/generator.h
```diff
@@ -1,5 +1,4 @@
-/*
-** Copyright 2007-2024, RTE (https://www.rte-france.com)
+/*** Copyright 2007-2024, RTE (https://www.rte-france.com)
```
flomnes: revert

flomnes (changes requested):
We should keep compatibility
- If `GenerateThermalTimeSeries` is called from solver.hxx, it should write to output/ts-generator/mc-XXX/etc
- If it is called from ts-generator/main.cpp, it should write to output/ts-generator directly (without mc year)

You may use `std::optional<uint>` to achieve this. We'll remove it when dropping support from ts-generation in antares-solver, so please add a TODO note. Another possibility is to let the caller provide the path, which I prefer.

src/solver/ts-generator/thermal.cpp
```diff
@@ -575,7 +559,7 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
                 double AVPDayInTheYear = AVP[dayInTheYear];
                 for (uint h = 0; h != 24; ++h)
                 {
-                    dstSeries[hour] = std::round(AVPDayInTheYear * modulation[hour]);
+                    dstSeries[hour] = std::round(AVPDayInTheYear * cluster.modulationCapacity[hour]);
```
flomnes: You can keep the modulation reference, and just change its target (`cluster.modulationCapacity`)

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -490,7 +491,8 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
     {
         auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
 
-        GenerateThermalTimeSeries(study, clusters, year, pResultWriter);
+        const std::string savePath = std::string("ts-generator") + SEP + "thermal" + SEP + "mc-0";
```
flomnes: 1. [style] Only define SEP where needed
2. [logic] `year` may not be 0 

```suggestion
#define SEP Yuni::IO::Separator
        const std::string savePath = std::string("ts-generator") + SEP + "thermal" + SEP + "mc-" + std::to_string(year);
#undef SEP
```

src/libs/antares/study/include/antares/study/area/links.h
```diff
@@ -21,6 +21,7 @@
 #ifndef __ANTARES_LIBS_STUDY_LINKS_H__
 #define __ANTARES_LIBS_STUDY_LINKS_H__
 
+#include <antares/study/parts/thermal/cluster.h>
```
flomnes: Everything should be defined in ts-generator/generator.h, so you won't need this awkward include.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -39,6 +39,7 @@
 #include <yuni/job/job.h>
 
 #include "antares/concurrency/concurrency.h"
+#define SEP Yuni::IO::Separator
```
flomnes:
```suggestion
```
see below

src/libs/antares/study/include/antares/study/area/links.h
payetvin: This will create circular dependency between the generator and the data structures.
In the future we want a separate tool that's depends only on structures and specialize what's needed after.

flomnes: If we move ThermalLaw to ts-generator.h,

ts-generator depends on ThermalCluster
ts-generator depends on AreaLink

That's all, right ?

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -490,8 +489,9 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
     if (refreshTSonCurrentYear)
     {
         auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
-
+#define SEP Yuni::IO::Separator
         const std::string savePath = std::string("ts-generator") + SEP + "thermal" + SEP + "mc-0";
```
flomnes: Please use `year` (see other comment)

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -489,8 +489,10 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
     if (refreshTSonCurrentYear)
     {
         auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
-
-        GenerateThermalTimeSeries(study, clusters, year, pResultWriter);
+#define SEP Yuni::IO::Separator
+        const std::string savePath = std::string("ts-generator") + SEP + "thermal" + SEP + "mc-l" + year;
```
flomnes:
```suggestion
        const std::string savePath = std::string("ts-generator") + SEP + "thermal" + SEP + "mc-" + std::to_string(year);
```

payetvin:
Add seed property and number of ts

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -489,8 +489,11 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
     if (refreshTSonCurrentYear)
     {
         auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
-
-        GenerateThermalTimeSeries(study, clusters, year, pResultWriter);
+#define SEP Yuni::IO::Separator
+        const std::string savePath = std::string("ts-generator") + SEP + "thermal" + SEP +
+            "mc-l" + std::to_string(year);
```
flomnes:
```suggestion
            "mc-" + std::to_string(year);
```

src/solver/ts-generator/thermal.cpp
```diff
@@ -608,24 +612,48 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
+    logs.info();
+    logs.info() << "Generating the links time-series";
+
+    auto generator = std::make_unique<GeneratorTempData>(study);
+
+    for (auto* link : links)
+    {
+        ThermalInterface clusterInterface(
+          link->tsGeneration, link->directCapacities, link->with->name);
+        (*generator)(*link->from, clusterInterface);
```
flomnes: - We need to distinguish direct & indirect capacities
- `link->directCapacities` and `link->indirectCapacities` should not be provided, use a temporary instead to make it clear that `link->directCapacities` and `link->indirectCapacities` won't be used after generating TS.
- We need to write generated TS to disk

src/solver/ts-generator/thermal.cpp
```diff
@@ -584,13 +568,58 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+
+    enum { precision = 0 };
+
+    std::string buffer;
+    if (directCapacity)
+        link.directCapacities.timeSeries.saveToBuffer(buffer, precision);
+    else
+        link.indirectCapacities.timeSeries.saveToBuffer(buffer, precision);
+
+    writer.addEntryFromBuffer(pTempFilename.c_str(), buffer);
```
flomnes:
```suggestion
void writeLinksResultsToDisk(const Data::Study& study,
                             Solver::IResultWriter& writer,
                             const Data::Matrix<>& ts,
                             const std::string& savePath,
                             bool directCapacity)
{
    if (study.parameters.noOutput)
        return;

    enum { precision = 0 };

    std::string buffer;
    ts.saveToBuffer(buffer, precision);

    writer.addEntryFromBuffer(savePath, buffer);
```

src/solver/ts-generator/thermal.cpp
```diff
@@ -608,24 +637,52 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
+    logs.info() << "Generating the links time-series";
+
+    auto generator = std::make_unique<GeneratorTempData>(study);
+
+    for (auto* link : links)
+    {
+        ThermalInterface clusterInterfaceDirect(
+                link->tsGeneration, link->directCapacities, link->with->name);
+        ThermalInterface clusterInterfaceIndirect(
+                link->tsGeneration, link->indirectCapacities, link->with->name);
```
flomnes: Don't use `link->directCapacities` and `link->indirectCapacities`, it's misleading. Someone could think this data will be used in memory, like thermal series, but it won't.

src/libs/antares/study/include/antares/study/area/links.h
```diff
@@ -204,6 +207,26 @@ class AreaLink final : public Yuni::NonCopyable<AreaLink>
     int linkWidth;
 
     friend struct CompareLinkName;
+
+    struct LinkTsGeneration
```
guilpier-code: **struct LinkTsGeneration** : 
- This class is about configuration parameters for TS generation only : it has nothing specific to **links**. It could be used for thermal as well.
-  As it's only a collection of parameter to make a TS, and has nothing specific to do with links, it should not be in this file. It should be moved to another source file about TS generation.
- Eventually, not sure this **struct** is useful (when looking at its usage). Maybe it should be removed.

We don't want to mix business rules and TS generation.

src/libs/antares/study/include/antares/study/parameters.h
```diff
@@ -268,6 +268,8 @@ class Parameters final
     uint nbTimeSeriesThermal;
     //! Nb of timeSeries : Solar
     uint nbTimeSeriesSolar;
+    //! Nb of timeSeries : Solar
+    uint nbTimeSeriesLinks;
```
guilpier-code: **nbTimeSeriesLinks** ==> **nbLinkTStoGenerate**, more specific about what this number represents. This number does not represent the number of ready-made time series. 
Variables **nbTimeSeriesSolar**, **nbTimeSeriesThermal**, ... should also be renamed (to be consistent, but will we take time to do it ?)

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -489,8 +489,11 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
     if (refreshTSonCurrentYear)
     {
         auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
-
-        GenerateThermalTimeSeries(study, clusters, year, pResultWriter);
+#define SEP Yuni::IO::Separator
```
guilpier-code: We should not use **Yuni** (as decided). Furthermore, its makes us define macro here. 
We should try to use **std::filesystem** instead.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -489,8 +489,11 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
     if (refreshTSonCurrentYear)
     {
         auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
```
guilpier-code: About **pData.haveToRefreshTSThermal** (and the use of bitwise operators to store booleans) : 

There has been a confusion between **have to refresh** and **have to generate** for a long time in **Antares**.
To me, this confusion comes from the fact that we have 2 **unsigned int** to store information about generating or refreshing about all TS. We encode this information in these 2 integers (with a bitwise operator), and further retrieves the information from theses integers (with another bitwise operator). This means we need complicated **enum** (with power of 2) to handle this.
I'm not sure we need this complexity. 
we may have this more simple : for each production means, we could have 2 booleans in **study.parameters** : generate ? refresh ? that's it.
We remove the **enum**, we remove **pData.haveToRefreshTS\<whatever\>** and all the too complicated things.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -489,8 +489,11 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
...
-
-        GenerateThermalTimeSeries(study, clusters, year, pResultWriter);
+#define SEP Yuni::IO::Separator
+        const std::string savePath = std::string("ts-generator") + SEP + "thermal" + SEP +
```
guilpier-code: **Question** : 
So we want to generate TS for thermal and we want to save them to a directory **ts-generator/thermal/mc-\<year\>**.
Are we not supposed to save them into **input/thermal/series/...**, and possibly erase the already existing ones ?

src/tools/ts-generator/main.cpp
```diff
@@ -34,7 +34,6 @@
 #include <antares/writer/result_format.h>
 #include <antares/checks/checkLoadedInputData.h>
 
-
 using namespace Antares;
 
 struct TsGeneratorSettings
```
guilpier-code: `struct TsGeneratorSettings` : 
We know we're in the TS generator, there's no need to repeat it.
And technically, these are script options (or command line arguments).
So I would have : 
```C++
struct Options
{
    ....
}
``` 
See further comments for more.

src/tools/ts-generator/main.cpp
```diff
@@ -92,6 +97,35 @@ std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
...
+        else
+            logs.warning() << "Link not found: " << linkID;
+    }
+
+    return links;
+}
+
 int main(int argc, char *argv[])
 {
     TsGeneratorSettings settings;
```
guilpier-code: According to me (and clean code), this main function should be shorter.
It contains too many implementation details we don't need to know about at this level.
It should contain only the big steps, and as few **if** statement as possible.
For example, we could have (take anything you're interested in, reject the rest) : 
```C++
int main(int argc, char *argv[])
{
    Options options();
    options.parse(argc, argv): // So the parser is created and used inside class Options
    options.check()
    
    ThermalClusterCollector clusterCollector(options);
    clusterCollector.run(); // Loading thermal clusters we're interested in
    std::vector<ThermalCluster> clustersForTSgeneration = clusterCollector.getList();
    
    for (auto& cluster : clustersForTSgeneration)
    {
        // Business rules and TS generation separation
        TSgenerationConfig ts_config; 
        ts_config.param1 = cluster->property1;
        ...
     
        TSgenerator tsGenerator(ts_config);
        matrix ts = tsGenerator.generate();
        // Here we print the TS
        // then we attach TS to the vector
    }
   
    // Same for link capacities here
}
```

guilpier-code: About `auto study = std::make_shared<Data::Study>(true);` :
We need to load the whole study to get several parameters (regarding thermal or link) used to generate TS.
Isn't that overkill ? 
Do we even need all data about a thermal cluster or a link ? we don't. 
We may even load some large ready made TS that want to replace with generated ones. It does not seem proper.

This could be avoided if code was more "modular", meaning that it would already have code units to load only what we're interested in here. But it's not.
That said, the planned work on loading input data might enhance this.

guilpier-code: About : 
```C++
Data::StudyLoadOptions studyOptions;
studyOptions.prepareOutput = true;
```
We're forced to writes these lines in order to be able to load a study we need a tiny part of.
We should not need this lines.

guilpier-code: `study->initializeRuntimeInfos();` : why do we need this ?

src/tools/ts-generator/main.cpp
```diff
@@ -141,16 +175,31 @@ int main(int argc, char *argv[])
     auto resultWriter = Solver::resultWriterFactory(
             Data::ResultFormat::legacyFilesDirectories, study->folderOutput, nullptr, nullDurationCollector);
 
-    std::vector<Data::ThermalCluster*> clusters;
+#define SEP Yuni::IO::Separator
```
guilpier-code: `#define SEP Yuni::IO::Separator`
We cannot wish to remove **Yuni** from our code and add new uses of **Yuni** at the same time.
Besides, we don't want to define macros.
So here, we should use **std::filesystem** features as much as possible.

src/tools/ts-generator/main.cpp
```diff
@@ -141,16 +175,31 @@ int main(int argc, char *argv[])
...
+    const std::string thermalSavePath = std::string("ts-generator") + SEP + "thermal";
```
guilpier-code: So, about the location where to print, we have : 
- one piece of the path in **study->folderOutput** we build the writer with.
- another piece of the path in variables **thermalSavePath** or **linksSavePath**
- another piece of the path represented by the area name.

And everything is assembled somewhere in the **generateThermalTimeSeries** function.
Should not the code be more simple ?
See this comment (put a link here)

src/tools/ts-generator/main.cpp
```diff
@@ -141,16 +175,31 @@ int main(int argc, char *argv[])
     auto resultWriter = Solver::resultWriterFactory(
```
guilpier-code: Question to @flomnes : 
Why does a **IResultWriter** need a **duration collector** ? 
Does the writer measures the time while it prints ?
Isn't that a violation of the **SRP** ?

src/tools/ts-generator/main.cpp
guilpier-code: `Antares::Check::checkMinStablePower(true, study->areas);` : 
Why do we need to make this check ?
I ask because it's unclear to me if we need the cluster's **minStablePower** to generate thermal TS

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
-        (*generator)(*cluster->parentArea, *cluster);
+    {
+        ThermalInterface clusterInterface(cluster);
+        (*generator)(*cluster->parentArea, clusterInterface);
 
-    delete generator;
+        if (archive)
+            writeThermalResultsToDisk(study, writer, *cluster->parentArea, *cluster, savePath);
+
+        cluster->calculationOfSpinning();
```
guilpier-code: **Question** :
Why do we change the TS inside the cluster after this TS was printed on disk ?
We're now sure that the TS used in the simulation is different from the ones printed... Isn't that a problem ?
I just noticed that this instruction was moved. 
But still, it's strange : what if, after generating some TS, user decides to re-run the same study with the previously generated thermal TS as ready-made TS ? 
She/he won't have the same results, I guess.

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
 {
     logs.info();
     logs.info() << "Generating the thermal time-series";
-    Solver::Progression::Task progression(study, year, Solver::Progression::sectTSGThermal);
 
-    auto* generator = new GeneratorTempData(study, progression, writer);
+    bool archive = (0 != (study.parameters.timeSeriesToArchive & Data::timeSeriesThermal));
 
-    generator->currentYear = year;
+    auto generator = std::make_unique<GeneratorTempData>
```
guilpier-code: `auto generator = std::make_unique<GeneratorTempData>(...)` : 
What's the point of having a smart pointer on a **GeneratorTempData** ?
A plain instance would be enough.

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
+        (study, study.parameters.nbTimeSeriesThermal);
 
     // TODO VP: parallel
     for (auto* cluster : clusters)
-        (*generator)(*cluster->parentArea, *cluster);
+    {
+        ThermalInterface clusterInterface(cluster);
+        (*generator)(*cluster->parentArea, clusterInterface);
```
guilpier-code: `(*generator)(*cluster->parentArea, clusterInterface);` : 
When I bump into this particular instruction, I always wonder "what does it do ?".
Well of course, it calls an **operator()** which has no name and therefore hides what it does (except if you think : "what does an generator do ? Generate of course.").
I still think we should avoid an **operator()** in this case, and rename this member function into **generate(...)**, even if it's kind of redundant.

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
 
-    generator->currentYear = year;
+    auto generator = std::make_unique<GeneratorTempData>
+        (study, study.parameters.nbTimeSeriesThermal);
 
     // TODO VP: parallel
     for (auto* cluster : clusters)
-        (*generator)(*cluster->parentArea, *cluster);
+    {
+        ThermalInterface clusterInterface(cluster);
```
guilpier-code: **ThermalInterface** is a not an interface in the sense we could expect.
- it contains only data (no member function), when interfaces usually contain only abstract methods
- interfaces are usually abstract : they cannot be instantiated, there are meant to be implemented / derived

Furthermore, its name contains **Thermal**, but it's used for links as well.

This class contains only the parameters used to generate TS (whether thermal or link capacities TS), so let's rename it (for instance) **TSconfigParams**.
But for me, there is more to be done : see other remarks in this file.

src/solver/ts-generator/thermal.cpp
```diff
@@ -95,54 +117,21 @@ class GeneratorTempData final
     std::vector<std::vector<double>> PPOW;
 
     Yuni::String pTempFilename;
```
guilpier-code: `Yuni::String pTempFilename` : unused or redefined, to be removed.

guilpier-code: Many data members are used in one member function only, so what's the point to keep them as data members ?

src/solver/ts-generator/thermal.cpp
```diff
@@ -35,31 +37,51 @@
...
+ name(areaDestName)
+{
+}
+
 namespace
 {
 class GeneratorTempData final
 {
 public:
-    GeneratorTempData(Data::Study& study,
```
guilpier-code: **GeneratorTempData** : this class is the real class for the TS generator (for thermal and link capacities).
I don't understand what it's temporary (like its name suggests) in it.
We should rename it into _what ?_

src/libs/antares/study/include/antares/study/area/links.h
```diff
@@ -21,6 +21,9 @@
 #ifndef __ANTARES_LIBS_STUDY_LINKS_H__
 #define __ANTARES_LIBS_STUDY_LINKS_H__
 
+//#include <antares/study/parts/thermal/cluster.h>
```
flomnes:
```suggestion
```

flomnes (changes requested):
Options `--links a,b;b,a`, as requested by our users.

src/solver/ts-generator/thermal.cpp
```diff
@@ -35,31 +37,51 @@
 
 namespace Antares::TSGenerator
 {
+ThermalInterface::ThermalInterface(Data::ThermalCluster* source) :
```
guilpier-code: Class **ThermalInterface** (or **TSconfigParams**, as suggested in another comment) should not depend on the business rules (thermal clusters, links, ...). We should separate business rules from TS generation. 
We could make this class a struct, remove the specialized constructors and fill its fields from outside.

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
+        std::ranges::for_each(area.links, [&links](auto& l) {
+            links.emplace_back(l.second, linkDirection::direct);
+            links.emplace_back(l.second, linkDirection::indirect);
+        });
+    });
+
+    return links;
+}
+
+void writeThermalResultsToDisk(const Data::Study& study,
```
guilpier-code: About write functions : 
they have duplicate code.
That because they make too many things.
For example the file path should be given from outside (as a parameter).

Besides, we mix business rules and printing TS job, that's another reason some code cannot be reused.

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
 
     // TODO VP: parallel
     for (auto* cluster : clusters)
-        (*generator)(*cluster->parentArea, *cluster);
+    {
+        ThermalInterface clusterInterface(cluster);
+        (*generator)(*cluster->parentArea, clusterInterface);
 
-    delete generator;
+        if (archive)
```
guilpier-code: **generateThermalTimeSeries(...)** and **generateLinkTimeSeries(...)** : 
Here we don't only generate TS. 
Besides generating TS, they print the TS, and we possibly change them afterwards (for thermal).
Furthermore, these functions mix TS generation and business rules.
My opinion is that these functions only not exist. 
A loop over clusters / links should be made at the upper level. At this level (we're in the TS generation level), we should only generate TS, whoever it's for (thermal, link capacities, ....).

src/solver/ts-generator/thermal.cpp
```diff
@@ -27,6 +27,8 @@
 #include <antares/logs/logs.h>
 #include <antares/writer/i_writer.h>
 
+#include <antares/solver/ts-generator/generator.h>
```
guilpier-code: About the current source file : 
It mixes business rules (thermal clusters, link capacities) and TS generation.
We should rename this file into something else related to TS generation only, and move code about thermal or links elsewhere (in other files).

Sometimes we mix thermal and links (some thermal function call links related code), which is very confusing (please see other remarks for this file).

src/tools/ts-generator/main.cpp
```diff
@@ -141,16 +171,33 @@ int main(int argc, char *argv[])
...
```
guilpier-code: `auto resultWriter = Solver::resultWriterFactory(....);`
It seems we don't need the factory here : we know want a **ImmediateFileResultWriter**, why not create one directly ?

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
+    pTempFilename.reserve(study.folderOutput.size() + 256);
+
+    pTempFilename.clear() << savePath << SEP << area.id << SEP << cluster.id() << ".txt";
+
+    enum { precision = 0 };
+
+    std::string buffer;
+    cluster.series.timeSeries.saveToBuffer(buffer, precision);
+
+    writer.addEntryFromBuffer(pTempFilename.c_str(), buffer);
```
guilpier-code: **Question** : 
why do we print TS twice ?
What I mean : we call **saveToBuffer** and **writer.addEntryFromBuffer**

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +585,122 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
+{
+    if (study.parameters.noOutput)
+        return;
+
+    Yuni::String pTempFilename;
+    pTempFilename.reserve(study.folderOutput.size() + 256);
+
+    pTempFilename.clear() << savePath << SEP << area.id << SEP << cluster.id() << ".txt";
+
+    enum { precision = 0 };
```
guilpier-code:
```C++
enum { precision = 0 };
cluster.series.timeSeries.saveToBuffer(buffer, precision);
```
To be replaced with : 
```C++
cluster.series.timeSeries.saveToBuffer(buffer, 0);
```

guilpier-code (changes requested):

src/tools/ts-generator/main.cpp
payetvin: containes `    MersenneTwister random[seedMax];`

src/tools/ts-generator/main.cpp
guilpier-code: Furthermore, about **TsGeneratorSettings.studyFolder**, shouldn't we have a **-d** (introducing the study path as a command line arg ?

src/tools/ts-generator/main.cpp
guilpier-code: So we're calling a function that makes many things for just a random seed thing.
Could we not improve this (if possible) ?

src/libs/antares/study/include/antares/study/area/links.h
payetvin: This struct will hold data read from files when loading each links.
Could be declared somewhere else, I prefer there to pack everything related to links together.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
payetvin: Will be removed

src/solver/simulation/include/antares/solver/simulation/solver.hxx
payetvin: Easier for antares web to get results from the output dir.

src/tools/ts-generator/main.cpp
payetvin: I prefer not to specify a letter for a mandatory arg

payetvin: renamed into `struct Settings`

src/tools/ts-generator/main.cpp
payetvin: Will be refactored when removing in memory

payetvin: Indeed this could/should be improved, and is one of the reason that we made a specific tool for it.
However it's way more complex to load individual things rather than the whole study, and may be done in the future if users want a faster generator.

payetvin: `runtimeInfos` is allocated dynamically
Could be improved after we remove the in memory gen, to much dependency right now to do so

src/tools/ts-generator/main.cpp
payetvin: done

src/tools/ts-generator/main.cpp
payetvin: Done this way to maintain backward compatibility with in memory gen, will be simplified

src/tools/ts-generator/main.cpp
payetvin: related to modulation

src/solver/ts-generator/thermal.cpp
payetvin: This is the current behavior for in memory ts, it's not saved on disk with the spinning, it's applied while loading them.
Changing the behavior will thus break results.

src/solver/ts-generator/thermal.cpp
payetvin: done

src/solver/ts-generator/thermal.cpp
payetvin: Agreed, should be a simple function

payetvin: done

src/solver/ts-generator/thermal.cpp
payetvin: saveToBuffer -> convert to str
addEntry -> write this str

src/solver/ts-generator/thermal.cpp
```diff
@@ -606,26 +578,91 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(Data::AreaList& areas,
...
-                               Solver::IResultWriter& writer)
+                               Solver::IResultWriter& writer,
+                               const std::string& savePath)
 {
     logs.info();
     logs.info() << "Generating the thermal time-series";
-    Solver::Progression::Task progression(study, year, Solver::Progression::sectTSGThermal);
 
-    auto* generator = new GeneratorTempData(study, progression, writer);
+    bool archive = study.parameters.timeSeriesToArchive & Data::timeSeriesThermal;
```
guilpier-code: About **bool archive** : 
- no need to test its value inside the loop as it's the same for every cluster. Instead we should return if **archive** is **false**, and remove the `if (archive)` from the loop.
- Actually **archive**'s value should be tested only if we call this function from the **solver**, but not from the **TS generator**. So testing **archive**'s value should not be done here, but from the current function's caller. 

In general, we should isolate code that's used in both the **TS generator** and the **solver**, and reject the other code (thus 2 pieces) elsewhere in separate places.

guilpier-code (changes requested):

src/solver/ts-generator/thermal.cpp
flomnes: It's because of `cluster->calculationOfSpinning()` at l. 621. 

I agree it would be best to do this
```cpp
generateTimeSeries();
if (archive) 
{
    writeResultsToDisks();
}
```
But actually we have
```cpp
bool generateThermalTimeSeries()
{
    bool archive = study.parameters.timeSeriesToArchive & Data::timeSeriesThermal;
    for (auto* cluster : clusters)
    {
        generator.generateTS(*cluster->parentArea, clusterInterface);
        if (archive) // compatibilty with in memory
        {
            writeResultsToDisk(study, writer, cluster->series.timeSeries, filePath);
        }
        cluster->calculationOfSpinning(); // SIDE EFFECT YAY
    }
    return true;
}
```

src/libs/antares/study/include/antares/study/area/links.h
guilpier-code: Was moved and renamed (**AvailabilityTSGeneratorData**)

src/solver/simulation/include/antares/solver/simulation/solver.hxx
guilpier-code: We still have `#define` and **Yuni** usage.
Is it complicated to make this change (though it was applied elsewhere in this PR, see ts generator **main.cpp**) ?

src/solver/simulation/include/antares/solver/simulation/solver.hxx
guilpier-code: ok, this remark is out of this PR's scope, but should be looked at (at some point).
I will create a **gopro** ticket for this.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
guilpier-code: Here we run **solver** (as you know), we generate thermal TS in memory, then we save these TS.
User is usually able to save the generated TS in **input** and/or in **output**.
But now we automatically save them in a 3rd place. 
When running the **solver**, do we need to save the TS at this location ?

src/tools/ts-generator/main.cpp
guilpier-code: ok

guilpier-code: I really don't understand why we need this runtime infos thing : class **StudyRuntimeInfos**. 
Everything we need (for example take the seed of random numbers) is loaded from disk.
What does "runtime" have to do with its real mission ? Isn't this class another global variable ?  
It seems that **StudyRuntimeInfos** is a kind of junk room, has too many responsibilities, and a meaningless name. 

This class should be broken into pieces : all elements it contains don't belong together, and if we want to use one of those, we are forced to use the whole class.

Hope it will be done at some point.

guilpier-code: Sorry, I can't see any place where `cluster->minStablePower` is used to generate a TS for thermal clusters.
If it's true and I'm not missing something, what would be the value and usefulness of this check ?

src/solver/ts-generator/thermal.cpp
guilpier-code: - If we're running the **TS generator** :  computing the spinning for TS in memory **and** after TS is printed makes no sense. We really should remove this computation from function **generateThermalTimeSeries()**.
- If we're running the **solver** :  computing the spinning for thermal TS after generation has a meaning. So we should call **ThermalClusterList::calculationOfSpinning()** after creating them : 

```C++
// solver.hxx
void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
{
    ...
    if (refreshTSonCurrentYear)
    {
        auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
        ...
        generateThermalTimeSeries(study, clusters, pResultWriter, savePath);
        study.areas.calculationOfSpinning(); // Here adding the computation
    }
    ....
}
```

src/solver/ts-generator/thermal.cpp
guilpier-code: **GeneratorTempData** : it's not temporary (**Temp**), and it's only behavior (not **Data**). So the name makes no sense.
We should rename it into **AvailableProdTSgenerator**.

src/solver/ts-generator/thermal.cpp
guilpier-code: Already done

src/solver/ts-generator/thermal.cpp
guilpier-code: We can ignore this.

src/tools/ts-generator/main.cpp
guilpier-code: ok, let's refactor some other time.

src/solver/ts-generator/availability.cpp
```diff
@@ -584,48 +547,114 @@ void GeneratorTempData::operator()(Data::Area& area, Data::ThermalCluster& clust
...
+    bool archive = study.parameters.timeSeriesToArchive & Data::timeSeriesThermal;
 
-    generator->currentYear = year;
+    auto generator = GeneratorTempData(study, study.parameters.nbTimeSeriesThermal);
 
     // TODO VP: parallel
     for (auto* cluster : clusters)
-        (*generator)(*cluster->parentArea, *cluster);
+    {
+        AvailabilityTSGeneratorData clusterInterface(cluster);
```
guilpier-code: `AvailabilityTSGeneratorData clusterInterface(cluster);` 

to be replaced with : 

```C++
AvailabilityTSGeneratorData tsConfigData(cluster);
generator.generateTS(*cluster->parentArea, tsConfigData);
```

Same thing with **linkInterface**

guilpier-code: Same thing with **linkInterface**

src/solver/ts-generator/thermal.cpp
guilpier-code: Done, **ThermalInterface** was renamed into **AvailabilityTSGeneratorData**

src/tools/ts-generator/main.cpp
guilpier-code: ok, let's close this

src/solver/ts-generator/thermal.cpp
guilpier-code: Done. it's now called **availability.cpp**

src/solver/ts-generator/thermal.cpp
guilpier-code: ok, but as said in [this remark](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1986#discussion_r1562506106), we don't need the  calculation of spinning here. So the suggested simplification can be applied.

docs/user-guide/04-migration-guides.md
```diff
@@ -1,5 +1,26 @@
 # Migration guides
 This is a list of all recent changes that came with new Antares Simulator features. The main goal of this document is to lower the costs of changing existing interfaces, both GUI and scripts.
+
+## v9.2.0
+### (TS-generator only) TS generation for links
```
guilpier-code: What's the quantity being generated for links ? 
**capacity** ?

docs/user-guide/04-migration-guides.md
```diff
@@ -1,5 +1,26 @@
...
+In files input/links/<link1>/properties.ini, add the following properties
```
guilpier-code, line 6: This doc is not very clear to me, especially when opening it with a markdown reader (I use VS Code).
This is a long list of different things.
Apparently, there are 4 files concerned by the links capacities TS generation.
So it should be separated into 4 parts.

guilpier-code (changes requested):

docs/user-guide/04-migration-guides.md
payetvin: yes

src/tools/ts-generator/main.cpp
```diff
@@ -43,6 +47,11 @@ struct Settings
     bool allThermal = false;
     /// generate TS for a list "area.cluster;area2.cluster2;"
     std::string thermalListToGen = "";
+
+    /// generate TS for all links if activated
+    bool allLinks = false;
+    /// generate TS for a list "area.link;area2.link2;"
+    std::string linksListToGen = "";
```
guilpier-code, line 54: **linksListToGen** : is not of list of links (instances of class **AreaLink**) as it suggests, it's a list of link names we want to generate TS for.
So **linkNamesForTSgeneration** is longer but more clear.

src/tools/ts-generator/main.cpp
```diff
@@ -95,8 +111,32 @@ std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
     return clusters;
 }
 
+TSGenerator::listOfLinks getLinksToGen(Data::AreaList& areas, const std::string& linksToGen)
+{
+    TSGenerator::listOfLinks links;
+    const auto ids = splitStringIntoPairs(linksToGen, ';', '.');
+
+    for (const auto& [areaFromID, areaWithID]: ids)
```
guilpier-code, line 119: **[areaFromID, areaWithID]**  ==> [areaSourceID, areaTargetID].
I never liked **with**, which stands for **to** or **target** or **destination** area in a link, much better.

src/tools/ts-generator/main.cpp
```diff
@@ -95,8 +111,32 @@ std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
     return clusters;
 }
 
+TSGenerator::listOfLinks getLinksToGen(Data::AreaList& areas, const std::string& linksToGen)
+{
+    TSGenerator::listOfLinks links;
+    const auto ids = splitStringIntoPairs(linksToGen, ';', '.');
```
guilpier-code, line 117: **splitStringIntoPairs** : is a bit over engineered.
We always call this function with the same 2 delimiters.
Why do we pass them as arguments then ?
We could just use them inside the function.

src/tools/ts-generator/main.cpp
```diff
@@ -95,8 +111,32 @@ std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
...
+
```
guilpier-code, line 118: **ids** : you extract pairs (as the name of the function suggests).
And you get **ids** ?
**pairs** would be better.

src/solver/ts-generator/availability.cpp
```diff
@@ -592,6 +609,23 @@ std::vector<Data::ThermalCluster*> getAllClustersToGen(const Data::AreaList& are
...
+                                {
+                                    if (!l.second->tsGeneration.forceNoGeneration)
+                                        links.push_back(l.second);
+                                });
+      });
+
+    return links;
+}
+
 void writeResultsToDisk(const Data::Study& study,
```
guilpier-code, line 629: **writeResultsToDisk** : never called to write some "results" but to write **time series**.
So we could rename it into **writeTimeSeriesToDisk** more clear.
Especially as the writer is an **IResultWriter** that does **not** write results, we can get confused (do we write TS or results ? And if results, what results ?).

src/solver/ts-generator/availability.cpp
```diff
@@ -636,4 +672,50 @@ bool generateThermalTimeSeries(Data::Study& study,
     return true;
```
guilpier-code, line 672: **_Deprecated comment (please see the comment that just follows)_**

In **generateThermalTimeSeries**, we save TS only if parameter **archive** is **true**.
But there is another condition to be respected when saving TS : **study.parameters.noOutput**.
And this condition evaluation is hidden inside **writeResultsToDisk**.
So **archive** is evaluated at a level, and **noOutput** at another level.
Besides, **writeResultsToDisk** is supposed to **write** only, but actually, it **writes if** a certain condition is respected **(noOutput == false**). So the name of the function is misleading in a way. 
I recommend to extract the **if** condition from **writeResultsToDisk**, and have : 
```C++
if (archive && not study.parameters.noOutput) // compatibilty with in memory
  {
      ...
      writeResultsToDisk(study, writer, cluster->series.timeSeries, filePath);
  }
```
Doing this, we simplify **writeResultsToDisk** by : 
- removing some code inside its body
- removing an argument now useless (**study**)

Of course, it has an impact on the links TS save as well

src/solver/ts-generator/availability.cpp
```diff
@@ -636,4 +672,50 @@ bool generateThermalTimeSeries(Data::Study& study,
...
 
+bool generateLinkTimeSeries(Data::Study& study,
+                            const listOfLinks& links,
+                            Solver::IResultWriter& writer,
+                            const std::string& savePath)
+{
+    logs.info();
+    logs.info() << "Generating the links time-series";
+
+    auto generator = GeneratorTempData(study,
```
guilpier-code, line 683: class **GeneratorTempData** : why do we name this class like that ?
What is **temporary** ? Does this class contains only **data** or behaviors as well ?
It's a time series generator, isn't it ?
So let's rename it as it should be : class **AvailabilityTimeSeriesGenerator**.
Or in short (if you prefer) : **AvailabilityTSgenerator**.

guilpier-code, line 683:
```C++
auto generator = GeneratorTempData(study,
                                   ...)
```
no need to pass the whole study to the generator constructor if we just need `study.parameters.derated`
**generator** could also be renamed  **tsGenerator**.

guilpier-code: And by the way, in **GeneratorTempData**, a **private** data member should be easily spotted by an hint (a trailing underscore). 
Some data members are not trailed by "_". 
They should.

guilpier-code: About : 
```C++
GeneratorTempData::GeneratorTempData(Data::Study& study, unsigned nbOfSeriesToGen, MersenneTwister& rndGenerator):
...
```
and particularly about `MersenneTwister& rndGenerator`.
**rndGenerator** is unclear : round generator ? ah ok, you mean **random generator**.
We should not hesitate to clarify and have  **MersenneTwister& randomGenerator**.
And what is true for the constructor argument is also true for the data member : **randomGenerator_**

src/libs/antares/study/include/antares/study/area/links.h
```diff
@@ -206,6 +208,9 @@ class AreaLink final: public Yuni::NonCopyable<AreaLink>
     int linkWidth;
 
     friend struct CompareLinkName;
+
+    LinkTsGeneration tsGeneration;
```
guilpier-code, line 212: `LinkTsGeneration tsGeneration;` : 

This represents data / parameters to generate TS for a given link.
But for links, TS are supposed to be generated outside **Antares Simulator** : during an execution of the solver, no TS for links is generated (unlike for thermal clusters).
So it is surprising and wrong that some code used to : 
- read data for link TS generation
- store this data

uselessly inflates **Antares**'s code base.
This code should be moved to TS generator scope.
This was made intentionally because reading a study's input is a block : either we read it as a whole, or we read nothing.
We really should make reading a study modular : split it into many reading entities.
Doing this would allow removing code from Antares.

src/tools/ts-generator/main.cpp
```diff
@@ -149,7 +196,8 @@ int main(int argc, char* argv[])
                                                     nullptr,
```
guilpier-code, line 196: As already pointed somewhere, we don't need to call a factory here.
Just create the instance you need.
The problem with that factory is that you're forced to use a **durationCollector**, which is unused for the kind of writer you need (`legacyFilesDirectories`)  !

Factories are useful in polymorphism context, but this not what we have here, regarding the TS generation and their printing on disk.

**But please see the other comments on writing (right after this one), they are more important.**

guilpier-code: More importantly : 
- First, we can see we need **folderOutput** to build a writer.  And right after, we build 2 pieces of path for printing TS on disk (**thermalSavePath** and **linksSavePath**). I dug a little bit and realized that eventually **folderOutput** is appended with these 2 pieces of path. This concatenation is done for us, by the writer, and internally. 
But we could build **thermalSavePath** and **linksSavePath** directly with their root (**folderOutput**) here.
- Then, we can see that this writer uses **IO::fileSetContent(output.string(), entryContent);** internally in order to print the generated TS on disk.
- Finally, we can see that we create a writer, and we pass it down and down (along with a piece of path) in the call stack until it's really used, Why not create it where you need it ? Why create it at a high level and pass it down ? The way we write is an implementation detail we don't need to know about here.   

What I recommend is : 
- We don't need the writer here, but at a deeper level. We should remove the writer creation (and the **durationCollector**) from here. Subsequently, we don't pass down the writer anymore.
- But we now complete **thermalSavePath** and **linksSavePath** by prepending them **folderOutput**.
- Using **IO::fileSetContent()** instead of a writer is simpler (requires less lines of code).
Of course, using **std** functionalities instead of **IO::fileSetContent()** (which calls **Yuni** things internally) would be even better, but we don't have to bother with that now.

src/solver/ts-generator/availability.cpp
```diff
@@ -636,4 +672,50 @@ bool generateThermalTimeSeries(Data::Study& study,
...
+        auto& tsGenStruct = link->tsGeneration;
+
+        if (!tsGenStruct.valid)
+        {
+            logs.error() << "Missing data for link " << link->from->id << "/" << link->with->id;
+            return false;
+        }
+
+        // DIRECT
+        AvailabilityTSGeneratorData tsConfigDataDirect(tsGenStruct, ts, tsGenStruct.modulationCapacityDirect, link->with->name);
```
guilpier-code, line 701: About : 
```C++
AvailabilityTSGeneratorData tsConfigDataDirect(tsGenStruct, 
                                               ts, 
                                               tsGenStruct.modulationCapacityDirect, 
                                               link->with->name);
```
There are too much arguments, we could simplify this.
- **tsGenStruct** appears twice. Once is enough, and we handle things internally.
- about **ts** : this is the result of the generation process, it has nothing to do among the input data that are required to generate these TS. Instead **ts** should (logically) be returned by the generation.
- If we need  **link->from** as argument of method **generateTS**, it's probably because it should be an input data featuring in  **AvailabilityTSGeneratorData**  as well. Which is expected because : 
  - **link->with->name** is already required to build an instance of **AvailabilityTSGeneratorData**
  - we actually need **link->from->name**, so it goes with **link->with->name**
- To simplify more, we could even create an instance of the TS generator inside the loop over links, that is a TS generator per link. It would avoid passing input data at different places. And I'm not sure it would be costly at execution.     

So, I would recommend to simplify the code this way :
```C++
bool generateLinkTimeSeries(...)
{
    if (study.parameters.noOutput) // This if statement should even be moved to the caller level
            return;
     
    for (const auto& link: links)
    {
        // We don't need to define and resize TS here (see below)
        AvailabilityTSGeneratorData tsConfigDataDirect(
                                                        // We put all we need here 
                                                      );
        auto tsGenerator = AvailabilityTSgenerator(tsConfigDataDirect);
        Matrix<double> generatedTS = tsGenerator.run();
        
        destinationPath = ...;
        writeTimeSeriesToDisk(generatedTS, destinationPath);
    }
}
```
Note that : 
- according to previous comments, we don't need **writer** or **study** as argument of **writeTimeSeriesToDisk**.
- Creating a **Matrix** of TS and resizing it is now done by the generator internally, which is not silly : the generator is responsible for generating these TS, so it must know how many.
- the block :
```C++
if (study.parameters.noOutput)
    return;
```
  should even be moved up in **main.cpp** for a good reason : if **noOutput** is **true**, TS for links must not be printed at all, and it must be clear as soon as possible.

src/solver/ts-generator/availability.cpp
```diff
@@ -636,4 +672,50 @@ bool generateThermalTimeSeries(Data::Study& study,
...
+    auto generator = GeneratorTempData(study,
+                                       study.parameters.nbLinkTStoGenerate,
+                                       study.runtime->random[Data::seedTsGenLinks]);
+
+    for (const auto& link: links)
+    {
+        Data::TimeSeries ts(link->timeseriesNumbers);
+        ts.resize(study.parameters.nbLinkTStoGenerate, HOURS_PER_YEAR);
+
+        auto& tsGenStruct = link->tsGeneration;
```
guilpier-code, line 692: About `tsGenStruct = link->tsGeneration;` : 
- **tsGenStruct** represents actually some data. But the container type is a **struct**. So **tsGenStruct** should be renamed into **tsGenData**.
- **link->tsGeneration** is not precise enough (**tsGeneration** is a process, but we want data here), so it should be renamed **link->tsGenerationData**.

src/solver/ts-generator/availability.cpp
```diff
@@ -636,4 +672,50 @@ bool generateThermalTimeSeries(Data::Study& study,
...
+    logs.info();
+    logs.info() << "Generating the links time-series";
+
+    auto generator = GeneratorTempData(study,
+                                       study.parameters.nbLinkTStoGenerate,
+                                       study.runtime->random[Data::seedTsGenLinks]);
+
+    for (const auto& link: links)
+    {
+        Data::TimeSeries ts(link->timeseriesNumbers);
```
guilpier-code, line 689: About `Data::TimeSeries ts(link->timeseriesNumbers);` : 
We don't need the time series numbers in the context of TS generation.
Actually we don't even need a **Data::TimeSeries** here. 
A simple **Matrix** is enough.

src/solver/ts-generator/availability.cpp
```diff
@@ -636,4 +672,50 @@ bool generateThermalTimeSeries(Data::Study& study,
...
+        ts.resize(study.parameters.nbLinkTStoGenerate, HOURS_PER_YEAR);
+
+        auto& tsGenStruct = link->tsGeneration;
+
+        if (!tsGenStruct.valid)
```
guilpier-code, line 694: About block : 
```C++
if (!tsGenStruct.valid)
 {
      logs.error() << "Missing data for link " << link->from->id << "/" << link->with->id;
      return false;
 }
```
This block does not belong here, but it belongs to the action of reading the data, or possibly checking it, but **not** when making something with it. 
If data is not **valid**, user should know it as soon as possible, not when we generate TS.

docs/user-guide/04-migration-guides.md
guilpier-code: Then we should say it. 
Otherwise it's unclear

src/tools/ts-generator/main.cpp
```diff
@@ -119,9 +159,16 @@ int main(int argc, char* argv[])
...
+    if (settings.allLinks && !settings.linksListToGen.empty())
+    {
+        logs.error() << "Conflicting options, either choose all links or a list";
+        return 1;
+    }
+
     auto study = std::make_shared<Data::Study>(true);
     Data::StudyLoadOptions studyOptions;
     studyOptions.prepareOutput = true;
+    studyOptions.linksLoadTSGen = true;
```
guilpier-code, line 171: About` studyOptions.linksLoadTSGen = true;` : 
We don't need this.
Indeed, imagine we can get a pointer to each link we want to generate TS for, and collect these pointer to a list.
Then we could call **link->loadTSGenTimeSeries(folder)** in the current main program.
The consequence would be : 
- We don't need **studyOptions.linksLoadTSGen** here or elsewhere in the code
- We don't need to call it when running a simulation, simulation that don't need to generate TS for links anyway.
- What else ?

Besides, getting pointer is easy, as we already process study links in function **getLinksToGen(...)**.

src/libs/antares/study/parameters.cpp
```diff
@@ -539,6 +539,10 @@ static bool SGDIntLoadFamily_General(Parameters& d,
     {
         return value.to<uint>(d.nbTimeSeriesSolar);
     }
+    if (key == "nbtimeserieslinks")
```
guilpier-code, line 531: We know generating TS for links should be fully separated from the Antares solver.
What is true for the code is true for data as well.
For now data required generating TS for links are mixed data with the study data.
Here is a good example : **nbtimeserieslinks** is still held by **generaldata.ini**. Same for **seed_links**.
Eventually, we should separate data for TS generation form study data.

guilpier-code (changes requested):

src/libs/antares/study/area/links.cpp
```diff
@@ -134,6 +137,58 @@ bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& fol
...
+    // file name without suffix, .txt for general infos and mod_direct/indirect.txt
+    fs::path preproFile = folder / "prepro" / with->id.c_str();
+
+    // Prepro
+    fs::path filepath = preproFile;
+    filepath += ".txt";
+    if (fs::exists(filepath))
+    {
+        anyFileWasLoaded = true;
+        tsGeneration.valid = tsGeneration.prepro->data.loadFromCSVFile(
```
guilpier-code, line 159: About **tsGeneration.valid** : 
We don't need **valid** as a data member, we should just return **false** if the reading goes wrong, **true** otherwise. 
If reading goes wrong, we don't go further.

About this, see also [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1986#discussion_r1631329839)

guilpier-code (dismissed):

src/solver/ts-generator/availability.cpp
guilpier-code: About **study.parameters.noOutput** :  
This parameter can only be set from the solver, and as a command line option. 
As such, it has nothing to do with the **links** TS generation in particular, and with the TS generation in general.
It should not be used in the TS generator.

Besides,  there's a really strange thing about **noOutput** : 
- it's declared in 2 places : in **study.parameters** and in **settings** (that is solver's cmd line options)
- In **study.parameters** : it receives a default value (**false**), but it's never read as a general parameter. In other word, it's not a general parameter. At some point it's given a value that depends on **settings**. This is mysterious, we should clarify this (create a ticket).

src/tools/ts-generator/main.cpp
payetvin: Agree, too much changes needed to be done here, will be done later

src/tools/ts-generator/main.cpp
payetvin: Will do

src/tools/ts-generator/main.cpp
payetvin: Function is stored in utils.cpp, if we specialize the function only for this use case we need to move it in src/tools

src/tools/ts-generator/main.cpp
payetvin: will do

src/solver/ts-generator/availability.cpp
payetvin: TimeSeries is a generic object, it doesn't specify if it's the one we generated or already existing ones
"results" is logic since we produce some data

src/solver/ts-generator/availability.cpp
payetvin: Need to be done in a separate PR, also is it even used ?

src/solver/ts-generator/availability.cpp
payetvin: Will do

payetvin: will do

src/tools/ts-generator/main.cpp
payetvin: writer is also needed for compatiblity with current "in memory" behavior

src/solver/ts-generator/availability.cpp
payetvin: will do

src/solver/ts-generator/availability.cpp
payetvin: will do

src/solver/ts-generator/availability.cpp
payetvin: will do

docs/user-guide/04-migration-guides.md
flomnes:
```suggestion
### (TS-generator only) TS generation for link capacities
```

docs/user-guide/04-migration-guides.md
```diff
@@ -1,5 +1,26 @@
...
+- law.planned (string "uniform"/"geometric")
+- law.forced (same)
+- volatility.planned (double in [0,1])
+- volatility.forced (same)
+
+- "prepro" timeseries => input/links/<link 1>/prepro/<link 2>_{direct, indirect}.txt, 365x6 values, respectively "forced outage duration", "planned outage duration", "forced outage rate", "planned outage rate", "minimum of groups in maintenance", "maximum of groups in maintenance".
+- "modulation" timeseries => input/links/<link 1>/prepro/<link 2>_mod_{direct, indirect}.txt, 8760x1 values each in [0, 1]
+- number of TS to generate => generaldata.ini/General/nbtimeserieslinks (unsigned int, default value 1)
+
+Add bool argument to optionally load time-series. This bool is false (default value) in antares-solver, and true in antares-ts-generator. Parameters are loaded in any case.-
```
flomnes:
```suggestion
```

flomnes (approved):

src/libs/antares/study/parameters.cpp
flomnes: See your PR @guilpier-code

src/solver/simulation/include/antares/solver/simulation/solver.hxx
guilpier-code: Done in #2200

src/tools/ts-generator/main.cpp
guilpier-code: In #2200, the result writer is removed, so that this comment is resolved now.

src/tools/ts-generator/main.cpp
guilpier-code: Removed in PR #2202

src/solver/ts-generator/thermal.cpp
guilpier-code: Done in PR #2171

src/tools/ts-generator/main.cpp
guilpier-code: Removed in #2200

src/solver/ts-generator/availability.cpp
guilpier-code: Fixed in #2200

src/tools/ts-generator/main.cpp
guilpier-code: `studyOptions.prepareOutput = true;` was removed in PR #2202

guilpier-code: `study->initializeRuntimeInfos();` was removed in #2202

src/tools/ts-generator/main.cpp
guilpier-code: Done in PR #2155

src/tools/ts-generator/main.cpp
guilpier-code: Done in PR #2155

src/tools/ts-generator/main.cpp
guilpier-code: Done in PR #2202

src/solver/ts-generator/availability.cpp
guilpier-code: Now renamed into **writeTStoDisk**. Done in some further PR.

src/solver/ts-generator/availability.cpp
guilpier-code: **noOutput** was removed from TS generation in PR #2202

src/solver/ts-generator/availability.cpp
guilpier-code: Done in a further and separate PR

guilpier-code: Done in #2202

src/libs/antares/study/include/antares/study/area/links.h
guilpier-code: This is the main purpose of #2155

src/tools/ts-generator/main.cpp
guilpier-code: Done in PR #2202

PR #1993, opened 2024-03-14
Expose API [ANT-1158]

JasonMarechal25:
Expose Antares Solver in a library. the library and target are called AntaresAPI. This API expose one entry point to run a simulation on a study. The return value of this is a data structure with the path to the simulation and the weekly problems.

- New component API
- Create a new project "api_client" and corresponding tests to detect API lib install regressions
- rename "utils" project in tests as "in-memory-study"
- promote this project at top level in tests. This allow in-memory-study to be used for any tests
- create SimulationObserver. Object passed throughout all "frames" of the simulation.


src/tests/src/api/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+        Boost::unit_test_framework
+        Antares::study
+        Antares::study-loader
+        Antares::tests::in-memory-study
+)
+
+target_include_directories(${EXECUTABLE_NAME}
+        PRIVATE
+        $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/api/private>
+        PUBLIC
```
sylvlecl: I guess you meant PRIVATE here

src/libs/antares/study-loader/include/antares/study-loader/IStudyLoader.h
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+#pragma once
+#include <antares/study/study.h>
+
+namespace Antares {
+class IStudyLoader {
+public:
+    virtual ~IStudyLoader() = default;
+    virtual std::shared_ptr<Antares::Data::Study> load() = 0;
```
sylvlecl: Since it's kind of a factory method, it should return a `unique_ptr`.
Then the caller may turn it into a `shared_ptr` if really necessary.

src/api/API.cpp
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#include "API.h"
+
+namespace Antares::API
+{
+SimulationResults APIInternal::run(IStudyLoader* study_loader) {
```
sylvlecl: `study_loader` should be a ref since it's required to be non-null.

src/api/API.cpp
```diff
@@ -0,0 +1,32 @@
...
+    study_ = study_loader->load();
```
sylvlecl: As it is now, we see that the loader has a much broader responsibility than just loading a study: it's also responsible for running a simulation.

We either need to rename it, or to restrict what it actually does.
From a high level view, one possibility seems to be to:
1. push down the study loader to `Application` function that runs the simulation
2. use the `Application` here in the `run` method

Another, probably better possibility would be to invert:
1. Move all "business logic" code from `Application` to another layer `Simulation` for example (similar to your APIInternal, but meant to be used not only by the public API)
2. Use the loader interface in that layer
3. Call this simulation layer from application, with the file tree study loader
4. Call also this simulation layer from API internal

src/tests/src/api/CMakeLists.txt
JasonMarechal25: No ?
Private includes are private and "include" dir is public for clients

src/libs/antares/study-loader/include/antares/study-loader/IStudyLoader.h
JasonMarechal25: Indeed it was my intent. The difficulty is that the "default" implementation FileTreeStudyLoader poses some difficulties for this.
In FileTreeStudyLoader "load" encapsulates the current "application.prepare" behaviour that create and load the study. This part, and most notably "load" passe around a shared_ptr. Difficulties are:
- It is not possible to transforme a shared_ptr into a unique else it would have only been a matter of returning the proper type.
- application prepare stack use shared_ptr. Changing for unique would mean changing all usage. Not difficult, just a bit of a tedious task
- application.prepare stack and inner mechanisms are opaque. It's not just all function call, else we could have just pass a pointer to the study with study_.get(). There are some objects that retains ownership of the study during their lifetime, justifying the use of shared_ptr. The lifetime of said objects probably doesn't exceed prepare() but since it's "black box" I didn't want to edit the interfaces and behaviour too much.

I'm still open to discussion and would very much prefer being able to return a unique_ptr

src/api/API.cpp
JasonMarechal25: Out dated.
New code for information and more inline with suggestion:
```
SimulationResults APIInternal::run(IStudyLoader* study_loader)
{
    study_ = study_loader->load();
    return execute();
}
```

src/tests/src/api/CMakeLists.txt
sylvlecl: Can't remember exactly why I wrote this, but indeed PUBLIC is not meaningful for executables, it's only useful for libraries: we will never use `target_link_library` with an executable.

src/api/include/antares/api/SimulationResults.h
```diff
@@ -0,0 +1,60 @@
...
+     */
+    std::filesystem::path simulationPath;
+    /**
+     * @brief weekly problems
+     */
+    Antares::Solver::LpsFromAntares antares_problems;
+    /**
+     * @brief An optional error that occurred during the simulation.
+     */
+    std::optional<Error> error;
```
sylvlecl, line 59: Weird, although the CI build passes, this doesn't compile in my env:
indeed `<optional>` is not included.

src/tests/src/api_internal/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+set(EXECUTABLE_NAME test-api)
+add_executable(${EXECUTABLE_NAME})
+
+target_sources(${EXECUTABLE_NAME}
+    PRIVATE
+    test_api.cpp
+    ${CMAKE_SOURCE_DIR}/api/API.cpp
```
sylvlecl: I think we should just link with solver_api lib, and not rebuild those cpp files.

src/tests/src/api_internal/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+        PRIVATE
+        Boost::unit_test_framework
+        Antares::study
+        Antares::study-loader
+        Antares::tests::in-memory-study
+        Antares::run-mode
+        Antares::lps
+)
+
+target_include_directories(${EXECUTABLE_NAME}
```
sylvlecl, line 16: Maybe a more elegant and re-usable way of including private headers of the tested lib, using its properties:
```cmake
target_include_directories(${EXECUTABLE_NAME}
    PRIVATE $<TARGET_PROPERTY:Antares::solver_api,INCLUDE_DIRECTORIES>)
```

See https://stackoverflow.com/a/39618354

The "public includes" of the lib don't have to be declared public for the executable, actually it is not meaningful for executables.

src/libs/antares/study-loader/include/antares/study-loader/IStudyLoader.h
sylvlecl: Ok let's keep it in mind for later days

src/api/include/antares/api/solver.h
```diff
@@ -0,0 +1,34 @@
...
+#include "SimulationResults.h"
+
+namespace Antares::API {
+/**
+ * @brief The PerformSimulation function is used to perform a simulation.
+ * @param study_path The path to the study to be simulated.
+ * @return SimulationResults object which contains the results of the simulation.
+ * @exception noexcept This function does not throw exceptions.
+ */
+SimulationResults PerformSimulation(std::filesystem::path study_path) noexcept;
```
sylvlecl: I would have expected the observer to go here in the public API instead, so that the client can decide what he does with intermediate data.

Then the implementation of `SimulationObserver` would go on the client side (xpansion).

src/api/include/antares/api/SimulationResults.h
```diff
@@ -0,0 +1,60 @@
...
+struct [[nodiscard("Contains results and potential error")]] SimulationResults
+{
+    /**
+     * @brief The path to the simulation (output).
+     */
+    std::filesystem::path simulationPath;
+    /**
+     * @brief weekly problems
+     */
+    Antares::Solver::LpsFromAntares antares_problems;
```
sylvlecl, line 55: See comment about public API:
I would expect this to not be part of the result, because it's linked to a very specific use case (xpansion), but to be collected via the observer only by clients who need it.

src/ext/yuni/src/yuni/core/system/cpu.h
```diff
@@ -1,3 +1,24 @@
+/*
```
sylvlecl: To be removed for consistency with other yuni headers.

src/api_client/README.md
```diff
@@ -0,0 +1,3 @@
+Client for antares solver API
+
+To demonstrate as much capabilities as possible this client should stay stand alone this mean not added as a sub_directory for cmake
```
sylvlecl: I wonder if we should move it to tests or not.

Or maybe be more explicit in the directory name, like "api_client_example", or having an intermediate "examples" subdirectory.

src/solver/simulation/include/antares/solver/simulation/solver.h
```diff
@@ -53,9 +54,10 @@ class ISimulation : public Impl
     ** \brief Constructor (with a given study)
     */
     ISimulation(Data::Study& study,
-                const ::Settings& settings,
+                const Settings& settings,
                 Benchmarking::IDurationCollector& duration_collector,
-                IResultWriter& resultWriter);
+                IResultWriter& resultWriter,
+                std::shared_ptr<Simulation::ISimulationObserver> simulationObserver);
```
sylvlecl: We must use a raw pointer here and all methods below:
the simulation will never own the observer, it's the caller of the simulation who is the clear owner of it, there is no shared ownership.

src/solver/simulation/include/antares/solver/simulation/ISimulationObserver.h
```diff
@@ -0,0 +1,65 @@
...
+ * @class ISimulationObserver
+ * @brief The ISimulationObserver class is an interface for observing the simulation.
+ * @details It declares the notifyHebdoProblem method.
+ */
+class ISimulationObserver
+{
+public:
+    virtual ~ISimulationObserver() = default;
+    /**
+     * @brief The notifyHebdoProblem method is used to notify of a problem during the simulation.
```
sylvlecl, line 40: We need to make clear that implementations need to be thread safe, because they will be called concurrently by computation threads.

src/api/SimulationObserver.cpp
```diff
@@ -0,0 +1,45 @@
...
+void SimulationObserver::notifyHebdoProblem(const PROBLEME_HEBDO* problemeHebdo,
+                                            int optimizationNumber,
+                                            std::string_view name)
+{
+    if (optimizationNumber != 1) return; //We only care about first optimization
+    Solver::HebdoProblemToLpsTranslator translator;
+    unsigned int const year = problemeHebdo->year + 1;
+    unsigned int const week = problemeHebdo->weekInTheYear + 1;
+    if (year == 1 && week == 1) {
+        lps_.replaceConstantData(translator.commonProblemData(problemeHebdo->ProblemeAResoudre.get()));
```
sylvlecl: Here you will need to make access to `lps_` thread safe with a mutex, because this code will be called concurrently from multiple threads.

src/solver/optimisation/include/antares/solver/optimisation/HebdoProblemToLpsTranslator.h
```diff
@@ -0,0 +1,82 @@
...
+    explicit HebdoProblemTranslationException(const std::string& string) noexcept;
+};
+
+/**
+ * @class HebdoProblemToLpsTranslator
+ * @brief Class for translating a weekly problem to a linear programming problem.
+ *
+ * This class is responsible for translating a weekly problem to a linear programming problem.
+ */
+class HebdoProblemToLpsTranslator
```
sylvlecl, line 50: It would make more sense to have plain functions here than a class: 
there is no undrlying state to maintain, we don't need an object here, it's only functions with an input and an output.

Nice ref about free functions :) : https://www.youtube.com/watch?v=WLDT1lDOsb4

docs/Architecture_Decision_Records/Remove_fetch_content.md
```diff
@@ -0,0 +1,25 @@
+# Remove usage of fetch_content for runtime dependencies
```
sylvlecl: We'd probably better do this in a preliminary PR

src/solver/lps/include/antares/solver/lps/LpsFromAntares.h
```diff
@@ -0,0 +1,142 @@
...
+#include <vector>
+
+namespace Antares::Solver
+{
+
+/**
+ * @struct ProblemHebdoId
+ * @brief The ProblemHebdoId struct is used to identify a weekly problem by year and week.
+ */
+struct ProblemHebdoId
```
sylvlecl: Let's keep english for our API --> Week(ly?)ProblemId

src/solver/lps/include/antares/solver/lps/LpsFromAntares.h
```diff
@@ -0,0 +1,142 @@
...
+// couplantes peuvent varier au cours du temps (ex: rendement d'une pompe à
+// chaleur qui varie selon la température, FlowBased ?, etc)
+/**
+ * @class ConstantDataFromAntares
+ * @brief The ConstantDataFromAntares class is used to store constant data across all weeks
+ * of Antares problems.
+ */
+struct ConstantDataFromAntares
+{
+    unsigned NombreDeVariables; // Mathématiquement : Nb colonnes de la matrice,
```
sylvlecl: Let's use english here too

src/libs/antares/study-loader/include/antares/study-loader/IStudyLoader.h
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2038

src/api/SimulationObserver.cpp
JasonMarechal25: I understand however I tend to prefer to stick to OOP paradigme in case I must stub behavior. I can do the change if required though because atm there is no stubing required

src/solver/optimisation/include/antares/solver/optimisation/HebdoProblemToLpsTranslator.h
JasonMarechal25: I understand however I tend to prefer to stick to OOP paradigme in case I must stub behavior. I can do the change if required though because atm there is no stubing required
(didn't watch the video yet)

docs/Architecture_Decision_Records/Remove_fetch_content.md
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2036

src/api/SimulationResults.cpp
flomnes, line 1: I don't think this file serves any purpose, though I could be wrong

src/api/include/antares/api/SimulationResults.h
```diff
@@ -0,0 +1,62 @@
...
+namespace Antares::API
+{
+/**
+ * @struct Error
+ * @brief The Error structure is used to represent an error that occurred during the simulation.
+ */
+struct Error {
+    /**
+     * @brief The reason for the error.
+     */
```
flomnes, line 37: enum or even `std::variant` ? For the client, it may be easier to decide what to do depending on that (recoverable error, etc.)

src/api/private/API.h
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#pragma once
+#include <filesystem>
+#include "antares/api/SimulationResults.h"
+#include "antares/study-loader/IStudyLoader.h"
+#include "antares/study/study.h"
```
flomnes: Maybe fwd decl for `Study` does the job here

src/api/private/SimulationObserver.h
```diff
@@ -0,0 +1,57 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#pragma once
+#include <antares/solver/simulation/ISimulationObserver.h>
+#include "antares/solver/lps/LpsFromAntares.h"
```
flomnes: Not homogenous, choose one and stick with it

src/api_client_example/tests/test.cpp
```diff
@@ -0,0 +1,40 @@
...
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#define BOOST_TEST_MODULE test_client_api
+#define BOOST_TEST_DYN_LINK
+
+#define WIN32_LEAN_AND_MEAN
```
flomnes:
```suggestion
```

src/api_client_example/tests/test.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include <boost/test/unit_test.hpp>
+#include "API_client.h"
+
+BOOST_AUTO_TEST_CASE(test_run) {
+    auto results = solve("dummy_study_test_client_api");
+    BOOST_CHECK(results.error);
+    BOOST_CHECK(!results.error->reason.empty());
+    auto c = results.error->reason;
+    std::cout << c << std::endl;
+    BOOST_CHECK(results.error->reason.find("Study") != std::string::npos);
```
flomnes, line 33: https://en.cppreference.com/w/cpp/string/basic_string/contains C++23 unfortunately

src/libs/antares/study/include/antares/study/parts/short-term-storage/cluster.h
```diff
@@ -1,23 +1,23 @@
...
+ * it under the terms of the Mozilla Public Licence 2.0 as published by
+ * the Mozilla Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Antares_Simulator is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
```
flomnes, line 18: Please fix your IDE's config

src/packaging/CMakeLists.txt
```diff
@@ -33,6 +35,7 @@ set(TARGET_LIBS #No alias
         exception
         benchmarking
         antares-solver-variable
+        lps
```
flomnes, line 38:
```suggestion
        antares-lps
```

src/solver/lps/LpsFromAntares.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+void LpsFromAntares::addWeeklyData(WeeklyProblemId id, WeeklyDataFromAntares&& data)
+{
+    weeklyProblems.emplace(id, std::move(data));
+}
+
+void LpsFromAntares::addWeeklyData(WeeklyProblemId id, const WeeklyDataFromAntares& data)
+{
+    weeklyProblems.emplace(id, data);
+}
```
flomnes: This is confusing. One version grabs ownership, the other does not. IMO we should keep only one.

src/solver/lps/LpsFromAntares.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+const WeeklyDataFromAntares& LpsFromAntares::weeklyData(WeeklyProblemId id) const
+{
+    auto it = weeklyProblems.find(id);
+    if (it == weeklyProblems.end())
+    {
+        return WeeklyDataFromAntares(); //TODO Better error handling
```
flomnes: Your error might be a `std::variant` of multiple possible errors
```
class WeekNotFound;
class EmptyProblem;
using Error = std::variant<WeekNotFound, EmptyProblem>;
```
So the client can use `std::visit` to handle errors.

src/solver/optimisation/HebdoProblemToLpsTranslator.cpp
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @param in The container from which to copy elements.
+ * @param out The container to which to copy elements.
+ */
+template<class T, class U>
+void copy(const T& in, U& out)
+{
+    std::ranges::copy(in, std::back_inserter(out));
+}
+}
```
flomnes, line 44: https://en.cppreference.com/w/cpp/algorithm/copy

Also usually the output is the 1st argument, like in `out = in`

src/solver/optimisation/HebdoProblemToLpsTranslator.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    if (problem->NombreDeContraintes > problem->IndicesDebutDeLigne.size()) {
+        throw WeeklyProblemTranslationException("ConstraintesCount exceed IndicesDebutDeLigne size");
+    }
+
+    if (problem->NombreDeContraintes > problem->NombreDeTermesDesLignes.size()) {
+        throw WeeklyProblemTranslationException("ConstraintesCount exceed NombreDeTermesDesLignes size");
+    }
+
+    auto ret = ConstantDataFromAntares();
```
flomnes:
```suggestion
    ConstantDataFromAntares ret;
```

src/solver/optimisation/include/antares/solver/optimisation/base_weekly_optimization.h
```diff
@@ -40,18 +41,23 @@ class WeeklyOptimization
...
-                                IResultWriter& writer);
+                                IResultWriter& writer,
+                                Simulation::ISimulationObserver& simulationObserver
+                                );
     Antares::Solver::Optimization::OptimizationOptions options_;
     PROBLEME_HEBDO* const problemeHebdo_ = nullptr;
     Antares::Data::AdequacyPatch::AdqPatchParams& adqPatchParams_;
     const uint thread_number_ = 0;
     IResultWriter& writer_;
+    std::reference_wrapper<Simulation::ISimulationObserver> simulationObserver_;
```
flomnes, line 61: Why not
```suggestion
    Simulation::ISimulationObserver& simulationObserver_;
```
?

src/solver/simulation/include/antares/solver/simulation/adequacy.h
```diff
@@ -102,6 +101,7 @@ class Adequacy
     Matrix<> pRES;
     IResultWriter& resultWriter;
 
+    std::reference_wrapper<Simulation::ISimulationObserver> simulationObserver_;
```
flomnes, line 104: Why not just a reference ?

src/tests/src/api_internal/test_api.cpp
```diff
@@ -0,0 +1,90 @@
...
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#define BOOST_TEST_MODULE test api
+#define BOOST_TEST_DYN_LINK
+
+#define WIN32_LEAN_AND_MEAN
```
flomnes:
```suggestion
```

src/tests/src/api_lib/test_api.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#define BOOST_TEST_MODULE test api
+#define BOOST_TEST_DYN_LINK
+
+#define WIN32_LEAN_AND_MEAN
```
flomnes:
```suggestion
```

src/tests/src/solver/lps/test_lps.cpp
```diff
@@ -0,0 +1,76 @@
...
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#define BOOST_TEST_MODULE test translator
+#define BOOST_TEST_DYN_LINK
+
+#define WIN32_LEAN_AND_MEAN
```
flomnes:
```suggestion
```

src/tests/src/solver/optimisation/translator/test_translator.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include "antares/solver/optimisation/opt_period_string_generator_base.h"
+
+using namespace Antares::Solver;
+
+class StubOptPeriodStringGenerator : public OptPeriodStringGenerator {
+public:
+    std::string to_string() const override {
+        return "Plop";
+    }
+};
```
flomnes: ~~Unused ?~~ Unused

src/api/SimulationObserver.cpp
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+#include "SimulationObserver.h"
+#include "antares/solver/optimisation/HebdoProblemToLpsTranslator.h"
+
+namespace Antares::API
+{
+namespace
+{
+auto translate(const PROBLEME_HEBDO* problemeHebdo,
```
flomnes:
```suggestion
auto translate(const PROBLEME_HEBDO& problemeHebdo,
```

src/api/private/SimulationObserver.h
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @class SimulationObserver
+ * @brief The SimulationObserver class is used to observe the simulation.
+ * @details It inherits from the ISimulationObserver interface and overrides the notifyHebdoProblem method.
+ */
+class SimulationObserver: public Solver::Simulation::ISimulationObserver
+{
+public:
+    /**
+     * @brief The notifyHebdoProblem method is used to notify of a problem during the simulation.
```
flomnes: Please make this clear. Before the optimization, after the construction of the problem

src/api/solver.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "antares/file-tree-study-loader/FileTreeStudyLoader.h"
+
+namespace Antares::API {
+
+SimulationResults PerformSimulation(std::filesystem::path study_path) noexcept
+{
+    try {
+        APIInternal api;
+        auto study_loader = std::make_unique<FileTreeStudyLoader>(study_path);
+        return api.run(*study_loader.get());
```
flomnes:
```suggestion
        FileTreeStudyLoader study_loader(study_path);
        return api.run(study_loader);
```

src/api/API.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return execute();
+}
+
+/**
+ * @brief The execute method is used to execute the simulation.
+ * @return SimulationResults object which contains the results of the simulation.
+ *
+ * This method is a copy of Application::execute
+ */
```
flomnes: with some modifications

src/api/SimulationObserver.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+}
+void SimulationObserver::notifyHebdoProblem(const PROBLEME_HEBDO* problemeHebdo,
+                                            int optimizationNumber,
+                                            std::string_view name)
+{
+    if (optimizationNumber != 1) return; //We only care about first optimization
+    Solver::HebdoProblemToLpsTranslator translator;
+    unsigned int const year = problemeHebdo->year + 1;
+    unsigned int const week = problemeHebdo->weekInTheYear + 1;
```
flomnes:
```suggestion
    const unsigned int year = problemeHebdo->year + 1;
    const unsigned int week = problemeHebdo->weekInTheYear + 1;
```

flomnes (commented):
- Rename `notifyHebdoProblem` -> `notify`
- Since the direct Sirius use will be removed at some point in the future, does this work with OR-Tools only ?

flomnes:
Please apply formatting & resolve conflicts

flomnes (commented):
LGTM. Remove unused functions among `setConstantData` and addWeeklyData

src/solver/misc/options.cpp
```diff
@@ -45,7 +45,7 @@ std::unique_ptr<Yuni::GetOpt::Parser> CreateParser(Settings& settings, StudyLoad
 {
     settings.reset();
 
-    auto parser = std::unique_ptr<Yuni::GetOpt::Parser>(new Yuni::GetOpt::Parser());
+    auto parser = std::make_unique<GetOpt::Parser>();
```
JasonMarechal25:
```suggestion
    auto parser = std::make_unique<Yuni::GetOpt::Parser>();
```

flomnes (approved):

PR #1996, opened 2024-03-15
Lay the groundwork for Scenarization@POC

flomnes:
This PR works by introducing a new sub-class `LinearProblemData::YearView`. The role of this class is to reference the time-series / scalar data for a given year.

We build `YearView` instances using the following correspondences
```cpp
unsigned tsIndex = data.groupYearToIndex_.at(group).at(year);
double scalarDataForScenario = scalarData[tsIndex];
```

Then we replace most instances of `LinearProblemData` by `LinearProblemData::YearView` so that `addXXX` methods only see what they need to use.

Please note that at least 50% of changes is just formatting. Sorry about that.

### TODO
- [x] Construct groups
- [x] Add test with scenarized data & groups
- [ ] Make `LinearProblemBuilder` re-usable ?

### DIFF
--- a/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
+++ b/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
@@ -96,19 +96,17 @@ struct SimplexResult
     mpsWriterFactory mps_writer_factory;
 };
 
-static SimplexResult OPT_TryToCallSimplex(
-        const OptimizationOptions& options,
-        PROBLEME_HEBDO* problemeHebdo,
-        Optimization::PROBLEME_SIMPLEXE_NOMME& Probleme,
-        const int NumIntervalle,
-        const int optimizationNumber,
-        const OptPeriodStringGenerator& optPeriodStringGenerator,
-        bool PremierPassage,
-        IResultWriter& writer)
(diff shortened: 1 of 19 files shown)


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -226,21 +226,24 @@ static SimplexResult OPT_TryToCallSimplex(
...
 
         // sinon renvoyer le builder ou le problem à une autre classe
         // Required for the balance constraint indices
-        gLinearProblemData.legacy.constraintMapping = &problemeHebdo->CorrespondanceCntNativesCntOptim;
+        gLinearProblemData.legacy.constraintMapping
+          = &problemeHebdo->CorrespondanceCntNativesCntOptim;
         gLinearProblemData.legacy.areaNames = &problemeHebdo->NomsDesPays;
         // TODO : add data here
-        linearProblemBuilder.build(gLinearProblemData);
+        linearProblemBuilder.build(gLinearProblemData[problemeHebdo->year]);
```
flomnes, line 242: Assuming `gLinearProblemData` contains the appropriate data, scenarization of new components should work natively for legacy problems.

flomnes:
See #2013

PR #2002, opened 2024-03-20
Add version 9.1

payetvin:


src/libs/antares/study/version.cpp
```diff
@@ -26,6 +26,8 @@
 using namespace Yuni;
 using namespace Antares::Data;
 
+static_assert(ANTARES_VERSION_LO == 1);
+
```
```suggestion
```

flomnes (commented):
Change src/CMakeLists.txt/ANTARES_VERSION_LO

flomnes (approved):

PR #2007, opened 2024-03-26
STS groups : make group categories dynamic rather than fixed 

guilpier-code:
We make groups of **S**hort **T**erm **S**torage dynamic. 
This means that accepted groups are no longer fixed in the code. Instead, any group can be defined by user.
It has consequences on the **Thematic Trimming**, which now accepts only one item (namely **STS BY GROUP**) regarding STS groups output variables.
Turning it to on prints all variables related to groups. 

Note : 
- We added a default group name (in case user doesn't specify a group name for a cluster) : "**sts-def-group**". Feel free to amend it.
- A new test (a study) was created. Please find it  [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/files/14802896/ST-dynamic-groups.zip). Look at its **user's notes** to know its intentions.
- upon a @sylvlecl suggestion, doc was updated accordingly

## Required
- [x] Merge https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR/pull/18/files, generate new reference results


src/libs/antares/study/include/antares/study/parts/short-term-storage/properties.h
```diff
@@ -64,14 +64,16 @@ class Properties
...
+    std::string groupName;
     /// cluster name
     std::string name;
 
     /// Enabled ?
     bool enabled = true;
 
-    static const std::map<std::string, enum Group> ST_STORAGE_PROPERTY_GROUP_ENUM;
 private:
+    static const std::map<std::string, enum Group> GROUP_TO_ENUM;
```
flomnes:
```suggestion
```

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -100,8 +101,10 @@ bool Properties::loadKey(const IniFile::Property* p)
...
     if (p->key == "group")
     {
-        if (auto it = Properties::ST_STORAGE_PROPERTY_GROUP_ENUM.find(p->value.c_str());
-            it != Properties::ST_STORAGE_PROPERTY_GROUP_ENUM.end())
+        this->groupName = p->value.c_str();
+
+        if (auto it = Properties::GROUP_TO_ENUM.find(p->value.c_str());
+            it != Properties::GROUP_TO_ENUM.end())
         {
             this->group = it->second;
```
flomnes:
```suggestion
return true;
```

src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -121,7 +124,10 @@ void Properties::save(IniFile& ini) const
 
     s->add("name", this->name);
 
-    for (const auto& [key, value] : ST_STORAGE_PROPERTY_GROUP_ENUM)
+    // gp : we'll have to print simply (instead of subsequent already existing lines) :
+    // s->add("group", this->groupName);
+
+    for (const auto& [key, value] : GROUP_TO_ENUM)
```
flomnes:
```suggestion
```

src/solver/variable/include/antares/solver/variable/economy/all.h
```diff
@@ -148,10 +151,13 @@ typedef                           // Prices
                      <Overflows        // Hydraulic overflows
                       <WaterValue      // Water values
                        <HydroCost      // Hydro costs
-                        <ShortTermStorageByGroup<STstorageInjectionByCluster<
-                          STstorageWithdrawalByCluster<STstorageLevelsByCluster<
-                          STstorageCashFlowByCluster<
-                            UnsupliedEnergy           // Unsuplied Energy
+                        // <ShortTermStorageByGroup
```
flomnes, line 149:
```suggestion
                         <ShortTermStorageByGroup
```

src/solver/variable/include/antares/solver/variable/economy/varDynamic.h
flomnes, line 1: Remove

src/libs/antares/study/include/antares/study/parts/short-term-storage/properties.h
guilpier-code: Done

src/libs/antares/study/parts/short-term-storage/properties.cpp
guilpier-code: Done

src/libs/antares/study/parts/short-term-storage/properties.cpp
guilpier-code: Done

src/solver/variable/include/antares/solver/variable/economy/all.h
guilpier-code: Done

src/solver/variable/include/antares/solver/variable/economy/varDynamic.h
guilpier-code: Done

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,379 @@
...
+        for (unsigned int column = 0; column < nbColumns_; column++)
+        {
+            results.variableCaption = caption(column);
+            results.variableUnit = unit(column);
+            pValuesForTheCurrentYear[numSpace][column]
+                .template buildAnnualSurveyReport<VCardType>(results, fileLevel, precision);
+        }
+    }
+
+    void buildSurveyReport(SurveyResults& results,
```
guilpier-code, line 366: **buildSurveyReport** is called when building the columns related to the current output variable in **synthesis** results.
If **buildSurveyReport** method is defined in an output var class, it is called.
If not, defined, it is the **IVariable** class's  **buildSurveyReport** that is called.
This function was a great source of confusion during this dev.
So I have 2 remarks about it : 
- We **really** **really** should rename it, because it has an horrible name, unrelated to what it does.
- take a look at **IVariable**'s version of this function : it calls **info.h**'s **BuildSurveyReport** function, and this latter function calls function **setClusterCaption** which is such an horrible function containing many **if**s. We should define a **buildSurveyReport** in variable classes for which **categoryFileLevel** is **Category::de_res** or **Category::de_sts**, so we could remove these **if**s. Actually this would allow to remove the horrible **setClusterCaption**, and make enclosing **info.h** : **BuildSurveyReport** to be the default behavior.

guilpier-code:
CI failure :
The CI fails because some of our studies are not valid anymore.
Indeed, some of their data related **Thematic Trimming** are no more supported.
These studies for example specify **PSP_OPEN_INJECTION** as an output variable to be printed, and possibly 26 more keywords related to short term storage groups.
But now we allow only one item for all STS groups output variables to be printed  : **STS BY GROUP**.
So we need to update these failing studies' thematic trimming input.

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,379 @@
+
```
sylvlecl: Copyright header is missing

src/tests/end-to-end/simple_study/simple-study.cpp
```diff
@@ -318,7 +318,7 @@ BOOST_AUTO_TEST_CASE(sts_initial_level)
     props.reservoirCapacity = 100;
     props.efficiencyFactor = .9;
     props.initialLevel = .443;
-    props.group = Group::PSP_open;
+	props.groupName = std::string("Some STS group");
```
sylvlecl: Wrong indent

src/libs/antares/study/include/antares/study/parts/short-term-storage/properties.h
```diff
@@ -62,15 +62,14 @@ class Properties
     bool initialLevelOptim = false;
     /// Efficiency factor between 0 and 1
     double efficiencyFactor = 1;
-    /// Used to sort outputs
-    Group group = Group::Other1;
+    // Used to sort outputs
+    std::string groupName = "sts-def-group";
```
sylvlecl: Question to be tackled:
should we keep Other1 here instead, for backwards compatibility?

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
sylvlecl: Seems to make sense!
Let's not do it in that PR though :)

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,379 @@
...
+        // Get the number of years in parallel
+        pNbYearsParallel = study->maxNbYearsInParallel;
+        pValuesForTheCurrentYear = new VCardType::IntermediateValuesBaseType[pNbYearsParallel];
+
+        // Vector of group names the clusters of the are belong to.
+        // In this Group names appear only once
+        for (auto& cluster : area->shortTermStorage.storagesByIndex)
+        {
+            std::string groupName = cluster.properties.groupName;
+            if (std::find(groupNames_.begin(), groupNames_.end(), groupName) == groupNames_.end())
```
sylvlecl: Note that a `std::set` would have given you name unicity and sorting without this and the std::sort at the end (replaced by a copy).
(and with more efficiency than a linear search in a vector, but here it does not matter since vectors are very small)
Something like
```cpp
std::set<std::string> names;
for (auto& cluster : area->shortTermStorage.storagesByIndex)
{
    names.insert(cluster.properties.groupName);
}
groupNames_ = std::vector<>(names.begin(), names.end());
```

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,379 @@
...
+    Antares::Memory::Stored<double>::ConstReturnType retrieveRawHourlyValuesForCurrentYear(
+      unsigned int column,
+      unsigned int numSpace) const
+    {
+        return pValuesForTheCurrentYear[numSpace][column].hour;
+    }
+
+    inline uint64_t memoryUsage() const
+    {
+        uint64_t r = (sizeof(IntermediateValues) * nbGroups_ + IntermediateValues::MemoryUsage())
```
sylvlecl: I guess we should use nbColumns_ in this method instead?

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,379 @@
...
+    };
+
+    typedef IntermediateValues IntermediateValuesDeepType;
+    typedef IntermediateValues* IntermediateValuesBaseType;
+    typedef IntermediateValuesBaseType* IntermediateValuesType;
+
+}; // class VCard
+
+/*!
+** \brief Energy generated by short term storage clusters
```
sylvlecl: Variables related to short term storage groups ?

sylvlecl (commented):
A few minor comments

src/tests/end-to-end/simple_study/simple-study.cpp
guilpier-code: Not in my **IDE** ! 
It seems like when visualing the code on **gitHub**, a tab character is not equivalent to 4 spaces, but larger.

guilpier-code: Corrected

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
guilpier-code: Yes, I tried the **std::set** instead : it seemed indeed a good way to have code simpler.
But at some point (see function `std::string caption(unsigned int column) const`)  we need to get a variable name from a column number, and a **std::vector** is adapted in this situation.
So we still could have a **std::set** instead, but we may need another container for associating a string to a number.
Of course, I could miss something. Tell me if there's better to do.

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
guilpier-code: Good eyes !
Corrected.
Even if we should remove this RAM usage, useless because only used in **GUI**, hopefully soon removed.
This memory usage consideration can be found in several places in the code.
Not sure removing it would be expensive, and many lines would vanish.
To me, it would worth it.

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
guilpier-code: Done

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
guilpier-code: Done

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
guilpier-code: ok, I understand your point better.
I made the simplifications you asked.

src/libs/antares/study/area/list.cpp
```diff
@@ -778,6 +778,7 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
     area.spreadSpilledEnergyCost = 0.;
 
     bool ret = true;
+    auto studyVersion = study.header.version;
```
```suggestion
    const auto studyVersion = study.header.version;
```

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+
+private:
+    //! Intermediate values for each year
+    typename VCardType::IntermediateValuesType pValuesForTheCurrentYear;
+    size_t nbColumns_ = 0;
+    std::vector<std::string> groupNames_; // Names of group containing the clusters of the area
+    std::map<std::string, unsigned int> groupToNumbers_; // Gives to each group (of area) a number
+    const std::vector<std::string> VAR_POSSIBLE_KINDS = { "injection", "withdrawal", "level" };
```
flomnes:
```suggestion
```

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,391 @@
...
+        r += sizeof(double) * nbColumns_ * maxHoursInAYear * pNbYearsParallel;
+        r += AncestorType::memoryUsage();
+        return r;
+    }
+
+    std::string caption(unsigned int column) const
+    {
+        std::string groupName = groupNames_[column / NB_COLS_PER_GROUP];
+        std::string variableKind = VAR_POSSIBLE_KINDS[column % NB_COLS_PER_GROUP];
+        return groupName + "_" + variableKind;
```
flomnes, line 317:
```suggestion
        static const std::vector<std::string> VAR_POSSIBLE_KINDS = { "injection", "withdrawal", "level" };
        std::string groupName = groupNames_[column / NB_COLS_PER_GROUP];
        std::string variableKind = VAR_POSSIBLE_KINDS[column % NB_COLS_PER_GROUP];
        return groupName + "_" + variableKind;
```

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,391 @@
...
+    {
+        // Get the number of years in parallel
+        pNbYearsParallel = study->maxNbYearsInParallel;
+        pValuesForTheCurrentYear = new VCardType::IntermediateValuesBaseType[pNbYearsParallel];
+
+        // Building the vector of group names the clusters belong to.
+        std::set<std::string> tmp_names;
+        for (auto& cluster : area->shortTermStorage.storagesByIndex)
+            tmp_names.insert(cluster.properties.groupName);
+        groupNames_ = { tmp_names.begin(), tmp_names.end() };
```
flomnes: Create a function for that

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,391 @@
...
+        
+        // Giving a number to each group
+        unsigned int groupNumber{0};
+        for (auto name : groupNames_)
+        {
+            groupToNumbers_[name] = groupNumber;
+            groupNumber++;
+        }
```
flomnes: Create a function for that

flomnes (commented):
Create functions, this will
- Not pollute the scope with temporary variables
- Allow unit-testing

src/libs/antares/study/include/antares/study/parts/short-term-storage/properties.h
flomnes:
```suggestion
    std::string groupName = "Other 1";
```

src/solver/variable/include/antares/solver/variable/adequacy/all.h
```diff
@@ -210,8 +210,6 @@ typedef // Prices
                                           Variable::Economy::WaterValue,
                                           Common::SpatialAggregate<
                                             Variable::Economy::HydroCost,
```
flomnes, line 212: What about districts ?

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -0,0 +1,406 @@
...
+        return r;
+    }
+
+    std::string caption(unsigned int column) const
+    {
+        static const std::vector<std::string> VAR_POSSIBLE_KINDS
+          = {"INJECTION", "WITHDRAWAL", "LEVEL"};
+        std::string groupName = groupNames_[column / NB_COLS_PER_GROUP];
+        std::string variableKind = VAR_POSSIBLE_KINDS[column % NB_COLS_PER_GROUP];
+        return groupName + "_" + variableKind;
```
flomnes:
```suggestion
        const std::string& groupName = groupNames_[column / NB_COLS_PER_GROUP];
        const std::string& variableKind = VAR_POSSIBLE_KINDS[column % NB_COLS_PER_GROUP];
        return groupName + "_" + variableKind;
```

flomnes (approved):

PR #2011, opened 2024-03-29
Fix style for STS by group

flomnes:
- Remove magic values (0, 1, 2), use `enum`
- Add a few `const` qualifiers
- Add const reference to results to improve readibility
- Formatting

### DIFF
--- a/src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
+++ b/src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
@@ -44,12 +44,12 @@ struct VCardSTSbyGroup
     }
 
     //! The synhesis results
-    typedef Results<R::AllYears::Average<           // The average values throughout all years
-                    R::AllYears::StdDeviation<      // The standard deviation values throughout all years
-                    R::AllYears::Min<               // The minimum values throughout all years
-                    R::AllYears::Max<               // The maximum values throughout all years
+    typedef Results<R::AllYears::Average< // The average values throughout all years
+      R::AllYears::StdDeviation<          // The standard deviation values throughout all years
+        R::AllYears::Min<                 // The minimum values throughout all years
+          R::AllYears::Max<               // The maximum values throughout all years
             >>>>>
-        ResultsType;
+      ResultsType;
 
     //! The VCard to look for for calculating spatial aggregates
     typedef VCardSTSbyGroup VCardForSpatialAggregate;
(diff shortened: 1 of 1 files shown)


src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -359,30 +378,25 @@ class STSbyGroup : public Variable::IVariable<STSbyGroup<NextT>,
...
 
-
 private:
     //! Intermediate values for each year
     typename VCardType::IntermediateValuesType pValuesForTheCurrentYear;
     size_t nbColumns_ = 0;
     std::vector<std::string> groupNames_; // Names of group containing the clusters of the area
     std::map<std::string, unsigned int> groupToNumbers_; // Gives to each group (of area) a number
-    const std::vector<std::string> VAR_POSSIBLE_KINDS = { "injection", "withdrawal", "level" };
+    const std::vector<std::string> VAR_POSSIBLE_KINDS = {"injection", "withdrawal", "level"};
```
guilpier-code: You didn't move **VAR_POSSIBLE_KINDS** to **caption(...)** as you said you would ?

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -88,18 +88,23 @@ struct VCardSTSbyGroup
 ** \brief Variables related to short term storage groups
 */
 template<class NextT = Container::EndOfList>
-class STSbyGroup : public Variable::IVariable<STSbyGroup<NextT>,
-                                              NextT,
-                                              VCardSTSbyGroup>
+class STSbyGroup : public Variable::IVariable<STSbyGroup<NextT>, NextT, VCardSTSbyGroup>
 {
+private:
```
guilpier-code, line 93: There is already a **private** section. **VariableType** should be moved down there.
Furthermore, why not 
```C++
enum class VariableType { injection, withdrawal, level };
```
instead ?

guilpier-code (changes requested):

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
flomnes: Yes, but then we must make it convertible to int

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
flomnes: No, I leave up to you

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
guilpier-code: ok, then would 
```C++
enum VariableType { injection = 0, withdrawal, level };
```
suit ?

flomnes: More like
```cpp
enum class VariableType : std::size_t { injection = 0, withdrawal = 1, level = 2};
```

guilpier-code: ok, it seems good to me

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
flomnes: done

src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
flomnes: We would need to `static_cast` values, which we don't really want

PR #2013, opened 2024-03-29
Second try at scenarization

flomnes:
Following @pet-mit's remarks

### DIFF
--- a/src/solver/optim/api/include/antares/optim/api/LinearProblemData.h
+++ b/src/solver/optim/api/include/antares/optim/api/LinearProblemData.h
@@ -33,46 +33,72 @@
 
 namespace Antares::optim::api
 {
-    class LinearProblemData final
+class LinearProblemData final
+{
+public:
+    using ScalarDataDict = std::map<std::string, std::vector<double>>;
+
+    using TimedData = std::vector<double>;
+    using TimedDataDict = std::map<std::string, std::vector<TimedData>>;
+
+    double getScalarData(const std::string& key, int scenario = 0) const
     {
-    public:
-        using ScalarDataDict = std::map<std::string, double>;
+        return scalarData_.at(key)[scenario];
+    }
 
-        using TimedData = std::vector<double>;
-        using TimedDataDict = std::map<std::string, TimedData>;
+    const TimedData& getTimedData(const std::string& key, int scenario = 0) const
+    {
+        return timedData_.at(key)[scenario];
+    }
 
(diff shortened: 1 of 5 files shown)


src/solver/optim/api/include/antares/optim/api/LinearProblemData.h
```diff
@@ -27,45 +27,78 @@
...
-    class LinearProblemData final
+class LinearProblemData final
+{
+public:
+    using ScalarDataDict = std::map<std::string, std::vector<double>>;
+
+    using TimedData = std::vector<double>;
+    using TimedDataDict = std::map<std::string, std::vector<TimedData>>;
+
+    double getScalarData(const std::string& key, int scenario = 0) const
```
flomnes: Use a default value for compatibility, to be removed later

src/solver/optim/api/include/antares/optim/api/LinearProblemData.h
```diff
@@ -27,45 +27,74 @@
...
+    using TimedDataDict = std::map<std::string, std::vector<TimedData>>;
+
+private:
+    // TODO : timestamps or timesteps?
+    std::vector<int> timeStamps_;
+    int timeResolutionInMinutes_;
+
+    ScalarDataDict scalarData_;
+    TimedDataDict timedData_;
+    // TODO : handle scenarios, and data vectorized on scenarios, on time, or on both
```
pet-mit, line 51: remove this TODO

src/solver/optim/api/include/antares/optim/api/LinearProblemData.h
```diff
@@ -27,45 +27,74 @@
...
 namespace Antares::optim::api
 {
-    class LinearProblemData final
+class LinearProblemData final
+{
+public:
+    using ScalarDataDict = std::map<std::string, std::vector<double>>;
+
+    using TimedData = std::vector<double>;
+    using TimedDataDict = std::map<std::string, std::vector<TimedData>>;
```
pet-mit, line 42: If I understand well, this implicitely means that 1st dimension is scenario and 2nd dimension is time.
I find it hard to read & maintain in time and over different files. Is there a way to have a multi-key map instead? Or define an extra struct / class? Or at least explicit it in the name of the attribute?

flomnes: Sure !

PR #2017, opened 2024-04-02
STS groups : output titles to uppercase

guilpier-code:
Any STS group name defined by the user is converted to uppercase in output results.

### DIFF
--- a/src/libs/antares/study/parts/short-term-storage/properties.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/properties.cpp
@@ -21,6 +21,7 @@
 
 #include <antares/logs/logs.h>
 #include <stdexcept>
+#include <boost/algorithm/string.hpp>
 
 #include "antares/study/parts/short-term-storage/properties.h"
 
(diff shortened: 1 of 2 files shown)


src/libs/antares/study/parts/short-term-storage/properties.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include <antares/logs/logs.h>
 #include <stdexcept>
+#include <boost/algorithm/string.hpp>
```
flomnes:
```suggestion
#include <boost/algorithm/string/case_conv.hpp>
```
<[boost/algorithm/string/case_conv.hpp](https://www.boost.org/doc/libs/1_84_0/boost/algorithm/string/case_conv.hpp)>

flomnes (changes requested):
Include only what you use

src/libs/antares/study/parts/short-term-storage/properties.cpp
guilpier-code: Done

flomnes (approved):

PR #2019, opened 2024-04-02
Add I/O for links' TS generation

flomnes:
Add read for
- "modulation" timeseries => `input/links/<link 1>/prepro/<link 2>_{direct, indirect}.txt`
- "prepro" timeseries => `input/links/<link 1>/prepro/<link 2>_mod_{direct, indirect}.txt`
- link parameters => `tsgen_direct_XXX`, `tsgen_indirect_XXX`, with XXX = unitcount, nominalcapacity, etc.
- number of TS to generate => `generaldata.ini/General/nbtimeserieslinks`

Add `bool` argument to optionally load time-series. This bool is `false` (default value) in `antares-solver`, and `true` in `antares-ts-generator`. Parameters are loaded in any case.

De-couple `PreproThermal` and `ThermalCluster` (remove the dependency of the 1st to the 2nd).


docs/reference-guide/13-file-format.md
```diff
@@ -1,6 +1,25 @@
...
+- tsgen_indirect_XXX
+with XXX in
+- unitcount (unsigned int, default 1)
+- nominalcapacity (float)
+- law.planned (string "uniform"/"geometric")
+- law.forced (same)
+- volatility.planned (double in [0,1])
+- volatility.forced (same)
+
+- "modulation" timeseries => input/links/<link 1>/prepro/<link 2>_{direct, indirect}.txt, 365x6 values, respectively "forced outage duration", "planned outage duration", "forced outage rate", "planned outage rate", "minimum of groups in maintenance", "maximum of groups in maintenance".
```
guilpier-code: There may be a confusion here, I guess : 
- **prepro** data file is supposed to contain the 6 columns
- **modulation** data file is supposed to contain the sole column (and have **\_mod\_** in its name)

src/libs/antares/study/area/links.cpp
```diff
@@ -471,7 +540,15 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
         link.comments.clear();
         link.displayComments = true;
 
-        ret = link.loadTimeSeries(study, folder) && ret;
+        if (loadTSGen)
```
guilpier-code: There are thermal consideration here, in the middle of a link loading function.
As said above, we should do some renaming.

src/libs/antares/study/area/links.cpp
```diff
@@ -132,6 +133,35 @@ bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& fol
     filename.clear() << capacitiesFolder << SEP << with->id << "_indirect.txt";
     success = indirectCapacities.loadFromFile(filename, false) && success;
 
+    if (loadTSGen)
```
guilpier-code: Please see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2019#discussion_r1553263253) first.

`if (loadTSGen)` (and the parameter **loadTSGen**)  : 

To avoid to make this function more complex with this new **if** statement, we should create a new method with the new loading ability, and remove **loadTSGen** as argument. 
We could eventually have (one level higher) : 
```C++
bool AreaLink::loadTimeSeries(const StudyVersion version, const AnyString& folder, bool loadTSGen)
{
    if (version < StudyVersion(8, 2))
       return loadCapacities_below_v810(folder) && 
              loadParameters_below_v810(folder);
    
    // Loading data of format over v820
    ret = loadCapacities_over_v820(folder); // No more loadTSGen argument here
    ret = loadParameters_over_v820(folder) && ret;
    if (loadTSGen)
       ret = loadTSgenerator(folder) && ret;
}
```
Please note :  
- As we're in class **AreaLink**, we don't need to repeat **link** in methods' names
- loadTSGen disappears from loading functions calls, but is used one level higher.

src/libs/antares/study/area/links.cpp
```diff
@@ -111,7 +112,7 @@ bool AreaLink::linkLoadTimeSeries_for_version_below_810(const AnyString& folder)
     return true;
 }
 
-bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& folder)
+bool AreaLink::linkLoadTimeSeries_for_version_820_and_later(const AnyString& folder, bool loadTSGen)
```
guilpier-code: **bool loadTSGen** : 
As you may already know,  if legacy code wasn't what it is, we would **not** need this new parameter at all.
We would only call the code that reads TS generations parameters for links from the **TS generator**, and we wouldn't call it from the **solver**, or the **cleaner**.
Instead, we have many occurrences of this parameter here in the reading functions, and some others at the higher levels. 

At some point, we should :
- separate data loading from : 
   - data checking and correcting 
   - business rules objects creation.
- make objects dedicated to loading data  (for example for a link) a collaboration of smaller, more individual, loading data object (for properties, for TS about capacities, for TS about modulation, ...), each one individually  usable in any context.

Doing this, we could load data on demand, for instance from the **TS generator**.

src/libs/antares/study/area/links.cpp
```diff
@@ -471,7 +548,20 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
...
+            link.tsGenerationDirect.prepro
+              = std::make_unique<Data::PreproThermal>(id_direct, link.tsGenerationDirect.unitCount);
+
+            const std::string id_indirect
+              = std::string(link.with->id) + "/" + std::string(link.from->id);
+            link.tsGenerationIndirect.prepro = std::make_unique<Data::PreproThermal>(
+              id_indirect, link.tsGenerationIndirect.unitCount);
+        }
+
+        ret = link.loadTimeSeries(study, folder, loadTSGen) && ret;
```
guilpier-code: About `ret = link.loadTimeSeries(study, folder, loadTSGen) && ret;` : 
**study** is passed only for study version. We should pass the study version instead of the whole study.

src/libs/antares/study/area/links.cpp
```diff
@@ -406,7 +472,14 @@ bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const
         link.filterYearByYear = stringIntoDatePrecision(value);
         return true;
     }
-
+    if (loadTSGen)
+    {
+        const std::string key_(key); // Conversion from Yuni::String
```
guilpier-code: Converting the current function argument **const Yuni::String& key** into **std::string** would be easy.
We wouldn't need the **key_** local variable.
We could have a call like : 
```C++
if (!AreaLinksInternalLoadFromProperty(link, key.to<std::string>(), value, loadTSGen))
    ...
```
And why not convert the value parameter as well ?

src/libs/antares/study/area/links.cpp
```diff
@@ -406,7 +472,14 @@ bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const
...
+        if (key_.starts_with("tsgen_direct"))
```
guilpier-code: Same idea as above (splitting the catching properties function) : 

```C++
// Inifile
const IniFile::Property* p = s->firstProperty;
for (; p; p = p->next)
{
    key = p->key;
    key.toLower();
    value = p->value;
    bool found = catchLinkProperties(link, key, value));
    if (loadTSGen)
       found = catchLinkTSgenerationProperties(link, key, value) || found;
    if (!found) 
       logs.warning() << '`' << p->key << "`: Unknown property";
}
```
Function names were clarified and simplified.

src/libs/antares/study/area/links.cpp
```diff
@@ -25,6 +25,7 @@
 #include "antares/utils/utils.h"
 #include "antares/study/area/links.h"
```
guilpier-code, line 27: About :
```C++
struct LinkTsGeneration
```
- Here we mix two different things : TS generation and business rules (links). This **struct** belongs more to TS generation, and should not be defined in this file.
- By the way, this **struct** could be used for a thermal cluster as well. So there is a kind of code duplication here.
- On the **Data::ThermalLaw** fields : should be renamed differently, as it is about statistic laws, and not specifically thermal laws.
- On the field `std::unique_ptr<Data::PreproThermal> prepro = nullptr;` : 
   - We don't need to initialize it to **nullptr**, it's done by default
   - about type **PreproThermal** : same as above. As it can be used for link capacities, it looses it's thermal-only characteristic, and should be renamed.

src/libs/antares/study/area/links.cpp
```diff
@@ -471,7 +548,20 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
         link.comments.clear();
         link.displayComments = true;
 
-        ret = link.loadTimeSeries(study, folder) && ret;
+        if (loadTSGen)
+        {
```
guilpier-code: About block : 
```C++
if (loadTSGen)
{
    ...
}
```
Too many implementation details we don't need to know about at this level.
We should call a function instead : 
```C++
if (loadTSGen)
    initializeTSgenerationPrepro(link);
```

guilpier-code: Furthermore, here we initialize the **prepro**, and in a further step (right after) we load its data matrix.
As we're handling the **prepro** here, why not do it in one step ?
So for me this initialize step should be moved to the **link.loadTimeSeries(...)** step (see remark above).
This would remove an **if (loadTSGen)** block.

src/libs/antares/study/cleaner/cleaner-v20.cpp
```diff
@@ -359,7 +359,7 @@ bool listOfFilesAnDirectoriesToKeep(StudyCleaningInfos* infos)
                 logs.verbosityLevel = Logs::Verbosity::Warning::level;
                 // load all links
                 buffer.clear() << infos->folder << "/input/links/" << area->id;
-                if (not AreaLinksLoadFromFolder(*study, arealist, area, buffer))
+                if (not AreaLinksLoadFromFolder(*study, arealist, area, buffer, false))
```
guilpier-code, line 362: New **bool** argument should not be needed. See [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2019#discussion_r1553263253)

guilpier-code (changes requested):

docs/reference-guide/13-file-format.md
payetvin: done

src/libs/antares/study/area/links.cpp
payetvin: I don't see thermal consideration here, loadTSGen is used to know if we're loading the links via the solver or the ts generator tool

src/solver/ts-generator/thermal.cpp
```diff
@@ -60,7 +60,7 @@ ThermalInterface::ThermalInterface(Data::AreaLink::LinkTsGeneration& source,
  plannedVolatility(source.plannedVolatility),
  forcedLaw(source.forcedLaw),
  plannedLaw(source.plannedLaw),
- prepro(source.prepro),
+ prepro(source.prepro.get()),
```
guilpier-code, line 63: Maybe already said in the review of [PR 1986](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/1986), but **thermal.cpp** is a bad name for this file.
It mostly contains code about TS generation (and a certain kind of TS generation).
So we should rename it for example **availabilityTSgeneration.cpp** (maybe not the right name, by the main idea is here).
The code unrelated to TS generation (for example code specific to links or thermal clusters) should be moved elsewhere

guilpier-code (changes requested):

src/libs/antares/study/area/links.cpp
```diff
@@ -170,12 +200,12 @@ void AreaLink::overrideTransmissionCapacityAccordingToGlobalParameter(
     }
 }
 
-bool AreaLink::loadTimeSeries(const Study& study, const AnyString& folder)
+bool AreaLink::loadTimeSeries(const StudyVersion &version, const AnyString& folder, bool loadTSGen)
 {
-    if (study.header.version < StudyVersion(8, 2))
+    if (version < StudyVersion(8, 2))
```
payetvin, line 238: Add a case for version >= 9.1 ?
Currently handled by loadTSGen

src/libs/antares/study/area/links.cpp
flomnes: Good catch ! Done and it indeed simplifies the `loadTimeSeries` function a lot !

src/libs/antares/study/area/links.cpp
flomnes: Done !

src/libs/antares/study/area/links.cpp
flomnes: You're right, my way is to avoid creating new instances of `Yuni::String`. However, it's not the point of this PR to migrate **existing** `Yuni::String`s into `std::string`s

src/libs/antares/study/area/links.cpp
flomnes:
```C++
// Inifile
const IniFile::Property* p = s->firstProperty;
for (; p; p = p->next)
{
    key = p->key;
    key.toLower();
    value = p->value;
    bool found = catchLinkProperties(link, key, value));
    if (!found && loadTSGen) // DIFF HERE
       found = catchLinkTSgenerationProperties(link, key, value);
    if (!found) 
       logs.warning() << '`' << p->key << "`: Unknown property";
}
```

src/libs/antares/study/area/links.cpp
guilpier-code: The type **PreproThermal** is a thermal consideration, is it not ?
About **loadTSGen**, see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2019#discussion_r1553263253)

src/libs/antares/study/area/links.cpp

src/libs/antares/study/area/links.cpp
flomnes: Done, except for PreproThermal (TODO)

src/libs/antares/study/cleaner/cleaner-v20.cpp
flomnes: I removed this argument from most functions, but it would be a bit more work to do it here

src/libs/antares/study/area/links.cpp
flomnes: Renamed `PreproThermal` -> `PreproAvailability`.

src/libs/antares/study/area/links.cpp
guilpier-code: > Done and it indeed simplifies the `loadTimeSeries` function a lot !

I'm not sure it's done  : we still have occurrences of the parameters **loadTSGen**. 
But doing what I recommend would require some amount of work.

src/libs/antares/study/cleaner/cleaner-v20.cpp
guilpier-code: You're right, it would require more work.
But we should keep in mind that we don't need this **loadTSGen** fundamentally.
It's a workaround because the code has low quality.

src/solver/ts-generator/thermal.cpp
guilpier-code: Not taken care of yet. 
Is there a problem to perform this change ?

src/libs/antares/study/area/links.cpp
flomnes: Yes, I don't want to go too deep here

src/libs/antares/study/area/links.cpp
flomnes: Yes, done. I had to check dependencies carefully

src/libs/antares/study/cleaner/cleaner-v20.cpp
flomnes: Yes, the loading functions are not de-coupled from the validation functions.

src/solver/ts-generator/thermal.cpp
flomnes: [done](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2019/commits/b647031d0cf14beedfc0ac21afe6fdd5ef302c57)

src/solver/ts-generator/include/antares/solver/ts-generator/prepro.h
```diff
@@ -24,14 +24,15 @@
...
+#include <antares/solver/ts-generator/law.h>
 #include <memory>
 
 namespace Antares::Data
 {
 /*!
 ** \brief Thermal
 */
-class PreproThermal
+class PreproAvailability
```
guilpier-code, line 35: **PreproAvailability** : watch for max value, which is **thermalPreproMax**

src/libs/antares/study/area/links.cpp
```diff
@@ -406,20 +469,64 @@ bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const
...
+    if (checkPrefixed("volatility.planned"))
+        return value.to(out.plannedVolatility);
 
+    if (checkPrefixed("volatility.forced"))
+        return value.to(out.forcedVolatility);
+
+    return false;
+}
+
+bool handleTSGenKey(Data::AreaLink& link, const std::string& key, const String& value)
```
guilpier-code, line 507: Now arg **key** is an **std::string &**.
Do we need to define **key_** ?
If not, we should remove `const std::string key_(key);`

src/solver/ts-generator/include/antares/solver/ts-generator/prepro.h
```diff
@@ -104,9 +110,27 @@ class PreproThermal
     // max x DAYS_PER_YEAR
     Matrix<> data;
     // Parent thermal cluster
-    std::weak_ptr<const ThermalCluster> itsThermalCluster;
-}; // class PreproThermal
+    YString id;
+    unsigned int unitCount;
+}; // class PreproAvailability
+
+struct LinkTsGeneration
```
guilpier-code, line 117: I still believe that there is a kind of code duplication between this type **LinkTsGeneration** and its thermal cluster counterpart. 
But it also would be not immediate to make it disappear. So I understand if we give up for now.

guilpier-code, line 117: I still believe that there is some kind of code duplication between this type **LinkTsGeneration** and its thermal cluster counterpart. 
But it also would be not immediate to make it disappear. So I understand if we give up for now.

guilpier-code (changes requested):

src/solver/ts-generator/include/antares/solver/ts-generator/prepro.h
flomnes: [done](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2019/commits/62a87063ff4648c0b45ba6c3830a4ac0509a0715)

src/libs/antares/study/area/links.cpp
flomnes: [done](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2019/commits/07278105c09907bf73a2d219f38cfb034f09e708)

src/solver/ts-generator/include/antares/solver/ts-generator/prepro.h
flomnes: Right

flomnes: Right

guilpier-code (approved):

PR #2025, opened 2024-04-10
Update clang-format with newer options and decisions

sylvlecl:
- Add clang-format options added between versions 6 and 17
 - Update existing configuration following project decisions


a-zakir:
why not put the .clangformat at the root of the project?

sylvlecl:
Final set of options is moved to and applied in #2067

PR #2029, opened 2024-04-10
Update formatting with clang-format changes

sylvlecl:


a-zakir (approved):

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -234,12 +239,12 @@ public:
     } // End of onExecute() method
 };
 
-
 template<class ImplementationType>
-inline ISimulation<ImplementationType>::ISimulation(Data::Study& study,
-    const ::Settings& settings,
-    Benchmarking::IDurationCollector& duration_collector,
-    IResultWriter& resultWriter) :
+inline ISimulation<ImplementationType>::ISimulation(
```
guilpier-code, line 243: Here ISimulation constructor : 
We would like the first arg right after the opening "(", but in this case, all args cannot start at the same column : the **duration_collector** arg is too long and exceeds 100 columns.
As a consequence, the first arg is rejected at the next line, so that all args can start at the same column.
But all args start after 2 spaces. We could prefer more indentation, like 4 spaces. But the args would start at the same column as the initialization list. 
So why not 8 spaces ?

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -523,19 +540,19 @@ uint ISimulation<ImplementationType>::buildSetsOfParallelYears(
         // created
         bool refreshing = false;
         refreshing = pData.haveToRefreshTSLoad && (y % pData.refreshIntervalLoad == 0);
-        refreshing
-          = refreshing || (pData.haveToRefreshTSSolar && (y % pData.refreshIntervalSolar == 0));
-        refreshing
-          = refreshing || (pData.haveToRefreshTSWind && (y % pData.refreshIntervalWind == 0));
-        refreshing
-          = refreshing || (pData.haveToRefreshTSHydro && (y % pData.refreshIntervalHydro == 0));
+        refreshing = refreshing
```
guilpier-code, line 543: operator || at end of line rather than at the beginning ? (so that each choice starts at the same column)

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -602,18 +619,23 @@ uint ISimulation<ImplementationType>::buildSetsOfParallelYears(
...
 
     } // End of loop over years
 
     return maxNbYearsPerformed;
 }
 
 template<class ImplementationType>
-void ISimulation<ImplementationType>::allocateMemoryForRandomNumbers(randomNumbers& randomForParallelYears)
+void ISimulation<ImplementationType>::allocateMemoryForRandomNumbers(
+  randomNumbers& randomForParallelYears)
```
guilpier-code, line 638: Same comment as above : first arg next line but 8 spaces indentation

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -695,74 +722,86 @@ void ISimulation<ImplementationType>::computeRandomNumbers(randomNumbers& random
...
-                // Current area's hydro starting (or initial) level computation
-                // (no matter if the year is performed or not, we always draw a random initial
-                // reservoir level to ensure the same results)
-                if (isPerformed)
-                    randomForYears.pYears[indexYear].pReservoirLevels[areaIndex] = randomLevel;
-            }
-
-            areaIndex++;
-        }); // each area
+        study.areas.each(
```
guilpier-code, line 739: lambda first part rejected next line : is that what we want ?  Especially if it's not indented in comparison to the previous line

src/solver/optimisation/base_weekly_optimization.cpp
```diff
@@ -19,10 +19,12 @@
 ** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
 */
 
-#include <memory>
 #include "antares/solver/optimisation/base_weekly_optimization.h"
-#include "antares/solver/optimisation/weekly_optimization.h"
+
+#include <memory>
```
guilpier-code, line 24: External inclusion **\<memory\>** could come become internal inclusion

src/solver/optimisation/base_weekly_optimization.cpp
```diff
@@ -41,18 +43,30 @@ WeeklyOptimization::WeeklyOptimization(const OptimizationOptions& options,
...
 }
 
-std::unique_ptr<WeeklyOptimization> WeeklyOptimization::create(
-    const Antares::Data::Study& study,
-    const OptimizationOptions& options,
-    AdqPatchParams& adqPatchParams,
-    PROBLEME_HEBDO* problemeHebdo,
-    uint thread_number,
-    IResultWriter& writer)
+std::unique_ptr<WeeklyOptimization> WeeklyOptimization::create(const Antares::Data::Study& study,
```
guilpier-code, line 46: **create** function : good example where return type could be moved up, as it is quite large

src/solver/hydro/management/daily.cpp
```diff
@@ -546,9 +561,7 @@ void HydroManagement::prepareDailyOptimalGenerations(Solver::Variable::State& st
                                                      uint y,
                                                      Antares::Data::Area::ScratchMap& scratchmap)
 {
-    areas_.each(
-      [&](Data::Area& area) {
-          prepareDailyOptimalGenerations(state, area, y, scratchmap);
-          });
+    areas_.each([&](Data::Area& area)
```
guilpier-code, line 564: Here the first part of the lambda is not rejected to the next line. In some other files (see **solver.hxx** for instance), it is. 
Strange.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -20,31 +20,29 @@
...
 {
 #include "spx_definition_arguments.h"
 #include "spx_fonctions.h"
-
 #include "srs_api.h"
 }
 
-#include <antares/logs/logs.h>
-#include <antares/antares/fatal-error.h>
+#include <chrono>
```
guilpier-code, line 36: STL inclusion first ?

guilpier-code (changes requested):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
sylvlecl: Yes, here we need to do something by hand because of the "extern C" block above. Clang-format will not remove it alone.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
sylvlecl: I don't like this formatting too much either. However there is only ONE parameter `ContinuationIndent` which decides for all types of continuation indent.
We need to check it's acceptable to use 8 for everything.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
sylvlecl: Why not, again there is only one parameter for all operators (including assignment), so it will change quite a lot of code.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
sylvlecl: This is because otherwise the function body will be too far right, I don't think we can do much about this.

A possibility is to use [OuterScope](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#lambdabodyindentation) instead of signature for lambda body indentation, but it will lead to weird formatting in other cases, where the lambda body will be on the left, but argument parameters on the right:
```cpp
        uint areaIndex = 0;
        study.areas.each(toto,
                         tata,
                         [&](Data::Area& area)
        {
            // looking for the initial reservoir level (begining of the year)
            auto& min = area.hydro.reservoirLevel[Data::PartHydro::minimum];
            auto& avg = area.hydro.reservoirLevel[Data::PartHydro::average];
            auto& max = area.hydro.reservoirLevel[Data::PartHydro::maximum];

```

src/solver/optimisation/base_weekly_optimization.cpp
sylvlecl: Not sure I understand, this is a standard library header so I think it's expected to have it here.

src/solver/optimisation/base_weekly_optimization.cpp
sylvlecl: Except if we prefer to **always** have return type on the previous line, this will not be possible here, because it will anyway be necessary to have each argument on its line.
So it will not be needed to have the return type on its own line, because the line will already be under 100 chars.

src/solver/hydro/management/daily.cpp
sylvlecl: Agreed that the behaviour is strange, but it does not seem to be possible to change it unfortunately (except by disabling completely single line lambdas).

sylvlecl:
Final set of options is applied in #2067

PR #2030, opened 2024-04-10
Cleanup link variables includes

sylvlecl:
**Description**

Currently, some code is shared between link variables for economy and adequacy
in an ugly way: the same files are included with a different preprocessor
definition for `LINK_NAMESPACE`. This leads to ununderstandable code
and include errors.

**Solution**

Code is shared through standard template mechanisms and type definitions.
No more "include templates" used with different preprocessor definitions.


sylvlecl:
This is a prerequisite for the PR about includes ordering.

flomnes (approved):

PR #2031, opened 2024-04-11
Remove unnecessary ".inc.hxx" files

sylvlecl:
Replaces ugly includes which behaviour is guided by
preprocessor definition, with standard template
instantiation mechanisms.


sylvlecl:
Example of previous implementation:
```cpp
#define NEXTTYPE Economy::VariablesPerArea
#include "antares/solver/variable/area.inc.hxx"
```

This does not make sense, this is exactly what templates are for:
instead the method will be compiled with type = `Economy::VariablesPerArea`
naturally when using `Areas<Economy::VariablesPerArea>`.

sylvlecl:
Note that the only difference with previous implementation is that the methods moved from cpp to hxx will be compiled every time they are included instead of once.

This is not really a problem, it's already the case of all other methods in variables lib.

If we wanted to keep this behaviour, we could use explicit template instanciation in cpp file, I considered it was not worth it.

JasonMarechal25 (approved):

PR #2032, opened 2024-04-11
Simplify code for timer

### LINKED ISSUE #2014: Use singleton for timer
# TODO
- [ ] Remove class `NullDurationCollector`
- [ ] Remove `pDurationCollector` members, `duration_collector` arguments
- [ ] Initialize singleton (See logs.h / logs.cpp, use `extern` keyword in header file)
- [ ] Change syntax whenever required (see below)

# Current
```cpp
void f(Benchmarking::IDurationCollector& durationCollector)
{
    Benchmarking::Timer timer_wait;
    std::lock_guard guard(pZipMutex); // Wait
    timer_wait.stop();
    durationCollector.addDuration("zip_wait", timer_wait.get_duration());
}
```
# Proposed
```cpp
#include <duration-collector.hpp>

int main() {
  DurationCollector collector;
// Sweet sweet syntactic sugar
  collector("hello") << [] {
    using namespace std::chrono_literals;
    std::this_thread::sleep_for(200ms);
  };
// Lambdas can capture stuff, too
  int a;
  collector("world") << [&a] {
    a++;
  };
  collector.summary();
}
/* Possible output
hello	200
world	0
*/
```
## Possible implementation
 ```cpp
#include <functional>
#include <thread>
#include <chrono>
#include <iostream>
#include <map>

class DurationCollector {
public:
  using clock = std::chrono::steady_clock;

  std::map<std::string, long> times;

  struct OperationTimer {
    OperationTimer(DurationCollector& collector,
		   const std::string& key): key(key), collector(collector) {}
    
    DurationCollector& collector;
    const std::string key;
  };

  OperationTimer operator()(const std::string& key) {
    return OperationTimer(*this, key);
  }

  friend void operator<<(const OperationTimer& op, std::function<void(void)> f) {
    auto start_ = clock::now();
    f();
    auto end_ = clock::now();
    auto duration_ms = std::chrono::duration_cast<std::chrono::milliseconds>(end_ - start_).count();
    op.collector.times[op.key] += duration_ms;
  }
  void summary() {
    for (auto& [k, v] : times)
      std::cout << k << '\t' << v << '\n';
  }
};
```

### PR DESCRIPTION
**Pull request by payetvin**
close  #2014

- Remove `IDurationCollector` and `NullDurationCollector`, keep only `DurationCollector`
- Add new syntax

### CODE REVIEW

File: src/solver/simulation/include/antares/solver/simulation/solver.hxx [hunk 1]
```diff
@@ -906,6 +899,7 @@ void ISimulation<ImplementationType>::computeAnnualCostsStatistics(
             pAnnualStatistics.updateTime.addCost(s.averageUpdateTime);
         }
     }
+    durationCollector.summary();
```
sylvlecl: Here we make the assumption that no other simulation is run in the same process:
we don't want to apply this limitation on the library.

For example, at a later point we could be willing to run multiple simulations from python or any other client libs.
This line will print, for simulation 1, the times of all other running simulations.

src/libs/antares/timer/include/antares/timer/timer.h
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Antares
+{
+
+class DurationCollector
+{
+public:
+    using clock = std::chrono::steady_clock;
+
+    std::map<std::string, long> times;
```
sylvlecl: Must be private

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -167,10 +168,10 @@ public:
             simulation_->prepareClustersInMustRunMode(scratchmap, y);
 
             // 4 - Hydraulic ventilation
-            Benchmarking::Timer timer;
-            hydroManagement.makeVentilation(randomReservoirLevel, state[numSpace], y, scratchmap);
-            timer.stop();
-            pDurationCollector.addDuration("hydro_ventilation", timer.get_duration());
+            durationCollector("hydro_ventilation") << [&]
```
sylvlecl: All those calls could override timers from other simulations running in the same process, we don't want it

src/libs/antares/timer/include/antares/timer/timer.h
```diff
@@ -0,0 +1,71 @@
...
+        return OperationTimer(*this, key);
+    }
+
+    friend void operator<<(const OperationTimer& op, std::function<void(void)> f)
+    {
+        auto start_ = clock::now();
+        f();
+        auto end_ = clock::now();
+        auto duration_ms = std::chrono::duration_cast<std::chrono::milliseconds>(end_ - start_).count();
+        op.collector.times[op.key] += duration_ms;
```
sylvlecl: Note that we have a loss compared to current collector: here we only have the sum of durations, but not the number of executions.

src/libs/antares/timer/include/antares/timer/timer.h
```diff
@@ -0,0 +1,71 @@
...
+        DurationCollector& collector;
+        const std::string key;
+    };
+
+    OperationTimer operator()(const std::string& key)
+    {
+        return OperationTimer(*this, key);
+    }
+
+    friend void operator<<(const OperationTimer& op, std::function<void(void)> f)
```
sylvlecl: It looks it's an operator of class `OperationTimer` --> it should inside this class instead (will also avoid the need for friend).

src/libs/antares/timer/include/antares/timer/timer.h
sylvlecl: Access to the map needs to be synchronized with a mutex since it will be accessed concurrently from multiple threads.
Should probably go into a method of the collector.

In the end the `DurationCollector` class should look exactly like the one already defined in `DurationCollector.h` but with the addition of the `OperationTimer` facility.

sylvlecl (changes requested):
- The new collector class lacks some of the feature of the one already defined in `DurationCollector.h`, in particular thread safety 
- In the end I don't think having a global is a good idea: we will mix data of concurrent simulations, which may be a use case in the future. We could distinguish with simulation identifiers, but then there is no benefit anymore of having a global
- We should add some unit tests

src/solver/simulation/include/antares/solver/simulation/solver.hxx
payetvin: Use of summary here is for test purposes, the recorded times will be written in a file

src/libs/antares/benchmarking/include/antares/benchmarking/DurationCollector.h
```diff
@@ -30,34 +32,42 @@
...
-};
 
-class DurationCollector : public IDurationCollector
-{
-public:
-    DurationCollector() = default;
-    virtual ~DurationCollector() = default;
+    friend void operator<<(const OperationTimer& op, const std::function<void(void)>& f)
+    {
+        static std::mutex mutex;
```
flomnes: `static` is very bad. `static` means this object will be shared across threads, across all objects. If you really need it, define the mutex next to the `std::map`.

src/libs/antares/benchmarking/include/antares/benchmarking/DurationCollector.h
```diff
@@ -30,34 +32,40 @@
...
+        const std::string key;
+    };
+
+    OperationTimer operator()(const std::string& key)
+    {
+        return OperationTimer(*this, key);
+    }
+
+    friend void operator<<(const OperationTimer& op, const std::function<void(void)>& f)
+    {
```
flomnes: Move definitions to a .cpp file

flomnes:
I think we should offer 2 syntaxes
- The one with the lambdas when no mutex / RAII stuff is present
- A RAII syntax when `std::mutex` are aquired for example
```cpp
{
durationCollector::Timer timer("blah");
std::lock_guard lock(mutex);
// timer.stop();
// autres trucs
// std::lock_guard::~lock_guard => unlock mutex
// DurationCollector::~Timer => record elapsed time here
}
```
Because mutex `lock/unlock` should be avoided, and the code looks less clear than before in this case.
```cpp
pDurationCollector("zip_wait") << [&] {
    pZipMutex.lock(); // Wait
};

pDurationCollector("zip_write") << [&] {
// stuff here
};
pZipMutex.unlock();
```

~~Alternative~~
```cpp
pDurationCollector("zip_wait") << [&] {
    std::lock_guard lock(mutex);
    pDurationCollector("zip_write") << [&] {
    // stuff here
    };
};

```

@sylvlecl WDYT ?

flomnes (changes requested):
- Abandon alternative syntax (it doesn't work), keep lock/unlock
- Add unit tests

flomnes (approved):

flomnes (changes requested):
LGTM. Add unit tests and we can merge

src/tests/src/solver/simulation/test-store-timeseries-number.cpp
```diff
@@ -64,7 +64,7 @@ BOOST_AUTO_TEST_CASE(BC_group_TestGroup_has_output_file) {
 
     auto working_tmp_dir = CREATE_TMP_DIR_BASED_ON_TEST_NAME();
 
-    Benchmarking::NullDurationCollector nullDurationCollector;
+    Benchmarking::DurationCollector nullDurationCollector;
```
sylvlecl: We'd better rename those occurences since they are not "null" any more.

src/libs/antares/writer/zip_writer.cpp
```diff
@@ -85,32 +84,29 @@ void ZipWriteJob<ContentT>::writeEntry()
...
+            logErrorAndThrow("Error opening entry " + pEntryPath + " (" + std::to_string(ret) + ")");
+        }
+        int32_t bw = mz_zip_writer_entry_write(pZipHandle, pContent.data(), pContent.size());
+        if (static_cast<unsigned int>(bw) != pContent.size())
+        {
+            logErrorAndThrow("Error writing entry " + pEntryPath + "(written = " + std::to_string(bw)
+                                                    + ", size = " + std::to_string(pContent.size()) + ")");
+        }
+    };
+    pZipMutex.unlock();
```
sylvlecl: Here we need to handle the case where an exception is thrown: we need to catch exceptions, release the mutex, and re-throw.

Or let's not use the new syntax here, it may be simper to keep the old way.

src/libs/antares/benchmarking/include/antares/benchmarking/DurationCollector.h
```diff
@@ -30,34 +32,29 @@
...
-    void addDuration(const std::string& name, int64_t duration) override;
+    OperationTimer operator()(const std::string& key);
+
+    friend void operator<<(const OperationTimer& op, const std::function<void(void)>& f);
+
+    long getTime(const std::string& name) const;
 
 private:
-    std::map<std::string, std::vector<int64_t>> duration_items_;
+    std::map<std::string, std::vector<long>> duration_items_;
```
sylvlecl: Maybe keep int64_t which is more portable.

sylvlecl (changes requested):
There is a possible missing release on a mutex which we need to fix.

Also we should rename previously "null" collectors.

flomnes (approved):

flomnes (approved):

sylvlecl (approved):

PR #2033, opened 2024-04-12
3rd try at scenario building

flomnes:
- 🧹🗑️ Remove simple & legacy tests - they make it too hard to move forward with the interface
- 📜 Add basic scenarization, allowing for multiple scenarios within the same LP (💫 new !)
- 📏 Formatting according to `src/.clang-format`

TODO
- [ ] 💪 Adapt "standard" test-case for more than one single scenario


src/tests/poc/legacy-workflow/simple-study.cpp
pet-mit, line 1: we need to keep testing that the new fillers are compatible with the legacy problem. instead of deleting this, you can replace the new fillers it uses with the ones use in "new-workflow/standard-test"

src/solver/optim/api/include/antares/optim/api/LinearProblemData.h
```diff
@@ -27,45 +27,96 @@
...
+    // TODO : timestamps or timesteps?
+    int timeResolutionInMinutes_;
+    std::map<std::string, std::vector<double>> scalarData_;
+    std::map<std::string, std::vector<std::vector<double>>> timedData_;
+    // TODO : handle scenarios, and data vectorized on scenarios, on time, or on both
+public:
+    explicit LinearProblemData(
+      int timeResolutionInMinutes,
+      const std::map<std::string, std::vector<double>>& scalarData,
+      const std::map<std::string, std::vector<std::vector<double>>>& timedData) :
```
pet-mit, line 47:
```suggestion
      const std::map<std::string, std::vector<double>>& scalarDataPerScenario,
      const std::map<std::string, std::vector<std::vector<double>>>& timedDataPerScenario) :
```

src/tests/poc/new-workflow/standard-test.cpp
pet-mit, line 1: I think it would be interesting to add a new test, inspired by the ones here, to test multiple scenarios

pet-mit (changes requested):

src/tests/poc/legacy-workflow/simple-study.cpp
flomnes: Yes. I tried but ran into the following issue : ports are not implemented for `LegacyLinearProblem`/`LegacyLinearProblemFillerImpl`. So it's a bit hard to connect battery / thermal to legacy balance constraints. This is not related to scenarization, so I'll open another PR for this.

pet-mit: when using the ComponentFiller of "BALANCE" type, the used AddBalanceConstraint in it will hook the new balance terms to the legacy constraint (meaning you don't need a port, it is implicit). you only have to make sure the ComponentFiller has the same node name as the legacy node

src/solver/optim/api/include/antares/optim/api/MipSolution.h
```diff
@@ -32,40 +32,54 @@
...
         }
-    };
-}
+        return solution;
+    }
+
+    double getObjectiveValue()
+    {
+        return objectiveValue_;
+    }
```
flomnes, line 83: `objectiveValue_` is set at construction. I think it should be extracted after `Solve`. But then we have to manage a cache with synchronization issues, etc. So I suggest not caching the objective value for now.

pet-mit: MipSolution is constructed after Solve

PR #2038, opened 2024-04-16
IStudyLoader return unique_ptr as expected of a factory

JasonMarechal25:
Had to make Study and all encapsulated objects movable to do so


a-zakir (approved):

JasonMarechal25:
Finally change is smaller that anticipated and can be made directly inside #1993

PR #2039, opened 2024-04-16
Apply formatting, alternative

sylvlecl:
Other formatting proposition with following updates compared to initial proposition, following comments from @guilpier-code : 
- `AlwaysBreakAfterReturnType: TopLevelDefinitions`
- `BreakBeforeBinaryOperators: None`
- `ContinuationIndentWidth: 8`
- `IndentWrappedFunctionNames: false`


sylvlecl:
Final set of options is applied in #2067

PR #2044, opened 2024-04-22
Create & forward declare BasisStatus struct, reduce build time

flomnes:
This is a wrapper around 2 vectors of OR-Tools-defined enums BasisStatus. By forward-declaring this class in most places, linear_solver.h is not included, which could avoid compiling it 29 times.

```
373174 ms: /home/leclercsyl/sources/antares/antares-simulator/src/solver/optimisation/include/antares/solver/optimisation/opt_structure_probleme_a_resoudre.h (included 124 times, avg 3009 ms), included via:
[...]
347162 ms: /home/leclercsyl/sources/antares/or-tools/install/debug/include/ortools/linear_solver/linear_solver.h (included 133 times, avg 2610 ms), included via:
  29x: opt_structure_probleme_a_resoudre.h 
```

A side effect is a better separation of responsibilities by having a dedicated class. ~~I couldn't use `std::unique_ptr` with incomplete type, but this should be fixable if need be.~~ [done](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2044/commits/c9fbd2502b6dfc0d79a0b078f5a0b0e1fe2e4baf)

Also, add unit 1 test for basis extraction.


src/solver/utils/include/antares/solver/utils/named_problem.h
```diff
@@ -27,22 +27,21 @@
...
 #include <algorithm>
 #include <iterator>
-#include "ortools/linear_solver/linear_solver.h"
 
 namespace Antares
 {
 namespace Optimization
 {
+struct BasisStatus;
+
```
flomnes, line 36: Forward decl here

src/solver/optimisation/include/antares/solver/optimisation/opt_structure_probleme_a_resoudre.h
```diff
@@ -24,23 +24,32 @@
...
 struct OptimizationOptions
 {
     bool useOrtools;
     std::string solverName;
 };
+} // namespace Antares::Solver::Optimization
 
+namespace Antares::Optimization
+{
+struct BasisStatus;
```
flomnes: Fwd decl here

src/solver/utils/include/antares/solver/utils/basis_status.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Antares::Optimization
+{
+struct BasisStatus
+{
+    bool exists() const;
+    void setStartingBasis(operations_research::MPSolver* solver) const;
+    void extractBasis(const operations_research::MPSolver* solver);
+    using Status = operations_research::MPSolver::BasisStatus;
+    std::vector<Status> StatutDesVariables;
```
sylvlecl: We should make it a class and have private fields here, since they do not need to be accessed outside.

src/solver/optimisation/include/antares/solver/optimisation/opt_structure_probleme_a_resoudre.h
```diff
@@ -112,12 +118,9 @@ struct PROBLEME_ANTARES_A_RESOUDRE
 
     std::vector<bool> VariablesEntieres; // true = int, false = continuous
 
-private:
-  using BasisStatus = operations_research::MPSolver::BasisStatus;
 public:
```
sylvlecl: public not needed any more

src/tests/src/solver/utils/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+set(src_solver_optimisation "${CMAKE_SOURCE_DIR}/solver/optimisation")
+
+## BASIS STATUS
+set(EXECUTABLE_NAME tests-basis-status)
+add_executable(${EXECUTABLE_NAME} basis_status.cpp)
+
+
+target_include_directories(${EXECUTABLE_NAME}
+							PRIVATE
+						    "${src_solver_optimisation}"
```
sylvlecl: This should not be needed, include dir comes with linked library solverUtils.

src/solver/optimisation/include/antares/solver/optimisation/opt_structure_probleme_a_resoudre.h
```diff
@@ -23,24 +23,30 @@
...
 
+namespace Antares::Optimization
+{
+struct BasisStatus;
 }
 
 /* Le probleme a resoudre */
 struct PROBLEME_ANTARES_A_RESOUDRE
 {
+    PROBLEME_ANTARES_A_RESOUDRE();
```
sylvlecl: The type is now not copyable / not movable: if this is what we want, we should make it explicit by at least deleting copy constructor and assignment, see https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types for background

src/solver/utils/include/antares/solver/utils/basis_status.h
```diff
@@ -0,0 +1,41 @@
...
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+// NOTE
```
sylvlecl: Actually, since now we hide the status as implementation detail in this class, we could go further and just use our own enum here.
This way, we won't need to include `linear_solver.h` here either, this would be even more secure.

(this would also allow to use a plain BasisStatus member in optimization struct if we prefer)

sylvlecl (commented):
As reported, it indeed reduces a lot parsing time during build (clang analysis).

A few comments about implementation.

src/solver/utils/include/antares/solver/utils/basis_status.h
flomnes: Right, only required in the test

src/solver/utils/include/antares/solver/utils/basis_status.h
flomnes: I disagree, conversion is painful and adds complexity and errors

sylvlecl (approved):

src/solver/utils/include/antares/solver/utils/basis_status.h
```diff
@@ -20,30 +20,42 @@
...
+
 // NOTE
 // Try not to include this header too often, if possible, forward-declare BasisStatus
 // since linear_solver.h takes time to compile
 
-#include <vector>
-#include "ortools/linear_solver/linear_solver.h"
-
-namespace Test {
+namespace Test
```
JasonMarechal25, line 25: Why do we have a namespace Test in implementations?

PR #2048, opened 2024-04-23
Doc/add thermal heuristic

PPlessiez:
Added a new doc page for the thermal heuristic + minor re-classification of the doc.


docs/optimisation/02-thermal-heuristic.md
```diff
@@ -0,0 +1,82 @@
...
+## General presentation of the resolution steps
+
+The linearised resolution of the weekly adequacy problem[^1] is summarized in the diagram below:
+
+![Global diagram](global_diagram.png)
+
+The general principle of the linear resolution is that two iterations of the weekly optimisation problem will be solved:
+- A first resolution in which integer variables are either linearised, or completely removed, depending on the "unit commitment mode" chosen (step 1).
+- A second resolution in which the NODU per cluster is fixed, and only the production level of each thermal cluster is set (step 3).
+
```
flomnes, line 20: The NODU is not fixed, it's lower bounded

docs/optimisation/02-thermal-heuristic.md
```diff
@@ -0,0 +1,82 @@
...
+Between these two steps, a heuristic is used to fix the values of the integer variables which will be used in the second resolution (step 2).
+
+Finally, a final annual heuristic is applied on the NODUs of the whole year to resolve potential inconsistencies caused by the decoupling of weeks.
```
flomnes: Some economic variables are computed after this step (OV COST & OP COST)

docs/optimisation/02-thermal-heuristic.md
```diff
@@ -0,0 +1,82 @@
...
+
+The way steps 1 and 2 are performed depends on a parameter set by the user in the "Advanced parameters" section, which is the "unit-commitment mode". This parameter can take two values:
+
+- The "fast" unit-commitment mode uses simplified methods to accelerate the weekly problem solving, while providing a consistent approximated solution.
+- The "accurate" unit-commitment mode embeds a more complex approach, which provides solutions which are closer to the optimum, at the cost of longer calculation times.
```
flomnes, line 28: The "milp" unit-commitment mode introduced in v8.8.0 ensures an exact weekly resolution, at the cost of even longer and more variable computation times. Variability is caused by the tree exploration in the branch & bound algorithm.

docs/optimisation/02-thermal-heuristic.md
PPlessiez: In this introduction, I prefferred to go with the "approximation" that the NODU is fixed in the explanation, to make it clear that the purpose of the second iteration is to set the power output of each cluster. In reality, the NODU is not fixed, you're right, but in the second iteration they usually stick to their lower bound. This is explained clearly in the part detailing each heuristic, but I wanted to keep it simple in the introduction.

docs/optimisation/02-thermal-heuristic.md
PPlessiez: I added this precision.

docs/optimisation/02-thermal-heuristic.md
PPlessiez: I added this info

flomnes (approved):

PR #2054, opened 2024-04-24
Move Run<X>Mode in dedicated files

JasonMarechal25:
Decouple Application with RunSimulationIn<X>Mode. Useful if other entry points want to run the simulation


sylvlecl (commented):
I wonder if it's worth it to create again a new lib for this: maybe we could just move those methods to `simulation` lib ?
What do you think ?

src/tests/src/api/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+set(EXECUTABLE_NAME test-api)
+add_executable(${EXECUTABLE_NAME})
+
+target_sources(${EXECUTABLE_NAME}
+    PRIVATE
+    test_api.cpp
+        ${CMAKE_SOURCE_DIR}/api/API.cpp
+)
```
flomnes: Can't see any test_api.cpp in this PR, is this a mistake ?

src/solver/CMakeLists.txt
```diff
@@ -46,8 +46,9 @@ OMESSAGE("  :: application solver")
 
 set(exec_name "antares-solver")
 
-add_library(solver-lib INTERFACE)
-
+add_library(solver-lib INTERFACE
+        ${SRCS}
+)
```
flomnes: I have a feeling this will add the icon to solver-lib (see l. 26-42), when what we really want is adding to executable antares-solver. Am I wrong ?

src/solver/run_mode/adequacy.cpp
```diff
@@ -0,0 +1,50 @@
...
+    simulation.run();
+
+    if (!(settings.noOutput || settings.tsGeneratorsOnly))
+    {
+        Benchmarking::Timer timer;
+        simulation.writeResults(/*synthesis:*/ true);
+        timer.stop();
+        durationCollector.addDuration("synthesis_export", timer.get_duration());
+
+        info = simulation.getOptimizationInfo();
```
flomnes: ⚠️ Conflict with #2032

src/solver/run_mode/economy.cpp
```diff
@@ -0,0 +1,50 @@
...
+    simulation.run();
+
+    if (!(settings.noOutput || settings.tsGeneratorsOnly))
+    {
+        Benchmarking::Timer timer;
+        simulation.writeResults(/*synthesis:*/ true);
+        timer.stop();
+        durationCollector.addDuration("synthesis_export", timer.get_duration());
+
+        info = simulation.getOptimizationInfo();
```
flomnes: ⚠️ Conflict with #2032

src/tests/src/api/CMakeLists.txt
JasonMarechal25: Maybe cherry-pick error

src/solver/run_mode/adequacy.cpp
JasonMarechal25: So be it

JasonMarechal25:
> I wonder if it's worth it to create again a new lib for this: maybe we could just move those methods to `simulation` lib ? What do you think ?

Indeed. After discussion it is not necessary

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -303,9 +303,6 @@ void ISimulation<ImplementationType>::run()
     // Determine if we have to use the preprocessors at least one time.
     pData.initialize(study.parameters);
 
-    // Prepro only ?
-    ImplementationType::preproOnly = settings.tsGeneratorsOnly;
-
```
flomnes: Change of behavior here.

flomnes (changes requested):

src/solver/simulation/include/antares/solver/simulation/solver.hxx
JasonMarechal25: Thanks.
Maybe I mistook this for local assignation but its parentType::member assignation :/

flomnes (approved):

PR #2056, opened 2024-04-24
Refactor Application

JasonMarechal25:
Extract some functions for clarity

### DIFF
--- a/src/solver/application/application.cpp
+++ b/src/solver/application/application.cpp
@@ -120,6 +120,8 @@ void Application::startSimulation(Data::StudyLoadOptions options)
 
     readDataForTheStudy(options);
 
+    postParametersChecks();
+
     pStudy->initializeProgressMeter(pSettings.tsGeneratorsOnly);
     if (pSettings.noOutput)
         pSettings.displayProgression = false;
@@ -133,6 +135,34 @@ void Application::startSimulation(Data::StudyLoadOptions options)
     }
     else
         logs.info() << "  The progression is disabled";
+}
+void Application::postParametersChecks() const
+{ // Some more checks require the existence of pParameters, hence of a study.
+    // Their execution is delayed up to this point.
+    checkOrtoolsUsage(
+      this->pParameters->unitCommitment.ucMode, this->pParameters->ortoolsUsed, this->pParameters->ortoolsSolver);
+
+    checkSimplexRangeHydroPricing(this->pParameters->simplexOptimizationRange,
+                                  this->pParameters->hydroPricing.hpMode);
+
+    checkSimplexRangeUnitCommitmentMode(this->pParameters->simplexOptimizationRange,
+                                        this->pParameters->unitCommitment.ucMode);
+
+    checkSimplexRangeHydroHeuristic(this->pParameters->simplexOptimizationRange, this->pStudy->areas);
+
+    if (this->pParameters->adqPatchParams.enabled)
+        this->pParameters->adqPatchParams.checkAdqPatchParams(
+          this->pParameters->mode,
+                                                        this->pStudy->areas,
+                                                        this->pParameters->include.hurdleCosts);
+
+    bool tsGenThermal
+      = (0 != (this->pParameters->timeSeriesToGenerate & Antares::Data::TimeSeriesType::timeSeriesThermal));
+
+    checkMinStablePower(tsGenThermal, this->pStudy->areas);
+
+    checkFuelCostColumnNumber(this->pStudy->areas);
+    checkCO2CostColumnNumber(this->pStudy->areas);
 }
 
 void Application::prepare(int argc, char* argv[])
--- a/src/solver/application/application.cpp
+++ b/src/solver/application/application.cpp
@@ -97,7 +97,7 @@ void Application::handleOptions(const Data::StudyLoadOptions& options)
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/solver/application/application.cpp
```diff
@@ -113,71 +95,31 @@ void Application::prepare(int argc, char* argv[])
...
-                                                        pStudy->areas,
-                                                        pParameters->include.hurdleCosts);
-
-    bool tsGenThermal
-      = (0 != (pParameters->timeSeriesToGenerate & Antares::Data::TimeSeriesType::timeSeriesThermal));
-
-    checkMinStablePower(tsGenThermal, pStudy->areas);
-
-    checkFuelCostColumnNumber(pStudy->areas);
-    checkCO2CostColumnNumber(pStudy->areas);
```
flomnes: These checks disappeared, right ? Is this intentional ?

flomnes (changes requested):

src/solver/application/application.cpp
JasonMarechal25: Totally unintentional. Probably a merge/rebase mistake.

flomnes (changes requested):
I'm not a fan of `this->pMember`, the p prefix should make it obvious that it's a member variable.

JasonMarechal25:
> I'm not a fan of `this->pMember`, the p prefix should make it obvious that it's a member variable.

Me neither, but I'm juste extracting function here, am not the author of the content ^^

Edit: My bad, they have been inserted by copilot

src/solver/application/application.cpp
```diff
@@ -113,45 +95,49 @@ void Application::prepare(int argc, char* argv[])
...
     readDataForTheStudy(options);
 
-    // Some more checks require the existence of pParameters, hence of a study.
+    postParametersChecks();
+
+    pStudy->initializeProgressMeter(pSettings.tsGeneratorsOnly);
+    if (pSettings.noOutput)
+        pSettings.displayProgression = false;
+
+    if (pSettings.displayProgression)
```
a-zakir, line 129:
```suggestion

    if (pSettings.displayProgression =!pSettings.noOutput; pSettings.displayProgression)
```

src/solver/application/application.cpp
```diff
@@ -176,21 +163,48 @@ void Application::prepare(int argc, char* argv[])
...
-        pStudy->progression.saveToFile(filename, *resultWriter);
-        pStudy->progression.start();
-    }
-    else
-        logs.info() << "  The progression is disabled";
+    // Load the local policy settings
+    LocalPolicy::Open();
+    LocalPolicy::CheckRootPrefix(argv[0]);
+
+    Resources::Initialize(argc, argv);
```
a-zakir, line 177: seems that this method cares only about argv 0 it will be clearer if you give it a const char* or const std::string or something like that?

src/solver/application/application.cpp
```diff
@@ -176,21 +163,48 @@ void Application::prepare(int argc, char* argv[])
...
+    // Something like bind("--foo", options.foo);
+    // So that option.foo will be assigned <value>
+    // if the user provides --foo <value>.
+    // CAUTION
+    // The parser contains references to members of pSettings and options,
+    // don't de-allocate these.
+    auto parser = CreateParser(pSettings, options);
+    // Parse the command line arguments
+
+    handleParserReturn(parser.get());
```
a-zakir, line 193: check if error or print help

src/solver/application/application.cpp
flomnes: Not really
```cpp
if (argc < 1 or argv[0] == NULL)
{
    logs.error() << "Impossible to find the root folder";
    return;
}
```

src/solver/application/include/antares/application/application.h
```diff
@@ -110,6 +113,10 @@ class Application final : public Yuni::IEventObserver<Application, Yuni::Policy:
                        Benchmarking::IDurationCollector& duration_collector);
 
     void writeComment(Data::Study& study);
+    void startSimulation(Data::StudyLoadOptions options);
```
a-zakir: by value? is it moved later?

flomnes: Indeed, `startSimulation` makes a copy while `readDataForTheStudy` works on a non-const ref

```cpp
void Application::readDataForTheStudy(Data::StudyLoadOptions& options);

void Application::startSimulation(Data::StudyLoadOptions options)
{
// ...
    readDataForTheStudy(options);
// ..
}
```
It should work but I find it a bit strange. Good catch :

src/solver/application/application.cpp
a-zakir: is this the method responsability?

src/solver/application/application.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/2062

src/solver/application/application.cpp
JasonMarechal25: I will leave it as it is. I want to introduce minimal change of code in this PR and try to keep it as simply encapsulating code in new function

src/solver/application/application.cpp
JasonMarechal25: This a larger topic about CLI options management (and ini parameters)

flomnes (approved):

PR #2064, opened 2024-04-26
Try to remove yuni from inifile

guilpier-code:
What was done so far :
- Removal of type **IO::File::Stream**
- Splitting and refactoring ugly function **AnalyzeIniLine(...)**
- Replacing few **Yuni::string** with **std::string** : 
  - **IniFile::filename()** now returns an **std::string**
  - class **IniFile**'s data member **pFilename** is now a **std::string** (and was renamed into **filename_**)
  - Replacement of **AnalyzeIniLine(...)** with some other code removes uses of **Yuni::string** (in args and locally)
- Some more simplifications

Note that the code has reduced  (if we except the addition of new unit tests), and is more clear now.
Please see comments added here to ease the review.


src/libs/antares/inifile/inifile.cpp
```diff
@@ -237,29 +238,20 @@ uint IniFile::Section::size() const
 
 bool IniFile::open(const AnyString& filename, bool warnings)
 {
-    // clear
     clear();
-    pFilename = filename;
+    std::string filePath = filename.to<std::string>();
+    const unsigned int lineMaxSize = 2048;
```
flomnes:
```suggestion
```

src/libs/antares/inifile/inifile.cpp
```diff
@@ -237,29 +238,20 @@ uint IniFile::Section::size() const
...
 
-    // Load the file
-    IO::File::Stream file;
-    if (file.open(filename))
+    std::ifstream file(filePath);
+    if (file.is_open())
```
flomnes: Proposal
```cxx
IniFile::readStream(std::istream& in, bool warnings)
{
   // most generic implementation
}
IniFile::open(const std::filesystem::path& file, bool warnings) 
{
   std::ifstream f(path);
   if (f.is_open())
   {
      return readStream(f, warnings);
   }
}
```
So we can have easy in-memory tests with `std::istringstream`

See [this diagram](https://en.cppreference.com/w/cpp/io) for more details.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -289,26 +281,25 @@ void IniFile::saveToString(std::string& str) const
 bool IniFile::save(const AnyString& filename) const
 {
     logs.debug() << "  :: writing `" << filename << '`';
-    IO::File::Stream f;
-    if (f.openRW(filename))
+    
+    std::ofstream f(filename.to<std::string>());
+    if (f.good())
```
flomnes: Same proposal here, don't impose files for output, use `std::ostream`.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -44,8 +45,8 @@ static inline IniFile::Section* AnalyzeIniLine(const String& pFilename,
 
     uint bytesRead = 0;
     Type type = typeUnknown;
-    char* p = line;
-    const char* key = line;
+    char* p = &line[0];
+    const char* key = &line[0];
```
flomnes: Horrible

src/libs/antares/inifile/inifile.cpp
```diff
@@ -24,15 +24,16 @@
...
 #include <antares/io/statistics.h>
 #include <sstream>
+#include <fstream>
 
 using namespace Yuni;
 
 namespace Antares
 {
-static inline IniFile::Section* AnalyzeIniLine(const String& pFilename,
+static inline IniFile::Section* AnalyzeIniLine(const std::string filename,
```
payetvin: use reference or string_view

src/libs/antares/inifile/inifile.cpp
guilpier-code: Yes indeed, but I wanted to touch this function as little as possible, to avoid introducing any side effect.
Any suggestion would be welcome.

This function isn't exactly the most beautiful I've crossed :-)

src/libs/antares/inifile/inifile.cpp
guilpier-code: I couldn't do exactly as you asked.
Indeed, handling the case where the file could not open correctly cannot easily be done inside the new  **readStream(...)** function.
Not handling this case in the function seems consistent : this function receives a stream and checks its correctness (as you suggested, it's a more generic function).
Handling a file opening error inside this function would have no sense.

guilpier-code: Please note that I made the new function **readStream(...)** **private**.
This makes it difficult to test that function in isolation. And yet, **readStream(...)** is a method called only once, inside a **public** method  (named **open(...)**) of class **iniFile**. So it shouldn't be part of the **iniFile** interface, and should remain **private**.

To overcome this problem, the current class **iniFile** should be split.
For example we coud have a class reading sections from a generic stream instead of file.
And another class could handle the files aspects and use the first one, via its interface.

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
```diff
@@ -19,99 +19,19 @@
...
+#include "antares/inifile/inifile.h"
 #include <antares/logs/logs.h>
 #include <antares/io/statistics.h>
-#include <sstream>
 
 using namespace Yuni;
 
 namespace Antares
 {
-static inline IniFile::Section* AnalyzeIniLine(const String& pFilename,
```
guilpier-code, line 33: Dirty function **AnalyzeIniLine** was removed, and replaced with a clearer block of lines (please see method **readStream(...)**).

src/libs/antares/inifile/include/antares/inifile/inifile.h
```diff
@@ -310,13 +308,15 @@ class IniFile final
...
 private:
+    void readStream(std::istream& in_stream, std::string filePath);
+
     /*!
     ** \brief The last filename used for loading/saving
     **
     ** \internal This variable to keep const-correctness for the
     ** public API (cf save).
     */
-    mutable YString pFilename;
```
guilpier-code, line 320: **pFilename** is now a `std::string`, instead of a `YString`
**pFilename** was renamed into  **filename_**;

src/libs/antares/benchmarking/file_content.cpp
```diff
@@ -65,8 +65,6 @@ std::string FileContent::saveToBufferAsIni()
         for (const auto& [key, value] : content)
             section->add(key, value);
     }
-    std::string buffer;
-    ini.saveToString(buffer);
-    return buffer;
+    return ini.saveToString();
```
guilpier-code: method **iniFile::saveToString()** now returns an `std::string`. 
More readable like this.

src/libs/antares/inifile/include/antares/inifile/inifile.h
```diff
@@ -201,10 +199,10 @@ class IniFile final
     */
     bool save(const AnyString& filename) const;
 
-    void saveToString(std::string& str) const;
+    std::string saveToString() const;
```
guilpier-code: As already said, **saveToString()** now returns a `std::string`, instead of taking a reference to a std::string and then changing it inside the function's body. 
More readable like this.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
+{
+    boost::trim_left(line);
+    return line.empty();
+}
+
+bool IniFile::readStream(std::istream& in_stream)
+{
+    std::string line;
+    IniFile::Section* currentSection(nullptr);
+    while (std::getline(in_stream, line))
```
guilpier-code, line 194: Lines inside **while loop** are meant to replace old function **AnalyzeIniLine**.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
+            continue;
 
-        delete[] line;
+        logs.error() << "Invalid INI line format : '" << line << "'";
+        return false;
+    }
+    return true;
+}
 
-        if (read)
```
guilpier-code, line 284: To be honest, local variable **read** and call of `Statistics::HasReadFromDisk(...);` were removed.
They were hampering the refactoring and were of no use but for the legacy **GUI**.
As its name suggests, the latter function collects disk loading volume for the **GUI**.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -283,29 +259,25 @@ void IniFile::saveToString(std::string& str) const
     if (written != 0)
```
guilpier-code, line 257:
```C++
if (written != 0)
        Statistics::HasWrittenToDisk(written);
```
This could also be removed (see previous comment)

src/libs/antares/inifile/inifile.cpp
```diff
@@ -283,29 +259,25 @@ void IniFile::saveToString(std::string& str) const
...
+    if (f.good())
     {
-        pFilename = filename;
         uint64_t written = 0;
-
-        // save all sections
         for (auto* section = firstSection; section; section = section->next)
             section->saveToStream(f, written);
-
         if (written != 0)
```
guilpier-code: `Statistics::HasWrittenToDisk(written);` : same as above : should be removed.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -283,29 +259,25 @@ void IniFile::saveToString(std::string& str) const
...
 bool IniFile::save(const AnyString& filename) const
 {
     logs.debug() << "  :: writing `" << filename << '`';
-    IO::File::Stream f;
-    if (f.openRW(filename))
+    
+    std::ofstream f(filename.to<std::string>());
+    if (f.good())
     {
-        pFilename = filename;
```
guilpier-code, line 324: `pFilename = filename;` : was removed, as it does not seem to be useful to store the name of a file we're saving. 
And tests on saving studies seem to confirm this uselessness.

flomnes (changes requested):
As discussed
- `IniFile` -> `IniContent`
- Publish `readStream` and `saveToStream`
- Make tests using these 2 functions above

src/libs/antares/inifile/inifile.cpp
guilpier-code: **AnalyzeIniLine** disappeared, see comments further

src/libs/antares/inifile/inifile.cpp
flomnes: `AnalyzeIniLine` was removed

src/libs/antares/inifile/inifile.cpp
flomnes: Or we could add a callback/virtual class mechanism for specific stuff like this
```cpp
class IEvent
{
public:
    virtual void written(uint size) = 0;
};
saveToString(IEvent& event);
```

flomnes (commented):
What about `Yuni::AnyString` ?

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
 }
 
-bool IniFile::open(const AnyString& filename, bool warnings)
+bool startingSection(std::string line)
 {
-    // clear
-    clear();
-    pFilename = filename;
+    boost::trim(line);
+    if (boost::starts_with(line, "[") && boost::ends_with(line, "]"))
```
flomnes: As of C++20, `starts_with` and `ends_with` are [available in the standard library](https://en.cppreference.com/w/cpp/string/basic_string/starts_with) as member functions of `std::string`. Consider using these instead.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
+}
+
+std::string getSectionName(std::string line)
+{
+    std::vector<std::string> splitLine;
+    boost::split(splitLine, line, boost::is_any_of("[]"));
+    return splitLine[1];
+}
+
+bool isProperty(std::string line)
```
flomnes:
```suggestion
static bool isProperty(const std::string& line)
```

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
+{
+    if (std::count(line.begin(), line.end(), '=') == 1)
+        return true;
+    return false;
```
flomnes:
```suggestion
    return std::count(line.begin(), line.end(), '=') == 1;
```

flomnes: Also, consider using [`std::ranges::count`](https://en.cppreference.com/w/cpp/algorithm/ranges/count)

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
+}
+
+std::pair<std::string, std::string> getKeyValuePair(std::string line)
+{
+    boost::trim(line);
+
+    std::vector<std::string> splitLine;
+    boost::split(splitLine, line, boost::is_any_of("="));
+
+    return std::make_pair(splitLine[0], splitLine[1]);
```
flomnes:
```suggestion
    return {splitLine[0], splitLine[1]};
```

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
+}
+
+bool isComment(std::string line)
+{
+    boost::trim_left(line);
+    if (boost::starts_with(line, "#") || boost::starts_with(line, ";"))
+        return true;
+    return false;
```
flomnes: 1. Use `line.starts_with`
2. One line instead of 3 `return x.starts_with(y) || x.starts_with(z);`

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
+    filename_ = filename; // storing filename for further use
+    std::string filePath = filename.to<std::string>();
+
+    if (std::ifstream file(filePath); file.is_open())
+    {
+        if (!readStream(file) && warnings)
+        {
+            logs.error() << "Invalid INI file : " << filePath;
+            return false;
+        }
```
flomnes: This is very error-prone. If `warning=false`, the function will return `true` even though `readStream` failed.

IMHO `warnings` should decide only on logs, the execution flow shouldn't care about `warnings`. Use nested `if` here.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -235,44 +153,102 @@ uint IniFile::Section::size() const
...
 
-    // Error
     if (warnings)
         logs.error() << "I/O error: " << filename << ": Impossible to read the file";
-    pFilename.clear();
     return false;
 }
 
-void IniFile::saveToString(std::string& str) const
+std::string IniFile::saveToString() const
```
flomnes:
```suggestion
std::string IniFile::toString() const
```

src/libs/antares/inifile/inifile.cpp
flomnes: I disagree. This is not in the scope of this PR.

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done : declaration of **lineMaxSize** was removed

src/libs/antares/inifile/inifile.cpp
```diff
@@ -253,51 +168,119 @@ uint IniFile::Section::size() const
...
-            lastSection = AnalyzeIniLine(pFilename, this, lastSection, line, read);
+            currentSection = addSection(getSectionName(line));
+            continue;
         }
 
-        delete[] line;
+        if (isProperty(line))
+        {
+            std::pair<std::string, std::string> pair = getKeyValuePair(line);
+            currentSection->add(pair.first, pair.second);
```
guilpier-code: `currentSection->add(pair.first, pair.second);` : what if we're reading a (key, value) pair and we're not in a section (meaning : **currentSection** is still **nullptr**)  ?
We should raise an error in this case.

src/libs/antares/inifile/inifile.cpp
guilpier-code: **Reminder** : as @sylvlecl said, **readStream(...)** should be made **public** in order to test it, and because it could be useful to have it **public** some day.

src/libs/antares/inifile/inifile.cpp
guilpier-code: ok, we could do that. 
But it's still more work & tests than a simple removal of : 
```C++
if (written != 0)
        Statistics::HasWrittenToDisk(written);
```
Furthermore, in the **GUI**, printed information about the amount of loaded data may be very approximative.
Besides, these information may also be useless, as the user can have a much more accurate estimation of her/his study's size on disk by asking directly to the system to do it.
By the way, `Statistics::HasWrittenToDisk(written);` increases a ... global variable.

src/libs/antares/inifile/inifile.cpp
guilpier-code: I know, but it's so simple to remove it (please see [the comment just above](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2064#discussion_r1598332129)).

src/libs/antares/inifile/inifile.cpp
guilpier-code: Indeed, a bit more readable like this

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

guilpier-code:
> What about `Yuni::AnyString` ?

Please see the PR's presentation comment (now more detailed)

src/libs/antares/inifile/inifile.cpp
guilpier-code: I tested your change proposal, but it does not compile, at least on **Windows**. 
It would have been nice though.

src/libs/antares/inifile/inifile.cpp
guilpier-code: ok, you're totally right.
I introduced an extra check to the original code.
In the original code, **warnings** is only used to log when an issue occurs while opening the ini file.
Now it's also used when a format issue occurs.
So I inadvertently changed the behavior.

So, in order to revert this new unsuitable behavior, the statement : 
`if (!readStream(file) && warnings)` 
should be replaced with : 
`if (!readStream(file))`

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

guilpier-code:
> As discussed
> 
> * `IniFile` -> `IniContent`
> * Publish `readStream` and `saveToStream`
> * Make tests using these 2 functions above

I'm not sure that renaming class **IniFile** into **IniContent** makes sense for now, because it is still coupled to the **file** concept : 
For example :
- it still contains a data member **filename_**
- it still has functions like **filename()** or **loaded()**

We should properly decouple data reading/saving from content managing.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -253,51 +160,110 @@ uint IniFile::Section::size() const
...
-    IO::File::Stream file;
-    if (file.open(filename))
+static std::string getSectionName(std::string line)
+{
+    std::vector<std::string> splitLine;
+    boost::split(splitLine, line, boost::is_any_of("[]"));
+    return splitLine[1];
+}
+
+static bool isProperty(std::string line)
```
sylvlecl: either use a const ref or a string_view

src/libs/antares/inifile/inifile.cpp
```diff
@@ -253,51 +160,110 @@ uint IniFile::Section::size() const
...
-    clear();
-    pFilename = filename;
+    boost::trim(line);
+    return line.starts_with("[") && line.ends_with("]");
+}
 
-    // Load the file
-    IO::File::Stream file;
-    if (file.open(filename))
+static std::string getSectionName(std::string line)
```
sylvlecl: either use a const ref or a string_view when possible to avoid copies

src/libs/antares/inifile/inifile.cpp
```diff
@@ -253,51 +160,110 @@ uint IniFile::Section::size() const
...
+    boost::split(splitLine, line, boost::is_any_of("[]"));
+    return splitLine[1];
+}
+
+static bool isProperty(std::string line)
+{
+    return std::ranges::count(line.begin(), line.end(), '=') == 1;
+}
+
+static std::pair<std::string, std::string> getKeyValuePair(std::string line)
```
sylvlecl: Note that you could create a small struct for naming key and value (only 4 lines of code), instead of using a std::pair<std::string, std::string> which is less expressive.

sylvlecl: Or juste return a `Property` (existing class).

src/libs/antares/inifile/inifile.cpp
```diff
@@ -253,51 +160,110 @@ uint IniFile::Section::size() const
...
+bool IniFile::readStream(std::istream& in_stream)
+{
+    std::string line;
+    IniFile::Section* currentSection(nullptr);
+    uint64_t read = 0;
+    while (std::getline(in_stream, line))
     {
-        enum
+        read += line.size();
+        if (startingSection(line))
```
sylvlecl: isStartingSection

src/libs/antares/inifile/inifile.cpp
```diff
@@ -253,51 +160,110 @@ uint IniFile::Section::size() const
...
-        auto* line = new char[lineHardLimit];
+            currentSection = addSection(getSectionName(line));
+            continue;
+        }
 
-        IniFile::Section* lastSection = nullptr;
-        uint64_t read = 0;
+        if (isProperty(line))
+        {
+            // Note : if a property not in a section, it's simply skipped
```
sylvlecl, line 205: Why not get an error instead?
Was it the previous behaviour ?

sylvlecl (dismissed):
The implementation is indeed MUCH simpler, that's nice !

However we need to add unit tests to cover this feature and ensure there is no regression in edge cases, like comments, properties which are not in a section, and so on.

Answer from @guilpier-code : it's now done

src/libs/antares/inifile/include/antares/inifile/inifile.h
```diff
@@ -68,7 +66,7 @@ class IniFile final
     class Section final
     {
     public:
-        Section();
+        Section() = default;
         explicit Section(const AnyString& name);
         ~Section();
```
flomnes, line 71: `= default` ?

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: I like your second idea, done.

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: I first raised an error in the code when a property was not inside a section.
But I realized that the previous behavior was to skip such properties.
That's why it's emphasized in the code as a comment.

src/libs/antares/inifile/include/antares/inifile/inifile.h
guilpier-code: yes, why not ?

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

guilpier-code:
I can see 3 possible directions for improving the class **IniFile** : 
- **Yuni** : remove some more **Yuni** things (mainly strings)
- **Implementation details** : several things can be done to improve, let's say, the implementation details. I mean without changing current classes interfaces. These improvements could make the code shrink quite a lot. For example : 
  - classes **IniFile** or **Section** internally implement a (chained) **list**. For both of them, a new **list** mechanism is duplicated. It's crazy :-). So let's replace it with a container from the **STL**. It could be a **std::list**, but a **std::vector** would work the same I guess. 
  - We have **each(...)** that are barley used : instead, **IniFile** and **Section** are traversed using classic **for loops**, even internally.
- **Design** :  as said above [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2064#issuecomment-2108117280), **IniFile** has drawbacks that would require to split it.

src/libs/antares/inifile/inifile.cpp
```diff
@@ -253,95 +147,143 @@ uint IniFile::Section::size() const
...
+            if (currentSection)
+            {
+                IniFile::Property p = getProperty(line);
+                currentSection->add(p.key, p.value);
```
flomnes, line 209:
```suggestion
                currentSection->add(Property(line));
```
Add a constructor for `Property` : `Property(std::string line)`

src/tests/src/libs/antares/inifile/test_inifile_io.cpp
flomnes, line 1: You should add a test or 2 on **keys** when reading INI files. Indeed, you look for sections & properties, but never check the actual content of properties.

guilpier-code: Done

src/libs/antares/inifile/inifile.cpp
guilpier-code: If we do that, we couple **reading concern** (a line) and **business logic** (Property).
As it's done now, **getProperty** is in the **reading** area, and depend on type **Property** (business).
If we introduce **Property(std::string line)** **reading** puts a step into business logic (dependency in the other direction).

guilpier-code: If we want to merge those 2 lines in one, we could have a new **Section::add(...)** function, taking a Property as argument, so we could have : 
`currentSection->add(getProperty(line));`

src/libs/antares/inifile/include/antares/inifile/inifile.h
flomnes: It was not `default` previously

flomnes (approved):

src/tests/src/libs/antares/inifile/test_inifile_io.cpp
```diff
@@ -0,0 +1,233 @@
+
```
sylvlecl: Missing header

src/tests/src/libs/antares/inifile/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+
+set(src_libs_antares "${CMAKE_SOURCE_DIR}/libs/antares")
```
sylvlecl: Not needed

sylvlecl (changes requested):
Last details

sylvlecl (approved):

sylvlecl (approved):
Just applied copyright + formatting

src/libs/antares/inifile/inifile.cpp
guilpier-code: Done

PR #2066, opened 2024-04-29
Replace Yuni::IO with std functions

payetvin:
The goal of this PR is to remove Yuni::IO functions in important parts of the code (libs and solver)
`std::filesystem::path` is used instead of Yuni strings when needed, more work will need to be done to use `path` on every level

No changes were made in:
- GUI
- deprecated tools
- functions that save the study


src/tools/batchrun/main.cpp
```diff
@@ -218,10 +218,10 @@ int main(int argc, char* argv[])
         }
         else
         {
-            String tmp;
-            IO::MakeAbsolute(tmp, *optSolver);
-            IO::Normalize(solver, tmp);
-            if (not IO::File::Exists(solver))
+            std::string tmp = *optSolver;
```
guilpier-code, line 238: Do we need **tmp** here ?

src/tools/batchrun/main.cpp
```diff
@@ -218,10 +218,10 @@ int main(int argc, char* argv[])
...
+            std::filesystem::path solverPath = std::filesystem::absolute(tmp);
```
guilpier-code: Could we not have directly : 
```C++
std::filesystem::path solverPath = 
        std::filesystem::absolute(optSolver.to<std::string>()).lexically_normal();
if (! std::filesystem::exists(solverPath))
{
    ...
}
```
?

src/libs/antares/study/area/links.cpp
```diff
@@ -428,17 +428,16 @@ void logLinkDataCheckErrorDirectIndirect(const AreaLink& link,
 }
 } // anonymous namespace
 
-bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyString& folder)
```
guilpier-code, line 476: Having : 
```C++
namespace fs = std::filesystem;
```
here (as it's done somewhere else in this PR) would good too, I think.

src/libs/antares/study/area/list.cpp
```diff
@@ -36,6 +36,8 @@
 
 using namespace Yuni;
 
+namespace fs = std::filesystem;
+
 namespace Antares::Data
 {
 namespace // anonymous
```
guilpier-code, line 45: About function : 
```C++
template<class StringT>
static bool AreaListLoadFromFolderSingleArea(Study& study,
                                             AreaList* list,
                                             Area& area,
                                             StringT& buffer,
                                             const StudyLoadOptions& options)
```

Parameter **buffer** is actually a local variable : we don't need to pass it to the function.
So, we should : 
- remove the **buffer** parameter and the associated template.
- make buffer a local variable.

Besides, I wonder if we could make buffer a **std::filesystem::path**.
Indeed : 
- it can easily turned into a **std::string**, vie **string()** member function.
- Locally (to this function), buffer very is often passed as a **AnyString**, which has a constructor taking a **std::string** as argument.

So every time we pass **buffer**, we could pass **buffer.string()** instead.
Is this possible ?

src/libs/antares/study/importlogs.cpp
```diff
@@ -36,9 +36,9 @@ void Study::importLogsToOutputFolder(Solver::IResultWriter& resultWriter) const
     if (!logs.logfile())
         return;
 
-    std::string logPath("simulation.log");
-    String from;
-    IO::Normalize(from, logs.logfile());
+    std::filesystem::path logPath = "simulation.log";
```
guilpier-code:
```C++
std::filesystem::path logPath = "simulation.log";
```
is not a log path, it's rather a filename. 
**logPath**  == > **logFileName** ?

guilpier-code: And I'am not even sure that we need **logPath** : 
```C++
resultWriter.addEntryFromFile("simulation.log", from.string());
```
should work

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -21,7 +21,6 @@
 
 #include <antares/logs/logs.h>
 #include <yuni/io/file.h>
```
guilpier-code: **Question** : 

if we replace : 
```C++
Yuni::IO::Directory::Create(folder);
```
with : 
```C++
std::filesystem::create_directories(folder);
```
, could we then remove the `#include <yuni/io/file.h>` ?

src/libs/antares/study/parts/thermal/ecoInput.cpp
```diff
@@ -58,27 +58,27 @@ bool EconomicInputData::saveToFolder(const AnyString& folder) const
...
 
     if (study.header.version >= StudyVersion(8, 7))
     {
-        YString filename;
+        fs::path filename;
         Yuni::Clob dataBuffer;
 
-        filename << folder << SEP << "fuelCost.txt";
-        if (IO::File::Exists(filename))
+        filename = fs::path(folder) / "fuelCost.txt";
```
guilpier-code:
```C++
fs::path filename = = fs::path(folder) / "fuelCost.txt";
```
instead of 2 lines

guilpier-code (dismissed):

src/libs/antares/writer/immediate_file_writer.cpp
```diff
@@ -83,62 +39,67 @@ ImmediateFileResultWriter::ImmediateFileResultWriter(const char* folderOutput):
...
-    Yuni::String fullPath;
+    fs::path fullPath;
     if (!prepareDirectoryHierarchy(pOutputFolder, entryPath, fullPath))
     {
         return;
     }
 
-    switch (Yuni::IO::File::Copy(filePath.c_str(), fullPath))
+    std::error_code ec;
+    fs::copy(filePath, fullPath, ec);
```
sylvlecl: I think it would be simpler to use the message provided by the STL:
```cpp
try {
    fs::copy(filePath, fullPath);
} catch (const fs::filesystem_error& exc) {
    logs.error() << exc.what();
}
```

For example it provides such message:
```
filesystem error: cannot copy: No such file or directory [source] [target]
```
which is richer than just the error code message "No such file or directory".

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -54,7 +56,7 @@ bool STStorageInput::createSTStorageClustersFromIniFile(const std::string& path)
         return true;
     }
 
-    logs.debug() << "  :: loading `" << pathIni << "`";
+    logs.debug() << "  :: loading `" << pathIni.string() << "`";
```
flomnes: Add overloads to Yuni's logs on type `fs::path` ?

src/libs/antares/writer/in_memory_writer.cpp
```diff
@@ -77,27 +84,30 @@ void InMemoryWriter::addEntryFromBuffer(const std::string& entryPath, std::strin
     addToMap(pEntries, entryPath, entryContent, pMapMutex, pDurationCollector);
 }
 
-void InMemoryWriter::addEntryFromFile(const std::string& entryPath, const std::string& filePath)
+static std::string readFile(const fs::path& filePath)
```
sylvlecl: For this kind of very generic function (reading a file to a string), we could share it in a technical lib, for example we already have an "IO" library where it could go.

This function would throw an exception on failure, which we could just let through here.

src/libs/antares/writer/zip_writer.cpp
```diff
@@ -157,29 +159,29 @@ void ZipWriter::addEntryFromBuffer(const std::string& entryPath, std::string& en
     addEntryFromBufferHelper<std::string>(entryPath, entryContent);
 }
 
-void ZipWriter::addEntryFromFile(const std::string& entryPath, const std::string& filePath)
+static std::string readFile(const fs::path& filePath)
```
sylvlecl, line 160: This could be removed from here by sharing the function in IO lib

sylvlecl (commented):
A few technical suggestion: in particular share implementation of reading file to string in IO lib

src/libs/antares/study/importlogs.cpp
```diff
@@ -51,12 +50,12 @@ void Study::importLogsToOutputFolder(Solver::IResultWriter& resultWriter) const
         logs.closeLogfile();
     }
 
-    resultWriter.addEntryFromFile(logPath, from.c_str());
+    resultWriter.addEntryFromFile("simulation.log", from.string());
```
flomnes, line 53: Not sure about this. "simulation.log" refers to $PWD/simulation.log, which is probably not what we want

src/libs/antares/study/header.cpp
```diff
@@ -205,27 +205,25 @@ bool StudyHeader::saveToFile(const AnyString& filename, bool upgradeVersion)
...
         return StudyVersion::unknown();
     }
 
-    // foldernormalization
-    String abspath, directory;
-    IO::MakeAbsolute(abspath, folder);
-    IO::Normalize(directory, abspath);
+    // folder normalization
+    fs::path abspath = fs::absolute(folder);
+    abspath = abspath.lexically_normal();
```
flomnes, line 217: Add unit tests on this ? To make sure that old & new behavior are consistent

src/libs/antares/study/importlogs.cpp
payetvin: I don't think this change the previous behavior, we just avoid using a single use var

src/libs/antares/io/include/antares/io/file.h
```diff
@@ -21,14 +21,17 @@
...
+#include <filesystem>
+
 /*!
 ** \brief Reset the content of a file
 **
 ** This routine will wait if there is not enough disk space
 */
 bool IOFileSetContent(const AnyString& filename, const AnyString& content);
 
+std::string readFile(const std::filesystem::path&);
```
sylvlecl, line 37: Let's wrap it into a namespace like `Antares::io`

src/ext/yuni/src/yuni/core/string/traits/append.h
```diff
@@ -294,6 +295,18 @@ class Append<CStringT, std::list<T>> final
     }
 };
 
+namespace fs = std::filesystem;
```
JasonMarechal25: You're declaring fs but don't use it

src/libs/antares/io/file.cpp
```diff
@@ -24,29 +24,53 @@
...
-    retryTimeout = 35, // seconds
-};
+    std::ifstream file(filePath, std::ios_base::binary | std::ios_base::in);
+    if (!file.is_open())
+    {
+        logErrorAndThrow(filePath.string() + ": file does not exist");
+    }
+
+    using Iterator = std::istreambuf_iterator<char>;
+    std::string content(Iterator{file}, Iterator{});
```
JasonMarechal25, line 63: https://stackoverflow.com/a/2602060
If readfile is used to open large file it may good to improve the implementation.
An issue may be opened to talk about the subject

src/libs/antares/study/area/links.cpp
```diff
@@ -473,17 +475,16 @@ bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const
...
     /* Initialize */
-    String buffer;
-    buffer << folder << SEP << "properties.ini";
+    fs::path path = folder / "properties.ini";
 
     IniFile ini;
-    if (!ini.open(buffer))
+    if (!ini.open(path.string()))
     {
         return 0;
```
JasonMarechal25:
```suggestion
        return false;
```

src/libs/antares/study/area/links.cpp
```diff
@@ -496,7 +497,7 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const AnyStr
     for (auto* s = ini.firstSection; s; s = s->next)
     {
         // Getting the name of the area
-        buffer.clear();
+        String buffer;
         TransformNameIntoID(s->name, buffer);
```
JasonMarechal25: We could make a new version of TransformNameIntoID that return a std::string and avoid the second parameter.

It would make some code cleaner.

src/libs/antares/study/header.cpp
JasonMarechal25, line 1: Do  we have unit tests for those function ?

If not would it be good to add some and ensure that we have the same behaviour before/after ?

src/libs/antares/study/header.cpp
```diff
@@ -205,27 +205,25 @@ bool StudyHeader::saveToFile(const AnyString& filename, bool upgradeVersion)
...
+StudyVersion StudyHeader::tryToFindTheVersion(const std::string& folder)
 {
     if (folder.empty()) // trivial check
     {
         return StudyVersion::unknown();
     }
 
-    // foldernormalization
-    String abspath, directory;
-    IO::MakeAbsolute(abspath, folder);
```
JasonMarechal25, line 217: MakeAbsolute behaviour is weird due to is IsAbsolute
Windows style path are supported under linux contrary to fs::absolute.
On unix:
IO::IsAbsolute(C:/plop) => true
filesystem::path::is_absolute(C:/plop) => false

It sould not be an issue (or even supported), but well, it bares mentionning

src/libs/antares/study/header.cpp
```diff
@@ -205,27 +205,25 @@ bool StudyHeader::saveToFile(const AnyString& filename, bool upgradeVersion)
...
-    IO::Normalize(directory, abspath);
+    // folder normalization
+    fs::path abspath = fs::absolute(folder);
+    abspath = abspath.lexically_normal();
 
-    if (not directory.empty() and IO::Directory::Exists(directory))
+    if (!abspath.empty() && fs::exists(abspath))
```
JasonMarechal25: Is empty  test mendatory ? fs::exists("") should return false

src/libs/antares/study/include/antares/study/area/area.h
```diff
@@ -34,6 +35,7 @@
 #include "antares/study/filter.h"
 #include "antares/study/parts/parts.h"
 
+#include "../parts/parts.h"
```
JasonMarechal25: Please use full path

src/libs/antares/io/file.cpp
```diff
@@ -24,29 +24,53 @@
...
+    if (!file)
+    {
+        logErrorAndThrow("Read failed '" + filePath.string() + "'");
+    }
+    return content;
+}
 
-bool IOFileSetContent(const AnyString& filename, const AnyString& content)
+bool fileSetContent(const std::string& filename, const std::string& content)
```
guilpier-code, line 71: Fonction **fileSetContent** : 
it's the one and only function of this source file still containing references to **Yuni**.
It's uselessly complicated : 
- several attempts to open a file in writing mode 
- trying to identify the cause of the failure

Nobody does that.
It can be handled by STL well enough.
We should take example on function **readFile**, shorter and more clear (in the history of this source file, **readFile** came out much more recently than **fileSetContent**).
So we should rewrite this function with **std** things, like **readFile**.
It would be quick and easy to test, as **fileSetContent** is called twice in the whole code, and only from **immediate_file_writer.cpp**.
After that, no more reference to **Yuni** in this source file.

src/libs/antares/study/area/list.cpp
guilpier-code: Should be done one day.

src/ui/simulator/windows/sets.cpp
```diff
@@ -581,7 +581,8 @@ void Sets::saveToStudy()
     pRichEdit->SaveFile(wxStringFromUTF8(pTempFile), wxRICHTEXT_TYPE_TEXT);
     // ... and reloading it
     // IO::File::LoadFromFile(study->simulation.comments, pTempFile);
-    study->setsOfAreas.loadFromFile(pTempFile);
+    std::filesystem::path p = std::filesystem::path(pTempFile.c_str());
```
flomnes:
```suggestion
    std::filesystem::path p(pTempFile.c_str());
```

src/tests/src/libs/antares/test_utils.cpp
```diff
@@ -83,3 +88,29 @@ BOOST_AUTO_TEST_CASE(test_transform_name_into_id)
...
+
+    Yuni::String yuniNorm;
+    Yuni::IO::Normalize(yuniNorm, p1.string());
+
+    BOOST_CHECK(p1.lexically_normal().string() == yuniNorm);
+
+    Yuni::IO::Normalize(yuniNorm, p2.string());
+
+    BOOST_CHECK(p2.lexically_normal().string() == yuniNorm);
+}
```
flomnes: Make two test-cases, don't mix up tests.

src/solver/misc/options.cpp
```diff
@@ -201,10 +198,10 @@ void checkAndCorrectSettingsAndOptions(Settings& settings, Data::StudyLoadOption
     const auto& optPID = settings.PID;
     if (!optPID.empty())
     {
-        IO::File::Stream pidfile;
-        if (pidfile.openRW(optPID))
+        std::ofstream pidfile(optPID);
+        if (pidfile.is_open())
```
```suggestion
        if (std::ofstream pidfile(optPID); pidfile.is_open())
```

src/tests/src/libs/antares/test_utils.cpp
flomnes, line 1: Add tests for `IsAbsolute`

flomnes:
`antares-solver` can't launch a simulation after this PR. Locally, I get
```
[2024-05-16 11:33:51][solver][fatal] The folder `/home/omnesflo/Téléchargements/valid-named-mps/binding-constraints/` does not seem to be a valid study
[2024-05-16 11:33:51][solver][fatal] Aborting now. (warning: no file log available)
[2024-05-16 11:33:51][solver][warns] No log file available
```

src/libs/antares/writer/immediate_file_writer.cpp
```diff
@@ -83,62 +39,67 @@ ImmediateFileResultWriter::ImmediateFileResultWriter(const char* folderOutput):
 
 ImmediateFileResultWriter::~ImmediateFileResultWriter() = default;
 
-static bool prepareDirectoryHierarchy(const YString& root,
-                                      const std::string& entryPath,
-                                      Yuni::String& output)
+static bool prepareDirectoryHierarchy(const fs::path& root,
```
guilpier-code, line 42: I find this class difficult to understand.

- It seems it handles **results** (I understand **simulation results**) but it also handles logs (when it's called from **importlogs.cpp**)
- Method **addEntryFromFile** : every time I cross this method, I wonder what it does. I must read the body of this function to understand, and I find it hard to read. 
  - the function's name is unclear to me, 
  - what the arguments represent is mysterious to me, 
  - I don't understand why we make a file system copy.
- Function **prepareDirectoryHierarchy** has name unclear to me : actually it concatenates path elements and then create the resulting path on disk. We should remove this function and keep 2 functions, one for concatenation, the other for creation on disk. 
Besides, it has an output argument used to store the result of this function. This result should be returned instead.
- This class is supposed to flush something but it does not : the **flush** function is defined but empty. It looks like a problem of conception.

guilpier-code (dismissed):

src/ui/simulator/windows/sets.cpp
JasonMarechal25:
```suggestion
    std::filesystem::path p{pTempFile.c_str()};
```

JasonMarechal25 (approved):

src/libs/antares/io/file.cpp
flomnes: I agree. This was meant to allow the user to free up some space. But it only makes sense on a desktop environment

flomnes (approved):

src/libs/antares/writer/immediate_file_writer.cpp
flomnes: out of scope

PR #2068, opened 2024-04-29
[CR23] Remove prefix "hgp", use "h" for max pumping & generation

flomnes:
### ⚠️ MERGE THIS AFTER #2073 

Using prefix "hgp" made it too complex for the user to create a valid dataset, and too tricky to validate it: need to loop over all MC years, read prefixes "h" and "hgp" to check that minGen[h] <= maxGen[h] for every hour.


src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -112,16 +112,16 @@ static void ConvertDailyTSintoHourlyTS(const Matrix<double>::ColumnType& dailyCo
...
-    ror(timeseriesNumbers),
-    storage(timeseriesNumbers),
-    mingen(timeseriesNumbers),
-    maxHourlyGenPower(timeseriesNumbersHydroMaxPower),
-    maxHourlyPumpPower(timeseriesNumbersHydroMaxPower)
+ ror(timeseriesNumbers),
+ storage(timeseriesNumbers),
+ mingen(timeseriesNumbers),
+ maxHourlyGenPower(timeseriesNumbers),
+ maxHourlyPumpPower(timeseriesNumbers)
```
flomnes: The real change is here. The rest is deletion.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -223,20 +171,18 @@ bool DataSeriesHydro::LoadMaxPower(const AreaName& areaID, const AnyString& fold
...
 
     return ret;
 }
 
-void DataSeriesHydro::buildHourlyMaxPowerFromDailyTS(const Matrix<double>::ColumnType& DailyMaxGenPower,
-                                               const Matrix<double>::ColumnType& DailyMaxPumpPower)
+void DataSeriesHydro::buildHourlyMaxPowerFromDailyTS(
+  const Matrix<double>::ColumnType& DailyMaxGenPower,
+  const Matrix<double>::ColumnType& DailyMaxPumpPower)
 {
```
guilpier-code, line 178: About : 
```C++
 maxHourlyGenPower.reset(TScount_, HOURS_PER_YEAR);
 maxHourlyPumpPower.reset(TScount_, HOURS_PER_YEAR);
```
**TScount_** is used here to resize the max power TS, knowing that we are building power TS from daily TS sized to **1**.
So, to make very clear, we should have : 
```C++
 maxHourlyGenPower.reset(1, HOURS_PER_YEAR);
 maxHourlyPumpPower.reset(1, HOURS_PER_YEAR);
```
This is more important than it looks : by the time of this comment, there are ambiguities and questions on **TScount_** (is it a number for ready made TS only ? Or does it take into account generated TS ? Some other comments on this PR also involve **TScount_**).
So the less used, the better.

src/libs/antares/study/area/list.cpp
```diff
@@ -894,15 +894,10 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
-
-            if (study.usedByTheSolver)
-            {
-                hydroSeries->EqualizeMaxPowerTSsizes(area);
-            }
-            else
-                hydroSeries->setHydroModulability(area);
         }
 
+        hydroSeries->computeTSCount();
```
guilpier-code: `hydroSeries->computeTSCount();` : 
The underlying private data member we compute here is **TScount_**
We compute the underlying private data member **TScount_** here, knowing that it will be possibly recomputed in case some **ror** and **storage** are generated (and so resized).
So we can wonder if it worth it leaving it here like that, alone, or computed once for all, after : 
`TSGenerator::ResizeGeneratedTimeSeries` is called.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -145,37 +98,33 @@ void DataSeriesHydro::copyMaxPowerTS(const DataSeriesHydro& source)
...
-    generationTScount_ = nbSeries;
-}
-
-void DataSeriesHydro::resizeMaxPowerTS(uint nbSeries)
-{
-    maxHourlyGenPower.reset(nbSeries, HOURS_PER_YEAR);
-    maxHourlyPumpPower.reset(nbSeries, HOURS_PER_YEAR);
-
-    maxPowerTScount_ = nbSeries;
+    TScount_ = nbSeries;
```
guilpier-code: `TScount_ = nbSeries;` : 
Why do we compute **TScount_** in a one way in **computeTSCount()** and in another way here ?
In **computeTSCount()**, we learn that **TScount_** is about 5 hydro TS. Now it's about 2 of them. I'm confused.

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -145,37 +98,33 @@ void DataSeriesHydro::copyMaxPowerTS(const DataSeriesHydro& source)
...
 {
-    generationTScount_ = std::max(storage.numberOfColumns(), ror.numberOfColumns());
+   const std::vector<uint32_t> nbColumns({
+        storage.numberOfColumns(),
+        ror.numberOfColumns(),
+        mingen.numberOfColumns(),
+        maxHourlyGenPower.numberOfColumns(),
+        maxHourlyPumpPower.numberOfColumns()
+    });
+    TScount_ = *std::max_element(nbColumns.begin(), nbColumns.end());
```
guilpier-code: **TScount_** : its a max. Shouldn't we rename into **maxTScount_** ?

src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -133,7 +86,7 @@ void DataSeriesHydro::copyGenerationTS(const DataSeriesHydro& source)
     storage.timeSeries = source.storage.timeSeries;
     mingen.timeSeries = source.mingen.timeSeries;
 
-    generationTScount_ = source.generationTScount_;
+    TScount_ = source.TScount_;
```
guilpier-code: `TScount_ = source.TScount_;` : 
**TScount_** is useful only in the **solver**, not in the **GUI**.
So we don't have to update it here I guess.

src/libs/antares/study/area/list.cpp
```diff
@@ -894,15 +894,10 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
...
+
```
guilpier-code: _Not exactly in the scope of this PR_

`hydroSeries->resizeTSinDeratedMode(....);` : 
If we're in **derated** mode, what prevents **ror** and **storage** to be resized in case they are generated ? 
If nothing does, should we not expect unwanted consequences ?

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
```diff
@@ -128,9 +128,10 @@ bool HydroMaxTimeSeriesReader::read(const AnyString& folder, bool usedBySolver)
 {
     bool ret = loadDailyMaxPowersAndEnergies(folder, usedBySolver);
     copyDailyMaxEnergy();
+    hydro_.series->computeTSCount();
```
guilpier-code: `hydro_.series->computeTSCount();` : 
Please see **.../area/list.cpp**'s review [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2068#discussion_r1592519627) for a comment on that matter.

guilpier-code (changes requested):

src/libs/antares/study/parts/hydro/series.cpp
flomnes: OK, yes.

src/libs/antares/study/area/list.cpp
flomnes: I think it would be best not to store this value, but instead to compute it every time. Fewer errors.

src/libs/antares/study/parts/hydro/series.cpp
flomnes: @guilpier-code I removed `TScount_`, will now use `TScount()` instead.

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
flomnes: This method has been removed. We now compute the TS count every time it's necessary.

src/libs/antares/study/area/list.cpp
flomnes: Yes, derated mode is supposed to average all time-series and run a single MCY. In my opinion, we should still resize & average all TS, included those that are generated.

src/libs/antares/study/parts/hydro/series.cpp
flomnes: Yes, I removed it.

src/libs/antares/study/parts/hydro/series.cpp
flomnes: `TScount_` has been removed. This removal simplifies the code quite a bit : we don't need to recompute `TScount_` every now and then.

src/libs/antares/study/parts/hydro/series.cpp
flomnes: `computeTSCount` has been removed.

src/libs/antares/study/parts/hydro/series.cpp
flomnes: Actually, I used 1 as you suggested. Using anything else is a **logic error** and introduces regressions.

src/libs/antares/study/area/list.cpp
guilpier-code: Yes but we duplicate code.
Ideally simulation should not know if TS are generated or read.
This TS acquisition should be done while loading input data.
This would avoid have further complications like that.
I know that TS generation will disappear, even if we don't know when.
Anyway, this issue will be handled later, in a way or another, if it is gods will.

guilpier-code: Yes but it's slow and complicated. It could be more simple.
Ideally simulation should not know if TS are generated or read.
This TS acquisition should be done while loading input data.
It would allow :
- to would avoid have further complications like that.
- maybe to make some more simplifications in **timeseries-numbers.cpp**

I know that TS generation will disappear, but we don't know when.
Anyway, this issue will be handled later, in a way or another. 
We should create a ticket to study the benefit we would get by taking this path.

guilpier-code (approved):

PR #2073, opened 2024-04-30
Rationalize consistency checks on the number of columns

### LINKED ISSUE #2069: mingen is erased when generating TS for hydro
Branch develop, commit b84250397.

See **src/libs/antares/study/parts/hydro/series.cpp:160**
```cpp
void DataSeriesHydro::resizeGenerationTS(uint nbSeries)
{
    storage.reset(nbSeries, DAYS_PER_YEAR);
    ror.reset(nbSeries, HOURS_PER_YEAR);
    mingen.reset(nbSeries, HOURS_PER_YEAR);

    generationTScount_ = nbSeries;
}
```

Also **src/solver/ts-generator/generator.cpp:49**
```cpp
// Hydro
if (params.timeSeriesToGenerate & Data::timeSeriesHydro)
{
    Data::DataSeriesHydro* const series = area.hydro.series;
    const uint nbSeries = params.nbTimeSeriesHydro;
    series->resizeGenerationTS(nbSeries);
}
```

I was able to reproduce on study **valid-named-mps/binding-constraints**, by setting mingen to 1 for area AT. The upper bound is as expected (**problem-1-1--optim-nb-1.mps**)
```
 UP BNDVALUE  HydProd::area<at>::hour<0>  4176.178291061
```
but the lower bound is absent, meaning it is =0 (expected 1).

flomnes:
# Context
This is a preliminary work to place `maxHydroPumping` & `maxHydroGen` under sc-builder prefix "h", and remove prefix "hgp" introduced in CR23 (see #2068).

This should also fix #2069.

# Description
- Add homogeneity in the checks performed on the number of columns for
   - Links (direct & indirect) (prefix "ntc)
   - Hydro (ror, storage & mingen) (prefix "h")
   - Thermal (prefix "t", not useful for now, but we may use this mechanism for fuelCost & co2Cost, to replace the custom `checkFuelCostColumnNumber` and `checkCO2CostColumnNumber` functions)
- Declare TSNumbers before series. Otherwise we get a segfault since data members are initialized in the order where they appear in the class declaration.
- Remove `EqualizeGenerationTSsizes` mechanism for `{ror, storage}` and `{ror, mingen}`. If only one column is present, use it, which is more or less equivalent to the column duplication made by `EqualizeGenerationTSsizes`.

# TODO
- [x] Add unit tests, especially to check that errors are emitted when they should.

# Related work
[Improve logs](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2080)


src/libs/antares/study/parts/hydro/series.cpp
```diff
@@ -157,11 +157,15 @@ void DataSeriesHydro::reset()
...
+void DataSeriesHydro::computeTSCount()
+{
+    generationTScount_ = std::max(storage.numberOfColumns(), ror.numberOfColumns());
+}
+
 void DataSeriesHydro::resizeGenerationTS(uint nbSeries)
 {
     storage.reset(nbSeries, DAYS_PER_YEAR);
     ror.reset(nbSeries, HOURS_PER_YEAR);
-    mingen.reset(nbSeries, HOURS_PER_YEAR);
```
flomnes, line 174: Fix #2069

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -682,6 +682,39 @@ bool TimeSeriesNumbers::checkAllElementsIdenticalOrOne(std::vector<uint> w)
     return std::adjacent_find(w.begin(), first_one, std::not_equal_to<uint>()) == first_one;
 }
 
+bool TimeSeriesNumbers::CheckNumberOfColumns(const Study& study)
```
guilpier-code: `bool TimeSeriesNumbers::CheckNumberOfColumns(const Study& study)` : 
argument **study** used only for `study.areas`. We should now the arg type to area list.

src/libs/antares/series/include/antares/series/series.h
```diff
@@ -35,10 +35,17 @@ namespace Antares::Data
 class TimeSeries
 {
 public:
-    using numbers = Matrix<uint32_t>;
+    struct Numbers : public Matrix<uint32_t> // TODO[FOM] remove inheritance, it was added mostly to avoid breaking existing code
```
sylvlecl: Indeed, would be better to use composition here with a class encapsulating both TS numbers and registered timeseries.
Do you prefer to postpone it to later work?

src/libs/antares/series/series.cpp
```diff
@@ -31,9 +33,31 @@ using namespace Yuni;
...
+        width.push_back(s->numberOfColumns());
 
-TimeSeries::TimeSeries(numbers& tsNumbers) : timeseriesNumbers(tsNumbers)
-{}
+    return checkAllElementsIdenticalOrOne(width);
+}
+
+TimeSeries::TimeSeries(Numbers& tsNumbers) : timeseriesNumbers(tsNumbers)
+{
+    tsNumbers.registerSeries(this);
```
sylvlecl: We should probably forbid copying of `TimeSeries`, to make sure that this is not called multiple times for the same timeseries.

src/libs/antares/series/series.cpp
```diff
@@ -31,9 +33,31 @@ using namespace Yuni;
...
 {
+void TimeSeries::Numbers::registerSeries(const TimeSeries* s)
+{
+    series.push_back(s);
+}
+
+// TODO[FOM] Code duplication
+static bool checkAllElementsIdenticalOrOne(std::vector<uint> w)
+{
+    auto first_one = std::remove(w.begin(), w.end(), 1); // Reject all 1 to the end
```
sylvlecl, line 46: Not important but note that it's not the most efficient way of implementing this functions logic: without copying and moving anything, we could just register the first value different from 1, continue, and stop as soon as we encounter another one.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -682,6 +682,39 @@ bool TimeSeriesNumbers::checkAllElementsIdenticalOrOne(std::vector<uint> w)
...
+{
+    std::map<const Antares::Data::TimeSeries::Numbers*, std::string> toCheck;
```
sylvlecl: Indexing on a pointer should be avoided: the key order will vary from one execution to another, leading to different orders in execution and messages.
It's better to have a vector of pairs instead (or small structs for naming stuff), because we don't care about finding based on the pointer.

src/solver/simulation/timeseries-numbers.cpp
sylvlecl: Having this method which validates all timeseries with the same logic is a lot better than before I think.

A thought for later though: when adding new timeseries numbers, it will still be quite easy to miss this check. Having a per-object `validate` method that would be called at the start of a computation could make it more "visible", so that we don't miss this validation step. This is linked to the broader validation topic we have already started to discuss.

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -682,6 +682,39 @@ bool TimeSeriesNumbers::checkAllElementsIdenticalOrOne(std::vector<uint> w)
...
+    for (const auto& [_, area] : study.areas)
```
guilpier-code: I can understand why you check TS number of columns for **links** and **hydro**.
But why **thermal** clusters ?

sylvlecl (changes requested):
Minor technical requests

src/solver/simulation/timeseries-numbers.cpp
```diff
@@ -682,6 +682,39 @@ bool TimeSeriesNumbers::checkAllElementsIdenticalOrOne(std::vector<uint> w)
...
+    {
```
guilpier-code, line 756: This function has 2 parts because it seems you try to avoid duplication about logging.
But it looks too complicated and difficult to read to me.
We could only have one block for **links**  and one for **hydro**, and avoid the current last block (I skip **thermal** because of previous comment).

src/libs/antares/series/include/antares/series/series.h
flomnes: Yes I'd rather do it later

src/libs/antares/series/series.cpp
flomnes: Yes, we had issues in the past with this function but you're right. We have a lot of UTs already so we could change implementation

src/solver/simulation/timeseries-numbers.cpp
flomnes: I agree, this is **bad**

src/libs/antares/series/include/antares/series/series.h
guilpier-code: **TimeSeries** should know nothing about TS numbers, I think class **TimeSeries** should be left alone.

On the other hand, class **Numbers** (to be renamed into **TSnumbers**) should be the one who knows about **TimeSeries**, because it checks some **TimeSeries** nmuber of columns consistency.

Instead, current classes **TimeSeries** and **TimeSeries** have a mutual dependency.

So, to me : 
- This class **Numbers** should be renamed into **TSnumbers**, and moved out from class **TimeSeries** (probably in a separate source file).
- We should reverse the registration process : make **Numbers** register **TimeSeries**, and not (as it is done now) make **TimeSeries** allow **Numbers** to register them (which is more complicated). 

For example, `DataSeriesHydro::DataSeriesHydro()` should make `timeseriesNumbers` register `ror`, `storage `and `mingen` : 

```C++
DataSeriesHydro::DataSeriesHydro() :
    timeseriesNumbers.register(ror, "ror"),
    timeseriesNumbers.register(storage, "storage"),
    timeseriesNumbers.register(mingen, "mingen")
{
    ...
}
```

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -339,6 +339,11 @@ void ISimulation<ImplementationType>::run()
         // for a single simulation
         study.resizeAllTimeseriesNumbers(1 + study.runtime->rangeLimits.year[Data::rangeEnd]);
         // Now, we will prepare the time-series numbers
+        if (not TimeSeriesNumbers::CheckNumberOfColumns(study))
```
guilpier-code: Good : resizing due to TS generation was already made at this stage

guilpier-code (changes requested):

src/libs/antares/study/include/antares/study/parts/hydro/series.h
```diff
@@ -144,10 +144,6 @@ class DataSeriesHydro
     */
     TimeSeries maxHourlyPumpPower;
 
-    // TS's number matrices for Generation and Maximum Power
-    Matrix<uint32_t> timeseriesNumbers;
-    Matrix<uint32_t> timeseriesNumbersHydroMaxPower;
-
     // Equalizing max generation and max pumping numbers of TS's    
     void EqualizeMaxPowerTSsizes(Area& area);
```
guilpier-code, line 147: We still have an **Equalize** function for max power pumping and generation.
We'll have to remove it at some point.

guilpier-code (changes requested):

src/libs/antares/study/include/antares/study/parts/hydro/series.h
flomnes: See #2068

src/solver/simulation/include/antares/solver/simulation/solver.hxx
flomnes: Yes !

src/solver/simulation/timeseries-numbers.cpp
flomnes: This has been done in PR #2080, see https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2080/files#diff-1bdb2c0970ab047ae53d0bcb2f829af2f797fac087c50d5f69ef08e9bbe217e1

src/solver/simulation/timeseries-numbers.cpp
flomnes: It can be safely removed without any effect.

src/libs/antares/series/include/antares/series/series.h
flomnes: Yes, will do

src/solver/simulation/timeseries-numbers.cpp
flomnes: I will remove thermal. I can split up the for loop if needed.

src/libs/antares/series/series.cpp
flomnes: I'd rather prevent copies, but it seems that `AreaScratchpad` must be copyable (used in std::vector), so build fails if I
- Delete copy/move constructor and operator
- Delete copy, keep move constructor and operator.

src/solver/simulation/timeseries-numbers.cpp
flomnes: DONE

src/solver/simulation/timeseries-numbers.cpp
flomnes: DONE and MERGED

src/solver/simulation/timeseries-numbers.cpp
flomnes: DONE

src/libs/antares/series/include/antares/series/series.h
flomnes: DONE

flomnes: @sylvlecl done here #2085

sylvlecl (approved):

guilpier-code (approved):

PR #2076, opened 2024-05-02
Bump upload-artifact to v4, use unique names for artifacts

### LINKED ISSUE #2075: upload-artifact@v3 deprecated, consider upgrading to v4
TL;DR
> Please see [the documentation](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) in the project repositories for guidance on how to migrate your workflows.

> The deprecation of v3 will be similar to the [previously announced v1 and v2 deprecation plans](https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/), **which is scheduled to take place on June 30, 2024**. Version tags will not be removed from the project repositories, however, **attempting to use a version of the actions after the deprecation date will result in a workflow failure**.

Source https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

flomnes:
[Migration doc](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md)

Instead of one big large artifact (named "artifact"), we get multiple artifacts
image

It may, or may not break the release process. **EDIT** Won't break, since **assets** are completely independent of **artifacts**.
```yml
    - name: Publish assets
      if: ${{ env.IS_RELEASE == 'true' }}
      env:
        tag: ${{ github.event.inputs.release_tag }}
      run: |
        gh release upload "$tag" _build/*.tar.gz _build/*.deb
```

### DIFF
--- a/.github/workflows/build-userguide.yml
+++ b/.github/workflows/build-userguide.yml
@@ -58,7 +58,7 @@ jobs:
         echo "PDF_PATH_ASSET= [\"docs/pdf-doc-generation-with-sphinx/$PDF_NAME\"]" >> $GITHUB_ENV
 
     - name: user guide upload
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: pdf-reference-guide
         path: ${{ env.PDF_PATH }}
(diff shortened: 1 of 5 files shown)


sylvlecl (approved):

flomnes:
Of course, CentOS7 fails. See [here](https://github.com/actions/runner/issues/2906) for the reason.

PR #2077, opened 2024-05-02
Use ubuntu-latest to upload assets & artifacts

flomnes:
~~Note the mount-point trick !~~ **EDIT** Data is not preserved from one job to the next one, unfortunately.

### DIFF
--- a/.github/workflows/centos7.yml
+++ b/.github/workflows/centos7.yml
@@ -139,5 +139,4 @@ jobs:
         GITHUB_TOKEN: ${{ github.token }}
         tag: ${{ github.event.inputs.release_tag }}
       run: |
-        apt install -y gh
         gh release upload "$tag" /build/*.tar.gz /build/*.rpm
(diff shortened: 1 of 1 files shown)


.github/workflows/centos7.yml
```diff
@@ -119,24 +114,27 @@ jobs:
...
       if: ${{ env.IS_RELEASE == 'true' }}
       env:
         GITHUB_TOKEN: ${{ github.token }}
         tag: ${{ github.event.inputs.release_tag }}
       run: |
-        gh release upload "$tag" _build/*.tar.gz _build/*.rpm
-
-  
\ No newline at end of file
+        apt install -y gh
```
flomnes: Installed by default on [GH's virtual environments](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cli-tools), not needed

flomnes:
Won't work, data isn't persistent across jobs. Only across steps within a job.

PR #2080, opened 2024-05-03
Add logs to checks on number of columns, refactor

flomnes:
Add more context to allow the user to correct invalid data.

Extract of logs after this PR
```
[2024-05-03 13:43:44][solver][error] Inconsistent number of columns for link fr / gb (direct-capacity: 3, indirect-capacity: 2)
[2024-05-03 13:43:44][solver][error] Please check that all series have the same number of columns or 1 column, or a combination of 1 column and the same number of columns
```

Also, small refactor to split the implementation of function `TimeSeriesNumbers::CheckNumberOfColumns` in 2.

Using a `std::map` induces a slight change of behavior : unique keys. TS with the same key will be erased. This is mostly good, for example if multiple `AreaScratchpad` are created for a given area (parallel simulations).


src/libs/antares/series/series.cpp
```diff
@@ -45,18 +46,37 @@ static bool checkAllElementsIdenticalOrOne(std::vector<uint> w)
     return std::adjacent_find(w.begin(), first_one, std::not_equal_to<uint>()) == first_one;
 }
 
-bool TimeSeries::Numbers::checkSeriesNumberOfColumnsConsistency() const
+static std::string errorMessage(const std::map<std::string, const TimeSeries*>& series)
```
guilpier-code, line 49: We should try : 
```C++
#include <boost/algorithm/string/join.hpp>
#include <boost/range/adaptors.hpp>
...
static std::string errorMessage(const std::map<std::string, const TimeSeries*>& series)
{
    return boost::algorithm::join(series | boost::adaptors::map_keys, ", ");
}
```
I did not test it, sorry

src/libs/antares/study/area/scratchpad.cpp
```diff
@@ -57,8 +57,8 @@ void CalculateDailyMeanPower(const Matrix<double>::ColumnType& hourlyColumn,
 }
 
 AreaScratchpad::AreaScratchpad(const StudyRuntimeInfos& rinfos, Area& area) :
- meanMaxDailyGenPower(area.hydro.series->timeseriesNumbersHydroMaxPower),
- meanMaxDailyPumpPower(area.hydro.series->timeseriesNumbersHydroMaxPower)
+  meanMaxDailyGenPower(area.hydro.series->timeseriesNumbersHydroMaxPower, "dailyGenPower"),
```
guilpier-code: As recommended in [PR 2073](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2073) (please see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2073#discussion_r1591156960)), we should not have to make this change

src/libs/antares/study/parts/common/cluster.cpp
```diff
@@ -30,7 +30,7 @@ namespace Antares::Data
 {
 Cluster::Cluster(Area* parent) :
     parentArea(parent),
-    series(tsNumbers)
+    series(tsNumbers, "thermal-availability")
```
guilpier-code: As recommended in https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2073 (please see https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2073#discussion_r1591156960), we should not have to make this change

src/libs/antares/study/parts/load/container.cpp
```diff
@@ -28,7 +28,7 @@ using namespace Yuni;
 
 namespace Antares::Data::Load
 {
-Container::Container() : prepro(nullptr), series(tsNumbers)
+Container::Container() : prepro(nullptr), series(tsNumbers, "load")
```
guilpier-code: No need for this change (see previous comment).

src/libs/antares/study/parts/solar/container.cpp
```diff
@@ -28,7 +28,7 @@ using namespace Yuni;
 
 namespace Antares::Data::Solar
 {
-Container::Container() : prepro(nullptr), series(tsNumbers)
+Container::Container() : prepro(nullptr), series(tsNumbers, "solar")
```
guilpier-code: No need for this change (see previous comment).

src/libs/antares/study/parts/wind/container.cpp
```diff
@@ -27,7 +27,7 @@ using namespace Yuni;
 
 namespace Antares::Data::Wind
 {
-Container::Container() : prepro(nullptr), series(tsNumbers)
+Container::Container() : prepro(nullptr), series(tsNumbers, "wind")
```
guilpier-code: No need for this change (see previous comment).

src/tests/src/libs/antares/study/series/timeseries-tests.cpp
```diff
@@ -42,7 +42,7 @@ struct Fixture
     Fixture(const Fixture&& f) = delete;
     Fixture& operator=(const Fixture& f) = delete;
     Fixture& operator=(const Fixture&& f) = delete;
-    Fixture() : ts(tsnum)
+    Fixture() : ts(tsnum, "")
```
guilpier-code: No need for this change (see previous comment).

guilpier-code (changes requested):

src/libs/antares/series/series.cpp
flomnes: It won't work, since we want to call this function on the keys
```cpp
[](const TimeSeries* s) { return s->numberOfColumns();}
```

guilpier-code: Ok. 
As, in the **std::map**'s values (reference to **TimeSeries**), we only use their number of columns, could we not directly have number of columns as values in this **std::map** ?
It would allow to use this boost join(...) algorithm

guilpier-code (changes requested):

flomnes:
@guilpier-code 
> Ok.
> As, in the std::map's values (reference to TimeSeries), we only use their number of columns, could we not directly have number of columns as values in this std::map ?
> It would allow to use this boost join(...) algorithm

Yes, but then we need to build a `std::map<std::string, int>`, which is a bit annoying.

PR #2084, opened 2024-05-13
Fix directory path for windows

sylvlecl:
**Analysis**

On windows:
```cpp
// this returns true as expected
fs::create_directories("/path/to/my/new/directory");

// thisreturns false because of trailing slash ???
fs::create_directories("/path/to/my/new/directory/");
```

It looks more or less like a bug of the STL ... the last slash is considered as "not created" which would explain it returns false ...

**The fix**

Use `parent_path` instead of `remove_filename`, to get the correct path of parent directory without the trailing slash.


sylvlecl:
Apparently considered "not a bug" ... 

https://developercommunity.visualstudio.com/t/stdfilesystemcreate-directories-returns-false-if-p/278829

payetvin (approved):

PR #2085, opened 2024-05-13
Use composition for TimeSeriesNumbers (instead of inheritance)

flomnes:
Also, use `timeSeriesNumbers[y]` instead of `timeSeriesNumbers[0][y]`.

### DIFF
--- a/src/libs/antares/series/series.cpp
+++ b/src/libs/antares/series/series.cpp
@@ -80,17 +80,6 @@ void TimeSeriesNumbers::reset(uint h)
     tsNumbers.reset(1, h);
 }
 
-namespace // anonymous
-{
-struct TSNumbersPredicate
-{
-    uint32_t operator()(uint32_t value) const
-    {
-        return value + 1;
-    }
-};
-} // namespace
-
 void TimeSeriesNumbers::clear()
 {
     tsNumbers.clear();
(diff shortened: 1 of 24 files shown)


src/libs/antares/series/series.cpp
```diff
@@ -61,6 +61,47 @@ static std::string errorMessage(const std::map<std::string, const TimeSeries*>&
...
+} // namespace
+
+void TimeSeriesNumbers::clear()
+{
+    tsNumbers.clear();
+}
+
+void TimeSeriesNumbers::saveToBuffer(std::string& data) const
+{
+    static TSNumbersPredicate predicate;
```
sylvlecl: We could probably use a lambda here and remove the struct, the function will be more self-contained.
Maybe it can also be `constexpr`?
By the way predicate seems like an odd name but it's an issue of the matric class !

sylvlecl (approved):
Nice simplification!

src/libs/antares/series/series.cpp
flomnes: Done ! https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2085/commits/50067f8e167197b9d409ac4c1782b0061f8cc9d5

It seems to work based on the fact that expression `object()` is valid both for `TSNumbersPredicate` (with `operator()`) and for a lambda function.

```cpp
template<class T, class ReadWriteT, class PredicateT>
void multiple_columns__dumper<T, ReadWriteT, PredicateT>::run()
{
// ...
        MatrixScalar<ReadWriteT>::Append(
          this->buffer_, (ReadWriteT)this->predicate_((this->mtx_)->entry[0][y]), this->format_);
// ...
}
```

PR #2089, opened 2024-05-14
Build boost from sources for CI

JasonMarechal25:


.github/workflows/compile-boost/action.yml
```diff
@@ -0,0 +1,20 @@
...
+  - id: install-boost
+    shell: bash
+    run: |
+      source /opt/rh/gcc-toolset-11/enable || true
+      export LD_LIBRARY_PATH=/usr/lib64/openmpi/lib:$LD_LIBRARY_PATH
+      export PATH=/usr/lib64/openmpi/bin:$PATH
+      wget https://github.com/boostorg/boost/releases/download/boost-1.81.0/boost-1.81.0.tar.gz
+      tar xvf boost-1.81.0.tar.gz
+      cd boost-1.81.0
+      ./bootstrap.sh --prefix=${{inputs.prefix}}/ --with-libraries=test
```
flomnes, line 19: We use other stuff as well
```cpp
#include <boost/algorithm/string/case_conv.hpp>
boost::to_lower(newgrp);
```

but I think this is header-only, so it may be installed regardeless of `--with-libraries` ? To be checked

flomnes (changes requested):
See #2091

PR #2091, opened 2024-05-14
Use VCPKG to install boost, also on Linux

sylvlecl:
This will allow us to easily select the boost version we want, independently from the OS.

In order to simplify things, this PR includes the **following change**
 - use header version of boost test

**Why**
currently, we use the dynamic link version. This implies to install boost test as a dynamic library, independently of other libraries.
This is feasible with VCPKG with a custom "triplet".
However, the use of this dynamic library has shown more issues than benefits:
- the actual benefits in size and build time are completely negligible
- the fact that we need the dynamic link version si hidden: if a user installs the static version, he will not understand why tests don't run. By using the header version, we ensure the tests will work in any case.


sylvlecl:
This has been handled in #2127, we now link statically with boost-test.

PR #2092, opened 2024-05-14
Revert "Bump upload-artifact to v4, use unique names for artifacts"

flomnes:
Reverts AntaresSimulatorTeam/Antares_Simulator#2076

OL8 & CentOS7 are incompatible with NodeJS 20. There seems to be no obvious solution.


flomnes:
See #2093

PR #2098, opened 2024-05-16
Add BOOST_TEST_DYN_LINK definition (CMake), fix build on OL8

flomnes:


JasonMarechal25:
Can you explain why OL8 particulary need DYN_LINK ?

flomnes:
@JasonMarechal25 We'll just revert to the pre-#1985 state.

PR #2103, opened 2024-05-21
Fix enum deprecation warning for DataLevel

flomnes:
On modern compilers (gcc>=11, clang>=17), this enum appears hundreds if not thousands of times because of templates, so it's very annoying.


sylvlecl (approved):
Ok! I did not think of completely removing the enum, because it goes in the direction of "less typing", but indeed it's more consistent with the current usage to having const ints !

PR #2104, opened 2024-05-21
Fix enum deprecation warning for FileLevel

flomnes:
On modern compilers (gcc>=11, clang>=17), this enum appears hundreds if not thousands of times because of templates, so it's very annoying.

### DIFF
--- a/src/solver/variable/include/antares/solver/variable/categories.h
+++ b/src/solver/variable/include/antares/solver/variable/categories.h
@@ -16,7 +16,7 @@
 ** Mozilla Public Licence 2.0 for more details.
 **
 ** You should have received a copy of the Mozilla Public Licence 2.0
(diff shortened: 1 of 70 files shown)


sylvlecl (approved):

src/solver/variable/include/antares/solver/variable/categories.h
```diff
@@ -16,7 +16,7 @@
 ** Mozilla Public Licence 2.0 for more details.
 **
 ** You should have received a copy of the Mozilla Public Licence 2.0
-** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+* along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
```
flomnes:
```suggestion
** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
```

PR #2109, opened 2024-05-22
inifile : forgotten commit of an previous PR

guilpier-code:
See #2064


flomnes:
@guilpier-code Can you use interactive rebase to include only the latest commit (fc0ef2f2f5f6350a64fbcc8488e126e829ce24fa) ? Currently you have lots of conflicts, and you're trying to merge many commits that have already been merged. Thank you

==> @guilpier-code  says : **Done**

src/libs/antares/inifile/include/antares/inifile/inifile.h
```diff
@@ -43,22 +42,21 @@ class IniFile final
...
+        Property() = default;
         explicit Property(const AnyString& key);
         template<class U>
         Property(const AnyString& key, const U& value);
-        ~Property();
+        ~Property() = default;
 
-        template<class StreamT>
-        void saveToStream(StreamT& file, uint64_t& written) const;
+        void saveToStream(std::ostream& file, uint64_t& written) const;
```
a-zakir: ofstream is more suitable for "file", unless file is not only a file

flomnes: We use `std::stringstream` for unit tests (in-memory tests)

src/libs/antares/inifile/inifile.cpp
```diff
@@ -48,6 +48,11 @@ inline void IniFile::Property::saveToStream(std::ostream& stream_out, uint64_t&
     stream_out << key << " = " << value << '\n';
 }
 
+void IniFile::Section::add(const Property property)
```
flomnes:
```suggestion
void IniFile::Section::add(const Property& property)
```

flomnes (changes requested):
Apply formatting (clang-format)

flomnes (approved):

PR #2112, opened 2024-05-24
Fix target branch for new releases

flomnes:
```
gh release create --help
      --target branch                Target branch or full commit SHA (default [main branch])
```


flomnes:
target_branch is already present

PR #2115, opened 2024-05-24
Fix missing synthesis results for links

flomnes:
Regression introduced in v9.1.0-rc2 or v9.1.0-rc3

`buildSurveyReport` for links is not called if `isPrinted[0] == false` for the STSbyGroup variable. This is not an intended behavior.

### DIFF
--- a/src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
+++ b/src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
@@ -402,7 +402,8 @@ public:
         // ------------------------------
         if (!AncestorType::isPrinted[0])
         {
-            return;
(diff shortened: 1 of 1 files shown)


src/solver/variable/include/antares/solver/variable/economy/STSbyGroup.h
```diff
@@ -398,32 +399,31 @@ class STSbyGroup: public Variable::IVariable<STSbyGroup<NextT>, NextT, VCardSTSb
                            int fileLevel,
                            int precision) const
     {
-        // Building syntheses results
+        // Building synthesis results
         // ------------------------------
-        if (!AncestorType::isPrinted[0])
-        {
-            return;
-        }
```
flomnes, line 406: The issue was here

guilpier-code (approved):

PR #1751, opened 2023-11-08
Expose solver as a lib

JasonMarechal25:
Expose solver as a lib to allow external application (e.g. xpansion) to link to it and perform simulation directly without invoking external process in CLI.

- Move Application as a separate library. Scaffold for future work
- Expose solver-lib library. Existing solver executable now link to this library
- Refactor several solver libraries to fit proper library structure. Better and necessary to resolve some path.

PR #2128, opened 2024-05-29
Parallel years cleaning : few removal of "numspace"

guilpier-code:
See ticket [#1751](https://gopro-tickets.rte-france.com/browse/ANT-1751)

For now, just a first analyze of how we could do to isolate **years** (from the simulation and what should not be inside a **year**).
We take advantage of this to make some removals of **numSpace**.

**Comments** : 
- **year job** and **simulation** are really intertwined, this is a first and important reason why a **year** are not isolated. 
  For instance, when running a year of simulation, we call a function of the **simulation** object, not of the **year** object.
- although what we first thought, **each year job creates a fresh new weekly optimization solver object**. This object solves every week problem, using the previous week's optimization solution as a base. Before creating a solver object, a year first **destroys the previous year's solver** (as well as the start base it holds).
- I've heard that CI tests for parallel (currently passed at night run) don't generate the same output results as the same tests run in sequential. We assumed that it was because of the bases. Indeed, if the same simulation year is run starting from 2 different bases, it can lead to different simulation results. But given the previous point, that's not true : the simulation results should be identical whether we run a simulation sequentially or not. We may have broken something in the parallelism by not testing the it on CI.

**Goals** : 
We want to isolate year jobs to make cleaner parallelism.
Each year job should be designed so that it can be run in a thread independently from any other year job.
In particular, a year job should hold its own memory spaces, and not shared memory as done currently.


src/solver/hydro/management/management.cpp
```diff
@@ -117,12 +117,10 @@ double BetaVariable(double a, double b, MersenneTwister& random)
 HydroManagement::HydroManagement(const Data::AreaList& areas,
                                  const Data::Parameters& params,
                                  const Date::Calendar& calendar,
-                                 unsigned int maxNbYearsInParallel,
                                  Solver::IResultWriter& resultWriter):
     areas_(areas),
     calendar_(calendar),
     parameters_(params),
-    maxNbYearsInParallel_(maxNbYearsInParallel),
```
guilpier-code, line 125: **maxNbYearsInParallel_** unused, and for good reasons : inside a year, why would we need the number of parallel years ?
So we remove it.

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -113,65 +113,6 @@ static void RecalculDesEchangesMoyens(Data::Study& study,
     }
 }
 
-void PrepareDataFromClustersInMustrunMode(Data::Study& study,
```
guilpier-code: This function (**PrepareDataFromClustersInMustrunMode**) has many drawbacks : 
- It's too long
- It's supposed to offer a treatment common to mode **adequacy** and **economy** but its body clearly separates **adequacy** mode and the other modes (**economy**)
- it's complicated : for example, we have 2 loops over thermal clusters of an area, though we could only have one.

So we removed this function from **common-eco-adq.cpp** and put down these calculations in specialized functions associated to each simulation mode.

src/solver/simulation/adequacy.cpp
```diff
@@ -411,7 +411,33 @@ void Adequacy::simulationEnd()
 
 void Adequacy::prepareClustersInMustRunMode(Data::Area::ScratchMap& scratchmap, uint year)
 {
-    PrepareDataFromClustersInMustrunMode(study, scratchmap, year);
-}
+    for (uint i = 0; i < study.areas.size(); ++i)
+    {
+        auto &area = *study.areas[i];
+        auto &scratchpad = scratchmap.at(&area);
```
flomnes:
```suggestion
        auto *area = study.areas[i];
        auto &scratchpad = scratchmap.at(area);
```

src/solver/simulation/economy.cpp
```diff
@@ -272,7 +272,21 @@ void Economy::simulationEnd()
 
 void Economy::prepareClustersInMustRunMode(Data::Area::ScratchMap& scratchmap, uint year)
 {
-    PrepareDataFromClustersInMustrunMode(study, scratchmap, year);
+    for (uint i = 0; i < study.areas.size(); ++i)
+    {
+        auto& area = *study.areas[i];
+        auto& scratchpad = scratchmap.at(&area);
```
flomnes:
```suggestion
        auto* area = study.areas[i];
        auto& scratchpad = scratchmap.at(area);
```

flomnes (approved):

src/solver/simulation/adequacy.cpp
```diff
@@ -68,15 +68,7 @@ bool Adequacy::simulationBegin()
         pProblemesHebdo.resize(pNbMaxPerformedYearsInParallel);
         for (uint numSpace = 0; numSpace < pNbMaxPerformedYearsInParallel; numSpace++)
         {
-            SIM_InitialisationProblemeHebdo(study, pProblemesHebdo[numSpace], 168, numSpace);
```
guilpier-code, line 71: Here we used to give value 168 as argument, and then check its value (through problem hebdo).
- What the point in using a hard coded value end then check its value
- Furthermore, **nbHoursInAWeek** is existing, is more meaningful, and is used elsewhere in this source file. So why not use it ?

We use **nbHoursInAWeek** instead of **168** and remove the useless check.

flomnes (changes requested):
Thanks for the awesome work @guilpier-code ! Can you split this PR in 2 ? The double count issue for adequacy needs a bit more investigation

src/solver/simulation/common-eco-adq.cpp
flomnes: See #2168

flomnes (approved):

PR #2132, opened 2024-05-30
Perform hydro checks prior to the simulation [ANT-1720] 

a-zakir:


src/solver/application/include/antares/application/ScenarioBuilderOwner.h
```diff
@@ -0,0 +1,42 @@
...
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+#include "antares/study/study.h"
+
+namespace Antares {
```
payetvin: namespace Antares::Solver

src/solver/hydro/management/HydroInputsChecker.cpp
```diff
@@ -0,0 +1,211 @@
...
+        //            prepareEffectiveDemand();
+        //            prepareMonthlyOptimalGenerations(randomReservoirLevel,
+        //            year);
+        //-----------------------------
+    }
+}
+
+bool HydroInputsChecker::checksOnGenerationPowerBounds(uint year) const
+{
+    return (checkMinGeneration(year) && checkGenerationPowerConsistency(year)) ? true : false;
```
```suggestion
    return checkMinGeneration(year) && checkGenerationPowerConsistency(year);
```

src/solver/hydro/management/daily.cpp
```diff
@@ -227,7 +227,7 @@ inline void HydroManagement::prepareDailyOptimalGenerations(
 {
     const auto srcinflows = area.hydro.series->storage.getColumn(y);
 
-    auto& data = tmpDataByArea_[&area];
+    auto& data = area.hydro.hydro_management_data[y];
```
flomnes: This is actually a pretty bad idea, since `area` is shared by threads you'll get race conditions, etc.

flomnes (changes requested):
I think (and @guilpier-code & @sylvlecl will agree) that adding temporary data to a `Study` is a bad idea. It would be best to keep it outside as it is now.

a-zakir:
> I think (and @guilpier-code & @sylvlecl will agree) that adding temporary data to a `Study` is a bad idea. It would be best to keep it outside as it is now.

actually tmpDataByArea is not temporary

src/solver/hydro/management/daily.cpp
a-zakir: ok

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -315,7 +317,7 @@ void ISimulation<ImplementationType>::run()
 
     ImplementationType::setNbPerformedYearsInParallel(pNbMaxPerformedYearsInParallel);
 
-    TSGenerator::ResizeGeneratedTimeSeries(study.areas, study.parameters);
+    // TSGenerator::ResizeGeneratedTimeSeries(study.areas, study.parameters);
```
flomnes: remove commented code

flomnes (commented):
1. `hydro_management_data` -> `managementData`

src/solver/simulation/include/antares/solver/simulation/solver.hxx
a-zakir: done

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,50 @@
...
+    double MLE[12];
+    //! Monthly optimal generation
+    double MOG[12];
+    //! Monthly optimal level
+    double MOL[12];
+    //! Monthly target generations
+    double MTG[12];
+    //! inflows
+    double inflows[12];
+    //! monthly minimal generation
```
flomnes: Use `std::array` instead ?

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,50 @@
...
+    std::array<double, 12> mingens;
+
+    //! Net demand, for each day of the year, for each area
+    double DLN[dayYearCount];
+    //! Daily local effective load
+    double DLE[dayYearCount];
+    //! Daily optimized Generation
+    double DOG[dayYearCount];
+    //! daily minimal generation
```
flomnes: `std::array` ?

src/solver/application/ScenarioBuilderOwner.cpp
```diff
@@ -0,0 +1,56 @@
...
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+
+#include "antares/application/ScenarioBuilderOwner.h"
+#include <antares/antares/fatal-error.h>
+#include "antares/solver/simulation/apply-scenario.h"
+#include "antares/solver/simulation/timeseries-numbers.h"
+#include "antares/solver/ts-generator/generator.h"
```
flomnes: Use `< >`

src/solver/application/application.cpp
```diff
@@ -33,6 +33,7 @@
 #include <antares/sys/policy.h>
 #include <antares/writer/writer_factory.h>
 #include "antares/antares/version.h"
+#include "antares/application/ScenarioBuilderOwner.h"
 #include "antares/config/config.h"
 #include "antares/file-tree-study-loader/FileTreeStudyLoader.h"
 #include "antares/signal-handling/public.h"
```
flomnes: Use `< >`

src/solver/application/include/antares/application/ScenarioBuilderOwner.h
```diff
@@ -0,0 +1,39 @@
...
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+#include "antares/study/study.h"
```
flomnes: Forward declaration is enough here (only refs to `Data::Study`, no need to know what's inside)

src/solver/hydro/include/antares/solver/hydro/management/HydroInputsChecker.h
```diff
@@ -0,0 +1,62 @@
...
+** Antares_Simulator is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+#include <antares/study/area/area.h>
```
flomnes, line 22: You may include study.h here

src/solver/hydro/management/HydroInputsChecker.cpp
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        //---------------------  May be later ? //
+        //            prepareNetDemand(year,
+        //            parameters_.mode,
+        //            scratchmap);
+        //            prepareEffectiveDemand();
+        //            prepareMonthlyOptimalGenerations(randomReservoirLevel,
+        //            year);
+        //-----------------------------
```
flomnes: Remove this commented code ?

src/solver/hydro/include/antares/solver/hydro/management/HydroInputsChecker.h
```diff
@@ -0,0 +1,62 @@
...
+{
+
+class HydroInputsChecker
+{
+public:
+    HydroInputsChecker(Antares::Data::Study& study, Solver::IResultWriter& resultWriter);
+    void Execute();
+
+private:
+    Antares::Data::Study& study_;
```
flomnes: I see that you already have a few references to member data of the Study class. If you can, avoid referencing Study, since it's a "god class", it makes it difficult to see what is actually meant to be used.

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,63 @@
...
+    //! Monthly optimal level
+    std::array<double, 12> MOL;
+    //! Monthly target generations
+    std::array<double, 12> MTG;
+};
+
+// TODO Rename
+struct HydroSpecific
+{
+    HydroDaily daily;
```
flomnes: You may want to use `double`s in `HydroDaily`, and an `std::array<HydroDaily, 12>` here, so you have 1 array instead of multiple. The same goes for monthly data.

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,63 @@
...
+    //! Net demand, for each day of the year, for each area
+    std::array<double, dayYearCount> DLN;
+    //! Daily local effective load
+    std::array<double, dayYearCount> DLE;
+};
+
+struct HydroMonthly
+{
+    //! Monthly local effective demand
+    std::array<double, 12> MLE;
```
flomnes: You may use the comment above to rename these variables. Quite unexpressive right now

src/solver/application/include/antares/application/ScenarioBuilderOwner.h
```diff
@@ -0,0 +1,44 @@
...
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+namespace Antares
```
flomnes, line 23: Use namespace nesting from cpp17

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,63 @@
 
 namespace Antares::Data
 {
+
+enum dayYearCount
+{
+    //! The maximum number of days in a year
+    dayYearCount = 366
```
flomnes: Use `constexpr` if possible

flomnes (commented):
Mostly style & improvement suggestions.

If a function only uses `HydroSpecific::monthly`, you can just pass the `monthly` object. Same for `daily`.

src/libs/antares/study/include/antares/study/parts/hydro/container.h
a-zakir: thank you

a-zakir:
> Mostly style & improvement suggestions.
> 
> If a function only uses `HydroSpecific::monthly`, you can just pass the `monthly` object. Same for `daily`.

data structs have changed, i think the code is tidy enough & Antares::Data::TimeDependantHydroManagementData is already compact

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,60 @@
 
 namespace Antares::Data
 {
+
+    //! The maximum number of days in a year
+constexpr size_t dayYearCount = 366;
+
+struct HydroDailyLoadDemand
```
guilpier-code: **HydroDailyLoadDemand** : there is some kind of repetition here.
**load** and **demand** are synonymous.
**HydroDailyDemand** is enough. 
Given the context we're in (hydro), **DailyDemand** could even do the job.

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,60 @@
...
+{
+    //! Net demand, for each day of the year, for each area
+    double DLN = 0.;
+    //! Daily local effective load
+    double DLE = 0.;
+};
+
+struct HydroMonthlyDemandGeneration
```
guilpier-code: **HydroMonthlyDemandGeneration** : 
- This **struct** gather things that are quite different : **MLE** is a demand (or load if you prefer), whereas **MTG** is an hydro generation, and **MOL** is an hydro level.
- The name of the **struct** doesn't make sense to me. What is a hydro monthly demand generation ?

In the end, what we want to compute here is a hydro monthly generation target.
So a **struct** named **MonthlyGenerationTargetData** would make more sense.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -1015,65 +989,74 @@ void ISimulation<ImplementationType>::loopThroughYears(uint firstYear,
...
     pQueueService->maximumThreadCount(pNbMaxPerformedYearsInParallel);
+    HydroInputsChecker hydroInputsChecker(study);
 
-    // Loop over sets of parallel years
-    std::vector<setOfParallelYears>::iterator set_it;
-    for (set_it = setsOfParallelYears.begin(); set_it != setsOfParallelYears.end(); ++set_it)
+    // Loop over sets of parallel years to check hydro inputs
+    for (const auto& batch: setsOfParallelYears)
+    {
+        // if refresh break
```
guilpier-code: About `// if refresh break` : 
This comment is less clear than the code. 
Code is clear enough. We should remove the comment.

src/solver/application/ScenarioBuilderOwner.cpp
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+#include <antares/antares/fatal-error.h>
+#include <antares/application/ScenarioBuilderOwner.h>
+#include "antares/solver/simulation/apply-scenario.h"
+#include "antares/solver/simulation/timeseries-numbers.h"
+#include "antares/solver/ts-generator/generator.h"
+#include "antares/study/study.h"
+
+Antares::Solver::ScenarioBuilderOwner::ScenarioBuilderOwner(Data::Study& study):
```
guilpier-code, line 29: Class **ScenarioBuilderOwner** : not sure we need a class in order to call free functions about TS numbers.
Is it not a bit over engineered ?

src/solver/application/ScenarioBuilderOwner.cpp
```diff
@@ -0,0 +1,57 @@
...
+    study_(study)
+{
+}
+
+void Antares::Solver::ScenarioBuilderOwner::callScenarioBuilder()  {
```
guilpier-code, line 34: Class **ScenarioBuilderOwner** : 
This name suggests the class is about **scenario builder** and only **scenario builder**.
Looking closely, this class is about : 
- **resizing** time series,  
- **resizing** TS numbers to the number of MC years, 
- **generating** TS numbers, 
- and finally applying (taking only TS numbers chosen by user) the scenario builder rules.

The 3 first bullets are not really related to scenario builder. 
It is misleading.
Either we think harder to find a better name (I'm afraid there isn't any), or we leave the original function calls.

src/solver/hydro/management/HydroInputsChecker.cpp
```diff
@@ -0,0 +1,206 @@
...
+    {
+        checked_years_[year] = false;
+    }
+}
+
+void HydroInputsChecker::Execute(uint year)
+{
+    if (!checked_years_[year])
+    {
+        checked_years_[year] = true;
```
guilpier-code: **checked_years_** : seems useless.
Why do we need this data member ?

If it turns out to be useless, remove data members **firstYear_** and **endYear_**, useless too.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -1015,65 +989,74 @@ void ISimulation<ImplementationType>::loopThroughYears(uint firstYear,
 
     // Number of threads to perform the jobs waiting in the queue
     pQueueService->maximumThreadCount(pNbMaxPerformedYearsInParallel);
+    HydroInputsChecker hydroInputsChecker(study);
 
-    // Loop over sets of parallel years
-    std::vector<setOfParallelYears>::iterator set_it;
-    for (set_it = setsOfParallelYears.begin(); set_it != setsOfParallelYears.end(); ++set_it)
+    // Loop over sets of parallel years to check hydro inputs
+    for (const auto& batch: setsOfParallelYears)
```
guilpier-code, line 968: About block : 
```C++
for (const auto& batch: setsOfParallelYears)
{
    if (batch.regenerateTS)
    {
        break;
    }
    for (auto year: batch.yearsIndices)
    {
        hydroInputsChecker.Execute(year);
    }
}
```
Is this block useful ?
Indeed, there is a further loop over batches, and there's a check on hydro input as well.
With this first block, don't we double check ?

guilpier-code, line 968: **I'm sure you're aware** that, by checking hydro input just before a batch of MC years is run, we still have the problem for which we opened this PR :  if a year's hydro data is not valid inside a batch, we stop in the middle of a simulation. 
And if the simulation is run in sequential mode, we gain nothing.

I know that **hydro TS refresh** is the cause of this poor gain we've here : it prevents us to move checks on hydro input before any batch of MC year is run.

Unless TS refresh disappears, we'll be hampered with this project.

Possible solutions : 
- As discussed with @a-zakir, a temporary solution could be : 
  - to reset the hydro random generator after TS were generated a first time **only for needs of hydro input checks**, 
  - and then to generate the hydro Ts a second time for real hydro management.
- If we succeed in having the checks on hydro input before running the MC years batch, we still have to store hydro data used for each area and year. Why not keep the hydro checks inside MC years, but just not run years for which data is not valid ?

src/solver/hydro/include/antares/solver/hydro/management/HydroInputsChecker.h
```diff
@@ -0,0 +1,62 @@
...
+private:
+    Data::AreaList& areas_;
+    const Data::Parameters& parameters_;
+    const Date::Calendar& calendar_;
+    Data::SimulationMode simulationMode_;
+    const uint firstYear_;
+    const uint endYear_;
+    PrepareInflows prepareInflows_;
+    MinGenerationScaling minGenerationScaling_;
+    bool refresh_ = false;
```
guilpier-code: - `bool refresh_ = false;`
- `const Data::Parameters& parameters_`
- `Data::SimulationMode simulationMode_;`

These data members are unused, and should be removed

src/libs/antares/study/include/antares/study/parts/hydro/container.h
```diff
@@ -28,6 +28,60 @@
...
+
+//!  Hydro Management Data for a given area
+struct TimeDependantHydroManagementData
+{
+    std::array<HydroDailyLoadDemand, dayYearCount> daily{0};
+    std::array<HydroMonthlyDemandGeneration, 12> monthly{0};
+};
+
+//! Area Hydro Management Data for a given year
+struct AreaDependantHydroManagementData
```
guilpier-code, line 64: **struct TimeDependantHydroManagementData** : 
This **struct** is used only during hydro ventilation (hydro management if you prefer).
It's not used when checking hydro data.
So we could ask : why is it defined here, and not in class **HydroManagement** ?

Same for other struct above.

guilpier-code (dismissed):

src/solver/application/ScenarioBuilderOwner.cpp
a-zakir: i like classes, it's chic

src/solver/application/ScenarioBuilderOwner.cpp
a-zakir: conservative are you?

src/solver/hydro/management/HydroInputsChecker.cpp
a-zakir: i'm the batch n°2 (with TS refresh 💧 ) , the batch before me didn't refresh so he was checked in the 1st loop right?
[myself batch n2] i'm thrown off the 1st loop and my checks are done in the 2nd loop

src/solver/simulation/include/antares/solver/simulation/solver.hxx
a-zakir: nope thanks to checked_years_

a-zakir: i understand it's a bit confusing

src/libs/antares/study/include/antares/study/parts/hydro/container.h
a-zakir: renamed

src/libs/antares/study/include/antares/study/parts/hydro/container.h
a-zakir: renamed

src/solver/simulation/include/antares/solver/simulation/solver.hxx
a-zakir: Ts generation on the fly and refresh will be abandoned soon, therefore the outer loop ensure that hydro checks are done before the simulation loop

src/solver/hydro/include/antares/solver/hydro/management/HydroInputsChecker.h
a-zakir: thanks they were used in a previous version

src/solver/application/ScenarioBuilderOwner.cpp
guilpier-code: > conservative are you?

I'm sorry, I don't understand your answer, could be more specific ?

src/solver/hydro/management/HydroInputsChecker.cpp
flomnes: Yes you can just ignore refresh

src/solver/application/ScenarioBuilderOwner.cpp
guilpier-code: I've nothing against classes, I'm even a fan of classes. 
But this class contain heterogeneous things (see the [comment below](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2132#discussion_r1646523749) ), and its name does

src/solver/simulation/include/antares/solver/simulation/solver.hxx
guilpier-code: Still on this hydro checking **loop over batches** : 
TS in general, and hydro TS in particular are generated by a call to **regenerateTimeSeries**.
When running this loop (before main loop over batches), **hydro TS haven't been generated** yet, they were only resized by calling **TSGenerator :: ResizeGeneratedTimeSeries**.
So the checks on hydro input are based on hydro TS for which every value is **zero**.

a-zakir: not all time regenerateTimeSeries is called, it's the case when (generate & refresh) are used. This options will be obsolete soon and all TS won"t be zero

guilpier-code: Ok, my bad, I missed something

src/solver/hydro/management/HydroInputsChecker.cpp
flomnes: checked_years_ was removed.

flomnes (approved):

PR #2138, opened 2024-05-31
Restore correct behavior when options -h/--list-solvers are provided

flomnes:
#2056 introduced a slight change of behavior : after this PR, when the user requests help (--help) or queries the available OR-Tools solvers (--list-solvers), a warning message is displayed after the requested information.

This is a bit annoying, so here is a fix.

### DIFF
--- a/src/solver/application/application.cpp
+++ b/src/solver/application/application.cpp
@@ -63,7 +63,7 @@ Application::Application()
     resetProcessPriority();
 }
 
-void Application::parseCommandLine(Data::StudyLoadOptions& options)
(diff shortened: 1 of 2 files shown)


src/solver/application/application.cpp
```diff
@@ -337,7 +338,8 @@ void Application::prepare(int argc, char* argv[])
 
     parseCommandLine(options);
 
-    handleOptions(options);
+    if (!handleOptions(options)) // -h, --list-solvers, etc.
+       return;
```
flomnes: No need to perform checks, the user just wants help / list solvers

JasonMarechal25 (approved):

PR #2140, opened 2024-05-31
Fix circular dependencies on CMake targets

### LINKED ISSUE #2137: circular dependency
circular dependency between Antares::solverUtils lib and model_antares

a-zakir:
close #2137


.github/workflows/centos7.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches:
       - develop
       - dependabot/*
+      - fix/circular-deps
```
flomnes:
```suggestion
```

flomnes (approved):

flomnes (approved):

PR #2143, opened 2024-06-05
Remove state from hydro ventilation

guilpier-code:
**State** type is passed to the hydro ventilation.
The object **state** is used by the hydro ventilation to reach the weekly optimization problem, in order to update the hydro initial level of a MC year.
This important update is hidden in the depth of the code whereas it should be more highlighted (when a year starts).
Furthermore, removing state alleviates a bit the hydro ventilation, already too complicated.

Eventually, removing dependency of class **HydroManagement** to **State** allows to : 
- highlight the update of hydro initial level when beginning a year
-  have class **HydroManagement** loose a dependency to **State**

### DIFF
--- a/src/solver/hydro/include/antares/solver/hydro/management/management.h
+++ b/src/solver/hydro/include/antares/solver/hydro/management/management.h
@@ -112,7 +112,6 @@ public:
 
     //! Perform the hydro ventilation
     void makeVentilation(double* randomReservoirLevel,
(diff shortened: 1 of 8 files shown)


a-zakir:
very nice!

a-zakir (approved):

JasonMarechal25 (approved):

PR #2149, opened 2024-06-07
Move data validation to specialized functions [ANT-1213]

payetvin:
The goal of this PR is to separate loading and validation of data.
This separation will allow to have a modular loading.

### DIFF
--- a/src/libs/antares/study/include/antares/study/parts/hydro/prepro.h
+++ b/src/libs/antares/study/include/antares/study/parts/hydro/prepro.h
@@ -97,6 +97,7 @@ public:
     */
     bool loadFromFolder(Study& s, const AreaName& areaID, const char folder[]);
 
+    bool validate(const std::string& areaID) const;
     /*!
     ** \brief Save hydro settings for the prepro into a folder
     **
--- a/src/libs/antares/study/include/antares/study/parts/hydro/prepro.h
+++ b/src/libs/antares/study/include/antares/study/parts/hydro/prepro.h
@@ -95,9 +95,9 @@ public:
     ** \param folder The source folder (ex: `input/hydro/prepro`)
     ** \return A non-zero value if the operation succeeded, 0 otherwise
     */
-    bool loadFromFolder(Study& s, const AreaName& areaID, const char folder[]);
+    bool loadFromFolder(Study& s, const AreaName& areaID, const std::string& folder);
 
-    bool validate(const std::string& areaID) const;
+    bool validate(const std::string& areaID);
     /*!
     ** \brief Save hydro settings for the prepro into a folder
     **
--- a/src/libs/antares/study/parts/hydro/container.cpp
+++ b/src/libs/antares/study/parts/hydro/container.cpp
@@ -281,7 +281,7 @@ bool PartHydro::LoadFromFolder(Study& study, const AnyString& folder)
     if ((section = ini.find("inter-daily-breakdown")))
     {
         ret = loadProperties(study, section->firstProperty, buffer, [](IniFile::Property* p, Area* area){
-            return p->value.to<double>(area->hydro.intermonthlyBreakdown);
+            return p->value.to<double>(area->hydro.interDailyBreakdown);
         }) && ret;
     }
 
--- a/src/libs/antares/study/parts/hydro/container.cpp
+++ b/src/libs/antares/study/parts/hydro/container.cpp
@@ -108,12 +108,11 @@ void PartHydro::reset()
 
 using setProperty = std::function<bool (IniFile::Property*, Area*)>;
 
-static bool loadProperties(Study& study, IniFile::Property* property, const std::string& filename, setProperty funcToSetProperty)
+template<class T>
+static bool loadProperties(Study& study, IniFile::Property* property, const std::string& filename, T PartHydro::*ptr)
 {
     if (!property)
         return false;
-
-    bool ret = true;
 
     // Browse all properties
     for (; property; property = property->next)
@@ -124,14 +123,15 @@ static bool loadProperties(Study& study, IniFile::Property* property, const std:
         Area* area = study.areas.find(id);
         if (area)
         {
-            ret = funcToSetProperty(property, area) && ret;
+            return property->value.to<T>(area->hydro.*ptr);
         }
         else
         {
             logs.warning() << filename << ": `" << id << "`: Unknown area";
+            return false;
         }
     }
-    return ret;
+    return true;
 }
 
 bool PartHydro::LoadFromFolder(Study& study, const AnyString& folder)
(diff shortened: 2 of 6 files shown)


src/libs/antares/study/include/antares/study/parts/hydro/prepro.h
```diff
@@ -95,8 +95,9 @@ class PreproHydro
     ** \param folder The source folder (ex: `input/hydro/prepro`)
     ** \return A non-zero value if the operation succeeded, 0 otherwise
     */
-    bool loadFromFolder(Study& s, const AreaName& areaID, const char folder[]);
+    bool loadFromFolder(Study& s, const AreaName& areaID, const std::string& folder);
```
JasonMarechal25, line 98: Is it possible (or usefull) to use std::filesystem::path for folder ?

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
```diff
@@ -47,6 +47,24 @@ HydroMaxTimeSeriesReader::HydroMaxTimeSeriesReader(PartHydro& hydro,
...
+{
+    for (uint i = 0; i < 4U; ++i)
+    {
+        auto& col = dailyMaxPumpAndGen[i];
+        for (uint day = 0; day < DAYS_PER_YEAR; ++day)
+        {
+            if (col[day] < 0. || (i % 2U /*column hours*/ && col[day] > 24.))
+            {
+                logs.error() << areaName << ": invalid power or energy value";
+                return false;
```
JasonMarechal25, line 60: Carefull
With the early return you are returning after the first error. Previous behaviour is to check all values and log all errors.

src/libs/antares/study/parts/hydro/prepro.cpp
```diff
@@ -145,11 +145,10 @@ bool PreproHydro::saveToFolder(const AreaName& areaID, const char* folder)
     return false;
 }
 
-bool PreproHydro::loadFromFolder(Study& s, const AreaName& areaID, const char* folder)
+bool PreproHydro::loadFromFolder(Study& s, const AreaName& areaID, const std::string& folder)
 {
     /* Asserts */
-    assert(folder);
-    assert('\0' != *folder);
+    assert(!folder.empty());
```
JasonMarechal25: Asserts only occur in debug. @flomnes Should we create a dedicated eror or exception for this ?

src/libs/antares/study/parts/hydro/prepro.cpp
```diff
@@ -161,6 +160,17 @@ bool PreproHydro::loadFromFolder(Study& s, const AreaName& areaID, const char* f
 
     buffer.clear() << folder << SEP << areaID << SEP << "prepro.ini";
     bool ret = PreproHydroLoadSettings(this, buffer);
+
+    buffer.clear() << folder << SEP << areaID << SEP << "energy.txt";
+    ret = data.loadFromCSVFile(buffer, hydroPreproMax, 12, mtrxOption, &s.dataBuffer) && ret;
```
JasonMarechal25: Avoid magic number "12". Please declare a constant for it

JasonMarechal25 (dismissed):

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -106,6 +106,34 @@ void PartHydro::reset()
...
 }
 
+using setProperty = std::function<void (IniFile::Property*, Area*)>;
+
+static bool loadProperties(Study& study, IniFile::Property* property, const std::string& filename, setProperty funcToSetProperty)
+{
+    if (!property)
+        return false;
+
+    bool ret = true;
```
flomnes, line 118: useless ?

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -106,6 +106,34 @@ void PartHydro::reset()
...
+        if (area)
+        {
+            funcToSetProperty(property, area);
+        }
+        else
+        {
+            logs.warning() << filename << ": `" << id << "`: Unknown area";
+        }
+    }
+    return ret;
```
flomnes, line 137: This function always returns true. Maybe `funcToSetProperty` should return a `bool` ? Since these functions can issue an error, it would make sense.

src/libs/antares/study/parts/hydro/prepro.cpp
flomnes: No need, this check is irrelevant. If `folder` is empty, we'll have an error later (file/directory does not exist, probably).

flomnes (changes requested):
Please split loading / validation.
```cpp
bool loadParameters();
bool validateParameters(); // if (area.hydro.xxx < 1.) { return false;}, etc.
```

src/libs/antares/study/parts/hydro/container.cpp
payetvin: ret is supposed to be captured by funcToSetProperty, i agree that returning bool may be clearer

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -248,115 +276,47 @@ bool PartHydro::LoadFromFolder(Study& study, const AnyString& folder)
...
 
-    const char* const sectionName = "inter-daily-breakdown";
-
     IniFile::Section* section;
-    IniFile::Property* property;
 
-    if ((section = ini.find(sectionName)))
+    if ((section = ini.find("inter-daily-breakdown")))
     {
-        if ((property = section->firstProperty))
```
a-zakir, line 258: (property = section->firstProperty) initialization is missing on new code

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
payetvin: if (!errorPower  avoided going into this scope after the first error, I simplified while keeping the same behavior

src/libs/antares/study/include/antares/study/parts/hydro/prepro.h
payetvin: will be done in another PR

src/libs/antares/study/parts/hydro/container.cpp
payetvin: `loadProperties(study, section->firstProperty,`
given to the function to simplify

src/libs/antares/study/parts/hydro/hydromaxtimeseriesreader.cpp
flomnes: No you don't keep the same behavior. The user will only see the 1st error, not the following ones if any.

flomnes (changes requested):
That's a good start @payetvin ! Please revert to the former behavior and we're good

flomnes (changes requested):
Missing leewayDown / leewayUp comparison check.

flomnes (approved):

PR #2155, opened 2024-06-11
Remove TS links from solver

guilpier-code:
The aim of this PR is removing from **Antares solver** the code related to TS generation associated to **links** capacity.
It follows PR #1986 

Note that : 
- Loading the study is now no longer required to generate TS associated to links.
- Handling errors related to each link's TS generation was simplified : 
  For a link, if we meet a problem while loading data required to generated its TS, a **warning** is raised and the TS generation for this link will be skipped.

What was done : 
- Reading data for links TS generation :  
   We now don't need the study to be loaded in order to generate TS. 
   Some new code retrieves data from study
- Generating TS associated to links : 
  Code of base branch was adapted to use data having a different format.
- Code related to link TS generation in **solver** was removed.
  An exception though : due to new data (related to links TS generation) next to data consumed by solver, we had to keep special skipping code for it, otherwise we get reading errors.

### DIFF
--- a/src/libs/antares/study/area/links.cpp
+++ b/src/libs/antares/study/area/links.cpp
@@ -306,14 +306,12 @@ AreaLink* AreaAddLinkBetweenAreas(Area* area, Area* with, bool warning)
 namespace // anonymous
 {
 
-bool isPropertyUsedForLinkTSgeneration(const std::string key)
+bool isPropertyUsedForLinkTSgeneration(const std::string& key)
 {
     std::array<std::string, 7> listKeys
         = {"unitcount", "nominalcapacity", "law.planned", "law.forced",
            "volatility.planned", "volatility.forced", "force-no-generation"};
-    if (std::find(listKeys.begin(), listKeys.end(), key) != listKeys.end())
-        return true;
-    return false;
+    return std::find(listKeys.begin(), listKeys.end(), key) != listKeys.end();
 }
 
 bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const String& value)
@@ -459,13 +457,9 @@ bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const 
         link.filterYearByYear = stringIntoDatePrecision(value);
         return true;
     }
-    if (isPropertyUsedForLinkTSgeneration(key.to<std::string>()))
-    {
-        // Properties used by TS generator only.
-        // We just skip them (otherwise : reading error)
-        return true;
-    }
-    return false;
+    // Properties used by TS generator only.
+    // We just skip them (otherwise : reading error)
+    return isPropertyUsedForLinkTSgeneration(key.to<std::string>());
 }
 
 [[noreturn]] void logLinkDataCheckError(const AreaLink& link, const String& msg, int hour)
@@ -508,7 +502,7 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* areaList, Area* area, const
     for (auto* s = ini.firstSection; s; s = s->next)
     {
         // Getting the name of the area
-        std::string targetAreaName = transformNameIntoID(s->name);
+        const std::string targetAreaName = transformNameIntoID(s->name);
 
         // Trying to find it
         Area* targetArea = AreaListLFind(areaList, targetAreaName.c_str());
--- a/src/libs/antares/study/area/links.cpp
+++ b/src/libs/antares/study/area/links.cpp
@@ -510,7 +510,6 @@ bool handleTSGenKey(Data::LinkTsGeneration& out,
                     const std::string& key,
                     const String& value)
 {
-
     if (key == "unitcount")
     {
         return value.to<uint>(out.unitCount);
--- a/src/libs/antares/study/area/links.cpp
+++ b/src/libs/antares/study/area/links.cpp
@@ -23,10 +23,7 @@
 
 #include <filesystem>
 #include <limits>
-
-#include <boost/algorithm/string/case_conv.hpp>
-
-#include <yuni/yuni.h>
+#include <array>
 
 #include <antares/exception/LoadingError.hpp>
 #include <antares/logs/logs.h>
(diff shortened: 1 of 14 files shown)


src/libs/antares/study/area/links.cpp
```diff
@@ -360,7 +305,18 @@ AreaLink* AreaAddLinkBetweenAreas(Area* area, Area* with, bool warning)
 
 namespace // anonymous
 {
-bool handleKey(Data::AreaLink& link, const String& key, const String& value)
+
+bool isPropertyUsedForLinkTSgeneration(const std::string key)
```
flomnes:
```suggestion
bool isPropertyUsedForLinkTSgeneration(const std::string& key)
```

src/libs/antares/study/area/links.cpp
```diff
@@ -503,57 +459,15 @@ bool handleKey(Data::AreaLink& link, const String& key, const String& value)
...
-        return value.to<uint>(out.unitCount);
-    }
-
-    if (key == "nominalcapacity")
-    {
-        return value.to<double>(out.nominalCapacity);
-    }
-
-    if (key == "law.planned")
+    if (isPropertyUsedForLinkTSgeneration(key.to<std::string>()))
```
flomnes:
```suggestion
    return isPropertyUsedForLinkTSgeneration(key.to<std::string>());
```

src/libs/antares/study/area/links.cpp
```diff
@@ -360,7 +305,18 @@ AreaLink* AreaAddLinkBetweenAreas(Area* area, Area* with, bool warning)
...
-bool handleKey(Data::AreaLink& link, const String& key, const String& value)
+
+bool isPropertyUsedForLinkTSgeneration(const std::string key)
+{
+    std::array<std::string, 7> listKeys
+        = {"unitcount", "nominalcapacity", "law.planned", "law.forced",
+           "volatility.planned", "volatility.forced", "force-no-generation"};
+    if (std::find(listKeys.begin(), listKeys.end(), key) != listKeys.end())
+        return true;
+    return false;
```
flomnes:
```suggestion
    return std::find(listKeys.begin(), listKeys.end(), key) != listKeys.end();
```

src/libs/antares/study/area/links.cpp
```diff
@@ -572,7 +486,7 @@ bool AreaLinksInternalLoadFromProperty(AreaLink& link, const String& key, const
 }
 } // anonymous namespace
 
-bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const fs::path& folder, bool loadTSGen)
+bool AreaLinksLoadFromFolder(Study& study, AreaList* areaList, Area* area, const fs::path& folder)
```
flomnes, line 483: 👍🏻

src/libs/antares/study/area/links.cpp
```diff
@@ -594,16 +508,16 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* l, Area* area, const fs::pa
     for (auto* s = ini.firstSection; s; s = s->next)
     {
         // Getting the name of the area
-        std::string buffer = transformNameIntoID(s->name);
+        std::string targetAreaName = transformNameIntoID(s->name);
```
flomnes:
```suggestion
        const std::string targetAreaName = transformNameIntoID(s->name);
```

src/solver/ts-generator/availability.cpp
```diff
@@ -73,8 +74,9 @@ class GeneratorTempData final
 {
 public:
     explicit GeneratorTempData(Data::Study&, unsigned, MersenneTwister&);
+    explicit GeneratorTempData(bool, unsigned, MersenneTwister&);
 
-    void generateTS(const Data::Area& area, AvailabilityTSGeneratorData& cluster) const;
+    void generateTS(std::string, AvailabilityTSGeneratorData&) const;
```
flomnes:
```suggestion
    void generateTS(const std::string&, AvailabilityTSGeneratorData&) const;
```

src/libs/antares/study/area/links.cpp
guilpier-code: Done

src/libs/antares/study/area/links.cpp
guilpier-code: Done

src/libs/antares/study/area/links.cpp
guilpier-code: Of course ! Thanks. Done

src/libs/antares/study/area/links.cpp
guilpier-code: Done

src/solver/ts-generator/availability.cpp
guilpier-code: I simply removed this **std::string**  argument : 
indeed, this **source area name** is used inside function **generateTS** in case **prepro** was not properly loaded.
But the **prepro** loading itself already checked that **prepro** was loaded properly and is supposed to have raised an error otherwise.
So I don't see the point of making this check again when generating the TS.

payetvin:
A new file tools/ts-generator/links.cpp should be created to handle links

src/tools/ts-generator/main.cpp
```diff
@@ -111,28 +113,311 @@ std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
...
+{
+    std::vector<LinkTSgenerationParams> to_return(linksFromCmdLine.size());
+    // gp : following loop should be improved : we shouldn't need an index
+    unsigned int index = 0;
+    for (const auto& link_pair : linksFromCmdLine)
+    {
+        to_return[index].namesPair = link_pair;
+        index++;
     }
+    return to_return;
```
flomnes: Use `reserve` and `push_back` to avoid using indices while avoiding reallocations.

flomnes (approved):
LGTM. One last suggestion to help

PR #2160, opened 2024-06-14
Remove artificial dependency on libuuid

flomnes:


flomnes:
The GUI needs this, do it later.

PR #2167, opened 2024-06-14
Reorganize cleaner files a bit (rename cleaner-v20.cpp, etc.)

flomnes:


src/libs/antares/study/cleaner/study-cleaning-info.cpp
flomnes, line 1: Formerly cleaner-v20.cpp

PR #2171, opened 2024-06-14
Link TS generation : splitting into multiple files

guilpier-code:
This PR aims at moving some code about link TS generation, from **main** program to new source files.
It follows PR #2155 

To be done : 

- [x] move headers (*.h) to the right place
- [x] update this branch with its base branch (@flomnes made a change on base branch)

We may also take advantage of this PR to : 
- [ ] Thermal TS generation :
  - [ ] avoid loading the **whole study** when **no** thermal TS generation is required (this is currently the case)
  - [ ] move implementation details about thermal TS generation from **main** program to new source files
- [ ] Separate loading / extracting data from study and generate the TS


src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
+
+#include "linksTSgenerator.h"
+#include <yuni/datetime/timestamp.h>
+
+namespace Antares::TSGenerator
+{
+// ==================
+// Free functions
+// ==================
+std::vector<std::string> extractTargetAreas(fs::path sourceLinkDir)
```
a-zakir: const ref is overrated, is it?

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+{
+    std::vector<std::string> to_return;
+    fs::path pathToIni = sourceLinkDir / "properties.ini";
+    IniFile ini;
+    ini.open(pathToIni); // gp : we should handle reading issues
+    for (auto* s = ini.firstSection; s; s = s->next)
```
a-zakir: auto can handle pointer

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+    {
+        std::string targetAreaName = transformNameIntoID(s->name);
+        to_return.push_back(targetAreaName);
```
a-zakir:
```suggestion
        to_return.push_back(transformNameIntoID(s->name));
```

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+    }
+    if (key == "seed-tsgen-links")
+    {
+        unsigned int seed {0};
+        if (! value.to<unsigned int>(seed))
+            return false;
+        params.random.reset(seed);
+        return true;
+    }
+    return true; // gp : should we return true here ?
```
a-zakir: indeed, you could log that the key is unknown and therefore skipped

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+std::vector<LinkTSgenerationParams> CreateLinkList(const LinkPairs& linksFromCmdLine)
+{
+    std::vector<LinkTSgenerationParams> to_return(linksFromCmdLine.size());
+    // gp : following loop should be improved : we shouldn't need an index
+    unsigned int index = 0;
+    for (const auto& link_pair : linksFromCmdLine)
+    {
+        to_return[index].namesPair = link_pair;
+        index++;
+    }
```
a-zakir:
```suggestion
    std::transform(linksFromCmdLine.begin(), linksFromCmdLine.end(), std::back_inserter(to_return), [](const LinkPair& link_pair){
    LinkTSgenerationParams ret;
    ret.namesPair = link_pair;
    return ret;
    } );
```

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+            return false;
+        params.random.reset(seed);
+        return true;
+    }
+    return true; // gp : should we return true here ?
+}
+
+std::vector<LinkTSgenerationParams> CreateLinkList(const LinkPairs& linksFromCmdLine)
+{
+    std::vector<LinkTSgenerationParams> to_return(linksFromCmdLine.size());
```
a-zakir:
```suggestion
    std::vector<LinkTSgenerationParams> to_return;
```

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+    // gp : following loop should be improved : we shouldn't need an index
+    unsigned int index = 0;
```
a-zakir:
```suggestion

```

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+    }
+    return to_return;
+}
+
+LinkTSgenerationParams* findLinkInList(const LinkPair& link_to_find,
+                                       std::vector<LinkTSgenerationParams>& linkList)
+{
+    for(auto& link : linkList)
+    {
+        if (link.namesPair == link_to_find)
```
a-zakir, line 86: why don't you re-use pairs_match?

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+        fs::path sourceAreaDir = toLinksDir / sourceAreaName;
+        if (! readLinkPreproTimeSeries(link, sourceAreaDir))
+        {
+            logs.warning() << "Could not load all prepro data for link '"
+                           << link.namesPair.first << "." << link.namesPair.second << "'";
+        }
+    }
+}
+
+std::string DateAndTime()
```
a-zakir: there is a function, free from Yuni in [application.cpp](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/bce0fa7d3c62bdec6a92aa824d147ddfaec1ffbc/src/solver/application/application.cpp#L429)

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -0,0 +1,346 @@
...
+    for (auto const& item : fs::directory_iterator{linksDir})
+    {
+        if (item.is_directory())
+        {
+            std::string sourceAreaName = item.path().filename().generic_string();
+            auto targetAreas = extractTargetAreas(item);
+            for (auto& targetAreaName : targetAreas)
+            {
+                auto linkPair = std::make_pair(sourceAreaName, targetAreaName);
+                to_return.push_back(linkPair);
```
a-zakir, line 279: std::transform?

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: I did this, because it makes code shorter, but intermediate variable **targetAreaName** had the advantage to be very clear.

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: I got inspired by your suggestion and eventually used **std::for_each** instead of **std::transform**.
Indeed, in my opinion, **std::for_each** is a bit more adapted to this particular situation, but the idea is the same.

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: I think it's clear and short enough as it is.

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: Done

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: Done

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: Done

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: Done

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: Done. Use of **index** disappeared.

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: Taken care of.

src/tools/ts-generator/linksTSgenerator.cpp
guilpier-code: What you suggest is not what we want : we're reading a **generaldata.ini** file, which (as you know) contains many properties. 
But here, we want to extract some of them. 
If we return **false** as soon as we meet a property we don't want to read, a lot of warnings will be raised.

src/solver/application/application.cpp
```diff
@@ -451,7 +437,7 @@ void Application::resetLogFilename() const
     }
 
     logfile /= "solver-";               // append the filename
-    logfile += timeToString() + ".log"; // complete filename with timestamp and extension
+    logfile += FormattedTime("%Y%m%d-%H%M%S") + ".log"; // complete filename with timestamp and extension
```
a-zakir: i'm surprised, how is this works? fs::path+std::string 😮

src/tools/ts-generator/linksTSgenerator.cpp
a-zakir: ok

src/solver/application/application.cpp
guilpier-code: `FormattedTime("%Y%m%d-%H%M%S")` is a **std::string**.
So : `FormattedTime("%Y%m%d-%H%M%S") + ".log"` is a **std::string** as well.
Then we use operator **+=** of class **fs::path**, which takes a **std::string** as argument, appends this string to the path.

a-zakir: thanks, i didn't notice the += ope

a-zakir (approved):

src/libs/antares/utils/utils.cpp
```diff
@@ -96,6 +96,20 @@ void BeautifyName(std::string& out, const std::string& oldname)
...
+    using namespace std::chrono;
+    auto time = system_clock::to_time_t(system_clock::now());
+    std::tm local_time = *std::localtime(&time);
+
+    char time_buffer[256];
+    std::strftime(time_buffer, sizeof(time_buffer), format.c_str(), &local_time);
+
+    std::string currentTime = time_buffer;
+
+    return currentTime;
```
flomnes:
```suggestion
    return std::string(time_buffer);
```

src/tools/ts-generator/linksTSgenerator.cpp
flomnes: Good point ! `pairs_match` is a bit less strict than `pair<>::operator==`, since it returns `true` if `a.first==b.second && a.second == b.first` (not the case with `pair<>::operator==`)

flomnes (approved):

PR #2173, opened 2024-06-17
Thermal validation and loading [ANT-1213]

payetvin:
Modular loading: separating loading and validation functions

### DIFF
--- a/src/libs/antares/study/parts/thermal/cluster_list.cpp
+++ b/src/libs/antares/study/parts/thermal/cluster_list.cpp
@@ -605,12 +605,22 @@ bool ThermalClusterList::loadPreproFromFolder(Study& study, const AnyString& fol
     Clob buffer;
     auto hasPrepro = [](auto c) { return (bool)c->prepro; };
 
-    auto loadAndCheckPrepro = [&buffer, &folder, &study, &globalThermalTSgeneration](auto c)
+    auto loadPrepro = [&buffer, &folder, &study, &globalThermalTSgeneration](auto& c)
     {
         assert(c->parentArea && "cluster: invalid parent area");
         buffer.clear() << folder << SEP << c->parentArea->id << SEP << c->id();
 
-        bool result = c->prepro->loadFromFolder(study, buffer);
+        return c->prepro->loadFromFolder(study, buffer);
+    };
+
+    return std::ranges::all_of(allClusters_ | std::views::filter(hasPrepro), loadPrepro);
+}
+
+bool ThermalClusterList::validatePrepro()
+{
+    return std::ranges::all_of(allClusters_ | std::views::filter(hasPrepro), [](auto& c)
+    {
+        bool result = true;
 
         if (study.usedByTheSolver && globalThermalTSgeneration)
         {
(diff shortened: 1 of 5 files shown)


src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -159,52 +159,57 @@ bool ThermalClusterList::loadFromFolder(Study& study, const AnyString& folder, A
...
+
+
+bool ThermalClusterList::validateClusters(const Parameters& parameters) const
+{
+    bool ret = true;
+
+    for (const auto& cluster : allClusters_)
+    {
+        // update the minUpDownTime
+        cluster->minUpDownTime = std::max(cluster->minUpTime, cluster->minDownTime);
```
JasonMarechal25, line 188: Need a bit of clarification, maybe more details in the comment. It's strange to see a std::max on a minX value.
Let's add information on why we're doing this

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -606,18 +585,31 @@ bool ThermalClusterList::saveEconomicCosts(const AnyString& folder) const
...
+{
+    auto hasPrepro = [](auto c) { return (bool)c->prepro; };
+
+    const bool globalThermalTSgeneration = study.parameters.timeSeriesToGenerate
+                                           & timeSeriesThermal;
+
+    return std::ranges::all_of(allClusters_ | std::views::filter(hasPrepro),
+            [&study, &globalThermalTSgeneration](auto& c)
+    {
+        bool result = true;
```
JasonMarechal25: if NOT (study.usedByTheSolver && globalThermalTSgeneration)  that is to say if !study.usedByTheSolver or !globalThermalTSgeneration result will be true. Moreover it is constant for all element.
It seems that we can return true before the computation of all_of if either value is false.

JasonMarechal25 (changes requested):

src/libs/antares/study/parts/thermal/cluster_list.cpp
payetvin:
```
    //! Max entre . minUp/minDown time (1..168)
    uint minUpDownTime;
```
The naming isn't great, can be changed to maxMinUpMinDownTime

src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -159,51 +159,69 @@ bool ThermalClusterList::loadFromFolder(Study& study, const AnyString& folder, A
...
+}
+
+
+bool ThermalClusterList::validateClusters(const Parameters& parameters) const
+{
+    bool ret = true;
+
+    for (const auto& cluster : allClusters_)
+    {
+        if (cluster->minUpTime < 1)
```
JasonMarechal25: Please replace with: std::clamp(cluster->minUpTime, 1, 168)
And replace 1 and 168 with named constant if possible.

JasonMarechal25 (changes requested):

src/libs/antares/study/parts/thermal/cluster_list.cpp
JasonMarechal25: It seems that some value "underflow". Some value stored as unint fall under 0, going back to UINT_MAX. When clamped they are clamped to 168.
At some point somewhere else we wrongly convert some value or don't check bounds. But it is outside of this PR to solve this.

JasonMarechal25 (approved):

PR #2179, opened 2024-06-20
Separation of loading and validation for header [ANT-1213] 

payetvin:


src/libs/antares/study/include/antares/study/header.h
```diff
@@ -109,7 +108,8 @@ class StudyHeader final
      */
     static StudyVersion tryToFindTheVersion(const std::string& folder);
 
-public:
+    bool validateVersion();
```
flomnes, line 111:
```suggestion
    bool validate() const;
```

src/libs/antares/study/header.cpp
```diff
@@ -166,28 +166,41 @@ bool StudyHeader::internalLoadFromINIFile(const IniFile& ini, bool warnings)
...
+        }
+    }
+    else
+    {
+        logs.error() << "Version before 7.0 not supported, please upgrade";
+        return false;
+    }
+
+    return true;
+}
```
flomnes: Move to `StudyVersion` ? So that you can just write `version.validate()`

JasonMarechal25 (approved):

a-zakir (approved):

PR #2183, opened 2024-06-21
Fix initialization order warnings

payetvin:


flomnes (changes requested):
As discussed, use constructor delegation to avoid code duplication

flomnes (approved):

PR #2189, opened 2024-06-25
Resize data collections in PROBLEME_A_RESOUDRE

JasonMarechal25:
Some collections in PROBLEME_A_RESOUDRE are optimistically sized early but never resized to their proper size. When exporting those collections to clients by API it unnecessarily.

Resizing for NombreDeVariables could be done earlier but for NombreDeContraintes we need to wait until we have build all constraints with ConstraintBuilder.

ConstraintsMatrixCoeff and ColumnIndexes collections are initially oversized (By a large factor).
Before exporting data through API we properly dimension them and remove extra 0 padding


src/solver/optimisation/opt_construction_matrice_des_contraintes_outils.cpp
```diff
@@ -30,8 +30,8 @@ void OPT_ChargerLaContrainteDansLaMatriceDesContraintes(
   int NombreDeTermesDeLaContrainte,
   char SensContrainte)
 {
-    int& nombreDeTermes = ProblemeAResoudre->NombreDeTermesDansLaMatriceDesContraintes;
```
flomnes: Does not using a reference here make any difference ?

src/solver/optimisation/opt_optimisation_lineaire.cpp
```diff
@@ -139,6 +139,34 @@ void runThermalHeuristic(PROBLEME_HEBDO* problemeHebdo)
...
+                    unsigned nombreDeContraintes)
+{
+    ProblemeAResoudre->CoutQuadratique.resize(nombreDeVariables);
+    ProblemeAResoudre->CoutLineaire.resize(nombreDeVariables);
+    ProblemeAResoudre->TypeDeVariable.resize(nombreDeVariables);
+    ProblemeAResoudre->Xmin.resize(nombreDeVariables);
+    ProblemeAResoudre->Xmax.resize(nombreDeVariables);
+    ProblemeAResoudre->X.resize(nombreDeVariables);
+    ProblemeAResoudre->AdresseOuPlacerLaValeurDesVariablesOptimisees.resize(nombreDeVariables);
+    ProblemeAResoudre->AdresseOuPlacerLaValeurDesCoutsReduits.resize(nombreDeVariables);
```
flomnes, line 175: nombreDeContraintes, to be checked

src/solver/optimisation/opt_optimisation_lineaire.cpp
```diff
@@ -139,6 +139,34 @@ void runThermalHeuristic(PROBLEME_HEBDO* problemeHebdo)
...
+    ProblemeAResoudre->PositionDeLaVariable.resize(nombreDeVariables);
+    ProblemeAResoudre->Pi.resize(nombreDeVariables);
```
flomnes: `nbTermes` for `Pi`

src/solver/optimisation/opt_optimisation_lineaire.cpp
```diff
@@ -139,6 +139,34 @@ void runThermalHeuristic(PROBLEME_HEBDO* problemeHebdo)
...
+    ProblemeAResoudre->Colonne.resize(nombreDeVariables);
```
flomnes: `nbTermes` for `Colonne`

src/solver/optimisation/opt_construction_matrice_des_contraintes_outils.cpp
JasonMarechal25: Yes. I'm reafecting the result at the end of the scope.
The idea is to have a clear affectation of the variables. With a reference it's kinda hidden

a-zakir: forget it! this file is deleted in develop

src/solver/optimisation/opt_optimisation_lineaire.cpp
JasonMarechal25: Pi is initially sized with nombreDeVariables in OPT_AllocateFromNumberOfVariableConstraints.
Why resize it to nbThermes ?

src/solver/optimisation/opt_optimisation_lineaire.cpp
JasonMarechal25: Colonne is initially sized with nombreDeVariables in OPT_AllocateFromNumberOfVariableConstraints.
Why resize it to nbThermes ?

src/solver/optimisation/opt_optimisation_lineaire.cpp
flomnes: `Pi` and `Colonne` are just temporary variables used to fill the constraint matrix. They should not be used after this step, so there is no need to resize them IMO.

src/solver/optimisation/opt_optimisation_lineaire.cpp
flomnes: See answer above

flomnes (approved):

PR #2202, opened 2024-06-26
TS generator : refactor thermal part

guilpier-code:
Aim of this PR : see the title.

What was done (see helping **gitHub** comments in this PR) : 
- simplifications : 
  - [x] Standardizing TS generation with link TS : remove a constructor in a class containing data for TS generation.
  - [x] Remove use of **RuntimeInfos** in **main.cpp** (many code lines used to be called there, just to get a random generator).
  - [x] Remove check on min stable power, useless.
  - [x] Function that actually generates the TS no longer takes the TS to be generated as argument, but returns it.
  - [x] **TS generator** code no longer contains code related to whether we should skip the TS writing on disk. It writes, that'all. This kind of code belongs to **solver**, not to **TS generator**.
  - [x] Use **std::filesystem** whenever possible instead of **Yuni**
  
- Unwanted behavior correction : 
  - [x] Execute links or thermal TS generation code only depending on user requirements : indeed, thermal TS generation still needs the whole study to be loaded, but link TS generation does not (it retrieves from study only the required data). Being forced to load the study if only link TS generation is required is a clear performance issue. We avoid that here.
  - [x] When data files required to generate TS (modulation files & prepro files) don't exist, we used to ignore it (this can lead to a crash). We now raise an error before TS generation time comes. 
- Moving code : _to be completed_

### DIFF
--- a/src/solver/simulation/include/antares/solver/simulation/solver.hxx
+++ b/src/solver/simulation/include/antares/solver/simulation/solver.hxx
@@ -464,7 +464,7 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
 	    namespace fs = std::filesystem;
             fs::path savePath = fs::path(study.folderOutput.to<std::string>()) / "ts-generator"
                                 / "thermal" / "mc-" / std::to_string(year);
-            generateThermalTimeSeries(study, clusters, savePath.string());
+            generateThermalTimeSeries(study, clusters, savePath);
 
             // apply the spinning if we generated some in memory clusters
             for (auto* cluster: clusters)
--- a/src/solver/simulation/include/antares/solver/simulation/solver.hxx
+++ b/src/solver/simulation/include/antares/solver/simulation/solver.hxx
@@ -464,7 +464,7 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
 	    namespace fs = std::filesystem;
             fs::path savePath = fs::path(study.folderOutput.to<std::string>()) / "ts-generator"
                                 / "thermal" / "mc-" / std::to_string(year);
-            generateThermalTimeSeries(study, clusters, savePath);
+            generateThermalTimeSeries(study, clusters, study.runtime->random[Data::seedTsGenThermal], savePath);
 
             // apply the spinning if we generated some in memory clusters
             for (auto* cluster: clusters)
--- a/src/solver/simulation/include/antares/solver/simulation/solver.hxx
+++ b/src/solver/simulation/include/antares/solver/simulation/solver.hxx
@@ -461,10 +461,16 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
         if (refreshTSonCurrentYear)
         {
             auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
-	    namespace fs = std::filesystem;
-            fs::path savePath = fs::path(study.folderOutput.to<std::string>()) / "ts-generator"
-                                / "thermal" / "mc-" / std::to_string(year);
-            generateThermalTimeSeries(study, clusters, study.runtime->random[Data::seedTsGenThermal], savePath);
+            generateThermalTimeSeries(study, clusters, study.runtime->random[Data::seedTsGenThermal]);
+
+            bool archive = study.parameters.timeSeriesToArchive & Data::timeSeriesThermal;
+            bool doWeWrite = archive && !study.parameters.noOutput;
+            if (doWeWrite)
+            {
+                fs::path savePath = fs::path(study.folderOutput.to<std::string>()) / "ts-generator"
+                                    / "thermal" / "mc-" / std::to_string(year);
+                writeThermalTimeSeries(clusters, savePath);
+            }
 
             // apply the spinning if we generated some in memory clusters
             for (auto* cluster: clusters)
--- a/src/solver/ts-generator/availability.cpp
+++ b/src/solver/ts-generator/availability.cpp
@@ -612,8 +612,7 @@ void writeTStoDisk(const Matrix<>& series,
 
 bool generateThermalTimeSeries(Data::Study& study,
                                const std::vector<Data::ThermalCluster*>& clusters,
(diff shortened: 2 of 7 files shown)


src/tools/ts-generator/main.cpp
```diff
@@ -80,34 +75,15 @@ int main(int argc, char* argv[])
...
     Data::StudyLoadOptions studyOptions;
-    studyOptions.prepareOutput = true;
-
     if (!study->loadFromFolder(settings.studyFolder, studyOptions))
     {
         logs.error() << "Invalid study given to the generator";
         return 1;
     }
 
-    study->initializeRuntimeInfos();
```
guilpier-code, line 94: `study->initializeRuntimeInfos();` : disappeared.
We no longer need to manipulate the **RunTimeInfos** stuff only to reach the random generator.
A **Mersenne twister** instance is now defined and used explicitly a bit further in this file (see below).

src/tools/ts-generator/main.cpp
```diff
@@ -80,34 +75,15 @@ int main(int argc, char* argv[])
...
-    // Force the writing of generated TS into output/YYYYMMDD-HHSSeco/ts-generator/thermal[/mc-0]
-    study->parameters.timeSeriesToArchive |= Antares::Data::timeSeriesThermal;
-
-    try
-    {
-        Antares::Check::checkMinStablePower(true, study->areas);
```
guilpier-code, line 100: `Antares::Check::checkMinStablePower(true, study->areas);` : disappeared.
Indeed, we don't need to check min stable power to generate a thermal TS : min stable power is never used to build a thermal TS.

src/tools/ts-generator/main.cpp
```diff
@@ -80,34 +75,15 @@ int main(int argc, char* argv[])
     if (! checkOptions(settings))
         return 1;
 
+    // ============ THERMAL : Getting data for generating time-series =========
     auto study = std::make_shared<Data::Study>(true);
     Data::StudyLoadOptions studyOptions;
-    studyOptions.prepareOutput = true;
```
guilpier-code, line 86: `studyOptions.prepareOutput = true;` : disappeared.
In the context of running the **solver**, **prepareOutput** is set as a command line option, and is used to allow / disable / restrain simulation output results.
In the context of running the TS generator, the user wants to generate and print TS. 
He/she doesn't want to know about **prepareOutput**.

src/tools/ts-generator/linksTSgenerator.cpp
```diff
@@ -166,52 +166,56 @@ void readIniProperties(std::vector<LinkTSgenerationParams>& linkList, fs::path t
...
+bool readLinkPreproTimeSeries(LinkTSgenerationParams& link,
+                              fs::path sourceAreaDir)
+{
+    bool to_return = true;
+    const auto preproId = link.namesPair.first + "/" + link.namesPair.second;
+    link.prepro = std::make_unique<Data::PreproAvailability>(preproId, link.unitCount);
+
+    auto preproFileRoot = sourceAreaDir / "prepro" / link.namesPair.second;
+
+    // Testing files existence
```
guilpier-code, line 185: `// Testing files existence` : 
This file existence test was added because it used to exist in another form in this function, but disappeared by mistake.
So we reintroduce it under this new form.

src/solver/ts-generator/availability.cpp
```diff
@@ -198,13 +178,16 @@ int AvailabilityTSgenerator::durationGenerator(Data::StatisticalLaw law,
     return 0;
 }
 
-void AvailabilityTSgenerator::run(AvailabilityTSGeneratorData& tsGenerationData) const
+Matrix<double> AvailabilityTSgenerator::run(AvailabilityTSGeneratorData& tsGenerationData) const
```
guilpier-code, line 181: **AvailabilityTSgenerator::run(...)** now returns the generated TS, and no longer take the generated TS as argument.
It's more natural and clear like this, and allows to simplify it's calls (for example, see **generateLinkTimeSeries(...)** below)

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -489,9 +489,16 @@ void ISimulation<ImplementationType>::regenerateTimeSeries(uint year)
         if (refreshTSonCurrentYear)
         {
             auto clusters = getAllClustersToGen(study.areas, pData.haveToRefreshTSThermal);
-            fs::path savePath = fs::path(study.folderOutput.to<std::string>()) / "ts-generator"
-                                / "thermal" / "mc-" / std::to_string(year);
-            generateThermalTimeSeries(study, clusters, savePath.string());
+            generateThermalTimeSeries(study, clusters, study.runtime->random[Data::seedTsGenThermal]);
+
+            bool archive = study.parameters.timeSeriesToArchive & Data::timeSeriesThermal;
```
guilpier-code, line 466: Details about if we print the TS or not were moved here (to the solver). 
Indeed, these details do not belong to the TS generation functions (as they used to) : TS generator executable generates and prints, and does not want to know if study data requires printing TS or not.

src/tools/ts-generator/main.cpp
```diff
@@ -44,23 +39,23 @@ std::vector<Data::ThermalCluster*> getClustersToGen(Data::AreaList& areas,
...
+    const auto pairsAreaCluster = splitStringIntoPairs(clustersToGen, ';', '.');
 
-    for (const auto& [areaID, clusterID]: ids)
+    for (const auto& [areaName, clusterName]: pairsAreaCluster)
     {
-        logs.info() << "Searching for area: " << areaID << " and cluster: " << clusterID;
+        logs.info() << "Searching for area: " << areaName << " and cluster: " << clusterName;
 
-        auto* area = areas.find(areaID);
+        auto* area = areas.find(areaName);
```
payetvin: `areas.find()` and `thermal.list.findInAll()` takes IDs, which is not the same as names (some char are prohibited in IDs). That's why the list given to the exe is list of IDs.

Revert to old name to explicit the difference

flomnes (approved):

PR #2203, opened 2024-06-27
Harmonize time constants

### LINKED ISSUE #2188: Constants in solver/variable
`src/solver/variable/include/antares/solver/variable/constants.h`

This file contains constants for hours in a year that doesn't match our other constants like `HOUR_PER_YEAR`
We should harmonize this code

payetvin:
close #2188


flomnes (dismissed):

src/libs/antares/date/date.cpp
```diff
@@ -137,12 +137,12 @@ bool StringToMonth(MonthName& out, AnyString text)
         return false;
     }
     text.trim();
-    CString<12, false> t = text;
+    CString<MONTHS_PER_YEAR, false> t = text;
```
flomnes: Revert or use `auto`

flomnes (approved):
One comment and we're good

PR #2204, opened 2024-06-27
Collect hydro validation errors

a-zakir:
collect *n* hydro validation errors (10 per area ) before exiting


src/solver/hydro/management/HydroErrorsCollector.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+void HydroErrorsCollector::IncreaseCounterForArea(const Antares::Data::Area* area)
+{
+    if (errors_limit_reached_ || !area)
+    {
+        return;
+    }
+
+    area_errors_counter_[area]++;
+    errors_limit_reached_ = area_errors_counter_[area] > 10 || errors_limit_reached_;
```
a-zakir: #TODO 
throw exception here?

src/solver/hydro/include/antares/solver/hydro/management/HydroErrorsCollector.h
```diff
@@ -0,0 +1,43 @@
...
+
+class HydroErrorsCollector
+{
+public:
+    HydroErrorsCollector() = default;
+    void IncreaseCounterForArea(const Antares::Data::Area* area);
+    bool StopExecution() const;
+    bool ErrorsLimitReached() const;
+    void FatalErrorHit();
+    // void RecordFatalErrors(const std::string& msg, uint year);
```
flomnes: That a bit too complicated for me

We should just handle limits internally, the caller should see this

```cpp
collector(area, "Error1");
collector(area, "Error2");
collector.toLog(logs.error);
```

a-zakir: ok nice

src/solver/hydro/management/HydroErrorsCollector.cpp
a-zakir: done

src/solver/hydro/management/HydroErrorsCollector.cpp
```diff
@@ -0,0 +1,36 @@
...
+{
+    logs.error() << "In Area " << area_name << " " << message;
+    area_errors_counter_[area_name]++;
+    errors_limit_reached_ = area_errors_counter_[area_name] > 10;
+    stop_ = true;
+
+    if (errors_limit_reached_)
+    {
+        throw FatalError("Hydro validation has failed !");
+    }
```
flomnes: I think it would be best to have this behavior instead
```cpp
if (!errors_limit_reached_) {
    logs.error() << "...";
}
```
There's no need to `throw`, let higher level functions handle that. Otherwise the responsability of `HydroErrorsCollector` becomes unclear, we might even have cases where 9 errors are considered valid, but 11 are considered invalid, etc.

src/solver/hydro/management/HydroInputsChecker.cpp
```diff
@@ -50,27 +46,24 @@ void HydroInputsChecker::Execute(uint year)
 {
     prepareInflows_.Run(year);
     minGenerationScaling_.Run(year);
-
     if (!checksOnGenerationPowerBounds(year))
     {
-        throw FatalError("hydro inputs checks: invalid minimum generation");
+        std::ostringstream msg;
+        msg << "hydro inputs checks: invalid minimum generation in year " << year;
+        errorCollector_.Collect(msg.str());
```
```suggestion
        errorCollector_ << "hydro inputs checks: invalid minimum generation in year " << year;
```
Who doesn't like a bit of syntactic sugar ?

src/solver/hydro/management/HydroErrorsCollector.cpp
```diff
@@ -0,0 +1,36 @@
...
+    logs.error() << message;
+    stop_ = true;
+}
+
+void HydroErrorsCollector::CheckForFatalErrors() const
+{
+    if (stop_ || errors_limit_reached_)
+    {
+        throw FatalError("Hydro validation has failed !");
+    }
```
flomnes: We should `throw` if any error have been encoutered, not depending on a threshold.

flomnes (changes requested):
- Minor suggestion for improvement using `operator<<` overloading
- Major suggestion to remove throws, use threshold only to filter errors (the initial goal is to avoid displaying too many error messages)

Almost good !

src/solver/hydro/management/HydroErrorsCollector.cpp
a-zakir: this method is called after all hydro checks (case max {errors for all area} <10)

src/solver/hydro/include/antares/solver/hydro/management/HydroErrorsCollector.h
```diff
@@ -0,0 +1,57 @@
...
+{
+public:
+    class AreaReference
+    {
+    public:
+        AreaReference(HydroErrorsCollector* collector, const std::string& name);
+        friend void operator<<(const AreaReference& ref, const std::string& msg);
+
+        ~AreaReference()
+        {
```
flomnes: Remove this (only useful for debugging)

src/solver/hydro/management/HydroErrorsCollector.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <yuni/core/logs.h>
+
+#include "antares/antares/fatal-error.h"
+
+namespace Antares
+{
+
+void HydroErrorsCollector::CheckForErrors() const
+{
+    if (!aresErrorMap_.empty())
```
flomnes:
```suggestion
    if (!areasErrorMap_.empty())
```

src/solver/hydro/management/HydroInputsChecker.cpp
```diff
@@ -205,4 +197,43 @@ bool HydroInputsChecker::checkGenerationPowerConsistency(uint year) const
...
+                                                         area.name,
+                                                         initialLevel,
+                                                         finalLevel,
+                                                         year,
+                                                         parameters_.simulationDays.end,
+                                                         parameters_.firstMonthInYear,
+                                                         errorCollector_);
+          if (!validator.check())
+          {
+              logs.error() << "hydro final level : infeasibility";
```
```suggestion
              errorCollector_(area.name) << "hydro final level : infeasibility";
```

flomnes (approved):

PR #2222, opened 2024-07-02
Fix infeasible analyzer

guilpier-code:
This PR is attached to ticket [ticket ANT-1825](https://gopro-tickets.rte-france.com/browse/ANT-1825).
Some improvements were made or tried by taking care of the ticket.
The result is this PR.

**What was done** : 
- hydro scope :
  - [x] move the code that prints the TS numbers on disk :  due to recent changes it was moved after the loop through MC years. So, we have to wait until simulation ends to know which TS numbers where drawn. So we move that code back.
- infeasibility analyzer scope 
  - [x] **Constraint :: extractItems()** was simplified
  - [x] Some renaming, meant to be more expressive
  - [x] Add a type of constraint to be spotted by the analyzer in case of infeasibility    
  - [ ] what else ?

### DIFF
--- a/src/solver/infeasible-problem-analysis/constraint.cpp
+++ b/src/solver/infeasible-problem-analysis/constraint.cpp
@@ -34,21 +34,21 @@ const std::string kUnknown = "<unknown>";
 
 namespace Antares::Optimization
 {
-Constraint::Constraint(const std::string& input, const double slackValue):
-    mInput(input),
-    mSlackValue(slackValue)
+Constraint::Constraint(const std::string& name, const double slackValue):
+    name_(name),
+    slackValue_(slackValue)
 {
 }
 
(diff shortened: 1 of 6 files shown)


flomnes (changes requested):
Looks good to me. However, could you please split this into 3 PRs ?
- (A) Renaming
- (B) Use `boost::algorithm::split_regex` 
- (C) Move the write of ts-numbers

A possibility is 
```
(develop) <----- C
          \-----  A <----- B
```

But at your convenience.

guilpier-code:
Was divided into 4 pull requests : 
- #2225 
- #2226 
- #2227
- #2228

guilpier-code:
Was divided into 4 pull requests : 
- #2225 
- #2226 
- #2227
- #2228

PR #2231, opened 2024-07-03
Infeasibility more cleaning

guilpier-code:
We're on the road to make **infeasibility analyzer** more changeable, more precisely when adding a constraint type to the list of constraints to be detected in case of infeasibity.    
This PR is a step towards this purpose.  
It contains some heterogeneous simplifications and renaming.
In order to ease the review, some comments were added in this PR.


src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -35,16 +36,15 @@ const std::string kUnknown = "<unknown>";
...
+    name_(name),
     slackValue_(slackValue)
 {
 }
 
-std::size_t Constraint::extractComponentsFromName()
+void Constraint::extractComponentsFromName()
 {
     boost::algorithm::split_regex(nameComponents_, name_, boost::regex("::"));
-    return nameComponents_.size();
```
guilpier-code, line 47: No need to return the size : this was returned only to test if this size is zero, but it can never be zero, as constraints are filtered based on regex pattern : 
```C++
const std::string constraint_name_pattern = "^AreaHydroLevel::|::hourly::|::daily::|::weekly::|"
                                                "^FictiveLoads::|^Level::|"
                                                "^HydroPower::";
```
This forces the constraint's name to contain "**::**" and something else (like **AreaHydroLevel**, **daily**, ...).

src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -61,67 +61,36 @@ class StringIsNotWellFormated: public std::runtime_error
     }
 };
 
-std::string StringBetweenAngleBrackets(const std::string& str)
+std::string StringBetweenAngleBrackets(const std::string& constraintName)
 {
-    const auto& begin = str.begin();
-    const auto& end = str.end();
+    std::vector<std::string> split_name;
```
guilpier-code, line 66: Here we base the extraction of the string between angle bracket on the **split algorithm**, rather than the **find algorithm**. It makes code slightly more simple.
Error handling was a bit simplified.
In the end, function is shorter and easier to understand, I guess.

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/constraint.h
```diff
@@ -40,27 +40,23 @@ enum class ConstraintType
 class Constraint
 {
 public:
-    // Construct object
```
guilpier-code, line 43: Here, some comments were removed : they added no value to the code itself.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -71,11 +72,8 @@ void InfeasibleProblemReport::sortConstraintsByType()
 {
     for (auto& c: constraints_)
     {
-        if (c.extractComponentsFromName() == 0)
```
guilpier-code, line 74: As already said above, **extractComponentsFromName** was returning the number of components a constraint's name is composed with, but this number can never be zero.
So the **if** statement was removed.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -114,14 +116,12 @@ void InfeasibleProblemReport::logSuspiciousConstraints()
     {
         Antares::logs.error() << "* Binding constraints,";
     }
-
-    Antares::logs.error() << "* Negative hurdle costs on lines with infinite capacity (rare).";
```
guilpier-code, line 118: Couldn't see any use for this log line : 
it's printed in all cases of infeasibility, even if here are no link involved (indeed, **hurdle cost**s are related to links).
Besides, it can be misleading for user, forced to wonder why he's got this message.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -114,14 +116,12 @@ void InfeasibleProblemReport::logSuspiciousConstraints()
...
 }
 
 void InfeasibleProblemReport::prettyPrint()
 {
-    sortConstraintsByType();
```
guilpier-code, line 123: The call to `sortConstraintsByType()` was moved above, with another call to a sort function.
 Besides, this call is not related to client function's name, which is supposed to print only.

src/tests/src/solver/infeasible-problem-analysis/test-unfeasible-problem-analyzer.cpp
```diff
@@ -171,9 +171,9 @@ std::unique_ptr<MPSolver> createUnfeasibleProblem(const std::string& constraintN
 }
 
 static const std::string validConstraintNames[] = {
-  "BC::hourly::hour<36>",
-  "BC::daily::day<67>",
-  "BC::weekly::week<12>",
+  "BC-name-1::hourly::hour<36>",
```
guilpier-code, line 174: Here I changed a bit the binding constraints' names.
Originally, these names started with "BC::", and it made think that BC names in **mps** files necessarily start with this prefix. This not the case, this prefix is the name of the BC.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -86,6 +84,10 @@ void InfeasibleProblemReport::logSuspiciousConstraints()
     {
         Antares::logs.error() << c.prettyPrint();
     }
+}
+
+void InfeasibleProblemReport::logInfeasibilityCauses()
```
guilpier-code, line 89: Here we split logs into 2 separate functions : 
- one that logs suspicious constraints 
- one that logs of infeasibility causes

src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -75,67 +61,36 @@ class StringIsNotWellFormated: public std::runtime_error
     }
 };
 
-std::string StringBetweenAngleBrackets(const std::string& str)
+std::string StringBetweenAngleBrackets(const std::string& constraintName)
```
a-zakir, line 64: this function is agnostic to its input and therefore the param should not be renamed;

src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -75,67 +61,36 @@ class StringIsNotWellFormated: public std::runtime_error
...
-
-    if (left == end)
+    std::string err_msg = "Error: ";
+    if (split_name.size() < 3)
     {
-        std::ostringstream stream;
-        stream << std::string("Error the string: ") << std::quoted(str)
-               << " does not contains the left angle bracket " << std::quoted("<");
-        throw StringIsNotWellFormated(stream.str());
+        err_msg += "constraint name '" + constraintName + "' misses '<' and/or '>' bracket";
```
a-zakir, line 72: or make it a class method

src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -158,61 +113,46 @@ ConstraintType Constraint::getType() const
     {
         return ConstraintType::hydro_reservoir_level;
     }
+    if (nameComponents_.at(0) == "HydroPower")
```
a-zakir: it's works because Antares provides constraints and vars names, trimmed!

flomnes (approved):

PR #2232, opened 2024-07-04
Infeasibility anaylsis : make it more changeable

guilpier-code:
Work in progress...


src/solver/infeasible-problem-analysis/constraint.cpp
```diff
@@ -36,9 +36,8 @@ const std::string kUnknown = "<unknown>";
 
 namespace Antares::Optimization
 {
-Constraint::Constraint(const std::string& name, const double slackValue):
-    name_(name),
-    slackValue_(slackValue)
+Constraint::Constraint(const std::string& name):
+    name_(name)
```
flomnes: I think this class becomes useless

src/solver/infeasible-problem-analysis/detected-infeasible-constraints.h
```diff
@@ -0,0 +1,79 @@
...
+{
+public:
+    void setConstraintName(std::string constraintName);
+    std::string regexId();
+    virtual std::string infeasisibity() = 0;
+    virtual std::string infeasisibityCause() = 0;
+
+protected:
+    std::string name_;
+    std::string regexId_;
```
flomnes: Remove protected variable, make corresponding getter regexID virtual

src/solver/infeasible-problem-analysis/constraint-slack-analysis.cpp
```diff
@@ -31,8 +32,29 @@
...
+    detectedConstraints_.push_back(std::make_shared<HourlyBC>());
+    detectedConstraints_.push_back(std::make_shared<DailyBC>());
+    detectedConstraints_.push_back(std::make_shared<WeeklyBC>());
+    detectedConstraints_.push_back(std::make_shared<FictitiousLoad>());
+    detectedConstraints_.push_back(std::make_shared<HydroLevel>());
+    detectedConstraints_.push_back(std::make_shared<STS>());
+    detectedConstraints_.push_back(std::make_shared<HydroProduction>());
+
+    std::vector<std::string> patterns;
+    std::for_each(detectedConstraints_.begin(), detectedConstraints_.end(),
```
flomnes: Use accumulate + back_inserter ?

src/solver/infeasible-problem-analysis/constraint.cpp
guilpier-code: Done

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -21,107 +21,65 @@
...
-    for (const MPVariable* slack: slackVariables)
+    for (const auto& slackVar: slackVariables_)
     {
-        constraints_.emplace_back(slack->name(), slack->solution_value());
+        for (const auto& cType: constraintTypes_)
+        {
+            if (std::regex_search(slackVar->name(), std::regex(cType->regexId())))
+            {
+                constraints_.push_back(cType->clone());
+                constraints_.back()->setConstraintName(slackVar->name());
```
flomnes:
```suggestion
                constraints_.push_back(cType->clone(slackVar->name()));
                continue;
```

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -0,0 +1,238 @@
...
+{
+    return std::make_shared<WeeklyBC>(*this);
+}
+
+std::string WeeklyBC::infeasisibity()
+{
+    return "Weekly BC '" + shortName(splitName_);
+}
+
+std::string WeeklyBC::infeasisibityCause()
```
a-zakir:
```suggestion
std::string WeeklyBC::infeasibilityCause()
```

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
```diff
@@ -0,0 +1,87 @@
...
+    virtual std::shared_ptr<WatchedConstraint> clone() const = 0;
+    virtual std::string infeasisibity() = 0;
+    virtual std::string infeasisibityCause() = 0;
+
+protected:
+    std::string name_;
+    std::string regexId_;
+    std::vector<std::string> splitName_;
+};
+
```
a-zakir, line 29: i think it's not necessary to have all these derived classes!
the only differences are the regex and the infeasibility message

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -0,0 +1,238 @@
...
+
+// =======================================
+// Weekly BC constraint
+// =======================================
+WeeklyBC::WeeklyBC()
+{
+    regexId_ = "::weekly::";
+}
+
+std::shared_ptr<WatchedConstraint> WeeklyBC::clone() const
```
a-zakir: this is not a clone, it's a ref to the same object

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -0,0 +1,238 @@
...
+    return "* Weekly binding constraints.";
+}
+
+// =======================================
+// Fictitious load constraint
+// =======================================
+FictitiousLoad::FictitiousLoad()
+{
+    regexId_ = "^FictiveLoads::";
+}
```
a-zakir: well if you insist, at least this could be in-class initialized

src/solver/infeasible-problem-analysis/constraint-slack-analysis.cpp
```diff
@@ -53,9 +71,30 @@ void ConstraintSlackAnalysis::run(MPSolver* problem)
...
+    trimSlackVariables();
+}
+
+void ConstraintSlackAnalysis::selectConstraintsToWatch(MPSolver* problem)
+{
+    std::vector<std::string> patterns;
+    std::for_each(watchedConstraintTypes_.begin(),
+                  watchedConstraintTypes_.end(),
+                  [&](auto& c) { patterns.push_back(c->regexId()); });
+    std::string constraint_name_pattern{boost::algorithm::join(patterns, "|")};
```
a-zakir: i think regex construction should not be splitted into multiple files!
for the sake of clarity, either here or in "watched constraint" classes

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/report.h
```diff
@@ -38,20 +42,18 @@ class InfeasibleProblemReport
...
-    std::vector<Constraint> constraints_;
-    std::map<ConstraintType, unsigned int> nbConstraintsByType_;
-    const unsigned int nbMaxVariables = 10;
+private:
+    void buildConstraintsFromSlackVars();
+    void filterConstraintsToOneByType();
+
+    const std::vector<std::shared_ptr<WatchedConstraint>>& constraintTypes_;
+    const std::vector<const MPVariable*>& slackVariables_;
+    std::vector<std::shared_ptr<WatchedConstraint>> constraints_;
```
a-zakir, line 50: 3 times std::vector<std::shared_ptr<WatchedConstraint>>
what's went wrong!?

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -0,0 +1,238 @@
...
+    return StringBetweenAngleBrackets(splitName.at(2));
+}
+
+namespace Antares::Optimization
+{
+
+// =======================================
+// Generic constraint logger
+// =======================================
+void WatchedConstraint::setConstraintName(std::string name)
```
a-zakir:
```suggestion
void WatchedConstraint::setConstraintName(const std::string& name)
```

a-zakir: or better set the name in the ctor?

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
guilpier-code: If you suggest we make one class with 2 member functions, each of them containing a switch or a series of **if** statements, I'm sorry to disagree. 
This is what we used to have, and what I want to avoid : it's contrary to good design rules.
Indeed, it violates **SRP** and fosters big classes/functions.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: I do agree

guilpier-code: How would you do that ?

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
a-zakir: you only need the regex and the time granularity as ctor inputs

a-zakir: and the later one can be deduced from the constraint name 'after the split

a-zakir: in fact std::string timeStep(std::vector<std::string> splitName) can also return the timstep type (hour, day or weekly)

a-zakir: std::string **WatchedConstraint**::infeasisibity()
{
    return shortName(splitName_) + "' at "+ timeStepType_+ + timeStep(splitName_);
}

a-zakir: and infeasisibityCause seems to be constant for one type ;

guilpier-code: Bodies of **infeasisibity()** and **infeasisibityCause()** can be quite different from a class to another, how would you handle with the messages they return ?

A coded example with 2 constraints would be appreciated.

a-zakir: see my last comment for infeasisibity() method 
and the infeasibilityCause can be (- const std::string member data) obtained from the constructor

src/solver/infeasible-problem-analysis/constraint-slack-analysis.cpp
guilpier-code: Now this block is gone, we no longer need **std::for_each** or **std::accumulate**

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: **clone()** function were removed, we now use a **factory** instead.

src/solver/infeasible-problem-analysis/report.cpp
guilpier-code: The inner loop was replaced with one line.

src/solver/infeasible-problem-analysis/constraint-slack-analysis.cpp
guilpier-code: Due to introduction of a factory, this part was moved, and also simplified. 
See file **watched-constraints.h/.cpp** files.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: Done, thanks

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
guilpier-code: ok, so @a-zakir suggests that we change messages of **infeasibility** and **infeasibility cause**, that is standardize them, in order to have one constraint class instead of one per constraint.
@flomnes : what do you think ?

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/report.h
guilpier-code: You right. This part is now simpler, some data members were removed.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: **setConstraintName** was removed, and replaced with a constructor, as you said.

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
flomnes: I say you may try what @a-zakir suggests, here's a [possible implementation](https://gist.github.com/flomnes/620f2767094586dbb4180914fb7bd19b)

src/solver/infeasible-problem-analysis/constraint-slack-analysis.cpp
```diff
@@ -53,9 +59,22 @@ void ConstraintSlackAnalysis::run(MPSolver* problem)
...
+    trimSlackVariables();
 }
 
-void ConstraintSlackAnalysis::addSlackVariables(MPSolver* problem)
+void ConstraintSlackAnalysis::selectConstraintsToWatch(MPSolver* problem)
+{
+    ConstraintsFactory factory;
+    std::regex rgx = factory.constraintsFilter();
+    std::copy_if(problem->constraints().begin(),
+                 problem->constraints().end(),
```
flomnes:
```suggestion
    std::ranges::copy_if(problem->constraints(),
```
https://en.cppreference.com/w/cpp/algorithm/ranges/copy

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/constraint-slack-analysis.h
```diff
@@ -52,13 +53,15 @@ class ConstraintSlackAnalysis: public UnfeasibilityAnalysis
...
-    void addSlackVariables(operations_research::MPSolver* problem);
+    void sortSlackVariablesByValue();
+    void trimSlackVariables();
 
+    std::vector<operations_research::MPConstraint*> constraintsToWatch_;
     std::vector<const operations_research::MPVariable*> slackVariables_;
-    const std::string constraint_name_pattern = "^AreaHydroLevel::|::hourly::|::daily::|::weekly::|"
-                                                "^FictiveLoads::|^Level::|"
-                                                "^HydroPower::";
+    const unsigned int nbMaxSlackVarsToKeep = 10;
```
flomnes:
```suggestion
    static const unsigned int nbMaxSlackVarsToKeep = 10;
```

flomnes: No need for multiple instances, one is enough

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -21,107 +21,58 @@
...
-        Antares::logs.error() << c.prettyPrint();
+        Antares::logs.error() << c->infeasibility();
     }
 }
 
 void InfeasibleProblemReport::logInfeasibilityCauses()
 {
     Antares::logs.error() << "Possible causes of infeasibility:";
-    if (nbConstraintsByType_[ConstraintType::hydro_reservoir_level] > 0)
+    for (const auto& c: uniqueConstraintByType_)
```
flomnes: You can use `std::unique` to remove duplicates. No need to construct this list, just use `constraints_` and remove duplicates (according to type)

See exemple [here](https://en.cppreference.com/w/cpp/algorithm/unique)

Or you can construct a `std::set` as described above.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -21,107 +21,58 @@
...
+    for (const auto& c: constraints_)
     {
-        c.extractComponentsFromName();
-        nbConstraintsByType_[c.type()]++;
+        std::string cType = typeid(*c).name();
+        if (std::find(pickedTypes.begin(), pickedTypes.end(), cType) == pickedTypes.end())
+        {
+            uniqueConstraintByType_.push_back(c);
+            pickedTypes.push_back(cType);
+        }
```
flomnes: You can use `std::unique`, or just a `std::set` with a custom comparator as described [here](https://gist.github.com/flomnes/f85f7b279ec3b6330eed4176357d1d16). Note that you could also use `std::type_info::hash_code` (type `size_t`) as a key to avoid string comparisons.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
flomnes: Actually, we're not sure `infeasibility` will be called. So there's no need to construct the string in case we need it later.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+ConstraintsFactory::ConstraintsFactory()
+{
+    regex_to_ctypes_["::hourly::"] = createHourlyBC;
+    regex_to_ctypes_["::daily::"] = createDailyBC;
+    regex_to_ctypes_["::weekly::"] = createWeeklyBC;
+    regex_to_ctypes_["^FictiveLoads::"] = createFictitiousLoad;
+    regex_to_ctypes_["^AreaHydroLevel::"] = createHydroLevel;
+    regex_to_ctypes_["^Level::"] = createSTS;
+    regex_to_ctypes_["^HydroPower::"] = createHydroProduction;
```
flomnes:
```suggestion
    regex_to_ctypes_["::hourly::"] = std::make_shared<HourlyBC>;
    regex_to_ctypes_["::daily::"] = //;
    regex_to_ctypes_["::weekly::"] = //;
    regex_to_ctypes_["^FictiveLoads::"] = //;
    regex_to_ctypes_["^AreaHydroLevel::"] = //;
    regex_to_ctypes_["^Level::"] = //;
    regex_to_ctypes_["^HydroPower::"] = //;
```

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
```diff
@@ -0,0 +1,104 @@
...
+#include <regex>
+#include <string>
+#include <vector>
+
+namespace Antares::Optimization
+{
+class WatchedConstraint
+{
+public:
+    WatchedConstraint(std::string constraintName);
```
flomnes:
```suggestion
    WatchedConstraint(const std::string& constraintName);
```
avoid copies, or use `std::move` later.

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
```diff
@@ -0,0 +1,104 @@
...
+    virtual std::string infeasibility() = 0;
+    virtual std::string infeasibilityCause() = 0;
+
+protected:
+    const std::vector<std::string>& splitName() const;
+
+protected:
+    std::string constraintName_;
```
flomnes:
```suggestion
    std::string fullName_;
```

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+std::shared_ptr<WatchedConstraint> createSTS(std::string varName)
+{
+    return std::make_shared<STS>(varName);
+}
+
+std::shared_ptr<WatchedConstraint> createHydroProduction(std::string varName)
+{
+    return std::make_shared<HydroProduction>(varName);
+}
```
flomnes:
```suggestion
```

You don't need all this, `std::make_shared<T>` is already a function.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -0,0 +1,217 @@
...
+{
+    return std::ranges::find_if(regex_to_ctypes_,
+                                [&varName](auto& p)
+                                { return std::regex_search(varName, std::regex(p.first)); })
+      ->second(varName);
+}
+
+std::regex ConstraintsFactory::constraintsFilter()
+{
+    return std::regex(boost::algorithm::join(regex_ids_, "|"));
```
flomnes:
```suggestion
    return std::regex(boost::algorithm::join(regex_to_ctypes_ | std::views::keys, "|"));
```
does this work ? If so, `regex_ids_` can be removed.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: Watched constraints no longer have a **regexId_** attribute : it's now associated to the class within in the factory.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: This isn't what we want, I guess : 
We don't want to associate to a "**regex**" an instance of a watched constraint (for example of **HourlyBC**), but rather a means to construct the type of the watched constraint : for example, to **"::hourly::"**, we'd like to associate the type a function return a fresh instance of **HourlyBC**.

So, I just changed the code to now have : 
```c++
ConstraintsFactory::ConstraintsFactory()
{
    regex_to_ctypes_ = {
      {"::hourly::", [](std::string name) { return std::make_shared<HourlyBC>(name); }},
      {"::daily::", [](std::string name) { return std::make_shared<DailyBC>(name); }},
      ...
      {"^HydroPower::", [](std::string name) { return std::make_shared<HydroProduction>(name); }}};

    auto keyView = std::views::keys(regex_to_ctypes_);
    regex_ids_ = {keyView.begin(), keyView.end()};
}
```
And I removed the creation functions.
Simpler and shorter like that.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: Please see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2232#discussion_r1672596805)

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
guilpier-code: Done

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/watched-constraints.h
guilpier-code: Done

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: Sorry I couldn't make it compile like that.

> does this work ?

yes it does

src/solver/infeasible-problem-analysis/report.cpp
flomnes: More simple : return a `std::set<std::string>` containing the possible infeasibility causes. By nature, all elements of a set are unique, so there's no need to remove duplicates.

flomnes (approved):

src/solver/infeasible-problem-analysis/include/antares/solver/infeasible-problem-analysis/constraint-slack-analysis.h
guilpier-code: done

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -40,26 +39,35 @@ void InfeasibleProblemReport::buildConstraintsFromSlackVars(
...
 
+bool compTypeName(const std::shared_ptr<WatchedConstraint> a, const std::shared_ptr<WatchedConstraint> b)
+{
+    return typeid(*a).name() < typeid(*b).name();
+}
+
+bool sameType(const std::shared_ptr<WatchedConstraint> a, const std::shared_ptr<WatchedConstraint> b)
+{
+    return typeid(*a).name() == typeid(*b).name();
+}
```
flomnes: Use [`type_index`](https://en.cppreference.com/w/cpp/types/type_index) for these operators, don't re-invent the wheel.

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -40,26 +39,35 @@ void InfeasibleProblemReport::buildConstraintsFromSlackVars(
...
-        if (std::find(pickedTypes.begin(), pickedTypes.end(), cType) == pickedTypes.end())
-        {
-            uniqueConstraintByType_.push_back(c);
-            pickedTypes.push_back(cType);
-        }
-    }
+    std::stable_sort(constraints_.begin(), constraints_.end(), compTypeName);
+    auto first_duplicate = std::unique(constraints_.begin(), constraints_.end(), sameType);
+    constraints_.erase(first_duplicate, constraints_.end());
+    std::sort(constraints_.begin(), constraints_.end(), compValue);
```
flomnes: 1. Add comments ("eliminate duplicates, etc.")
2. Sort by descending order ? Highest value first

flomnes (commented):
`const double` -> `double` should be better (passed by copy)

src/solver/infeasible-problem-analysis/report.cpp
guilpier-code: We finally chose to call **std algorithms** in a row instead.

src/solver/infeasible-problem-analysis/report.cpp
guilpier-code: We finally chose to call **std algorithms** in a row instead.

src/solver/infeasible-problem-analysis/report.cpp
guilpier-code: Done

flomnes (approved):

PR #2256, opened 2024-07-11
Minor fixes, use std::unique_ptr, add safety

flomnes:
- Use `std::unique_ptr` since we don't share objects
- Remove lambda functions (see https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique)
- Use `const std::string&`
- Make check on `create`, add safety on no result found
- Remove unused variables in tests

### DIFF
--- a/src/solver/infeasible-problem-analysis/watched-constraints.cpp
+++ b/src/solver/infeasible-problem-analysis/watched-constraints.cpp
@@ -154,22 +154,25 @@ std::string HydroProduction::infeasibilityCause()
 // --- Constraints factory ---
 ConstraintsFactory::ConstraintsFactory()
 {
-    regex_to_ctypes_ = {
-      {"::hourly::", [](std::string name) { return std::make_shared<HourlyBC>(name); }},
-      {"::daily::", [](std::string name) { return std::make_shared<DailyBC>(name); }},
-      {"::weekly::", [](std::string name) { return std::make_shared<WeeklyBC>(name); }},
-      {"^FictiveLoads::", [](std::string name) { return std::make_shared<FictitiousLoad>(name); }},
-      {"^AreaHydroLevel::", [](std::string name) { return std::make_shared<HydroLevel>(name); }},
-      {"^Level::", [](std::string name) { return std::make_shared<STS>(name); }},
-      {"^HydroPower::", [](std::string name) { return std::make_shared<HydroProduction>(name); }}};
+    regex_to_ctypes_ = {{"::hourly::", std::make_unique<HourlyBC, const std::string&>},
+                        {"::daily::", std::make_unique<DailyBC, const std::string&>},
+                        {"::weekly::", std::make_unique<WeeklyBC, const std::string&>},
+                        {"^FictiveLoads::", std::make_unique<FictitiousLoad, const std::string&>},
+                        {"^AreaHydroLevel::", std::make_unique<HydroLevel, const std::string&>},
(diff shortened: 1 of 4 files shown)


src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -154,22 +154,25 @@ std::string HydroProduction::infeasibilityCause()
...
+    regex_to_ctypes_ = {{"::hourly::", std::make_unique<HourlyBC, const std::string&>},
+                        {"::daily::", std::make_unique<DailyBC, const std::string&>},
+                        {"::weekly::", std::make_unique<WeeklyBC, const std::string&>},
+                        {"^FictiveLoads::", std::make_unique<FictitiousLoad, const std::string&>},
+                        {"^AreaHydroLevel::", std::make_unique<HydroLevel, const std::string&>},
+                        {"^Level::", std::make_unique<STS, const std::string&>},
+                        {"^HydroPower::", std::make_unique<HydroProduction, const std::string&>}};
+}
+
+std::unique_ptr<WatchedConstraint> ConstraintsFactory::create(const std::string& name) const
```
guilpier-code, line 165: About **ConstraintsFactory::create(...)** :
It seems impossible, from the **name**, not to find the right map value : the **name** necessarily matches a key in the map.
Indeed, the slack variables are filtered according to the map keys before they are passed to this **create()** function.
 That's the reason why I didn't introduce the check you just added.

guilpier-code (changes requested):

src/solver/infeasible-problem-analysis/watched-constraints.cpp
flomnes: As of now, it works. But if someone else changes the code in 6 months, they might find a crash here.

guilpier-code: ok, if you like it like that. 
At least, if someone adds a new constraint to watch, the problem you point won't happen.

flomnes: Yes, that's the whole point

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -154,22 +154,25 @@ std::string HydroProduction::infeasibilityCause()
...
-      {"^HydroPower::", [](std::string name) { return std::make_shared<HydroProduction>(name); }}};
-}
-
-std::shared_ptr<WatchedConstraint> ConstraintsFactory::create(std::string name) const
-{
-    return std::ranges::find_if(regex_to_ctypes_,
-                                [&name](auto& pair)
-                                { return std::regex_search(name, std::regex(pair.first)); })
-      ->second(name);
+    regex_to_ctypes_ = {{"::hourly::", std::make_unique<HourlyBC, const std::string&>},
```
guilpier-code, line 156: ok, I did not realize that, regarding **std::make_unique**, one could add the target **T**'s constructor arguments' types as template parameters. Good to know.

flomnes: If you want the function pointer, it's mandatory since `std::make_unique` has 6 versions. The compiler can't determine what version you're calling in this context. Here we want the version that calls the constructor `HourlyBC::HourlyBC(const std::string&)`, so that's how we tell the compiler.

guilpier-code (approved):

Issue #2244, opened 2024-07-09
Remove manual dynamic allocations


flomnes:
Dynamic allocations are considered vulnerabilities in our internal SonarCloud quality profile.

- 230 matches on `= new`
- 166 matches on `delete[]`
- 152 matches on `delete` (a few false positives)
- 2 matches on `malloc` and `free`

Some have already been replaced by `std::vector`. Other options include `std::shared_ptr` or `std::unique_ptr`.

pet-mit:
ANT-2120

PR #2270, opened 2024-07-17
Remove manual dynamic allocations in Xcast

payetvin:
#2244


src/libs/antares/study/area/list.cpp
```diff
@@ -449,8 +449,7 @@ const AreaLink* AreaList::findLink(const AreaName& area, const AreaName& with) c
 
 void AreaList::clear()
 {
-    delete[] byIndex;
-    byIndex = nullptr;
+    byIndex.clear();
```
guilpier-code, line 452: I guess that the **delete** instruction in : 
```C++
for (auto i = copy.begin(); i != end; ++i)
{
    delete i->second;
}
```
would require some work, and cannot be done in this one, right ?

src/libs/antares/study/area/list.cpp
```diff
@@ -449,8 +449,7 @@ const AreaLink* AreaList::findLink(const AreaName& area, const AreaName& with) c
...
 
     if (!areas.empty())
```
guilpier-code: Note that tests like : 
```C++
if (!areas.empty())
{
    ...
}
```
are useless. 
I guess we could remove them safely (keeping the inner block, of course), right ?

src/solver/ts-generator/xcast/xcast.cpp
```diff
@@ -58,11 +58,6 @@ XCast::XCast(Data::Study& study, Data::TimeSeriesType ts, IResultWriter& writer)
 {
 }
 
-XCast::~XCast()
```
guilpier-code, line 61: About class **XCast**'s constructor : 
```c++
    pNeverInitialized(true),
    pAccuracyOnCorrelation(false),
```
We should prefer initializing these data members inside class declaration

guilpier-code (changes requested):

src/solver/ts-generator/xcast/xcast.cpp
payetvin: Agreed, will do

src/libs/antares/study/area/list.cpp
payetvin: Indeed the Area object is used in more part of the code, I chose to stop this PR here before it become to much

guilpier-code (approved):

PR #2273, opened 2024-07-18
Remove dynamic manually allocated memory

payetvin:


flomnes (commented):
`Correlation::retrieveMontlyMatrixArray` is unused, consider removing it. It would be a good thing since `std::unique_ptr::get` is "dangerous", it can easily lead to dangling pointers
```cpp
auto ptr = std::make_unique<Cow>();
Cow* cow = ptr.get();
// ptr goes out of scope, is deleted
cow->moo(); // segfault
```

src/libs/antares/study/include/antares/study/study.h
```diff
@@ -586,7 +580,7 @@ class Study: public Yuni::NonCopyable<Study>, public LayerData
     //! \name Scenario Builder
     //@{
     //! Rules for building scenarios (can be null)
-    ScenarioBuilder::Sets* scenarioRules = nullptr;
+    std::shared_ptr<ScenarioBuilder::Sets> scenarioRules;
```
JasonMarechal25: Can't be declared by value easily. Create circular include dependency. Would require classes in antares/study/scenario-builder to be split header/impl

flomnes (approved):

PR #2274, opened 2024-07-19
Remove duplication in simulation run

guilpier-code:
This work just started...

**Situation** : 
There are code duplications :
- When running the simulation, functions we use to do it  (**runSimulationInEconomicMode()** and  **runSimulationInAdequacyMode()**) are very much the same, the only difference is the C++ type of simulation (**economy** vs **adequacy**).
- The way simulation is run from **API** or from classic **Application** look pretty close as well : see methods **APIInternal::execute()** and **Application::execute()**.

**Problem** : 
Difficult to remove the first duplication without removing the second one.

**Purpose** : 
Remove them both.


src/solver/application/include/antares/application/application.h
```diff
@@ -95,6 +96,9 @@ class Application final: public Yuni::IEventObserver<Application, Yuni::Policy::
     void runSimulationInAdequacyMode();
```
guilpier-code: These methods are about to be removed
```C++
void runSimulationInAdequacyMode();
void runSimulationInEconomicMode();
```

src/api/API.cpp
```diff
@@ -104,8 +105,6 @@ SimulationResults APIInternal::execute() const
     // Importing Time-Series if asked
     study_->importTimeseriesIntoInput();
 
-    // Stop the display of the progression
```
guilpier-code, line 107:
```C++
study_->progression.stop();
```
This instruction was removed from **API**'s simulation run : it's related to old GUI only, so irrelevant here.

src/api/API.cpp
```diff
@@ -104,8 +105,6 @@ SimulationResults APIInternal::execute() const
     // Importing Time-Series if asked
     study_->importTimeseriesIntoInput();
```
guilpier-code, line 89:
```C++
study_->importTimeseriesIntoInput();
```
Do we need this in **API** ?

src/solver/application/include/antares/application/application.h
flomnes: Why not now?

flomnes (approved):

src/solver/application/include/antares/application/application.h
guilpier-code: > Why not now?

it's still a draft : unfinished work

guilpier-code: now removed

src/solver/simulation/include/antares/solver/simulation/simulation-runner.h
```diff
@@ -22,17 +21,33 @@
...
+private:
+    template<class simulationType>
+    Benchmarking::OptimizationInfo runSimulation();
+
+    Antares::Data::Study& study_;
+    const Settings& settings_;
+    Benchmarking::DurationCollector& durationCollector_;
+    IResultWriter& resultWriter_;
+    Simulation::ISimulationObserver& simulationObserver_;
+};
```
flomnes: A class is not required here, a free function would be enough.

```suggestion
template<class simulationType>
Benchmarking::OptimizationInfo runSimulation(Antares::Data::Study& study,
                 const Settings& settings,
                 Benchmarking::DurationCollector& durationCollector,
                 IResultWriter& resultWriter,
                 Simulation::ISimulationObserver& simulationObserver);
```

src/api/API.cpp
flomnes: Probably not

src/solver/simulation/include/antares/solver/simulation/simulation-runner.h
```diff
@@ -22,17 +21,33 @@
 
 #pragma once
 
-#include "antares/infoCollection/StudyInfoCollector.h"
-#include "antares/solver/misc/options.h"
-#include "antares/solver/simulation/ISimulationObserver.h"
-#include "antares/writer/i_writer.h"
+#include "antares/solver/simulation/adequacy.h"
+#include "antares/solver/simulation/economy.h"
```
flomnes: These headers are not required here, please move them to **simulation-runner.cpp**

flomnes (commented):
Minor remarks

src/solver/simulation/include/antares/solver/simulation/simulation-runner.h
guilpier-code: Yes I hesitated on that point.

src/solver/simulation/include/antares/solver/simulation/simulation-runner.h
guilpier-code: ok, they will have to be moved to **simulation-runner.cpp**, where they are needed.

src/solver/simulation/simulation-runner.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+Benchmarking::OptimizationInfo SimulationRunner::run()
+{
+    study_.computePThetaInfForThermalClusters();
+
+    if (study_.runtime.mode == Data::SimulationMode::Adequacy)
+    {
+        return runSimulation<Solver::Simulation::ISimulation<Solver::Simulation::Adequacy>>();
+    }
```
JasonMarechal25: Please add else if case for economy for clarity.
Alternatively we can define a map <simulationMode, simulationtype>

src/solver/simulation/simulation-runner.cpp
```diff
@@ -0,0 +1,73 @@
...
+    return runSimulation<Solver::Simulation::ISimulation<Solver::Simulation::Economy>>();
+}
+
+template<class simulationType>
+Benchmarking::OptimizationInfo SimulationRunner::runSimulation()
+{
+    Simulation::NullSimulationObserver observer;
```
JasonMarechal25: observer not used

src/solver/simulation/simulation-runner.cpp
JasonMarechal25: Since runtime.mode is an enum it would be better to keep handling it with a switch case. And an except/error/log in the default case.
If we add a new mode it will default to economy in this case which may not be intended

JasonMarechal25 (dismissed):

src/solver/simulation/include/antares/solver/simulation/simulation-runner.h
guilpier-code: I agree.
Done.

src/solver/simulation/include/antares/solver/simulation/simulation-runner.h
guilpier-code: Done

src/solver/simulation/simulation-runner.cpp
guilpier-code: Done

src/solver/simulation/simulation-runner.cpp
guilpier-code: Done (**if** replaced with a **switch**)
If we add a new mode, we'll have to make changes in many places.
We'll take (I hope) this opportunity to use polymorphism instead of switching or use if everywhere.

For now **economy** mode remains the default case.

**About adding an exception** : 
Well, we all know that another case than Economy, adequacy or Expansion is impossible for now.
So do we really need an exception ?
However, if an exception is still needed, then with this exception we add a new behavior, and we should create a ticket for that.

flomnes (approved):

PR #2279, opened 2024-07-22
Districts : simplifying the code

### LINKED ISSUE #2276: Many .hxx files
Sometimes we have to use .hxx files to implement templates, so we have no choice.

But sometimes it's not necessary, .hxx files are not associated to templates
- src/libs/antares/study/include/antares/study/scenario-builder/rules.hxx
- src/libs/antares/study/include/antares/study/scenario-builder/sets.hxx (`template` StringT should be removed)
- etc.

This can lead to increased build times, since .hxx files are potentially compiled multiple times (in different CMake libraries).

guilpier-code:
Type **SetsOfAreas** (see **study.h**) is used to print aggregated results for a group of areas.
Its friend **SetsOfLinks** is defined but never used.
We remove it and make subsequent simplifications : 
- [x] class **Sets** no longer needs to be a template class, or have template methods (So **sets.hxx** is removed, and **sets.cpp** is created)
- [x] class **SetHandlerAreas** (specific operations very much related to sets of areas) needs to be moved from **Study** code area into **sets.h/.cpp**

What should be done in the future : 
- [ ] class **SetHandlerAreas** should be questioned as a class : does it constitute a cohesive code entity ? Should it not be integrated  to class **Sets** ? What should we do with its code so that **Sets** are clearer and simpler (next point may give a solution for this point) ?
- [ ] class **Sets** should be refactored : indeed, it contains many **private** **std::vector**s / **std::map**s (**pMap**, **pOptions**, **pByIndex**, ...), associating a property to a set of area. We make many loops over these containers. This is chaotic design. 
   Instead,  we should define a class **Set** (for a single set), and make existing class **Sets** a collection of **Set**. 
 Properties of a set should be contained in class **Set**. Besides the fact it would be more natural and clear, I suspect it would cause many simplifications and code reductions.


src/libs/antares/study/include/antares/study/sets.h
```diff
@@ -194,8 +195,6 @@ class Sets final
 
     uint size() const;
 
-    void rebuildIndexes();
```
guilpier-code, line 197: **void rebuildIndexes()** is used only internally, so it was made **private**.

src/libs/antares/study/include/antares/study/sets.h
```diff
@@ -238,12 +237,6 @@ class Sets final
     */
     void defaultForAreas();
 
-    /*!
-    ** \brief Rebuild the lists of a group from the rules
-    */
-    template<class HandlerT>
-    void rebuildFromRules(const IDType& id, HandlerT& handler);
```
guilpier-code, line 245: **void rebuildFromRules(...)** : same as above

src/libs/antares/study/include/antares/study/sets.hxx
```diff
@@ -21,78 +21,11 @@
 #ifndef __ANTARES_LIBS_STUDY_SETS_HXX__
 #define __ANTARES_LIBS_STUDY_SETS_HXX__
 
-namespace Antares
+namespace Antares::Data
 {
-namespace Data
```
guilpier-code, line 26: **Sets** is no longer a class template. So many method were moved to a new **sets.cpp** source file.
There are methods left : they have an extra template parameter. 
These template parameters may be removed very soon, as they are useless : used with only one effective type.

flomnes (approved):

src/libs/antares/study/area/sets.cpp
```diff
@@ -18,145 +18,174 @@
...
 {
-    using namespace Yuni;
-    const typename MapType::const_iterator end = pMap.end();
-    for (typename MapType::const_iterator i = pMap.begin(); i != end; ++i)
-    {
-        log.info() << "   found `" << i->first << "` (" << (uint)i->second->size() << ' '
-                   << (i->second->size() < 2 ? "item" : "items")
-                   << ((!hasOutput(i->first)) ? ", no output" : "") << ')';
-    }
+    typename MapOptions::const_iterator i = pOptions.find(s);
```
flomnes:
```suggestion
    auto i = pOptions.find(s);
```

src/libs/antares/study/area/sets.cpp
```diff
@@ -18,145 +18,174 @@
...
+    return (i != pOptions.end()) ? i->second.output : false;
+}
+
+uint Sets::resultSize(const Yuni::ShortString128& s) const
+{
+    typename MapOptions::const_iterator i = pOptions.find(s);
```
flomnes:
```suggestion
    auto i = pOptions.find(s);
```

src/libs/antares/study/area/sets.cpp
```diff
@@ -18,145 +18,174 @@
...
+    return (i != pOptions.end()) ? i->second.resultSize : 0;
 }
 
-template<class T>
-void Sets<T>::defaultForAreas()
+Sets::IDType Sets::caption(const Yuni::ShortString128& s) const
+{
+    typename MapOptions::const_iterator i = pOptions.find(s);
```
flomnes:
```suggestion
    auto i = pOptions.find(s);
```

src/libs/antares/study/area/sets.cpp
```diff
@@ -400,64 +367,103 @@ void Sets<T>::rebuildIndexes()
...
-template<class T>
-template<class StringT>
-inline uint Sets<T>::resultSize(const StringT& s) const
+void Sets::dumpToLogs() const
 {
-    // Assert, if a C* container can not be found at compile time
-    static_assert(Yuni::Traits::CString<StringT>::valid);
+    using namespace Yuni;
+    const typename MapType::const_iterator end = pMap.end();
+    for (typename MapType::const_iterator i = pMap.begin(); i != end; ++i)
```
flomnes:
```suggestion
    for (const auto& [name, content] : pMap)
```

src/libs/antares/study/area/sets.cpp
```diff
@@ -400,64 +367,103 @@ void Sets<T>::rebuildIndexes()
...
+{
+    std::ranges::for_each(otherSet, [&set](auto* area) { set.erase(area); });
+}
+
+bool SetHandlerAreas::applyFilter(Sets::SetAreasType& set, const std::string& value)
+{
+    if (value == "add-all")
+    {
+        auto end = areas_.end();
+        for (auto i = areas_.begin(); i != end; ++i)
```
flomnes: Use modern syntax ?

flomnes (commented):
Minor remarks.
- `typename` makes no sense if there is no template, use `auto` instead
- Use modern style for loops

guilpier-code:
> * `typename` makes no sense if there is no template, use `auto` instead

Done : all **typename**s were removed from **sets.h/.cpp**

src/libs/antares/study/area/sets.cpp
guilpier-code: I'll do that, no problem, but we should keep in mind that any **std::map** / **std::vector** used as a data members of class **Sets** (for instance **pMap**) should not exist : please see PR presentation comment.
**std::map** and  **std::vector** are used because of a chaotic architecture of class **Sets**.
Making changes on these **std::map** and  **std::vector** manipulations is making changes on things that should disappear.

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

src/libs/antares/study/area/sets.cpp
```diff
@@ -105,9 +105,9 @@ void Sets::defaultForAreas()
 
 void Sets::rebuildAllFromRules(SetHandlerAreas& handler)
 {
-    for (uint i = 0; i != pMap.size(); ++i)
+    for (auto setId: pNameByIndex)
```
flomnes:
```suggestion
    for (const auto& setId: pNameByIndex)
```

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

src/libs/antares/study/area/sets.cpp
```diff
@@ -356,7 +356,7 @@ void Sets::rebuildIndexes()
     pByIndex.resize(pMap.size());
 
     uint index = 0;
-    for (auto& [setId, set] : pMap)
+    for (auto& [setId, set]: pMap)
```
flomnes:
```suggestion
    for (const auto& [setId, set]: pMap)
```

src/libs/antares/study/area/sets.cpp
```diff
@@ -382,7 +382,7 @@ uint Sets::resultSize(const uint index) const
 void Sets::dumpToLogs() const
 {
     using namespace Yuni;
-    for (auto& [setId, set] : pMap)
+    for (auto& [setId, set]: pMap)
```
flomnes:
```suggestion
    for (const auto& [setId, set]: pMap)
```

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

flomnes (dismissed):

flomnes:
Add unit / study tests

guilpier-code:
> Add unit / study tests

I added many tests to be run on CI (see [this commit](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2279/commits/cf8fa438d20a52c67d1b796ed191080d79934303)), knowing that district results are taken into account when running CI tests.

src/libs/antares/study/area/sets.cpp
guilpier-code: Done

src/libs/antares/study/area/sets.cpp
```diff
@@ -18,118 +18,94 @@
...
     using namespace Yuni;
     clear();
     Options opts;
     opts.caption = "All areas";
     opts.comments = "Spatial aggregates on all areas";
     opts.output = false;
-    opts.rules.push_back(Rule(ruleFilter, new String("add-all")));
-    auto item = std::make_shared<T>();
-    add("all areas", item, opts);
+    opts.rules.push_back(Rule(ruleFilter, "add-all"));
```
flomnes:
```suggestion
    opts.rules.emplace_back(ruleFilter, "add-all");
```

src/libs/antares/study/area/sets.cpp
```diff
@@ -239,17 +209,17 @@ bool Sets<T>::loadFromFile(const std::filesystem::path& filename)
 
                 if (p->key == "+")
                 {
-                    opts.rules.push_back(Rule(ruleAdd, new String(value)));
+                    opts.rules.push_back(Rule(ruleAdd, value.to<std::string>()));
```
flomnes: This is so much better, the previous code was very confusing regarding memory ownership

.github/workflows/ubuntu.yml
flomnes, line 1: Revert this

.github/workflows/windows-vcpkg.yml
flomnes, line 1: Revert

flomnes (commented):
Minor remarks
- Revert workflow files
- Use `emplace_back(args...)` instead of `push_back(T(args...))`

guilpier-code:
> Add unit / study tests

Unit tests could be added easier by doing recommended refactoring (see PR presentation)

src/libs/antares/study/area/sets.cpp
guilpier-code: Construction in place. I get it.

flomnes (approved):

PR #2280, opened 2024-07-22
Add short test to coverage analysis

a-zakir:
- Add "short-tests" study batch to coverage analysis
- Use Release build type for SonarCloud CI (previously Debug) to avoid saturating the RAM on GH runners (Debug + Coverage instrumentation is a bit too much)
- (minor) Display error code in case of failure for study tests


src/analyzer/atsp/atsp.cpp
```diff
@@ -10,7 +10,7 @@
 ** the Mozilla Foundation, either version 2 of the License, or
 ** (at your option) any later version.
 **
-** Antares_Simulator is distributed in the hope that it will be useful,
+** XNothingX in the hope that it will be useful,
```
flomnes: 🤔

flomnes (approved):

PR #2284, opened 2024-07-24
Use a formula for the number of cores used

### LINKED ISSUE #2266: Number of cores in parallel
`getRawNumberCoresPerLevel` in study.cpp

We should have a formula for every number of core instead of this horrible switch case

payetvin:
close #2266

### DIFF
--- a/docs/user-guide/solver/optional-features/multi-threading.md
+++ b/docs/user-guide/solver/optional-features/multi-threading.md
@@ -44,35 +44,43 @@ The Table indicates either the refresh status (No) or the refresh span (the asso
 
 In 9.2 We changed the formula for the number of cores to simplify. Here's the old values and the new ones.
 
-Previously
-[nbCores] ncMin ncLow ncAvg ncHigh ncMax
-[1] 1 1 1 1 1
-[2] 1 1 1 2 2
-[3] 1 2 2 2 3
-[4] 1 2 2 3 4
-[5] 1 2 3 4 5
-[6] 1 2 3 4 6
-[7] 1 2 3 5 7
-[8] 1 2 4 6 8
-[9] 1 3 5 7 8
-[10] 1 3 5 8 9
-[11] 1 3 6 8 10
-[12] 1 3 6 9 11
+### Previously
 
-Now
-[nbCores] ncMin ncLow ncAvg ncHigh ncMax
-[1] 1 1 1 1 1
-[2] 1 1 1 2 2
-[3] 1 1 2 3 3
-[4] 1 1 2 3 4
-[5] 1 2 3 4 5
-[6] 1 2 3 5 6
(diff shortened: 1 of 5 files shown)


flomnes (changes requested):
Good ! But then you don't need a `std::map` then, so please remove it.
```cpp
uint computeNumberOfCores(uint nbLogicalCores, Antares::Data::NumberOfCoresMode ncMode)
{
switch (ncMode)
{
    case ncMin: return 1;
    case ncLow: return std::ceil(nbLogicalCores / 4.);
    // etc.
}
}
```

flomnes:
Could you also add some tests ?

docs/user-guide/solver/optional-features/multi-threading.md
```diff
@@ -54,6 +42,45 @@ Examples of reduction from an initial allowance of 12 cores are given hereafter.
 
 The Table indicates either the refresh status (No) or the refresh span (the associated refresh status "yes" is implicit).
 
+In 9.2 We changed the formula for the number of cores to simplify. Here's the old values and the new ones.
+
+### Previously
```
flomnes:
```suggestion
### Before v9.2
```

docs/user-guide/solver/optional-features/multi-threading.md
```diff
@@ -54,6 +42,45 @@ Examples of reduction from an initial allowance of 12 cores are given hereafter.
...
+-|          _7_          |     1     |     2     |     3     |     5      |     7     |
+-|          _8_          |     1     |     2     |     4     |     6      |     8     |
+-|          _9_          |     1     |     3     |     5     |     7      |     8     |
+-|         _10_          |     1     |     3     |     5     |     8      |     9     |
+-|         _11_          |     1     |     3     |     6     |     8      |    10     |
+-|         _12_          |     1     |     3     |     6     |     9      |    11     |
+-|      _S &gt; 12_      |     1     | Ceil(S/4) | Ceil(S/2) | Ceil(3S/4) |    S-1    |
+
+
+### Since 9.2
```
flomnes:
```suggestion
### Starting from v9.2
```

docs/user-guide/solver/optional-features/multi-threading.md
```diff
@@ -54,6 +42,45 @@ Examples of reduction from an initial allowance of 12 cores are given hereafter.
 
 The Table indicates either the refresh status (No) or the refresh span (the associated refresh status "yes" is implicit).
 
+In 9.2 We changed the formula for the number of cores to simplify. Here's the old values and the new ones.
```
flomnes: Please place "Starting from v9.2" first, the older one after

flomnes (approved):

PR #2285, opened 2024-07-25
Feature/test name translator

a-zakir:


flomnes (approved):
The intent of these tests is not very clear

PR #2292, opened 2024-07-31
Use raw-pointer & delete visitor

flomnes:
Pros
- Memory management is simple. Just good old raw pointers

Cons
- Memory leaks, etc.


flomnes:
Sonar wasn't happy, we neither.

PR #2298, opened 2024-08-01
Use std::unique_ptr

flomnes:
TODO : use `std::unique_ptr` in node constructors ?


a-zakir:
to fully take owenship of child nodes, every node could expose addChild methods:

`
Add a;
a.addLeftChild<Literal, double>(2.1);-->template<class ChildType, class... Val> addLeftChild( Val... pargs){n1_ = std::make_unique<ChildType>(pargs...)
`

PR #2307, opened 2024-08-06
Use memory manager instead of std::unique_ptr

flomnes:


src/solver/expressions/include/antares/solver/expressions/Registry.hxx
```diff
@@ -0,0 +1,50 @@
...
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+#include <memory>
+#include <mutex>
+#include <vector>
```
flomnes, line 26: add `#include <concept>`

a-zakir (approved):

PR #2311, opened 2024-08-07
Re-enable full build of ortools after new repo paradigm (ortools-rte)

JasonMarechal25:
- Require https://github.com/rte-france/or-tools-rte/pull/13 and new release
- `set(ortools_REF ${ORTOOLS_TAG})` is open to debate. For now I suppose ortools-rte and google/ortools to have the same tags, but this allows some differences (like patches on ortools-rte versions)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,16 +1,16 @@
 cmake_minimum_required(VERSION 3.14) # FetchContent_MakeAvailable
 
 # Version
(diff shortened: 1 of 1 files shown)


flomnes:
This would be useful, but I'm not sure we want to keep this confusing machinery for the long term. And most of all, leak details about OR-Tools into Antares_Simulator's build instructions.

PR #2317, opened 2024-08-07
Add context for evaluation, VariableNode

flomnes:
- Add evaluation context for variable & parameter nodes
- Add forgotten boilerplate code for variable nodes (Visitor.h)
- Add 2 simple tests: one for a variable node, one for a parameter node


src/solver/expressions/visitors/EvalVisitor.cpp
```diff
@@ -1,23 +1,53 @@
...
+*/
 
 #include "antares/solver/expressions/visitors/EvalVisitor.h"
 
 #include <antares/solver/expressions/nodes/ExpressionsNodes.h>
 
 namespace Antares::Solver::Expressions
 {
+EvalVisitor::EvalVisitor(EvaluationContext context):
+    context_(std::move(context))
```
a-zakir, line 29: why?

flomnes: We `std::move` a copy, the original is untouched

a-zakir (approved):

PR #2329, opened 2024-08-14
Remove BinaryNode, UnaryNode, use variadic constructor

flomnes:


src/solver/expressions/include/antares/solver/expressions/nodes/ParentNode.h
```diff
@@ -41,6 +41,13 @@ class ParentNode: public Node
     {
     }
 
+    template<typename... Args>
+    explicit ParentNode(Args... args):
+        children_{{args...}} // Initialize the array with the provided Node* arguments
+    {
```
a-zakir, line 47: we have lost two specialized class in favor of complex code with an assert...
you could use std::initializer_list<Node*>

PR #2333, opened 2024-08-14
Substitution from context visitor

flomnes:


src/tests/src/solver/expressions/test_expressions.cpp
```diff
@@ -375,3 +377,61 @@ BOOST_FIXTURE_TEST_CASE(simple_constant_expression, Registry<Node>)
...
+
+private:
+    Node* visit(const Nodes::ComponentVariableNode& component_variable_node) override
+    {
+        if (auto it = std::find(ctx_.variables.begin(),
+                                ctx_.variables.end(),
+                                component_variable_node);
+            it != ctx_.variables.end())
+        {
+            return const_cast<Node*>(dynamic_cast<const Node*>(&(*it)));
```
flomnes: danger zone

src/solver/expressions/visitors/SubstitutionVisitor.cpp
```diff
@@ -0,0 +1,50 @@
...
+    ctx_(ctx),
+    registry_(registry)
+{
+}
+
+Nodes::Node* SubstitutionVisitor::visit(const Nodes::ComponentVariableNode& component_variable_node)
+{
+    auto it = std::find_if(ctx_.variables.begin(),
+                           ctx_.variables.end(),
+                           [&component_variable_node](Nodes::ComponentVariableNode* x)
```
a-zakir:
```suggestion
                           [&component_variable_node](const auto* x)
```

src/solver/expressions/include/antares/solver/expressions/visitors/SubstitutionVisitor.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <antares/solver/expressions/Registry.hxx>
+#include <antares/solver/expressions/nodes/NodesForwardDeclaration.h>
+#include <antares/solver/expressions/visitors/CloneVisitor.h>
+
+namespace Antares::Solver::Visitors
+{
+struct SubstitutionContext
+{
+    std::vector<Nodes::ComponentVariableNode*> variables;
```
a-zakir: if the ComponentVariableNode::id is unique, variables could be an unordered_map, faster and with direct access

flomnes: I now use an `std::unordered_set`

src/solver/expressions/visitors/SubstitutionVisitor.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <algorithm>
+
+#include <antares/solver/expressions/nodes/ComponentNode.h>
+#include <antares/solver/expressions/visitors/SubstitutionVisitor.h>
+
+namespace Antares::Solver::Visitors
+{
+SubstitutionVisitor::SubstitutionVisitor(SubstitutionContext& ctx, Registry<Nodes::Node>& registry):
+    CloneVisitor(registry),
```
a-zakir, line 30: make Registry<...> the first arg

flomnes: done 714d20310

src/solver/expressions/include/antares/solver/expressions/visitors/SubstitutionVisitor.h
flomnes: Actually it's a bit more complicated. `std::set::find` only works if we lookup elements, there is no `std::set::find_if`. So we have to use `std::find_if` that has linear complexity. I left a comment in the code, so we can come back to it later if need be.

PR #2334, opened 2024-08-16
avoid to break constness && dynamic cast

a-zakir:


src/tests/src/solver/expressions/test_expressions.cpp
```diff
@@ -406,16 +406,17 @@ class SubstitutionVisitor: public CloneVisitor
...
+
         else
         {
             return CloneVisitor::visit(component_variable_node);
         }
     }
 };
 
-void fillContext(ANTLRContext& ctx)
+void fillContext(ANTLRContext& ctx, Registry<Node>& registry)
```
flomnes: It doesn't seem to be necessary to pass a registry here (unused argument).

a-zakir: yes, from an old commit

flomnes (approved):

PR #2335, opened 2024-08-16
Equality visitor

a-zakir:
draft

### DIFF
--- a/src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
+++ b/src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
@@ -34,7 +34,7 @@ std::optional<RetT> tryType(const Node& node1, const Node& node2, VisitorT& visi
 {
     auto x = dynamic_cast<const NodeT*>(&node1);
     auto y = dynamic_cast<const NodeT*>(&node2);
-    if (x && y)
+    if (x && y && (typeid(*x) == typeid(*y)))
     {
         return visitor.visit(*x, *y);
     }
@@ -44,6 +44,11 @@ std::optional<RetT> tryType(const Node& node1, const Node& node2, VisitorT& visi
     }
 }
 } // namespace
+
+struct TwoNodesVisitorException: std::runtime_error
+{
+    using std::runtime_error::runtime_error;
+};
 
 template<class R>
 class TwoNodesVisitor
@@ -75,8 +80,8 @@ public:
                 return ret.value();
             }
         }
-        logs.error() << "Antares::Solver::Nodes Visitor: unsupported Node!";
-        return R();
+        throw TwoNodesVisitorException(
+          "Antares::Solver::Nodes TwoNodesVisitor: the inputs nodes must be of the same type!");
     }
 
     virtual R visit(const AddNode&, const AddNode&) = 0;
--- a/src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
+++ b/src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
@@ -45,11 +45,6 @@ std::optional<RetT> tryType(const Node& node1, const Node& node2, VisitorT& visi
 }
 } // namespace
 
-struct TwoNodesVisitorException: std::runtime_error
-{
-    using std::runtime_error::runtime_error;
-};
-
 template<class R>
 class TwoNodesVisitor
 {
@@ -80,8 +75,8 @@ public:
                 return ret.value();
             }
         }
-        throw TwoNodesVisitorException(
-          "Antares::Solver::Nodes TwoNodesVisitor: the inputs nodes must be of the same type!");
+        logs.error()
+          << "Antares::Solver::Nodes TwoNodesVisitor: the inputs nodes must be of the same type!";
     }
 
     virtual R visit(const AddNode&, const AddNode&) = 0;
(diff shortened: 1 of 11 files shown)


src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
```diff
@@ -0,0 +1,98 @@
...
+
+template<class R>
+class TwoNodesVisitor
+{
+public:
+    virtual ~TwoNodesVisitor() = default;
+
+    R dispatch(const Node& node1, const Node& node2)
+    {
+        using FunctionT = std::optional<R> (*)(const Node&, const Node&, TwoNodesVisitor<R>&);
```
flomnes, line 56: What about different node types ? Is there a default case for this ? For example `AddNode` and `MultiplicationNode`.

flomnes (commented):
The key is to have something like
```cpp
bool operator==(const Node& node1, const Node& node2)
```
that works for all node types. I'm not sure your implementation does that.

src/tests/src/solver/expressions/test_expressions.cpp
```diff
@@ -375,3 +376,37 @@ BOOST_FIXTURE_TEST_CASE(simple_constant_expression, Registry<Node>)
...
+    VariableNode y2("y");
+    // Mult : 56.*y
+    Node* v1 = create<MultiplicationNode>(&lit3, &y1);
+    Node* v2 = create<MultiplicationNode>(&lit4, &y2);
+
+    Node* add1 = create<AddNode>(u1, v1);
+    CloneVisitor cloneVisitor(*this);
+    Node* add1_clone = cloneVisitor.dispatch(*add1);
+
+    Node* add2 = create<AddNode>(u2, v2);
```
flomnes: Write a function to create the tree
```cpp
Node* root1 = createExpressionTree();
Node* root2 = createExpressionTree();
BOOST_CHECK(equalityVisitor.dispatch(*root1, *root2));
```

src/tests/src/solver/expressions/test_expressions.cpp
```diff
@@ -375,3 +376,37 @@ BOOST_FIXTURE_TEST_CASE(simple_constant_expression, Registry<Node>)
...
+
+    PrintVisitor printVisitor;
+    EqualityVisitor equalityVisitor;
+
+    BOOST_CHECK_EQUAL(printVisitor.dispatch(*add1), printVisitor.dispatch(*add1_clone));
+    BOOST_CHECK_EQUAL(equalityVisitor.dispatch(*add1, *add2), true);
+    // TODO test depth
+    AddNode random_add(create<LiteralNode>(25.), create<VariableNode>("rand_var"));
```
flomnes: You may also compare a root to itself
```cpp
BOOST_CHECK(equalityVisitor.dispatch(*root, *root));
```

src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
```diff
@@ -0,0 +1,98 @@
...
+                                         &tryType<R, TwoNodesVisitor<R>, ComponentVariableNode>,
+                                         &tryType<R, TwoNodesVisitor<R>, ComponentParameterNode>};
+        for (auto f: functions)
+        {
+            if (auto ret = f(node1, node2, *this); ret.has_value())
+            {
+                return ret.value();
+            }
+        }
+        logs.error() << "Antares::Solver::Nodes Visitor: unsupported Node!";
```
flomnes: If `node1` and `node2` are of different types, we will encounter this line. So I think we should handle errors better here.
Example
```cpp
Literal lit1(2), lit2(2);
AddNode node1(&lit1, &lit2);
MultiplicationNode node2(&lit1, &lit2);
EqualityVisitor equal;
// Here, we go through l. 78 above, but the caller only gets a log error
// and an `R()`, not so easy to catch
equal.dispatch(&node1, &node2); 
```

src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
```diff
@@ -0,0 +1,98 @@
...
+{
+    auto x = dynamic_cast<const NodeT*>(&node1);
+    auto y = dynamic_cast<const NodeT*>(&node2);
+    if (x && y)
+    {
+        return visitor.visit(*x, *y);
+    }
+    else
+    {
+        return {};
```
flomnes, line 43: Different node types

src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
a-zakir: yes i have to check node consistency

src/tests/src/solver/expressions/test_expressions.cpp
flomnes: BTW it's more like
```cpp
Node* root1 = createExpressionTree(registry);
```

src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
a-zakir: update: the code print warning message and returns false

src/solver/expressions/include/antares/solver/expressions/nodes/TwoNodesVisitor.h
a-zakir: the return type is handled in the dispatch method

PR #2337, opened 2024-08-19
Enable parallel run for API

JasonMarechal25:
Enable parallel by default


flomnes (commented):
Why not something like
```cpp
Options options;
options.field1 = "value1";
options.field2 = true;
// etc.
Antares::Solver::Application application(options);
```
?
We don't really need to parse arguments here

JasonMarechal25:
> Why not something like
> 
> ```c++
> Options options;
> options.field1 = "value1";
> options.field2 = true;
> // etc.
> Antares::Solver::Application application(options);
> ```
> 
> ? We don't really need to parse arguments here

This is the current behaviour of _Application_. _Application::prepare()_ parse the command line and do a lot of stuff on settings and options.
Passing options directly would indeed be better but it's not trivial as we would need to keep the current behaviour.

flomnes (approved):

PR #2339, opened 2024-08-19
Time index visitor

a-zakir:


src/solver/expressions/nodes/TimeIndex.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+bool TimeIndex::IsScenarioVarying() const
+{
+    return scenario_varying_;
+}
+
+bool TimeIndex::operator==(const TimeIndex& other) const
+{
+    return (time_varying_ == other.time_varying_) && (scenario_varying_ == other.scenario_varying_);
+}
```
flomnes: Use a single `enum class` ? So you don't need to define trivial operators such as this one

flomnes (commented):
- Use `enum class` + free functions
- No need for `operator+`, `operator*`, etc. since they all do the same thing. You can just define `TimeIndex operator||(TimeIndex a, TimeIndex b)` (currently `Connect`) and use it in visitors
- To avoid changing nodes (new members & constructors), you can define a `std::map<Node*, TimeIndex>` member for `TimeIndexVisitor`.

src/solver/expressions/include/antares/solver/expressions/visitors/EvaluationContext.h
```diff
@@ -1,23 +1,35 @@
...
+    ValueType getParameterValue(const KeyType& key) const
+    {
+        return parameters_.at(key);
+    }
 
 private:
-    std::map<std::string, double> parameters_;
-    std::map<std::string, double> variables_;
+    std::map<KeyType, ValueType> parameters_;
+    std::map<KeyType, ValueType> variables_;
```
flomnes: In my opinion you don't need a template here. You can just declare a single
`std::map<Node*, TimeIndex>`.

For the use cases, you just need to populate this map for all leaves (parameters, literal, etc.)

src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
```diff
@@ -28,7 +28,7 @@ class LinearStatus
 public:
     enum Status
     {
-        CONSTANT,
+        CONSTANT_EXPR,
```
flomnes: Conflicts with #2343

src/solver/expressions/nodes/VariableNode.cpp
```diff
@@ -0,0 +1,38 @@
...
+    Leaf(value),
+    TimeIndex()
+{
+}
+
+VariableNode::VariableNode(const std::string& value, bool time_varying, bool scenario_varying):
+    Leaf(value),
+    TimeIndex(time_varying, scenario_varying)
+{
+}
```
flomnes, line 37: In my opinion it's not necessary to define new nodes, or add attributes to existing nodes.

src/tests/src/solver/expressions/test_expressions.cpp
```diff
@@ -359,5 +360,33 @@ BOOST_FIXTURE_TEST_CASE(simple_constant_expression, Registry<Node>)
...
+
+BOOST_FIXTURE_TEST_CASE(simple_time_dependant_expression, Registry<Node>)
+{
+    PrintVisitor printVisitor;
+    TimeIndexVisitor timeIndexVisitor;
+    // LiteralNode --> constant in time and for all scenarios
+    LiteralNode literalNode(65.);
+    BOOST_CHECK_EQUAL(timeIndexVisitor.dispatch(literalNode).GetTimeIndex().IsTimeVarying(), false);
+    BOOST_CHECK_EQUAL(timeIndexVisitor.dispatch(literalNode).GetTimeIndex().IsScenarioVarying(),
+                      false);
```
flomnes: You could just extract the enum value, it would be shorter

flomnes:
Duplicate of #2349

PR #2340, opened 2024-08-19
Fix clang++ warning about typeid

flomnes:
Expression with side effects will be evaluated despite being used as an operand to 'typeid'

### DIFF
--- a/src/solver/infeasible-problem-analysis/report.cpp
+++ b/src/solver/infeasible-problem-analysis/report.cpp
@@ -49,13 +49,18 @@ void InfeasibleProblemReport::buildConstraintsFromSlackVars(
 bool lessTypeName(const std::shared_ptr<WatchedConstraint> a,
                   const std::shared_ptr<WatchedConstraint> b)
 {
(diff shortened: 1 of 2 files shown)


payetvin (approved):

src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -49,13 +49,18 @@ void InfeasibleProblemReport::buildConstraintsFromSlackVars(
 bool lessTypeName(const std::shared_ptr<WatchedConstraint> a,
                   const std::shared_ptr<WatchedConstraint> b)
 {
-    return std::type_index(typeid(*a)) < std::type_index(typeid(*b));
+    const WatchedConstraint* a_raw = a.get();
```
guilpier-code, line 52: Could you explain a bit more the problem we want to avoid by making this change (which seems to be equivalent) ? Thanx.

flomnes: `*a` dereferences the pointer, while `a.get()` doesn't.

PR #2343, opened 2024-08-20
Use enum class & free functions for `LinearStatus`, add unit tests

flomnes:
Some functions may have changed behavior, ~~we need more unit tests~~ I added **a lot** of unit tests.


src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
```diff
@@ -20,121 +20,80 @@
...
+        return a;
+    case LinearStatus::LINEAR:
+        return b;
+
+    default:
+        return LinearStatus::NON_LINEAR;
     }
+}
 
-    constexpr bool operator==(Status status) const
```
guilpier-code, line 122: I can see that old **operator==**,  **operator!=** and **operator-** disappeared.
Probably because it's useless, or we don't see a usage for them.

src/solver/expressions/visitors/LinearStatus.cpp
```diff
@@ -0,0 +1,40 @@
...
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+
+#include <antares/solver/expressions/visitors/LinearStatus.h>
+
+namespace Antares::Solver::Visitors
+{
+std::ostream& operator<<(std::ostream& os, LinearStatus s)
```
guilpier-code: I can't see any use of this **operator<<**. Is there any ?
If not, we should remove it.

src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
```diff
@@ -20,121 +20,115 @@
...
-            }
-            else
-            {
-                return LinearStatus::NON_LINEAR;
-            }
-        default:
-            return LinearStatus::NON_LINEAR;
-        }
-    }
+constexpr char pair(LinearStatus a, LinearStatus b)
```
guilpier-code, line 35: What's the problem with using classic **std::pair** ?
Why do we need the help of **char** ?

guilpier-code (changes requested):

src/solver/expressions/visitors/LinearStatus.cpp
flomnes: It's only useful for tests (`BOOST_CHECK_EQUAL`). Shall move it to test_expression.cpp ?

src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
flomnes: `switch..case` only accepts integers (int, unsigned int, char, etc.), `std::pair` is not an integer or convertible to an integer

src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
flomnes: Yes these operators are generated automatically by the compiler for `enum class`es.

src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
guilpier-code: ok, then why do we use switch.. case, and not a std::map that associates std::pair of enum value to enum value.
Is there a problem with this solution ?

flomnes: Yes, it's not `constexpr`.

src/tests/src/solver/expressions/test_expressions.cpp
```diff
@@ -260,6 +279,90 @@ BOOST_FIXTURE_TEST_CASE(comparison_node, Registry<Node>)
...
+
+    BOOST_CHECK_EQUAL(LinearStatus::NON_LINEAR / LinearStatus::NON_LINEAR,
+                      LinearStatus::NON_LINEAR);
+}
+
+static const std::vector<LinearStatus> LinearStatus_ALL = {LinearStatus::LINEAR,
+                                                           LinearStatus::NON_LINEAR,
+                                                           LinearStatus::CONSTANT};
+
+BOOST_AUTO_TEST_CASE(linear_status_minus)
```
guilpier-code: Instead of loops for over the same list of values, could we not do like what we did for **infeasible problems** unit tests : 
```C++
BOOST_DATA_TEST_CASE(linear_status_minus,
                     bdata::make(LinearStatus_ALL),
                     x)
{
    BOOST_CHECK_EQUAL(x, -x);
}
```
or something like that...

guilpier-code (changes requested):

src/solver/expressions/visitors/LinearStatus.cpp
guilpier-code: Yes, you're right, we should. This operator should not be in production code if not used in it.

src/tests/src/solver/expressions/test_expressions.cpp
flomnes: Done 2ebb72192

src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
```diff
@@ -20,121 +20,112 @@
...
-        return status_;
-    }
+constexpr char pair(LinearStatus a, LinearStatus b)
+{
+    return static_cast<char>(a) << 4 | static_cast<char>(b);
+}
 
-    // Comparisons
-    constexpr bool operator==(LinearStatus a) const
+constexpr LinearStatus operator*(LinearStatus a, LinearStatus b)
```
guilpier-code, line 40: As **LinearStatus::NON_LINEAR** is the default case, we could ignore cases where we return **LinearStatus::NON_LINEAR**.
If you think this leads to a lack of clarity, ignore this comment.

guilpier-code (changes requested):

src/solver/expressions/include/antares/solver/expressions/visitors/LinearStatus.h
flomnes: I think it's best to keep all cases, for clarity.

guilpier-code (approved):

PR #2346, opened 2024-08-21
Fix various compilation warnings

flomnes:
- Comparison of a bool
- Loss of integer precision
- Conversion of double -> unsigned

### DIFF
--- a/src/libs/antares/utils/utils.cpp
+++ b/src/libs/antares/utils/utils.cpp
@@ -147,7 +147,7 @@ bool isZero(double d)
 
 double round(double d, unsigned precision)
 {
-    unsigned factor = std::pow(10, precision);
+    double factor = std::pow(10, precision);
     return std::round(d * factor) / factor;
 }
 
--- a/src/libs/antares/utils/utils.cpp
+++ b/src/libs/antares/utils/utils.cpp
@@ -147,7 +147,7 @@ bool isZero(double d)
 
 double round(double d, unsigned precision)
 {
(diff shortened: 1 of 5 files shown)


src/libs/antares/utils/utils.cpp
```diff
@@ -147,7 +147,7 @@ bool isZero(double d)
 
 double round(double d, unsigned precision)
 {
-    unsigned factor = std::pow(10, precision);
+    double factor = std::pow(10, precision);
```
a-zakir: auto factor = std:pow(10.,precision)?

flomnes: Done

src/libs/antares/writer/zip_writer.cpp
```diff
@@ -99,7 +99,7 @@ void ZipWriteJob<ContentT>::writeEntry()
     {
         logErrorAndThrow("Error opening entry " + pEntryPath + " (" + std::to_string(ret) + ")");
     }
-    int32_t bw = mz_zip_writer_entry_write(pZipHandle, pContent.data(), pContent.size());
+    int32_t bw = mz_zip_writer_entry_write(pZipHandle, pContent.data(), (int32_t)pContent.size());
```
JasonMarechal25:
```suggestion
    int32_t bw = mz_zip_writer_entry_write(pZipHandle, pContent.data(), static_cast<int32_t>(pContent.size()));
```

JasonMarechal25 (approved):

PR #2347, opened 2024-08-21
Cucumber test example

pet-mit:
Add support for cucumber tests for functional tests, using "behave" for Python.
Some first scenarios, copied from the legacy non-regression tests, as a showcase.


flomnes:
You can upload logs as artifacts to see what's going on
```yaml
      - name: Upload logs on failure
        if: ${{ failure() }}
        uses: actions/upload-artifact@v4
        with:
          name: cucumber build log
          path: ${{ github.workspace }}/vcpkg/buildtrees/cucumber-cpp/install-x64-linux-release-rel-out.log
```

src/ports/cucumber-cpp/vcpkg.json
```diff
@@ -0,0 +1,32 @@
...
+      "version>=": "1.22.1"
+    },
+    {
+      "name": "nlohmann-json",
+      "version>=": "3.10.5"
+    },
+    {
+      "name": "nlohmann-json",
+      "version>=": "3.10.5"
+    }
```
flomnes: Duplicate ?

src/vcpkg.json
```diff
@@ -34,6 +34,14 @@
...
       ]
+    },
+    {
+      "name": "cucumber-cpp",
+      "version>=": "0.7.0"
+    },
+    {
+      "name": "gtest",
+      "version>=": "1.11.0"
     }
```
flomnes: cucumber-cpp (optionnally) depends on gtest, so it would make more sense to have this in src/ports/cucumber-cpp/vcpkg.json in my opinion.

pet-mit: thanks for the tip!

pet-mit: actually both cucumber-cpp and the tests/cucumber sources (AntaresSteps.cpp) depend on GTest

src/ports/cucumber-cpp/vcpkg.json
pet-mit: I think it's important to list all the dependencies of cucumber-cpp in its port file, even if they are already installed by the parent vcpkg.json, don't you think ? this would make it more portable (but I don't see it in vcpkg.json)

flomnes: What I mean is that this dependency appears twice in the same file, which is prob a mistake

pet-mit: ah you're right haha!

src/tests/cucumber_python/features/steps/antares_steps.py
```diff
@@ -0,0 +1,82 @@
...
+        assert_double_close(float(row["MAX"]), get_annual_system_cost(context)["MAX"], 0.001)
+
+@then('the simulation takes less than {seconds} seconds')
+def check_simu_time(context, seconds):
+    actual_simu_time = parse_simu_time(context.output_path)
+    assert actual_simu_time <= float(seconds)
+
+@then('in area "{area}", during year {year}, loss of load lasts {lodl_hours} hours')
+def check_lodl_duration(context, area, year, lodl_hours):
+    assert int(lodl_hours) == get_hourly_values(context, area.lower(), int(year))["LOLD"].sum()
```
flomnes: LOLD

src/tests/cucumber/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+file(WRITE "${CMAKE_CURRENT_SOURCE_DIR}/latest_binary_dir.txt" "${CMAKE_BINARY_DIR}")
```
a-zakir:
```suggestion
file(GENERATE OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/conf.yaml CONTENT "antares-solver : $<TARGET_FILE:antares-solver>")
```

a-zakir: this will produce a yaml file with a key-value pair :
antares-solver : /path/to/antares-solver

src/tests/cucumber/features/steps/simulator_utils.py
```diff
@@ -0,0 +1,51 @@
...
+def get_solver_path():
+    file = open('latest_binary_dir.txt', 'r')
+    if os.name == 'nt':
+        exec_name = "antares-solver.exe"
+    else:
+        exec_name = "antares-solver"
+    solver_path = os.path.join(file.readline(), "solver", exec_name)
+    solver_full_path = str(Path(solver_path).resolve())
+    return solver_full_path
+
```
a-zakir:
```suggestion
def get_solver_path():
  #config_file = path to conf.yaml
    with open(config_file) as file:
            content = yaml.full_load(file)
    return content.get("antares-solver")

```

src/tests/cucumber/CMakeLists.txt
pet-mit: thanks !

a-zakir:
great work! i think the main problem in the project is the manner we launch e2e tests: download, unzip && test;
in the other hand this has the benefit to not clone large repo (antares_test_nr). i don't know what is the best choice:
1/ clone partially the test repo 
2/ download test archives 
both ways could be combined with the test tag (short, medium etc..)

.github/workflows/ubuntu.yml
```diff
@@ -231,6 +235,11 @@ jobs:
           batch-name: short-tests
           os: ${{ env.os }}
 
+      - name: Run cucumber on short-tests
+        uses: ./.github/workflows/cucumber-tests
+        with:
+          feature: "features/short_tests.feature"
```
a-zakir, line 241: here partial clone Antares_Simulator_Tests_NR ( only short tests sub dir)

.github/workflows/ubuntu.yml
```diff
@@ -273,6 +282,11 @@ jobs:
           batch-name: medium-tests
           os: ${{ env.os }}
 
+      - name: Run cucumber on medium-tests
+        uses: ./.github/workflows/cucumber-tests
+        with:
+          feature: "features/medium_tests.feature"
```
a-zakir, line 288: same

.github/workflows/ubuntu.yml
a-zakir: or to benchmark whether an archive is suitable ( something like [simtest](https://github.com/AntaresSimulatorTeam/SimTest/releases/tag/v9.2.0c) assets but without reference-results

src/tests/cucumber/features/steps/output_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    annual_system_cost = {}
+    for line in file.readlines():
+        for key in keys:
+            if key in line:
+                annual_system_cost[key] = float(line.split(key + " : ")[1])
+    return annual_system_cost
+
+
+def parse_simu_time(output_path: str) -> float:
+    file = open(os.path.join(output_path, "simulation.log"), 'r')
```
a-zakir: ther is maybe straightforward way to check some data, like in output/ info.antares-output for execution time (not sure)

a-zakir: or in execution_info.ini

.github/workflows/ubuntu.yml
pet-mit: I understand your point of view. However, I am reluctant to create an implicit adherence between the feature file naming (or tags) and the structure of the Antares_Simulator_Tests_NR file structure. I think it can possibly lead to confusion, especially since no explicit rule exists to prevent, for example, running a test under "medium-tests" from the "short_test" feature file.

a-zakir (approved):

PR #2352, opened 2024-08-22
Split expression-lib tests 

a-zakir:


flomnes (commented):
May I suggest one unit translation per test suite ?

src/tests/src/solver/expressions/module_test_declaration.cpp
flomnes, line 1: test_main.cpp ?

flomnes (approved):

PR #2358, opened 2024-08-26
Test improvement trial

guilpier-code:
This is a suggestion for a unit tests enhancement.
Unit tests are a bit simpler like this.
The main idea is to introduce a **fixture** containing more things than just a **registry**. 
What do you think ?


a-zakir:
nice !

flomnes (approved):

PR #2354, opened 2024-08-22
sonar issues from #2290

a-zakir:

PR #2362, opened 2024-08-27
optimize trytype loop

a-zakir:
The current dispatch method iterates through a list of functions **trytype**, sequentially checking each one to see if it can handle the given node type. This can be inefficient.
in this PR we use unordered_map as look up table between node type and trytype function:
less comparisons && faster access than sequential iteration


src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -45,32 +47,46 @@ class NodeVisitor
...
+                                  LessThanOrEqualNode,
+                                  GreaterThanOrEqualNode,
+                                  NegationNode,
+                                  ParameterNode,
+                                  VariableNode,
+                                  LiteralNode,
+                                  PortFieldNode,
+                                  ComponentVariableNode,
+                                  ComponentParameterNode>();
+        if (auto ret = my_map.at(typeid(node))(node, *this, args...); ret.has_value())
```
flomnes: There are two possibilities for errors
- The map does not contain an entry
- The function was found, but returns `std::nullopt`

We should have a different error message for each case. In the first case, you must catch `std::out_of_range` exception thrown by `std::map::at`.

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -45,32 +47,46 @@ class NodeVisitor
...
+     *
+     * @tparam NodeTypes A variadic pack of node types to be included in the map.
+     * @return An `std::unordered_map` containing the associations between node types and their
+     * corresponding visitor functions.
+     */
+    template<class... NodeTypes>
+    std::unordered_map<std::type_index, FunctionT> NodesVisitList() const
+    {
+        std::unordered_map<std::type_index, FunctionT> my_map;
+        ([&] { my_map[typeid(NodeTypes)] = &tryVisit<R, NodeVisitor<R, Args...>, NodeTypes>; }(),
```
flomnes:
```suggestion
        ([&my_map, &NodeTypes] { my_map[typeid(NodeTypes)] = &tryVisit<R, NodeVisitor<R, Args...>, NodeTypes>; }(),
```

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -45,32 +47,46 @@ class NodeVisitor
...
-          &tryVisit<R, NodeVisitor<R, Args...>, NegationNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, ParameterNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, VariableNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, LiteralNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, PortFieldNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, ComponentVariableNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, ComponentParameterNode>};
-        for (auto f: allNodeVisitList)
+        const static std::unordered_map<std::type_index, FunctionT>
+          my_map = NodesVisitList<AddNode,
```
flomnes:
```suggestion
        const static auto nodeDispatchFunctions = NodesVisitList<AddNode,
```

flomnes (commented):
If you use `typeid` to determine the polymorphic type of node objects, you no longer need to check the result of `dynamic_cast` in templated function `tryVisit`.

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: i wrapped the the _if_ cond in a try/catch blocks.
if the  _if_ cond body is not computed an exception is raised
in the catch block, a message is printed and the original exception (which could originate from any part of the code) is re-thrown

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -32,47 +34,76 @@ namespace
...
+            using namespace std::string_literals;
+            // node type is unknown or not found in the map: Throw an exception
+            throw TheVisitorIsNotImplementedForYourNode("Visitor ("s + typeid(*this).name()
+                                                        + ") cannot handle nodes of type ("
+                                                        + node_type.name() + ").");
+        }
+        catch (std::exception&)
+        {
+            logs.error() << "Antares::Solver::Visitor: could not visit the node!";
+            throw;
```
flomnes: Please use shorter blocks with more specific exception types
```cpp
try
{
    my_map.at(...)
}
catch (const std::out_of_range& ex) 
{
// ...
}
```

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -32,47 +34,82 @@ namespace
...
+        std::unordered_map<std::type_index, FunctionT> nodeDispatchFunctions;
+        (
+          [&nodeDispatchFunctions] {
+              nodeDispatchFunctions[typeid(NodeTypes)] = &tryVisit<R,
+                                                                   NodeVisitor<R, Args...>,
+                                                                   NodeTypes>;
+          }(),
+          ...);
+        return nodeDispatchFunctions;
+    }
```
flomnes: This could be a free function (outside `NodeVisitor`)

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -45,37 +43,59 @@ class NodeVisitor
...
+        {
+            return nodeDispatchFunctions.at(typeid(node))(node, *this, args...);
+        }
+        catch (std::exception&)
         {
-            if (auto ret = f(node, *this, args...); ret.has_value())
-            {
-                return ret.value();
-            }
+            logs.error() << "Antares::Solver::Visitor: could not visit the node!";
```
flomnes, line 113:
```suggestion
            logs.error() << "Antares::Solver::Visitor: could not visit the node! (exception " << ex.what() << ")";
```

https://en.cppreference.com/w/cpp/error/exception/what

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -45,37 +43,59 @@ class NodeVisitor
...
+        std::unordered_map<std::type_index, FunctionT> nodeDispatchFunctions;
+        (
+          [&nodeDispatchFunctions] {
+              nodeDispatchFunctions[typeid(NodeTypes)] = &tryVisit<R,
+                                                                   NodeVisitor<R, Args...>,
+                                                                   NodeTypes>;
+          }(),
+          ...);
+        return nodeDispatchFunctions;
+    }
```
flomnes: You may move this out of class `NodeVisitor`

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: two variadic parameters, it will look like
template<class R, class... NodeTypes, class... Args> // not ok

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: not necessary, when the exception is re-thrown;

flomnes (approved):

flomnes:
Waiting for @guilpier-code 's approval

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -49,42 +75,43 @@ class NodeVisitor
...
+        {
+            return nodeDispatchFunctions.at(typeid(node))(node, *this, args...);
+        }
+        catch (std::exception&)
         {
-            if (auto ret = f(node, *this, args...); ret.has_value())
-            {
-                return ret.value();
-            }
+            logs.error() << "Antares::Solver::Visitor: could not visit the node!";
```
guilpier-code, line 113: About : 
```C++
logs.error() << "Antares::Solver::Visitor: could not visit the node!";
```
Could be interesting to know which node type is missing in the map. We don't have this information in the message.

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -30,15 +30,41 @@ namespace Antares::Solver::Nodes
...
-    return std::nullopt;
+    auto* x = dynamic_cast<const NodeT*>(&node);
+    return visitor.visit(*x, args...);
 }
 } // namespace
+template<class R, class... Args>
+class NodeVisitor;
+
+template<class R, class... Args>
+struct NodeDispatchFunctionsProvider
```
guilpier-code: Very clever, I mean it.
But the code here became quite complex (it was a pain to understand the mechanisms at work in this PR).

A justification for all this complexity is an expected better efficiency.
We just talked about this with @flomnes : we wonder if, given the limited size of the map, we would get a performance benefit from that new change.
Maybe we could measure it.

guilpier-code (changes requested):

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -49,42 +75,43 @@ class NodeVisitor
...
-          &tryVisit<R, NodeVisitor<R, Args...>, GreaterThanOrEqualNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, NegationNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, ParameterNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, VariableNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, LiteralNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, PortFieldNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, ComponentVariableNode>,
-          &tryVisit<R, NodeVisitor<R, Args...>, ComponentParameterNode>};
-        for (auto f: allNodeVisitList)
+        const static auto nodeDispatchFunctions = NodeDispatchFunctionsProvider<R, Args...>::
```
guilpier-code: **NodeDispatchFunctionsProvider** could be renamed **NodeVisitsProvider** more in related the underlying design pattern.

Same for **nodeDispatchFunctions** : ==> **nodeVisitList** ?

guilpier-code (changes requested):

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: yes, this will be done in [2354](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2354)

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: the number of node types is not large (nb_Node_type<20), but in a deep/large tree the benefit is to avoid nb_Node_type * nb_nodes [trytype loop ]

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -30,15 +30,41 @@ namespace Antares::Solver::Nodes
...
-    return std::nullopt;
+    auto* x = dynamic_cast<const NodeT*>(&node);
+    return visitor.visit(*x, args...);
 }
 } // namespace
+template<class R, class... Args>
+class NodeVisitor;
+
+template<class R, class... Args>
+struct NodeVisitsProvider
```
guilpier-code, line 44: **Question** :
To avoid too much complexity (essentially due to template, and the very powerful **variadic pack** C++ feature), could we not more simply have :
```C++
R dispatch(const Node& node, Args... args)
{
    static const std::unordered_map<std::type_index, FunctionT> nodeDispatchFunctions = 
    {
        {typeid(AddNode), &tryVisit<R, NodeVisitor<R, Args...>, AddNode>},
        {typeid(SubtractionNode), &tryVisit<R, NodeVisitor<R, Args...>, SubtractionNode>},
        {typeid(MultiplicationNode), &tryVisit<R, NodeVisitor<R, Args...>, MultiplicationNode>},
        ....
    }
    
    try
    {
        return nodeDispatchFunctions.at(typeid(node))(node, *this, args...);
    }
    catch (std::exception&)
    {
        logs.error() << "Antares::Solver::Visitor: could not visit the node!";
        throw;
    }
}
```
?
Would it even be possible ?

I suppose that the template **struct**, its template **NodesVisitList** function and the **variadic pack** system inside were meant to avoid **code duplication** in the **node type** that we can see in the code exerpt just above, and I can see that the goal is reached, but the price is a bit high in code complexity.

guilpier-code (changes requested):

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: yes that's exactly what is done by the compiler in NodeVisitsProvider::NodesVisitList

PR #2363, opened 2024-08-28
Remove manually allocated memory

payetvin:
#2244

### DIFF
--- a/src/libs/antares/study/area/list.cpp
+++ b/src/libs/antares/study/area/list.cpp
@@ -1335,7 +1335,7 @@ void AreaListEnsureDataSolarPrepro(AreaList* l)
       {
           if (!area.solar.prepro)
           {
-              area.solar.prepro = new Antares::Data::Solar::Prepro();
+              area.solar.prepro = std::make_unique<Antares::Data::Solar::Prepro>();
           }
       });
 }
--- a/src/libs/antares/study/area/list.cpp
+++ b/src/libs/antares/study/area/list.cpp
@@ -934,11 +934,10 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
             ret = area.hydro.prepro->validate(area.id) && ret;
         }
 
-        auto* hydroSeries = area.hydro.series;
         if (!options.loadOnlyNeeded || !area.hydro.prepro) // Series
         {
             buffer.clear() << study.folderInput << SEP << "hydro" << SEP << "series";
-            ret = hydroSeries->loadGenerationTS(area.id, buffer, studyVersion) && ret;
+            ret = area.hydro.series->loadGenerationTS(area.id, buffer, studyVersion) && ret;
         }
 
         if (studyVersion < StudyVersion(9, 1))
@@ -953,10 +952,10 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
         else
         {
             buffer.clear() << study.folderInput << SEP << "hydro" << SEP << "series";
-            ret = hydroSeries->LoadMaxPower(area.id, buffer) && ret;
+            ret = area.hydro.series->LoadMaxPower(area.id, buffer) && ret;
         }
 
-        hydroSeries->resizeTSinDeratedMode(study.parameters.derated,
+        area.hydro.series->resizeTSinDeratedMode(study.parameters.derated,
                                            studyVersion,
                                            study.usedByTheSolver);
     }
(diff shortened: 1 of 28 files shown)


src/libs/antares/study/area/list.cpp
```diff
@@ -934,11 +934,10 @@ static bool AreaListLoadFromFolderSingleArea(Study& study,
             ret = area.hydro.prepro->validate(area.id) && ret;
         }
 
-        auto* hydroSeries = area.hydro.series;
         if (!options.loadOnlyNeeded || !area.hydro.prepro) // Series
         {
             buffer.clear() << study.folderInput << SEP << "hydro" << SEP << "series";
-            ret = hydroSeries->loadGenerationTS(area.id, buffer, studyVersion) && ret;
+            ret = area.hydro.series->loadGenerationTS(area.id, buffer, studyVersion) && ret;
```
flomnes, line 940: No need to remove the alias, you could still have
```cpp
auto* hydroSeries = area.hydro.series.get();
// ...
ret = hydroSeries->loadGenerationTS(/* ... */);
```

src/libs/antares/study/study.cpp
```diff
@@ -979,22 +979,22 @@ bool Study::clusterRename(Cluster* cluster, ClusterName newName)
...
 {
-    areas.each([](Data::Area& area) { FreeAndNil(area.hydro.prepro); });
+    areas.each([](Data::Area& area) { area.hydro.prepro.reset(); });
 }
 
 void Study::destroyAllWindTSGeneratorData()
 {
-    areas.each([](Data::Area& area) { FreeAndNil(area.wind.prepro); });
+    areas.each([](Data::Area& area) { area.wind.prepro.reset(); });
 }
```
flomnes, line 998: You probably no longer need these functions. Memory will be released when `Study` will be released, probably just a little later than now.

The only possible change of behaviour I foresee is if `prepro` objects are re-used, in which case we want to reset them beforehand.

After looking at the code, it's exactly what happens, because of TS refresh. Since this feature is to be removed "soon", please just add a comment to remind us that these functions (and probably more...) should be removed when removing TS refresh.

src/solver/ts-generator/include/antares/solver/ts-generator/xcast/studydata.h
```diff
@@ -77,7 +77,7 @@ class StudyData final
     //! List of all areas (sub-set of the complete list)
     Data::Area::Vector localareas;
     //! Correlation coefficients for each month
-    const Matrix<float>* correlation[12];
+    std::array<Matrix<float>, 12> correlation;
```
flomnes, line 80: You lost a `const` qualifier here, the equivalent might be
```cpp
std::array<const Matrix<float>, 12> correlation;
```

src/solver/ts-generator/xcast/studydata.cpp
```diff
@@ -92,8 +66,8 @@ void StudyData::reloadDataFromAreaList(const Data::Correlation& originalCorrelat
         {
         case Data::Correlation::modeAnnual:
         {
-            auto* m = new Matrix<float>();
-            prepareMatrix(*m, originalCorrelation.annual);
+            auto m = Matrix<float>();
```
flomnes:
```suggestion
            Matrix<float> m;
```

src/solver/ts-generator/xcast/studydata.cpp
```diff
@@ -105,23 +79,16 @@ void StudyData::reloadDataFromAreaList(const Data::Correlation& originalCorrelat
         {
             for (uint realmonth = 0; realmonth != 12; ++realmonth)
             {
-                auto* m = new Matrix<float>();
+                auto m = Matrix<float>();
                 correlation[realmonth] = m;
```
flomnes:
```suggestion
                correlation[realmonth] = Matrix<float>();
```

src/libs/antares/study/area/list.cpp
payetvin: The alias is used only once and the length is not a lot different, useless imo

src/solver/ts-generator/include/antares/solver/ts-generator/xcast/studydata.h
payetvin: using const here doesn't makes a lot of sense, those matrix will be modified

flomnes (changes requested):
Please revert **include/antares/solver/variable** and open a new PR

flomnes (approved):

PR #2364, opened 2024-08-28
Properly set common data in API output when using playlist

JasonMarechal25:
Common data are only computed once.
In the current implementation this computation is done only for the week 1 of the year 1. However when using playlist the year 1 may not exists

This PR fix this issue by computing common data is no data exist yet.


flomnes:
@JasonMarechal25 Can you provide more context on this ?

src/api/SimulationObserver.cpp
```diff
@@ -30,13 +30,14 @@ namespace
...
-    Solver::ConstantDataFromAntares common_data;
-    if (year == 1 && week == 1)
+    std::optional<Solver::ConstantDataFromAntares> common_data;
+    bool translateCommonData = false;
+    static std::once_flag flag;
+    std::call_once(flag, [&translateCommonData]() { translateCommonData = true; });
+    if (translateCommonData)
     {
         common_data = translator.commonProblemData(problemeHebdo.ProblemeAResoudre.get());
     }
```
flomnes: ### Style/simplification

You can eliminate `bool translateCommonData` by making `common_data` a `static` variable.

```cpp
auto translate(const PROBLEME_HEBDO& problemeHebdo,
               std::string_view name,
               const Solver::HebdoProblemToLpsTranslator& translator)
{
    auto weekly_data = translator.translate(problemeHebdo.ProblemeAResoudre.get(), name);
    static std::optional<Solver::ConstantDataFromAntares> common_data;
    static std::once_flag flag;

    std::call_once(flag, [&]() {
        common_data = translator.commonProblemData(problemeHebdo.ProblemeAResoudre.get());
    });

    return std::make_pair(common_data, weekly_data);
}
```

src/api/private/SimulationObserver.h
```diff
@@ -46,6 +46,7 @@ class SimulationObserver: public Solver::Simulation::ISimulationObserver
     void notifyHebdoProblem(const PROBLEME_HEBDO& problemeHebdo,
                             int optimizationNumber,
                             std::string_view name) override;
+    bool shouldTranslateCommonData() const;
```
flomnes:
```suggestion
```

src/api/SimulationObserver.cpp
```diff
@@ -56,18 +57,18 @@ void SimulationObserver::notifyHebdoProblem(const PROBLEME_HEBDO& problemeHebdo,
     const unsigned int year = problemeHebdo.year + 1;
     const unsigned int week = problemeHebdo.weekInTheYear + 1;
     // common_data and weekly_data computed before the mutex lock to prevent blocking the thread
-    auto [common_data, weekly_data] = translate(problemeHebdo, name, translator, year, week);
-    std::lock_guard lock(mutex_);
-    if (year == 1 && week == 1)
+    auto [common_data, weekly_data] = translate(problemeHebdo, name, translator);
```
flomnes: ### Performance/nitpicking

I'm afraid a copy of `std::optional<Solver::ConstantDataFromAntares>` is made here. Since this structure contains a lot of `std::vector`s, this should be avoided in favor of move operations.

EDIT : I can confirm that it indeed happens, since adding the following ctor declarations in `ConstantDataFromAntares`'s definition results in a compilation error at this point.
```cpp
    ConstantDataFromAntares() = default;
    ConstantDataFromAntares(ConstantDataFromAntares&&) = default;
    ConstantDataFromAntares(const ConstantDataFromAntares&) = delete;
```

This is a separate issue though, and not a serious one if `setConstantData` is called only once.

flomnes (approved):

PR #2369, opened 2024-09-03
Better dependencies with cmake, antares matrix

payetvin:


src/libs/antares/series/series.cpp
```diff
@@ -31,6 +31,8 @@
 
 using namespace Yuni;
 
+#define HOURS_PER_YEAR 8760
```
flomnes: Use already defined constant from src/libs/antares/include/antares/antares/constants.h

flomnes (approved):

a-zakir:
target_link_libraries(series PUBLIC Antares::array antares-core) -->target_link_libraries(antares-core
        PUBLIC
   ...
            Antares::study
....
) -->target_link_libraries(study
        PUBLIC
        ...
        Antares::series
      ...
)

PR #2380, opened 2024-09-06
Add contribution guidelines

flomnes:
About data members, I like the `m_` prefix, since it makes it very easy to identify member variables.

It's not our current convention, and it would take some effort to apply it to our current code base.

[source](https://github.com/organicmaps/organicmaps/blob/master/docs/CPP_STYLE.md)

### DIFF
--- a/docs/developer-guide/6-Contributing.md
+++ b/docs/developer-guide/6-Contributing.md
@@ -10,13 +10,13 @@ Below are our specific (but not all!) exceptions to the Google's coding standard
 - We use `#pragma once` instead of the `#define` Guard in header files.
 - Includes are sorted and grouped by directory, there should be newlines between different directories.
 - Order of directories in includes: "current_dir/current_file.hpp", includes from other dirs sorted by dependencies (e.g. indexer, then coding, then base), "defines.hpp", C++ standard library headers, boost headers, 3party.
-- We ARE using C++ exceptions.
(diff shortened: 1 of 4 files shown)


docs/developer-guide/6-Contributing.md
```diff
@@ -0,0 +1,245 @@
+# C++ Style Guide
+
+In general, [Google's coding standard](https://google.github.io/styleguide/cppguide.html) is used, and we strongly encourage to read it.
+
+Below are our specific (but not all!) exceptions to the Google's coding standard:
+
+- All C++ code should conform to the C++20 standard.
+- We use `.cpp` and `.hpp` files, not `.cc` and `.h` (`.c` and `.h` are used for C code), in UTF-8 encoding.
```
a-zakir: Actually we don’t use .hpp but .h

docs/developer-guide/6-Contributing.md
```diff
@@ -0,0 +1,245 @@
...
+- File names are lowercase with underscores, like `file_reader.cpp`.
```
a-zakir: For future files?

docs/developer-guide/6-Contributing.md
pet-mit, line 1: we should unify this with "CONTRIBUTING.md" which is at the root of the repo

flomnes: done

pet-mit (approved):

PR #2381, opened 2024-09-09
Update deprecated actions

payetvin:


flomnes:
Please enable OL8 & CentOS7 on this PR to check if the updated versions work. dependabot should have suggested this change already, so I think we rejected this because these versions don't work.

flomnes (approved):

PR #2383, opened 2024-09-09
Modeleur 2.1: Lib for modeling objects [ANT-1877]

payetvin:


src/solver/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ add_subdirectory(utils)
 add_subdirectory(variable)
 add_subdirectory(modeler)
 add_subdirectory(expressions)
+add_subdirectory(model)
```
pet-mit: in my opinion, "model" vs "modeler" can be confusing

src/solver/model/include/antares/solver/model/model.h
```diff
@@ -0,0 +1,59 @@
...
+
+private:
+    std::string id_;
+    Expression objective_;
+
+    std::map<std::string, Parameter> parameters_;
+    std::map<std::string, Variable> variables_;
+
+    std::map<std::string, Constraint> constraints_;
+    std::map<std::string, Constraint> bindingConstraints_;
```
pet-mit: do we really need to separate constraints from "binding" constraints? 
I think that this is a risk of code duplication, since the expression visitors will be the ones to know if a constraint is binding (ie has a PortFiel in it)

src/solver/model/include/antares/solver/model/portType.h
```diff
@@ -0,0 +1,48 @@
...
+namespace Antares::Solver::Model
+{
+
+class PortField
+{
+private:
+    std::string name;
+};
+
+class PortType
```
pet-mit: why are PortField and PortType in the same header file, and Port and PortFieldDefinitaion in another one?

src/solver/model/include/antares/solver/model/portType.h
```diff
@@ -0,0 +1,48 @@
...
+#include <string>
+#include <vector>
+
+namespace Antares::Solver::Model
+{
+
+class PortField
+{
+private:
+    std::string name;
```
pet-mit: shouldn't it also have a ValueType?

src/solver/libModelObject/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ set(SRC_model
     include/antares/solver/model/constraint.h
     include/antares/solver/model/expression.h
     include/antares/solver/model/port.h
+    include/antares/solver/model/portField.h
```
pet-mit: rename the includes

pet-mit (changes requested):

src/solver/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ add_subdirectory(utils)
 add_subdirectory(variable)
 add_subdirectory(modeler)
 add_subdirectory(expressions)
+add_subdirectory(libModelObject)
```
guilpier-code, line 24: About directory **libModelObject** : 
Shouldn't this directory be a sub-directory of **modeler** ?

src/solver/libModelObject/CMakeLists.txt
guilpier-code: @pet-mit  : what do you mean ? Rename how ?

src/solver/libModelObject/include/antares/solver/model/model.h
```diff
@@ -0,0 +1,59 @@
...
+ * Defines a model that can be referenced by actual components.
+ * A model defines the behaviour of those components.
+ */
+class Model
+{
+public:
+    Model();
+    ~Model() = default;
+
+    void getAllConstraints();
```
guilpier-code: `void getAllConstraints();` : 

We cannot ask a getter to return **void**. 
It should return a list of constraints, obviously.

guilpier-code: **Very cosmetic** : 
`void getAllConstraints();` ==> `void getConstraints();` or even `void constraints();`, shorter, and not meaningless

src/solver/libModelObject/include/antares/solver/model/model.h
```diff
@@ -0,0 +1,59 @@
...
+
+private:
+    std::string id_;
+    Expression objective_;
+
+    std::map<std::string, Parameter> parameters_;
+    std::map<std::string, Variable> variables_;
+
+    std::map<std::string, Constraint> constraints_;
```
guilpier-code: About `std::map<std::string, Constraint> constraints_;`

Something could be clarified : 
- in this map, what the value represent is clear. But what do the keys represent ? Constraint names ?
  If so, this map should be named **names_to_constraints** (for example).
- Note that the constraint names is already a data member of class **Constraint**. So, to me, it feels like a redundancy : 
  - **either** : we have a map associating a name to a **Constraint**, but the **Constraint** does not contain a name, 
  - **or** : we have a vector of contraints, and each constraint knows its name.

But maybe all this is not yet decided / clarified

src/solver/libModelObject/include/antares/solver/model/parameter.h
```diff
@@ -0,0 +1,48 @@
...
+{
+public:
+    Parameter();
+    ~Parameter() = default;
+
+private:
+    std::string name_;
+    ValueType type_;
+    bool timeDependent_;
+    bool scenarioDependent_;
```
guilpier-code: For these 2 boolean data members, is there any default value, like **false** ?
```C++
bool timeDependent_ = false;
bool scenarioDependent_  false;
```

src/solver/libModelObject/CMakeLists.txt
pet-mit: from model to libModelObject, they are two separate things

src/solver/CMakeLists.txt
pet-mit: no, they are completely different : 
- modeler is how we model an optimization problem
- libModelObject is how we model the Antares business problem
- later, we'll have a bridge (ComponentFiller) between the two

guilpier-code: ok, got it

src/solver/libModelObject/include/antares/solver/model/parameter.h
payetvin: I don't know if it's been decided yet

payetvin: should be true

src/solver/libModelObject/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+project(Model)
+
+set(SRC_model
+    model.cpp
+
+    include/antares/solver/model/library.h
```
JasonMarechal25: path should be include/antares/solver/libModelObject for consistency

src/solver/libModelObject/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+project(Model)
```
JasonMarechal25: Project and folder should strive to have the same name

JasonMarechal25 (dismissed):

pet-mit (approved):

PR #2387, opened 2024-09-10
Add iterators on ASTs, allowing for loops, std::find/std::find_if, std::count/std::count_if, etc.

flomnes:


src/solver/expressions/iterators/pre-order.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+    Node* current = nodeStack.top();
+    nodeStack.pop();
+
+    const auto children = childrenLeftToRight(current);
+    // Push children in reverse order to process them in left-to-right order
+    for (auto it = children.rbegin(); it != children.rend(); ++it)
+    {
+        nodeStack.push(*it);
+    }
```
a-zakir:
```suggestion
   for (auto it : children | std::views::reverse)
    {
        nodeStack.push(it);
    }
```

flomnes: It works but it makes the code a bit uglier (IMO)
```cpp
    for (auto it : children | std::views::reverse)
    {
        nodeStack.push(&*it);
    }
```

a-zakir: &* is for what?

flomnes: You're right, it's not necessary

a-zakir (approved):

PR #2391, opened 2024-09-11
API modeler - 1.1.b

guilpier-code:
The goal of this PR is to introduce and unit test class  **LinearProblemBuilder**.
This class is an orchestration class, and therefore makes other classes collaborate.
What was done : 
- Remove **LinearProblemBuilder::solve()** : 
  this method was planned in the specifications, but it extends the builder responsibility scope.
  Once modified by the builder, the problem can be solved by calling its own **solve()** method.
- Add methods **numVariables()** and **numConstraints()** to base class **ILinearProblem** for unit tests reason.
- **CMakeLists.txt** file for modeler API was split into 2 **\*.cmake** files
- **Trial for a dependency simplification** :
  In the original / specified architecture, **LinearProblemBuilder** collaborates with classes : 
  - **LinearProblemData**
  - **ILinearProblem** (interface for LP)
  - **LinearProblemFiller** : this class also depends on previous classes

  Here, we allowed ourselves to reduce dependencies of this builder class (is this simplification relevant ?) :
class **LinearProblemBuilder** currently only depends on **LinearProblemFiller**, while fillers keeps its planned dependencies on **LinearProblemData** and **ILinearProblem**.
Indeed, any filler  is in charge to modify problem (using LP data),  the but the builder only manipulates fillers.
- **Mocking fillers** : fillers are the entities responsible for creating variables and constraints in the linear problem. To do that, they need the LP itself as well as the LP data (see fillers constructor). We created an abstract base filler. All concrete filler inherit from it. Class **LinearProblemBuilder** is constructed from a vector of fillers.
- Several unit tests were created for class **LinearProblemBuilder**


src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
```diff
@@ -30,9 +32,19 @@ namespace Antares::Solver::Modeler::Api
...
-    virtual void addObjectiveCoefficients(LinearProblem* problem, LinearProblemData* data) = 0;
+    explicit LinearProblemFiller(std::shared_ptr<ILinearProblem> pb, LinearProblemData& data):
+        linearProblem_(pb),
+        LPdata_(data)
+    {
+    }
+
+    virtual void addVariables() = 0;
+    virtual void addConstraints() = 0;
+    virtual void addObjective() = 0;
```
pet-mit: I think we better keep the old signatures in order to allow the same fillers to be used to fill multiple matrices (LinearProblems), for example the same thermal filler can be used to fill multiple 1-week problems

src/tests/src/solver/modeler/api/mock-fillers/OneConstraintFiller.cpp
pet-mit, line 1: since it is a mock, if it suits you, I suggest we put the code in the .h file for simplicity

pet-mit (changes requested):

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblem.h
```diff
@@ -47,10 +47,12 @@ class ILinearProblem
     /// Create a integer variable
     virtual IMipVariable* addIntVariable(double lb, double ub, const std::string& name) = 0;
     virtual IMipVariable* getVariable(const std::string& name) const = 0;
+    virtual int numVariables() const = 0;
```
a-zakir: getVariableCount()
or
variableCount()

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblem.h
```diff
@@ -47,10 +47,12 @@ class ILinearProblem
...
 
     /// Add a bounded constraint to the problem
     virtual IMipConstraint* addConstraint(double lb, double ub, const std::string& name) = 0;
     virtual IMipConstraint* getConstraint(const std::string& name) const = 0;
+    virtual int numConstraints() const = 0;
```
a-zakir: getConstraintCount()
or
constraintCount()

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblem.h
guilpier-code: ok. Done.

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblem.h
guilpier-code: Done.

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemBuilder.h
```diff
@@ -24,17 +24,18 @@
...
 namespace Antares::Solver::Modeler::Api
 {
 
 class LinearProblemBuilder
 {
 public:
-    virtual void LinearProblemBuilder(std::vector<LinearProblemFiller*> fillers) = 0;
-    virtual void build(LinearProblemData* data) = 0;
-    virtual MipSolution* solve() = 0;
+    LinearProblemBuilder(const std::vector<std::shared_ptr<LinearProblemFiller>>& fillers);
```
a-zakir: the use of shared_ptr is questionnable

guilpier-code: Could you be more specific ?
Why is it a bad practice ?
What would you suggest instead ?

src/tests/src/solver/modeler/api/mock-fillers/OneConstraintFiller.cpp
guilpier-code: Done.

src/tests/src/solver/modeler/api/CMakeLists.txt
```diff
@@ -1,20 +1,2 @@
...
-)
-
-# Storing tests-ts-numbers under the folder Unit-tests in the IDE
-set_target_properties(${EXECUTABLE_NAME} PROPERTIES FOLDER Unit-tests)
-
-add_test(NAME test-modeler-api-ortools COMMAND ${EXECUTABLE_NAME})
-
-set_property(TEST test-modeler-api-ortools PROPERTY LABELS unit)
+include(LinearProblemWithOrtools.cmake)
+include(LinearProblemBuilder.cmake)
```
a-zakir: this will produce two executable, one for each .cmake, i think it's not necessary;
having more exe to build as we add test is too costly, consider 1 exe for multiple unit translations

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemBuilder.h
a-zakir: if the ownership of LinearProblemFiller is managed externally , you could simply use raw pointer or if you which to ensure non-nullity of LinearProblemFiller use ref

a-zakir: | Feature                     | `shared_ptr<LinearProblemFiller>`      | `LinearProblemFiller*`               | `LinearProblemFiller&`                |
|-----------------------------|----------------------------------------|--------------------------------------|---------------------------------------|
| **Memory Management**        | Automatic (via reference counting)     | Manual (requires `new`/`delete`)     | Manual (no ownership semantics)       |
| **Ownership**                | Shared ownership                      | No ownership                         | No ownership                          |
| **Null-Safety**              | Can be null                           | Can be null                          | Cannot be null                        |
| **Memory Overhead**          | Extra overhead for reference counting | No overhead                          | No overhead                           |
| **Lifetime Guarantee**       | Managed, object destroyed when last owner releases | External management                 | External management                   |
| **Preferred Use Case**       | When multiple owners exist or unsure of lifetime management | When you know the lifetime management outside the builder | When fillers are guaranteed to outlive the builder |

a-zakir: the use of new/delete is not mandatory, for ex if LinearProblemFiller is obtained from a unique_ptr::get()

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
```diff
@@ -30,9 +32,19 @@ namespace Antares::Solver::Modeler::Api
...
+
+protected:
```
a-zakir: why protected? for tests?

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemBuilder.h
guilpier-code: ok. Done.

guilpier-code: It was also done when handling the LP (linear problem) inside the LP fillers

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
guilpier-code: For now, all LP fillers should keep a reference to the LP (as well as to LP data).
To me, the best way so that they all contain such references is to have their base class containing it.
According to Sonar, **protected** data members are usually a bad practice, because it introduces too much coupling between a class and its children.
However, when a class hierarchy is only 2 leveled, and meant to use polymorphism, like here, I guess that this way to have common data members among classes is not troublesome.

Do you see a better way to have code in common ? (I would be interested)

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
```diff
@@ -30,9 +32,19 @@ namespace Antares::Solver::Modeler::Api
...
+    ILinearProblem& linearProblem_;
+    const LinearProblemData& LPdata_;
```
a-zakir: this class should stay an interface without member data;
I understand the need of these members, as an alternative put them in the builder

a-zakir: and without ctor of course

guilpier-code: Could you explain your point of view ?

@pet-mit asks for the same changes in [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2391#discussion_r1757090621), and justifies it by saying : what if a filler has to change of LP ? For example when a new weekly optimization LP is required in **Antares**. 
I'm about to take care of this.

That said : 
The code as it currently is makes sense to me : fillers are the **only** objects in charge of adding variables and constraints to problems. So it seems natural they contain a reference to the **LP** and the **LP data**.
Furthermore, there are drawbacks in the fact that fillers do not have the **LP** and the **LP data** as data members : 
- we must pass the **LP** and the **LP data** each time we call a filler's method, so the code gets more complicated.
- **builder** gets necessarily new dependencies to class **ILinearProblem** and **LinearProblemData**, so it leads to more dependencies, which is not recommended by code quality.

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
guilpier-code: ok.
As said in this PR's presentation and more accurately in [this (later) comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2391#discussion_r1758716311), there are drawbacks in doing what you ask : for dependency and simplicity reasons.

A possible solution to this could be to add a new concrete method **updateLinearProblem(...)** to fillers interface : 
```C++
class LinearProblemFiller
{
public:
    explicit LinearProblemFiller(ILinearProblem& pb, LinearProblemData& data):
        linearProblem_(pb),
        LPdata_(data)
    {
    }

    void updateLinearProblem(ILinearProblem& pb); // New method
   
    virtual void addVariables() = 0;
    virtual void addConstraints() = 0;
    virtual void addObjective() = 0;

protected:
    ILinearProblem& linearProblem_;
    const LinearProblemData& LPdata_;
};
```
Would it make sense to you ?

pet-mit: I agree with @a-zakir 's comment that this should stay an interface and contain no attributes, to better respect the single-responsibility pattern.
As for the drawbacks : 
- the three fillers' methods will only be called once in the builder, so it's not that big a drawback for them to have 2 arguments
- the builder actually is supposed to be the orchestrator of the LinearProblem and the Fillers (and that's its only purpose), so I'm willing to accept the dependencies

guilpier-code: ok. Done.

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
guilpier-code: Class **LinearProblemFiller** no longer has data member.

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
guilpier-code: This class no longer has data members.

src/tests/src/solver/modeler/api/CMakeLists.txt
guilpier-code: When you say "too costly", what do you mean ?
And if you're right, which unit tests policy should we apply ?
For example, should we necessarily regroup unit tests of a sub directory in an executable ?
At which level of sub directory do we stop gathering unit tests ?

We should discuss about this in a team meeting, and identify what we want to do exactly : Is it really suitable ? Are there any benefits or drawbacks ? Should we generalize this ?

pet-mit (approved):

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
```diff
@@ -30,9 +32,10 @@ namespace Antares::Solver::Modeler::Api
 class LinearProblemFiller
 {
 public:
-    virtual void addVariables(LinearProblem* problem, LinearProblemData* data) = 0;
-    virtual void addConstraints(LinearProblem* problem, LinearProblemData* data) = 0;
-    virtual void addObjectiveCoefficients(LinearProblem* problem, LinearProblemData* data) = 0;
+    explicit LinearProblemFiller() = default;
```
a-zakir: seems unused?

a-zakir:
you have left two minors code smell

guilpier-code:
> you have left two minors code smell

src/solver/modeler/api/include/antares/solver/modeler/api/linearProblemFiller.h
guilpier-code: **LinearProblemFiller()** removed

guilpier-code:
> you have left two minors code smell

Taken care of

src/tests/src/solver/modeler/api/CMakeLists.txt
a-zakir: generating 2 exe take more times and memory than one exe, right?
multiply this for every lib we would create:
lib_test
 |_ sublib1_exe
 |_ sublib2_exe                      VS     lib_test_exe
 : 
 |_ sublibn_exe

a-zakir (approved):

flomnes (approved):

PR #2395, opened 2024-09-12
Fix mingen issues with several modifications

flomnes:
1. [X] Remove truncation from weekly turbine credits (was done previously)
2. [X] Always enable monitoring of stock levels (including overflow)
3. [X] Take overflow into account in the "HydroPower" constraint (was done in 9.1.0)
4. [X] Remove recomputation of reservoir levels & overflow 
5. [X] Correct infeasibility in hydro heuristic problem, related to mingen > 0


src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -572,6 +584,14 @@ bool PartHydro::SaveToFolder(const AreaList& areas, const AnyString& folder)
           {
               allSections.sPowerToLevel->add(area.id, true);
           }
+          if (area.hydro.overflowCost)
+          {
+              allSections.sOverflowCost->add(area.id, area.hydro.overflowCost);
+          }
+          if (area.hydro.levelCost)
+          {
+              allSections.sLevelCost->add(area.id, area.hydro.overflowCost);
```
nikolaredstork: Should in this line need to be levelCost instead of overflowCost?

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -152,6 +152,9 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
 
         problem.CoutDeDefaillanceNegative[i] = area.thermal.spilledEnergyCost;
```
guilpier-code, line 168: It seems that removing the weekly truncation involves removing the whole block  : 
```C++
double weekGenerationTarget = 1.;
double marginGen = 1.;

if (area.hydro.reservoirManagement && area.hydro.useHeuristicTarget
    && not area.hydro.useLeeway)
{
     double weekTarget_tmp = 0.;
     for (uint j = 0; j < 7; ++j)
     {
          uint day = study.calendar.hours[PasDeTempsDebut + j * 24].dayYear;
         weekTarget_tmp += hydroVentilationResults[k]
                        HydrauliqueModulableQuotidien[day];
     }

     if (weekTarget_tmp != 0.)
     {
         weekGenerationTarget = weekTarget_tmp;
     }
     marginGen = weekGenerationTarget;
}
```
As well as the truncation coefficient `marginGen / weekGenerationTarget`.

guilpier-code (changes requested):

.github/workflows/windows-vcpkg.yml
```diff
@@ -25,8 +25,8 @@ on:
 env:
   GITHUB_TOKEN: ${{ github.token }}
   IS_RELEASE: ${{ github.event_name == 'workflow_dispatch' }}
-  RUN_SIMPLE_TESTS: ${{ github.event_name == 'push' || inputs.run-tests == 'true' }}
-  RUN_EXTENDED_TESTS: ${{ github.event_name == 'schedule' || inputs.run-tests == 'true' }}
+  RUN_SIMPLE_TESTS: 'false'
+  RUN_EXTENDED_TESTS: 'false'
```
guilpier-code, line 29: Whatever the purpose of this change is (it could be so that CI completes and generates artefacts), we'll have not to forget reverting it.

src/solver/simulation/sim_calcul_economique.cpp
flomnes: You mean that since `marginGen = weekGenerationTarget`, `marginGen / weekGenerationTarget` is always 1 ?

guilpier-code: Exactly. 
This code is a piece of something (credit troncation) that once existed but no longer does.
It was not removed completely.
And this rest has no effect on the code

guilpier-code (changes requested):

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -330,6 +332,12 @@ bool PartHydro::LoadFromFolder(Study& study, const AnyString& folder)
               && ret;
     }
 
+    if (IniFile::Section* section = ini.find("overflow cost"))
+    {
+        ret = loadProperties(study, section->firstProperty, buffer, &PartHydro::overflowCost)
+              && ret;
+    }
+
     return ret;
```
nikolaredstork, line 339: I assume that levelCost also need to be loaded. If not, levelCost will be always 0, since it is assigned to 0 in class definition.

src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -89,6 +89,8 @@ void PartHydro::reset()
     allocation.clear();
     // allocation.fromArea(<current area>, 1.); // Area::reset()
 
+    overflowCost = 0.;
+
```
nikolaredstork: Should levelCost also have to be reset to 0 in `reset` function?

flomnes: a67c0894d

src/libs/antares/study/parts/hydro/container.cpp
flomnes: 4189a3e96

src/solver/simulation/sim_calcul_economique.cpp
guilpier-code: Exactly.
When we removed the troncation, we did not remove all what should have been.

src/solver/hydro/daily/h2o_j_initialiser_les_bornes_des_variables.cpp
```diff
@@ -53,7 +53,7 @@ void H2O_J_InitialiserLesBornesdesVariables(DONNEES_MENSUELLES* DonneesMensuelle
         int Var = CorrespondanceDesVariables.NumeroDeVariableTurbine[Pdt];
         Xmax[Var] = TurbineMax[Pdt];
 
-        Xmin[Var] = std::min(TurbineMax[Pdt], std::max(TurbineCible[Pdt], TurbineMin[Pdt]));
+        Xmin[Var] = std::min(TurbineMax[Pdt], TurbineMin[Pdt]);
```
guilpier-code, line 56: So we no longer need to use `TurbineCible` in this function **H2O_J_InitialiserLesBornesdesVariables(...)**.
Hence we should remove line : 
```C++
const std::vector<double>& TurbineCible = DonneesMensuelles->TurbineCible;
```

guilpier-code, line 56: About : 
```C++
Xmax[Var] = TurbineMax[Pdt];
Xmin[Var] = std::min(TurbineMax[Pdt], TurbineMin[Pdt]);
```
It's strange that we give the lower bound of variable **Turbine** like this.
Reader could expect instead : 
```C++
Xmin[Var] = TurbineMin[Pdt];
Xmax[Var] = TurbineMax[Pdt];
```
It seems that **TurbineMax** is not necessarily greater than **TurbineMin** for every day, which is strange.

src/solver/optimisation/adequacy_patch_csr/adq_patch_post_process_list.cpp
```diff
@@ -51,8 +51,6 @@ AdqPatchPostProcessList::AdqPatchPostProcessList(const AdqPatchParams& adqPatchP
     // Here a post process particular to adq patch
     post_process_list.push_back(
       std::make_unique<DTGmarginForAdqPatchPostProcessCmd>(problemeHebdo_, areas, thread_number));
-    post_process_list.push_back(
```
guilpier-code, line 55: In the current PR presentation, it's said that the re-computation of hydro levels after weekly optimization should be removed.
And indeed, the post process class **HydroLevelsUpdatePostProcessCmd** was largely simplified : it now only converts levels from their adimensionnel version into **MWh**.
In that context, we can also see that this post-process is still called **twice** for modes **adequacy** & **economy**, which leads to wrong hydro levels values.

What we should do : 
- This class should therefore be renamed into **HydroLevelsConversionIntoMWhCmd** for more clarity.
- Its method **execute(...)** should be called once (for a given simulation mode).
- We should take advantage of this to remove boolean **remixWasRun**, no longer useful.
- If it turns out the other boolean **computeAnyway** is not really needed, then remove it.

guilpier-code (changes requested):

src/solver/hydro/daily/h2o_j_initialiser_les_bornes_des_variables.cpp
nikolaredstork: I agree, since there is a precheck for each hour in `checkGenerationPowerConsistency` function, I don't see how TurbineMax in any situation can be lower than TurbineMin, but maybe I am wrong.

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -181,17 +185,13 @@ void SIM_InitialisationProblemeHebdo(Data::Study& study,
                                                                               .useHeuristicTarget
                                                                            || area.hydro.useLeeway);
 
-        problem.CaracteristiquesHydrauliques[i].SuiviNiveauHoraire
-          = area.hydro.reservoirManagement && (problem.OptimisationAuPasHebdomadaire)
-            && (!area.hydro.useHeuristicTarget
-                || problem.CaracteristiquesHydrauliques[i].PresenceDePompageModulable);
+        problem.CaracteristiquesHydrauliques[i].SuiviNiveauHoraire = area.hydro.reservoirManagement;
```
guilpier-code, line 202: About **SuiviNiveauHoraire** : if this boolean equals **reservoirManagement**, let's call it  **reservoirManagement** and avoid a parameter to be named in two different ways.
It would lower complexity.

src/solver/optimisation/opt_decompte_variables_et_contraintes.cpp
```diff
@@ -132,16 +132,16 @@ int OPT_DecompteDesVariablesEtDesContraintesDuProblemeAOptimiser(PROBLEME_HEBDO*
...
 
-        if (!Pump && !TurbEntreBornes && !MonitorHourlyLev
+        if (!Pump && !TurbEntreBornes
             && problemeHebdo->CaracteristiquesHydrauliques[pays].PresenceDHydrauliqueModulable)
         {
             ProblemeAResoudre->NombreDeContraintes++;
         }
 
-        if (Pump && !TurbEntreBornes && !MonitorHourlyLev)
+        ProblemeAResoudre->NombreDeContraintes += nombreDePasDeTempsPourUneOptimisation;
```
guilpier-code, line 140: About : 
```C++
ProblemeAResoudre->NombreDeContraintes += nombreDePasDeTempsPourUneOptimisation;
```
This instruction suddenly appears. 
Why ? Was the previous code incorrect ?
Or is it about overflows back in the game ?
We should add a comment.

guilpier-code, line 140: This file contains no comment that relates a count increment to a set of constraints.
It's unclear : which increment correspond to which set of constraints ? 
Besides, this way of counting constraints apart from the constraints themselves is very much error prone : it's very easy to add / remove constraints and not update the constraints count.
Adding / removing a constraint should automatically update the constraint count.

But this will be for another PR.

src/solver/optimisation/opt_gestion_des_couts_cas_lineaire.cpp
```diff
@@ -275,26 +275,19 @@ void OPT_InitialiserLesCoutsLineaire(PROBLEME_HEBDO* problemeHebdo,
```
guilpier-code, line 292: We should update the comment that explains the cost we give to the hydro overflow

guilpier-code (changes requested):

src/solver/simulation/sim_calcul_economique.cpp
flomnes: I agree, if we go on with this change, we'll simply remove `SuiviNiveauHoraire`.

src/solver/optimisation/opt_decompte_variables_et_contraintes.cpp
flomnes: No the estimation of variables & constraints is incorrect. There was an underestimation, which lead to segfaults.

flomnes:
See #2599

src/solver/optimisation/opt_decompte_variables_et_contraintes.cpp
flomnes: The estimation was already wrong in develop, we make it a bit more wrong here

PR #2396, opened 2024-09-13
Replace AddNode with SumNode

pet-mit:
In expressions, replace "AddNode" that takes only 2 operands, with "SumNode" that takes 0 to N operands.
This will allow shallower trees.

PR #2399, opened 2024-09-16
visualize ast with graphviz

a-zakir:
output : 
out

legends box-style could match those of the nodes


src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::stringstream dot;
+    dot << "digraph ExpressionTree {\n";
+    dot << result_.str();
+    dot << "}\n";
+    return dot.str();
+}
+
+std::string AstGraphVisitor::visit(const Nodes::PortFieldNode* node)
+{
+    throw AstGraphVisitorNotImplemented(name(), node->name());
```
flomnes: return PF(XXX, YYY)

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+std::string AstGraphVisitor::visit(const Nodes::ComponentVariableNode* node)
+{
+    throw AstGraphVisitorNotImplemented(name(), node->name());
```
flomnes: return CV(XXX, YYY)

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+std::string AstGraphVisitor::visit(const Nodes::ComponentParameterNode* node)
+{
+    throw AstGraphVisitorNotImplemented(name(), node->name());
```
flomnes: return "CP(XXX, YYY)"

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <antares/solver/expressions/nodes/ExpressionsNodes.h>
+
+namespace Antares::Solver::Visitors
+{
+
+std::string AstGraphVisitor::visit(const Nodes::AddNode* node)
+{
+    processBinaryOperation(node, "+");
+    return result_.str();
+}
```
flomnes: minor conflict #2396 (`AddNode` is replaced with `SumNode`)

src/solver/expressions/include/antares/solver/expressions/visitors/AstGraphVisitor.h
```diff
@@ -0,0 +1,70 @@
...
+    std::string visit(const Nodes::ParameterNode* node) override;
+    std::string visit(const Nodes::LiteralNode* node) override;
+    std::string visit(const Nodes::PortFieldNode* node) override;
+    std::string visit(const Nodes::ComponentVariableNode* node) override;
+    std::string visit(const Nodes::ComponentParameterNode* node) override;
+
+    int getNodeID(const Nodes::Node* node);
+    void emitNode(int id, const std::string& label);
+    void processBinaryOperation(const Nodes::BinaryNode* node, const std::string& label);
+    std::stringstream result_;
```
flomnes: Do you think you can eliminate this and write directly on a user-given stream ?
```cpp
AstGraphVisitor visitor;
visitor.dispatch(rootExpr, std::cout);
```
In this case `std::cout` would be passed to every visitor as `std::ostream&`, as an additional parameter.

a-zakir: yes of course !

src/solver/expressions/include/antares/solver/expressions/visitors/AstGraphVisitor.h
```diff
@@ -0,0 +1,85 @@
...
+                  const std::string& color = "azure",
+                  const std::string& shape = "box",
+                  const std::string& style = "rounded");
+
+    void processBinaryOperation(const Nodes::BinaryNode* node,
+                                const std::string& label,
+                                const std::string& color = "azure",
+                                const std::string& shape = "box",
+                                const std::string& style = "rounded");
+    std::ostream* out_stream_ = &std::cout;
```
flomnes:
```suggestion
    std::ostream& out_stream_;
```

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,201 @@
...
+        nodeIds_[node] = ++nodeCount_;
+    }
+    return nodeIds_[node];
+}
+
+void AstGraphVisitor::emitNode(int id,
+                               const std::string& label,
+                               const std::string& color,
+                               const std::string& shape,
+                               const std::string& style)
```
flomnes:
```cpp
struct BoxProperties {
// 4 arguments
};
emitNode(int id, const BoxProperties& style);
```

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+void AstGraphVisitor::EndTreeGraph()
+{
+    *out_stream_ << "}\n";
+}
+
+void AstGraphVisitor::setOutStream(std::ostream* outStream)
+{
+    out_stream_ = outStream;
+}
```
flomnes:
```suggestion
```

src/solver/expressions/include/antares/solver/expressions/visitors/AstGraphVisitor.h
```diff
@@ -0,0 +1,85 @@
...
+#include "antares/solver/expressions/visitors/NodeVisitor.h"
+
+namespace Antares::Solver::Visitors
+{
+
+class AstGraphVisitorNotImplemented: public std::invalid_argument
+{
+public:
+    AstGraphVisitorNotImplemented(const std::string& visitor, const std::string& node);
+};
```
flomnes:
```suggestion
```

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+AstGraphVisitorNotImplemented::AstGraphVisitorNotImplemented(const std::string& visitor,
+                                                             const std::string& node):
+    std::invalid_argument("Visitor" + visitor + " not implemented for node type " + node)
+{
+}
+
+void AstGraphVisitor::NewTreeGraph(const std::string& tree_name)
+{
+    *out_stream_ << "digraph " + tree_name + " {\n";
```
flomnes: Do we care about setting this ID ? It seems that it is only necessary to handle multiple graphs.

src/solver/expressions/include/antares/solver/expressions/visitors/AstGraphVisitor.h
```diff
@@ -0,0 +1,85 @@
...
+
+class AstGraphVisitor: public NodeVisitor<void>
```
flomnes: 1st one is the return type, 2nd one is the argument type
```suggestion
class AstGraphVisitor: public NodeVisitor<void, std::ostream&>
```
So `AstGraphVisitor` can go stateless.

Then
```cpp
std::ostream& operator<<(std::ostream& os, std::pair<AstGraphVisitor&, Node*> visitorExpr)
{
    beginTree(os);
    auto& [visitor, root] = visitorExpr;
    visitor.dispatch(root, os);
    endTree(os);
    return os;
}

// Caller
AstGraphVisitor visitor;
Node* root = fakeTree();
std::cout << std::pair(visitor, root) << std::endl;
```

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -26,103 +26,106 @@
...
+        os << "  " << id << " -> " << childId << ";\n";
+        dispatch(child, os);
     }
 }
 
-void AstGraphVisitor::visit(const Nodes::SubtractionNode* node)
+void AstGraphVisitor::visit(const Nodes::SubtractionNode* node, std::ostream& os)
 {
-    processBinaryOperation(node, "-", "moccasin", "oval", "filled, rounded");
+    processBinaryOperation(node, {"-", "moccasin", "oval", "filled, rounded"}, os);
```
flomnes: Make a few `constexpr BoxStyle`s instead of mixing code and design ;)

src/solver/expressions/visitors/AstGraphVisitor.cpp
```diff
@@ -0,0 +1,191 @@
...
+    os << "digraph " + tree_name + " {\n";
+    os << "node[style = filled]\n";
+}
+
+void AstGraphVisitor::EndTreeGraph(std::ostream& os)
+{
+    os << "}\n";
+}
+
+std::ostream& operator<<(std::ostream& os, std::pair<AstGraphVisitor&, Nodes::Node*>& visitorExpr)
```
flomnes:
```suggestion
std::ostream& operator<<(std::ostream& os, const std::pair<AstGraphVisitor&, Nodes::Node*>& visitorExpr)
```

src/tests/src/solver/expressions/test_AstGraphVisitor.cpp
```diff
@@ -0,0 +1,118 @@
...
+    Node* literalNode2 = create<LiteralNode>(53.);
+    Node* lessThanOrEqualNode = create<LessThanOrEqualNode>(literalNode2, equalNode);
+    Node* literalNode3 = create<LiteralNode>(54.);
+    Node* greaterThanOrEqualNode = create<GreaterThanOrEqualNode>(literalNode3,
+                                                                  lessThanOrEqualNode);
+
+    std::ofstream out("out.dot");
+    std::vector<std::ostream*> ostreams{&out, &std::cout};
+    std::stringstream redirectedStdout;
+    std::streambuf* initialBufferCout = std::cout.rdbuf(redirectedStdout.rdbuf());
```
flomnes: Why do you need to test `std::cout` ? You could write something as simple as
```cpp
std::ostringstream os;
os << std::pair(visitor, expr);
BOOST_CHECK_EQUAL(os.str(), expected_dot);
```

You probably don't need `rdbuf` and `std::streambuf`. All the tests can be done in memory, no need for disk I/O (C++ streams FTW)

src/solver/expressions/include/antares/solver/expressions/visitors/AstGraphVisitor.h
```diff
@@ -0,0 +1,82 @@
...
+
+struct BoxStyle
+{
+    std::string label;
+    std::string color = "azure";
+    std::string shape = "box";
+    std::string style = "rounded";
+};
+
+class AstGraphVisitor: public NodeVisitor<void, std::ostream&>
```
flomnes: This name isn't very explicit, why not `DOTStreamVisitor` ?

src/solver/expressions/include/antares/solver/expressions/visitors/AstGraphVisitor.h
```diff
@@ -0,0 +1,82 @@
...
+namespace Antares::Solver::Visitors
+{
+
+struct BoxStyle
+{
+    std::string label;
+    std::string color = "azure";
+    std::string shape = "box";
+    std::string style = "rounded";
+};
```
flomnes: Add a few `constexpr` instances

```suggestion
struct BoxStyle
{
    const std::string color;
    const std::string shape;
    const std::string style;
};
```
Place the `label` outside the `struct`.

src/tests/src/solver/expressions/test_AstGraphVisitor.cpp
a-zakir: Because you have mentioned std::cout !
And I don’t see why it’s bothering you 🙂

src/solver/expressions/include/antares/solver/expressions/visitors/AstGraphVisitor.h
a-zakir: Do you really think that DOTStream is more explicit than AstGraph?
I would say AstGraphDotStyle

flomnes: Yes, but I think DOT must be capitalized entirely (on official docs it's DOT)

a-zakir: ok

src/solver/expressions/include/antares/solver/expressions/visitors/AstDOTStyleVisitor.h
```diff
@@ -0,0 +1,83 @@
...
+    void emitNode(int id, const std::string& label, const BoxStyle& box_style, std::ostream& os);
+
+    void processBinaryOperation(const Nodes::BinaryNode* node,
+                                const std::string& label,
+                                const BoxStyle& box_style,
+                                std::ostream& os);
+
+private:
+    std::unordered_map<const Nodes::Node*, int> nodeIds_; // Mapping to store unique IDs for nodes
+    int nodeCount_ = 0;                                   // Counter to assign unique node IDs
```
flomnes:
```suggestion
    std::unordered_map<const Nodes::Node*, unsigned int> nodeIds_; // Mapping to store unique IDs for nodes
    unsigned int nodeCount_ = 0;                                   // Counter to assign unique node IDs
```

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,204 @@
...
+static constexpr BoxStyle SumStyle{"aqua", "hexagon", "filled, solid"};
+static constexpr BoxStyle BinaryStyle{"aqua", "oval", "filled, rounded"};
+static constexpr BoxStyle ComparisonStyle{"yellow", "diamond", "filled"};
+static constexpr BoxStyle NegationStyle{"tomato", "invtriangle", "filled, solid"};
+static constexpr BoxStyle LiteralStyle{"lightgray", "box", "filled, solid"};
+static constexpr BoxStyle VariableStyle{"gold", "box", "filled, solid"};
+static constexpr BoxStyle ParameterStyle{"wheat", "box", "filled, solid"};
+static constexpr BoxStyle ComponentParameterStyle{"springgreen", "octagon", "filled, solid"};
+static constexpr BoxStyle ComponentVariableStyle{"goldenrod", "octagon", "filled, solid"};
+static constexpr BoxStyle PortFieldStyle{"olive", "component", "filled, solid"};
```
flomnes, line 42: Use the anonymous namespace
```suggestion
namespace {
constexpr BoxStyle SumStyle{"aqua", "hexagon", "filled, solid"};
constexpr BoxStyle BinaryStyle{"aqua", "oval", "filled, rounded"};
constexpr BoxStyle ComparisonStyle{"yellow", "diamond", "filled"};
constexpr BoxStyle NegationStyle{"tomato", "invtriangle", "filled, solid"};
constexpr BoxStyle LiteralStyle{"lightgray", "box", "filled, solid"};
constexpr BoxStyle VariableStyle{"gold", "box", "filled, solid"};
constexpr BoxStyle ParameterStyle{"wheat", "box", "filled, solid"};
constexpr BoxStyle ComponentParameterStyle{"springgreen", "octagon", "filled, solid"};
constexpr BoxStyle ComponentVariableStyle{"goldenrod", "octagon", "filled, solid"};
constexpr BoxStyle PortFieldStyle{"olive", "component", "filled, solid"};
}
```

a-zakir: That’s code smell

a-zakir:
output : 
out

legends box-style could match those of the nodes

src/solver/expressions/include/antares/solver/expressions/visitors/AstDOTStyleVisitor.h
```diff
@@ -64,19 +123,70 @@ class AstDOTStyleVisitor: public NodeVisitor<void, std::ostream&>
...
+     *
+     * Writes the DOT representation of a node to the output stream with its associated label
+     * and style.
+     *
+     * @param id The unique ID of the node.
+     * @param label The label to be used for the node.
+     * @param box_style The style to be applied to the node's box.
+     * @param os The output stream to which the node representation is written.
+     */
     void emitNode(int id, const std::string& label, const BoxStyle& box_style, std::ostream& os);
```
flomnes:
```suggestion
    void emitNode(unsigned int id, const std::string& label, const BoxStyle& box_style, std::ostream& os);
```

src/tests/src/solver/expressions/test_AstGraphVisitor.cpp
flomnes: You don't need to make your tests depend on `std::ostream` internals and `std::cout`. Running tests with only a `std::ostringstream` is fine.

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return std::pair{expected_msg, greaterThanOrEqualNode};
+}
+
+BOOST_FIXTURE_TEST_CASE(tree_with_all_type_node, Registry<Node>)
+{
+    auto [expected_msg, expr] = buildAllTypeNode(*this);
+
+    const auto filename = std::filesystem::temp_directory_path() / "out.dot";
+    std::ofstream out(filename);
+    std::ostringstream os3;
```
flomnes: Keep only this output stream

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include <antares/solver/expressions/nodes/ExpressionsNodes.h>
+#include <antares/solver/expressions/visitors/AstDOTStyleVisitor.h>
+
+using namespace Antares::Solver;
+using namespace Antares::Solver::Nodes;
+using namespace Antares::Solver::Visitors;
+
+BOOST_AUTO_TEST_SUITE(_AstGraphVisitor_)
+
+static auto buildAllTypeNode(Registry<Node>& registry)
```
guilpier-code: `static auto buildAllTypeNode(Registry<Node>& registry)` : 
This definition could be done outside the **\_AstGraphVisitor\_** scope (and before it).

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    const auto filename = std::filesystem::temp_directory_path() / "out.dot";
+    std::ofstream out(filename);
+    std::ostringstream os3;
+    std::vector<std::ostream*> ostreams{&out, &std::cout, &os3};
+    std::stringstream redirectedStdout;
+    std::streambuf* initialBufferCout = std::cout.rdbuf(redirectedStdout.rdbuf());
+
+    AstDOTStyleVisitor astGraphVisitor;
+    for (auto* os: ostreams)
```
guilpier-code:
```C++
for (auto* os: ostreams)
{ ... }
```
Not sure this loop is required : using one of this **streams** to get the result of the test is enough, we don't get anything more by multiplying the output devices.
 We could choose a **std::ostringstream** or a **std::stringstream**, but **std::ofstream** touches file system and we should avoid it (makes the test slower and makes the test depend on the filesystem : see [here](https://stackoverflow.com/questions/17167309/is-file-access-in-unit-tests-bad)).

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,169 @@
...
+    Node* literalNode3 = registry.create<LiteralNode>(54.);
+    Node* greaterThanOrEqualNode = registry.create<GreaterThanOrEqualNode>(literalNode3,
+                                                                           lessThanOrEqualNode);
+
+    return std::pair{expected_msg, greaterThanOrEqualNode};
+}
+
+BOOST_FIXTURE_TEST_CASE(tree_with_all_type_node, Registry<Node>)
+{
+    auto [expected_msg, expr] = buildAllTypeNode(*this);
```
guilpier-code: About : 
`auto [expected_msg, expr] = buildAllTypeNode(*this);` : 

We could remove this line (and simplify the test) by defining a fixture, for example something like : 
```C++
class Fixture
{
public:
    Fixture()
    {
       makeExpression();
       expectedDotContent();
    }
   
    std::string expected_dot_content;
    Node* expression = nullptr;
    
private:
    makeExpression()
    {
        // Filling the expression
    };
    expectedDotContent()
    {
        // Defining the dot content
    }
   
    Registry<Node>& registry_;
};

...
BOOST_FIXTURE_TEST_CASE(my_test_title, Fixture)
{
    ...
}
```
By doing this, **expected_dot_content** and **expression** are directly available in the test.

src/tests/src/solver/expressions/test_AstGraphVisitor.cpp
a-zakir: I do not agree with you; I’m testing what was written in stdout not its internal mechanism

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+    else if (auto& id_map = nodeIds_[node_name]; id_map.find(node) == id_map.end())
+    {
+        id_map[node] = ++nodeCount_;
+    }
+
+    return nodeIds_[node->name()][node];
+}
+
+void AstDOTStyleVisitor::emitNode(int id,
```
guilpier-code: This method only uses its arguments and nothing else. 
Why is it a class a method and not a just free function ?

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    
+    std::ostringstream os;
+
+    AstDOTStyleVisitor astGraphVisitor;
+    
+    std::pair<AstDOTStyleVisitor&, Node*> pair1(astGraphVisitor, expr);
+    os << pair1;
```
guilpier-code: I know that @a-zakir and @flomnes discussed on what follows and made a decision, but I'm still going to leave of comment about it : 
It's about : 
```C++
AstDOTStyleVisitor astGraphVisitor;
std::pair<AstDOTStyleVisitor&, Node*> pair1(astGraphVisitor, expr);
*os << pair1;
```
The use of this visitor seems not very friendly to me ; it feels we do strange things because we can't do more simple.
Indeed :  
- First defining a **std::pair** of 2 different objects feels unusual. And giving this pair to a stream is not very friendly.
- We put the result in a **std::ostream** and not in a **std::string** (which seems more immediate)
- as we test a visitor, and more precisely the only operation it has (the dispatch / visiting), we could expect something like (like we could expect from any visitor) : 
  ```C++
  AstDOTStyleVisitor astGraphVisitor(...);
  std::string dot_content = astGraphVisitor.dispatch(expr);
  ```
  as the dot content is the result of the visit / dispatch operation.

Could we apply some of these points or it's not suitable or doable ?

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,252 @@
...
+    auto rightId = getNodeID(right);
+
+    os << "  " << id << " -> " << leftId << ";\n";
+    os << "  " << id << " -> " << rightId << ";\n";
+
+    dispatch(left, os);
+    dispatch(right, os);
+}
+
+void AstDOTStyleVisitor::NewTreeGraph(std::ostream& os, const std::string& tree_name)
```
guilpier-code, line 233: `AstDOTStyleVisitor::NewTreeGraph` : this member function don't not use data member. 
Why make it a data member and not a free function ?

flomnes (approved):

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
flomnes: We could define a wrapper such as
```cpp
std::string expressionToDOT(Node* expr)
{
    AstDOTStyleVisitor astGraphVisitor;
    std::ostringstream os;
    os << std::pair(astGraphVisitor, expr);
    return os.str();
}
```
streams are more generic than strings, and they allow e.g to write into a file without using a RAM buffer
```cpp
std::string generateLargeBufferOfData();
file << generateLargeBufferOfData(); // COPY stuff, allocate memory, etc.

void generateLargeBufferOfData(std::ostream& os):
generateLargeBufferOfData(file); // Stream is written directly to file, without large memory allocations, etc.
```

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,252 @@
...
+       << box_style.style << "\", color=\"" << box_style.color << "\"];\n";
+}
+
+// Process binary operation nodes like Add, Subtract, etc.
+void AstDOTStyleVisitor::processBinaryOperation(const Nodes::BinaryNode* node,
+                                                const std::string& label,
+                                                const BoxStyle& box_style,
+                                                std::ostream& os)
+{
+    auto id = getNodeID(node);
```
guilpier-code, line 217: Stupid question (sorry I can't find the answer by myself) :
Why do we need **getNodeID** ?
I mean, when traversing the tree, could we not, for each node, just increment **nodeCount_** and use it to compose the local info to be sent to the stream ?

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
flomnes: This was removed.

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
a-zakir: this utility function is specific to this particular test suite and is not needed by other test suites, it make sense to place it inside the BOOST_AUTO_TEST_SUITE.

a-zakir: If my reply addresses the issue, please mark this thread as resolved.

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
a-zakir: i don't see simplification, on the contrary here you define Fixture class with methods, data member to replace one Function

a-zakir: updated;
If my reply addresses the issue, please mark this thread as resolved.

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
a-zakir: at the origin the method was able to modify object of the enclosing class. Still this method is closely related to AstDOTStyleVisitor and only used by this class. i think it belongs here

a-zakir: If my reply addresses the issue, please mark this thread as resolved.

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
a-zakir: i am totaly ok with your remark @guilpier-code  the oveloard of << operator is ugly

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
a-zakir: see my [reply](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2399#discussion_r1766437874)

a-zakir: If my reply addresses the issue, please mark this thread as resolved.

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
flomnes: No, a node may not have been identified yet

a-zakir: good question! node in the DOT lang must be identified with an unique id ...
why not " just increment nodeCount_ and use it to compose the local info to be sent to the stream" --> because theoretically a node can have multiple parent :

                 A
                / \
               B  C
              /\  /\ 
             .. ..  B

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
guilpier-code: This suggestion is about simplicity **for the test**, and also clarity : 
- By defining the fixture, the test looses a line and gets more simple and readable.
- **buildAllTypeNode(...)** function makes 2 responsibilities actually, and should be split in two. 
  That's the reason why :  
  - it has to return 2 things under the form of a **std::pair** containing different object.
  - the function's name does not describe accurately what it does : the name forgets the expected dot content definition
- **buildAllTypeNode(...)** takes the **this** as an argument. If you know it refers to **Registry** passed to test as a fixture, good for you. But it's not obvious to everyone, and one has to think to understand.

I suggest that because literature on tests insists on having tests simple and short (and testing one thing only, which this test seem to do). It also says that preparation of tests should be preferably made in fixtures.

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
guilpier-code: **GetLegend** is also closely related to your class, would you make it a **AstDOTStyleVisitor**'s method then ?

Making a free function help a class method do its job is a common thing, and by sometimes preferring free functions, we can simplify the class interface (class should be as simple as possible).
You could also read [this](https://danielsieger.com/blog/2023/05/01/cpp-member-vs-free-functions.html) on that matter.

a-zakir: the article conclude "_Let me close by saying that in practice the choice is not always that obvious. There might be cases when you design an API around free functions only to later realize that some function does indeed needs access to the internals of the class. This has been exactly the case in the motivating example above, and so I resorted to a friend function. 😕_"

a-zakir: i prefer GetLegend in the class

src/tests/src/solver/expressions/test_AstDOTStyleVisitor.cpp
a-zakir: no, no and no! buildAllTypeNode doesn't have 2 responsibilities but only one:
 Create an expression, and because it creates this expression, it's the only place where we can build the expected output!

a-zakir: how one can guess  number chosen by another? coding is not magic

guilpier-code: we don't connect on this, but I still resolve this comment.

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
guilpier-code: we don't connect on this, but I still resolve this comment.

src/solver/expressions/include/antares/solver/expressions/visitors/AstDOTStyleVisitor.h
```diff
@@ -0,0 +1,193 @@
...
+                                const std::string& label,
+                                const BoxStyle& box_style,
+                                std::ostream& os);
+
+    /**
+     * @brief A map of nodes to their unique IDs.
+     *
+     * This map is used to keep track of assigned IDs for each node in the AST.
+     */
+    std::unordered_map<std::string, std::unordered_map<const Nodes::Node*, unsigned int>> nodeIds_;
```
guilpier-code: About **nodeIds_** : 
To me, it's a complex structure, and if I understand it correctly, it contains 2 types of information : 
- for each node, its number of occurrences in the tree
- the number of nodes of each C++ type

A possible way to clarify this would be : 
- To split this container in two more simple containers.
- The first one would remain a private data member and would still be filled when traversing the tree (like the current one).
- The second one (numbers of instances for each type) would be computed in **GetLegend**, and does need to be a data member, which would simplify the class.

If we do this, class is simpler, as well as **getNodeIds**, and this last function will do exactly what its name says it does.

flomnes: Yes good idea, @guilpier-code can you open a PR to implement it ?

guilpier-code: > Yes good idea, @guilpier-code can you open a PR to implement it ?

@a-zakir : what do you think ? If you agree and want to do it, please do

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
guilpier-code: ok

src/solver/expressions/include/antares/solver/expressions/visitors/AstDOTStyleVisitor.h
a-zakir: as i explained to you, this is the most efficient way. The maps are strictly related, but if you think you can ameliorate the code, be My guest and open a Pr.

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+std::string AstDOTStyleVisitor::name() const
+{
+    return "AstGraphVisitor";
+}
+
+unsigned int AstDOTStyleVisitor::getNodeID(const Nodes::Node* node)
+{
+    const auto& node_name = node->name();
+    if (nodeIds_.find(node_name) == nodeIds_.end())
```
flomnes, line 190: You can use ‘contains‘ for simpler code

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+    auto id = getNodeID(node);
+    emitNode(id,
+             "CP(" + node->getComponentId() + "," + node->getComponentName() + ")",
+             NodeStyle::ComponentParameterStyle,
+             os);
+}
+
+std::string AstDOTStyleVisitor::name() const
+{
+    return "AstGraphVisitor";
```
flomnes: AstDOTStyleVisitor

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+}
+
+unsigned int AstDOTStyleVisitor::getNodeID(const Nodes::Node* node)
+{
+    const auto& node_name = node->name();
+    if (nodeIds_.find(node_name) == nodeIds_.end())
+    {
+        nodeIds_[node->name()][node] = ++nodeCount_;
```
flomnes, line 192: It seems like you don't need to mix node IDs and category counting, you might consider two separate maps here

src/solver/expressions/include/antares/solver/expressions/visitors/AstDOTStyleVisitor.h
```diff
@@ -0,0 +1,196 @@
...
+                                const std::string& label,
+                                const BoxStyle& box_style,
+                                std::ostream& os);
+
+    /**
+     * @brief A map of nodes to their unique IDs.
+     *
+     * This map is used to keep track of assigned IDs for each node in the AST.
+     */
+    std::map<std::string, std::map<const Nodes::Node*, unsigned int>> nodeIds_;
```
flomnes, line 173: Using pointers as map keys can lead to flaky tests. 

Also, do you need such detailed info ? Why not just map<string, size_t> (type of node -> number of occurrences) ?

flomnes (changes requested):

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+static constexpr BoxStyle LiteralStyle{"lightgray", "box", "filled, solid"};
+static constexpr BoxStyle VariableStyle{"gold", "box", "filled, solid"};
+static constexpr BoxStyle ParameterStyle{"wheat", "box", "filled, solid"};
+static constexpr BoxStyle ComponentParameterStyle{"springgreen", "octagon", "filled, solid"};
+static constexpr BoxStyle ComponentVariableStyle{"goldenrod", "octagon", "filled, solid"};
+static constexpr BoxStyle PortFieldStyle{"olive", "component", "filled, solid"};
+} // namespace NodeStyle
+
+std::string GetLegend(
+  const std::map<std::string, std::map<const Nodes::Node*, unsigned int>>& nodeIds)
```
flomnes: Pass a `std::ostream&` directly, don't use temporaries.

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+)raw";
+    std::set<std::string> legend_id;
+    for (const auto& [node_type, id_map]: nodeIds)
+    {
+        const auto [it, inserted] = legend_id.insert("legend_" + node_type);
+
+        result << *it << " [ label =\" " << node_type << ": " << id_map.size() << "\"]\n";
+    }
+    if (legend_id.size() > 1)
```
flomnes:
```suggestion
    if (!legend_id.empty())
```

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+    os << "digraph " + tree_name + " {\n";
+    os << "node[style = filled]\n";
+}
+
+void AstDOTStyleVisitor::EndTreeGraph(std::ostream& os)
+{
+    // Graph title showing the total number of nodes
+    os << "label=\"AST Diagram(Total nodes : " << nodeCount_ << ")\"\n";
+    os << "labelloc = \"t\"\n";
+    os << GetLegend(nodeIds_);
```
flomnes:
```suggestion
    GetLegend(nodeIds_, os);
```

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
a-zakir: size > 1  ;)

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,254 @@
...
+label = "Legend";
+style = dashed;
+fontsize = 16;
+color = lightgrey;
+node [shape=plaintext];
+
+)raw";
+
+    auto order = nodeIds.size() <=> 1;
+    if (order == std::strong_ordering::greater)
```
flomnes: The spaceship operator is nice, but it makes the code less readable IMO.

```suggestion
    if (nodeIds.size() > 1)
```

src/solver/expressions/visitors/AstDOTStyleVisitor.cpp
```diff
@@ -0,0 +1,254 @@
...
+             ++it, ++next_it)
+        {
+            auto current_legend_id = std::string("legend_") + it->first;
+            auto next_legend_id = std::string("legend_") + next_it->first;
+            os << current_legend_id << " [ label =\" " << it->first << ": " << it->second.size()
+               << "\"]\n";
+            os << current_legend_id << " -> " << next_legend_id << " [style=invis];\n";
+        }
+    }
+    else if (order == std::strong_ordering::equal)
```
flomnes:
```suggestion
    else if (nodeIds.size() == 1)
```

flomnes (approved):

PR #2401, opened 2024-09-18
Protect study path with space when using cucumber

JasonMarechal25:
Study path with spaces can lead to error if the CLI interpreter assume the path being multiple space separated parameters.

- Surround study path parameter with double quotes
- Update documentation: use `python -m pip` instead of `pip`

### DIFF
--- a/src/tests/cucumber/features/steps/common_steps/simulator_utils.py
+++ b/src/tests/cucumber/features/steps/common_steps/simulator_utils.py
@@ -27,7 +27,7 @@ def init_simu(context):
 
 
 def build_antares_solver_command(context):
-    command = [context.config.userdata["antares-solver"], "-i", str(context.study_path)]
+    command = [context.config.userdata["antares-solver"], "-i", f'"{context.study_path}"']
     solver = "sirius"
     if "solver" in context.config.userdata:
         solver = context.config.userdata["solver"]


src/tests/cucumber/requirements.txt
```diff
@@ -1,2 +1,3 @@
 behave
-pyyaml
\ No newline at end of file
+pyyaml
+testresources
```
pet-mit: why do you need this new requirement ?

src/tests/cucumber/features/steps/simulator_utils.py
```diff
@@ -35,7 +55,7 @@ def activate_simu_outputs(context):
 
 
 def build_antares_solver_command(context):
-    command = [SOLVER_PATH, "-i", str(context.study_path)]
+    command = [SOLVER_PATH, "-i", f'"{context.study_path}"']
```
flomnes: Please revert all other changes and keep only this one.

src/tests/cucumber/requirements.txt
JasonMarechal25: Without it I had an error installing requirements saying it was missing

src/tests/cucumber/features/steps/simulator_utils.py
JasonMarechal25: I have removed the copyright notice but other changes are legitimate.

src/tests/cucumber/requirements.txt
JasonMarechal25: I have removed the addition.
Locally I uninstalled testresources and it works properly

pet-mit (approved):

PR #2406, opened 2024-09-18
Modeler 2.4c: portfield substitution

payetvin:


src/solver/expressions/include/antares/solver/expressions/visitors/PortfieldSubstitutionVisitor.h
```diff
@@ -0,0 +1,67 @@
...
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+#include <boost/functional/hash.hpp>
```
flomnes: You can just declare `KeyHasher` here and define it in a .cpp file
```cpp
struct KeyHasher
{
    std::size_t operator()(const Nodes::PortFieldNode& n) const;
};
```
Then include `boost/functional/hash.hpp` in the .cpp file (implementation detail).

src/solver/expressions/include/antares/solver/expressions/visitors/PortfieldSubstitutionVisitor.h
```diff
@@ -0,0 +1,67 @@
...
+{
+  std::size_t operator()(const Nodes::PortFieldNode& n) const
+  {
+      std::size_t seed = 0;
+
+      boost::hash_combine(seed, boost::hash_value(n.getPortName()));
+      boost::hash_combine(seed, boost::hash_value(n.getFieldName()));
+
+      return seed;
+  }
```
flomnes: Move to .cpp

src/solver/expressions/visitors/PortfieldSubstitutionVisitor.cpp
```diff
@@ -0,0 +1,49 @@
...
+    ctx_(ctx)
+{
+}
+
+Nodes::Node* PortfieldSubstitutionVisitor::visit(const Nodes::PortFieldNode* node)
+{
+    if (ctx_.portfield.contains(*node))
+    {
+        return (ctx_.portfield.at(*node));
+    }
```
flomnes: ### Performance

`std::unordered_map::contains` is nice, but here we perform 2 lookup operations
1. Check if `ctx_.portfield` contains `*node`
2. If so, lookup the the corresponding value and return it

I think we'd better stick to the good old `std::unordered_map::find`, to get the iterator, check it against `ctx_.portfield.end()`, etc.

According to the doc, [`contains`](https://en.cppreference.com/w/cpp/container/unordered_map/contains), [`at`](https://en.cppreference.com/w/cpp/container/unordered_map/at) and [`find`](https://en.cppreference.com/w/cpp/container/unordered_map/find) are all constant on average, worst case linear in the size of the container.

src/solver/expressions/include/antares/solver/expressions/visitors/PortfieldSubstitutionVisitor.h
```diff
@@ -0,0 +1,61 @@
...
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+#include <map>
+
+#include <boost/functional/hash.hpp>
+
```
flomnes:
```suggestion
```

flomnes (approved):

PR #2407, opened 2024-09-18
Place CSR after hydro remix [ANT-2070]

flomnes:


payetvin (approved):
We should not forget to reactivate and regenerate results for CSR tests

guilpier-code (approved):

PR #2410, opened 2024-09-20
Remove last global variable

### LINKED ISSUE #1719: Remove last global variable
`transitMoyenInterconnexionsRecalculQuadratique`

payetvin:
close #1719


src/libs/antares/study/include/antares/study/runtime/runtime.h
```diff
@@ -131,6 +131,8 @@ class StudyRuntimeInfos
     */
     bool quadraticOptimizationHasFailed;
```
guilpier-code, line 132: Not directly related to this PR, but very easy to fix : 
Declaration : `void initializeThermalClustersInMustRunMode(Study& study) const;` 
is never defined, and a fortiori never called.
We should remove the declaration.

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -343,6 +343,12 @@ bool StudyRuntimeInfos::loadFromStudy(Study& study)
     // Check if some clusters request TS generation
     checkThermalTSGeneration(study);
 
+    transitMoyenInterconnexionsRecalculQuadratique.resize(study.runtime.interconnectionsCount());
+    for (uint i = 0; i != study.runtime.interconnectionsCount(); i++)
+    {
+        transitMoyenInterconnexionsRecalculQuadratique[i].assign(HOURS_PER_YEAR, 0.);
+    }
```
guilpier-code: In block : 
```C++
transitMoyenInterconnexionsRecalculQuadratique.resize(study.runtime.interconnectionsCount());
for (uint i = 0; i != study.runtime.interconnectionsCount(); i++)
{
    transitMoyenInterconnexionsRecalculQuadratique[i].assign(HOURS_PER_YEAR, 0.);
}
```
when we call `study.runtime.interconnectionsCount()`, we're already in the context of **study.runtime**, because we are in class **StudyRuntimeInfos**.

Simple call to `interconnectionsCount()` is enough I guess.

src/solver/optimisation/opt_calcul_des_pmin_MUT_MDT.cpp
```diff
@@ -23,7 +23,6 @@
 
 #include "antares/solver/optimisation/opt_fonctions.h"
```
guilpier-code, line 24: ### Don't treat this remark, it'll be done in a separate PR 

Same as above : 
We should replace : 

```c++
#include "antares/solver/optimisation/opt_fonctions.h"
#include "antares/solver/optimisation/opt_structure_probleme_a_resoudre.h"
#include "antares/solver/simulation/sim_extern_variables_globales.h"
#include "antares/solver/simulation/sim_structure_donnees.h"
#include "antares/solver/simulation/simulation.h"
``` 
with ; 
```c++
#include "antares/solver/simulation/sim_structure_probleme_economique.h"
```

guilpier-code (changes requested):

src/solver/variable/include/antares/solver/variable/economy/links/flowQuad.h
```diff
@@ -268,6 +269,8 @@ class FlowQuad: public Variable::IVariable<FlowQuad<NextT>, NextT, VCardFlowQuad
 private:
     uint pLinkGlobalIndex;
     uint pNbHours;
+    std::vector<std::vector<double>>* transitMoyenInterco;
```
guilpier-code: About `std::vector<std::vector<double>>* transitMoyenInterco;`

Why not a reference to the vector (instead of a raw pointer) ?
It would avoid uses of symbol `&` and notation` (*transitMoyenInterco)`, always a bit harder to read.

guilpier-code: _Cosmetic_
About `std::vector<std::vector<double>>* transitMoyenInterco;`

Maybe renaming it into **transitMoyenInterco_** (trailing underscore) or **pTransitMoyenInterco** to tell reader it's a private data member.

guilpier-code (changes requested):

src/solver/variable/include/antares/solver/variable/economy/links/flowQuad.h
payetvin: Reference members must be initialized at construction, here we initialize it with another function

guilpier-code: True, I missed this point

guilpier-code: Maybe adding a **const** then : 
`const std::vector<std::vector<double>>* transitMoyenInterco;`

==> No, sorry, it'll force us to initialize the data member in the constructor. Forget that

guilpier-code (approved):

PR #2411, opened 2024-09-20
Headers files : remove useless inclusions 

guilpier-code:
When reviewing PR #2410, we noticed that some headers inclusions are irrelevant, especially in the **OPT** part.
The current PR is about fixing this.

### DIFF
--- a/src/solver/hydro/management/monthly.cpp
+++ b/src/solver/hydro/management/monthly.cpp
@@ -199,7 +199,7 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
               case OUI:
               {
 #ifndef NDEBUG
-                      CheckHydroAllocationProblem(area, problem, initReservoirLvlMonth, lvi);
+                  CheckHydroAllocationProblem(area, problem, initReservoirLvlMonth, lvi);
 #endif
 
                   for (uint month = 0; month != MONTHS_PER_YEAR; ++month)
--- a/src/solver/hydro/management/monthly.cpp
+++ b/src/solver/hydro/management/monthly.cpp
@@ -198,10 +198,9 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
               {
               case OUI:
               {
-                  if (false)
-                  {
+#ifndef NDEBUG
                       CheckHydroAllocationProblem(area, problem, initReservoirLvlMonth, lvi);
-                  }
+#endif
 
                   for (uint month = 0; month != MONTHS_PER_YEAR; ++month)
                   {
--- a/src/solver/hydro/management/monthly.cpp
+++ b/src/solver/hydro/management/monthly.cpp
@@ -198,7 +198,7 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
               {
               case OUI:
               {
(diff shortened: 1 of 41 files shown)


flomnes (approved):

src/solver/optimisation/opt_appel_solveur_quadratique.cpp
```diff
@@ -21,24 +21,22 @@
...
+#include <setjmp.h>
 extern "C"
 {
-#include "pi_define.h"
-#include "pi_definition_arguments.h"
-#include "pi_fonctions.h"
+#include <pi_constantes_externes.h>
+#include <pi_definition_arguments.h>
+#include <pi_fonctions.h>
 }
```
guilpier-code, line 36: `#include <setjmp.h>` : 
We should make a ticket about this.
As the related comment says, this inclusion should belong to **Sirius**, but because **Sirius** does not do it, **Antares** has to do it instead, or code does not compile.
Why not change **Sirius** ? Any reason ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -268,7 +256,7 @@ static SimplexResult OPT_TryToCallSimplex(const OptimizationOptions& options,
             logs.info() << " Solver: Standard resolution failed";
             logs.info() << " Solver: Retry in safe mode"; // second trial w/o scaling
 
-            if (Logs::Verbosity::Debug::enabled)
+            if (false) // To be turned to true if we want to print this
```
flomnes: Better yet 
```cpp
logs.debug() << "message";
```
So you can drop the `if` and only display info in debug mode

src/solver/optimisation/opt_appel_solveur_quadratique.cpp
flomnes: No particular reason, we don't want to touch Sirius too much (deprecated)

src/solver/hydro/management/monthly.cpp
```diff
@@ -206,7 +199,7 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
               {
               case OUI:
               {
-                  if (Logs::Verbosity::Debug::enabled)
+                  if (false) // Put true to have extra check
```
flomnes: Please revert this

flomnes (changes requested):

src/solver/hydro/management/monthly.cpp
guilpier-code: `Logs::Verbosity::Debug::enabled` : 
- belongs to **Yuni** 
- is a preprocessor directive

It has all flaws :-)
Back to **Yuni** dependency in this file ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp

src/solver/hydro/management/monthly.cpp
flomnes: Then we should have a replacement for this, for exemple
```cpp
#ifndef NDEBUG
// print some info, make extra checks, etc.
#endif
```
See https://stackoverflow.com/a/8594122

guilpier-code: Done

flomnes (approved):

PR #2414, opened 2024-09-23
Better handling of discontinued parameters

payetvin:


src/libs/antares/study/parameters.cpp
```diff
@@ -1124,6 +1124,36 @@ static bool SGDIntLoadFamily_Legacy(Parameters& d,
         return true;
     }
 
+    if (key == "set-to-null-ntc-between-physical-out-for-first-step")
```
guilpier-code: We have a code duplication we did not have before, with the use of **legacyLocalMatchingKeys**.
There may be a reason for this. Which one ?

src/libs/antares/study/parameters.cpp
```diff
@@ -1111,15 +1111,30 @@ static bool SGDIntLoadFamily_Legacy(Parameters& d,
...
-              << "Hydro hot start not supported with this solver, please use a version < 9.2";
+            logs.warning() << "Parameter set-to-null-ntc-between-physical-out-for-first-step "
+                              " is deprecated, please remove it from the study";
+        }
+        return true;
+    }
+
+    if (key == "enable-first-step") // ignored since 9.2
+    {
+        if (value == "false")
```
flomnes: We can't handle enable-first-step=true anymore, even though it was the default value previously.
```suggestion
        if (value == "true")
```

flomnes (commented):
Merge nested `if`

flomnes (approved):

PR #2419, opened 2024-09-24
Force enable-first-step=false

flomnes:


src/libs/antares/study/parameters/adq-patch-params.cpp
```diff
@@ -128,8 +136,6 @@ void CurtailmentSharing::addProperties(IniFile::Section* section) const
 // ------------------------
 void AdqPatchParams::reset()
 {
-    enabled = false;
```
payetvin: This `enabled` var is related to adq patch, not the same as the one in local matching
This change is also present on develop, I'll do a PR to fix it

payetvin (dismissed):

src/libs/antares/study/parameters/adq-patch-params.cpp
payetvin: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2420

flomnes: done

src/libs/antares/study/parameters/adq-patch-params.cpp
```diff
@@ -25,7 +25,15 @@ bool LocalMatching::updateFromKeyValue(const Yuni::String& key, const Yuni::Stri
...
     if (key == "enable-first-step")
-        return value.to<bool>(enabled);
+    {
+        if (value == "true")
+        {
+            logs.warning() << "Property enable-first-step has been disabled, it is known to cause "
+                              "errors and inconsistent results";
+        }
+        return true;
+    }
```
guilpier-code: _Cosmetic_

Or : 
```C++
if (key == "enable-first-step" && value == "true")
{
    logs.warning() << "Property enable-first-step has been disabled, it is known to cause "
                      "errors and inconsistent results"
    return true;
 }
```

src/libs/antares/study/parameters/adq-patch-params.h
```diff
@@ -59,7 +59,7 @@ enum class AdqPatchPTO
 
 struct LocalMatching
```
guilpier-code, line 60: `struct LocalMatching` : do we keep this **struct** to have a minimal change ?

flomnes: yes

guilpier-code (approved):

PR #2431, opened 2024-09-26
[ANT-2206] yaml to model

JasonMarechal25:
# Description
Expose `Antares::Solver::ObjectModel::Library Antares::Solver::ModelConverter::convert(const Antares::Solver::ModelParser::Library& library)`

# Implementation details
Convert a Library resulting from yaml parsing to a library in the context ob ObjectModel

Use builder pattern to construct complexes object such as library and models

Declaring a lib ModelConverter as the same level as ObjectModel and ObjectParser prevent either from depending on both. Conceptually they're independent from each others

Defines two ValueType enums in two different contexts. Allow for diverging evolution

# Limitations

- [ANT-2228](https://gopro-tickets.rte-france.com/browse/ANT-2228) Gérer les exception de champs superflus
- [ANT-2230](https://gopro-tickets.rte-france.com/browse/ANT-2230) Gérer les champs optionnels
- [ANT-2233](https://gopro-tickets.rte-france.com/browse/ANT-2233) Pour les valueType, gérer la capitalisation
- [ANT-2233](https://gopro-tickets.rte-france.com/browse/ANT-2233) value_type pour les Parameters
- [ANT-2235](https://gopro-tickets.rte-france.com/browse/ANT-2235) contraintes et binding_constraints
- [ANT-2236](https://gopro-tickets.rte-france.com/browse/ANT-2236) Constraintes et expressions, inadéquation models/yaml
- Ports are not handled in the scope of this PR/ANT-2206


src/solver/libModelObject/include/antares/solver/libObjectModel/valueType.h
```diff
@@ -31,4 +33,25 @@ enum class ValueType
...
+    case ValueType::BOOL:
+        os << "BOOL"s;
+        break;
+    default:
+        os << "UNKNOWN"s;
+        break;
+    }
+    return os;
+}
+
```
flomnes: Why is this function needed ?

src/solver/libModelObject/library.cpp
```diff
@@ -0,0 +1,101 @@
...
+ * \return Reference to the LibraryBuilder object.
+ *
+ * inputs it not garanteed to be valid after the call
+ */
+LibraryBuilder& LibraryBuilder::withModel(std::vector<Model>& models)
+{
+    std::transform(models.begin(),
+                   models.end(),
+                   std::inserter(library_.models_, library_.models_.end()),
+                   [](/*Non const to prevent copy*/ Model& model)
```
flomnes, line 87:
```suggestion
                   [](/*Non const to prevent copy*/ Model&& model)
```

src/solver/libModelObject/model.cpp
```diff
@@ -19,14 +19,120 @@
...
+    return *this;
+}
+
+/**
+ * \brief Sets the variables of the model.
+ *
+ * \param variables A vector of Variable objects to set.
+ * \return Reference to the ModelBuilder object.
+ *
+ * inputs it not garanteed to be valid after the call
```
flomnes, line 90: To better reflect this, we might as well use `std::vector<T>&&`
```cpp
void takeControl(std::vector<T>&& vec)
{
    
}
```

src/solver/modelConverter/modelConverter.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+#include <stdexcept>
+
+#include "antares/solver/libObjectModel/constraint.h"
+#include "antares/solver/libObjectModel/library.h"
+#include "antares/solver/libObjectModel/model.h"
+#include "antares/solver/libObjectModel/parameter.h"
+#include "antares/solver/libObjectModel/port.h"
+#include "antares/solver/libObjectModel/portType.h"
+#include "antares/solver/libObjectModel/variable.h"
```
flomnes, line 32: Consider adding a "main header" for convenience

src/solver/modelParser/include/antares/solver/modelParser/model.h
```diff
@@ -34,11 +34,56 @@ struct Parameter
...
+        break;
+    case ValueType::BOOL:
+        os << "BOOL"s;
+        break;
+    default:
+        os << "UNKNOWN"s;
+        break;
+    }
+    return os;
+}
```
flomnes, line 80: If this is only required in the tests, then please move it there.

src/tests/src/solver/modelParser/testModelTranslator.cpp
```diff
@@ -0,0 +1,237 @@
...
+    BOOST_REQUIRE_EQUAL(lib.portTypes().at("port2").fields().size(), 2);
+    BOOST_CHECK_EQUAL(lib.portTypes().at("port2").fields()[0].Name(), "field3");
+    BOOST_CHECK_EQUAL(lib.portTypes().at("port2").fields()[1].Name(), "field4");
+}
+
+// Test library with models
+BOOST_AUTO_TEST_CASE(test_library_empty_models)
+{
+    ModelParser::Library library;
+    ModelParser::Model model1{"model1", "description", {}, {}, {}, {}, {}, "objectives"};
```
flomnes: Can you at least add a comment for each `{}` ?

src/tests/src/solver/modelParser/testModelTranslator.cpp
```diff
@@ -0,0 +1,237 @@
...
+    library.models = {model1};
+    ObjectModel::Library lib = ModelConverter::convert(library);
+    auto& model = lib.models().at("model1");
+    // BOOST_REQUIRE_EQUAL(model.Ports().size(), 2);
+    // auto& port1 = model.Ports().at("port1");
+    // auto& port2 = model.Ports().at("port2");
+    // BOOST_CHECK_EQUAL(port1.Name(), "port1");
+    //  BOOST_CHECK_EQUALS port1.Type()
+    // BOOST_CHECK_EQUAL(port2.Name(), "port2");
+    // BOOST_CHECK_EQUALS port2.Type()
```
flomnes, line 211: This test does nothing

src/solver/libModelObject/include/antares/solver/libObjectModel/parameter.h
```diff
@@ -35,14 +35,59 @@ namespace Antares::Solver::ObjectModel
...
+    {
+        NO = false,
+        YES = true
+    };
+
+    enum class ScenarioDependent : bool
+    {
+        NO = false,
+        YES = true
+    };
```
pet-mit, line 53: I'm not a big fan of this. we can workaround the code readability issue you stated by adding an in-line comment before the name, or by implementing a ParameterBuilder, like you did for Library and Model (but that would be a lot more complex, maybe do it in another PR if you want to)

It would be something like : 
modelBuilder.newParameter().withId(...).withType(...).isTimeDependent(true).isScenarioDependent(false).add()
.newParameter()...

src/solver/libModelObject/include/antares/solver/libObjectModel/model.h
```diff
@@ -39,22 +39,61 @@ namespace Antares::Solver::ObjectModel
...
 };
 
+class ModelBuilder
+{
+public:
+    ModelBuilder& withId(std::string_view id);
+    ModelBuilder& withObjective(Expression objective);
+    ModelBuilder& withParameters(std::vector<Parameter>& parameters);
+    ModelBuilder& withVariables(std::vector<Variable>& variables);
+    ModelBuilder& withPorts(std::vector<Port>& ports);
```
pet-mit: you may make it even easier to use, if you implement withParameter(Parameter& parameter), withVariable, and withPort, so the user can add objects iteratively.
Better yet, you can create a ParameterBuilder, a VariableBuilder, and a PortBuilder (see my previous comment for an example)

src/solver/libModelObject/include/antares/solver/libObjectModel/library.h
```diff
@@ -32,15 +33,56 @@ namespace Antares::Solver::ObjectModel
...
+ * Follow builder pattern:
+ * builder.Library().withId("id").withDescription("description").withPortType(portList).withModel(modelList).build();
+ */
+class LibraryBuilder
+{
+public:
+    LibraryBuilder& withId(const std::string& id);
+    LibraryBuilder& withDescription(const std::string& description);
+    LibraryBuilder& withPortType(std::vector<PortType>& portTypes);
+    LibraryBuilder& withModel(std::vector<Model>& models);
```
pet-mit: if you want to go further : instead of adding a built model, you can declare a "newModel()" method that returns a ModelBuilder

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -313,11 +313,84 @@ BOOST_AUTO_TEST_CASE(test_library_model_multiple_variables)
...
+        library:
+            id: "lib_id"
+            description: "lib_description"
+            port-types: []
+            models:
+                - id: "model_id"
+                  description: "model_description"
+                  parameters: []
+                  variables:
+                      - name: "var_name"
```
pet-mit: change all "name" occurences in yaml to "id". in Model lib too

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -313,11 +313,84 @@ BOOST_AUTO_TEST_CASE(test_library_model_multiple_variables)
...
+                        lower-bound: "near-zero"
+                        upper-bound: "pmax"
+                  ports: []
+                  port-field-definitions: []
+                  constraints: []
+                  objective: "objective"
+        )"s;
+    Antares::Solver::ModelParser::Library libraryObj = parser.parse(library);
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].lower_bound, "near-zero");
```
pet-mit, line 345: check variable id

src/tests/src/solver/modelParser/testModelTranslator.cpp
pet-mit, line 1: I find these tests very hard to read. though I understand that they are unit tests, I think we should at least complement them with some tests on using yaml texts

pet-mit (changes requested):
nice move with the builders. some comments if you want to take it further + on other parts of the code

src/solver/libModelObject/include/antares/solver/libObjectModel/valueType.h
JasonMarechal25: For google test message when comparing values

src/tests/src/solver/modelParser/testModelTranslator.cpp
JasonMarechal25: Yes. *boost::unit_test::disabled()
It's here as a skeleton for when ports will be implemented properly
Can be deleted if you prefer so

src/solver/libModelObject/library.cpp
JasonMarechal25: Not possible.

> The signature of the function should be equivalent to the following:
> 
>  Ret fun(const Type &a);
> 
> The signature does not need to have const &.
> The type  Type must be such that an object of type InputIt can be dereferenced and then implicitly converted to  Type. The type Ret must be such that an object of type OutputIt can be dereferenced and assigned a value of type Ret. ​

src/solver/libModelObject/model.cpp
JasonMarechal25: Nice.
I kept the names to stay in the convention of builder but changed to rvalue ref. Client are forced to std::move(value) so no mistake possible

src/solver/modelConverter/modelConverter.cpp
JasonMarechal25: If further clients need to include all headers every time, indeed we can expose a meta-header in libObjectModel

src/tests/src/solver/modelParser/testModelTranslator.cpp
JasonMarechal25: There is one in a subsequent PR: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2448

src/tests/src/solver/modelParser/testModelTranslator.cpp
flomnes: No, you may keep it here

src/solver/libModelObject/include/antares/solver/libObjectModel/model.h
JasonMarechal25: I don't think there is value to this at the moment.
The current interface stem from the usage where models are created by parsing yaml, not built be hand.
At least I don't think a value by value design would plug in nicely. If you have a example I'm open to it

src/solver/libModelObject/include/antares/solver/libObjectModel/library.h
JasonMarechal25: Same as https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2431#discussion_r1784571025

src/solver/libModelObject/model.cpp
```diff
@@ -19,14 +19,120 @@
...
+ */
+ModelBuilder& ModelBuilder::withParameters(std::vector<Parameter>&& parameters)
+{
+    std::transform(parameters.begin(),
+                   parameters.end(),
+                   std::inserter(model_.parameters_, model_.parameters_.end()),
+                   [](/*Non const to prevent copy*/ Parameter& parameter)
+                   { return std::make_pair(parameter.Name(), std::move(parameter)); });
```
flomnes:
```suggestion
                   { return std::make_pair(parameter.Id(), std::move(parameter)); });
```

src/solver/libModelObject/model.cpp
```diff
@@ -19,14 +19,120 @@
...
+ *
+ * inputs it not garanteed to be valid after the call
+ */
+ModelBuilder& ModelBuilder::withVariables(std::vector<Variable>&& variables)
+{
+    std::transform(variables.begin(),
+                   variables.end(),
+                   std::inserter(model_.variables_, model_.variables_.end()),
+                   [](/*Non const to prevent copy*/ Variable& variable)
+                   { return std::make_pair(variable.Name(), std::move(variable)); });
```
flomnes:
```suggestion
                   { return std::make_pair(variable.Id(), std::move(variable)); });
```

flomnes (changes requested):

src/tests/src/solver/modelParser/testModelParser.cpp
JasonMarechal25: Done

src/tests/src/solver/modelParser/testModelParser.cpp
JasonMarechal25: Done

src/solver/modelParser/include/antares/solver/modelParser/model.h
```diff
@@ -29,21 +29,66 @@ namespace Antares::Solver::ModelParser
 // Define structures
 struct Parameter
 {
```
guilpier-code, line 32: question : as the current file contains the declaration / definition of type **Library**, which is the containing type (it has a a data member of all the other types declared in this file, among which is type **Model**), why not name this file **Library.h** ?

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -77,7 +77,7 @@ BOOST_AUTO_TEST_CASE(test_library_port_types)
                 - id: "porttype_id"
```
guilpier-code, line 75: Test **test_library_id_description** : 
considering the name of the test, are the last two checks 
```C++
 BOOST_CHECK(libraryObj.port_types.empty());
 BOOST_CHECK(libraryObj.models.empty());
```
relevant ?

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -312,12 +312,86 @@ BOOST_AUTO_TEST_CASE(test_library_model_multiple_variables)
...
+    Antares::Solver::ModelParser::Library libraryObj = parser.parse(library);
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].id, "var_name");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].lower_bound, "near-zero");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].upper_bound, "pmax");
+}
+
+// variable variable-type
+BOOST_AUTO_TEST_CASE(test_library_model_variables_type)
+{
+    Antares::Solver::ModelParser::Parser parser;
```
guilpier-code, line 352: As instanciation of the parser can be found in every test :
```c++
Antares::Solver::ModelParser::Parser parser;
```
we could define a **fixture** to be given to every test as argument

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -312,12 +312,86 @@ BOOST_AUTO_TEST_CASE(test_library_model_multiple_variables)
...
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].id, "var_name1");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].lower_bound, "0");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].upper_bound, "1");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[1].id, "var_name2");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[1].lower_bound, "-1");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[1].upper_bound, "2");
+}
+
+// variable bounds are strings expressions
+BOOST_AUTO_TEST_CASE(test_library_model_variables_bounds)
```
guilpier-code: _Really cosmetic_ : 
in order to highlight the unit test pattern **arrange, act, assert**, we should separate these 3 blocks with empty lines

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -312,12 +312,86 @@ BOOST_AUTO_TEST_CASE(test_library_model_multiple_variables)
...
+                  objective: "objective"
+        )"s;
+    Antares::Solver::ModelParser::Library libraryObj = parser.parse(library);
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].id, "var_name");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].lower_bound, "near-zero");
+    BOOST_CHECK_EQUAL(libraryObj.models[0].variables[0].upper_bound, "pmax");
+}
+
+// variable variable-type
+BOOST_AUTO_TEST_CASE(test_library_model_variables_type)
```
guilpier-code: test names : 
- suffixing each test name with **test_** does not seem very helpful
- the action here is the parsing. Why not make "parse" appear in the name of the test
- we should not hesitate to make names of tests look like a sentence, for example : 
  **test_library_model_variables_type**  ==>  **parsing_a_model_with_many_variables**

src/solver/modelParser/include/antares/solver/modelParser/model.h
JasonMarechal25: Good point.
I was thinking "model" in the context of "modeler". I think Library.h may indeed be more accurate

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -298,10 +298,10 @@ BOOST_AUTO_TEST_CASE(test_library_model_multiple_variables)
...
                   variables:
-                      - name: "var_name1"
+                      - id: "var_name1"
                         lower-bound: 0
                         upper-bound: 1
-                      - name: "var_name2"
+                      - id: "var_name2"
                         lower-bound: -1
                         upper-bound: 2
                   ports: []
```
guilpier-code, line 305: question  : 
From what I can see, there are no **Model** without fields : 
```yaml
variables: []
ports: []
port-field-definitions: []
constraints: []
objective: "objective"
```
even if these fields are empty. These fields seem necessary, are they ?

src/tests/src/solver/modelParser/testModelParser.cpp
JasonMarechal25: Yes, but it's only one declaration so the fixture doesn't bring much value for now

src/tests/src/solver/modelParser/testModelParser.cpp
JasonMarechal25: Indeed
I've renamed tests with meaningful names following your suggestion

src/tests/src/solver/modelParser/testModelParser.cpp
guilpier-code: Up to you.
(to me, the parser creation belongs to a set up, it makes each test a bit shorter, and contains only what is variable from a test to another).

src/solver/modelParser/include/antares/solver/modelParser/Library.h
```diff
@@ -29,21 +29,66 @@ namespace Antares::Solver::ModelParser
...
+    case ValueType::INTEGER:
+        return "INTEGER"s;
+    case ValueType::BOOL:
+        return "BOOL"s;
+    default:
+        return "UNKNOWN"s;
+    }
+}
+
+inline std::ostream& operator<<(std::ostream& os, const ValueType& value_type)
```
guilpier-code, line 61:
```c++
std::ostream& operator<<(std::ostream& os, const ValueType& value_type)
```
is defined twice : here and in **testModelTranslator.cpp**.
As you already know, we should avoid code duplication.

src/tests/src/solver/modelParser/testModelTranslator.cpp
```diff
@@ -0,0 +1,270 @@
...
+    default:
+        os << "UNKNOWN"s;
+        break;
+    }
+    return os;
+}
+} // namespace Antares::Solver::ObjectModel
+
+// Test empty library
+BOOST_AUTO_TEST_CASE(test_library_empty)
```
guilpier-code: Some suggestions for tests in **testModelParser.cpp** (tests names, steps separation, fixture, ...) could also be applied here.

src/tests/src/solver/modelParser/testModelTranslator.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// Test library with port types
+BOOST_AUTO_TEST_CASE(test_library_port_types_empty_fields)
+{
+    ModelParser::Library library;
+    ModelParser::PortType portType1{"port1", "flow port", {}};
+    ModelParser::PortType portType2{"port2", "impedance port", {}};
+    library.port_types = {portType1, portType2};
+    ObjectModel::Library lib = ModelConverter::convert(library);
```
guilpier-code, line 87: Here for instance, the **convert** (the action in the test) is lost among other steps (**arrange** and **assert**) and melted with them, so we have to look for it. 
Some empty lines around would be welcome.

src/solver/modelConverter/include/antares/solver/modelConverter/modelConverter.h
```diff
@@ -0,0 +1,41 @@
...
+{
+namespace ObjectModel
+{
+class Library;
+}
+
+namespace ModelParser
+{
+class Library;
+}
```
guilpier-code, line 35: About : 
```c++
namespace ModelParser
{
class Library;
}
```
I have warnings at compile time because of this.
Compiler complains that **ModelParser::Library** is a **struct**, not a **class**

src/solver/modelConverter/modelConverter.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * \brief Converts a ModelParser::Library object to an ObjectModel::Library object.
+ *
+ * \param library The ModelParser::Library object to convert.
+ * \return The corresponding ObjectModel::Library object.
+ */
+Antares::Solver::ObjectModel::Library convert(const Antares::Solver::ModelParser::Library& library)
+{
+    Antares::Solver::ObjectModel::LibraryBuilder builder;
```
guilpier-code, line 200: At first glance, I wonder why we would need a **builder** to build a **ObjectModel :: Library**, which is external to the class **ObjectModel :: Library** : I mean, constructors are made for this, right ? So why not use them ?
Is it a question of programming style ?
Or is there a stronger reason for using an external builder.

This comment apply to other builder classes in this file

src/solver/modelConverter/modelConverter.cpp
```diff
@@ -0,0 +1,211 @@
...
+    std::vector<Antares::Solver::ObjectModel::PortType> portTypes = convertTypes(library);
```
guilpier-code, line 201: _Cosmetic_

About : 
```c++
std::vector<Antares::Solver::ObjectModel::PortType> portTypes = convertTypes(library);
```
Why not (instead) :
 ```c++
std::vector<Antares::Solver::ObjectModel::PortType> portTypes = convertTypes(library.port_types);
```
?
Maybe more natural. 
Note that this remark applies to multiple convert functions in this file

File: src/solver/libModelObject/model.cpp [hunk 27]
```diff
@@ -19,14 +19,120 @@
...
 namespace Antares::Solver::ObjectModel
 {
 
-std::vector<Constraint*> getConstraints()
+/**
+ * \brief Builds and returns the Model object.
+ *
+ * \return The constructed Model object.
+ */
+Model ModelBuilder::build()
```
guilpier-code, line 37: Main question here (already asked about **modelConverter.cpp**) : why separate construction from the rest of the class **Model** (mainly getters) by using a model builder ?

src/tests/src/solver/modelParser/testModelParser.cpp
JasonMarechal25: They can be omitted by users and default to empty vector or empty string: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2448/files

src/solver/modelParser/include/antares/solver/modelParser/Library.h
JasonMarechal25: Not the same ValueType, different context. One is ObjectModel::ValueType, one is ModelParser::ValueType

src/solver/modelConverter/modelConverter.cpp
JasonMarechal25: https://stackoverflow.com/questions/29881135/difference-between-builder-pattern-and-constructor

Since convert() is the only client it doesn't matter much but:

Mainly because a constructor to those objects could become hard to manage in the long time.
Model has 6 members, you can define default value with empty map or string for each of them. However if a client only want to provide one of the later parameters in the list, they need to still provide empty values for the first ones. Only values after the one provided by the client are defaulted.

If a attribute need to be added of removed, the builder implementation can change and no change is made to the object constructor or API.

In the same way it encapsulate logics in different with method. Those method can be overloaded, deleted, or added without impacting the object API and constructor.

src/solver/modelConverter/modelConverter.cpp
JasonMarechal25: No particular reason

src/solver/libModelObject/model.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2431#discussion_r1787284646

src/solver/modelParser/include/antares/solver/modelParser/Library.h
guilpier-code: Could a template form of this **operator<<** be possible / suitable ?

src/solver/modelConverter/modelConverter.cpp
guilpier-code, line 201: Renaming : 

**convertTypes** ==> **convertPortTypes** (**port** seems essential here)

src/solver/modelConverter/modelConverter.cpp
guilpier-code: ok, seems compelling

src/solver/modelParser/include/antares/solver/modelParser/Library.h
guilpier-code: > Not the same ValueType, different context. One is ObjectModel::ValueType, one is ModelParser::ValueType

This a problem. It seems clear that if a `ModelParser::ValueType` changes,` ObjectModel::ValueType` will change accordingly. So this 2 types are very much related, not to say identical.

I understand that they belong to separate worlds (namespaces) and we hesitate to make this type unique for both worlds, but there is also a code duplication issue.

JasonMarechal25: > It seems clear that if a ModelParser::ValueType changes, ObjectModel::ValueType will change accordingly. So this 2 types are very much related, not to say identical.

Wrong. Take for exemple https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2447/files where they start diverging.

src/solver/libModelObject/include/antares/solver/libObjectModel/parameter.h
pet-mit: @JasonMarechal25 
I still feel that this can become cumbersome in the future (just use booleans), but it's not blocking for me for this PR

pet-mit (approved):

src/solver/libModelObject/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 project(LibObjectModel)
 
 set(SRC_model
+    library.cpp
```
a-zakir, line 4: library is a vague name

a-zakir: too vernacular

src/solver/libModelObject/include/antares/solver/libObjectModel/expression.h
```diff
@@ -28,8 +28,20 @@ namespace Antares::Solver::ObjectModel
 class Expression
 {
 public:
-    Expression();
-    ~Expression() = default;
+    Expression() = default;
+
+    explicit Expression(std::string value):
+        value(value)
```
a-zakir:
```suggestion
        value(std::move(value))
```

a-zakir: since is not const&

a-zakir: could you explain why you did not put implementations in a .cpp?

src/solver/libModelObject/include/antares/solver/libObjectModel/expression.h
```diff
@@ -28,8 +28,20 @@ namespace Antares::Solver::ObjectModel
...
+    {
+    }
+
+    const std::string& Value() const
+    {
+        return value;
+    }
+
+private:
+    std::string value;
```
a-zakir: use the convention private_

src/solver/libModelObject/include/antares/solver/libObjectModel/library.h
```diff
@@ -32,15 +33,56 @@ namespace Antares::Solver::ObjectModel
...
-    Library();
+    Library() = default;
     ~Library() = default;
 
+    const std::string& id() const
+    {
+        return id_;
+    }
+
+    const std::string& description() const
```
a-zakir: rename it to Description()

src/solver/libModelObject/include/antares/solver/libObjectModel/library.h
```diff
@@ -32,15 +33,56 @@ namespace Antares::Solver::ObjectModel
...
+        return models_;
+    }
+
 private:
+    friend class LibraryBuilder;
+
     std::string id_;
     std::string description_;
 
     std::map<std::string, PortType> portTypes_;
```
a-zakir: if the order doesn't matter, consider using unordered_map

src/solver/libModelObject/include/antares/solver/libObjectModel/constraint.h
```diff
@@ -32,11 +32,24 @@ namespace Antares::Solver::ObjectModel
...
+        expression_(std::move(expression))
+    {
+    }
+
+    const std::string& Id() const
+    {
+        return id_;
+    }
+
+    Expression expression() const
```
a-zakir, line 46: mismatch between getters name style

src/solver/libModelObject/include/antares/solver/libObjectModel/library.h
a-zakir: rename it to Description()

src/solver/libModelObject/include/antares/solver/libObjectModel/model.h
```diff
@@ -39,22 +39,61 @@ namespace Antares::Solver::ObjectModel
...
 
+class ModelBuilder
+{
+public:
+    ModelBuilder& withId(std::string_view id);
+    ModelBuilder& withObjective(Expression objective);
+    ModelBuilder& withParameters(std::vector<Parameter>&& parameters);
+    ModelBuilder& withVariables(std::vector<Variable>&& variables);
+    ModelBuilder& withPorts(std::vector<Port>&& ports);
+    Model build();
```
a-zakir, line 91: build is just a getter, add comment to en-light that

src/solver/libModelObject/include/antares/solver/libObjectModel/valueType.h
```diff
@@ -19,6 +19,8 @@
 ** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
 */
 #pragma once
+#include <iostream>
+#include <string>
```
a-zakir:
```suggestion
```

a-zakir (approved):

src/solver/libModelObject/include/antares/solver/libObjectModel/expression.h
JasonMarechal25: Since it's just an empty constructor and a getter I left them in header. No other reason

src/solver/libModelObject/include/antares/solver/libObjectModel/constraint.h
JasonMarechal25: Yes. It can't be called Expression() because it clash with the type _Expression_

flomnes (approved):

PR #2433, opened 2024-09-27
[ANT-2206] Parse yaml

JasonMarechal25:
- Parse user models described as CPP
- Expect to read one library per content
- Only accept yaml content as std::string
- Output object is Antares::Solver::ModelParser::Library
- CMake : remove WITH_YAMLCPP (always ON)


src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -0,0 +1,264 @@
...
+        )";
+    Antares::Solver::ModelParser::Library libraryObj = parser.parse(library);
+    BOOST_CHECK(libraryObj.models.size() == 1);
+    BOOST_CHECK(libraryObj.models[0].id == "model_id");
+    BOOST_CHECK(libraryObj.models[0].description == "model_description");
+    BOOST_CHECK(libraryObj.models[0].parameters.size() == 0);
+    BOOST_CHECK(libraryObj.models[0].variables.size() == 0);
+    BOOST_CHECK(libraryObj.models[0].ports.size() == 0);
+    BOOST_CHECK(libraryObj.models[0].port_field_definitions.size() == 0);
+    BOOST_CHECK(libraryObj.models[0].constraints.size() == 0);
```
flomnes:
```suggestion
    auto& model = libraryObj.models[0];
    BOOST_CHECK(model.parameters.empty());
    BOOST_CHECK(model.variables.empty());
    BOOST_CHECK(model.ports.empty());
    BOOST_CHECK(model.port_field_definitions.empty());
    BOOST_CHECK(model.constraints.empty());
```

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -0,0 +1,264 @@
...
+                  variables: []
+                  ports: []
+                  port-field-definitions: []
+                  constraints: []
+                  objective: "objective"
+        )";
+    Antares::Solver::ModelParser::Library libraryObj = parser.parse(library);
+    BOOST_CHECK(libraryObj.models.size() == 1);
+    BOOST_CHECK(libraryObj.models[0].id == "model_id");
+    BOOST_CHECK(libraryObj.models[0].description == "model_description");
```
flomnes: Use `BOOST_REQUIRE` to exit in case of invalid size (avoid subsequent segfault)

```suggestion
    BOOST_REQUIRE_EQUAL(libraryObj.models.size(), 1);
    BOOST_CHECK_EQUAL(libraryObj.models[0].id, "model_id");
    BOOST_CHECK_EQUAL(libraryObj.models[0].description, "model_description");
```

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -0,0 +1,264 @@
...
+    BOOST_CHECK(libraryObj.port_types[0].description == "porttype_description");
+    BOOST_CHECK(libraryObj.port_types[0].fields.size() == 1);
+    BOOST_CHECK(libraryObj.port_types[0].fields[0] == "port_name");
+}
+
+// Test library with models
+BOOST_AUTO_TEST_CASE(test_library_models)
+{
+    Antares::Solver::ModelParser::Parser parser;
+    auto library = R"(
```
flomnes:
```suggestion
    const std::string library = R"(
```

src/solver/modelParser/encoders.hxx
```diff
@@ -0,0 +1,239 @@
...
+        node["ports"] = rhs.ports;
+        node["port-field-definitions"] = rhs.port_field_definitions;
+        node["constraints"] = rhs.constraints;
+        node["objective"] = rhs.objective;
+        return node;
+    }
+
+    static bool decode(const Node& node, Antares::Solver::ModelParser::Model& rhs)
+    {
+        rhs.id = node["id"].as<std::string>();
```
flomnes, line 119: if(!node.IsMap()) return false;

flomnes (changes requested):
You could polish error-handling `node.as<XXX>` can throw exceptions, these exceptions are not caught currently.

Also, you should add tests for error-handling (provide an invalid YAML, check for errors)

src/tests/src/solver/modelParser/testModelParser.cpp
flomnes, line 1: You only test vectors of size 0 or 1. You should go up to 2

a-zakir:
Nice job!
Here some minor remarks :
Are the encode methods mandatory ? If not are we going to use them?
Even though using structs is the most straightforward way to read the yaml, have you considered basic class with member data private, maybe making friends the decode methods?

src/tests/src/solver/modelParser/testModelParser.cpp
JasonMarechal25: New tests added

flomnes (approved):

PR #2440, opened 2024-09-27
Add cucumber submodule for tests, remove Antares_Simulator_Tests_NR

flomnes:
Also, remove some columns from "021 Four areas - DC law" (slight result change)


pet-mit (commented):
The idea of the cucumber tests in the end is to migrate all existing tests in Antares_Simulator_Tests_NR to the new cucumber "run & check" scripts. So we will still need all studies that are in that repo. What purpose does this PR serve?

flomnes:
> The idea of the cucumber tests in the end is to migrate all existing tests in Antares_Simulator_Tests_NR to the new cucumber "run & check" scripts. So we will still need all studies that are in that repo. What purpose does this PR serve?

To have a smaller repository used only by cucumber tests (faster clone/checkout/etc.)

src/tests/cucumber/features/medium_tests.feature
```diff
@@ -2,7 +2,7 @@ Feature: medium tests
 
   @fast @medium @incomplete
   Scenario: 035 Mixed Expansion - Smart grid model 2
-    Given the study path is "medium-tests/035 Mixed Expansion - Smart grid model 2"
+    Given the study path is "035 Mixed Expansion - Smart grid model 2"
```
a-zakir, line 5: you can take advantage of this new repo to remove withespaces from study folder

src/tests/cucumber/features/short_tests.feature
```diff
@@ -36,10 +36,10 @@ Feature: short tests
...
+    Given the study path is "021 Four areas - DC law-simplified"
     When I run antares simulator
     Then the simulation takes less than 20 seconds
     And the simulation succeeds
     And the annual system cost is
-      | EXP       | STD       | MIN       | MAX       |
-      | 7.972e+10 | 2.258e+10 | 5.613e+10 | 1.082e+11 |
\ No newline at end of file
+      | EXP       | STD         | MIN         | MAX         |
+      | 6.5397e+10| 2.97462e+09 | 6.10076e+10 | 6.84837e+10 |
```
a-zakir, line 45: why?

flomnes: Slight change of results after the study was cleaned up (100 TS -> 3 TS)

PR #2448, opened 2024-10-01
[ANT-2206] Full exemple

JasonMarechal25:
Ful exemple of the parsing + conversion to model of a library


src/solver/modelParser/encoders.hxx
```diff
@@ -54,15 +61,15 @@ struct convert<Antares::Solver::ModelParser::ValueType>
         {
             return false;
         }
-        if (node.as<std::string>() == "FLOAT")
+        if (node.as<std::string>() == "continuous")
```
flomnes: you don't need to convert to `std::string` N times, you can convert it once and for all (l. 64, l. 68, l. 72). The compiler may optimize this away, not sure though

src/tests/src/solver/modelParser/enum_operators.h
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Antares::Solver::ObjectModel
+{
+inline std::ostream& operator<<(std::ostream& os, const ValueType& value_type)
+{
+    using namespace std::string_literals;
+    switch (value_type)
+    {
+    case ValueType::FLOAT:
+        os << "FLOAT"s;
```
flomnes:
```suggestion
        os << "float"s;
```

src/tests/src/solver/modelParser/enum_operators.h
```diff
@@ -0,0 +1,76 @@
...
+        break;
+    case ValueType::INTEGER:
+        os << "INTEGER"s;
```
flomnes:
```suggestion
        os << "integer"s;
```

src/tests/src/solver/modelParser/enum_operators.h
```diff
@@ -0,0 +1,76 @@
...
+        break;
+    case ValueType::BOOL:
+        os << "BOOL"s;
```
flomnes:
```suggestion
        os << "boolean"s;
```

src/tests/src/solver/modelParser/testModelParser.cpp
```diff
@@ -230,7 +232,7 @@ BOOST_AUTO_TEST_CASE(model_can_contain_multiple_parameters)
                   description: "model_description"
                   parameters:
                       - id: "param_name1"
-                        time-dependent: false
+                        time-dependent: FALSE
                         scenario-dependent: false
```
flomnes, line 236: I find it odd to mix FALSE, false, TRUE, true

flomnes (commented):
Please harmonize true/TRUE, FLOAT/continuous, etc.

src/tests/src/solver/modelParser/testModelParser.cpp
pet-mit: It seems to me that he is just making use of and testing all the possibilities offered by the yml parser by default. IMO we can keep them for now, and if we need to limit the possibilities, do it in another PR ?

JasonMarechal25: That's it

pet-mit (approved):
OK for me if it's of with @flomnes

src/tests/src/solver/modelParser/test_full.cpp
```diff
@@ -0,0 +1,510 @@
...
+using namespace std::string_literals;
+using namespace Antares::Solver;
+
+void checkParameter(const ObjectModel::Parameter& parameter,
+                    const std::string& name,
+                    bool timeDependent,
+                    bool scenarioDependent,
+                    ObjectModel::ValueType type)
+{
+    std::cout << "Parameter: " << parameter.Id() << std::endl;
```
flomnes, line 45:
```suggestion
```

src/tests/src/solver/modelParser/test_full.cpp
```diff
@@ -0,0 +1,510 @@
...
+    BOOST_CHECK_EQUAL(parameter.Type(), type);
+}
+
+void checkVariable(const ObjectModel::Variable& variable,
+                   const std::string& name,
+                   const std::string& lowerBound,
+                   const std::string& upperBound,
+                   ObjectModel::ValueType type)
+{
+    std::cout << "Variable: " << variable.Id() << std::endl;
```
flomnes, line 58:
```suggestion
```

src/tests/src/solver/modelParser/test_full.cpp
```diff
@@ -0,0 +1,510 @@
...
+    BOOST_CHECK_EQUAL(variable.LowerBound().Value(), lowerBound);
+    BOOST_CHECK_EQUAL(variable.UpperBound().Value(), upperBound);
+    BOOST_CHECK_EQUAL(variable.Type(), type);
+}
+
+void checkConstraint(const ObjectModel::Constraint& constraint,
+                     const std::string& name,
+                     const std::string& expression)
+{
+    std::cout << "Constraint: " << constraint.Id() << std::endl;
```
flomnes, line 69:
```suggestion
```

src/tests/src/solver/modelParser/enum_operators.h
```diff
@@ -0,0 +1,76 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#pragma once
+#include <iostream>
```
flomnes:
```suggestion
#include <ostream>
```

src/solver/modelParser/encoders.hxx
```diff
@@ -54,15 +61,16 @@ struct convert<Antares::Solver::ModelParser::ValueType>
         {
             return false;
         }
-        if (node.as<std::string>() == "FLOAT")
+        auto value = node.as<std::string>();
```
flomnes:
```suggestion
        const auto value = node.as<std::string>();
```

src/solver/modelParser/encoders.hxx
```diff
@@ -29,6 +29,13 @@
 // Implement convert specializations
 namespace YAML
 {
+
+template<typename T>
+inline T as_fallback_default(const Node& n)
+{
+    return n.as<T>(T());
```
flomnes, line 46: Can you add a comment to explain why we need this ?

flomnes (changes requested):
Minor remarks

src/tests/src/solver/modelParser/test_full.cpp
JasonMarechal25: Helpful when an error occurs. Else we don't really know what failed.

flomnes (approved):

PR #2449, opened 2024-10-01
Fix a crash occurring when first year not played

guilpier-code:
Sometimes, a unit test from **simple-study.cpp** won't pass. 
This test is named **two_MC_years__thermal_cluster_fullfills_area_demand_on_2nd_year_as_well**.

**Cause** : 
when executing code (see **solver.hxx**) : 
```c++
auto task = std::make_shared<yearJob<ImplementationType>>(
        ...
        y,
        ...
        numSpace,
        ...
        state[numSpace],
        ...);
```
When a year **y** is inactive in the playlist, variable **numSpace** is 99999, and `state[numSpace]` is undefined.

**More thoughts** : 
- this crash was first introduced in PR #2128 
- Associated commit : **be3301cdb3b2b7ed232db72e68a874c353d51609**
- It's unclear why we did not catch this crash sooner. Test **two_MC_years__thermal_cluster_fullfills_area_demand_on_2nd_year_as_well** should have failed already.
  Anyway, it fails at every run on **Windows** and in **debug** mode

### DIFF
--- a/src/solver/simulation/include/antares/solver/simulation/solver.hxx
+++ b/src/solver/simulation/include/antares/solver/simulation/solver.hxx
@@ -51,7 +51,7 @@ public:
             randomNumbers& pRandomForParallelYears,
             bool pPerformCalculations,
             Data::Study& pStudy,
-            Variable::State& pState,
+            std::vector<Variable::State>& pStates,
             bool pYearByYear,
             Benchmarking::DurationCollector& durationCollector,
             IResultWriter& resultWriter,
@@ -65,14 +65,14 @@ public:
         randomForParallelYears(pRandomForParallelYears),
         performCalculations(pPerformCalculations),
         study(pStudy),
-        state(pState),
+        states(pStates),
         yearByYear(pYearByYear),
         pDurationCollector(durationCollector),
         pResultWriter(resultWriter),
         simulationObserver_(simulationObserver),
         hydroManagement(study.areas, study.parameters, study.calendar, resultWriter)
     {
-        scratchmap = study.areas.buildScratchMap(numSpace);
+        // scratchmap = study.areas.buildScratchMap(numSpace);
     }
 
     yearJob(const yearJob&) = delete;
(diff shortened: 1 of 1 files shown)


flomnes:
@guilpier-code Good catch ! Can you point an action run that failed ? I'm surprised we didn't catch this bug earlier.

Also, could you handle the formatting issue ?

guilpier-code:
> @guilpier-code Good catch ! Can you point an action run that failed ? I'm surprised we didn't catch this bug earlier.

Please see the current PR's presentation comment. I updated it to answer your question.

> Also, could you handle the formatting issue ?

Done

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -65,14 +65,14 @@ public:
...
-        state(pState),
+        states(pStates),
         yearByYear(pYearByYear),
         pDurationCollector(durationCollector),
         pResultWriter(resultWriter),
         simulationObserver_(simulationObserver),
         hydroManagement(study.areas, study.parameters, study.calendar, resultWriter)
     {
-        scratchmap = study.areas.buildScratchMap(numSpace);
+        // scratchmap = study.areas.buildScratchMap(numSpace);
```
flomnes:
```suggestion
```

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -89,13 +89,13 @@ private:
...
     Data::Study& study;
-    Variable::State& state;
+    std::vector<Variable::State>& states;
     bool yearByYear;
     Benchmarking::DurationCollector& pDurationCollector;
     IResultWriter& pResultWriter;
     std::reference_wrapper<ISimulationObserver> simulationObserver_;
     HydroManagement hydroManagement;
-    Antares::Data::Area::ScratchMap scratchmap;
+    // Antares::Data::Area::ScratchMap scratchmap;
```
flomnes:
```suggestion
```

flomnes (commented):
Remove commented-out code

src/solver/simulation/include/antares/solver/simulation/solver.hxx
guilpier-code: Done

src/solver/simulation/include/antares/solver/simulation/solver.hxx
guilpier-code: Done

flomnes (approved):

PR #2451, opened 2024-10-04
Remove timeseries refresh [ANT-1335]

payetvin:
Remove the refresh feature: no known users, add unnecessary complexity to the code

TODO: 
- [x] Remove parallel test activation for every commit
- [x] Add doc


JasonMarechal25 (changes requested):
Only missing documentation update for refresh

flomnes (changes requested):
LGTM
We should exit with an error if key `refreshtimeseries` is not empty (change of behavior)

See #2838

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -286,53 +286,67 @@ void ISimulation<ImplementationType>::run()
...
 
-    // Export ts-numbers into output
-    TimeSeriesNumbers::StoreTimeSeriesNumbersIntoOuput(study, pResultWriter);
+        regenerateTimeSeries(0);
 
-    if (not ImplementationType::simulationBegin())
-    {
-        return;
+        // Destroy the TS Generators if any
+        // It will export the time-series into the output in the same time
```
flomnes: Consider a function rename for more expressiveness
```suggestion
        // It will export the time-series into the output at the same time
```

JasonMarechal25 (approved):

flomnes (approved):

PR #2450, opened 2024-10-02
Remove direct use of Sirius solver (ANT-1152)

guilpier-code:
This PR intends to be a first work in order to remove the direct call of **Sirius** as a solver, **regarding the optimization weekly problem only**.
We want to keep the use of **Sirius**, but only called through **or-tools** interface.
The associated **gopro** ticket is [this one](https://gopro-tickets.rte-france.com/browse/ANT-1152).
Here we intend to : 
- remove the use of command line option **--use-ortools**
- rename current command line option **--ortools-solver** into **--solver**
- handle direct consequences of these changes.
- adapt legacy GUI (especially remove **use ortools** check box)

So this PR does not intend to be thorough : these first changes leave remaining dead code to be removed 
and renaming to be done. This further changes will be done in a coming PR.

**Remaining to be done** : 
- or-tools (even if we use **Sirius** behind it) does not print correctly MPS for named problems : optimization variables names are composed with time steps, set once for all (0, ..., 167) before running any year, and not updated at each week (we could expect step numbers to be inside [0, ..., 8759] depending on the week). 
  So we left the code that print these correctly. We'll have to find a solution for this.
  We could leave the responsibility of printing these named MPS to Antares code (as it did when Sirius was run alone) for a moment
- Banish all direct calls to **Sirius** from **Antares Simulator** (calls to **Sirius** are done in **hydro heuristics**, **adequacy patch CSR** and annual **quadratic** post-treatments of flows) does not seem to be suitable and urgent for now.
- What else ?

PR #2452, opened 2024-10-04
Call to solver : trial for simplification

guilpier-code:
This PR follows PR #2450 (that removes direct call to **Sirius** and forces Sirius behind **ortools**).
This is a trial for simplifications of code calling the optimization weekly solver call (mainly in **opt_appel_solveur_lineaire.cpp**)

What was done : 
- Correlate **solver** destruction (delete) with setting **solver** to null
  (to keep in mind : deleting a **nullptr** is harmless)
- what else ?

To be done : 
- in **ORTOOLS_Simplexe(...)**, instead of updating `Probleme->ExistenceDUneSolution` and use it in , we could use solver (type **MPSolver**) returned status.
# CAUTION : 
Remove changes on CI yaml files !

### DIFF
--- a/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
+++ b/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
@@ -91,10 +91,6 @@ static SimplexResult OPT_TryToCallSimplex(const OptimizationOptions& options,
     assert(opt >= 0 && opt < 2);
     OptimizationStatistics& optimizationStatistics = problemeHebdo->optimizationStatistics[opt];
     TIME_MEASURE timeMeasure;
-    if (!PremierPassage)
-    {
-        solver = nullptr;
-    }
 
     if (solver == nullptr)
     {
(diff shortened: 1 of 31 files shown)


a-zakir:
> * Correlate **solver** destruction (delete) with setting **solver** to null
>       (to keep in mind : deleting a **nullptr** is harmless)
> 
>     * what else ?

i'd say building unique_ptr from MPSolverFactory so we don't have to call ORTOOLS_LibererProbleme.
if resource release is important you could use std::unique_ptr<T,Deleter>::release

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -212,26 +196,15 @@ static SimplexResult OPT_TryToCallSimplex(const OptimizationOptions& options,
     ProblemeAResoudre->ExistenceDUneSolution = Probleme.ExistenceDUneSolution;
     if (ProblemeAResoudre->ExistenceDUneSolution != OUI_SPX && PremierPassage)
     {
-        if (ProblemeAResoudre->ExistenceDUneSolution != SPX_ERREUR_INTERNE)
```
guilpier-code, line 245: We used to test `ExistenceDUneSolution != SPX_ERREUR_INTERNE`.
But `ExistenceDUneSolution == SPX_ERREUR_INTERNE` is impossible : indeed, ExistenceDUneSolution is given by **ortools** now, which does not know value **SPX_ERREUR_INTERNE**.

flomnes:
This PR has become obsolete, I will close it for now

PR #2463, opened 2024-10-14
batchrun forwards options to antares-solver [ANT-2314]

flomnes:
So there's no need to batchrun when/if new options appear for antares-solver.


sylvlecl (approved):
The only drawback is that we lose the help message

PR #2468, opened 2024-10-21
8.8: Don't apply reverse spinning if the cluster is no force gen [ANT-2293]

payetvin:


JasonMarechal25:
Is it possible to add a test (cucumber or unit) on this behavior ? Ideally We'd like to see it fail before your correction and pass after.

PR #2470, opened 2024-10-23
Reserve model : trial for a fix on reserve model

guilpier-code:
This work is a trial to fix a crash occurring when running **Antares Simulator** on a study named **BP23_A-Reference_2031**.
This crash was reported by a student supervised by @thibaulttoujouse.

The current work does not fix the crash, but is worth to be integrated in the reserve model work.

What was done : 
- **commit 1** : fix on a compilation issue on **Windows** + **debug**
- **commit 2** : cherry pick of a crash fix commit from another branch to see if it fixes the crash on study **BP23_A-Reference_2031**.

### DIFF
--- a/src/libs/antares/utils/include/antares/utils/utils.h
+++ b/src/libs/antares/utils/include/antares/utils/utils.h
@@ -36,6 +36,9 @@ namespace Antares
 */
 template<class StringT>
 void TransformNameIntoID(const AnyString& name, StringT& out);
+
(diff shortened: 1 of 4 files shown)


guilpier-code:
The CI fails on Windows and Linux for now.
We should see why (if needed) : 
- is it due to changes introduced by this commit ?
- is it due to base branch content ?

PR #2471, opened 2024-10-23
Modeler 4.4: expression parsing [ANT-2313]

payetvin:
This pull request includes several updates and improvements to the ANTLR grammar and related files for the `Expr` language. The changes involve modifications to the grammar rules, updates to the CMake configuration, and synchronization with a newer version of ANTLR.

### Grammar and Parsing Improvements:
* Refactored `Expr.g4` grammar to introduce new rules and improve parsing capabilities, including the addition of `atom`, `shift_expr`, and `right_expr` rules. [[1]](diffhunk://#diff-9904320a1807a7a3514cd161403920248cbde82e6c2dbc9cf855c5c134dd62dcL20-R68) [[2]](diffhunk://#diff-9904320a1807a7a3514cd161403920248cbde82e6c2dbc9cf855c5c134dd62dcL45-L48)
* Updated `Expr.tokens` to reflect new token definitions and changes in the grammar.
* Removed redundant rules and tokens from `Expr.g4` and `Expr.tokens`. [[1]](diffhunk://#diff-9904320a1807a7a3514cd161403920248cbde82e6c2dbc9cf855c5c134dd62dcL45-L48) [[2]](diffhunk://#diff-cb10c1b9428e9e48c48e524757dfc48dcc6ae8dd2ad921eff65851cbc3c2e5c1L10-R30)

### Build Configuration Updates:
* Updated `sonar-project.properties` to exclude `src/libs/antares/antlr-interface` from Sonar analysis. antlr-interface contains generated code that trigger defect in analysis
* Modified `CMakeLists.txt` to always include the `antlr-interface` directory and removed conditional logic for building with ANTLR4. [[1]](diffhunk://#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4L186-L188) [[2]](diffhunk://#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4R194-R195) [[3]](diffhunk://#diff-9e299f60c14464c86511d6c9a4e7c081765abb4840b57ea4dc25238311006ce9L31-L33)

### Synchronization with ANTLR 4.13.2:
* Updated generated files to be compatible with ANTLR 4.13.2, including `ExprBaseVisitor.cpp`, `ExprBaseVisitor.h`, and `ExprLexer.cpp`. [[1]](diffhunk://#diff-bda2e00a884ff4b113d535c8239b493bfc88c5c4f1d38c4687c2f28e44cdfde8L2-R2) [[2]](diffhunk://#diff-abcc97c1794721be0f33a540339645ac1d5d9d3dd6cbe941235200bad3dcacb9L2-R2) [[3]](diffhunk://#diff-6adabe5fce15f2c6367db67aa33a2e02bb220796c9ff4e143160752a637ca3c0L2-R2)
* Adjusted `ExprLexer.cpp` to use `std::unique_ptr` for static data initialization. [[1]](diffhunk://#diff-6adabe5fce15f2c6367db67aa33a2e02bb220796c9ff4e143160752a637ca3c0L48-R48) [[2]](diffhunk://#diff-6adabe5fce15f2c6367db67aa33a2e02bb220796c9ff4e143160752a637ca3c0L61-R62)

### Visitor Pattern Enhancements:
* Updated `ExprBaseVisitor` to include new visit methods for the added grammar rules. [[1]](diffhunk://#diff-abcc97c1794721be0f33a540339645ac1d5d9d3dd6cbe941235200bad3dcacb9L22-R57) [[2]](diffhunk://#diff-abcc97c1794721be0f33a540339645ac1d5d9d3dd6cbe941235200bad3dcacb9L58-R106)

These changes collectively enhance the flexibility and maintainability of the grammar and build configuration, while ensuring compatibility with the latest ANTLR version.


src/libs/antares/antlr-interface/ExprBaseVisitor.h
```diff
@@ -1,5 +1,5 @@
 
-// Generated from Expr.g4 by ANTLR 4.13.1
+// Generated from Expr.g4 by ANTLR 4.7.2
```
a-zakir: use v 4.13.1?

a-zakir:
add antlr4 [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/2fc326517d3b5174212df46aed3794825d62028d/src/vcpkg.json#L40)

src/solver/modelConverter/include/antares/solver/modelConverter/convertorVisitor.h
```diff
@@ -0,0 +1,75 @@
...
+#include "antares/solver/modelParser/Library.h"
+
+namespace Antares::Solver::ModelConverter
+{
+
+Nodes::Node* convertExpressionToNode(const std::string& exprStr,
+                                     Registry<Nodes::Node>& registry,
+                                     const ModelParser::Model& model);
+
+// Visitor to convert nodes to Antares::Solver::Nodes
```
pet-mit:
```suggestion
// Visitor to convert ANTLR expressions to Antares::Solver::Nodes
```

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+    {
+        if (param.id == context->getText())
+        {
+            is_parameter = true;
+            break;
+        }
+    }
+
+    return (is_parameter) ? static_cast<Node*>(registry_.create<ParameterNode>(context->getText()))
+                          : static_cast<Node*>(registry_.create<VariableNode>(context->getText()));
```
pet-mit: in case it is not a parameter, I think it would be better to check that it is a variable indeed. else, throw an exception. this will make the model parsing throw at parse time instead of at runtime

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+    auto* right = toNodePtr(visit(context->expr(1)));
+
+    std::string op = context->op->getText();
+    return (op == "+") ? static_cast<Node*>(registry_.create<SumNode>(left, right))
+                       : static_cast<Node*>(registry_.create<SubtractionNode>(left, right));
+}
+
+std::any ConvertorVisitor::visitPortField([[maybe_unused]] ExprParser::PortFieldContext* context)
+{
+    return std::any();
```
pet-mit:
```suggestion
    // TODO implement this
    throw ...
```

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+std::any ConvertorVisitor::visitTimeSum([[maybe_unused]] ExprParser::TimeSumContext* context)
+{
+    return std::any();
+}
+
+std::any ConvertorVisitor::visitAllTimeSum([[maybe_unused]] ExprParser::AllTimeSumContext* context)
+{
+    return std::any();
+}
```
pet-mit: same

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+  [[maybe_unused]] ExprParser::RightMuldivContext* context)
+{
+    return std::any();
+}
+
+std::any ConvertorVisitor::visitSignedExpression(
+  [[maybe_unused]] ExprParser::SignedExpressionContext* context)
+{
+    return std::any();
+}
```
pet-mit: same

src/tests/src/solver/modelParser/testModelTranslator.cpp
pet-mit, line 1: add asserts to this class to check that expressions are converted to Node*

pet-mit (changes requested):

src/tests/cucumber/behave.ini
pet-mit, line 1: remove this file

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -69,20 +70,35 @@ std::any ConvertorVisitor::visit(antlr4::tree::ParseTree* tree)
     return tree->accept(this);
 }
 
+class NoParameterOrVariableWithThisName: public std::exception
+{
+public:
+    const char* what() const noexcept override
+    {
+        return "No parameter or variable found for this identifier";
```
pet-mit: add identifier to exception message?

pet-mit (changes requested):

sonar-project.properties
```diff
@@ -12,8 +12,8 @@ sonar.sources=src
 sonar.tests=src/tests
 sonar.sourceEncoding=UTF-8
 
-sonar.exclusions=src/ext/**,src/tests/**,src/ui/**
-sonar.coverage.exclusions=src/ext/**,src/tests/**,src/analyzer/**,src/distrib/**,src/tools/**,src/ui/**
+sonar.exclusions=src/ext/**,src/tests/**,src/ui/**,src/libs/antares/antlr-interface/**
```
guilpier-code: _Cosmetic_ 
We should have one line per item (if possible), it would be more clear, and easier for a human eye to browse it

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,283 @@
...
+    std::string op = context->op->getText();
+    return (op == "+") ? static_cast<Node*>(registry_.create<SumNode>(left, right))
+                       : static_cast<Node*>(registry_.create<SubtractionNode>(left, right));
+}
+
+// TODO implement this
+std::any ConvertorVisitor::visitPortField([[maybe_unused]] ExprParser::PortFieldContext* context)
+{
+    throw std::runtime_error(std::string("Node portfield not implemented yet"));
+    return std::any();
```
a-zakir: the return is unreachable

a-zakir: why this method is not implemented?

a-zakir:
https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2475

src/solver/modelConverter/convertorVisitor.cpp
JasonMarechal25: A lot of feature are out of scope of this PR and will be implemented in subsequent PRs

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,283 @@
...
+        msg_ = "No parameter or variable found for this identifier: " + id;
+    }
+
+    const char* what() const noexcept override
+    {
+        return msg_.c_str();
+    }
+
+private:
+    std::string msg_;
```
flomnes:
```suggestion
    NoParameterOrVariableWithThisName(const std::string& id) : std::exception::exception("No parameter or variable found for this identifier: " + id)
    {
    }
```

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    auto* node = ModelConverter::convertExpressionToNode("", registry, model);
+    BOOST_CHECK_EQUAL(node, nullptr);
+}
+
+BOOST_FIXTURE_TEST_CASE(negation, Fixture)
+{
+    std::string expression = "-7";
+    auto* n = ModelConverter::convertExpressionToNode(expression, registry, model);
+    BOOST_CHECK_EQUAL(n->name(), "NegationNode");
+    auto* nodeNeg = dynamic_cast<Nodes::NegationNode*>(n);
```
flomnes, line 57: You need to check the result of `dynamic_cast` too

flomnes: You need to check the result of `dynamic_cast` too

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    auto* sub = registry.create<Nodes::SubtractionNode>(l4, l1);
+    auto* mult = registry.create<Nodes::MultiplicationNode>(l12, sub);
+    auto* sum1 = registry.create<Nodes::SumNode>(mult, param);
+    auto* sum2 = registry.create<Nodes::SumNode>(l42, l3);
+    auto* sum3 = registry.create<Nodes::SumNode>(sum2, var);
+    auto* neg = registry.create<Nodes::NegationNode>(sum3);
+    auto* div = registry.create<Nodes::DivisionNode>(sum1, neg);
+
+    Visitors::CompareVisitor cmp;
+    BOOST_CHECK(cmp.dispatch(n, div));
```
flomnes: 🥇

flomnes (commented):
We need to catch errors better (syntax errors, etc.), and add tests for failures

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+#define WIN32_LEAN_AND_MEAN
+
+#include <boost/test/unit_test.hpp>
+
+#include "antares/solver/expressions/Registry.hxx"
```
guilpier-code, line 26: None of : 
- `#include "antares/solver/expressions/visitors/AstDOTStyleVisitor.h"`
- `#include "antares/solver/libObjectModel/library.h"`

is required here. Both should be removed.

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,283 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#include <iostream>
+
+/* #include <antares/logs/logs.h> */
```
guilpier-code: `/* #include <antares/logs/logs.h> */` : should be removed

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+};
+
+static Nodes::LiteralNode* toLiteral(Nodes::Node* n)
+{
+    return dynamic_cast<Nodes::LiteralNode*>(n);
+}
+
+BOOST_FIXTURE_TEST_CASE(empty_expression, Fixture)
+{
+    auto* node = ModelConverter::convertExpressionToNode("", registry, model);
```
guilpier-code: About the **registry** : 
As you know, the **registry** comes with a tree, it makes sure we avoid memory leaks when destroying the tree (so, a registry and the associated tree should have the same life time and be more dependent, but this is another matter that should be discussed at some point).
On the other hand, we have unit tests converting an expression into a tree and checking the created tree.

**These unit tests (and any calling code) don't have to and don't want to see the registry.**

**Suggestion** : 
instead of a function, make the converter a **class**, and define an instance of this class in the **fixture**, and run the converter in the tests : 
```C++
struct Fixture
{
    ModelParser::Model model;
    Antares::Solver::Registry<Antares::Solver::Nodes::Node> registry;
    ModelConverter::ExpressionToNodeConverter converter(model, registry);
};
...

BOOST_FIXTURE_TEST_CASE(empty_expression, Fixture)
{
    auto* node = converter.run("");
    BOOST_CHECK_EQUAL(node, nullptr);
}
...
```
As you can see, there is no **registry** in tests, only an expression, a converter and a generated tree.

guilpier-code (changes requested):

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    BOOST_CHECK_EQUAL(node, nullptr);
+}
+
+BOOST_FIXTURE_TEST_CASE(negation, Fixture)
```
guilpier-code: _Crash_
On **Windows** + **debug**, test **negation** crashes.

**After investigation** : it turns out that changing the name of the fixture (from **Fixture** into **ExpressionIntoNode** for instance) causes the crash to disappear.
In the end, we can notice that **testModelTranslator.cpp** also contains a fixture named **Fixture**, and at linking target **TestModelParser**, the 2 fixtures are exchanged in some way.

guilpier-code (changes requested):

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+{
+    std::string expression = "-7";
+    auto* n = ModelConverter::convertExpressionToNode(expression, registry, model);
+    BOOST_CHECK_EQUAL(n->name(), "NegationNode");
+    auto* nodeNeg = dynamic_cast<Nodes::NegationNode*>(n);
+    BOOST_CHECK_EQUAL(toLiteral(nodeNeg->child())->value(), 7);
```
guilpier-code, line 70: result of **toLiteral(Node\* n)** could be **nullptr** as well. 
So a check is also required here.

guilpier-code (changes requested):

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+}
+
+BOOST_FIXTURE_TEST_CASE(identifier, Fixture)
+{
+    ModelParser::Library library;
```
guilpier-code: `ModelParser::Library library;` : unused, to be removed

guilpier-code (changes requested):

src/solver/modelConverter/include/antares/solver/modelConverter/convertorVisitor.h
```diff
@@ -0,0 +1,74 @@
...
+    std::any visitUnsignedAtom(ExprParser::UnsignedAtomContext* context) override;
+    std::any visitRightAtom(ExprParser::RightAtomContext* context) override;
+    std::any visitSignedExpression(ExprParser::SignedExpressionContext* context) override;
+    std::any visitShiftAddsub(ExprParser::ShiftAddsubContext* context) override;
+    std::any visitShiftMuldiv(ExprParser::ShiftMuldivContext* context) override;
+    std::any visitRightMuldiv(ExprParser::RightMuldivContext* context) override;
+    std::any visitRightExpression(ExprParser::RightExpressionContext* context) override;
+
+    Registry<Nodes::Node>& registry_;
+    const ModelParser::Model& model_;
```
guilpier-code: These 2 data members should be **private** (make them private still allow a successful compilation)

guilpier-code (changes requested):

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+      .constraints = {},
+      .objective = "objectives"};
+
+    std::string expression = "abc"; // not a param or var
+    BOOST_CHECK_EXCEPTION(ModelConverter::convertExpressionToNode(expression, registry, model0),
+                          std::exception,
+                          expectedMessage);
+}
+
+BOOST_FIXTURE_TEST_CASE(AddSub, Fixture)
```
guilpier-code: There are 2 unit tests here.
Making 2 unit tests instead of one would allow to be more specific and clear in the name of tests

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    expression = "6 - 3";
+    n = ModelConverter::convertExpressionToNode(expression, registry, model);
+    BOOST_CHECK_EQUAL(n->name(), "SubtractionNode");
+
+    auto* nodeSub = dynamic_cast<Nodes::SubtractionNode*>(n);
+    BOOST_CHECK_EQUAL(toLiteral(nodeSub->left())->value(), 6);
+    BOOST_CHECK_EQUAL(toLiteral(nodeSub->right())->value(), 3);
+}
+
+BOOST_FIXTURE_TEST_CASE(mulDiv, Fixture)
```
guilpier-code: Same as above : there 2 unit tests here.

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include "antares/solver/expressions/Registry.hxx"
+#include "antares/solver/expressions/visitors/AstDOTStyleVisitor.h"
+#include "antares/solver/expressions/visitors/CompareVisitor.h"
+#include "antares/solver/libObjectModel/library.h"
+#include "antares/solver/modelConverter/convertorVisitor.h"
+#include "antares/solver/modelConverter/modelConverter.h"
+#include "antares/solver/modelParser/Library.h"
+
+using namespace Antares::Solver;
```
guilpier-code, line 31: About `using namespace Antares::Solver;` : if we're announcing we're using a namespace, then why specify it in the **Fixture** ?
```C++
Antares::Solver::Registry<Antares::Solver::Nodes::Node> registry;
```
should become : 
```C++
Registry<Nodes::Node> registry;
```
shorter.

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+BOOST_FIXTURE_TEST_CASE(negation, Fixture)
+{
+    std::string expression = "-7";
+    auto* n = ModelConverter::convertExpressionToNode(expression, registry, model);
+    BOOST_CHECK_EQUAL(n->name(), "NegationNode");
+    auto* nodeNeg = dynamic_cast<Nodes::NegationNode*>(n);
+    BOOST_CHECK_EQUAL(toLiteral(nodeNeg->child())->value(), 7);
+}
+
+BOOST_FIXTURE_TEST_CASE(identifier, Fixture)
```
guilpier-code: In this test, we don't need the **Model** coming with the fixture.
And there are many tests like this.
I recommend we define a **Model** inside the test (and remove Model from the fixture).
If we do so, only the registry remains in the fixture, so we should remove the fixture, and pass the **registry** to the unit tests directly.

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    expression = "6 / 3";
+    n = ModelConverter::convertExpressionToNode(expression, registry, model);
+    BOOST_CHECK_EQUAL(n->name(), "DivisionNode");
+
+    auto* nodeDiv = dynamic_cast<Nodes::DivisionNode*>(n);
+    BOOST_CHECK_EQUAL(toLiteral(nodeDiv->left())->value(), 6);
+    BOOST_CHECK_EQUAL(toLiteral(nodeDiv->right())->value(), 3);
+}
+
+BOOST_FIXTURE_TEST_CASE(comparison, Fixture)
```
guilpier-code: There are 3 unit tests in this tests

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+bool expectedMessage(const std::exception& ex)
+{
+    BOOST_CHECK_EQUAL(ex.what(),
+                      std::string("No parameter or variable found for this identifier: abc"));
+    return true;
+}
+
+BOOST_FIXTURE_TEST_CASE(identifierNotFound, Fixture)
+{
+    ModelParser::Library library;
```
guilpier-code: `ModelParser::Library library;` : not needed

src/tests/src/solver/modelParser/testConvertorVisitor.cpp
```diff
@@ -0,0 +1,200 @@
...
+    ModelParser::Model model0{
```
guilpier-code: Do we need a **model** as complex as that for this test ?
It seems we don't.
We should replace this model with a very simple one.

guilpier-code (changes requested):

src/tests/src/solver/modelParser/testModelTranslator.cpp
```diff
@@ -33,36 +32,39 @@
```
guilpier-code: `#include <iostream>` : not needed here. To be removed.

src/tests/src/solver/modelParser/testModelTranslator.cpp
```diff
@@ -33,36 +32,39 @@
...
 {
     ModelParser::Library library;
-    ObjectModel::Library lib = ModelConverter::convert(library);
+    Registry<Nodes::Node> registry;
+};
+
+// Test empty library
+BOOST_FIXTURE_TEST_CASE(Empty_library_is_valid, Fixture)
+{
+    ObjectModel::Library lib = ModelConverter::convert(library, registry);
```
guilpier-code: **_Design_** :
```C++
ObjectModel::Library lib = ModelConverter::convert(library, registry);
```
Here we have a parsed library, and we want to convert it into an object Library. 
So we a source and a target.
Why would we need a **registry** here ? It seems unrelated to what we're doing now in the unit test.

Ok, it's used in the depth of the code when we build tree / expressions (associated to variables , constraints and objective) form **std::string**. But do we have to pass it all the way down to where it's needed ?
The **registry** should be hidden and created either : 
- to where it's really needed (in classes **Variables**, **Constraints**, ...)
- directly in the top node of the AST tree it's supposed be associated with

But not here, in my opinion.
Here we should have : 
```C++
ObjectModel::Library lib = ModelConverter::convert(library);
```

guilpier-code (changes requested):

src/tests/src/solver/modelParser/testModelTranslator.cpp
pet-mit: maybe hide one Registry instance inside each Expression instance ?

src/solver/modelConverter/include/antares/solver/modelConverter/convertorVisitor.h
```diff
@@ -0,0 +1,74 @@
...
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#pragma once
+
+#include <ExprVisitor.h>
+
+#include <antares/solver/expressions/Registry.hxx>
+#include <antares/solver/expressions/nodes/ExpressionsNodes.h>
```
guilpier-code:
```C++
#include <antares/solver/expressions/nodes/ExpressionsNodes.h>
```
In theory, instead of this inclusion, we need : 
```C++
#include <antares/solver/expressions/nodes/Node.h>
```
As only the declaration of **Nodes::Node** is required.

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+namespace Antares::Solver::ModelConverter
+{
+
+using namespace Antares::Solver::Nodes;
+
+static Node* toNodePtr(const std::any& a)
+{
+    return std::any_cast<Node*>(a);
+}
```
guilpier-code: Is this function needed (since a simple call to **std::any_cast** is supposed to be expressive and short enough) ?

src/solver/modelConverter/convertorVisitor.cpp
```diff
@@ -0,0 +1,283 @@
...
+    for (const auto& var: model_.variables)
+    {
+        if (var.id == context->getText())
+        {
+            return static_cast<Node*>(registry_.create<VariableNode>(var.id));
+        }
+    }
+
+    /* logs.error() << "No parameter or variable found with this name: " << context->getText(); */
+    throw NoParameterOrVariableWithThisName(context->getText());
```
guilpier-code, line 129:
```C++
throw NoParameterOrVariableWithThisName(context->getText());
```
or we could throw a **std exception** instead (`std::runtime_error` ?) constructible with a message.
And we wouldn't need class **NoParameterOrVariableWithThisName**

src/solver/libModelObject/include/antares/solver/libObjectModel/expression.h
```diff
@@ -30,8 +32,9 @@ class Expression
 public:
     Expression() = default;
 
-    explicit Expression(std::string value):
-        value_(std::move(value))
+    explicit Expression(std::string value, Nodes::Node* node):
```
guilpier-code: _Cosmetic_ : 
`Nodes::Node* node` ==> `Nodes::Node* topNode` ?

src/solver/expressions/visitors/EvalVisitor.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include "antares/solver/expressions/visitors/EvalVisitor.h"
 
+#include <cmath>
```
guilpier-code, line 24: `#include <cmath>` : we don't need that

guilpier-code (dismissed):

src/solver/modelConverter/include/antares/solver/modelConverter/convertorVisitor.h
flomnes: Nope, `Registry` needs the full definition. Having an incomplete type doesn't work.

flomnes: No, we use e.g `ParameterNode` here

src/solver/modelConverter/convertorVisitor.cpp
flomnes: It can be removed, right

src/solver/modelConverter/convertorVisitor.cpp
flomnes: yes, I had the same idea and did exactly this, before reading this comment

src/solver/libModelObject/include/antares/solver/libObjectModel/expression.h
flomnes: root would be an even better name IMO

src/solver/expressions/visitors/EvalVisitor.cpp
flomnes: Actually it's required, @a-zakir will confirm this

src/solver/libModelObject/include/antares/solver/libObjectModel/expression.h
guilpier-code: I do agree, **rootNode** is better

src/solver/modelConverter/include/antares/solver/modelConverter/convertorVisitor.h
guilpier-code: > No, we use e.g `ParameterNode` here

**ParameterNode** is required in **convertorVisitor.cpp**, not in the current header file.

src/solver/expressions/visitors/EvalVisitor.cpp
guilpier-code: Why is it required for in this file ?
Sorry, I can't see.

a-zakir: to get std::isfinite for Ubuntu

guilpier-code: oops, sorry.

src/tests/src/solver/modelParser/testModelTranslator.cpp
flomnes: See #2501 for a fix

flomnes (approved):

pet-mit (approved):

PR #2473, opened 2024-10-29
Add new solver documentation

pet-mit:


flomnes (commented):
I'm not a fan of the naming convention "legacy" / "new". Here are a few suggestions
- fixed/configurable
- static/dynamic
- predefined/customized
- base/configured
- core/extension

What do you think ?

pet-mit:
> I'm not a fan of the naming convention "legacy" / "new". Here are a few suggestions
> 
> * fixed/configurable
> * static/dynamic
> * predefined/customized
> * base/configured
> * core/extension
> 
> What do you think ?

I quite like "static/dynamic"
(And I think we should avoid using synonyms of core/extension, because they imply that we need the "core" to be able to use the "extension", which is not true)
@tbittar thoughts?

tbittar:
> > I'm not a fan of the naming convention "legacy" / "new". Here are a few suggestions
> > 
> > * fixed/configurable
> > * static/dynamic
> > * predefined/customized
> > * base/configured
> > * core/extension
> > 
> > What do you think ?
> 
> I quite like "static/dynamic" (And I think we should avoid using synonyms of core/extension, because they imply that we need the "core" to be able to use the "extension", which is not true) @tbittar thoughts?

Static / dynamic is my favorite choice among the proposals

flomnes (approved):

PR #2475, opened 2024-10-29
Fix build errors caused by conflicts between windows.h and antlr4-runtime.h

a-zakir:


src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -269,5 +293,13 @@ class NodeVisitor: public IName
      * @return The result of processing the ComponentParameterNode.
      */
     virtual R visit(const Nodes::ComponentParameterNode*, Args... args) = 0;
+
+    void setNodeVisitorCustomLog(const NodeVisitorCustomLog& nodeVisitorCustomLog)
+    {
+        nodeVisitorCustomLog_ = nodeVisitorCustomLog;
+    }
```
guilpier-code: About : **setNodeVisitorCustomLog(...)**
is unused. 
Removal ?

guilpier-code (changes requested):

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: The client of this class has to use this method in order to customize logging;

For instance the client would create an  NodeVisitorCustomLog cust_log = {.info = my_info_log, .warning =etc...};
where my_info could be:

void my_info_log(const std::string& msg){
logs.info(msg);
}

src/solver/expressions/visitors/EvalVisitor.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include "antares/solver/expressions/visitors/EvalVisitor.h"
 
+#include <cmath>
```
guilpier-code, line 24: `#include <cmath>` : seems useless (removing it still allows a successful compilation)

guilpier-code (changes requested):

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -19,18 +19,33 @@
...
+    using LogFunction = std::function<void(const std::string&)>;
+
+    LogFunction info;
+    LogFunction warning;
+    LogFunction error;
+};
+
+void ToYuniInfo(const std::string& msg);
+
+void ToYuniError(const std::string& msg);
```
flomnes: This declaration is unused

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -269,5 +284,13 @@ class NodeVisitor: public IName
...
      */
     virtual R visit(const Nodes::ComponentParameterNode*, Args... args) = 0;
+
+    void setNodeVisitorCustomLog(const NodeVisitorCustomLog& nodeVisitorCustomLog)
+    {
+        nodeVisitorCustomLog_ = nodeVisitorCustomLog;
+    }
+
+private:
+    NodeVisitorCustomLog nodeVisitorCustomLog_ = RedirectToStandardOutputs();
```
flomnes:
```suggestion
    LogSink nodeVisitorCustomLog_ = RedirectToAntaresLogs();
```

src/solver/expressions/visitors/NodeVisitor.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+void ToYuniInfo(const std::string& msg)
+{
+    logs.info() << msg;
+}
+
+inline void ToYuniError(const std::string& msg)
+{
+    logs.error() << msg;
+}
```
flomnes:
```suggestion
static void ToYuniInfo(const std::string& msg)
{
    logs.info() << msg;
}

static void ToYuniError(const std::string& msg)
{
    logs.error() << msg;
}
```

src/solver/expressions/visitors/NodeVisitor.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+NodeVisitorCustomLog RedirectToStandardOutputs()
+{
+    return {.info = ToYuniInfo, .warning = ToYuniInfo, .error = ToYuniError};
```
flomnes: Please add a proper warning function

```suggestion
    return {.info = ToYuniInfo, .warning = ToYuniWarning, .error = ToYuniError};
```

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -119,7 +134,7 @@ class NodeVisitor: public IName
         }
         catch (std::exception&)
         {
-            logs.error() << "Antares::Solver::Visitor: could not visit the node!";
+            nodeVisitorCustomLog_.error("Antares::Solver::Visitor: could not visit the node!");
```
flomnes:
```suggestion
            log_.error("Antares::Solver::Visitor: could not visit the node!");
```

src/solver/expressions/visitors/NodeVisitor.cpp
flomnes, line 1: Please rename this function to something like logSink, etc.

src/solver/expressions/visitors/NodeVisitor.cpp
```diff
@@ -0,0 +1,41 @@
...
+{
+    logs.info() << msg;
+}
+
+inline void ToYuniError(const std::string& msg)
+{
+    logs.error() << msg;
+}
+
+NodeVisitorCustomLog RedirectToStandardOutputs()
```
flomnes:
```suggestion
LogSink RedirectToAntaresLogs()
```

flomnes (changes requested):
Thanks for this, it works

As @guilpier-code said, you can remove the setter since it's unused (for now)

Please add a comment explaining why we need a LogSink (antlr4-runtime.h & windows.h conflict)

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
```diff
@@ -19,18 +19,33 @@
 ** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
 */
 #pragma once
+#include <functional>
+#include <iostream>
 #include <optional>
 #include <typeindex>
 #include <vector>
 
-#include <antares/logs/logs.h>
```
guilpier-code, line 26: About `#include <antares/logs/logs.h>` : 
As we know, this is the inclusion that causes a compilation error on Windows.
Just to be clear, this header indirectly `#include <windows.h>`, that holds an incompatibility with **antlr4**.
As a proof, replacing `#include <antares/logs/logs.h>` with `#include <windows.h>` in the current **logs.h** causes the same compilation error.
Note that `#include <winsock2.h>` is also a cause for incompatibility, as this header includes **windows.h**

a-zakir:
> Thanks for this, it works
> 
> As @guilpier-code said, you can remove the setter since it's unused (for now)
> 
> Please add a comment explaining why we need a LogSink (antlr4-runtime.h & windows.h conflict)

in fact, the method is no longer needed

src/solver/expressions/include/antares/solver/expressions/visitors/NodeVisitor.h
a-zakir: [update] the default log is now yuni::logs, i can safely remove the setter

src/solver/expressions/visitors/EvalVisitor.cpp
a-zakir: done

src/solver/expressions/visitors/NodeVisitor.cpp
```diff
@@ -0,0 +1,41 @@
...
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+
+#include <antares/logs/logs.h>
+#include <antares/solver/expressions/visitors/NodeVisitor.h>
+
+namespace Antares::Solver::Visitors
+{
+void ToYuniInfo(const std::string& msg)
```
guilpier-code: We can do it like that.
Or we could define an interface class (with the same public methods as current **Yuni** logger), and create a **Yuni** implementation, internally calling the **Yuni** logger.

guilpier-code:
We identified 3 solutions in order to solve the compilation issue, due to the current conflict.
(the first 2 are quite the same, they count as one solution).
- Do as it's solved now (replacing direct call to **Yuni** logger with functions defined in a **.cpp** that includes the **Yuni** logger header **logs.h**)
- We also can create a logger interface, as suggested [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2475#discussion_r1843458761)
- change **Yuni** code, in order to include `<mutex>` instead of system custom headers related to multi threading.
- Replace **Yuni** logger with another one, compatible with **antlr**. This new logger should replace **Yuni** logger at some point anyway.

src/solver/expressions/visitors/NodeVisitor.cpp
a-zakir: Yes, same spirit as LogSink; the interface implies using a pointer or something like that, injected into the ctor? or from a setter?

flomnes: No need for an interface class at this point

flomnes (approved):

PR #2483, opened 2024-11-04
Unit test for reverse calculation

payetvin:
Add new test checks the behavior of the `thermal_cluster` when the generation behavior is forced to no generation.
Verify the `thermal_cluster` behavior when `tsGenBehavior` is set to `LocalTSGenerationBehavior::forceNoGen` and the time series is filled with a constant value.

### DIFF
--- a/src/tests/src/libs/antares/study/test_study.cpp
+++ b/src/tests/src/libs/antares/study/test_study.cpp
@@ -210,6 +210,8 @@ BOOST_FIXTURE_TEST_CASE(WithForceNoGenOptionTimeSeriesNotGeneratedForReverseSpin
                         ThermalClusterStudy)
 {
     cluster->tsGenBehavior = LocalTSGenerationBehavior::forceNoGen;
+    cluster->spinning = 0.9; // An arbitrary value != 1.0 is chosen here, otherwise the reverse
+                             // calculation of spinning does nothing
     auto& ts = cluster->series.timeSeries;
 
     ts.resize(1, 8760);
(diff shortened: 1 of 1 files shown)


src/tests/src/libs/antares/study/test_study.cpp
```diff
@@ -199,6 +199,16 @@ BOOST_FIXTURE_TEST_CASE(thermal_cluster_delete, ThermalClusterStudy)
     BOOST_CHECK(areaA->thermal.list.empty());
 }
 
+BOOST_FIXTURE_TEST_CASE(thermal_cluster_forceNoGen, ThermalClusterStudy)
```
JasonMarechal25: Can we find a better name for this test to better describe the behaviour we want to validate.
Maybe : WithForceNoGenOptionTimeSeriesNotGeneratedForReverseSpinning

src/tests/src/libs/antares/study/test_study.cpp
```diff
@@ -199,6 +199,16 @@ BOOST_FIXTURE_TEST_CASE(thermal_cluster_delete, ThermalClusterStudy)
...
+{
+    cluster->tsGenBehavior = LocalTSGenerationBehavior::forceNoGen;
+    cluster->series.timeSeries.resize(1, 8760);
+    cluster->series.timeSeries.fill(100);
+    cluster->reverseCalculationOfSpinning();
+
+    BOOST_CHECK_EQUAL(cluster->series[0][0], 100);
```
JasonMarechal25: Can you check all value, maybe with std::[ranges::]allof

JasonMarechal25 (approved):

src/tests/src/libs/antares/study/test_study.cpp
JasonMarechal25: _

JasonMarechal25 (changes requested):

JasonMarechal25 (approved):

flomnes (approved):

PR #2500, opened 2024-11-20
New System model (2.2) [ANT-2208]

pet-mit:
Add objects that represent the new model for the simulated system:
- a Component is a physical instance of a Model in the simulated system
- a System is a collection of Components


src/tests/src/study/system-model/test_main.cpp
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#define BOOST_TEST_MODULE systemModel
+#define WIN32_LEAN_AND_MEAN
+
+#include <boost/test/unit_test.hpp>
+
+std::function<bool(const std::exception&)> checkMessage(std::string expected_message)
```
a-zakir: defined twice!

src/tests/src/study/system-model/utils.h
```diff
@@ -0,0 +1,38 @@
...
+
+inline std::function<bool(const std::exception&)> checkMessage(std::string expected_message)
+{
+    auto predicate = [expected_message](const std::exception& e)
+    {
+        BOOST_CHECK_EQUAL(e.what(), expected_message);
+        return true;
+    };
+    return predicate;
+}
```
guilpier-code: This function **checkMessage** is (or is about to be) very much used, in fact each time we use **BOOST_CHECK_EXCEPTION**.
We should care about code duplication.

src/tests/src/study/system-model/utils.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Antares_Simulator is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
```
guilpier-code: **utils.h** is the name of the current file.
As we look at whet ii contains, it should be more specific.
Besides, we have too many files already named **utils.h**.

src/tests/src/study/system-model/test_component.cpp
```diff
@@ -0,0 +1,213 @@
...
+                                      Parameter("param2",
+                                                ValueType::FLOAT,
+                                                Parameter::TimeDependent::NO,
+                                                Parameter::ScenarioDependent::NO)});
+    }
+    auto model = model_builder.build();
+    return model;
+}
+
+BOOST_AUTO_TEST_SUITE(_Component_)
```
guilpier-code: Each time we call **BOOST_FIXTURE_TEST_CASE**, we call it with the same fixture (which is **ComponentTestFixture**).
We can avoid this by using **BOOST_FIXTURE_TEST_SUITE** instead of **BOOST_AUTO_TEST_SUITE**.

Before you decide to apply this remark (or not), please see other remarks. They may make this one obsolete.

src/tests/src/study/system-model/test_component.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+#include <boost/test/unit_test.hpp>
+
+#include "antares/study/system-model/component.h"
+
+#include "utils.h"
+
+using namespace Antares::Study::SystemModel;
+
+struct ComponentTestFixture
```
guilpier-code: **ComponentTestFixture** only creates a **Model**.
Why not name it struct **createModelFixture** then ?

src/tests/src/study/system-model/test_component.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+    auto model = model_builder.build();
+    return model;
+}
+
+BOOST_AUTO_TEST_SUITE(_Component_)
+
+BOOST_FIXTURE_TEST_CASE(nominal_build_with_parameters, ComponentTestFixture)
+{
+    Model model = createModel(true);
```
guilpier-code: Every test starts with : 
```c++
Model model = createModel(true); // or 'false'
ComponentBuilder component_builder;
auto component = component_builder.withModel(&model);
```
So it seems several things could be moved into a fixture, depending on how these tests will end up after **modeler** is complete.
We could have a **createDefaultComponentFixture** (creating a component from a default Model) and a more elaborate one **createComponentFromModelWithParametersFixture**

src/tests/src/study/system-model/test_component.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include "utils.h"
+
+using namespace Antares::Study::SystemModel;
+
+struct ComponentTestFixture
+{
+    static Model createModel(bool with_parameters);
+};
+
+Model ComponentTestFixture::createModel(bool with_parameters)
```
guilpier-code: Sorry, not a fan of bool arguments.
For example,` createModel(true);` is not very expressive. 
Instead, we could  define 2 fixtures : one defining a simple **Model** (**createDefaultModel**), another one using the first and adding parameters to this **Model** (**createModelWithParameters**).

Please see also[ this remark](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2500#discussion_r1853797243), they complete one another

guilpier-code (changes requested):

src/tests/src/study/system-model/test_component.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+    Model model = createModel(true);
+    ComponentBuilder component_builder;
+    auto component = component_builder.withId("component")
+                       .withModel(&model)
+                       .withParameterValues({{"param1", 5}, {"param2", 3}})
+                       .withScenarioGroupId("scenario_group")
+                       .build();
+    BOOST_CHECK_EQUAL(component.Id(), "component");
+    BOOST_CHECK_EQUAL(component.getModel(), &model);
```
guilpier-code, line 64: About :
```C++
BOOST_CHECK_EQUAL(component.getModel(), &model);
```
Rather than with its address, should we identify a **Model** with its **id** ?

src/tests/src/study/system-model/test_component.cpp
```diff
@@ -0,0 +1,213 @@
...
+    BOOST_CHECK_EQUAL(component.getScenarioGroupId(), "scenario_group");
+}
+
+BOOST_FIXTURE_TEST_CASE(nominal_build_without_parameters1, ComponentTestFixture)
+{
+    Model model = createModel(false);
+    ComponentBuilder component_builder;
+    auto component = component_builder.withId("component2")
+                       .withModel(&model)
+                       .withParameterValues({})
```
guilpier-code, line 78: About` .withParameterValues({})` : do we need that ? 
I mean, by default a component has an empty list of parameters, right ?
There are 2 occurrences of it in these tests.

guilpier-code (changes requested):

src/study/system-model/component.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <ranges>
+
+#include <antares/study/system-model/component.h>
+
+namespace Antares::Study::SystemModel
+{
+
+Component::Component(const ComponentData& data)
+{
+    // Check that mandatory attributes are not empty
```
guilpier-code, line 31: Why not move all these checks on incoming data in a private member function ?
```C++
isComponentDataValid(const ComponentData& data);
```
The current constructor's body would be more simple.

guilpier-code (changes requested):

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -0,0 +1,102 @@
...
+                       .withScenarioGroupId("scenario_group")
+                       .build();
+    return component;
+}
+
+BOOST_AUTO_TEST_SUITE(_System_)
+
+BOOST_FIXTURE_TEST_CASE(nominal_build, SystemTestFixture)
+{
+    SystemBuilder builder;
```
guilpier-code: Is there a reason why we don't move `SystemBuilder builder;` into the fixture ?

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "antares/study/system-model/system.h"
+
+#include "utils.h"
+
+using namespace Antares::Study::SystemModel;
+
+struct SystemTestFixture
+{
+    static Component createComponent(std::string id);
+};
```
guilpier-code: **createComponent** seems to be actually a simple free function, called in different way in unit tests. For that, it does not have to be declared in a fixture.
As you know, fixtures are about **set up** and **tear down** automation for each test.
We should make it a free function, having nothing to do with the fixture.

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+Component SystemTestFixture::createComponent(std::string id)
```
guilpier-code: Please see [this remark](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2500#discussion_r1853887762) first.

**SystemTestFixture** : what precision does **Test** give to this name ?
it could be called **SystemBuilderCreationFixture**, and contain only the definition of a **SystemBuilder**

guilpier-code (changes requested):

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * Antares_Simulator is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
```
guilpier-code, line 21: From what I can see, we unit test the **system builder** here, not the **system**.
So should not this file be renamed into **test_system_builder.cpp** ?

Ok, We can't build a component without using the component builder.

src/tests/src/study/system-model/utils.h
pet-mit: I'll rename it and move it up

src/tests/src/study/system-model/utils.h
pet-mit: I'll rename it and move it up

src/tests/src/study/system-model/test_component.cpp
pet-mit: I would rather keep the name of the test class so it is unique (to avoid compile issues if we use the same name in another cpp file of the same executable)

src/tests/src/study/system-model/test_component.cpp
pet-mit: if the component identifies its model by its id, then the component should have access to the model library. then it becomes much more difficult to use

src/tests/src/study/system-model/test_component.cpp
pet-mit: no we don't need it ; this is to test that the behavior of calling withParameterValues with an empty map is equivalent to when we don't call it at all

src/study/system-model/component.cpp
pet-mit: ok

src/tests/src/study/system-model/test_system.cpp
pet-mit: again, this is to avoid name clash of this function if we create another one in another test file with the same name. putting it into a unique fixture makes it unique

src/tests/src/study/system-model/test_system.cpp
pet-mit: "SystemTest" for me is the unique name of this file that tests System, hence making the fixture name unique

src/tests/src/study/system-model/test_component.cpp
pet-mit: ok

src/tests/src/study/system-model/test_component.cpp
pet-mit: ok

guilpier-code (changes requested):

src/tests/src/study/system-model/test_component.cpp
pet-mit: ok

src/tests/src/study/system-model/test_component.cpp
guilpier-code: Sorry I'm not fluent with this matter, but class **Model** has **Id()** method. Couldn't we use it here ?

src/tests/src/study/system-model/test_component.cpp
pet-mit: we can't add parameters to a model once it is created, so we would have to duplicate the code

src/tests/src/study/system-model/test_component.cpp
pet-mit: I would rather keep the tests as they are, because having the different steps separated (creting a model, then a component builder, than using the component builder) make them more readable, and easier to understand what is being tested

src/study/system-model/component.cpp
guilpier-code: or even better : in a free fonction, as a method would be useless here.

src/tests/src/study/system-model/test_component.cpp
pet-mit: we can but it wouldn't be enough, I actually want to test that the getModel() points to the right object, not only to a model with the same ID (two different models can have the same ID if they are stored in two separate libraries)

src/study/system-model/component.cpp
pet-mit: can you read my last commit ?

src/study/system-model/system.cpp
```diff
@@ -0,0 +1,87 @@
...
+    // Check that mandatory attributes are not empty
+    if (id.empty())
+    {
+        throw std::invalid_argument("A system can't have an empty id");
+    }
+    if (components.empty())
+    {
+        throw std::invalid_argument("A system must contain at least one component");
+    }
+    std::transform(components.begin(),
```
guilpier-code: We could even call **std::ranges::transform** instead of **std::transform** : 
```C++
std::ranges::transform(components,
                       std::inserter(components_, components_.end()),
                       [this](/*Non const to prevent copy*/ Component& component)
                       {
                           ...
                           return std::make_pair(component.Id(), std::move(component));
                       });
```

guilpier-code (changes requested):

src/study/system-model/system.cpp
guilpier-code: As the lambda function captures **this**, why not make it a member function and have a **System**'s private method : 
```C++
std::make_pair<std::string, Component>  System::makeComponent(Component& component)
{
    if (components_.contains(component.Id()))
    {
            throw std::invalid_argument(
                    "System has at least two components with the same id ('"
                    + component.Id() + "'), this is not supported");
    }
    return std::make_pair(component.Id(), std::move(component));
}
```
Doing this, we would have : 
```c++
System::System(const std::string_view id, std::vector<Component> components)
{
   ...
   std::ranges::transform(components,
                          std::inserter(components_, components_.end()),
                          System::makeComponent);
}
```
?

guilpier-code (changes requested):

src/tests/src/study/system-model/utils.h
guilpier-code: There are other places where we call **BOOST_CHECK_EXCEPTION** (see **testModelerLinearProblemWithOrtools.cpp** for instance), and where **checkMessage** should be called as well. We should create a ticket for this, or do it now.

src/tests/src/study/system-model/utils.h
guilpier-code: I saw you moved **checkMessage** to **src/tests/unit_test_utils.h**.
There is already a folder **src/tests/src/utils**, it may be the right place for it.

pet-mit: you're right, I didn't see it before

src/tests/src/study/system-model/utils.h
pet-mit: I prefer not to mix subjects, I created an issue (ANT-2473)

src/study/system-model/component.cpp
guilpier-code: **checkComponentDataValidity** should be a free function : 

```C++
static void checkComponentDataValidity(const ComponentData& data)
{
    ...
}
```

src/tests/src/study/system-model/test_system.cpp
guilpier-code: **createComponent** has nothing to do with  **set up** or **tear down**, so it does not belong to a fixture.
As you know, tests have 3 steps : **Arrange**, **Act** and **Assert**. 
Calls to your **createComponent** belongs to the **Act** step.

We should make it a free function.
If you're afraid of linkage mismatch, declare it as **static**, that's what the keyword is for : internal linkage.

Besides, link mismatch are allowed because we have several unit tests pages in an executable. Is it suitable ?
But this is another subject.

src/tests/src/study/system-model/test_system.cpp
guilpier-code: **SystemTestFixture** is encapsulating some code. So it is natural that the name of the fixture tells about what its code is about.

src/tests/src/study/system-model/test_system.cpp
pet-mit: ok, i'll make it static meanwhile

src/tests/src/study/system-model/utils.h
guilpier-code: good. Then I resolve this conversation.

src/study/system-model/system.cpp
pet-mit: I'm not sure I can make this work, because addComponent is not static?

src/tests/src/study/system-model/test_component.cpp
guilpier-code: Unit test have 3 steps : **Arrange**, **Act** , **Assert**.
If a piece of code belongs to **Act**, then it should **not** be in the fixture (which is necessarily about set up / tear down).
So **createModelWithParameters()** and **createModelWithoutParameters()** should be free functions.

src/study/system-model/system.cpp
guilpier-code: `System::makeComponent` is a method of a class **System**. 
This class is unique. 
There won't be any linkage mismatch.

pet-mit: done

guilpier-code (approved):

PR #2502, opened 2024-11-25
Allow the passing of optimization options through the API

flomnes:
- Add an `OptimizationOptions` argument to public function `PerformSimulation`
- Add an `OptimizationOptions` argument to private methods `API::run` and `API::execute`
- In the existing tests, use the default instance for `OptimizationOptions`, that is

```cpp
struct OptimizationOptions
{
    //! Force ortools use
    bool ortoolsUsed = false;
    //! The solver name, sirius is the default
    std::string ortoolsSolver = "sirius";
    bool solverLogs = false;
    std::string solverParameters;
};
```
- Add a test with COIN
- Move headers around (remove unused logs.h causing issues, etc.)

### DIFF
--- a/src/solver/utils/ortools_utils.cpp
+++ b/src/solver/utils/ortools_utils.cpp
@@ -21,6 +21,7 @@
 #include "antares/solver/utils/ortools_utils.h"
 
 #include <filesystem>
+#include <optional>
 
 #include <antares/exception/AssertionError.hpp>
 #include <antares/exception/LoadingError.hpp>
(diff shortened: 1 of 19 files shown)


src/tests/src/api_lib/test_api.cpp
pet-mit, line 1: can you add a test to test the new behaviour when we set the desired solver?

flomnes: See 9f5296548

src/solver/utils/ortools_utils.cpp
```diff
@@ -354,33 +355,42 @@ std::string availableOrToolsSolversString()
     return solvers.str();
 }
 
-MPSolver* MPSolverFactory(const bool isMip, const std::string& solverName)
+static std::optional<std::string> translateSolverName(const std::string& solverName, bool isMip)
```
flomnes, line 357: The previous code didn't handle `std::out_of_range` exceptions properly.

src/tests/src/api_internal/test_api.cpp
```diff
@@ -96,3 +98,35 @@ BOOST_AUTO_TEST_CASE(result_contains_problems)
...
+                                                                 .solverParameters = ""};
+
+    auto results = api.run(*study_loader.get(), opt);
+
+    BOOST_CHECK(!results.antares_problems.empty());
+    BOOST_CHECK(!results.error);
+    BOOST_CHECK_EQUAL(results.antares_problems.weeklyProblems.size(), 52);
+}
+
+// Test where we use an invalid OR-Tools
```
pet-mit:
```suggestion
// Test where we use an invalid OR-Tools solver
```

src/tests/src/api_internal/test_api.cpp
```diff
@@ -96,3 +98,35 @@ BOOST_AUTO_TEST_CASE(result_contains_problems)
...
+    Antares::API::APIInternal api;
+    auto study_loader = std::make_unique<InMemoryStudyLoader>();
+    const Antares::Solver::Optimization::OptimizationOptions opt{
+      .ortoolsUsed = true,
+      .ortoolsSolver = "this-solver-does-not-exist",
+      .solverLogs = true,
+      .solverParameters = ""};
+
+    auto shouldThrow = [&api, &study_loader, &opt] { return api.run(*study_loader.get(), opt); };
+    BOOST_CHECK_THROW(shouldThrow(), Antares::FatalError);
```
pet-mit: can you wait for #2500 to be merged and use the new unit_test_utils.h to check the message of the exception ? (see examples in test_component and test_system)

src/solver/utils/ortools_utils.cpp
```diff
@@ -354,33 +355,42 @@ std::string availableOrToolsSolversString()
...
+                                 + " not found. Please use one of the following "
+                                 + availableOrToolsSolversString();
+    const Antares::Data::AssertionError assertionError(notFound);
 
-            throw Antares::Data::AssertionError(msg_to_throw);
-        }
+    auto internalSolverName = translateSolverName(solverName, isMip);
+    if (!internalSolverName)
+    {
+        throw assertionError;
```
pet-mit: shouldn't we throw an std::invalid_argument exception here?

pet-mit (changes requested):

src/solver/utils/ortools_utils.cpp
a-zakir: what's the point to catch the exception at l371 to finally throw another here?

flomnes: That's right

flomnes: To `throw` a more meaningful exception. `std::out_of_range` has no meaning from the caller's point of view.

flomnes: So we now throw an `std::invalid_argument` exception !

a-zakir: yes but you could do it the catch

pet-mit (approved):

PR #2510, opened 2024-11-28
Fix/centos build (#2509)

a-zakir:


.github/workflows/centos7.yml
```diff
@@ -29,135 +30,77 @@ jobs:
...
+          mkdir archive
+          docker cp $container_id:/workspace/Antares_Simulator/_build/archive/ archive/ 
+          docker rm $container_id
+
+      - name: Install gh if needed
+        if: ${{ env.IS_RELEASE == 'true' }}
+        run: |
+          wget https://github.com/cli/cli/releases/download/v2.52.0/gh_2.52.0_linux_amd64.rpm
+          rpm -i gh_2.52.0_linux_amd64.rpm
+          gh --version
```
flomnes: On the ubuntu runners, gh is installed by default
```suggestion
```

flomnes (commented):
For ccache, it requires updating the online cache for persistency between runs.

.github/workflows/centos7.yml
```diff
@@ -2,10 +2,11 @@ name: Centos7 CI (push and/or release)
 
 on:
   release:
-    types: [created]
+    types: [ created ]
   push:
     branches:
       - develop
+      - fix/centos-build
```
payetvin: remove this

payetvin (changes requested):

.github/workflows/centos7.yml
a-zakir: done

payetvin (approved):

PR #2530, opened 2024-12-10
Modeler 4.6: system import [ANT-2207]

payetvin:
This PR add the ability to read a YAML file containing all the infos about the electric system and which models they're related to, converting it into system-model structures.


src/solver/systemParser/converter.cpp
```diff
@@ -0,0 +1,133 @@
...
+class ModelNotFound: public std::runtime_error
+{
+public:
+    explicit ModelNotFound(const std::string& s):
+        runtime_error("No model found with this name: " + s)
+    {
+    }
+};
+
+static std::pair<std::string, std::string> splitLibraryModelString(const std::string& s)
```
flomnes, line 61: Use a dedicated `struct` for the return type

src/solver/systemParser/converter.cpp
```diff
@@ -0,0 +1,133 @@
...
+    {
+        throw ErrorWhileSplittingLibraryAndModel(s);
+    }
+
+    std::string library = s.substr(0, pos);
+    std::string model = s.substr(pos + 1);
+    return {library, model};
+}
+
+static const SystemModel::Model* getModel(const std::vector<SystemModel::Library>& libraries,
```
flomnes:
```suggestion
static const SystemModel::Model& getModel(const std::vector<SystemModel::Library>& libraries,
```

src/solver/systemParser/converter.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    try
+    {
+        auto& model = lib->Models().at(modelId);
+        return &model;
+    }
+    catch (const std::out_of_range&)
+    {
+        throw ModelNotFound(modelId);
+    }
```
flomnes: Use `std::find` or `std::find_if` here instead of `try..catch`

src/solver/systemParser/converter.cpp
payetvin: I think it's fine to use a pair and get var like this in this context `const auto [libraryId, modelId] = getSplit...`

flomnes (approved):

PR #2539, opened 2024-12-18
Modeler 2.5a: Add parameters and YAML parser

flomnes:


src/solver/modeler/parameters/parseModelerParameters.cpp
```diff
@@ -0,0 +1,20 @@
+#include <filesystem>
+#include <fstream>
+
+#include "encoder.hxx"
+
+namespace Antares::Solver
+{
+static std::string loadFileToString(const std::filesystem::path& filePath)
```
payetvin: Use the function in `src/libs/antares/io/file.cpp`

payetvin (changes requested):

payetvin (approved):

PR #2540, opened 2024-12-19
Modeler 2.5a: Read models and system yaml files [ANT-2306]

payetvin:
Add the main modeler as a new binary
Directory loadFiles added, used to call previous work to read yaml files by handling I/O and errors related to yaml parsing


src/solver/modeler/loadFiles/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+        PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+)
+
+# Link dependencies (if any)
+target_link_libraries(loadModelerFiles
+        PUBLIC
+        Antares::antares-study-system-model
+        PRIVATE
+        io
```
flomnes:
```suggestion
        Antares::io
```

src/solver/modeler/loadFiles/readLibraries.cpp
```diff
@@ -0,0 +1,60 @@
...
+std::vector<Study::SystemModel::Library> loadLibraries(const fs::path& studyPath)
+{
+    std::vector<Study::SystemModel::Library> libraries;
+
+    const fs::path directoryPath = studyPath / "input" / "model-libraries";
+    for (const auto& entry: fs::directory_iterator(directoryPath))
+    {
+        if (entry.path().extension() != "yml")
+        {
+            logs.info() << "File ignored because of wrong extension: " << entry.path();
```
flomnes: OK but what is a valid extension ?
```suggestion
            logs.info() << entry.path() " ignored, only files having the `yml` extension are loaded";
```

flomnes (commented):
Good start ! Now we need a `main`

src/solver/modeler/loadFiles/CMakeLists.txt
payetvin: doesn't exists

payetvin: Added this in io/CmakeLists.txt `add_library(Antares::io ALIAS io)`

src/solver/modeler/loadFiles/readParameters.cpp
```diff
@@ -0,0 +1,61 @@
...
+ModelerParameters loadParameters(const fs::path& studyPath)
+{
+    try
+    {
+        const std::string paramStr = IO::readFile(studyPath / "parameters.yml");
+        return parseModelerParameters(paramStr);
+    }
+    catch (const YAML::Exception& e)
+    {
+        logs.error() << "Error while parsing the yaml parameters file";
```
flomnes:
```suggestion
        const auto fullPath = studyPath / "parameters.yml";
        const std::string paramStr = IO::readFile(fullPath);
        // ...
        logs.error() << "Error while parsing parameter file " << fullPath;
```

flomnes (approved):

PR #2546, opened 2024-12-20
For short-term storage cumulative constraints [ANT-1855]

a-zakir:
add week-dependent rhs
New syntax to declare multiple contraints that shares same rhs vector and type


src/libs/antares/study/include/antares/study/parts/short-term-storage/AdditionalConstraints.h
```diff
@@ -22,20 +22,30 @@
...
     std::string variable;
     std::string operatorType;
-    std::set<int> hours;
-    double rhs;
+    // TODO a lot unused entries
+    //std::array<double, HOURS_PER_YEAR> rhs = {};
+    std::vector<double> rhs = {};
 
-    unsigned int globalIndex = 0;
+    std::vector<SingleAdditionalConstraint> constraints = {};
```
guilpier-code:
```C++
std::vector<SingleAdditionalConstraint> constraints = {};
```
to be replaced with : 
```C++
std::vector<SingleAdditionalConstraint> constraints;
```

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -99,43 +100,77 @@ bool STStorageInput::LoadConstraintsFromIniFile(const fs::path& parent_path)
...
-                std::string hour;
-                while (std::getline(ss, hour, ','))
+                //
+                // std::stringstream ss(value.c_str());
+                // std::string hour;
+                // while (std::getline(ss, hour, ','))
+                // {
+                //     int hourVal = std::stoi(hour);
+                //     constraint.hours.insert(hourVal);
+                // }
```
guilpier-code: Let's not forget to remove this commented code if not needed

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -99,43 +100,77 @@ bool STStorageInput::LoadConstraintsFromIniFile(const fs::path& parent_path)
...
+
+                // Split the `hours` field into multiple groups
```
guilpier-code: in specification, we talk about **hour lists**, not **groups**. 
We should either stick to the plan and have **hourLists** in the code.

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -87,8 +88,8 @@ bool STStorageInput::LoadConstraintsFromIniFile(const fs::path& parent_path)
 
     for (auto* section = ini.firstSection; section; section = section->next)
     {
-        AdditionalConstraint constraint;
-        constraint.name = section->name.c_str();
+        AdditionalConstraints additional_constraints;
```
guilpier-code: _**Design**_
This fonction is now very long.
Indeed, the code it contains : 
1.  makes too many things : it reads multiple files :  **additional-constraints.ini** and all the **rhs_\<constraint-name\>.txt**
2.  mixes important things and details (we should hide details and highlight big steps).
     Examples : 
     - at this level, we don't need to know about how we retrieve list of hours (using regexp, isolate hour lists, ...), it could be done in a separate private method.
     - when reading RHS from .txt files there are many details on retrieved data validation that should be hidden here and rejected in a private method.

**Suggestions** : 
- **LoadConstraintsFromIniFile** should be renamed into **LoadAdditionalConstraints(** constraints_folder **)**
- This method could call a private method for each of the following steps : 
   - **ReadConstraintsFromIni(...)** : one loop on sections + call to private function **readHourList(...)** when crossing the key **"hours"**
   - **ReadAssociatedRHS(...)**
   - **validateConstraints(...)**

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -99,43 +100,77 @@ bool STStorageInput::LoadConstraintsFromIniFile(const fs::path& parent_path)
...
-            {
-                property->value.to<double>(constraint.rhs);
-            }
         }
 
-        if (auto ret = constraint.validate(); !ret.ok)
+        // try to read the rhs
+        loadFile(parent_path / ("rhs_" + additional_constraints.name + ".txt"),
+                 additional_constraints.rhs);
+        fillIfEmpty(additional_constraints.rhs, 0.0);
```
guilpier-code: _**Requires refactoring, so maybe for another pull request**_
About : 
```c++
loadFile(some_path, additional_constraints.rhs);
fillIfEmpty(additional_constraints.rhs, 0.0);
```
I'm not sure it's the right design.
Indeed : 
- **struct AdditionalConstraints** could have a constructor, that automatically sizes the vectors the **struct** contains, and fill it with default values, using `std::vector::assign` for instance. 
  As a consequence, we would not have to make a call to **fillIfEmpty(...)** here.
- **loadFile** should not have to reserve memory or makes push backs for the vector, it should only fill the vector.

Doing this, we would have one line instead of one, and code would be a bit simpler and expected.
It's done in the same way elsewhere (when loading **PMAX-injection.txt**, ...), but it does mean it's right.

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/container.cpp
guilpier-code: `AdditionalConstraints additional_constraints;` : this is a local variable, and due to the context we're in, we could shorten the name : 
`AdditionalConstraints constraints;`

src/libs/antares/study/parts/short-term-storage/AdditionalConstraints.cpp
```diff
@@ -18,11 +18,12 @@
...
 ** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
 */
-#include "antares/study/parts/short-term-storage/AdditionalConstraint.h"
+#include "antares/study/parts/short-term-storage/AdditionalConstraints.h"
+#include <algorithm>
 
 namespace Antares::Data::ShortTermStorage
 {
-AdditionalConstraint::ValidateResult AdditionalConstraint::validate() const
+AdditionalConstraints::ValidateResult AdditionalConstraints::validate() const
```
guilpier-code: We don't need to define a **struct ValidateResult** : it can be replaced with a `std::pair<bool, std::string>`.

That said, instead of returning something, we could throw exception (`std::runtime_error` ?) inside **validate()**.
This would simplify the function a bit : 
- it would return **void**
- it would avoid the strange default return `{true, ""}`

Calling it would be a bit simpler and conventional : 
```C++
try
{
    additional_constraints.validate();
}
catch (const std::runtime_error& ex)
{
    logs.error() << "Invalid constraint in section: " << section->name;
    logs.error() << ex.what();
    return false;
}
```

src/libs/antares/study/parts/short-term-storage/AdditionalConstraints.cpp
```diff
@@ -39,26 +40,38 @@ AdditionalConstraint::ValidateResult AdditionalConstraint::validate() const
...
-    return !hours.empty() && *hours.begin() >= 1 && *std::prev(hours.end()) <= 168;
+    return !hours.empty() && *hours.begin()
+           >= 1 && *std::prev(
+                   hours.end()) <= 168;
 }
 
-bool AdditionalConstraint::isValidVariable() const
+bool AdditionalConstraints::isValidHours() const
+{
+    return std::ranges::all_of(constraints.begin(),
```
guilpier-code: Not sure we need to specify `constraints.begin() `and `constraints.end()` in the `std::ranges::all_of` algorithm.
```C++
 return std::ranges::all_of(constraints,
                            [](const auto& constraint)
                            {
                                return constraint.isValidHoursRange();
                            });
```
is enough

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -61,10 +61,13 @@ static void importShortTermStorages(
...
             toInsert.name = st.properties.name;
             toInsert.additional_constraints = st.additional_constraints;
-            for (auto& constraint: toInsert.additional_constraints)
+            for (auto& additional_constraints: toInsert.additional_constraints)
             {
-                constraint.globalIndex = clusterCumulativeConstraintGlobalIndex;
-                ++clusterCumulativeConstraintGlobalIndex;
+                for (auto& [_, globalIndex,__]: additional_constraints.constraints)
+                {
+                    globalIndex = clusterCumulativeConstraintGlobalIndex;
```
guilpier-code: `clusterCumulativeConstraintGlobalIndex` : long name to say `constraintGlobalIndex`, shorter.

src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp
```diff
@@ -130,28 +130,33 @@ void ShortTermStorageCumulation::add(int pays)
 
     for (const auto& storage: data.ShortTermStorage[pays])
     {
-        for (const auto& constraint: storage.additional_constraints)
+        for (const auto& additional_constraints: storage.additional_constraints)
```
guilpier-code: Sorry, but I find this code painful to understand. 
I find it difficult to make the correspondance between the code and the content of the specifications, whereas it should be quite immediate.

I'll try to explain why :
- Function **cumulationConstraintFromVariable** : this function creates a constraint. Indeed, it returns a child of class **CumulationConstraint**. So I would rename it into **cumulationConstraintFactory(...)**
- **constraintHelper** : why do we give this variable that name ? This local variable is the result of a function that returns a  **CumulationConstraint**. So let's name this local variable **cumulationConstraint**. I personally never append or prefix a name (variable, class) with **helper** or **manager**, because it's not meaningful enough, it's too vague. 
- **class CumulationConstraint** : there is design problem
   This class is an interface and requires its children to define the **build(...)** function.
   **build(...)** takes 3 arguments. The 3rd argument is used only by one child, and useless to the other 2.
   To fix this, we could pass the **builder** and the **storage** to the **constraint factory** (function **cumulationConstraintFromVariable**), and pass the **storage** internally to the constructor of the cumulation constraint that needs it.
   Doing this, we simplify the call form : 
   `constraintHelper->build(builder, index, storage);`
   to : 
   `constraintHelper->build(index);`
   
   We could even go further and replace : 
   ```c++
   builder.updateHourWithinWeek(hour - 1);
   constraintHelper->build(builder, index, storage);
   ```
   with only :
    ```c++
   constraintHelper->build(index, hour); // Notice the new 'hour' as arg
   ```
- **namer** : useless to build a constraint itself, but used to build the constraint name to print in a MPS file. So we mix the construction of the constraint itself and its MPS name in the same function. We could do both things  separately (in different functions), so that each piece is simpler and more understandable, but it's already done in one piece in many places. So is it worth it to separate things here and not elsewhere ?
- **index** : when we use it, we don't know which objet this index is related to, so we have to navigate to its definition to realize it's a storage index. Let's name it **storageIndex**.
- About **ConvertSense(...)** : 
  - this conversion should have already been made at that (for example when reading input data or right after), it's not the place to do it. Doing it some place else would lighten the code a little bit more.
  - renaming : we talk about **operator type** on one hand, and about **sense** on the other hand. We have to choose which unique term to use. I ask to @JMJ-rte, he recommends **operator**.
    For more clarity, this function should be called **convertOperatorToChar**

guilpier-code (changes requested):

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
```diff
@@ -451,3 +455,478 @@ BOOST_FIXTURE_TEST_CASE(check_series_save, Fixture)
 }
 
 BOOST_AUTO_TEST_SUITE_END()
+
+BOOST_AUTO_TEST_SUITE(AdditionalConstraintsTests)
+
+BOOST_AUTO_TEST_CASE(Validate_ClusterIdEmpty)
```
guilpier-code, line 460: we should not hesitate to make test title look like a sentence :  
**Validate_ClusterIdEmpty** ==> **cluster_id_is_empty___we_get_validation_error**

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
```diff
@@ -451,3 +455,478 @@ BOOST_FIXTURE_TEST_CASE(check_series_save, Fixture)
...
+    constraints.variable = "injection";
+    constraints.operatorType = "invalid"; // Invalid operator type
+
+    auto [ok, error_msg] = constraints.validate();
+    BOOST_CHECK_EQUAL(ok, false);
+    BOOST_CHECK_EQUAL(error_msg,
+                      "Invalid operator type. Must be 'less', 'equal', or 'greater'.");
+}
+
+BOOST_AUTO_TEST_CASE(Validate_InvalidHours)
```
guilpier-code: In this test, we mean to test 3 configurations leading to a validation error.
Here we indeed have an error, but : 
- we don't which situation causes the error
- the 2 other configurations may not lead to an error

So we need 3 tests instead of one.

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
```diff
@@ -451,3 +455,478 @@ BOOST_FIXTURE_TEST_CASE(check_series_save, Fixture)
...
+{
+    ShortTermStorage::AdditionalConstraints constraints;
+    constraints.cluster_id = "ClusterA";
+    constraints.variable = "injection";
+    constraints.operatorType = "less";
```
guilpier-code, line 502: For many tests, we : 
- create an additional constraint
- give the same value to its fields

We should use a fixture creating an additional constraint and giving a default value to its fields

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
```diff
@@ -451,3 +455,478 @@ BOOST_FIXTURE_TEST_CASE(check_series_save, Fixture)
...
+
+    ShortTermStorage::SingleAdditionalConstraint constraint1;
+    constraint1.hours = {1, 2, 3}; // Valid hours
+
+    ShortTermStorage::SingleAdditionalConstraint constraint2;
+    constraint2.hours = {100, 150, 168}; // Valid hours
+
+    constraints.constraints.push_back(constraint1);
+    constraints.constraints.push_back(constraint2);
```
guilpier-code: `constraints.constraints = {constraint1, constraint2};`

guilpier-code (changes requested):

src/libs/antares/study/parts/short-term-storage/container.cpp
payetvin: additionalConstraints contains a member named constraints
constraints->constraints is not clear even in a local context

src/libs/antares/study/parts/short-term-storage/AdditionalConstraints.cpp
payetvin: pair done

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
```diff
@@ -22,13 +22,17 @@
 
 #define WIN32_LEAN_AND_MEAN
 
+#include <absl/strings/internal/str_format/extension.h>
```
flomnes: Is this required ? It seems to introduce a new dependency on absl

src/libs/antares/study/parts/short-term-storage/AdditionalConstraints.cpp
a-zakir: i don't thinh that std::pair<bool, std::string> is more meaningful than struct other { bool ok; string msg;} come on!
pair<bool, string> a; a.first = true;

other o; o.ok = true;

src/libs/antares/study/parts/short-term-storage/container.cpp
a-zakir: not related to that list; here group refers to the principle of regex group:

src/solver/simulation/sim_calcul_economique.cpp
a-zakir: constraintGlobalIndex; what constraint?
there is a lost of informations!

src/libs/antares/study/parts/short-term-storage/AdditionalConstraints.cpp
flomnes: I have to agree with @a-zakir on this one

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
a-zakir: i did update

src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp
a-zakir: ok for the first 3 points, about namer, ths object is instantiated once outside the for loops and one of its method is called inside those loops, i don't see what i can do about it unless to recreate it for every ite of the inner loop via a function.

on concertSense grammar, you'll notice that less, greater and equal are not striclty operator, they are words and their  global defintion is Sense

src/libs/antares/study/include/antares/study/parts/short-term-storage/additionalConstraints.h
```diff
@@ -22,32 +22,48 @@
...
 {
     std::string name;
     std::string cluster_id;
     std::string variable;
     std::string operatorType;
-    std::set<int> hours;
-    double rhs;
+    bool enabled = true;
+    // TODO a lot unused entries
+    // std::array<double, HOURS_PER_YEAR> rhs = {};
```
guilpier-code:
```C++
// std::array<double, HOURS_PER_YEAR> rhs = {};
```
comment to be removed

src/libs/antares/study/include/antares/study/parts/short-term-storage/additionalConstraints.h
```diff
@@ -22,32 +22,48 @@
...
+struct AdditionalConstraints
 {
     std::string name;
     std::string cluster_id;
     std::string variable;
     std::string operatorType;
-    std::set<int> hours;
-    double rhs;
+    bool enabled = true;
+    // TODO a lot unused entries
```
guilpier-code:
```C++
// TODO a lot unused entries
```
What do you mean ? 
If the comment is meant to remain there, it should be cristal clear or we will forget what it refers to

src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp
guilpier-code: Acceptable for me

guilpier-code (dismissed):

src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp
guilpier-code: Acceptable like this for me

src/libs/antares/study/parts/short-term-storage/AdditionalConstraints.cpp
guilpier-code: The suggestion of using a `std::pair<bool, std::string>` instead of a **struct** was a part of this remark.
What about the other part (with the code sample) ?

src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp
a-zakir: ok

src/libs/antares/study/include/antares/study/parts/short-term-storage/additionalConstraints.h
a-zakir: only rhs values corresponding to the hours set are used. (rhs is a vector 8760 lines)

src/libs/antares/study/parts/short-term-storage/AdditionalConstraints.cpp
a-zakir: we don't necessarily want to throw exception, the return boolean is used by the caller 
bool AreaList::**loadFromFolder** --> bool STStorageInput::**loadAdditionalConstraints** --> AdditionalConstraints::ValidateResult AdditionalConstraints::**validate**()

a-zakir: in your suggestion:
``` C++
try
{
    additional_constraints.validate();
}
catch (const std::runtime_error& ex)
{
    logs.error() << "Invalid constraint in section: " << section->name;
    logs.error() << ex.what();
    return false;
}

```

  i am doubtful with throwing an exception to catch it the line after.
  can you give an example of code in the project that use this technique?

guilpier-code: Well, the error handling in **Antares Simulator** in not necessarily at the state of art.
This suggestion has the advantage of : 
- simplifying the **validate()** function, especially its return type,
- make the caller as clear as it was before, with the **try... catch...** block (which indicates we're dealing with errors)

But in the end, this point is a detail

flomnes (approved):

flomnes (approved):

PR #2549, opened 2024-12-26
Branch 8.8: Use vcpkg

payetvin:


.github/workflows/install-cmake-328/action.yml
```diff
@@ -1,4 +1,4 @@
-name: "Install cmake 3.28 using devtoolset 10"
+name: "Install cmake 3.28 using devtoolset 10 if possible"
```
a-zakir, line 1: where this workflow is used?
cmake 3.28 can be installed with pip

src/CMakeLists.txt
```diff
@@ -246,7 +209,7 @@ endif()
 find_package(sirius_solver)
 
 if (NOT sirius_solver_FOUND)
-	message (FATAL_ERROR "Sirius solver not found. Sirius solver can be compiled with -DBUILD_sirius=ON or you can specify previous dependency install directory with -DCMAKE_PREFIX_PATH or -DDEPS_INSTALL_DIR")
+	message (FATAL_ERROR "Sirius solver not found. Sirius solver can be compiled with -DBUILD_sirius=ON or you can specify previous dependency install directory with -DCMAKE_PREFIX_PAT ")
```
a-zakir: is it true that BUILD_sirius=ON will build Sirius?

a-zakir: i don't think so

flomnes:
```suggestion
	message (FATAL_ERROR "Sirius solver not found. Sirius solver can be compiled with -DBUILD_sirius=ON or you can specify a dependency install directory with -DCMAKE_PREFIX_PATH")
```

flomnes (changes requested):

flomnes (approved):

PR #2550, opened 2024-12-30
Add attribute `enabled` to additional constraints, adapt tests [ANT-1855]

flomnes:


src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
```diff
@@ -61,19 +61,14 @@ static void shortTermStorageCumulationRHS(
...
-                                                            return sum + additional_constraints.rhs[
-                                                                    weekFirstHour + hour - 1];
-                                                        });
+                                      .ShortTermStorageCumulation[constraint.globalIndex];
+
+                    SecondMembre[cnt] = std::accumulate(
+                      constraint.hours.begin(),
+                      constraint.hours.end(),
+                      0.0,
+                      [weekFirstHour, &additional_constraints](const double sum, const int hour)
```
a-zakir: here you should check if the constraint is enabled

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -60,15 +60,19 @@ static void importShortTermStorages(
...
-            toInsert.additional_constraints = st.additional_constraints;
-            for (auto& additional_constraints: toInsert.additional_constraints)
+            for (auto& constr: st.additional_constraints)
             {
-                for (auto& [_, globalIndex,__]: additional_constraints.constraints)
+                if (constr.enabled)
                 {
-                    globalIndex = clusterCumulativeConstraintGlobalIndex;
-                    ++clusterCumulativeConstraintGlobalIndex;
+                    for (auto c: constr.constraints)
```
a-zakir: no! catch the constraint by reference

src/solver/optimisation/opt_gestion_second_membre_cas_lineaire.cpp
flomnes: There is no need for this, because the addition of constraints to `shortTermStorageInput` is already done in a selective manner (see sim_calcul_economique.cpp)

src/solver/simulation/sim_calcul_economique.cpp
flomnes: you are totally right, without the reference we modify a copy which is not what we want.

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -102,6 +102,10 @@ bool STStorageInput::LoadConstraintsFromIniFile(const fs::path& parent_path)
                 value.to<std::string>(clusterName);
                 additional_constraints.cluster_id = transformNameIntoID(clusterName);
             }
+            else if (key == "enabled")
+            {
+                value.to<bool>(additional_constraints.enabled);
+            }
```
a-zakir, line 148: if it's disabled, why bother reading and storing it?

a-zakir: especially rhs and all related code to find the sts object

flomnes: Possible reasons
- GUI/save => no longer the case
- Indexes => we want to keep the same indices when disabling an object for reproducibility reasons

But we would need to look for the "enabled" key first, and decide later. Another dev would be required

flomnes: I suggest that we skip rhs & finding the STS if the constraint is disabled

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -155,13 +160,19 @@ bool STStorageInput::loadAdditionalConstraints(const fs::path& parentPath)
...
         }
 
-        if (const auto rhsPath = parentPath / (
-                                     "rhs_" + additionalConstraints.name
-                                     + ".txt"); !readRHS(additionalConstraints, rhsPath))
+        // We don't want load RHS and link the STS time if the constraint is disabled
+        if (!additionalConstraints.enabled)
+        {
+            return true;
+        }
```
flomnes, line 167: If the constraint is disabled, we don't load the timeseries and link the STS object.

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
```diff
@@ -943,4 +945,45 @@ BOOST_DATA_TEST_CASE(Validate_AllVariableOperatorCombinationsFromFile,
...
+    iniFile << "hours=[1,2,3]\n";
+    iniFile.close();
+
+    // Write the `rhs_constraint1.txt` file
+    std::ofstream rhsFile(testPath / "rhs_constraint1.txt");
+    for (int i = 0; i < HOURS_PER_YEAR; ++i)
+    {
+        rhsFile << i * 1.0 << "\n";
+    }
+    rhsFile.close();
```
a-zakir:
```suggestion
```

a-zakir (approved):

PR #2560, opened 2025-01-07
Better handling of variable bounds for BOOL variables

flomnes:
Make it so that the user doesn't need to provide lower & upper bounds for Boolean variables.

### DIFF
--- a/src/solver/optim-model-filler/ComponentFiller.cpp
+++ b/src/solver/optim-model-filler/ComponentFiller.cpp
@@ -43,10 +43,33 @@ void ComponentFiller::addVariables(Solver::Modeler::Api::ILinearProblem& pb,
     auto evaluator = std::make_unique<Solver::Visitors::EvalVisitor>(evaluationContext_);
     for (const auto& variable: component_.getModel()->Variables() | std::views::values)
     {
-        pb.addVariable(evaluator->dispatch(variable.LowerBound().RootNode()),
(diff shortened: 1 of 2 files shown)


src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -40,13 +40,36 @@ void ComponentFiller::addVariables(Solver::Modeler::Api::ILinearProblem& pb,
...
-        pb.addVariable(evaluator->dispatch(variable.LowerBound().RootNode()),
-                       evaluator->dispatch(variable.UpperBound().RootNode()),
-                       variable.Type() != Study::SystemModel::ValueType::FLOAT,
-                       component_.Id() + "." + variable.Id());
+        const auto variableID = component_.Id() + "." + variable.Id();
+        switch (variable.Type())
+        {
+        case Study::SystemModel::ValueType::BOOL:
+        {
+            pb.addVariable(0., 1., true, variableID);
```
pet-mit: the user might define lb / ub (for example ub can be a model parameter). it's only in the case where the user doesn't define lb/ub that we should us these default values. maybe set them in the Variable object model to keep this code simple?

pet-mit (changes requested):

pet-mit (changes requested):
ok for me, just add unit tests

flomnes:
This topic is too complex to be handled without a proper spec

PR #2562, opened 2025-01-07
Print solution values in the logs if a solution exists

flomnes:


src/solver/modeler/main.cpp
```diff
@@ -75,6 +75,26 @@ int main(int argc, const char** argv)
...
+        auto* solution = pb.solve(parameters.solverLogs);
+        switch (solution->getStatus())
+        {
+        case Antares::Solver::Modeler::Api::MipStatus::OPTIMAL:
+        case Antares::Solver::Modeler::Api::MipStatus::FEASIBLE:
+            if (!parameters.noOutput)
+            {
+                logs.info() << "Variables";
+                // TODO don't rely on this specific function
+                for (auto* var: pb.MpSolver()->variables())
```
payetvin: We could add a function to MipSolution to get all solutions and avoid using MpSolver

src/solver/modeler/ortoolsImpl/include/antares/solver/modeler/ortoolsImpl/linearProblem.h
```diff
@@ -67,8 +67,6 @@ class OrtoolsLinearProblem: public Api::ILinearProblem
     OrtoolsMipSolution* solve(bool verboseSolver) override;
 
     double infinity() const override;
-
-protected:
```
payetvin: Should stay protected to keep implementation details away in `src/solver/modeler/main.cpp`

payetvin (changes requested):

src/solver/modeler/ortoolsImpl/include/antares/solver/modeler/ortoolsImpl/linearProblem.h
flomnes: Yes, the API currently lacks the ability to loop over variables ! All we have is
```
IMipVariable* getVariable(const std::string& name)
```
which is kind of useless when we want to loop over all variables (unless we already know their name, even then it's not very convenient).

payetvin (approved):

PR #2565, opened 2025-01-08
Write raw optimization results [ANT-2302]

flomnes:
Allow the user to export solutions through new boolean option `include-export-solutions` (default=`false`).

### DIFF
--- a/docs/user-guide/solver/static-modeler/04-parameters.md
+++ b/docs/user-guide/solver/static-modeler/04-parameters.md
@@ -434,6 +434,17 @@ _**This section is under construction**_
 > _**Note:**_ You can find more information on this parameter [here](../03-appendix.md#details-on-the-include-exportmps-parameter).
 
 ---
+#### include-export-solutions
+- **Expected value:** `true` or `false`
+- **Required:** no
+- **Default value:** `false`
+- **Usage:** set to `true` to activate writing the raw optimization results, that is
+  - values and reduced costs for each variable
+  - marginal cost for each constraint
+
+This is an advanced option intended to help developers and advanced users better understand their simulation results.
+
+---
 #### include-split-exported-mps
 [//]: # (TODO: document this parameter, seems to belong to another category)
 _**This section is under construction**_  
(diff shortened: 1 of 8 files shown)


docs/user-guide/solver/static-modeler/04-parameters.md
```diff
@@ -433,6 +433,17 @@ _**This section is under construction**_
 
 > _**Note:**_ You can find more information on this parameter [here](../03-appendix.md#details-on-the-include-exportmps-parameter).
 
+---
+#### include-export-solutions
+- **Expected value:** `true` or `false`
+- **Required:** no
+- **Default value:** `false`
+- **Usage:** set to `true` to activate writing the raw optimization results, that is
```
pet-mit, line 441:
```suggestion
- **Usage:** set to `true` to activate writing the raw optimization results under ... (put path & filename here), that is
```

pet-mit (approved):

src/solver/optimisation/opt_optimisation_lineaire.cpp
```diff
@@ -62,6 +62,38 @@ void OPT_EcrireResultatFonctionObjectiveAuFormatTXT(
     writer.addEntryFromBuffer(filename, buffer);
 }
 
+void OPT_WriteSolution(const PROBLEME_ANTARES_A_RESOUDRE& pb,
+                       const OptPeriodStringGenerator& optPeriodStringGenerator,
+                       int optimizationNumber,
+                       Solver::IResultWriter& writer)
+{
+    Yuni::Clob buffer;
```
guilpier-code, line 70: `Yuni::Clob buffer;` : 
Is it possible to get rid of **Clob** and use a `std::string` instead (faster and more standard) ?
The `IResultWriter` has a function : 
```C++
void addEntryFromBuffer(const std::filesystem::path& entryPath,
                        std::string& entryContent)
```
Can we not use it ?

guilpier-code (changes requested):

PR #2566, opened 2025-01-08
Accurate shave peaks algorithm integration

guilpier-code:


src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -220,6 +221,92 @@ static bool Remix(const Data::AreaList& areas,
...
+    }
+    return to_return;
+}
+
+std::vector<double> extractLoadForCurrentWeek(const Data::Area& area,
+                                              const unsigned int year,
+                                              const unsigned int firstHourOfWeek)
+{
+    std::vector<double> load_to_return(168, 0.);
+    for (int h = 0; h < 168; h++)
```
flomnes: Don't use 168, use a constant instead

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -220,6 +221,92 @@ static bool Remix(const Data::AreaList& areas,
...
+    return load_to_return;
+}
+
+std::vector<double> extractHydroPmin(const Data::Area& area,
+                                     const unsigned int year,
+                                     const unsigned int firstHourOfWeek)
+{
+    // area->hydro.series->mingen.timeSeries
+    std::vector<double> hydroPmin(168, 0.);
+    for (int h = 0; h < 168; h++)
```
flomnes: Use a constant instead

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -220,6 +221,92 @@ static bool Remix(const Data::AreaList& areas,
...
+                                     .NiveauInitialReservoir;
+          const double capacity = area.hydro.reservoirCapacity;
+          const auto& inflows = problem.CaracteristiquesHydrauliques[area.index]
+                                  .ApportNaturelHoraire;
+          const auto& ovf = weeklyResults.debordementsHoraires;
+          const auto& pump = weeklyResults.PompageHoraire;
+          const auto& spillage = weeklyResults.ValeursHorairesDeDefaillanceNegative;
+
+          const auto& dtgMrgArray = area.scratchpad[numSpace].dispatchableGenerationMargin;
+          const std::vector<double> dtgMrg(dtgMrgArray, dtgMrgArray + 168);
```
flomnes: Use a constant

src/solver/simulation/common-hydro-remix.cpp
guilpier-code: ok, **168** was also used in previous hydro remix algo, as well as in many places in **Simulator**.
I'll locally change this, but we should think about doing something about it

src/solver/simulation/common-hydro-remix.cpp
guilpier-code: Done

src/solver/simulation/common-hydro-remix.cpp
guilpier-code: Done

flomnes (approved):

PR #2589, opened 2025-01-23
Remove new/delete for AreaUI

flomnes:


a-zakir:
I guess the last commit is ok

a-zakir (approved):

PR #2608, opened 2025-01-30
New output variable : partial min gen for thermal clusters [ANT-2425]

guilpier-code:
- Related ticket is [ANT-2425](https://gopro-tickets.rte-france.com/browse/ANT-2425).
- A study that tests this new behavior can be found [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR/pull/43) (please have a look to its users notes).


src/libs/antares/study/include/antares/study/parts/thermal/cluster.h
```diff
@@ -293,6 +293,16 @@ class ThermalCluster final: public Cluster, public std::enable_shared_from_this<
...
+    bool hasMinGen() const
+    {
+        return anyMinGen_;
+    }
+
+    void anyMinGen(bool minGen)
+    {
+        anyMinGen_ = minGen;
+    }
+
```
flomnes: Use above `PthetaInf`, no need for a setter here

```cpp
class ThermalCluster
{
public:
    bool hasMinGen() const
    {
        if (!hasMinGen_.has_value())
        {
            hasMinGen_ = std::ranges::any_of(/*...*/);
        }
        return hasMinGen_.value();
    }

private:
    mutable std::optional<bool> hasMinGen_;
};

```

src/libs/antares/study/include/antares/study/parts/thermal/cluster.h
```diff
@@ -307,6 +317,7 @@ class ThermalCluster final: public Cluster, public std::enable_shared_from_this<
     CostProvider& getCostProvider();
 
 private:
+    bool anyMinGen_ = false;
```
flomnes: Not necessary

src/libs/antares/study/study.cpp
```diff
@@ -1422,13 +1422,15 @@ void Study::computePThetaInfForThermalClusters() const
...
         {
             for (uint k = 0; k < HOURS_PER_YEAR; k++)
             {
-                cluster->PthetaInf[k] = cluster->modulation[Data::thermalMinGenModulation][k]
-                                        * cluster->unitCount * cluster->nominalCapacity;
+                c->PthetaInf[k] = c->modulation[Data::thermalMinGenModulation][k] * c->unitCount
+                                  * c->nominalCapacity;
             }
+
+            c->anyMinGen(std::ranges::any_of(c->PthetaInf, [](double& e) { return e > 0.5; }));
```
flomnes: This threshold is completely arbitrary

src/libs/antares/study/include/antares/study/parts/thermal/cluster_list.h
```diff
@@ -94,6 +94,11 @@ class ThermalClusterList: public ClusterList<ThermalCluster>
                | std::views::filter(std::not_fn(&ThermalCluster::isMustRun));
     }
 
+    auto each_has_min_gen() const
+    {
+        return each_enabled_and_not_mustrun() | std::views::filter(&ThermalCluster::hasMinGen);
+    }
```
flomnes: Probably not necessary

src/solver/variable/include/antares/solver/variable/economy/minOfDispatchableGen.h
```diff
@@ -0,0 +1,319 @@
...
+        for (auto& cluster: area->thermal.list.each_has_min_gen())
+        {
+            unsigned int groupNumber = groupToNumbers_[cluster->getGroup()];
+            double minGen = cluster->PthetaInf[state.hourInTheYear];
+            double production = thermal[area->index]
+                                  .thermalClustersProductions[cluster->enabledIndex];
+
+            pValuesForTheCurrentYear[numSpace][groupNumber][state.hourInTheYear] += std::min(
+              production,
+              minGen);
```
flomnes: Not sure I understand this, isn't this always the case that 

production >= minGen

?

guilpier-code: - **minGen** is made from only : 
   - **min gen modulation** (which is a TS of doubles > 0.) 
   -  **nominal capacity**.
- **available capacity** is a TS on its own.

If, for some hours in the week, **available capacity** (which is an upper bound for production) decreases under **minGen**,  then the lower bound for production at these hours is **available capacity**. So, for these hours, production is exactly **available capacity**.

Indeed, we have (see **common-eco-adq.cpp**) : 
```c++
Pt.PuissanceMinDuPalierThermique[hourInWeek] = 
    (Pt.PuissanceDisponibleDuPalierThermique[hourInWeek] < cluster->PthetaInf[hourInYear]) ?
        Pt.PuissanceDisponibleDuPalierThermique[hourInWeek] : 
        cluster->PthetaInf[hourInYear];
```

src/libs/antares/study/study.cpp
guilpier-code: Done : replaced with `return !Utils::isZero(e);`

src/libs/antares/study/include/antares/study/parts/thermal/cluster.h
guilpier-code: Not used to see a getter with an **if** block, it adds as bit of complexity.
Normally, `hasMinGen_` should be set at the cluster construction, so we would not need the **if** block here.
I did the changes you ask though.

src/libs/antares/study/include/antares/study/parts/thermal/cluster.h
guilpier-code: Replaced with `std::optional<bool> hasMinGen_;`, as requested

src/libs/antares/study/include/antares/study/parts/thermal/cluster_list.h
guilpier-code: I removed `hasMinGenCount()` from class **ThermalClusterList**.
But for this one (`each_has_min_gen()`), what do you suggest ?

flomnes: I suggest not using this filter since it's extremely unusual for output variables to be printed based on such a condition. Print the value for every enabled cluster.

guilpier-code: cluster list is a business concept, and output variables is not.
So you mean that `each_has_min_gen()` unnecessarily bloats class **ThermalClusterList** with a function used only in output ?

If you want `each_has_min_gen()` to disappear from class **ThermalClusterList**, we could create instead a free function more in the scope of output variables (**utils.h** ?) that would do the job and would avoid code duplication.

flomnes: I think we should produce this variable for all enabled cluster, no matter if they have a min-gen or not

guilpier-code: > I think we should produce this variable for all enabled cluster, no matter if they have a min-gen or not

it's a change of specification. To be discussed with our PO.

flomnes (commented):
As discussed, you can remove filters on `hasMinGen`, etc. and write the output variable no matter what

src/tests/run-study-tests/check_on_results/output_compare.py
```diff
@@ -85,9 +85,13 @@ def compare_simulation_files(simulation_files, tol):
     at_least_one_diff = False
     col_name_where_diff = ""
     for file_pair in simulation_files:
+        rows_to_skip = [0,1,2,3,5,6]
```
guilpier-code, line 88: In this PR, we had regressions on **details** output files : indeed, adding a new output variables for every cluster results in adding a new column in details files. This new column's title is the name of the cluster, again (as for every variable associated to a cluster).
So now, for a details file, we give the columns the subtitles (Units) as their names.

src/solver/variable/include/antares/solver/variable/economy/minDispatchableGenByPlant.h
```diff
@@ -0,0 +1,281 @@
...
+#include "antares/solver/variable/variable.h"
+
+namespace Antares::Solver::Variable::Economy
+{
+struct VCardMinDispatchableGenByPlant
+{
+    //! Caption
+    static std::string Caption()
+    {
+        return "Min DTG by plant";
```
flomnes:
```suggestion
        return "MIN DTG by plant";
```
looks more natural

src/solver/variable/include/antares/solver/variable/economy/minDispatchableGenByPlant.h
```diff
@@ -0,0 +1,281 @@
...
+    void initializeFromArea(Data::Study* study, Data::Area* area)
+    {
+        // Get the number of years in parallel
+        pNbYearsParallel = study->maxNbYearsInParallel;
+        pValuesForTheCurrentYear.resize(pNbYearsParallel);
+
+        // Get the area
+        nbClusters_ = std::ranges::count_if(area->thermal.list.all(),
+                                            [](auto c)
+                                            { return c->isEnabled() && !c->isMustRun(); });
```
flomnes: Please be consistent with other "by plant" variables (e.g Profit by plant, DTG by plant, etc.)

```cpp
        pSize = area->thermal.list.enabledCount();
```

src/solver/variable/include/antares/solver/variable/economy/minDispatchableGenByPlant.h
```diff
@@ -0,0 +1,281 @@
...
+        // Next variable
+        NextType::computeSummary(numSpaceToYear, nbYearsForCurrentSummary);
+    }
+
+    void hourForEachArea(State& state, unsigned int numSpace)
+    {
+        auto& area = state.area;
+        auto& thermal = state.thermal;
+        unsigned indexCluster = 0;
+        for (auto& cluster: area->thermal.list.each_enabled_and_not_mustrun())
```
flomnes: As discussed, please be consistent with other "by plant" variables

=> use `thermal.list.each_enabled()`

src/solver/variable/include/antares/solver/variable/economy/minDispatchableGenByPlant.h
```diff
@@ -0,0 +1,281 @@
...
+        {
+            double minGen = cluster->PthetaInf[state.hourInTheYear];
+            double production = thermal[area->index]
+                                  .thermalClustersProductions[cluster->enabledIndex];
+
+            pValuesForTheCurrentYear[numSpace][indexCluster++].hour[state.hourInTheYear] += std::
+              min(production, minGen);
```
flomnes:
```suggestion
            pValuesForTheCurrentYear[numSpace][cluster->enabledIndex].hour[state.hourInTheYear] += std::
              min(production, minGen);
```

src/solver/variable/include/antares/solver/variable/economy/minDispatchableGenByPlant.h
guilpier-code: Done

src/solver/variable/include/antares/solver/variable/economy/minDispatchableGenByPlant.h
guilpier-code: Done

src/solver/variable/include/antares/solver/variable/economy/minDispatchableGenByPlant.h
guilpier-code: Done

flomnes (approved):

PR #2612, opened 2025-01-30
Define exception classes as subclasses

flomnes:
In order to avoid prefixes, we use sub-classes instead of separate classes

For each class, we define a few sub-classes so that we don't need prefixes.


guilpier-code (approved):

guilpier-code:
Indeed, better like this

PR #2614, opened 2025-01-30
Derive ElemAlreadyExists from std::runtime_error

flomnes:
`std::exception`'s constructors don't accept `std::strings`, so it's a bit strange to override method `what`

https://en.cppreference.com/w/cpp/error/exception/exception


a-zakir:
I am always against this practice!

a-zakir (approved):

PR #2616, opened 2025-02-03
Refactor modeler code

pet-mit:
- move "expressions" code to src/expressions
- move optimisation sources to src/optimisatio,
- move yml importers to src/io


a-zakir (approved):

a-zakir:
this line must be updated in order to avoid sonar issues:
https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/4c28b49eac075ed34277f65aa1a539938f435148/sonar-project.properties#L37

src/expressions/include/antares/expressions/iterators/pre-order.h
```diff
@@ -3,7 +3,7 @@
 #include <stack>
 #include <vector>
 
-namespace Antares::Solver::Nodes
```
guilpier-code, line 6: Why is this file named **pre-order.h** whereas it contains declaration of class **AST** ?
**AST** seems to be the main topic here, at least the feature we should highlight.
Class **ASTPreOrderIterator** is an implementation detail.
Rename into **ast.h** ?

guilpier-code (changes requested):

src/expressions/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+)
+target_link_libraries(expressions
+        PUBLIC
+        Boost::headers
+        Antares::logs
+)
+
+
+add_library(expressions-iterators
```
guilpier-code, line 90: This remark follows [this one](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2616#discussion_r1942871952).
About **expressions-iterators** target : as it mainly defines class AST, and AST is central to expressions, these files should be part of **expressions** target.

src/expressions/include/antares/expressions/visitors/InvalidNode.h
```diff
@@ -22,11 +22,11 @@
 #pragma once
 #include <stdexcept>
 
-namespace Antares::Solver::Visitors
+namespace Antares::Expressions::Visitors
 {
 class InvalidNode: public std::invalid_argument
 {
 public:
     explicit InvalidNode(const std::string& node_name = "");
```
guilpier-code: Dangerous to have a default argument here (`node_name = ""`).
This means we allow : 
```C++
throw InvalidNode();
```
without specifying any node name.
In this case, if exception is caught, associated error message will be cryptic, as we don't know which node is concerned.

src/expressions/visitors/InvalidNode.cpp
```diff
@@ -0,0 +1,10 @@
+#include <antares/expressions/visitors/InvalidNode.h>
+
+namespace Antares::Expressions::Visitors
+{
+
+InvalidNode::InvalidNode(const std::string& node_name):
+    std::invalid_argument("Antares::Expressions::Nodes Visitor: invalid node type " + node_name)
```
guilpier-code:
```c++
"Antares::Expressions::Nodes Visitor: ..."
```
Does user cares about hard coded namespace things when she reads the error message of an invalid node ?
```c++
"Invalid node type " + node_name
```
is enough

guilpier-code (dismissed):

guilpier-code:
Attempted small changes on the current PR : see #2619

src/expressions/include/antares/expressions/iterators/pre-order.h
pet-mit: @a-zakir opinion ?

src/expressions/visitors/InvalidNode.cpp
pet-mit: I made it more "human": 
~~~
"Node visitor encountered an invalid node type: " + node_name
~~~

PR #2617, opened 2025-02-03
Add modeler non-regression tests

pet-mit:


src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+    def __read_csv(self, file_name) -> pd.DataFrame:
+        absolute_path = os.path.join(self.study_output_path, file_name.replace("/", os.sep))
+        return pd.read_csv(absolute_path, header=None, sep=' ', low_memory=False)
+
+    def get_optimal_value(self, var : str) -> float:
+        for row in self.results.iterrows():
+            if row[1][0] == var:
+                return row[1][1]
```
flomnes, line 23: Maybe use a dict here ? for example `{"var1": 42.0, "var2":41.0}`

src/solver/modeler/main.cpp
```diff
@@ -90,7 +90,7 @@ int main(int argc, const char** argv)
     std::filesystem::path studyPath(argv[1]);
     logs.info() << "Study path: " << studyPath;
 
-    if (!std::filesystem::is_directory(studyPath))
+    if (!is_directory(studyPath))
```
flomnes: While this works thanks to ADL, I think it's better to keep the namespaces for clarity & safety.
If later we added another `NS::is_directory(const std::filesystem::path&)` function, the compiler might resolve to a different function and this kind of error is very hard to detect.

src/solver/modeler/main.cpp
```diff
@@ -105,6 +105,22 @@ int main(int argc, const char** argv)
...
+        if (!parameters.noOutput)
+        {
+            logs.info() << "Output folder : " << outputPath;
+            if (!is_directory(outputPath))
+            {
+                if (!create_directory(outputPath))
+                {
+                    logs.error() << "Failed to create output directory. Exiting simulation.";
+                    throw std::runtime_error(
+                      "Failed to create output directory. Exiting simulation.");
```
flomnes: Currently, this exception isn't caught. I suggest instead
```suggestion
                    return EXIT_FAILURE;
```

src/solver/modeler/main.cpp
```diff
@@ -118,20 +134,20 @@ int main(int argc, const char** argv)
         if (!parameters.noOutput)
         {
             logs.info() << "Writing problem.lp...";
-            auto mps_path = std::filesystem::current_path() / "problem.lp";
+            auto mps_path = outputPath / "problem.lp";
```
flomnes: Naming
```suggestion
            auto lp_path = outputPath / "problem.lp";
```

flomnes (commented):
LGTM. Maybe find better names for studies ? "Two generators one load - single node", etc. I find that "study_2.4.2" is not very explicit

src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
pet-mit: too hard to do with pandas, or I don't know it well. I suggest we improve this later

flomnes (approved):

PR #2620, opened 2025-02-06
Make loading of renewable clusters less strict, add logs

flomnes:
Renewable clusters aren't strictly necessary in a study

### DIFF
--- a/src/libs/antares/study/area/list.cpp
+++ b/src/libs/antares/study/area/list.cpp
@@ -1215,15 +1215,23 @@ bool AreaList::loadFromFolder(const StudyLoadOptions& options)
     {
         // The cluster list must be loaded before the method ensureDataIsInitialized is called
         // in order to allocate data with all renewable clusters.
-        fs::path renewClusterPath = pStudy.folderInput / "renewables" / "clusters";
 
-        auto end = areas.end();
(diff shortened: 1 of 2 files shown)


src/libs/antares/study/area/list.cpp
```diff
@@ -1215,15 +1215,23 @@ bool AreaList::loadFromFolder(const StudyLoadOptions& options)
     {
         // The cluster list must be loaded before the method ensureDataIsInitialized is called
         // in order to allocate data with all renewable clusters.
-        fs::path renewClusterPath = pStudy.folderInput / "renewables" / "clusters";
 
-        auto end = areas.end();
-        for (auto i = areas.begin(); i != end; ++i)
+        fs::path renewClusterPath = pStudy.folderInput / "renewables" / "clusters";
+        if (!fs::is_directory(renewClusterPath))
```
guilpier-code: or more readable : 
```c++
if (fs::is_directory(renewClusterPath))
{
    // Read
}
else
{
    // Logs
}
```

guilpier-code (changes requested):

src/libs/antares/study/area/list.cpp
flomnes: done

PR #2621, opened 2025-02-06
[ANT-2033] Import modeler data-series

pet-mit:
- import data series from input/data-series, stored in tsv or csv files, seperated with tabs
- store in LinearProblemData for later use


src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <fstream>
+
+#include <boost/regex.hpp>
+
+#include <antares/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.h>
+
+namespace Antares::IO::Inputs::DataSeriesCsvImporter
+{
+using namespace std;
+using namespace boost;
```
flomnes: Please use fully qualified names for boost and std

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,111 @@
...
+        vector<double> row;
+        getline(infile, line);
+        std::stringstream ss(line);
+        while (getline(ss, element, csvSeparator))
+        {
+            row.push_back(stod(element));
+        }
+        result.push_back(row);
+    }
+    return move(result);
```
flomnes: Returning by move shouldn't be necessary in this case

flomnes (commented):
You shouldn't assume that the CSV file is well-formed
It could have garbage characters
Two rows could have a different number of columns
etc.

guilpier-code:
Please relate your PR to a ticket and specification in the description

src/solver/modeler/main.cpp
```diff
@@ -119,7 +135,7 @@ int main(int argc, const char** argv)
         logs.info() << "linear problem of System loaded";
         OrtoolsLinearProblem ortools_linear_problem(true, parameters.solver);
 
-        system_linear_problem.Provide(ortools_linear_problem, parameters);
+        system_linear_problem.Provide(ortools_linear_problem, parameters, dataSeriesRepository);
```
guilpier-code, line 138: _Not strictly related to this PR_

Instead of : 
```c++
OrtoolsLinearProblem ortools_linear_problem(true, parameters.solver);
system_linear_problem.Provide(ortools_linear_problem, parameters, 
                              dataSeriesRepository);
```
I would recommend : 
```c++
OrtoolsLinearProblem ortools_linear_problem = 
      system_linear_problem_builder.build(parameters, dataSeriesRepository);
```
Shorter and clearer. 
Indeed, the output of current **Provide** function is the linear problem, so why not make it clear.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+        throw std::invalid_argument("Not a directory: " + path.string());
+    }
+    DataSeriesRepository repo{};
+    for (const auto& entry: std::filesystem::directory_iterator(path))
+    {
+        if (!is_regular_file(entry))
+        {
+            continue;
+        }
+        if (entry.path().extension() == ".csv" || entry.path().extension() == ".tsv")
```
flomnes: What's the expected behavior if both `data.tsv` and `data.csv` are present ?

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+TimeSeriesSet TimeSeriesSetImporter::importFromFile(const std::filesystem::path& path,
+                                                    char csvSeparator)
+{
+    auto csvMatrix = csvToMatrix(path, csvSeparator);
+    // We have to transpose the matrix
+    // TODO: we may want to improve this by reading directly into the TimeSeriesSet object, or
+    // by creating a specific IDataSeries implementation
+    int nTimestamps = csvMatrix.size();
+    TimeSeriesSet timeSeriesSet(path.stem(), nTimestamps);
```
guilpier-code: Line : 
```c++
TimeSeriesSet timeSeriesSet(path.stem(), nTimestamps);
```
does not compile on **Win** (strangely, it does on **Linux**).
Could be fixed with : 
```c++
TimeSeriesSet timeSeriesSet(path.stem().string(), nTimestamps);
```
This makes unit tests pass.

guilpier-code (changes requested):

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+TimeSeriesSet TimeSeriesSetImporter::importFromFile(const std::filesystem::path& path,
+                                                    char csvSeparator)
+{
+    auto csvMatrix = csvToMatrix(path, csvSeparator);
+    // We have to transpose the matrix
+    // TODO: we may want to improve this by reading directly into the TimeSeriesSet object, or
+    // by creating a specific IDataSeries implementation
+    int nTimestamps = csvMatrix.size();
```
guilpier-code: **time stamps** ?
Do you still prefer this to **hours** (coming from the business nomenclature) ?

src/io/inputs/data-series-csv-importer/include/antares/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.h
```diff
@@ -0,0 +1,57 @@
...
+ * Every csv file found represents a DataSeries, its id will be the name of the file
+ */
+namespace Antares::IO::Inputs::DataSeriesCsvImporter
+{
+
+class TimeSeriesSetImporter
+{
+public:
+    TimeSeriesSetImporter() = delete; // must not be used
+    static Optimisation::LinearProblemDataImpl::TimeSeriesSet importFromFile(
```
guilpier-code: Why is the read function **static** ?
For which benefit ?

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
pet-mit: My initial thought was to allow both extensions for more flexibility, since they are the same formats except for the extension (and some tools like libre office Calc only allow writing csv). But I hadn't thought that by allowing both, the user can create two files with the same name, but different extensions. I'll have to add some checks and tests to prohibit this.

src/io/inputs/data-series-csv-importer/include/antares/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.h
```diff
@@ -0,0 +1,57 @@
...
+      const std::filesystem::path& path,
+      char csvSeparator = '\t');
+
+private:
+    static std::vector<std::vector<double>> csvToMatrix(const std::filesystem::path& path,
```
guilpier-code: There may be some kind of code duplication by having `csvSeparator = '\t'` twice.
In case it's obvious this class reads TS from a file, we could have **csvSeparator** as a data member : 
```c++
class TimeSeriesSetImporter
{
public:
    TimeSeriesSetImporter(char separator = '\n') : csvSeparator_(separator)
    {}
    ...

private:
    char csvSeparator_;
    ...
};
```

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+        {
+            throw std::invalid_argument(path.filename().string()
+                                        + ": rows have inconsistent number of columns");
+        }
+        result.push_back(row);
+    }
+    return result;
+}
+
+TimeSeriesSet TimeSeriesSetImporter::importFromFile(const std::filesystem::path& path,
```
guilpier-code: Or we could also have : 
```c++
TimeSeriesSet TimeSeriesSetImporter::importFromFile(const std::filesystem::path& path,
                                                    char csvSeparator)
{
    auto csvMatrix = csvToMatrix(path, csvSeparator);
    TimeSeriesSet timeSeriesSet(path.stem().string(), csvMatrix);
    return timeSeriesSet;
}
```
Indeed, I would not be shocked if class **TimeSeriesSet**  had a constructor that takes a "matrix" (`std::vector<std::vector<double>>`) as an argument.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+        {
+            std::unique_ptr<IDataSeries> timeSeriesSet = std::make_unique<TimeSeriesSet>(
+              TimeSeriesSetImporter::importFromFile(entry, csvSeparator));
+            repo.addDataSeries(std::move(timeSeriesSet));
+        }
+    }
+    return repo;
+}
+
+std::vector<std::vector<double>> TimeSeriesSetImporter::csvToMatrix(
```
guilpier-code: **_Draft remark, not exactly finished_**

This function is long and tedious.
Could it be shorter and more efficient ?
After all, we're not the first ones to read these matrix files ?

guilpier-code (changes requested):

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+        throw std::invalid_argument("Could not open file " + path.filename().string());
+    }
+    std::string line;
+    std::string element;
+    bool empty_line_found = false;
+    while (!infile.eof())
+    {
+        std::vector<double> row;
+        getline(infile, line);
+        if (line.empty())
```
guilpier-code, line 81: About : 
```c++
if(line.empty())
{
    empty_line_found = true;
    continue;
}
if (empty_line_found)
{
    // only accept empty lines in the end of the file
    throw std::invalid_argument(path.filename().string()
                 + ": empty line in the middle of the file");
}
```
- block `if (empty_line_found)` seems dead code (because of the previous statement `continue`)
- If we meet empty lines while reading, should we not just skip them ? Do we care about empty lines ?

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+    std::vector<std::vector<double>> result;
+    std::ifstream infile(path, std::ios_base::binary | std::ios_base::in);
+    if (!infile.is_open())
+    {
+        throw std::invalid_argument("Could not open file " + path.filename().string());
+    }
+    std::string line;
+    std::string element;
+    bool empty_line_found = false;
+    while (!infile.eof())
```
guilpier-code: Using : 
```c++
while (std::getline(infile, line))
{
    ...
}
```
makes code a tiny bit shorter.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+            continue;
+        }
+        if (empty_line_found)
+        {
+            // only accept empty lines in the end of the file
+            throw std::invalid_argument(path.filename().string()
+                                        + ": empty line in the middle of the file");
+        }
+        std::stringstream ss(line);
+        while (getline(ss, element, csvSeparator))
```
guilpier-code: Instead of using a while loop here (and rather use functional style) : 
```c++
while (getline(ss, element, csvSeparator))
{
    ...
}
```
we could (possibly) : 
- use `boost::split(splitLine, line, boost::is_any_of(csvSeparator));`
- use `boost::trim(line);`
- use `std::ranges::transform(...)` to convert `std::string` into `double`

guilpier-code: To ease the code readability, we could create a **readLine(...)** method / free function that contains the code to read a line.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#include <fstream>
+
+#include <boost/regex.hpp>
```
guilpier-code:
```c++
#include <boost/regex.hpp>
```
is unused

src/solver/modeler/main.cpp
pet-mit: system_linear_problem.Provide is currently independent from the ortools implementation and only relies on the ILinearProblem interface. I guess that's why it was separated like that.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
pet-mit: we actually used the term "timestep" in fillercontect and parameters file. i'll use that

src/io/inputs/data-series-csv-importer/include/antares/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.h
pet-mit: because there's no reason to have to create an instance of the importer. it makes it harder and more confusing to use than just a static method.

src/io/inputs/data-series-csv-importer/include/antares/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.h
pet-mit: same as above, I prefer to avoid instanciating utilitary classes when there's no real reason to. but I can remove the default value from the csvToMatrix method if you prefer ; it is private after all.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
pet-mit: but then we'd have to transpose the csvMatrix first, or create a new implementation of iDataSeries that inverts the indices (bc timeSeriesSet uses the rank as first index, and the timestep as the second, while the csv matrix has the rows (timesteps) as first index, and columns (scenarios) as second index)

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
pet-mit: I think the reason it's long is all the error cases handling. otherwise it would have fit into 10 lines

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -0,0 +1,138 @@
...
+using namespace Optimisation::LinearProblemDataImpl;
+
+DataSeriesRepository DataSeriesRepoImporter::importFromDirectory(const std::filesystem::path& path,
+                                                                 char csvSeparator)
+{
+    if (!is_directory(path))
+    {
+        throw std::invalid_argument("Not a directory: " + path.string());
+    }
+    DataSeriesRepository repo{};
```
guilpier-code: Please see [this](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2629)

guilpier-code (changes requested):

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
guilpier-code: **csvToMatrix** : does it use class **TimeSeriesSetImporter**'s data members ?
Otherwise, should not it be a free function ?

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
guilpier-code: ok.
Could we do : 
```c++
TimeSeriesSet TimeSeriesSetImporter::importFromFile(const std::filesystem::path& path,
                                                    char csvSeparator)
{
    auto csvMatrix = csvToMatrix(path, csvSeparator);
    return matrixToTimeSeries(csvMatrix, path.stem().string());
}
```
?

guilpier-code (changes requested):

src/solver/modeler/main.cpp
guilpier-code: ok.
`SystemLinearProblem::Provide(...)` seems to be a builder / factory.
I guess all we need to build a particular **ILinearProblem** resides in **parameters**.
Here, a builder could know from **parameters** that it should build an **ortools** problem.

pet-mit: yes it would be nice to be able to configure the implementation from outside the code. but for now the usage of ortools is hard-coded.

src/io/inputs/data-series-csv-importer/include/antares/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.h
guilpier-code: Class **TimeSeriesSetImporter** has no data member, and has 2 static methods.
So it seems we don't need a class here, but 2 separate free functions.
It's easier to call a simple free function than a static class member function : no need to prefix the call with class name.

guilpier-code (changes requested):

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
pet-mit: we accept an empty line in the end of the file (because some editors, like libreoffice calc, insert an empty line in the end of the file). if the empty line is met before, then the file is wrong (see test "empty_line")

guilpier-code (approved):

PR #2622, opened 2025-02-07
[ANT-2749] Support time dependency

a-zakir:


src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -117,4 +125,97 @@ LinearExpression LinearExpression::negate() const
...
+    for (auto [key, value]: other_linear_expressions)
+    {
+        if (result.contains(key))
+        {
+            result[key] += value;
+        }
+        else
+        {
+            result[key] = value;
+        }
```
flomnes: This shouldn't be necessary
```cpp
std::map<int, int> m;
m[0]++;
assert(m[0] == 1);
```

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -37,42 +45,195 @@ class EvalVisitorNotImplemented: public std::invalid_argument
...
+
+                if (!std::isfinite(result))
+                {
+                    throw EvalVisitorDivisionException(lhs, rhs, "is not a finite number");
+                }
+            }
+            catch (const std::exception& ex)
+            {
+                throw EvalVisitorDivisionException(lhs, rhs, ex.what());
+            }
```
flomnes: The try..catch adds nothing, better to remove it for clarity

Also, I'd compare rhs to zero instead of this, but I may be wrong

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -37,42 +45,195 @@ class EvalVisitorNotImplemented: public std::invalid_argument
...
+    {
+        return applyOperator(right, std::minus<>());
+    }
+
+    EvaluationResult operator*(const EvaluationResult& right) const
+    {
+        return applyOperator(right, std::multiplies<>());
+    }
+
+    struct SafeDivides
```
flomnes, line 73:
```suggestion
    struct CheckedDivisionFunctor
```

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -37,42 +45,195 @@ class EvalVisitorNotImplemented: public std::invalid_argument
...
+    }
+    else if (right.evaluationResultType == EvaluationResultType::CONSTANT)
+    {
+        result.values_ = values_;
+        for (double& v: result.values_)
+        {
+            v = op(v, right.value_);
+        }
+    }
+    else if (values_.size() == right.values_.size())
```
flomnes: error in case of size mismatch ?

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -37,42 +45,195 @@ class EvalVisitorNotImplemented: public std::invalid_argument
...
+    }
+
+    struct SafeDivides
+    {
+        double operator()(double lhs, double rhs) const
+        {
+            // if (rhs == 0.0)
+            // {
+            //     throw std::runtime_error("Division by zero in EvaluationResult.");
+            // }
```
flomnes:
```suggestion
```

src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -21,109 +21,128 @@
...
 }
 
-double EvalVisitor::visit(const Nodes::MultiplicationNode* node)
+EvaluationResult EvalVisitor::visit(const Nodes::MultiplicationNode* node)
 {
     return dispatch(node->left()) * dispatch(node->right());
 }
 
-double EvalVisitor::visit(const Nodes::DivisionNode* node)
+EvaluationResult EvalVisitor::visit(const Nodes::DivisionNode* node)
```
flomnes, line 58: Missing implementation here

src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -21,109 +21,128 @@
...
+        return EvaluationResult{context_.getSystemParameterValueAsDouble(node->value())};
+    }
+    else
+    {
+        std::vector<double> params;
+        params.reserve(dataSeriesKeys_.timeSteps.size());
+        for (auto timeStep: dataSeriesKeys_.timeSteps)
+        {
+            params.emplace_back(
+              context_.getParameterValue(context_.getSystemParameterValue(node->value()),
```
flomnes: The API for `context_` is a bit messy

src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/linearProblem.h
```diff
@@ -45,16 +45,53 @@ class ILinearProblem
...
+                                                      const std::vector<double>& ub,
+                                                      const std::string& name,
+                                                      unsigned int number_new_variables)
+      = 0;
+    /// Create range of continuous variables
+    virtual std::vector<IMipVariable*> addNumVariable(const std::vector<double>& lb,
+                                                      double ub,
+                                                      const std::string& name,
+                                                      unsigned int number_new_variables)
+      = 0;
```
flomnes: Please don't complexify `ILinearProblem` so much. This complexity belongs somewhere else (too many functions)

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -44,6 +44,20 @@ bool checkTimeSteps(Optimisation::LinearProblemApi::FillContext& ctx)
...
+    if (evaluation_result.getEvaluationResultType()
+        == Expressions::Visitors::EvaluationResultType::CONSTANT)
+    {
+        return std::variant<double, std::vector<double>>{evaluation_result.value()};
+    }
+    else
+    {
+        return std::variant<double, std::vector<double>>{evaluation_result.values()};
+    }
+}
```
flomnes: # Style
You can use `std::variant<double, std::vector<double>>` as the return type instead of `auto` and drop the `std::variant<...>` in the return statements

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -37,42 +45,195 @@ class EvalVisitorNotImplemented: public std::invalid_argument
...
+    EvaluationResult applyOperator(const EvaluationResult& right, Op op) const;
+    template<typename Op>
+    EvaluationResult applyUnaryOperator(Op op) const;
+};
+
+template<typename Op>
+EvaluationResult EvaluationResult::applyOperator(const EvaluationResult& right, Op op) const
+{
+    EvaluationResult result(0.0);
+
```
a-zakir, line 149: Update type

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
tbittar: I agree with Florian, I would compare `rhs` to zero with respect to a given tolerance (e.g. 1e-16) to avoid numerical issues with float equality comparisons. But it seems you have already tried it in the commented part, so there is a maybe a good reason for this ? Probably you can handle more cases like dividing a very large number with a very small one ?

This kind of equality comparison between floats and 0, 1 or other interesting value may be often used (e.g. for online simplification of expressions later on, avoiding having unecessarily large trees : 1 * x -> x), so it can be interesting to have somewhere a function that performs float equality comparison given a numerical tolerance

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -117,4 +125,115 @@ LinearExpression LinearExpression::negate() const
...
+TimeDependentLinearExpression TimeDependentLinearExpression::operator-(
+  const TimeDependentLinearExpression& other) const
+{
+    // checkOtherLength(other);
+    const auto& linear_expressions = GetLinearExpressions();
+    const auto& other_linear_expressions = other.GetLinearExpressions();
+    std::map<unsigned int, LinearExpression> result;
+    for (size_t i = 0; i < linear_expressions.size(); ++i)
+    {
+        result[i] = linear_expressions.at(i) - other_linear_expressions.at(i);
```
tbittar: When adding or subtracting linear expression, we may end up with variables having coefficient zero (as they are not deleted from the map), does this lead to errors when adding such variables in the solver ? My insight is that it could...

src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/ILinearProblemData.h
```diff
@@ -2,17 +2,30 @@
 #pragma once
 
 #include <string>
+#include <vector>
 
 namespace Antares::Optimisation::LinearProblemApi
 {
 
+struct DataSeriesKeys
```
pet-mit: what is the purpose of this struct ?

src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/linearProblem.h
a-zakir: ok

src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/ILinearProblemData.h
a-zakir: i was using it to "transport"  scenario and scenarioGroup from main to where they were supposed to be used.
Since scenario is not managed in this pr, i will have to update

src/expressions/visitors/EvaluationContext.cpp
```diff
@@ -14,8 +17,59 @@ double EvaluationContext::getVariableValue(const std::string& key) const
...
+    catch (const std::invalid_argument&)
+    {
+        throw CouldNotEvaluateConstantParameter<std::invalid_argument>(
+          "Parameter '" + key + "' has an invalid numerical format: '" + value + "'.");
+    }
+    catch (const std::out_of_range&)
+    {
+        throw CouldNotEvaluateConstantParameter<std::out_of_range>(
+          "Parameter '" + key + "' is out of numerical range: '" + value + "'.");
+    }
```
flomnes: Extract this to a dedicated `noexcept` conversion function

src/optimisation/linear-problem-data-impl/include/antares/optimisation/linear-problem-data-impl/dataSeries.h
```diff
@@ -15,6 +16,7 @@ class IDataSeries
     }
 
     virtual double getData(unsigned int rank, unsigned int hour) = 0;
+    virtual std::vector<double> getData(unsigned int rank) = 0;
```
flomnes: Is it possible to avoid copies here ?
```suggestion
    virtual const std::vector<double>& getData(unsigned int rank) = 0;
```

src/optimisation/linear-problem-data-impl/include/antares/optimisation/linear-problem-data-impl/linearProblemData.h
```diff
@@ -45,6 +45,9 @@ class LinearProblemData: public LinearProblemApi::ILinearProblemData
                    const std::string& scenarioGroup,
                    const unsigned scenario,
                    const unsigned hour) override;
+    std::vector<double> getData(const std::string& dataSetId,
+                                const std::string& scenarioGroup,
+                                const unsigned scenario) override;
```
flomnes:
```suggestion
    const std::vector<double>& getData(const std::string& dataSetId,
                                const std::string& scenarioGroup,
                                const unsigned scenario) override;
```

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -54,21 +175,37 @@ void ComponentFiller::addVariables(Optimisation::LinearProblemApi::ILinearProble
...
+        const auto& lb = evaluator.dispatch(variable.LowerBound().RootNode());
+        const auto& ub = evaluator.dispatch(variable.UpperBound().RootNode());
         if (variable.isTimeDependent())
         {
-            pb.addVariable(evaluator.dispatch(variable.LowerBound().RootNode()),
-                           evaluator.dispatch(variable.UpperBound().RootNode()),
-                           variable.Type() != Study::SystemModel::ValueType::FLOAT,
-                           component_.Id() + "." + variable.Id(),
-                           ctx.getNumberOfTimestep());
+            std::visit(
```
flomnes, line 182: Please add a comment here to explain that you're using `std::visit` to handle the 4 cases: double/double, vector/double, double/vector and vector/vector.

src/solver/optim-model-filler/ReadLinearConstraintVisitor.cpp
```diff
@@ -39,92 +42,129 @@ std::string ReadLinearConstraintVisitor::name() const
...
+    return constraints;
 }
 
-LinearConstraint ReadLinearConstraintVisitor::visit(const GreaterThanOrEqualNode* node)
+std::vector<LinearConstraint> ReadLinearConstraintVisitor::visit(const GreaterThanOrEqualNode* node)
 {
+    // auto leftMinusRight = linear_expression_visitor_.dispatch(node->left())
+    //                       - linear_expression_visitor_.dispatch(node->right());
+    // return LinearConstraint{.coef_per_var = leftMinusRight.coefPerVar(),
+    //                         .lb = -leftMinusRight.offset()};
```
flomnes:
```suggestion
```

flomnes (commented):
Good job, only a few last comments

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
a-zakir: update -> d7f4d2125d15d31d91dd986ca3c3f1e21d8d1ff4

src/solver/optim-model-filler/LinearExpression.cpp
a-zakir: Well i don't know

src/expressions/visitors/EvaluationContext.cpp
a-zakir: noexcept(false)? isn't this the default behavior?

src/optimisation/linear-problem-data-impl/include/antares/optimisation/linear-problem-data-impl/dataSeries.h
a-zakir: I've removed this function to avoid checking the consistency between the requested data and the vector size.

flomnes (commented):
`TimeDependentLinearExpression` is not tested

a-zakir:
> `TimeDependentLinearExpression` is not tested
c48c256d8d4aca4b0006675c7d7e3e538a6ac3d2

src/expressions/include/antares/expressions/visitors/EvaluationContext.h
```diff
@@ -44,18 +61,34 @@ class EvaluationContext
...
+    {
+        using T::T;
+    };
 
 private:
     /**
      * @brief A map storing parameter values.
      */
-    std::map<std::string, double> parameters_;
+    std::map<std::string, ContextParameter> system_parameters_;
```
pet-mit: this class isn't supposed to know what a system is. maybe some name like this?
```suggestion
    std::map<std::string, ContextParameter> parameters_types_and_values_;
```

src/expressions/include/antares/expressions/visitors/EvaluationContext.h
```diff
@@ -44,18 +61,34 @@ class EvaluationContext
...
 
     /**
      * @brief A map storing variable values.
      */
     std::map<std::string, double> variables_;
+    Optimisation::LinearProblemApi::ILinearProblemData& data_;
```
pet-mit, line 91: This kind of usage makes me think that we should move & rename ILinearProblemData to something more generic, to imply that it is independent from LinearProblem. But it's not in the scope of this PR.

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -38,41 +48,219 @@ class EvalVisitorNotImplemented: public std::invalid_argument
...
-    double visit(const Nodes::GreaterThanOrEqualNode* node) override;
-    double visit(const Nodes::NegationNode* node) override;
-    double visit(const Nodes::VariableNode* node) override;
-    double visit(const Nodes::ParameterNode* node) override;
-    double visit(const Nodes::LiteralNode* node) override;
-    double visit(const Nodes::PortFieldNode* node) override;
-    double visit(const Nodes::PortFieldSumNode* node) override;
-    double visit(const Nodes::ComponentVariableNode* node) override;
-    double visit(const Nodes::ComponentParameterNode* node) override;
+    Optimisation::LinearProblemApi::FillContext fillContext_;
```
pet-mit, line 249: also we should move & rename FillContext so that visitors don't depend on LinearProblemApi

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -52,22 +55,51 @@ class LinearExpression
...
+
+    LinearExpression& operator+=(const LinearExpression& value);
 
 private:
     double offset_ = 0;
     std::map<std::string, double> coef_per_var_;
 };
+
+// time dependent parameter
+class TimeDependentLinearExpression
```
pet-mit: why is it in the same .h / .cpp as LinearExpression ?

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ComponentFiller.h
```diff
@@ -79,7 +78,41 @@ class ComponentFiller: public Optimisation::LinearProblemApi::LinearProblemFille
     bool IsThisVariableTimeDependent(const std::string& var_id) const;
 
     const Study::SystemModel::Component& component_;
-    Expressions::Visitors::EvaluationContext evaluationContext_;
     const std::map<std::string, Study::SystemModel::Variable>& modelVariable_;
 };
+
+// TODO move me
+class VariablesBulkAddition
```
pet-mit, line 84: as these are helper functions for ComponentFiller, I think you can create free static functions in ComponentFiller.cpp, or at least make this class private to ComponentFiller

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -31,9 +32,121 @@
...
+    {
+        variables[i] = linear_problem_.addVariable(lb[i],
+                                                   ub,
+                                                   integer,
+                                                   name + "_" + std::to_string(i));
+    }
+    return variables;
+}
+
+std::vector<Optimisation::LinearProblemApi::IMipVariable*> VariablesBulkAddition::addVariable(
```
pet-mit, line 99: avoid duplication by calling the method with lb and ub both vectors inside this function?

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -91,24 +218,22 @@ void ComponentFiller::addStaticConstraint(Optimisation::LinearProblemApi::ILinea
...
+                                    component_.Id() + "." + constraint_id + '_'
+                                      + std::to_string(constraint_count));
         for (const auto& [var_id, coef]: linear_constraint.coef_per_var)
         {
-            // TODO FIXME the coefficient needs to be time-dependent
             if (IsThisVariableTimeDependent(var_id))
             {
                 auto* variable = pb.getVariable(component_.Id() + "." + var_id + '_'
-                                                + std::to_string(cstr));
+                                                + std::to_string(constraint_count));
```
pet-mit: Now that the visitor is able to produce one LinearConstraint per t, I believe that it is its role to select the variable's time index that should be used in the constraint. Thus the ID of the variable should be complete in the coef_per_var map, minus the component ID, for now (we'll see how things change with the dictionary)

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -169,11 +302,12 @@ void ComponentFiller::addObjective(Optimisation::LinearProblemApi::ILinearProble
     {
         if (IsThisVariableTimeDependent(var_id))
         {
-            for (auto var_pos = 0; var_pos != ctx.getNumberOfTimestep(); ++var_pos)
+            for (auto var_pos = ctx.getFirstTimeStep(); var_pos <= ctx.getLastTimeStep(); ++var_pos)
```
pet-mit, line 302: same here

src/study/system-model/component.cpp
```diff
@@ -96,7 +96,7 @@ ComponentBuilder& ComponentBuilder::withModel(const Model* model)
  * \return Reference to the ComponentBuilder object.
  */
 ComponentBuilder& ComponentBuilder::withParameterValues(
-  std::map<std::string, double> parameter_values)
+  std::map<std::string, Expressions::Visitors::ContextParameter> parameter_values)
```
pet-mit: what do you mean by the name "ContextParameter"? I find it confusing

src/tests/cucumber/features/steps/common_steps/steps.py
```diff
@@ -34,6 +36,15 @@ def simu_success(context):
     assert context.return_code == 0
 
 
+@then('the modeler simulation succeeds')
+def simu_success(context):
+    print(f"context.return_code = {context.return_code} ***********************")
+    assert context.return_code == 0
+    context.output_path = os.path.join(context.study_path, "output")  # TODO : fixme parse_output_folder_from_logs(out)
+    context.moh = modeler_output_handler(context.output_path)
```
pet-mit: instead of adding this step, just check that return_code==0 before instantiating "moh" in "when i run modeler" step:
- save you from creating a new step
- it is not the role of "then the simulation succeeds" step to import outputs, since the user may chose not to run this check (but they will still need access to the outputs)

pet-mit (changes requested):

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ComponentFiller.h
a-zakir: it was private, i decided to expose to unit test it

src/expressions/include/antares/expressions/visitors/EvaluationContext.h
a-zakir: done

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
a-zakir: done

src/solver/optim-model-filler/ComponentFiller.cpp
a-zakir: until then i propose to get the time step corresponding to each constraints from ReadLinearConstraintVisitor -> [pr](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2656/files)

pet-mit: the PR is too complex, I was thinking it would be enough to add the "_ts" suffix to the variable name inside the visitor. maybe wait until https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2655 is merged and use the variable full key instead of the string variable ID

src/tests/resources/modeler/1_1/parameters.yml
```diff
@@ -1,6 +1,6 @@
-solver: sirius
```
pet-mit: the CI can't run XPRESS

src/solver/optim-model-filler/ComponentFiller.cpp
a-zakir: still the ts number is needed

src/solver/optim-model-filler/ComponentFiller.cpp
a-zakir: and create useless vector in case one of ub or lb is a simple value?

pet-mit: you're right it might fire back on us. let's leave it like this for now

src/tests/cucumber/features/steps/common_steps/steps.py
a-zakir: done

src/tests/resources/modeler/1_1/parameters.yml
a-zakir: indeed

src/study/system-model/component.cpp
a-zakir: renamed ParameterTypeAndValue

pet-mit (approved):

PR #2625, opened 2025-02-07
Fix the upgrade from v8.8 to 9.2 of hydro power matrices [ANT-2742]

guilpier-code:
Fixes problem described in issue [ANT-2742](https://gopro-tickets.rte-france.com/browse/ANT-2742)

This fix is a quick fix : code could be cleaner. 
Some refactoring could make code more readable when reading it again in a few months / years.
In particular, we could create an interface for reading and saving this data (hydro power) and create 2 classes implementing this interface and encapsulating each case (read/write daily or hourly)

### DIFF
--- a/src/libs/antares/study/area/list.cpp
+++ b/src/libs/antares/study/area/list.cpp
@@ -774,7 +774,7 @@ bool AreaList::saveToFolder(const AnyString& folder) const
     // Hydro
     // The hydro files must be saved after the area has been invalidated
     buffer.clear() << folder << SEP << "input" << SEP << "hydro";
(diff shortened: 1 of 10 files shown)


flomnes (commented):
LGTM. `enums` don't need to be passed by reference
```c++
enum Color
{
    Red, Blue
};
void printColor(Color c);
```

Also, formatting

guilpier-code:
> LGTM. `enums` don't need to be passed by reference
> 
> ```c++
> enum Color
> {
>     Red, Blue
> };
> void printColor(Color c);
> ```
> 
> Also, formatting

Done

flomnes (approved):

PR #2626, opened 2025-02-07
Improve modeler tests

pet-mit:
- Added a modeler test with an int variable (have to used scip bc sirius does not support this)
- Added the objective value to the output file
- Added test steps to test the objective value
- Added test steps to make it more simple to test variable values
- Added description to existing tests


src/tests/resources/modeler/epic2/us2.5/study_2.5.4/parameters.yml
```diff
@@ -0,0 +1,6 @@
+solver: sirius
```
flomnes: sirius doesn't support MILP problems

src/tests/resources/modeler/epic2/us2.5/study_2.5.4/output/problem.lp
flomnes, line 1: Does the test also check the produced LP file ?

src/tests/resources/modeler/epic2/us2.5/study_2.5.4/parameters.yml
pet-mit: thanks I forgot. maybe we should add an explicit exception / log

src/tests/resources/modeler/epic2/us2.5/study_2.5.4/output/problem.lp
pet-mit: no since it's supposed to be an end-to-end test

flomnes (approved):

PR #2630, opened 2025-02-10
Improve vcpkg configuration

JasonMarechal25:
Use prescribed configuration with two files, one for application (vcpkg.json) one for configuration (vcpkg-configuration.json).

Use sirius in Antares registry instead of local port.


src/vcpkg-configuration.json
```diff
@@ -0,0 +1,25 @@
...
+    "kind": "git",
+    "baseline": "b322364f06308bdd24823f9d8f03fe0cc86fd46f",
+    "repository": "https://github.com/microsoft/vcpkg"
+  },
+  "registries": [
+    {
+      "kind": "artifact",
+      "location": "https://github.com/microsoft/vcpkg-ce-catalog/archive/refs/heads/main.zip",
+      "name": "microsoft"
+    },
```
flomnes: Why is this registry needed here ? It seems out-of-date/unmaintained

src/vcpkg-configuration.json
```diff
@@ -0,0 +1,25 @@
...
+    {
+      "kind": "git",
+      "repository": "https://github.com/AntaresSimulatorTeam/antares-vcpkg-registry",
+      "baseline": "3fa279503f88d764778bcaacdacd3aa27e323025",
+      "packages": [
+        "sirius-solver"
```
flomnes: Please use antares-integration-v1.7 (currently v1.5) on that registry

flomnes (commented):
If we're going to use an external registry, I think it would be nice to keep a trace of the versions of each dependency.

Supposing https://github.com/AntaresSimulatorTeam/antares-vcpkg-registry was deleted, there would be no way to determine what version of e.g sirius-solver was used for e.g Antares Simulator 9.2.0

src/vcpkg-configuration.json
JasonMarechal25: This is generated by default by vcpkg. I will check

src/vcpkg-configuration.json
JasonMarechal25: I've updated the registry with sirius 1.7

flomnes (approved):

PR #2633, opened 2025-02-11
Fix migration guide

flomnes:
- Fix `hydro-pmax-format` -> `hydro-pmax` (as in the code)
- Remove section related to TS generation for links
- Add precisions where needed


MartinBelthle (approved):
Thanks for the notif. 
Also, you could remove the whole part concerning the links generation as we said it will be handled in AntaresWeb

flomnes:
> Thanks for the notif. Also, you could remove the whole part concerning the links generation as we said it will be handled in AntaresWeb

done

PR #2635, opened 2025-02-11
Fix binding constraint TS loading in 8.8 [ANT-2746]

payetvin:


src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
```diff
@@ -123,6 +123,9 @@ std::vector<std::shared_ptr<BindingConstraint>> BindingConstraintLoader::load(En
             if (!Math::Zero(o))
                 bc->offset(clstr, o);
 
+            if (!clstr->isActive())
+                bc->pEnabled = false;
```
flomnes: Problem here, see review message

flomnes (changes requested):
# Issues with the GUI
1. Load a study with an enabled BC containing an inactive thermal cluster
2. The loader disables the BC (this code)
3. Save the study. Since `enabled` property for the BC is set to `false`, it is also set to false in bindingconstraints.ini. The initial value is overwritten

# Issues with antares-solver : breaking change
The current behavior is to ignore inactive clusters in binding constraints. This line of code will disable the BC completely, which is a breaking change.

# Suggested change
Inactive thermal clusters are already filtered out during the translation from structure `Antares::Data::BindingConstraint` to optimization problems (see method `BindingConstraint::initLinkArrays`), so there is no need to filter them when loading from files.

```cpp
auto cEnd = pClusterWeights.end();
for (auto i = pClusterWeights.begin(); i != cEnd; ++i)
{
    if (i->first->isActive())
    // add term to binding constraint
}
```
My suggestion is to remove l. 126 and 127 from BindingConstraint.cpp.

src/libs/antares/study/binding_constraint/BindingConstraintSaver.cpp
```diff
@@ -65,6 +65,7 @@ bool BindingConstraintSaver::saveToEnv(EnvForSaving& env, const BindingConstrain
 
     // Exporting the matrix
     env.matrixFilename.clear() << bindingConstraint->timeSeriesFileName(env);
-    return bindingConstraint->RHSTimeSeries_.saveToCSVFile(env.matrixFilename.c_str());
+    return bindingConstraint->RHSTimeSeries_.saveToCSVFile(env.matrixFilename.c_str(), 1, false, true);
```
flomnes: Don't change the save function
```suggestion
    return bindingConstraint->RHSTimeSeries_.saveToCSVFile(env.matrixFilename.c_str(), 1, false, true);
```
Modify function `BindingConstraintGroupRepository::timeSeriesWidthConsistentInGroups` to be less restrictive, accepting width of 1 (see function `checkAllElementsIdenticalOrOne`)

src/libs/antares/study/binding_constraint/BindingConstraint.cpp
```diff
@@ -545,6 +545,10 @@ double BindingConstraint::weight(const AreaLink* lnk) const
 
 double BindingConstraint::weight(const ThermalCluster* cluster) const
 {
+    if (!cluster->isEnabled())
```
flomnes: Add a comment : "this method should only be used in the GUI, it doesn't handle inactive clusters well"

```suggestion
```

src/libs/antares/study/binding_constraint/BindingConstraint.cpp
```diff
@@ -555,9 +559,13 @@ int BindingConstraint::offset(const AreaLink* lnk) const
...
 }
 
-int BindingConstraint::offset(const ThermalCluster* lnk) const
+int BindingConstraint::offset(const ThermalCluster* cluster) const
 {
-    auto i = pClusterOffsets.find(lnk);
+    if (!cluster->isEnabled())
+    {
+        return 0;
+    }
```
flomnes:
```suggestion
```

flomnes (changes requested):
As discussed,
- Weights & offsets: just remove condition `isActive` from `BindingConstraint::weight(const ThermalCluster* cluster, double w)` and `BindingConstraint::offset(const ThermalCluster* cluster, int o)`
- Time-series: make method `BindingConstraintGroupRepository::timeSeriesWidthConsistentInGroups()` more tolerant with TS of width 1

src/libs/antares/study/binding_constraint/BindingConstraintGroupRepository.cpp
```diff
@@ -36,7 +36,7 @@ namespace Antares::Data {
                                                 return false;
                                             auto width = (*constraints.begin())->RHSTimeSeries().width;
                                             bool isConsistent = std::all_of(constraints.begin(), constraints.end(), [&width](const std::shared_ptr<BindingConstraint>& bc){
-                                                              bool sameWidth = bc->RHSTimeSeries().width == width;
+                                                              bool sameWidth = bc->RHSTimeSeries().width == width || bc->RHSTimeSeries().width == 1;
```
flomnes: Not the right behavior. Please backport the previous functions

PR #2641, opened 2025-02-14
Add end to end tests to sonarcloud

payetvin:


pet-mit:
you can also add modeler cucumber tests

flomnes:
Some studies require too much RAM when the line coverage analysis is enabled and thus can't be run on standard GH runners

flomnes (approved):

PR #2643, opened 2025-02-17
Fix std::bad_array_new_length exception in SurveyResults::SurveyResults

flomnes:
This happens only if there is at least one area and no output variables (example study valid-filter-outputs/no-geo-trim--noVar)

The original code is equivalent to 
```cpp
bool* x = new bool[0]{false};
```
which throws a `std::bad_array_new_length` exception as shown [here](https://en.cppreference.com/w/cpp/memory/new/bad_array_new_length).

### DIFF
--- a/src/solver/variable/surveyresults/surveyresults.cpp
+++ b/src/solver/variable/surveyresults/surveyresults.cpp
@@ -558,7 +558,11 @@ SurveyResults::SurveyResults(const Data::Study& s, const Yuni::String& o, IResul
     digestNonApplicableStatus = new bool*[digestSize];
     for (uint i = 0; i < digestSize; i++)
     {
-        digestNonApplicableStatus[i] = new bool[maxVariables]{false};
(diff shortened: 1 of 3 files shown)


payetvin (approved):

src/solver/variable/surveyresults/surveyresults.cpp
```diff
@@ -558,7 +558,11 @@ SurveyResults::SurveyResults(const Data::Study& s, const Yuni::String& o, IResul
     digestNonApplicableStatus = new bool*[digestSize];
     for (uint i = 0; i < digestSize; i++)
     {
-        digestNonApplicableStatus[i] = new bool[maxVariables]{false};
+        digestNonApplicableStatus[i] = new bool[maxVariables];
```
JasonMarechal25, line 561: **new**
Can't we do better than using new ?

flomnes: Yes, absolutely

PR #2648, opened 2025-02-19
Improve logging & loading of STS elements introduced in v9.2.0 [ANT-2730]

flomnes:
- Add logging to allow for easier debugging, especially for optional files
- Only try to load new STS stuff if the study version indicates that we should

**One test fails (ST-storage-cost) because the study is in version 8.6.0. But it contains new costs introduced in v9.2.0, that are not loaded. This failure is therefore expected and perfectly normal. Reference results need to be (once again) regenerated.**


JasonMarechal25 (approved):
LGTM

src/tests/src/libs/antares/study/short-term-storage-input/short-term-storage-input-output.cpp
```diff
@@ -270,37 +270,37 @@ BOOST_FIXTURE_TEST_CASE(check_series_folder_loading_different_values, Fixture)
 {
     createFileSeries(8760);
 
-    BOOST_CHECK(series.loadFromFolder(folder));
+    BOOST_CHECK(series.loadFromFolder(folder, StudyVersion::latest()));
```
a-zakir: add test with old versions

a-zakir (approved):

PR #2655, opened 2025-02-21
Add variable dictionnary in ComponentFiller

flomnes:
- Remove unused return value for `VariablesBulkAddition::addVariable`
- Add wrapper around 4-key templated container. For variables, it is non-owning (raw pointers)
- Adapt existing tests


src/solver/optim-model-filler/include/antares/solver/optim-model-filler/variableDict.h
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return hmv.at(k.getPartialKey())[k.getTimestep()][k.getScenario()];
+    }
+
+    inline Value& operator[](const FullKey& k)
+    {
+        return hmv[k.getPartialKey()][k.getTimestep()][k.getScenario()];
+    }
+
+    inline const TwoIndexVector& operator[](const PartialKey& k) const
```
pet-mit: I feel that the PartialKey may become confusing. What prevents us from using the full key when using this dictionary ?

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/variableDict.h
```diff
@@ -0,0 +1,108 @@
...
+
+class hash
+{
+public:
+    std::size_t operator()(const PartialKey& p) const;
+};
+
+std::string buildVariableName(const FullKey& key, int timestep, int scenario);
+
+template<class Value>
```
pet-mit: I think we can avoid templating this class since it is not useful and adds extra complexity

src/solver/optim-model-filler/CMakeLists.txt
```diff
@@ -5,10 +5,12 @@ set(SRC_optim_model_filler
         include/antares/solver/optim-model-filler/LinearExpression.h
         include/antares/solver/optim-model-filler/ReadLinearConstraintVisitor.h
         include/antares/solver/optim-model-filler/ReadLinearExpressionVisitor.h
+        include/antares/solver/optim-model-filler/variableDict.h
         ComponentFiller.cpp
         LinearExpression.cpp
         ReadLinearConstraintVisitor.cpp
         ReadLinearExpressionVisitor.cpp
+	variableDict.cpp
```
pet-mit: let's rename this to VariableDictionary.h/cpp

src/solver/optim-model-filler/variableDict.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    std::string ret = key.getComponent() + "." + key.getVariable();
+    if (key.getTimestep() > 1)
+    {
+        ret += "_" + std::to_string(timestep);
+    }
+    if (key.getScenario() > 1)
+    {
+        ret += "_" + std::to_string(scenario);
+    }
```
pet-mit: why the if's ?

pet-mit: I think this is what leads to cucumber failure on modeler

flomnes: Previously it was implemented like this, but I agree we should always add the scenario & timestep number.

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/variableDict.h
flomnes: I agree it's useless, let's remove this `PartialKey` or make it private

src/solver/optim-model-filler/variableDict.cpp
a-zakir: i agree with @pet-mit , we need different treatment between time dependent and static variables

a-zakir: That said,  always include the time step and scenario for time-dependent variables, and only component_id.variable_name for others.

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -232,13 +228,15 @@ void ComponentFiller::addTimeDependentConstraints(
...
-                                                + std::to_string(constraint_count));
+                // TODO scenario = 0
+                auto* variable = variableDict(component_.Id(), var_id, constraint_count, 0);
+
                 ct->setCoefficient(variable, coef);
             }
             else
             {
-                auto* variable = pb.getVariable(component_.Id() + "." + var_id);
+                auto* variable = variableDict(component_.Id(), var_id, 0, 0);
```
a-zakir: why?

flomnes: Because this variable is time & scenario independent

a-zakir: Then it should not be tied to time and scenario 0

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -175,28 +157,42 @@ void ComponentFiller::addVariables(Optimisation::LinearProblemApi::ILinearProble
     {
         const auto& lb = evaluator.dispatch(variable.LowerBound().RootNode());
         const auto& ub = evaluator.dispatch(variable.UpperBound().RootNode());
+        constexpr int numberOfScenarios = 1; // FIXME ONE SCENARIO ONLY
         if (variable.isTimeDependent())
         {
+            const FullKey key(component_.Id(),
+                              variable.Id(),
+                              ctx.getLastTimeStep() - ctx.getFirstTimeStep() + 1,
```
a-zakir: well, in the previous code it was possible to have a first time step other than 0. In think this could be source of a malicious bug

a-zakir: i mean first and last ts were actually used ; (t_0 = 3, tf = 5)  ≠ (t_0 = 0, tf = 2)

src/solver/optim-model-filler/VariableDictionary.cpp
```diff
@@ -0,0 +1,230 @@
...
+{
+    return nbScenarios.has_value();
+}
+
+std::vector<int> Dimensions::getTimesteps() const
+{
+    if (timeInterval)
+    {
+        std::vector<int> ret(timeInterval->finalTime - timeInterval->initialTime + 1);
+        for (int t = timeInterval->initialTime; t <= timeInterval->finalTime; ++t)
```
a-zakir: Std::iota?

src/solver/optim-model-filler/VariableDictionary.cpp
```diff
@@ -0,0 +1,230 @@
...
+}
+
+std::vector<int> Dimensions::getScenarioIndices() const
+{
+    if (nbScenarios)
+    {
+        std::vector<int> ret(*nbScenarios);
+        for (int s = 0; s < *nbScenarios; ++s)
+        {
+            ret[s] = s;
```
a-zakir: same

src/solver/optim-model-filler/VariableDictionary.cpp
```diff
@@ -0,0 +1,230 @@
...
+        {
+            ret[t - timeInterval->initialTime] = t;
+        }
+        return ret;
+    }
+    else
+    {
+        // Arbitrary
+        return {0};
+    }
```
a-zakir:
```suggestion
    if (timeInterval)
    {
        std::vector<int> ret(timeInterval->finalTime - timeInterval->initialTime + 1);
        std::iota(ret.begin(), ret.end(), timeInterval->initialTime);
        return ret;
    }
    else
    {
        // Arbitrary
        return {0};
    }
```

src/solver/optim-model-filler/VariableDictionary.cpp
```diff
@@ -0,0 +1,230 @@
...
+std::vector<int> Dimensions::getScenarioIndices() const
+{
+    if (nbScenarios)
+    {
+        std::vector<int> ret(*nbScenarios);
+        for (int s = 0; s < *nbScenarios; ++s)
+        {
+            ret[s] = s;
+        }
+        return ret;
```
a-zakir:
```suggestion
        std::ioata(ret.begin(), ret.end(), 0);
        return ret;
```

a-zakir (approved):

PR #2662, opened 2025-02-24
Modeler 4.7: Import connections and ports

payetvin:


src/io/inputs/model-converter/include/antares/io/inputs/model-converter/modelConverter.h
```diff
@@ -50,5 +50,34 @@ class UnknownTypeException: public std::runtime_error
     explicit UnknownTypeException(YmlModel::ValueType type);
 };
 
+class ObjectWithThisIdAlreadyExists: public std::runtime_error
```
JasonMarechal25: Maybe "DuplicatePortIdError" ?

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -139,17 +170,82 @@ std::vector<Antares::Study::SystemModel::Variable> convertVariables(const YmlMod
...
-  [[maybe_unused]] const Antares::IO::Inputs::YmlModel::Model& model)
+  const Antares::IO::Inputs::YmlModel::Model& model,
+  const std::vector<Antares::Study::SystemModel::PortType>& portTypes)
+{
+    std::vector<Antares::Study::SystemModel::Port> ports;
+    ports.reserve(model.ports.size());
+    for (const auto& port: model.ports)
+    {
+        // Can't have port with the same ID
+        if (std::ranges::find_if(ports, [&port](const auto& p) { return p.Id() == port.id; })
```
JasonMarechal25, line 204: In line with legacy behaviour maybe we should collect all errors.
Somethings like :
for port in model.ports {
  for each port {
     if duplicate id => errors.add(ObjectWithThisIdAlreadyExists);
     if idNotFound => errors.add(PortTypeNotFound)
 }
 }
 if (error) {
 		log
 		throw/exit/whatever
 		}

JasonMarechal25, line 204: Instead of find_if you can store ids in a map.
std::map<PortID, bool> already_foundId;

for port in model.ports{
   if (already_foundId[port.id])) => error
   already_foundId[port.id] = true;
...

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/modelConverter.h
```diff
@@ -47,7 +47,49 @@ namespace Antares::IO::Inputs::ModelConverter
 class UnknownTypeException: public std::runtime_error
```
JasonMarechal25, line 51: Do we have unit test that catch each exceptions ?

payetvin: Yes

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -33,49 +33,88 @@
...
-        std::vector<Antares::Study::SystemModel::PortField> fields;
+        if (portType.fields.empty()) // Can't have a port type without fields
+        {
+            throw PortTypeDoesntContainsFields(portType.id);
+        }
+        std::vector<Study::SystemModel::PortField> fields;
         for (const auto& field: portType.fields)
         {
-            fields.emplace_back(Antares::Study::SystemModel::PortField{field});
+            fields.emplace_back(Study::SystemModel::PortField{field});
```
flomnes:
```suggestion
            fields.emplace_back(field);
```

flomnes (commented):
Many `std::ranges::find_if` could be replaced with std::map::contains (if we used maps instead of vectors)

src/study/system-model/model.cpp
```diff
@@ -136,4 +136,23 @@ ModelBuilder& ModelBuilder::withConstraints(std::vector<Constraint>&& constraint
...
+ * inputs it not garanteed to be valid after the call
+ */
+ModelBuilder& ModelBuilder::withPortFieldDefinitions(
+  std::vector<PortFieldDefinition>&& portFieldDefinitions)
+{
+    std::transform(portFieldDefinitions.begin(),
+                   portFieldDefinitions.end(),
+                   std::inserter(model_.portFieldDefinitions_, model_.portFieldDefinitions_.end()),
+                   [](/*Non const to prevent copy*/ PortFieldDefinition& pfd)
+                   { return std::make_pair(pfd.getPort().Id(), std::move(pfd)); });
```
flomnes: The order of evaluation for arguments is unspecified, if the 2nd statement is executed before the 1st, then you might read invalid (moved) memory.

Example https://gist.github.com/flomnes/5b62b714e5f90ce625954a32f5e3d310

See https://stackoverflow.com/questions/2934904/what-is-the-order-of-evaluation-for-function-arguments

src/tests/src/io/yml-importers/test_full.cpp
```diff
@@ -117,16 +117,16 @@ BOOST_AUTO_TEST_CASE(test_full)
         - port: injection_port
           field: flow
           definition: generation
+      binding-constraints:
+        - id: balance
+          expression: injection_port.flow = 0
```
flomnes, line 122: This expression is probably invalid, try
```cpp
sum_connections(injection_port.flow) = 0
```

payetvin: sum_connection isn't handled by the convertorVisitor

src/study/system-model/model.cpp
JasonMarechal25:
```suggestion
                   {
                       auto id = pfd.getPort().Id();
                       return std::make_pair(id, std::move(pfd)); 
                   });
```

payetvin: done

flomnes (approved):

PR #2663, opened 2025-02-25
cleanup cucumber test outputs

pet-mit:


src/tests/cucumber/features/environment.py
```diff
@@ -1,4 +1,5 @@
 # This file defines behave hooks to run before all tests
+import shutil
```
guilpier-code: Should not we have : 
```py
from shutil import rmtree
```
?

src/tests/cucumber/features/environment.py
```diff
@@ -8,3 +9,8 @@ def check_userdata_exists(context, data_name: str):
     if data_name not in context.config.userdata:
         raise Exception(
             f"The following parameter should be defined in 'behave.ini' or in the command line when calling behave: {data_name}")
+
+def after_scenario(context, scenario):
```
guilpier-code, line 14: Should we go a bit further and remove content of **my_study/logs** ?

guilpier-code (changes requested):

src/tests/cucumber/features/environment.py
```diff
@@ -1,4 +1,7 @@
 # This file defines behave hooks to run before all tests
+import os.path
+import pathlib
```
guilpier-code:
```py
import pathlib
```
is it useful ?

guilpier-code (changes requested):

guilpier-code (approved):

src/tests/cucumber/features/environment.py
```diff
@@ -8,3 +10,10 @@ def check_userdata_exists(context, data_name: str):
     if data_name not in context.config.userdata:
         raise Exception(
             f"The following parameter should be defined in 'behave.ini' or in the command line when calling behave: {data_name}")
+
+def after_scenario(context, scenario):
+    # post-processing a test: clean up output files to avoid taking up all the disk space
+    if context.output_path != None:
```
a-zakir, line 16:
```suggestion
    if context.return_code == 0 and context.output_path != None :
```

a-zakir: because if the simu fails, there will not  be any output

a-zakir:
```suggestion
    if hasattr(context, "output_path") and  context.output_path != None:
```

a-zakir: because if the **modeler** simu fails, there will not  be any output

PR #2666, opened 2025-02-26
[ANT-2890] Fix segfault loading scenario

flomnes:
The exact cause has not been exactly found but we found what could cause it and how to fix it.

The error happened in destructor of yuni strings in rules. By preventing string copy we "moved" the issue in BindingConstraintGroupRepository::operator[].

Valgrind reported invalid write in _BindingConstraintGroupRepository::operator[]_ and _Antares::Data::ScenarioBuilder::BindingConstraintsTSNumberData::reset_

The suspicion is an invalid usage of unique_ptr in lambdas.

## Fix
Instead of exposing the container of unique_ptr to consultation through iterators, we maintain a second vector of raw pointers. One vector is for ownership, one for data consultation.

### Group caching
Add a map to cache group when searching by name. This map may be redundant with the new vector of raw pointer. We could find some time to analyze performance (time and memory) difference between using the two structures or only one map.
### Pass by reference in Rules.cpp
A lot of strings were passed by copy. They are now passed by reference. Ideally with std::string instead of Yuni::string we could passe std::string_view if applicable.
### Add constness
Improve constness locally

### VCPKG management

Update vcpkg configuration to properly use manifest and configuration


flomnes:
See #2668

PR #2669, opened 2025-02-28
Improve modeler API & main class

pet-mit:
- Add a function to read data-series in loadModelerFiles lib
- Move modeler dir to src/
- Remove "problem.lp" cucumber output file that shouldn't have been pushed
- rename some stuff for harmony


src/modeler/loadFiles/CMakeLists.txt
```diff
@@ -21,12 +22,15 @@ target_include_directories(loadModelerFiles
...
         Antares::antares-study-system-model
+        Antares::linear-problem-api
         PRIVATE
         Antares::io
         Antares::yml-system
         Antares::yml-model
         Antares::model-converter
         Antares::modelerParameters
+        Antares::data-series-csv-importer
+        linear-problem-data-impl
```
JasonMarechal25: linear-problem-data-impl probably need an alias and be refered here as Antares::linear-problem-data-impl to be consistent with other libs

src/modeler/loadFiles/readDataSeries.cpp
```diff
@@ -0,0 +1,49 @@
...
+  const std::filesystem::path& studyPath)
+{
+    try
+    {
+        Optimisation::LinearProblemDataImpl::DataSeriesRepository dataSeriesRepository = IO::
+          Inputs::DataSeriesCsvImporter::DataSeriesRepoImporter::importFromDirectory(
+            studyPath / "input" / "data-series",
+            "\t");
+        logs.info() << "Data-series loaded";
+        return std::make_unique<Optimisation::LinearProblemDataImpl::LinearProblemData>(
```
JasonMarechal25, line 39: Here we pass the local dataSeriesRepository. Linear problem data constructor accept a non-const reference and move the value.
In LinearProblemData we should either accept a value as argument and move it, or accept an rvalue ref

src/modeler/main.cpp
```diff
@@ -64,13 +61,13 @@ class SystemLinearProblem
...
         LinearProblemBuilder linear_problem_builder(fillers_ptr);
-        LinearProblemData data(dataSeriesRepo);
         FillContext dummy_time_scenario_ctx = {parameters.firstTimeStep, parameters.lastTimeStep};
-        linear_problem_builder.build(pb, data, dummy_time_scenario_ctx);
+        linear_problem_builder.build(pb, dataSeries_, dummy_time_scenario_ctx);
     }
 
 private:
     const Study::SystemModel::System& system_;
+    ILinearProblemData& dataSeries_;
```
JasonMarechal25: Do we really need a non const reference ?
Can't we do with a copy or a const reference ?
Since data series is only used in Provide, maybe it make more sens to move dataSeries as a parameter of build.
SystemLinearProblem is not much of a system rather than a SystemLinearProblemBuilder

src/modeler/main.cpp
```diff
@@ -35,21 +33,20 @@ using namespace Antares::Optimisation::LinearProblemMpsolverImpl;
 using namespace Antares;
 using namespace Antares::Solver;
 using namespace Antares::Optimisation::LinearProblemApi;
-using namespace Antares::Optimisation::LinearProblemDataImpl;
 
 class SystemLinearProblem
```
JasonMarechal25: Another refactoring to do later would be to "empty" main() and move code in library. It would be more testable and usable in other components

JasonMarechal25 (changes requested):

src/modeler/main.cpp
pet-mit: I agree, it would be nice to have one struct/class with all modeler inputs ([example of code duplication](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2671/files#diff-e459b7fc3f521838da6235795415d8a558c0c49f0f11c26e5d5f8976fcc4bc9a) that would be avoided)

pet-mit: what do you suggest we do with the parameters, that are mandatory for modeler studies but that will be ignored in hybrid studies ?

src/modeler/loadFiles/readDataSeries.cpp
pet-mit: I think that the idea behind LinearProblemData (we should probably rename it btw) is that it can/should be unique. So I'll make it accept the repo value.

src/optimisation/linear-problem-data-impl/CMakeLists.txt
```diff
@@ -17,15 +19,17 @@ set(SRC_DATA_SERIES
         linearProblemData.cpp
 )
 
-add_library(linear-problem-data-impl ${SRC_DATA_SERIES})
-set_target_properties(linear-problem-data-impl PROPERTIES LINKER_LANGUAGE CXX)
+add_library(${PROJ} ${SRC_DATA_SERIES})
+add_library(Antares::${PROJ} ALIAS ${PROJ})
+
+set_target_properties(${PROJ} PROPERTIES LINKER_LANGUAGE CXX)
```
flomnes:
```suggestion
```

flomnes (approved):

JasonMarechal25 (approved):

PR #2673, opened 2025-03-03
Run-study-tests: skip studies executed in cucumber

pet-mit:
This allows to migrate non-regression tests to cucumber. Every test that is mgrated is deactivated automatically in run-study-tests


pet-mit:
included in #2682

PR #2674, opened 2025-03-03
[ANT-2890] Check scenario year for Binding constraints

payetvin:


src/libs/antares/study/scenario-builder/BindingConstraintsTSNumbersData.cpp
```diff
@@ -113,7 +113,10 @@ void BindingConstraintsTSNumberData::setTSnumber(const std::string& group_name,
                                                  uint value)
 {
     auto& group_ts_numbers = rules_[group_name];
-    group_ts_numbers[0][year] = value;
+    if (year < group_ts_numbers.width)
```
flomnes:
```suggestion
    if (year < group_ts_numbers.height)
```

flomnes (changes requested):

flomnes (approved):

PR #2690, opened 2025-03-07
[ANT-2140] MustRun clusters taken into account in BC

JasonMarechal25:
## MustRun clusters contribute to BC RHS

Add mustrun clusters RHS to BC RHS

## Move prepareBindingConstraint

Move prepareBindingConstraint in dedicated file to reduce sim_calcul_economique.cpp size


src/solver/simulation/sim_binding_constraints.cpp
```diff
@@ -0,0 +1,193 @@
...
+                              int PasDeTempsDebut,
+                              const Data::BindingConstraintsRepository& bindingConstraints,
+                              const Data::BindingConstraintGroupRepository& bcgroups,
+                              const uint weekFirstDay,
+                              int pasDeTemps)
+{
+    const auto activeConstraints = bindingConstraints.activeConstraints();
+    const auto constraintCount = activeConstraints.size();
+
+    for (unsigned constraintIndex = 0; constraintIndex != constraintCount; ++constraintIndex)
```
guilpier-code: **Design - not exactly in the scope of this PR  - cosmetic** 
It seems that active constraints are fetched and gathered when we call **activeConstraints()** only.
So I'd like to suggest to give each active constraint an index internally to **activeConstraints()** (each active constraint knows its own index), but I'm not sure it's suitable or worth it.
If it turns out it is suitable, it would certainly simplify the caller's code a bit, by hiding details (no more index computation and no more vector size) : 
```c++
void prepareBindingConstraint(...)
{
    const auto activeConstraints = bindingConstraints.activeConstraints();
    for (const auto& bc :  activeConstraints)
    {
        ...
        problem.MatriceDesContraintesCouplantes[bc->index()]...
       ...
    }
}
```
Up to you.

src/solver/simulation/sim_binding_constraints.cpp
```diff
@@ -0,0 +1,193 @@
...
+                   && "Invalid constraint data height");
+
+            std::vector<double>& sndMember = problem
+                                               .MatriceDesContraintesCouplantes[constraintIndex]
+                                               .SecondMembreDeLaContrainteCouplante;
+
+            for (unsigned day = 0; day != 7; ++day)
+            {
+                auto dailyClusterMustRunProduction = getClusterDailyProduction(
+
```
guilpier-code: There's an empty line here, could be removed.

src/solver/simulation/sim_binding_constraints.cpp
```diff
@@ -0,0 +1,193 @@
...
+            auto hourly_mustrun_production = getClustersHourlyProduction(
+              PasDeTempsDebut,
+              pasDeTemps,
+              clusterMustRunTimeSeriesAndWeight);
+            problem.MatriceDesContraintesCouplantes[constraintIndex]
+              .SecondMembreDeLaContrainteCouplante[pasDeTemps]
+              = column[PasDeTempsDebut + pasDeTemps] - hourly_mustrun_production;
+            break;
+        }
+        case Data::BindingConstraint::typeDaily:
```
guilpier-code: **Not exactly in the scope of the PR, but still uncomfortable** :

When calling the current function, we do it for a given time step.
In other words, we call it 168 times a week.
It means that, for daily constraints, we make 24 times the same operation of filling the RHS. 
And for weekly constraints, 168 times.

If I'm right, a possible solution would be to switch loop over time and loop over constraints in **sim_calcul_economique.cpp**, of course.

src/libs/antares/study/include/antares/study/binding_constraint/BindingConstraint.h
```diff
@@ -373,6 +373,8 @@ class BindingConstraint final: public Yuni::NonCopyable<BindingConstraint>
     template<class Env>
     std::string timeSeriesFileName(const Env& env) const;
 
+    const clusterWeightMap& clusters() const;
```
guilpier-code: **_This has to be discussed : it seems difficult to apply required changes._**

`const clusterWeightMap& clusters() const;` : 
The name of function is quite misleading, because what this function returns is not clusters, but a **std::map** associating clusters to their weight.
Actually, what clients of this function really need is a only clusters (and not an `std::map`), so the name of the function should remain.
So why not make **clusters()** return a vector of clusters (elegantly extracted from **pClusterWeights**)  ?

A consequence is that the client code would not have to get a **std::map**  and filter only the keys (that is the clusters)

src/solver/simulation/sim_binding_constraints.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+#include "antares/solver/simulation/sim_binding_constraints.h"
+
+#include <numeric>
+#include <span>
+#include <vector>
+
+#include "antares/solver/simulation/sim_structure_probleme_economique.h"
+#include "antares/study/binding_constraint/BindingConstraint.h"
+#include "antares/study/binding_constraint/BindingConstraintGroup.h"
```
guilpier-code: It seem we don't need 
`#include "antares/study/binding_constraint/BindingConstraintGroup.h"`

src/solver/simulation/sim_binding_constraints.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "antares/study/binding_constraint/BindingConstraintGroupRepository.h"
+#include "antares/study/binding_constraint/BindingConstraintsRepository.h"
+
+namespace
+{
+using Time_serie = std::span<const double>;
```
guilpier-code: `using Time_serie = std::span<const double>;` : 
for types, we usually use **ThisStyle**

src/solver/simulation/sim_binding_constraints.cpp
JasonMarechal25: We need it for 
        tsIndexForBc = group->timeseriesNumbers[year];

src/libs/antares/study/include/antares/study/binding_constraint/BindingConstraint.h
JasonMarechal25: I renamed it clustersAndWeights
We need the weight as well as the respective clusters

src/solver/simulation/sim_binding_constraints.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/2798

src/solver/simulation/sim_binding_constraints.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/2797

flomnes (approved):

PR #2692, opened 2025-03-07
Move deps to vcpkg

JasonMarechal25:
Use vcpkg to provide libuuid


JasonMarechal25:
I have issues locally. Mainly yuni doesn't find uuid. Closing for now

PR #2693, opened 2025-03-07
Enable extra linear solvers: PDLP & HiGHS

pet-mit:
- Enable usage of PDLP & HiGHS (needs latest release of or-tools-rte) in linear optimization
- Add some unit tests
- Update doc


JasonMarechal25 (approved):

src/solver/utils/ortools_utils.cpp
```diff
@@ -356,30 +358,29 @@ std::string availableOrToolsSolversString()
...
+        return names.LPSolverName.empty() ? std::optional<std::string>() : names.LPSolverName;
     }
 }
 
 MPSolver* MPSolverFactory(const bool isMip, const std::string& solverName)
 {
-    const std::string notFound = "Solver " + solverName + " not found";
+    const std::string notFound = "Solver " + solverName + " was not found or does not support "
+                                 + (isMip ? "MIP" : "LP") + " problems";
     const std::invalid_argument except(notFound);
```
flomnes: Move in the `if(!internalSolverName.has_value())` block, since this should not happen very often there is no need to construct objects "just in case something goes wrong"

src/tests/src/solver/utils/test_ortools_utils.cpp
```diff
@@ -0,0 +1,98 @@
...
+               SolverTestData{"xpress", true, SolverType::XPRESS_MIXED_INTEGER_PROGRAMMING},
+               SolverTestData{"coin", false, SolverType::CLP_LINEAR_PROGRAMMING},
+               SolverTestData{"coin", true, SolverType::CBC_MIXED_INTEGER_PROGRAMMING},
+               SolverTestData{"highs", false, SolverType::HIGHS_LINEAR_PROGRAMMING},
+               SolverTestData{"highs", true, SolverType::HIGHS_MIXED_INTEGER_PROGRAMMING},
+               SolverTestData{"glpk", false, SolverType::GLPK_LINEAR_PROGRAMMING},
+               SolverTestData{"glpk", true, SolverType::GLPK_MIXED_INTEGER_PROGRAMMING}}))
+{
+    auto solver = MPSolverFactory(sample.is_mip, sample.solver_name);
+    BOOST_CHECK(solver);
```
flomnes:
```suggestion
    BOOST_REQUIRE(solver);
```

If `solver==nullptr`, the following line will read invalid memory. `BOOST_REQUIRE` exits the current test, preventing invalid reads (in this case).

src/tests/src/solver/utils/test_ortools_utils.cpp
```diff
@@ -0,0 +1,98 @@
...
+    BOOST_CHECK_EQUAL(sample.expected_type, solver->ProblemType());
+}
+
+// PDLP only supports LP
+BOOST_AUTO_TEST_CASE(test_pdlp_support)
+{
+    auto lpSolver = MPSolverFactory(false, "pdlp");
+    BOOST_CHECK(lpSolver);
```
flomnes:
```suggestion
    BOOST_REQUIRE(lpSolver);
```

src/tests/src/solver/utils/test_ortools_utils.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+// SCIP only supports MIP
+BOOST_AUTO_TEST_CASE(test_scip_support)
+{
+    BOOST_CHECK_EXCEPTION(MPSolverFactory(false, "scip"),
+                          std::invalid_argument,
+                          checkMessage(
+                            "Solver scip was not found or does not support LP problems"));
+    auto mipSolver = MPSolverFactory(true, "scip");
+    BOOST_CHECK(mipSolver);
```
flomnes:
```suggestion
    BOOST_REQUIRE(mipSolver);
```

src/tests/src/solver/utils/test_ortools_utils.cpp
```diff
@@ -0,0 +1,98 @@
...
+    BOOST_CHECK_EQUAL(data.expected_type, solver->ProblemType());
+}
+
+BOOST_AUTO_TEST_SUITE(_ortools_utils_)
+
+// The following solvers support LP & MIP
+BOOST_DATA_TEST_CASE(
+  test_lp_and_mip_solvers_support,
+  bdata::make({SolverTestData{"sirius", false, SolverType::SIRIUS_LINEAR_PROGRAMMING},
+               SolverTestData{"sirius", true, SolverType::SIRIUS_MIXED_INTEGER_PROGRAMMING},
```
flomnes: Out of scope, but Sirius doesn't really support MILP problems (only binary variables)

In my opinion `SIRIUS_MIXED_INTEGER_PROGRAMMING` should not be used in Antares.

src/solver/utils/ortools_utils.cpp
```diff
@@ -322,7 +322,9 @@ const std::map<std::string, struct OrtoolsUtils::SolverNames> OrtoolsUtils::solv
   {"sirius", {"sirius_lp", "sirius"}},
   {"coin", {"clp", "cbc"}},
   {"glpk", {"glpk_lp", "glpk"}},
-  {"scip", {"scip", "scip"}}};
+  {"scip", {"", "scip"}}, // SCIP only supports MIPs
+  {"highs", {"highs_lp", "highs"}},
+  {"pdlp", {"pdlp", ""}}}; // PDLP only supports LPs
```
flomnes: Consider using `std::optional<std::string>` instead of empty strings, it may simplify e.g function `translateSolverName`

src/tests/src/solver/utils/test_ortools_utils.cpp
pet-mit: do you think we can safely prohibit it from being used in MIP mode?

src/solver/utils/ortools_utils.cpp
```diff
@@ -356,38 +365,40 @@ std::string availableOrToolsSolversString()
...
+    {
+        return names.LPSolverName;
     }
 }
 
 MPSolver* MPSolverFactory(const bool isMip, const std::string& solverName)
 {
-    const std::string notFound = "Solver " + solverName + " not found";
+    const std::string notFound = "Solver " + solverName + " was not found or does not support "
+                                 + (isMip ? "MIP" : "LP") + " problems";
```
```suggestion
```
This can now be removed (duplicate)

src/tests/src/solver/utils/test_ortools_utils.cpp
flomnes: It's already not enabled (see `OrtoolsUtils::solverMap`), so you could just remove it from the tests since (non-existent possibility).

src/solver/utils/ortools_utils.cpp
pet-mit: oops you're right

PR #2697, opened 2025-03-11
Solver options : when MILP, forbid to use options for optim 1 or 2

guilpier-code:


flomnes (changes requested):
The problem is that until the problem has been constructed we can't know if it will be solved using a MILP solver or an LP solver. This will become even more true when the hybrid modeler comes into place.

This check has to be done way later

guilpier-code:
> The problem is that until the problem has been constructed we can't know if it will be solved using a MILP solver or an LP solver. This will become even more true when the hybrid modeler comes into place.
> 
> This check has to be done way later

From what I see, user can set the parameter **unit-commitment-mode** to **fast**, **accurate** or **milp** in **generaldat.ini**.
I didn't know that resolution mode could be also set by the problem at execution. What do I miss ?

flomnes:
> > The problem is that until the problem has been constructed we can't know if it will be solved using a MILP solver or an LP solver. This will become even more true when the hybrid modeler comes into place.
> > This check has to be done way later
> 
> From what I see, user can set the parameter **unit-commitment-mode** to **fast**, **accurate** or **milp** in **generaldat.ini**. I didn't know that resolution mode could be also set by the problem at execution. What do I miss ?

In the future integer variables may arise from different sources (e.g hybrid studies with the modeler)

guilpier-code:
> > > The problem is that until the problem has been constructed we can't know if it will be solved using a MILP solver or an LP solver. This will become even more true when the hybrid modeler comes into place.
> > > This check has to be done way later
> > 
> > 
> > From what I see, user can set the parameter **unit-commitment-mode** to **fast**, **accurate** or **milp** in **generaldat.ini**. I didn't know that resolution mode could be also set by the problem at execution. What do I miss ?
> 
> In the future integer variables may arise from different sources (e.g hybrid studies with the modeler)

ok, but not yet. We'll change this when time comes.

PR #2703, opened 2025-03-14
2.9 TimeSum and AllTimeSum operators [ANT-2932][ANT-2933]

a-zakir:


src/tests/src/expressions/test_AstDOTStyleVisitor.cpp
```diff
@@ -75,15 +75,33 @@ class Fixture
     {
```
guilpier-code, line 73:
```c++
#include <fstream>
```
unused

src/tests/src/expressions/test_AstDOTStyleVisitor.cpp
```diff
@@ -75,15 +75,33 @@ class Fixture
...
         Node* timeShiftNode = registry_.create<TimeShiftNode>(parameterNode, literalNode2);
         Node* literalNode = registry_.create<LiteralNode>(6.62607015);
         Node* parm2 = registry_.create<ParameterNode>("par");
         Node* timeIndexNode = registry_.create<TimeIndexNode>(
           literalNode,
-          registry_.create<SumNode>(parm2, literalNode2));
+          registry_.create<SumNode>(parm2, literalNode3));
 
-        return registry_.create<SumNode>(gt, timeIndexNode, timeShiftNode);
+        Node* parameterNode2 = registry_.create<ParameterNode>("div");
```
guilpier-code, line 85: These AST **trees** are already large, and we make them larger.
If we want to tests new nodes as **AllTimeSumNode**, **TimeSumNode** or **SumNode**, we should make much more smaller tests, but more numerous : one for each kind of node.

These small tests would be more readable, and if one fails, we know quickly why.
Besides, with small tests, you can test several aspects of a class / object (in big tests, you can't, unless you make them bigger).
After many small tests, make more complicated tests if you like, in order to tests if all nodes work together.

Something else : suppose I'm new to **AST** nodes, or I need to remember about them. I want to know what is a **AllTimeSumNode**, how it works and what I should expect of it.
Well, in these big tests, it's difficult.

src/tests/src/expressions/test_CloneVisitor.cpp
```diff
@@ -100,6 +100,46 @@ BOOST_FIXTURE_TEST_CASE(clone_TimeIndexNode, Registry<Node>)
     BOOST_CHECK(rightChild->value() == "yolo");
```
guilpier-code, line 100: Test name : **clone_TimeShitNode** : to be renamed into "clone_TimeShi**F**tNode" !
I don't know what a time shit is :-)

src/tests/src/expressions/test_CloneVisitor.cpp
```diff
@@ -100,6 +100,46 @@ BOOST_FIXTURE_TEST_CASE(clone_TimeIndexNode, Registry<Node>)
...
 }
```
guilpier-code, line 101: In test **clone_TimeShitNode**, instruction : 
`BOOST_REQUIRE(cloneTimeShift);` 
should come before 
```c++
BOOST_CHECK(dynamic_cast<LiteralNode*>(cloneTimeShift->right())->value() == -546);
```

src/tests/src/expressions/test_TimeIndexVisitor.cpp
```diff
@@ -80,6 +80,109 @@ BOOST_FIXTURE_TEST_CASE(simple_time_dependant_expression, Registry<Node>)
     BOOST_CHECK_EQUAL(timeIndexVisitor.dispatch(expr), TimeIndex::VARYING_IN_TIME_AND_SCENARIO);
 }
 
+BOOST_FIXTURE_TEST_CASE(timeShift_expression, Registry<Node>)
+{
+    PrintVisitor printVisitor;
```
guilpier-code:
```c++
PrintVisitor printVisitor;
```
In many tests here, we instantiate a **PrintVisitor** and don't use it.

src/tests/src/expressions/test_TimeIndexVisitor.cpp
```diff
@@ -80,6 +80,109 @@ BOOST_FIXTURE_TEST_CASE(simple_time_dependant_expression, Registry<Node>)
...
+    // LiteralNode --> constant in time and for all scenarios
+    LiteralNode literalNode(65.);
```
guilpier-code: Duplication : 
- `TimeIndexVisitor timeIndexVisitor;` repeated in each test
- Same definition of `literalNode(65.)`, `parameterNode1 `and `variableNode1`, repeated in several tests.

We could have a fixture here.

src/tests/src/expressions/test_TimeIndexVisitor.cpp
```diff
@@ -80,6 +80,109 @@ BOOST_FIXTURE_TEST_CASE(simple_time_dependant_expression, Registry<Node>)
...
+
+    // Parameter --> constant in time and varying scenarios
+    ParameterNode parameterNode1("p1", TimeIndex::VARYING_IN_SCENARIO_ONLY);
+
+    // Variable time varying but constant across scenarios
+    VariableNode variableNode1("v1", TimeIndex::VARYING_IN_TIME_ONLY);
```
guilpier-code: Why adding a **1** at the end of **variableNode1** if we don't have a **variableNode2** ?
Same for **parameterNode1**.

src/tests/src/expressions/test_TimeIndexVisitor.cpp
guilpier-code: **_If time allows it_** 
In many tests here, we actually have 3 unit tests.
As you may know, small unit tests are strongly recommended : for clarity, for documenting a feature, and if a unit test fails, it's quicker to see what's wrong.

Problem is : too big unit tests usually have unclear names (not specific enough).
Splitting tests would allow improving their names, because then, these names become more specific. 

Moreover, using a fixture would allow to shorten the tests,

src/expressions/visitors/PrintVisitor.cpp
```diff
@@ -139,6 +139,17 @@ std::string PrintVisitor::visit(const Nodes::TimeIndexNode* node)
     return dispatch(node->left()) + "[ " + dispatch(node->right()) + " ]";
 }
```
guilpier-code, line 141: Inside `trimAndFormat(const std::string& in)` : 
we could use `boost::algorithm::trim(...)` or `boost::algorithm::trim_left(...)` instead of 3 lines.

src/expressions/visitors/TimeIndexVisitor.cpp
```diff
@@ -121,6 +121,17 @@ TimeIndex TimeIndexVisitor::visit(const Nodes::TimeIndexNode* timeIndexNode)
     return TimeIndex::CONSTANT_IN_TIME_AND_SCENARIO;
```
guilpier-code, line 121: **Question** : 
About : 
```c++
TimeIndex TimeIndexVisitor::visit(const Nodes::TimeShiftNode* timeShiftNode)
{
    return dispatch(timeShiftNode->left());
}
```
This means that, on expression **[1 - t]**, the **TimeIndexVisitor** returns **CONSTANT_IN_TIME_AND_SCENARIO**.
Is this expected ?

src/expressions/visitors/CompareVisitor.cpp
```diff
@@ -22,15 +22,24 @@
 #include <antares/expressions/visitors/CompareVisitor.h>
 
 template<class T, class V>
-static bool compareBinaryNode(V& visitor,
+static bool compareParentNode(V& visitor,
                               const T* node,
                               const Antares::Expressions::Nodes::Node* other)
 {
-    if (const T* other_node = dynamic_cast<const T*>(other))
+    if (const auto* other_node = dynamic_cast<const T*>(other))
```
flomnes: Why not `T*` ? It's more clear & shorter than `auto*`

flomnes (commented):
LGTM

Please move test data to [Antares_Simulator_Tests_NR](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR), see #2716

src/tests/src/expressions/test_TimeIndexVisitor.cpp
```diff
@@ -80,6 +80,109 @@ BOOST_FIXTURE_TEST_CASE(simple_time_dependant_expression, Registry<Node>)
...
+    LiteralNode from(65.);
+
+    // Parameter --> constant in time and varying scenarios
+    ParameterNode to("p1", TimeIndex::VARYING_IN_SCENARIO_ONLY);
+
+    // Variable time varying but constant across scenarios
+    VariableNode variableNode1("v1", TimeIndex::VARYING_IN_TIME_ONLY);
+    TimeIndexVisitor timeIndexVisitor;
+
+    TimeSumNode t1(&from, &to, &from);
```
guilpier-code:
```c++
TimeSumNode t1(&from, &to, &from);
```
is confusing : a time sum is supposed to have a start range (the **from**), an end range (the **to**) and have quantity to sum. 
Here (if I understand correctly), **from** is also the quantity to be summed. 
What does it mean ?
(I guess we reuse an object that happens to have the **TimeIndex** we're looking for )

src/tests/src/expressions/test_TimeIndexVisitor.cpp
```diff
@@ -80,6 +80,109 @@ BOOST_FIXTURE_TEST_CASE(simple_time_dependant_expression, Registry<Node>)
...
+
+    BOOST_CHECK_EQUAL(timeIndexVisitor.dispatch(&t1), TimeIndex::CONSTANT_IN_TIME_AND_SCENARIO);
+
+    TimeSumNode t2(&from, &to, &to);
```
guilpier-code: **General remark** : 
For reading confort, each time we have a composed **Node**, we should write, **as a comment** in the test code, what it represents in a **yaml** file.

For example : 
```c++
LiteralNode literalNode(65.);
ParameterNode parameterNode1("p1", TimeIndex::VARYING_IN_SCENARIO_ONLY);

// Represents : [65 + p1] 
TimeShiftNode constant_time_shift_node(&literalNode, &parameterNode1);
```
Indeed, it's not obvious for any eyes.

src/tests/src/expressions/test_TimeIndexVisitor.cpp
```diff
@@ -80,6 +80,109 @@ BOOST_FIXTURE_TEST_CASE(simple_time_dependant_expression, Registry<Node>)
...
+    BOOST_CHECK_EQUAL(timeIndexVisitor.dispatch(&t3), TimeIndex::CONSTANT_IN_TIME_AND_SCENARIO);
+}
+
+BOOST_FIXTURE_TEST_CASE(timeSumNode_expression, Registry<Node>)
+{
+    PrintVisitor printVisitor;
+    // LiteralNode --> constant in time and for all scenarios
+    LiteralNode from(65.);
+
+    // Parameter --> constant in time and varying scenarios
```
guilpier-code:
```c++
In lines like : 
// Parameter --> constant in time and varying scenarios
ParameterNode to("p1", TimeIndex::VARYING_IN_SCENARIO_ONLY);
```
not sure the comment is useful, as the following code line is clear enough.

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ReadLinearConstraintVisitor.h
```diff
@@ -86,5 +86,7 @@ class ReadLinearConstraintVisitor
       const Expressions::Nodes::ComponentParameterNode* node) override;
     std::vector<LinearConstraint> visit(const Expressions::Nodes::TimeShiftNode* node) override;
```
guilpier-code, line 87: Out of this PR's scope : 
A **ReadLinearConstraintVisitor** has a name. What for ? (it does not seem useful at first sight)

src/tests/src/expressions/test_LinearVisitor.cpp
```diff
@@ -327,6 +327,49 @@ BOOST_FIXTURE_TEST_CASE(CheckTimeShiftOnVariableLinearity, Registry<Node>)
                 == LinearStatus::LINEAR); // because literal1 is constant
 }
 
+BOOST_FIXTURE_TEST_CASE(CheckTimeSumOnLiteralLinearity, Registry<Node>)
+{
+    LinearityVisitor linearityVisitor;
+    CompareVisitor compareVisitor;
+    Node* literal1 = create<LiteralNode>(1.);
```
guilpier-code: `Node* literal1` : why give a number 1 to a literal if there's no second one ?

src/tests/src/expressions/test_LinearVisitor.cpp
```diff
@@ -327,6 +327,49 @@ BOOST_FIXTURE_TEST_CASE(CheckTimeShiftOnVariableLinearity, Registry<Node>)
                 == LinearStatus::LINEAR); // because literal1 is constant
 }
 
+BOOST_FIXTURE_TEST_CASE(CheckTimeSumOnLiteralLinearity, Registry<Node>)
+{
+    LinearityVisitor linearityVisitor;
+    CompareVisitor compareVisitor;
```
guilpier-code:
```c++
CompareVisitor compareVisitor;
```
unused, in several unit tests

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
```diff
@@ -610,13 +662,69 @@ BOOST_FIXTURE_TEST_CASE(evaluate_shifted_param, MyDummyFixture)
...
+    unsigned first = 0;
+    unsigned last = 2;
+    EvalVisitor evalVisitor(context, {first, last /*three hours*/});
+    return evalVisitor.dispatch(&root);
+}
+
+BOOST_FIXTURE_TEST_CASE(evaluate_alltimeSum_param, MyDummyFixture)
+{
+    LiteralNode from(0.0);
+    LiteralNode to(1.0);
```
guilpier-code, line 721:
```c++ 
LiteralNode from(0.0);
LiteralNode to(1.0);
```
are unused in this test.

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
```diff
@@ -610,13 +662,69 @@ BOOST_FIXTURE_TEST_CASE(evaluate_shifted_param, MyDummyFixture)
...
 }
 
+EvaluationResult CreateAndEvaluateTimeSumNode(Node* from, Node* to)
+{
+    ParameterNode param("my-param", TimeIndex::VARYING_IN_TIME_ONLY);
+    TimeSumNode root(from, to, &param);
+    const std::string value = "dummy";
+    MockLinearProblemData dummy_data;
+    EvaluationContext context(
+      {build_context_parameter_with("my-param", value, ParameterType::TIMESERIE)},
```
guilpier-code, line 679: In : 
```c++
build_context_parameter_with("my-param", value, ParameterType::TIMESERIE)
```
shouldn't the first argument be `param.value()` ?

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
```diff
@@ -610,13 +662,69 @@ BOOST_FIXTURE_TEST_CASE(evaluate_shifted_param, MyDummyFixture)
...
     LiteralNode literal_node(1.0);
-    const auto eval = CreateAndEvalutateTimeNode<TimeIndexNode, Node*>(&literal_node)
+    const auto eval = CreateAndEvaluateTimeNode<TimeIndexNode, Node*>(&literal_node)
                         .valueAsDouble();
     // from MockLinearProblemData  param TSdata is {0, 1, 2}
     // here we applied TimeIndex[1]
     BOOST_CHECK_EQUAL(eval, 1); //
 }
 
+EvaluationResult CreateAndEvaluateTimeSumNode(Node* from, Node* to)
```
guilpier-code, line 672: Functions **CreateAndEvaluate...(...)** : 
- they make several things (there is **And** in their name). So we could (at least) split them into 2 parts.
- They hide (in their body) some details that should appear clearly in the tests, because tests checks depend on this these details : we'd like to know : 
  - what quantity we sum (here a `ParameterNode param`)
  - how many time step we have (here we have 3 : {0, 1, 2})
  - which visitor we use in the test
- They are called once, so their purpose is not to be reused.
- They contain lines of code that can be found in each of them. So it would be worth make functions with these lines of code.

```c++
BOOST_FIXTURE_TEST_CASE(evaluate_timeSum_param_improve_trial, MyDummyFixture)
{
    // Arrange step
    LiteralNode from(0.0);
    LiteralNode to(1.0);
    ParameterNode param("my-param", TimeIndex::VARYING_IN_TIME_ONLY);
    TimeSumNode timeSumNode(&from, &to, &param);

    unsigned firstTimeStep = 0;
    unsigned lastTimeStep = 2;

    EvaluationContext context = BuildContextForParam(param); // Reusable
    EvalVisitor evalVisitor(context, {firstTimeStep, lastTimeStep});

    // Act step
    const auto result = evalVisitor.dispatch(&timeSumNode).valuesAsVector();

    // Assert step
    BOOST_CHECK_EQUAL(result.at(0), 0 + 1);
    BOOST_CHECK_EQUAL(result.at(1), 1 + 2); 
    BOOST_CHECK_EQUAL(result.at(2), 2 + 0);
}
```

src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -181,14 +199,34 @@ EvaluationResult::EvaluationResult(const std::variant<double, std::vector<double
...
-                 { return shift(l, timeShift); },
+      std::visit([&time_shift](const auto& l) -> std::variant<double, std::vector<double>>
+                 { return shift(l, time_shift); },
                  value_));
 }
 
+EvaluationResult EvaluationResult::timeSum(int from, int to) const
+{
+    EvaluationResult ret(0.);
+    for (auto shift = from; shift <= to; ++shift)
```
guilpier-code: `auto shift = from` : 
`int shift = from` would be (a bit) shorter and more clear.

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
```diff
@@ -336,10 +336,62 @@ BOOST_AUTO_TEST_CASE(ShiftResult_EmptyVector)
...
+
+BOOST_AUTO_TEST_CASE(TimeSum_VectorValue_PositiveShift)
+{
+    const EvaluationResult eval(std::vector<double>{1.0, 2.0, 3.0});
+    const auto sum = eval.timeSum(0, 2).valuesAsVector();
+    BOOST_CHECK_THROW(eval.timeSum(-1, 0).valueAsDouble(), EvaluationResult::EvalResultTypeError);
+    const std::vector<double> expected(3.0, 1.0 + 2.0 + 3.0);
+
+    BOOST_CHECK_EQUAL_COLLECTIONS(sum.begin(), sum.end(), expected.begin(), expected.end());
+}
```
guilpier-code, line 360: There are 2 tests here : 
```c++
BOOST_AUTO_TEST_CASE(TimeSum_VectorValue_PositiveShift)
{
    // Test 1
    const EvaluationResult eval(std::vector<double>{1.0, 2.0, 3.0});
    BOOST_CHECK_THROW(eval.timeSum(-1, 0).valueAsDouble(), EvaluationResult::EvalResultTypeError);
    
    // Test 2
    const auto sum = eval.timeSum(0, 2).valuesAsVector();
    const std::vector<double> expected(3.0, 1.0 + 2.0 + 3.0);
    BOOST_CHECK_EQUAL_COLLECTIONS(sum.begin(), sum.end(), expected.begin(), expected.end());
}
```
Having 2 tests would allow each of them to have a good name (describing what it does).

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
```diff
@@ -336,10 +336,62 @@ BOOST_AUTO_TEST_CASE(ShiftResult_EmptyVector)
...
+    const EvaluationResult sum = eval.timeSum(-2, 2);
+    BOOST_CHECK_EQUAL(sum.valueAsDouble(), 20.0); // 4.0 * 5 = 20.0
+
+    BOOST_CHECK_THROW(eval.timeSum(-1, 0).valuesAsVector(), EvaluationResult::EvalResultTypeError);
+}
+
+BOOST_AUTO_TEST_CASE(TimeSum_VectorValue_PositiveShift)
+{
+    const EvaluationResult eval(std::vector<double>{1.0, 2.0, 3.0});
+    const auto sum = eval.timeSum(0, 2).valuesAsVector();
```
guilpier-code, line 355: **sum** not needed on first check. 
Should be put down where it's needed, for clarity.

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
```diff
@@ -610,13 +662,69 @@ BOOST_FIXTURE_TEST_CASE(evaluate_shifted_param, MyDummyFixture)
...
+}
+
+BOOST_FIXTURE_TEST_CASE(evaluate_timeSum_param, MyDummyFixture)
+{
+    LiteralNode from(0.0);
+    LiteralNode to(1.0);
+    const auto eval = CreateAndEvaluateTimeSumNode(&from, &to).valuesAsVector();
+    // from MockLinearProblemData  param TSdata is {0, 1, 2}
+    // here we applied TimeSum from t+0 and to t+1
+    BOOST_CHECK_EQUAL(eval.at(0), 0 + 1); // add param.at(0)+param.at(1)
```
guilpier-code, line 696:
```c++
// add param.at(0)+param.at(1)
```
Shouldn't this comment be  : 
```c++
// = param[0]+param[1]
```
?

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
```diff
@@ -325,3 +325,38 @@ BOOST_AUTO_TEST_CASE(TimeShiftExpressionMul)
     Visitors::CompareVisitor cmp;
     BOOST_CHECK(cmp.dispatch(expr.node, timeShiftNode));
 }
+
+BOOST_AUTO_TEST_CASE(TimeSumExpression)
+{
+    Registry<Nodes::Node> registry;
```
guilpier-code, line 331: **registry** define in each test.
We should make a fixture, like it's done for other unit tests.

src/expressions/visitors/CompareVisitor.cpp
a-zakir: alright!

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
```diff
@@ -325,3 +325,38 @@ BOOST_AUTO_TEST_CASE(TimeShiftExpressionMul)
...
+
+    const auto [expression, div] = expected_expression(registry);
+    const auto expressionWithNumericalTimeIndex = "sum(t-89*param1 .. t+(param1/89)," + expression
+                                                  + ")";
+    auto expr = createMediumExpression().run(expressionWithNumericalTimeIndex);
```
guilpier-code, line 336:
```c++
auto expr = createMediumExpression().run(expressionWithNumericalTimeIndex);
```
is short but unclear.
It's difficult to understand what it does : we have to navigate in the code to get it.

What this instruction does is : 
```c++
auto model = createSomeModel();
auto expr = ModelConverter::convertExpressionToNode(exprWithTimeSum,
                                                    model);
```
More meaningful.

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
```diff
@@ -325,3 +325,38 @@ BOOST_AUTO_TEST_CASE(TimeShiftExpressionMul)
     Visitors::CompareVisitor cmp;
     BOOST_CHECK(cmp.dispatch(expr.node, timeShiftNode));
 }
+
+BOOST_AUTO_TEST_CASE(TimeSumExpression)
+{
+    Registry<Nodes::Node> registry;
+
+    const auto [expression, div] = expected_expression(registry);
+    const auto expressionWithNumericalTimeIndex = "sum(t-89*param1 .. t+(param1/89)," + expression
```
guilpier-code, line 334: **expressionWithNumericalTimeIndex** : is this an exact name for this expression ?
It seems this name is repeated from previous tests.

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
```diff
@@ -325,3 +325,38 @@ BOOST_AUTO_TEST_CASE(TimeShiftExpressionMul)
     Visitors::CompareVisitor cmp;
     BOOST_CHECK(cmp.dispatch(expr.node, timeShiftNode));
 }
+
+BOOST_AUTO_TEST_CASE(TimeSumExpression)
```
guilpier-code, line 329: Now it's done. 
But does this unit test have to be this complicated ?

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
guilpier-code, line 336: **expr** is an expected result
- it's name should say it.
- it could be defined next to the check (I mean : not in the middle of the test)

guilpier-code (changes requested):

src/tests/src/solver/optim-model-filler/test_linearExpression.cpp
```diff
@@ -21,6 +21,7 @@
 
 #define WIN32_LEAN_AND_MEAN
 
+#include <ranges>
```
guilpier-code, line 24: `#include <ranges>` seems unused

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -125,71 +125,11 @@ size_t TimeDependentLinearExpression::getSize() const
...
-    unsigned rangeSize = fillContext.getNumberOfTimestep();
-
-    // Normalize shift within bounds (to prevent negative indexing)
-    shift = (shift % static_cast<int>(rangeSize) + rangeSize) % static_cast<int>(rangeSize);
-
-    // Compute which key's value should be assigned to `key`
-    return fillContext.getFirstTimeStep()
-           + (key - fillContext.getFirstTimeStep() + shift) % rangeSize;
+    linearExpressions_ = add_maps(linearExpressions_, other.GetLinearExpressions());
+    return *this;
```
guilpier-code, line 132: _**Not strictly in this PR**_
question about block (inside **shiftLinearExpressions(...)**) : 
```c++
unsigned number_of_timestep = fillContext.getNumberOfTimestep();
if (number_of_timestep == 0)
{
     return TimeDependentLinearExpression(fillContext);
}
```
What if we removed this block ? Do we have a different behavior ? A crash ?

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
```diff
@@ -105,6 +105,48 @@ struct MockLinearProblemData: Antares::Optimisation::LinearProblemApi::ILinearPr
     }
 };
 
+BOOST_FIXTURE_TEST_CASE(visit_timeSum, MyDummyFixture)
+{ // param  ={0,1,2}
+    // 5 + sum( t-2 .. t+1, param) -->
+    // t = 0 : 5 +param.at(-2) + param.at(-1) + param.at(0) + param.at(1) --> 5 +1 + 2 + 0 + 1 = 9
+    // t = 1 : 5 +param.at(-1) + param.at(0) + param.at(1) + param.at(2) --> 5 + 2+0+1+2 = 10
```
guilpier-code, line 112: Nice comments, very welcome.
We could make them a bit shorter by replacing **param.at(-2)** by **param[-2]** (for instance)

src/expressions/visitors/TimeIndexVisitor.cpp
a-zakir: it returns the time Index of the expression

src/tests/src/expressions/test_TimeIndexVisitor.cpp
a-zakir: yes it is just to avoid creating another node, all from, to and expression are Nodes:
ex: to = ParamNode("Summer")+LiteralNode(21) // start 21 days after your summer beginnings

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ReadLinearConstraintVisitor.h
a-zakir: it transforms constraint Node into Solver Constraint,

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
a-zakir: both are wrong, i think // add param at  t = 0+param at t =1

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
```diff
@@ -105,6 +105,48 @@ struct MockLinearProblemData: Antares::Optimisation::LinearProblemApi::ILinearPr
...
+    BOOST_CHECK(linear_expressions.at(1).coefPerVar().empty());
+}
+
+BOOST_FIXTURE_TEST_CASE(visit_AllTimeSum, MyDummyFixture)
+{ // param  ={0,1,2}
+    // 5 + sum(param) -->
+    // 5 +param.at(0) + param.at(1) + param.at(2)  --> 5 + 0 + 1  + 2  = 8
+
+    Node* from = create<LiteralNode>(-2.);
+    Node* to = create<LiteralNode>(1.);
```
guilpier-code: Nodes **from** and **to** are not needed in this test.
As we read constraints containing only an **AllTimeSumNode**, these bounds make no sense.

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
```diff
@@ -105,6 +105,48 @@ struct MockLinearProblemData: Antares::Optimisation::LinearProblemApi::ILinearPr
...
+    EvaluationContext evaluation_context(
+      {{build_context_parameter_with("param", "something", ParameterType::TIMESERIE)}},
+      {},
+      my_data);
+    ReadLinearExpressionVisitor visitor(evaluation_context, {0, 2}, componentId);
+    auto linear_expressions = visitor.dispatch(sum).GetLinearExpressions();
+    BOOST_CHECK_EQUAL(linear_expressions.at(0).offset(), 9.);
+    BOOST_CHECK(linear_expressions.at(0).coefPerVar().empty());
+    BOOST_CHECK_EQUAL(linear_expressions.at(1).offset(), 10.);
+    BOOST_CHECK(linear_expressions.at(1).coefPerVar().empty());
```
guilpier-code, line 127: If we don't test the 3rd contraint, argument **{0, 2}** should be turned into **{0, 1}**

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
```diff
@@ -105,6 +105,48 @@ struct MockLinearProblemData: Antares::Optimisation::LinearProblemApi::ILinearPr
...
+    Node* to = create<LiteralNode>(1.);
+    Node* sum = create<SumNode>(create<LiteralNode>(5.),
+                                create<TimeSumNode>(from, to, create<ParameterNode>("param")));
+    MockLinearProblemData my_data;
+    EvaluationContext evaluation_context(
+      {{build_context_parameter_with("param", "something", ParameterType::TIMESERIE)}},
+      {},
+      my_data);
+    ReadLinearExpressionVisitor visitor(evaluation_context, {0, 2}, componentId);
+    auto linear_expressions = visitor.dispatch(sum).GetLinearExpressions();
```
guilpier-code, line 123: About : 
```c++
auto linear_expressions = visitor.dispatch(sum).GetLinearExpressions();
BOOST_CHECK_EQUAL(linear_expressions.at(0).offset(), 9.);
```
Here we'd like to use `TimeDependentLinearExpression::operator[](int index)` instead (seems more simple and more natural) : 
```c++
auto linear_expressions = visitor.dispatch(sum);
BOOST_CHECK_EQUAL(linear_expressions[0].offset(), 9.);
```
but we can't, because this **[] operator** returns a **TimeDependentLinearExpression**.
This returned type seems strange if we consider class **TimeDependentLinearExpression** as a container of **LinearExpression**.
Would it make sense to make the **[] operator** return a **LinearExpression** ?

By the way, this operator is currently unused in production code and in tests.

guilpier-code (changes requested):

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
```diff
@@ -59,9 +59,12 @@ class TimeDependentLinearExpression
     TimeDependentLinearExpression operator-() const;
     TimeDependentLinearExpression shiftLinearExpressions(int shiftValue) const;
     TimeDependentLinearExpression operator[](int index) const;
```
guilpier-code:
```c++
TimeDependentLinearExpression operator[](int index) const
```
the argument is a time step, not a fuzzy index.
We should indicate it : `TimeDependentLinearExpression operator[](int timeStep) const`

guilpier-code (changes requested):

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
a-zakir: operator[](int index) applied on TimeDependentLinearExpression returns TimeDependentLinearExpression and it's used in ReadLinearExpressionVisitor as well in tests test_linearExpression.cpp::OperatorIndexInvalid and test_linearExpression.cpp::ShiftLinearExpressionsNegativeGreaterThanSize. 
TimeDependentLinearExpression::GetLinearExpressions return map of <timeStep, LinearExpression>

a-zakir: and it's true that TimeDependentLinearExpression::GetLinearExpressions().at(Ts) == TimeDependentLinearExpression[Ts]

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
```diff
@@ -188,4 +188,22 @@ TimeDependentLinearExpression ReadLinearExpressionVisitor::visit(const TimeIndex
     const auto timeIndex = static_cast<int>(evalVisitor_.dispatch(node->right()).valueAsDouble());
     return ret[timeIndex];
 }
+
```
guilpier-code, line 191: _**Out of this PR's scope**_ 

About : 
- `ReadLinearExpressionVisitor::visit(const ParameterNode* node)`
- `ReadLinearExpressionVisitor::visit(const VariableNode* node)`

In both cases, method contains an **if...else...**.
For **constant** cases, loop is inside **TimeDependentLinearExpression**'s constructor, for the other cases, loop is outside the constructor (on client's side).
It would  make sense to move the loop inside constructor for both cases.

we could even go a bit further and define several (overloaded) constructors for class **TimeDependentLinearExpression**.
Goal would be to have : 
```c++
ReadLinearExpressionVisitor::visit(const VariableNode* node)
{
    return TimeDependentLinearExpression(node, fillContext_, componentId_);
}
...
ReadLinearExpressionVisitor::visit(const ParameterNode* node)
{
    return TimeDependentLinearExpression(node, fillContext_, context_);
}
```
(Would it make sense to have a constructor for each **leaf node** ?)
Inside constructors, if needed, we could use a factory of **TimeDependentLinearExpression**

guilpier-code (changes requested):

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
```diff
@@ -59,9 +59,12 @@ class TimeDependentLinearExpression
     TimeDependentLinearExpression operator-() const;
     TimeDependentLinearExpression shiftLinearExpressions(int shiftValue) const;
```
guilpier-code, line 60: `TimeDependentLinearExpression shiftLinearExpressions(int shiftValue) const;`
is called internally to class **TimeDependentLinearExpression**, and nowhere else (apart from unit tests).
So it should be **private** (a good practice is to keep classes interfaces as smaller as possible).
Unit testing each method of a class leads to have only public methods, which is bad practice.
If we're worried about what to do with unit tests that call this method from outside a **TimeDependentLinearExpression**, we should remember that we should test behaviors, not methods.
We should make small unit tests calling only a small amount of public methods.
Here, we should unit test **timeSumLinearExpressions** instead (the only method calling **shiftLinearExpressions**).

guilpier-code (dismissed):

src/expressions/visitors/PrintVisitor.cpp
a-zakir: debatable; i always prefer to use stl as much as possible rather than external libs

src/tests/src/expressions/test_LinearVisitor.cpp
a-zakir: one is for the value it contains, but i"ll rename it

src/expressions/visitors/PrintVisitor.cpp
guilpier-code: This `boost::algorithm::trim(...)` is already used somewhere else in code base.
More generally, we already use boost thoroughly. 
And as you know, features of **boost** tend to be moved to **STL**.

src/tests/src/expressions/test_PrintAndEvalNodes.cpp
a-zakir: the id is the "name" of the param while value ...

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
a-zakir: using [] is misleading, what is the [] with negative index?

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
a-zakir: disagree!

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
a-zakir: this will mix LinearExpression notion and Ast

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
a-zakir: it"s called from ReadLinearExpressionVisitor::visit(const TimeShiftNode* node) && i do test behavior

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ReadLinearConstraintVisitor.h
guilpier-code: Yes, but i was talking about `ReadLinearConstraintVisitor::name()`

src/tests/src/expressions/test_AstDOTStyleVisitor.cpp
a-zakir: Yes it’s difficult. Sorry we are not at school

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
a-zakir: Test is fine

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
a-zakir: I don’t agree

a-zakir: Style of code

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
a-zakir: Because the concept is not easy

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
a-zakir: Disagree

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
a-zakir: Yes and?

flomnes (approved):

PR #2707, opened 2025-03-17
Small typo

flomnes:


MartinBelthle (commented):
While you're at it there's a little typo inside the migration guide for the v9.2 inside the St storage new fields. The file is said to be ` input/st-storage/series//list.ini` where it's supposed to be `input/st-storage/clusters/list.ini`

payetvin (approved):

PR #2708, opened 2025-03-17
Fix undefined behavior in BCs [ANT-2994]

payetvin:


src/libs/antares/study/include/antares/study/binding_constraint/BindingConstraintGroup.h
```diff
@@ -43,15 +46,14 @@ class BindingConstraintGroup
     }
 
     void add(const std::shared_ptr<BindingConstraint>& constraint);
-    [[nodiscard]] std::set<std::shared_ptr<BindingConstraint>> constraints() const;
+    [[nodiscard]] Set constraints() const;
```
flomnes:
```suggestion
    [[nodiscard]] const Set& constraints() const;
```

flomnes (commented):
If unused, please remove the other `Set` typedef.

flomnes (approved):

PR #2714, opened 2025-03-18
Features/scalian warning doubles

sylvmara:
Removes doubles from clusters, areas and binding contraints, and handles memory errors when binding constraints timeseries have the same id.
References gopro 2127

### DIFF
--- a/src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
+++ b/src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
@@ -218,6 +218,20 @@ bool BindingConstraintsRepository::loadFromFolder(Study& study,
             if (env.section->firstProperty)
             {
                 auto new_bc = LoadBindingConstraint(env);
+                for (int indexCst = 0; indexCst < constraints_.size(); indexCst++)
+                {
+                    if (constraints_.at(indexCst).get()->id() == new_bc.begin()->get()->id())
+                    {
+                        logs.warning()
+                          << "Two binding constraints with id " << new_bc.begin()->get()->id();
+                    }
+
+                    if (constraints_.at(indexCst).get()->name() == new_bc.begin()->get()->name())
+                    {
+                        logs.warning()
+                          << "Two binding constraints named " << new_bc.begin()->get()->name();
+                    }
+                }
                 std::copy(new_bc.begin(), new_bc.end(), std::back_inserter(constraints_));
             }
         }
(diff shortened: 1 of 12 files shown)


src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
```diff
@@ -218,6 +218,20 @@ bool BindingConstraintsRepository::loadFromFolder(Study& study,
...
+                        logs.warning()
+                          << "Two binding constraints with id " << new_bc.begin()->get()->id();
+                    }
+
+                    if (constraints_.at(indexCst).get()->name() == new_bc.begin()->get()->name())
+                    {
+                        logs.warning()
+                          << "Two binding constraints named " << new_bc.begin()->get()->name();
+                    }
+                }
```
flomnes: `constraints_` is a `std::vector<std::shared_ptr<BindingConstraint>>`

After reading all the constraints, you should loop on this vector and check for duplicates instead of doing it N times

This would allow for a better separation of responsibility and performance.

Having a dedicated `bool checkDuplicates() const` member function would make unit testing a lot easier.

src/libs/antares/utils/utils.cpp
```diff
@@ -181,7 +181,14 @@ bool checkAllElementsIdenticalOrOne(std::vector<unsigned> w)
...
 {
     // Erase 1 from the vector
+    auto message = p.begin()->second;
     std::erase_if(p, [](const auto& pair) { return pair.first == 1; });
+    if (p.size() == 0)
+    {
+        logs.error() << "Error : empty timeseries, checkAllElementsIdenticalOrOne failed for "
+                     << message;
+        return false;
+    }
```
flomnes: I'm not sure I understand the problem, can you explain ?

flomnes (changes requested):
I think you should make dedicated functions for checking duplicates make unit testing possible. Our coverage is 60%, below that the PR cannot be merged.

Also keep in mind that function `TransformNameIntoID` is not injective (there exists two different inputs x and y such that `TransformNameIntoID(x) == TransformNameIntoID(y)`, so checking for name duplicates may not be enough

src/libs/antares/study/binding_constraint/BindingConstraintsRepository.cpp
sylvmara: This is done, I made it a void which casts warnings directly

PR #2719, opened 2025-03-20
Trial for improvements on 3.3

guilpier-code:
This a small trial to improve [this PR ](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2699), it's related more specifically to [this remark](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2699#discussion_r2005599916)

**Not passing the CI for now (memory issue).**

### DIFF
--- a/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
+++ b/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
@@ -85,7 +85,8 @@ struct SimplexResult
     mpsWriterFactory mps_writer_factory;
 };
 
-static std::vector<LinearProblemFiller*> buildComponentFillers(const Antares::Study::SystemModel::System* modelerSystem)
+static std::vector<LinearProblemFiller*> buildComponentFillers(
+  const Antares::Study::SystemModel::System* modelerSystem)
 {
     std::vector<LinearProblemFiller*> componentFillers;
     if (!modelerSystem)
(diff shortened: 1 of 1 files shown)


guilpier-code:
The last commit not skipped by CI is failing because of a memory issue.
I tried to solve it, but haven't yet succeeded.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -85,24 +85,19 @@ struct SimplexResult
...
-    for (const auto& [_, component]: modelerSystem->Components())
+    for (const auto& component: modelerSystem->Components() | std::views::values)
     {
-        auto cf = std::make_unique<ComponentFiller>(component);
-        componentFillers.push_back(std::move(cf));
-    }
-    for (auto& component_filler: componentFillers)
-    {
-        fillersCollection.push_back(component_filler.get());
+        componentFillers.push_back(std::make_unique<ComponentFiller>(component).get());
```
flomnes, line 98: This is not equivalent to the previous code.

Because you're not keeping the `std::unique_ptr` alive, it will release memory at the end of scope. `componentFillers` will contain dangling pointers after function `buildComponentFillers` returns it.

A strong indication for this is `std::make_unique<Stuff>(args...).get()`, which you might call a code smell.

flomnes (changes requested):

flomnes (commented):
The real problem is the constructor of class `LinearProblemBuilder` 
```cpp
LinearProblemBuilder::LinearProblemBuilder(const std::vector<LinearProblemFiller*>& fillers)
```
It accepts a vector of raw pointers, forcing the caller to keep objects alive.

This can be fixed by changing this constructor to 
```cpp
LinearProblemBuilder::LinearProblemBuilder(const std::vector<std::shared_ptr<LinearProblemFiller>>& fillers)
```
or `std::unique_ptr` (doesn't support copy, the compiler might complain about `std::vector`)

PR #2720, opened 2025-03-20
Remove macro ndebug

payetvin:


flomnes (approved):

src/libs/antares/study/binding_constraint/BindingConstraint.cpp
```diff
@@ -139,10 +139,8 @@ const char* BindingConstraint::MathOperatorToCString(BindingConstraint::Operator
 
 BindingConstraint::~BindingConstraint()
 {
-#ifndef NDEBUG
     pName = "<INVALID>";
     pID = "<INVALID>";
-#endif
```
flomnes: This is probably useless, make `~BindingConstraint() = default`

src/libs/antares/date/date.cpp
```diff
@@ -382,11 +382,9 @@ static inline DayOfTheWeek NextDayOfTheWeek(DayOfTheWeek weekday)
 
 void Calendar::reset()
 {
-#ifndef NDEBUG
     logs.debug() << "  reset calendar" << ", month : " << MonthToString(settings_.firstMonth)
                  << ", january 1rst : " << DayOfTheWeekToString(settings_.weekday1rstJanuary)
                  << ", first weekday : " << DayOfTheWeekToString(settings_.weekFirstDay);
-#endif
```
flomnes: Useless function calls in Release mode

src/libs/antares/study/runtime/runtime.cpp
```diff
@@ -287,10 +287,8 @@ void StudyRuntimeInfos::initializeRandomNumberGenerators(const Parameters& param
     logs.info() << "Initializing random number generators...";
     for (uint i = 0; i != Data::seedMax; ++i)
     {
-#ifndef NDEBUG
         logs.debug() << "  random number generator: " << Data::SeedToCString((Data::SeedIndex)i)
                      << ", seed: " << parameters.seed[i];
-#endif
```
flomnes: Useless function calls in Release mode

flomnes (changes requested):
Where function calls are needed, please keep the #ifndef NDEBUG..#endif`. We don't want to add useless function calls in Release mode

flomnes (approved):

PR #2721, opened 2025-03-21
Init cost provider before calling get

payetvin:


src/libs/antares/study/parts/thermal/cluster_list.cpp
```diff
@@ -622,6 +622,7 @@ bool ThermalClusterList::loadEconomicCosts(Study& study, const fs::path& folder)
                                    auto filePath = folder / c->parentArea->id.c_str() / c->id();
 
                                    bool result = c->ecoInput.loadFromFolder(study, filePath);
+                                   c->initCostProvider();
```
flomnes, line 625: tests-simple-study won't run, since the study is built in memory

flomnes:
In my opinion this belongs in `StudyRuntimeInfos`

flomnes:
Done in #2725

PR #2724, opened 2025-03-24
4.8 Establishing Component Connections

a-zakir:
Associated specifications are [here](https://gopro-collaboratif.rte-france.com/display/ANTR/4.8+-+Import+de+connexions+de+ports).


src/io/inputs/yml-system/converter.cpp
```diff
@@ -121,17 +121,88 @@ static SystemModel::Component createComponent(const YmlSystem::Component& c,
...
+    {
+        throw std::invalid_argument("Both ports '" + first_port.Id() + "' and '" + second_port.Id()
+                                    + "' are " + (isFirstPortSending ? "senders " : "receivers"));
+    }
+
+    if (first_port.Type() != second_port.Type())
+    {
+        throw std::invalid_argument("Ports '" + first_port.Id() + "' and '" + second_port.Id()
+                                    + "' are not of the same type!");
+    }
```
flomnes: You need to check this condition for each field, not for each port.

src/study/system-model/include/antares/study/system-model/system.h
```diff
@@ -58,22 +59,26 @@ class System
...
 };
 
 class SystemBuilder
 {
 public:
     SystemBuilder& withId(std::string_view id);
-    SystemBuilder& withComponents(std::vector<Component>& components);
-    System build() const;
+    SystemBuilder& withComponents(std::unordered_map<std::string, Component>&& components);
+    SystemBuilder& withConnections(std::vector<Connection>&& connections);
```
flomnes, line 78: You could have 2 builder classes if you want to enforce order of call at compile-time. Suppose you want method `withComponents` to be called before `withConnections`. Then you can write [something like this](https://gist.github.com/flomnes/13de4b2edaa92a7c0b4f48fb6139e8f4).

flomnes (commented):
Most important first

1. For each field, check that one component is "emitting" and one is "receiving"
2. You removed a few checks, can you explain why ?
3. Proposal for improvement of the `SystemBuilder`

src/study/system-model/include/antares/study/system-model/system.h
a-zakir: order does not matter, compo and connections are built before the Builder

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -32,10 +32,10 @@ using namespace Antares::Study::SystemModel;
...
 {
     SystemBuilder system_builder;
-    std::vector<Component> components;
+    std::unordered_map<std::string, Component> components;
 };
 
-static Component createComponent(std::string id)
+static std::pair<std::string, Component> createComponent(std::string id)
 {
     ModelBuilder model_builder;
```
guilpier-code, line 43: There's a memory flaw here : we build a **Model**, then reference it inside the **Component**.
But when the surrounding function goes out of scope, the **Model** is destroyed, and the returned **Component** is left with a pointer to invalid memory.

It's does not have consequence because we don't use the **Component**'s **Model** in these unit tests.
Having the **Model*** to **nullptr** would be clear. 
In fact, we should have : 
```c++
static std::pair<std::string, Component> createComponent(std::string id)
{
    ComponentBuilder component_builder;
    auto component = component_builder.withId(id)
                       .withScenarioGroupId("scenario_group")
                       .build();
    return {id, component};
}
```
or even better : 
```c++
struct SystemBuilderCreationFixture
{
    ...
    Model model; 
    std::pair<std::string, Component> createComponent(std::string id)
    {
        return {id, Component({id, &model, {}, "scenario_group"})};
    }
};
```
**\/!\\** : This last suggestion does not compile because the used **Component** constructor is **private**.
Don't know why. 
But it could be useful to have it **public** here : using necessarily a **ComponentBuilder**  to create a **Component** may be overkill.
**Note** : I created an instance of a **Model** in the fixture because  it must have an address

guilpier-code (changes requested):

src/tests/src/study/system-model/test_system.cpp
a-zakir: the remark is correct, but this Pr did not introduce the faulty line.

guilpier-code: If this flaw can also be found elsewhere, we should also correct it.

src/io/inputs/yml-system/converter.cpp
```diff
@@ -121,17 +122,92 @@ static SystemModel::Component createComponent(const YmlSystem::Component& c,
...
+                << " in both ports " + first_port.Id() << "' and '" << second_port.Id() << "'";
+            throw std::invalid_argument(msg.str());
+        }
+    }
+
+    return {SystemModel::ConnectionEntry(&first_component, &first_port),
+            SystemModel::ConnectionEntry(&second_component, &second_port)};
+}
+
 SystemModel::System convert(const YmlSystem::System& ymlSystem,
```
guilpier-code, line 187: Instead, we may like to see : 
```c++
SystemModel::System convert(const YmlSystem::System& ymlSystem,
                            const std::vector<SystemModel::Library>& libraries)
{
    auto components = convert(ymlSystem.components);
    auto connections = convert(ymlSystem.connections);
    
    SystemModel::SystemBuilder builder;
    return builder.withId(ymlSystem.id)
      .withComponents(std::move(components))
      .withConnections(std::move(connections))
      .build();
}
```

src/io/inputs/yml-system/converter.cpp
tbittar: I do not agree, the connection should be between two ports of the same type, as the port type defines the fields that are available

src/io/inputs/yml-system/converter.cpp
```diff
@@ -121,17 +122,92 @@ static SystemModel::Component createComponent(const YmlSystem::Component& c,
     return component;
 }
 
+static SystemModel::Connection createConnection(
+  const YmlSystem::Connection& connection,
+  const std::unordered_map<std::string, SystemModel::Component>& components,
+  const std::vector<SystemModel::Library>& libraries)
```
guilpier-code: Last argument of **createConnection(...)** (named **libraries**) is unused.

guilpier-code (changes requested):

src/io/inputs/yml-system/converter.cpp
```diff
@@ -121,17 +122,106 @@ static SystemModel::Component createComponent(const YmlSystem::Component& c,
...
+ *
+ * @param connection A YmlSystem::Connection object containing the connection details.
+ * @param components An unordered map of component IDs to SystemModel::Component objects.
+ *
+ * @return A SystemModel::Connection object representing the created connection.
+ *
+ * @throw std::invalid_argument if a component or port is not found, if the ports are not
+ *        of the same type, or if fields are incorrectly configured for sending/receiving.
+ */
+static SystemModel::Connection createConnection(
```
guilpier-code, line 235: About **createConnection(...)** : 
This function is long.
Indeed it does 2 things (many checks and one creation at the end), and defines lambdas.
- Lambdas ; 
  have to be short (that's why they have no name) and are usually called from algorithms. Here lambdas have names and are not particularly short.
We could make them free functions and define them out of **createConnection(...)**.
- Couldn't we make checks in different functions, and have something like : 
```c++
for (const auto& connection: ymlSystem.connections)
{
    checkComponentsExistence(connection, components);
    checkPortsExistence(connection, components);
    checkPortsCompatibility(connection, components);
    connections.push_back(createConnection(connection, components));
}
```
?

src/io/inputs/yml-system/converter.cpp
```diff
@@ -121,17 +122,106 @@ static SystemModel::Component createComponent(const YmlSystem::Component& c,
...
+{
+    auto findComponent = [&components](const std::string& id) -> const SystemModel::Component&
+    {
+        const auto it = components.find(id);
+        if (it == components.end())
+        {
+            throw std::invalid_argument("Component with id '" + id + "' not found in system.");
+        }
+        return it->second;
+    };
```
guilpier-code: Maybe we could have : 
```c++
auto findComponent = [&components](const std::string& id) -> const SystemModel::Component&
{
    if (! components.contains(id))
    {
        throw std::invalid_argument("Component with id '" + id + "' not found in system.");
    }
     return components[id];
};
```

guilpier-code (dismissed):

src/io/inputs/yml-system/converter.cpp
a-zakir: sorry, my code is more efficient 😆

flomnes: @a-zakir is right, the current implementation only requires one lookup

guilpier-code: True

guilpier-code: > sorry, my code is more efficient 😆

True

src/io/inputs/yml-system/converter.cpp
a-zakir: the code is fine

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -78,22 +255,37 @@ BOOST_FIXTURE_TEST_CASE(fail_on_no_component1, SystemBuilderCreationFixture)
...
-BOOST_FIXTURE_TEST_CASE(fail_on_components_with_same_id, SystemBuilderCreationFixture)
+BOOST_AUTO_TEST_CASE(PortFieldLessOperator)
 {
-    components = {createComponent("component1"),
-                  createComponent("component2"),
-                  createComponent("component2")};
-    system_builder.withId("system").withComponents({components});
-    BOOST_CHECK_EXCEPTION(system_builder.build(),
-                          std::invalid_argument,
-                          checkMessage("System has at least two components with the same id "
```
a-zakir, line 95: moved https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/4b932eb7f7ad3da891c444b82673fcab83baef00/src/tests/src/io/yml-importers/testSystemConverter.cpp#L411

a-zakir:
> Most important first
> 
>     1. For each field, check that one component is "emitting" and one is "receiving"
> 
>     2. You removed a few checks, can you explain why ?
> 
>     3. Proposal for improvement of the `SystemBuilder`

the creation and non-duplication component are made in the same loop. in  `SystemConverter::convert`

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -78,22 +255,37 @@ BOOST_FIXTURE_TEST_CASE(fail_on_no_component1, SystemBuilderCreationFixture)
...
+
+BOOST_AUTO_TEST_CASE(PortFieldeEqualityOperator)
+{
+    BOOST_CHECK_EQUAL(PortField("A") == PortField("B"), false);
+    BOOST_CHECK(PortField("A") == PortField("A"));
+}
+
+BOOST_AUTO_TEST_CASE(PrintSenderFieldRole)
+{
+    constexpr auto role = FieldRole::Sender;
```
flomnes:
```suggestion
    FieldRole role = FieldRole::Sender;
```

src/tests/src/study/system-model/test_system.cpp
```diff
@@ -78,22 +255,37 @@ BOOST_FIXTURE_TEST_CASE(fail_on_no_component1, SystemBuilderCreationFixture)
...
+    std::ostringstream os;
+    os << role;
+    BOOST_CHECK_EQUAL(os.str(), "Sender");
+}
+
+BOOST_AUTO_TEST_CASE(PrintReceiverFieldRole)
+{
+    constexpr auto role = FieldRole::Receiver;
```
flomnes:
```suggestion
    FieldRole role = FieldRole::Receiver;
```

flomnes (approved):

src/tests/src/study/system-model/test_system.cpp
a-zakir: done

PR #2727, opened 2025-03-26
Use std::identity from header <functional>

flomnes:
See https://en.cppreference.com/w/cpp/utility/functional/identity


a-zakir (approved):

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -104,10 +96,10 @@ using FullKeyMap = std::unordered_map<FullKey, double, FullKeyHash>;
  * auto result = add_maps(map3, map4);
  * ```
  */
-template<typename MapType, typename UnaryOp = IdentityFunction<typename MapType::mapped_type>>
+template<typename MapType, typename UnaryOp = std::identity<typename MapType::mapped_type>>
```
a-zakir: i think there is no need for template args

PR #2731, opened 2025-03-27
Allow empty series files for STS [ANT-3053]

flomnes:


MartinBelthle (approved):
Seems okay to me but I'm a noobie so perhaps ask also the review of someone else

JasonMarechal25 (approved):

PR #2733, opened 2025-03-27
Add checks for duplicates on study objects [ANT-2127]

flomnes:


src/tests/src/libs/antares/study/test_duplicates.cpp
```diff
@@ -0,0 +1,209 @@
...
+    errors.clear();
+    area->thermal.list.clearAll();
+
+    addCluster("cluster_1");
+    addCluster("cluster_2");
+
+    BOOST_CHECK(checkForDuplicates(*study));
+
+    BOOST_REQUIRE_EQUAL(errors.size(), 0);
+    errors.clear();
```
flomnes: Démarrer un nouveau test ici (éviter les effets de bord)

src/tests/src/libs/antares/study/test_duplicates.cpp
```diff
@@ -0,0 +1,209 @@
...
+    areaB->thermal.list.clearAll();
+
+    addCluster("cluster_1");
+    addCluster("cluster_2");
+
+    BOOST_CHECK(checkForDuplicates(*study));
+
+    BOOST_REQUIRE_EQUAL(errors.size(), 0);
+    errors.clear();
```
flomnes: Nouveau test, pas besoin d'appeler `errors.clear()` (un nouvel objet `MessageHandler` sera instancié)

src/tests/src/libs/antares/study/test_duplicates.cpp
```diff
@@ -0,0 +1,209 @@
...
+    area->renewable.list.clearAll();
+
+    addCluster("cluster_1");
+    addCluster("cluster_2");
+
+    BOOST_CHECK(checkForDuplicates(*study));
+
+    BOOST_REQUIRE_EQUAL(errors.size(), 0);
+    errors.clear();
```
flomnes: Nouveau test

src/tests/src/libs/antares/study/test_duplicates.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    errors.clear();
+    area->shortTermStorage.storagesByIndex.clear();
+
+    addCluster("cluster_1");
+    addCluster("cluster_2");
+
+    BOOST_CHECK(checkForDuplicates(*study));
+    BOOST_REQUIRE_EQUAL(errors.size(), 0);
+    errors.clear();
```
flomnes: Nouveau test

src/tests/src/libs/antares/study/test_duplicates.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    errors.clear();
+    study->bindingConstraints.clear();
+
+    auto bc_3 = study->bindingConstraints.add("dummy_1");
+    auto bc_4 = study->bindingConstraints.add("dummy_2");
+
+    BOOST_CHECK(checkForDuplicates(*study));
+    BOOST_REQUIRE_EQUAL(errors.size(), 0);
+    errors.clear();
```
flomnes: Nouveau test

flomnes (commented):
@sylvmara C'est presque bon. Je pense qu'il faut séparer les tests pour faciliter le diagnostic en cas d'échec et éviter les effets de bord.

flomnes (commented):
Please check multiple duplicates e.g "cluster", "cluster", "cluster2", "cluster2"

src/solver/application/application.cpp
```diff
@@ -128,6 +129,8 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
             throw Error::NoAreas();
         }
 
+        checkForDuplicates(study);
```
payetvin: Return value is ignored and expect that logs.error() will stop the simulation
Make it void if it's the desired behavior or use the return value

src/libs/antares/study/duplicates.cpp
```diff
@@ -0,0 +1,90 @@
...
+template<class T>
+std::string getName(const std::shared_ptr<T>& t)
+{
+    return t->name();
+}
+
+template<class Container>
+bool check(const Container& c, const std::string& objectType, const std::string& context)
+{
+    std::set<std::string> names;
```
payetvin, line 45: The syntax is clean but we're copying ids of every object and inserting them into a set.

If the objects in the container are ordered alphanumerically we could just check that `object.name != nextObject.name` without using a set ?

I'm not sure it will make a real difference performance wise, it's just an idea.

src/libs/antares/study/parts/short-term-storage/cluster.cpp
```diff
@@ -64,6 +64,11 @@ bool STStorageCluster::enabled() const
     return properties.enabled;
 }
 
+std::string STStorageCluster::name() const
```
payetvin: We're mixing `cluster.id` and `cluster.properties.name` inside and outside the class, we should only have one of those used

payetvin (changes requested):

src/libs/antares/study/duplicates.cpp
flomnes: Yes, but we need to assume that the objects are sorted

payetvin: It is for thermal/renew/sts, idk about BCs

flomnes (commented):
@sylvmara please keep the return code for function `check`, and check it in the application.cpp (`throw FatalError` if this return value is false).

payetvin (approved):

PR #2736, opened 2025-03-28
Fix memory leak in hydro

JasonMarechal25:
Fix memory leak during hydro

## Allocate object on stack

_H2O_J_Instanciation_ return _DONNEES_MENSUELLES_ that was not released. Instead of managing a heap allocated object we return a value object. _DONNEES_MENSUELLES_ is small since most of it's data is heap allocated in vector of Simplexe object.

_DONNEES_MENSUELLES_ is not passed or copy around, so there is no performance hit for large vector copy for exemple

## Namespace hydro object

Besides best practice there was two definition of _CORRESPONDANCE_DES_VARIABLES_ in the global namespace. It didn't cause issue until now but trying to call destructor on those struct failed. Since there was two definitions, one had priority and tried to destroy the wrong type

## Default value initialization

Can't hurt


src/solver/hydro/include/antares/solver/hydro/monthly/h2o_m_donnees_optimisation.h
```diff
@@ -98,39 +100,44 @@ typedef struct
...
+    double CoutDeLaSolution{0.};
+    double CoutDeLaSolutionBruite{0.};
+
+    ~PROBLEME_HYDRAULIQUE()
+    {
+        for (auto* problem: ProblemeSpx)
+        {
+            SPX_LibererProbleme(problem);
+        }
+    }
```
flomnes: Consider using `std::unique_ptr` with `SPX_LibererProbleme` as custom deleter

flomnes (approved):

PR #2737, opened 2025-03-28
Fix leak using matrix_to_buffer_dumper_factory::get_dumper

JasonMarechal25:
_get_dumper_ return a heap allocated object that is never released. It now returns a unique_ptr to ensure proper liberation of ressources


src/libs/antares/array/include/antares/array/matrix-to-buffer.h
```diff
@@ -63,24 +62,18 @@ class I_mtx_to_buffer_dumper
...
     }
 
     void set_print_format(bool isDecimal, uint precision);
     virtual void run() = 0;
 
-    ~I_mtx_to_buffer_dumper()
-    {
-        delete format_;
-    }
-
```
flomnes, line 78: Keep a `virtual` destructor, even if it's the default destructor

flomnes (commented):
LGTM, please keep the virtual destructor for class `I_mtx_to_buffer_dumper`

flomnes (approved):

PR #2744, opened 2025-04-04
Sum connections : adapt antlr4 grammar

guilpier-code:


src/expressions/antlr-interface/Expr.g4
```diff
@@ -25,13 +25,14 @@ expr
     | expr op=('/' | '*') expr                 # muldiv
     | expr op=('+' | '-') expr                 # addsub
     | expr COMPARISON expr                     # comparison
-    | 'sum' '(' expr ')'                    # allTimeSum
+    | 'sum' '(' expr ')'                       # allTimeSum
+    | 'sum_connections' '(' expr ')'           # portFieldSum
```
a-zakir, line 29: if sum_connections can take only port, enforce the rule by doing:
```suggestion
    | 'sum_connections' '(' portField ')'           # sumConnections
```
/!\ i suggest to rename the rule sumConnections

src/expressions/nodes/PortFieldSumNode.cpp
```diff
@@ -29,6 +32,20 @@ PortFieldSumNode::PortFieldSumNode(const std::string& port_name, const std::stri
 {
 }
 
+PortFieldSumNode::PortFieldSumNode(const Node* node):
```
a-zakir, line 35: unnecessary

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -317,6 +318,12 @@ std::any ConvertorVisitor::visitAllTimeSum([[maybe_unused]] ExprParser::AllTimeS
     return static_cast<Node*>(registry_.create<AllTimeSumNode>(expr));
 }
 
+std::any ConvertorVisitor::visitPortFieldSum(ExprParser::PortFieldSumContext* context)
+{
+    auto expr = std::any_cast<Node*>(context->expr()->accept(this));
```
a-zakir, line 330: just retrieve PortId and field Id, the port should have been built earlier.

a-zakir: prior to the sum_connections instruction

a-zakir:
outdated see https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2746

guilpier-code:
Not useful anymore

PR #2746, opened 2025-04-07
Work on port rules

a-zakir:


flomnes (commented):
You only test error cases, please test the nominal case too

a-zakir:
added checks -> 2b6daf7cc3a552651eeb50b

flomnes (approved):

PR #2759, opened 2025-04-11
Mauvaise valeur turbine max mensuelle [ANT-3095]

payetvin:
On calcule le turbinage max par mois avec la formule suivante, pour chaque jour du mois:
`puissance maximale par jour * nombre d'heures de génération max / capacité du réservoir`


src/solver/hydro/management/monthly.cpp
```diff
@@ -181,21 +206,35 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
...
+                  unsigned simulationMonth = calendar_.mapping.months[realmonth];
+                  unsigned firstDay = calendar_.months[simulationMonth].daysYear.first;
+
+                  problem.TurbineMax[month] = calcTurbineMaxForMonth(simulationMonth,
+                                                                     y,
+                                                                     area,
+                                                                     calendar_,
+                                                                     scratchmap);
+
+                  capaciteTurbinageAnnuelle += problem.TurbineMax[month];
```
guilpier-code: **capaciteTurbinageAnnuelle** : we go towards more english and less french.
So maybe **capaciteTurbinageAnnuelle** could be replaced with **yearlyMaxGeneration**

src/solver/hydro/management/monthly.cpp
```diff
@@ -181,21 +206,35 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
...
                   problem.Apport[month] = data.inflows[realmonth];
                   problem.VolumeMin[month] = minLvl[firstDay];
                   problem.VolumeMax[month] = maxLvl[firstDay];
               }
 
+              if (totalInflowsYear > capaciteTurbinageAnnuelle)
+              {
+                  logs.debug() << "Apports annuels > capacité turbinage annuelle";
+                  problem.TurbineMax.assign(12, totalInflowsYear);
+              }
```
guilpier-code: I'm confused : here we enlarge the monthly max generation. 
It would seem natural that monthly max generation previously computed from max daily power cannot be enlarged.
In other words, instead of a `max(inflows, yearlyGeneration) `, shouldn't we have a **min** ?

src/solver/hydro/management/monthly.cpp
```diff
@@ -145,13 +145,38 @@ double HydroManagement::prepareMonthlyTargetGenerations(
     return total;
 }
 
+static double calcTurbineMaxForMonth(unsigned simulationMonth,
+                                     unsigned y,
+                                     Data::Area& area,
+                                     const Date::Calendar& calendar,
+                                     Antares::Data::Area::ScratchMap& scratchmap)
```
guilpier-code, line 152: **Out of this PR's scope**

**meanMaxDailyGenPower** should be held by `area.hydro`, not by **scratchpad**.
Instead, we're forced to pass scratchpad **and** the area to this function (whereas this function could have less arguments).
And we're also forced to pass the scratchpad in other multiple places.
Should be cleaned some day, hydro is already a mess and deserves better, and this could be done quickly.

src/solver/hydro/monthly/h2o_m_resoudre_le_probleme_lineaire.cpp
```diff
@@ -126,7 +126,7 @@ void H2O_M_ResoudreLeProblemeLineaire(DONNEES_ANNUELLES& DonneesAnnuelles, int N
     {
         if (ProblemeLineairePartieVariable.ExistenceDUneSolution != SPX_ERREUR_INTERNE)
         {
-            SPX_LibererProbleme(ProbSpx);
+            ProblemeHydraulique.ProblemeSpx[NumeroDeReservoir].reset(nullptr);
```
guilpier-code: ok : seems equivalent to old deletion.

guilpier-code (changes requested):

guilpier-code (approved):

PR #2773, opened 2025-04-17
Additional constraints : fix hours field 

guilpier-code:
Fixes [this ticket](https://gopro-tickets.rte-france.com/browse/ANT-3091)

**Diagnosis** : 
When converting an **hours** field into a collection of group, we first have to check the its format.
In order to do this, we used to use a large **regex**  : `"(^(\[\d+(,\d+)*\])(,(\[\d+(,\d+)*\]))*$)"`.
This is a good solution to check the format of **hours** field.
Unfortunately, on **Windows**, this long **regex** applied to long expressions creates a memory lack issue, and a crash eventually occurs.

**Chosen solution** : 
Instead, we first split the **hours** field into groups, and then we check that each group has the right format.

**What was done** : 
- [x] as said above, the format check made in 2 steps : first split into groups, then check format of each group.   
- [x] design was a bit changed : as for check on format, the code was split into 4 functions instead of one bigger function, so we can see the steps of converting a string into groups of hours
- [x]  unit tests are added (there already were unit tests, but they first have not been spotted. Besides, they test a higher level function [**loadAdditionalConstraints**], which requires IO, so they are slow. New tests are more numerous and much faster. Old tests were modified and integrated in new tests)


a-zakir:
You can’t split into group something that is ill-formed

src/tests/src/libs/antares/study/short-term-storage-input/test-on-making-hour-groups-from-string.cpp
```diff
@@ -0,0 +1,169 @@
...
+    const auto groupsOfHours = makeGroupsOfHours(hoursField);
+
+    const std::vector<std::set<int>> expected_result = {{5, 1, 98}, {8, 79, 54, 32}, {109}};
+    BOOST_CHECK(groupsOfHours == expected_result);
+}
+
+BOOST_AUTO_TEST_CASE(hours_field_contains_unwnated_but_ignored_chars_between_brackets)
+{
+    std::string hoursField = "[5, 1, 23], {I will be ignored}[4, 19]";
+
```
a-zakir: this is less restrictive than the original code

guilpier-code: The original unit tests were modified and integrated.
So there are simply more tests now, isn't it ?

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    std::regex commaSeparatedDigits(R"(^\d+(,\d+)*$)");
+    for (const auto& group: groups)
+    {
+        if (!std::regex_match(group, commaSeparatedDigits))
+        {
+            throw std::invalid_argument("splitting each group into hours > wrong format");
+        }
+        std::vector<std::string> hoursAsStrings;
+        boost::split(hoursAsStrings, group, boost::is_any_of(","));
```
a-zakir, line 87: `boost::split` is slower than `std::getline`

guilpier-code: Ok.
How do you know that ?

flomnes:
Duplicate of #2777

PR #2774, opened 2025-04-17
Remove some warnings

payetvin:


src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -234,7 +234,7 @@ class NotImplemented: public std::runtime_error
 static bool isThePortIsRegistered(const std::string& portId,
                                   const std::vector<YmlModel::Port>& ports)
 {
-    for (const auto [id, _]: ports)
+    for (const auto& [id, _]: ports)
```
JasonMarechal25, line 237: You can use ports | std::views::keys

JasonMarechal25 (changes requested):

src/io/inputs/model-converter/convertorVisitor.cpp
JasonMarechal25: disregard. I thought  ports was a map

JasonMarechal25 (approved):

src/io/inputs/model-converter/convertorVisitor.cpp
payetvin: ports is a vector not a map, we're getting the id of the ports this way

PR #2777, opened 2025-04-18
Use antlr to process st additional constraints [ANT-3091]

a-zakir:


src/libs/antares/study/include/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h
```diff
@@ -0,0 +1,34 @@
+
```
JasonMarechal25, line 21: Licence is missing

src/libs/antares/additionalConstraintRhsExpression/HoursField.g4
```diff
@@ -0,0 +1,12 @@
+grammar HoursField;
```
JasonMarechal25, line 22: If it is possible (doesn't break antlr), could you add licence header please.

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
```diff
@@ -0,0 +1,56 @@
+#include "antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h"
```
JasonMarechal25, line 21: licence is missing

JasonMarechal25 (approved):

JasonMarechal25 (dismissed):
A couple of licence headers are missing

src/libs/antares/study/include/antares/study/parts/short-term-storage/HoursCollectorVisitor.h
```diff
@@ -0,0 +1,47 @@
...
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+#include <HoursFieldBaseVisitor.h>
+
+class HoursCollectorVisitor: public HoursFieldBaseVisitor
+{
+public:
+    std::vector<std::set<int>> result;
```
guilpier-code: `std::vector<std::set<int>> result;` : seems to be useless here.
It seems it should be a local variable of **visitHoursField(...)** just below

And, if so, maybe call it **groups**, more clear.

src/libs/antares/study/include/antares/study/parts/short-term-storage/HoursCollectorVisitor.h
```diff
@@ -0,0 +1,47 @@
...
+
+    std::any visitHoursField(HoursFieldParser::HoursFieldContext* ctx) override
+    {
+        for (auto groupCtx: ctx->group())
```
guilpier-code, line 30: **groupCtx** : could be renamed into **group**, shorter, unless **groupCtx** is actually better.

src/libs/antares/study/include/antares/study/parts/short-term-storage/HoursCollectorVisitor.h
```diff
@@ -0,0 +1,47 @@
...
+        {
+            result.push_back(std::any_cast<std::set<int>>(visit(groupCtx)));
+        }
+        return result;
+    }
+
+    std::any visitGroup(HoursFieldParser::GroupContext* ctx) override
+    {
+        std::set<int> hours;
+        for (auto hourCtx: ctx->hour())
```
guilpier-code, line 40: **hourCtx** ==> **hour** ?

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
```diff
@@ -0,0 +1,56 @@
...
+    CustomErrorListener customErrorListener;
+    antlr4::ANTLRInputStream stream(hoursField);
+    HoursFieldLexer lexer(&stream);
+    lexer.removeErrorListeners();
+    lexer.addErrorListener(&customErrorListener);
+    antlr4::CommonTokenStream tokens(&lexer);
+    HoursFieldParser parser(&tokens);
+
+    parser.removeErrorListeners();
+    parser.addErrorListener(&customErrorListener);
```
guilpier-code: These are details for the current function **makeGroupsOfHours(...)**, we could make it very simple and have : 
```c++
std::vector<std::set<int>> makeGroupsOfHours(std::string& hoursField)
{
    auto parser = buildHoursFieldParser(hoursField);

    auto* tree = parser.hoursField();
    try
    {
        HoursCollectorVisitor visitor;
        return std::any_cast<std::vector<std::set<int>>>(visitor.visit(tree));
    }
    catch (std::exception& e)
    {
        ...
    }
}
```

src/libs/antares/study/include/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h
```diff
@@ -0,0 +1,34 @@
...
+class CustomErrorListener: public antlr4::BaseErrorListener
+{
+public:
+    void syntaxError(antlr4::Recognizer* recognizer,
+                     antlr4::Token* offendingSymbol,
+                     size_t line,
+                     size_t charPositionInLine,
+                     const std::string& msg,
+                     std::exception_ptr e) override;
+};
```
guilpier-code, line 50: If `class CustomErrorListener` is useful outside function **makeGroupsOfHours(...)** then it should remain here, but if not, it should be moved to **makeGroupsOfHoursFromString.cpp** and not be exposed to the outside world.

src/libs/antares/study/include/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include <BaseErrorListener.h>
+#include <set>
+#include <string>
+#include <vector>
+
+namespace antlr4
+{
+class Recognizer;
+}
```
guilpier-code, line 32: Same thing for : 
```c++
namespace antlr4
{
class Recognizer;
}
```

guilpier-code (changes requested):

src/libs/antares/study/include/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h
a-zakir: who said that?

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
a-zakir: the code you suggest contains bugs,  parser depends on tokens (pointer) which depends on lexer another pointer etc...

src/libs/antares/study/include/antares/study/parts/short-term-storage/HoursCollectorVisitor.h
a-zakir: hour is the result of std::stoi(hourCtx->getText()), got from hour context

src/libs/antares/study/include/antares/study/parts/short-term-storage/HoursCollectorVisitor.h
a-zakir: group is the result of std::any_cast<std::set<int>>(visit(groupCtx)), got from groupCtx which is GroupContext*

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -185,9 +145,19 @@ bool STStorageInput::loadAdditionalConstraints(const fs::path& parentPath)
...
             {
-                return false;
+                try
+                {
+                    std::string hoursField = value.c_str();
+                    additionalConstraints.constraints = makeConstraints(hoursField);
+                }
+                catch (const std::exception& e)
+                {
+                    logs.error() << "Constraint " << additionalConstraints.name << " : " << e.what()
```
flomnes, line 157: File name instead

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -185,9 +145,19 @@ bool STStorageInput::loadAdditionalConstraints(const fs::path& parentPath)
             {
                 value.to<std::string>(additionalConstraints.operatorType);
             }
-            else if (key == "hours" && !loadHours(value.c_str(), additionalConstraints))
+            else if (key == "hours")
             {
-                return false;
+                try
+                {
+                    std::string hoursField = value.c_str();
```
flomnes, line 152: `to<std::string>`

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
```diff
@@ -0,0 +1,94 @@
...
+        tokens_(&lexer_),
+        parser_(&tokens_)
+    {
+        lexer_.removeErrorListeners();
+        lexer_.addErrorListener(&customErrorListener_);
+        parser_.removeErrorListeners();
+        parser_.addErrorListener(&customErrorListener_);
+    }
+
+    auto* hoursField()
```
flomnes, line 64: Not a huge fan of auto in this context, having the actual type would improve readability

flomnes (commented):
Style

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
```diff
@@ -0,0 +1,94 @@
...
+    std::ostringstream os;
+    os << "Syntax error at line " << line << ":" << charPositionInLine << " - " << msg << std::endl;
+    if (offendingSymbol)
+    {
+        os << "Offending symbol: " << offendingSymbol->getText() << std::endl;
+    }
+    throw ShortTermStorageAdditionalConstraintsError(os.str());
+}
+
+class GroupsHours
```
guilpier-code, line 48: `class GroupsHours` : this class is a parser, right ? I means : we give this object a string and we get a tree, result of a parsing.
Why not call it  **GroupsOfHoursParser** then ? And give it a **operator()**.
And have (for instance) : 
```c++
GroupsOfHoursParser parser(hoursField);
auto* tree = parser();
```

a-zakir:
i added very large rhs (100000) hours field

src/libs/antares/study/include/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h
guilpier-code: Will client code of  function **makeGroupsOfHours** need to know about **CustomErrorListener** ?
If not, it's because **CustomErrorListener** is "private" to the **makeGroupsOfHours**.
In that case, we should avoid exposing **CustomErrorListener** to the outside world, and declare + define it in the .cpp file.

A proof that the knowledge of **CustomErrorListener** is not needed outside is that, when moving  all about **CustomErrorListener** in the cpp file, the code still compiles.

guilpier-code (changes requested):

src/libs/antares/study/include/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h
a-zakir: a proof you said? what did you proven? i generally prefer to declare class in header, implement it in .cpp. NO ONE is forced to use it though.

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
a-zakir: no it is not a parser, i wonder if you really understand what is a parser?  read carefully, why i'm going to complicate nice, precise and easy-understandable code with ugly () operator?

guilpier-code: Please don't think I'm an angry guy when I do remarks. I'm not.
Skip it if you don't like this remark.

guilpier-code (approved):

src/libs/antares/study/include/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h
flomnes: What @guilpier-code meant to say is that we should expose in a public header only the classes & functions that are to be used by the clients of this library, as opposed to internal components.

flomnes (approved):

PR #2778, opened 2025-04-22
Some cleanup

JasonMarechal25:
Fix several warnings
Remove dead code


src/expressions/antlr-interface/ExprParser.cpp
```diff
@@ -1422,7 +1422,7 @@ bool ExprParser::sempred(RuleContext *context, size_t ruleIndex, size_t predicat
   return true;
 }
 
-bool ExprParser::exprSempred(ExprContext *_localctx, size_t predicateIndex) {
+bool ExprParser::exprSempred([[maybe_unused]] ExprContext *_localctx, size_t predicateIndex) {
```
a-zakir: this is generated by antlr, it should not be modified

JasonMarechal25: Good catch, thank you

flomnes:
I mostly agree with the changes proposed. Unfortunately the coverage is only about 40%, so we have 2 options
- Ignore the warning and merge as is (I don't like this option)
- Add some tests, which can be done by splitting this PR into multiple, smaller PRs

flomnes (approved):

PR #2785, opened 2025-04-28
Ceiling error with xpress and accurate [ANT-3167]

Juliette-Gerbaux:
When X = 1 + 1e-12, (int)ceil(X) = 2 and we want to remove the term 1e-12 before ceiling.


src/solver/optimisation/opt_nombre_min_groupes_demarres_couts_demarrage.cpp
```diff
@@ -154,7 +154,8 @@ void OPT_AjusterLeNombreMinDeGroupesDemarresCoutsDeDemarrage(PROBLEME_HEBDO* pro
                           PmaxDUnGroupeDuPalierThermique);
                     }
 
-                    NombreMinDeGroupesEnMarcheDuPalierThermique[pdtHebdo] = (int)ceil(X);
+                    NombreMinDeGroupesEnMarcheDuPalierThermique[pdtHebdo] = (int)ceil(
+                      std::round(X * 1000000) / 1000000);
```
flomnes: Use existing function `Utils::ceilDiv`, for reference

```cpp
static constexpr double largeValue = 1000000;

double ceilDiv(double numerator, double denominator)
{
    return std::ceil(std::round(numerator / denominator * largeValue) / largeValue);
}
```

flomnes (changes requested):

flomnes (approved):

PR #2790, opened 2025-04-29
Fix fixed variable value in quadratic optimization

pet-mit:


flomnes (approved):

pet-mit:
hello @hugo-antoine-rtei 
what do you think of this? 
Besides, I'm troubled by the fact that no test fails when changing "-" to "+". If RTE-i is at the source of the code, do you mind adding tests (preferably unit tests, else behave tests under "cucumber")? (and telling us if this is an actual bug or not .. )

hugo-antoine-rtei:
Hi @pet-mit,
I don't think RTE-i has modified this particular bit of code, and I don't really know what it's supposed to do
I'll leave this as approved.

hugo-antoine-rtei (approved):

PR #2792, opened 2025-05-02
[ANT-2902] 3.6 - Connect components to areas in linear optimization problem

pet-mit:
Allow the addition of components contribution to areas' balance constraint, as documentent here : https://antares-simulator.readthedocs.io/en/latest/user-guide/solver/08-hybrid-studies/#optimization-model


docs/user-guide/solver/08-hybrid-studies.md
```diff
@@ -88,4 +86,16 @@ connect modeler components to solver areas.
 
 #### Optimization model
 The linear expression defined by the connected component's port field definition is simply added to the **left-hand side** 
-of the area's [balance constraint](05-model.md#balance-between-load-and-generation).
\ No newline at end of file
+of the area's [balance constraint](05-model.md#balance-between-load-and-generation).  
+
+???+ warning
+    
+    The current convention of this constraint is:
```
JasonMarechal25, line 93: Convention will not be met by user. Are we asserting this in the code?

src/libs/antares/exception/include/antares/exception/RuntimeError.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+#pragma once
+
+#include <stdexcept>
+
+namespace Antares
+{
+namespace Error
+{
+class RuntimeError: public std::runtime_error
```
JasonMarechal25, line 36: What is the point of this class? Just to silence linter/sonar about using standard exceptions?

src/tests/cucumber/features/steps/common_steps/solver_steps.py
```diff
@@ -83,10 +83,32 @@ def check_annual_cost(context):
...
-    assert_double_close(0, context.soh.get_annual_system_cost()["STD"], 0.001)
-    assert_double_close(one_year_value, context.soh.get_annual_system_cost()["MIN"], 0.001)
-    assert_double_close(one_year_value, context.soh.get_annual_system_cost()["MAX"], 0.001)
+    assert_double_close(one_year_value, context.soh.get_annual_system_cost()["EXP"], 0.00001)
+    assert_double_close(0, context.soh.get_annual_system_cost()["STD"], 0.00001)
+    assert_double_close(one_year_value, context.soh.get_annual_system_cost()["MIN"], 0.00001)
+    assert_double_close(one_year_value, context.soh.get_annual_system_cost()["MAX"], 0.00001)
+
+
+@then('the annual system cost is {one_year_value1:g} with the linear solver {solver1} and {one_year_value2:g} with the others')
```
JasonMarechal25, line 92: I fear test cases will not be clear, or their errors won't be either.

Reading the then clause I expect value1 and value 2 to be checked for the respective linear and quadratic/other solver but the step will only check one of the two values depending on whether or not solver1 match the linear solver.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -132,25 +134,62 @@ static void writeModelerSolutions(const operations_research::MPSolver* solver,
     writer.addEntryFromBuffer(modelerSolutionFilename, content);
 }
 
+FillContext buildFillContext(PROBLEME_HEBDO* problemeHebdo, const int NumIntervalle)
```
JasonMarechal25:
```suggestion
FillContext buildFillContext(PROBLEME_HEBDO* problemeHebdo, int NumIntervalle)
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -132,25 +134,62 @@ static void writeModelerSolutions(const operations_research::MPSolver* solver,
...
+{
+    unsigned firstTimestamp, lastTimestamp;
+    if (problemeHebdo->OptimisationAuPasHebdomadaire)
+    {
+        firstTimestamp = problemeHebdo->weekInTheYear
+                         * static_cast<unsigned int>(problemeHebdo->NombreDePasDeTempsDUneJournee)
```
JasonMarechal25: Maybe not important but could `static_cast<unsigned int>(problemeHebdo->NombreDePasDeTempsDUneJournee)  * problemeHebdo->NombreDeJours;` be constant and computed only once ?

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
```diff
@@ -0,0 +1,69 @@
...
+#include "antares/solver/utils/named_problem.h"
+#include "antares/study/system-model/system.h"
+
+namespace Antares::Optimization
+{
+class ComponentToAreaConnectionFiller: public Optimisation::LinearProblemApi::LinearProblemFiller
+{
+public:
+    explicit ComponentToAreaConnectionFiller(const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe,
+                                             const unsigned int nTimestampsInProblem,
```
JasonMarechal25:
```suggestion
                                             unsigned int nTimestampsInProblem,
```

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "antares/solver/simulation/sim_structure_probleme_economique.h"
+
+using namespace Antares::Optimisation::LinearProblemApi;
+
+namespace Antares::Optimization
+{
+
+ComponentToAreaConnectionFiller::ComponentToAreaConnectionFiller(
+  const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe,
+  const unsigned int nTimestampsInProblem,
```
JasonMarechal25:
```suggestion
  unsigned int nTimestampsInProblem,
```

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
```diff
@@ -0,0 +1,69 @@
...
+    void addObjective(Optimisation::LinearProblemApi::ILinearProblem& pb,
+                      Optimisation::LinearProblemApi::ILinearProblemData& data,
+                      Optimisation::LinearProblemApi::FillContext& ctx) override;
+
+private:
+    const ModelerStudy::SystemModel::System* modelerSystem_;
+    const VariableDictionary& modelerVariableDictionary_;
+    std::map<std::pair<std::string, unsigned int>, std::string>
+      balanceConstraintPerAreaAndTimestamp_;
+    unsigned int nTimestampsInProblem_;
```
JasonMarechal25:
```suggestion
    const unsigned int nTimestampsInProblem_;
```

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
```diff
@@ -0,0 +1,69 @@
...
+                        Optimisation::LinearProblemApi::ILinearProblemData& data,
+                        Optimisation::LinearProblemApi::FillContext& ctx) override;
+    void addObjective(Optimisation::LinearProblemApi::ILinearProblem& pb,
+                      Optimisation::LinearProblemApi::ILinearProblemData& data,
+                      Optimisation::LinearProblemApi::FillContext& ctx) override;
+
+private:
+    const ModelerStudy::SystemModel::System* modelerSystem_;
+    const VariableDictionary& modelerVariableDictionary_;
+    std::map<std::pair<std::string, unsigned int>, std::string>
```
JasonMarechal25: Maybe we should introduce some using to improve readability.
What is the area, what is the balance constraint ? What is the pair ?

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+    return problemeSimplexe->ConstraintNames().at(index);
+}
+
+void ComponentToAreaConnectionFiller::parseConstraintIds(
+  const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe)
+{
+    for (int idxRow = 0; idxRow < problemeSimplexe->NombreDeContraintes; ++idxRow)
+    {
+        std::regex pattern(R"(AreaBalance::area<(.+)>::hour<(\d+)>)");
```
JasonMarechal25: If we expect a significant number of iterations, we should use another regex provider than std. std::regex is very slow.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+                                                   FillContext& ctx)
+{
+    // nothing to do
+}
+
+static std::string getConnectionFieldId(const ModelerStudy::SystemModel::Component& component,
+                                        const std::string& portId)
+{
+    auto optionalField = component.getModel()->Ports().at(portId).Type().AreaConnectionFieldId();
+    if (!optionalField.has_value())
```
JasonMarechal25: It's strange to read. We are enforcing optionalField to have a value. Is it optional or mandatory ?

src/libs/antares/exception/include/antares/exception/RuntimeError.hpp
```diff
@@ -0,0 +1,42 @@
...
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+**
+** SPDX-License-Identifier: MPL-2.0
+*/
+
+#pragma once
```
guilpier-code, line 28: **Question** : 
class **RuntimeError** adds nothing more to `std::runtime_error`.
Is there a particular reason why we add this new class ?

JasonMarechal25 (changes requested):

docs/user-guide/solver/08-hybrid-studies.md
pet-mit: we can't verify this in the code, it depends on the convention the user uses in their yml library. End users have already been notified and they validated this

src/libs/antares/exception/include/antares/exception/RuntimeError.hpp
pet-mit: Unfortunately Sonar doesn't like standard exceptions. The point is that this way, users will be able to catch and manage specifically Antares' runtime errors (and not other std runtime_errors). Personally I agree with this point of view

src/tests/cucumber/features/steps/common_steps/solver_steps.py
pet-mit: I can replace it with something like this one : https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/4e8bd1a373d0186be20c0363db987967a2bd0281/src/tests/cucumber/features/steps/common_steps/modeler_steps.py#L30

pet-mit: (in all cases, I think that this is only temporary, since we mustn't have different values depending on solver. I was able to pin-point the source of this difference and notify R&D about it)

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
pet-mit: when a component is connected to an area, the port type MUST define an area connection field. if it's not connected, then the port type does not have to.

in this case, the component is connected.

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
pet-mit: done

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
pet-mit: replaced with boost regex

JasonMarechal25 (approved):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -132,25 +134,56 @@ static void writeModelerSolutions(const operations_research::MPSolver* solver,
     writer.addEntryFromBuffer(modelerSolutionFilename, content);
 }
 
+FillContext buildFillContext(PROBLEME_HEBDO* problemeHebdo, int NumIntervalle)
```
guilpier-code, line 137: - **very cosmetic** : **unsigned int** ==> **unsigned** (shorter)
- **firstTimestamp** ==> **firstTimeStep** (same for last time step)
- **question** : couldn't this function be more simple ? I mean : 
   - `problemeHebdo->NombreDePasDeTempsDUneJournee` is supposed to be **24** (or **HOURS_PER_DAY** in **constants.h**)
   - `problemeHebdo->NombreDeJours;` is supposed to be **7** (if we're in a weekly optimization)
   - ...

It be could be something like (a bit shorter) : 
```c++
FillContext buildFillContext(PROBLEME_HEBDO* problemeHebdo, const int NumIntervalle)
{
    unsigned firstTimeStep = problemeHebdo->weekInTheYear * Constants::nbHoursInAWeek;
    unsigned lastTimeStep = firstTimeStep + Constants::nbHoursInAWeek - 1;
    if (!problemeHebdo->OptimisationAuPasHebdomadaire)
    {
        firstTimeStep += NumIntervalle * HOURS_PER_DAY;
        lastTimeStep = firstTimeStep + HOURS_PER_DAY - 1;
    }
    return FillContext(firstTimeStep, lastTimeStep);
}
```
What's wrong with that ?

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
```diff
@@ -0,0 +1,74 @@
...
+#include "antares/solver/simulation/sim_structure_probleme_economique.h"
+#include "antares/solver/utils/named_problem.h"
+#include "antares/study/system-model/system.h"
+
+namespace Antares::Optimization
+{
+class ComponentToAreaConnectionFiller: public Optimisation::LinearProblemApi::LinearProblemFiller
+{
+public:
+    explicit ComponentToAreaConnectionFiller(const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe,
```
guilpier-code, line 36: Why time stamp ?
We usually use **time step**. Are we wrong ?
Time stamp is usually associated to digital record, with a date and hour.
Seems strange in the context of Antares  Simulator.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+    {
+        return 'c' + std::to_string(index);
+    }
+    return problemeSimplexe->ConstraintNames().at(index);
+}
+
+void ComponentToAreaConnectionFiller::parseConstraintIds(
+  const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe)
+{
+    for (int idxRow = 0; idxRow < problemeSimplexe->NombreDeContraintes; ++idxRow)
```
guilpier-code, line 64: **Question :** 
Here we use indices to loop over constraint names, couldn't we directly loop over constraints names, and drop indices ?
```c++
for (auto& name : problemeSimplexe->ConstraintNames())
{
    ...
}
```

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+    // This should be in sync with LegacyFiller::GetConstraintName
+    if (!problemeSimplexe->UseNamedProblems()
+        || problemeSimplexe->ConstraintNames().at(index).empty())
+    {
+        return 'c' + std::to_string(index);
+    }
+    return problemeSimplexe->ConstraintNames().at(index);
+}
+
+void ComponentToAreaConnectionFiller::parseConstraintIds(
```
guilpier-code, line 61: Filtering **balance contraints** doesn't seem to be among responsibilities of your class **ComponentToAreaConnectionFiller**, it could be done inside **PROBLEME_SIMPLEXE_NOMME**, where any information to do this is available.

Moreover, **getLegacyConstraintName** seems to be a duplication of **LegacyFiller::GetConstraintName**.
So doing the work inside **PROBLEME_SIMPLEXE_NOMME** would allow to remove this duplication.

If we do that, we may not need neither : 
- method **parseConstraintIds** 
- nor free function **getLegacyConstraintName**
- or even data member **balanceConstraintPerAreaAndTimestamp_** (to be experienced)

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+  const ModelerStudy::SystemModel::System* modelerSystem,
+  const VariableDictionary& modelerVariableDictionary):
+    modelerSystem_(modelerSystem),
+    modelerVariableDictionary_(modelerVariableDictionary),
+    nTimestampsInProblem_(nTimestampsInProblem)
+{
+    parseConstraintIds(problemeSimplexe);
+}
+
+static std::string getLegacyConstraintName(const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe,
```
guilpier-code, line 49: This an example that shows that code contained in **getLegacyConstraintName** does not belong here :  we only use **problemeSimplexe** in this function, and nothing else.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+    return optionalField.value();
+}
+
+IMipConstraint* ComponentToAreaConnectionFiller::getBalanceConstraint(ILinearProblem& pb,
+                                                                      const std::string& areaId,
+                                                                      unsigned ts) const
+{
+    auto key = std::make_pair(areaId, ts % nTimestampsInProblem_);
+    if (balanceConstraintPerAreaAndTimestamp_.contains(key))
```
guilpier-code: Here (as @a-zakir would say) we have a find algorithm used twice : 
- when calling **contains(...)**
- when calling **at(...)**

To avoid this, we can use **std::map::find(...)**, which return an iterator pointing on the found item, or the end of the **std::map**. Doing this, **find** algorithm is used once.
```c++
auto it = my_map.find(key);
if (it == my_map.end())
{
    throw ...
}
return pb.lookupConstraint(it->id);
```

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (!optionalField.has_value())
+    {
+        throw Error::RuntimeError("Component \"" + component.Id()
+                                  + "\" is connected to an area using a port type that has no "
+                                    "area-connection field defined.");
+    }
+    return optionalField.value();
+}
+
+IMipConstraint* ComponentToAreaConnectionFiller::getBalanceConstraint(ILinearProblem& pb,
```
guilpier-code, line 98: **getBalanceConstraint(...)** : this function fetches any **Mip** constraint, not a balance constraint in particular.
It returns a balance constraint depending on what is given in input.
So it should be renamed into **MipConstraint(...)** (for instance).

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
guilpier-code: Please see further current source file's remarks before this one.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
guilpier-code: **Note** : same remark for **LegacyFiller::GetVariableName**.

src/study/system-model/include/antares/study/system-model/component.h
```diff
@@ -103,6 +103,8 @@ class Component
 
     std::optional<std::string> areaConnectedToPort(const std::string& portId) const;
 
+    const std::map<std::string, std::string>& areaConnectionPerPort() const;
+
 private:
```
guilpier-code, line 108: **Very cosmetic** : **areaConnections_** ==> **connectionsToAreas_** ? A bit more clear, but up to you.

But at least we could use the plural : there are several connections to areas, right ?

src/study/system-model/include/antares/study/system-model/component.h
```diff
@@ -103,6 +103,8 @@ class Component
 
     std::optional<std::string> areaConnectedToPort(const std::string& portId) const;
 
+    const std::map<std::string, std::string>& areaConnectionPerPort() const;
```
guilpier-code: **areaConnectionPerPort()** ==> **connectionsToAreasViaPort()**

I guess there are several connections, hence the plural.

src/study/system-model/component.cpp
```diff
@@ -123,6 +123,11 @@ std::optional<std::string> Component::areaConnectedToPort(const std::string& por
                                              : std::nullopt;
 }
```
guilpier-code, line 126: This remark is out of this scope but it's very easy to give it an answer : 
in **portType.h**, we have : 
```c++
PortType(...): ...
{
    if (areaConnectionFieldId.empty())
    {
        areaConnectionFieldId_ = std::nullopt;
    }
    else
    {
        // do something
    }
}
```
**areaConnectionFieldId_** is a `std::optional<std::string>`, which is `std::nullopt` by default.
Instead, we should do : 
```c++
PortType(...): ...
{
    if (!areaConnectionFieldId.empty())
    {
        // do something
    }
}
```

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+                                                   ILinearProblemData& data,
+                                                   FillContext& ctx)
+{
+    // nothing to do
+}
+
+static std::string getConnectionFieldId(const ModelerStudy::SystemModel::Component& component,
+                                        const std::string& portId)
+{
+    auto optionalField = component.getModel()->Ports().at(portId).Type().AreaConnectionFieldId();
```
guilpier-code: optionalField ==> **field** is enough. 
Indeed, when you define a **double** var, we don't call it **doubleVar**.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+void ComponentToAreaConnectionFiller::addComponentPortContributionToArea(
+  ILinearProblem& pb,
+  ILinearProblemData& data,
+  const FillContext& ctx,
+  const ModelerStudy::SystemModel::Component& component,
+  const std::string& portId,
+  const std::string& areaId)
+{
+    std::string fieldId = getConnectionFieldId(component, portId);
```
guilpier-code: `std::string fieldId` : you mean **injectionFiledId** ?

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+    const Expressions::Visitors::EvaluationContext
+      connectedComponentEvalContext(component.getParameterValues(), {}, data);
+    ReadLinearExpressionVisitor visitor(connectedComponentEvalContext, ctx, component);
+    auto timeDependentLinearExpression = visitor.dispatch(
+      component.nodeAtPortField(portId, fieldId));
+    // Transform areaId to lower case
+    std::string lowerAreaId = areaId;
+    std::transform(lowerAreaId.begin(), lowerAreaId.end(), lowerAreaId.begin(), ::tolower);
```
guilpier-code: Instead of : 
```c++
// Transform areaId to lower case
std::string lowerAreaId = areaId;
std::transform(lowerAreaId.begin(), lowerAreaId.end(), lowerAreaId.begin(), ::tolower);
```
We could have : 
```c++
auto lowerAreaId = toLower(areaId);
```
Note that the comment is no longer needed.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,168 @@
...
+        // Contribution is added to the left-hand side of the constraint
+        // We invert the sign bc modeler is in "gen>0, load<0" convention
+        // legacy constraint is in "gen<0, load>0" convention
+        for (const auto& [varKey, coef]: expression.coefPerVar())
+        {
+            auto* var = modelerVariableDictionary_[varKey];
+            areaBalanceConstraint->setCoefficient(var, -coef);
+        }
+        areaBalanceConstraint->setBounds(areaBalanceConstraint->getLb() + expression.offset(),
+                                         areaBalanceConstraint->getUb() + expression.offset());
```
guilpier-code: Maybe we could have : 
```c++
for (const auto& [ts, expression]: timeDependentLinearExpression.GetLinearExpressions())
{
    IMipConstraint* balanceConstraint = getBalanceConstraint(pb, lowerAreaId, ts);
    addExpressionToConstraint(expression, balanceConstraint);
}
```
with **addExpressionToConstraint** a private member function.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -132,25 +134,56 @@ static void writeModelerSolutions(const operations_research::MPSolver* solver,
...
     std::vector<LinearProblemFiller*> fillersCollection = {&legacyOrtoolsFiller};
 
     std::vector<std::unique_ptr<ComponentFiller>> componentFillers;
-    // All LP variables coordinates (component id, variable id, scenario, time step)
     VariableDictionary variableDictionary;
-    fillModelerComponents(componentFillers,
-                          fillersCollection,
-                          problemeHebdo->modelerSystem,
-                          variableDictionary);
+    ComponentToAreaConnectionFiller compatibilityFiller(
```
guilpier-code: **ComponentToAreaConnectionFiller compatibilityFiller** : type and variable names are quite different.
Is it because the first one is too long ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -312,18 +345,18 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
...
+        throw FatalError("Internal error: insufficient memory");
     }
 
     writeModelerSolutions(solver, Probleme, optimizationNumber, optPeriodStringGenerator, writer);
 
-    return {.success = true, .timeMeasure = timeMeasure, .mps_writer_factory = mps_writer_factory};
+    return {.success = true,
+            .timeMeasure = timeMeasure,
+            .mps_writer_factory = mps_writer_factory,
+            .objectiveValue = solver != nullptr ? solver->Objective().Value() : 0};
```
guilpier-code, line 359: **_Probably not in the scope of this PR_**
- Why a mps_writer_factory is taken in a **SimplexResult**. Seems to make no sense.
- We return a` success = true`, so why **solver** should be` != nullptr` ? We normally should return only `solver->Objective().Value()`

src/tests/src/solver/optimisation/test_componentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,330 @@
...
+      area: Area1
+    - component: connected_component_param
+      port: connection_port_param
+      area: Area1
+    )"s;
+
+struct ComponentToAreaConnectionFillerFixture
+{
+    std::unique_ptr<PROBLEME_SIMPLEXE_NOMME> problemeSimplexe;
+    unsigned int nTimestampsInProblem{};
```
guilpier-code: `unsigned int nTimestampsInProblem{};` : unused

src/solver/optimisation/CMakeLists.txt
```diff
@@ -205,6 +208,7 @@ target_link_libraries(model_antares
         Antares::benchmarking
         Antares::optimization-options
         Antares::lps
+        Boost::headers
```
guilpier-code: Can we put **Boost::headers** in the **PRIVATE** section ?

src/libs/antares/exception/include/antares/exception/RuntimeError.hpp
pet-mit: see discussion [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2792#discussion_r2090923423)

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
pet-mit: ok for first 2 suggestions.
concerning the other suggestions, I prefer that this function stays generic and not introduce more hard-coded constants in the code (SRP)

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
a-zakir: Yes, I would also remove this function. We could simply add an unordered_map field to ProblemHebdo, which can be populated when the  [AreaBalance](https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/96796036494c3f414234b18ab79e7846cf918e6f/src/solver/optimisation/constraints/AreaBalance.cpp#L46) objects are created.

pet-mit: thanks, I'll try this out

a-zakir: We don’t have ortoolsMip at this stage (areabalance.cpp) but could not this work with cstraint index ?

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
guilpier-code: You can also use `boost::algorithm::to_lower(areaId);`

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
pet-mit: I guess that, in the context of PROBLEME_HEBDO, the constraint index is the only thing that makes sense. 
However, when cross-referencing with ortoolsMip, we'll have to use the contraint ID.
So I think that it's not going to be very easy, I'll have to experiment

src/tests/src/solver/optimisation/test_componentToAreaConnectionFiller.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+    void addEmptyConstraints(std::vector<std::string>& names, double rhs)
+    {
+        for (const auto& name: names)
+        {
+            linearProblem.addConstraint(rhs, rhs, name);
+        }
+    }
+
+    void setUpLegacyLp(std::vector<std::string>& constraintNames, bool useNamedProblems, double rhs)
```
guilpier-code: `setUpLegacyLp` : does too many things... (draft remark, to be finished)

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
pet-mit: it does lookup the constraint in the balance constraints map

src/study/system-model/include/antares/study/system-model/component.h
pet-mit: renamed to portToAreaConnections_

src/study/system-model/include/antares/study/system-model/component.h
pet-mit: renamed to portToAreaConnections()

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
guilpier-code: > it does lookup the constraint in the balance constraints map

Indeed. Oops, sorry.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
pet-mit: changes may be deep. will open a new PR

src/tests/src/solver/optimisation/test_componentToAreaConnectionFiller.cpp
pet-mit: I separated the constraints part of the code

src/study/system-model/include/antares/study/system-model/portType.h
```diff
@@ -40,11 +40,7 @@ class PortType
         id_(id),
         fields_(std::move(fields))
     {
-        if (areaConnectionFieldId.empty())
-        {
-            areaConnectionFieldId_ = std::nullopt;
-        }
-        else
+        if (!areaConnectionFieldId.empty())
```
guilpier-code, line 43: In the end, do we need the `if (!areaConnectionFieldId.empty())` ?
I mean, would it make sense to have : 
```c++
PortType(const std::string& id,
         std::vector<PortField>&& fields,
         const std::string& areaConnectionFieldId = ""):
    id_(id),
    fields_(std::move(fields))
{
    if (!std::ranges::any_of(fields_,
                             [areaConnectionFieldId](const auto& field)
                             { return field.Id() == areaConnectionFieldId; }))
    {
         throw std::invalid_argument(
           "Field \"" + areaConnectionFieldId
           + "\" selected for area connections was not defined in PortType \"" + id_
           + "\".");
    }
    areaConnectionFieldId_ = std::optional(areaConnectionFieldId);
}
```
?

pet-mit: you're right, a field ID can't be empty

pet-mit: what you suggest would throw the exception if areaConnectionFieldId="" (which is the default case, when the port type can't be connected to an area)

pet-mit: but I can merge the if statements at least

guilpier-code: ok, i misunderstood.

guilpier-code (approved):

PR #2793, opened 2025-05-05
Output variable : near price cap [ANT-3038]

guilpier-code:
Fixes ticket [ANT-3038](https://gopro-tickets.rte-france.com/browse/ANT-3038)
A small study for tests was attached to the previous ticket.

### DIFF
--- a/src/solver/variable/economy/all.h
+++ b/src/solver/variable/economy/all.h
@@ -71,6 +71,7 @@
 #include "lold.h"
 #include "loldCsr.h"
 #include "lolp.h"
+#include "nearPriceCap.h"
 #include "lolpCsr.h"
 #include "max-mrg.h"
 #include "max-mrg-csr.h"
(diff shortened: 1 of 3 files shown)


src/solver/variable/economy/nearPriceCap.h
```diff
@@ -0,0 +1,260 @@
...
+    void hourBegin(unsigned int hourInTheYear)
+    {
+        NextType::hourBegin(hourInTheYear);
+    }
+
+    void hourForEachArea(State& state, unsigned int numSpace)
+    {
+        double mrgPrice = -state.hourlyResults->CoutsMarginauxHoraires[state.hourInTheWeek];
+
+        if (mrgPrice > unsuppliedEnergyCost - threshold + eps)
```
guilpier-code: About **eps** : 
- Is it needed ?
- Is it's value suitable ?
- Is it the right way to use it (for example, we could have `- eps` instead of `+ eps`) ?

a-zakir: if presented like this  std::abs(mrgPrice-(unsuppliedEnergyCost - threshold))> eps, it becomes clear

src/solver/variable/economy/all.h
```diff
@@ -168,90 +169,93 @@ typedef                               // Prices
...
+                                                                                    // + Fixed cost
+                                                                                    // per thermal
+                                                                                    // plant detail
+                                             <NbOfDispatchedUnits // Number of Units Dispatched
+                                              <NbOfDispatchedUnitsByPlant // Number of Units
+                                                                          // Dispatched by plant
+                                               <ProfitByPlant
+                                                // Links
+                                                <Variable::Economy::Links // All links
+                                                 >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
```
a-zakir, line 189: i don't understand anything about this, i trust you :)

flomnes: It's basically a chained template, each function calling the next with the same name (convention).

src/solver/variable/economy/nearPriceCap.h
guilpier-code: What if your **marginal price** is, let's say, **100**, and your **unsupplied energy cost** is **200**, and **threshold** is **5** ?
In that case, we have : 

`std::abs(mrgPrice-(unsuppliedEnergyCost - threshold)) > eps`

but **marginal price** is very much below **unsupplied energy cost**. 
In this case the **near price cap** should remain **null**.
So I'm not sure the suggested formula works.

a-zakir: ah the sign is bot good, it should be <

src/solver/variable/economy/nearPriceCap.h
```diff
@@ -0,0 +1,243 @@
...
+              results, fileLevel, precision);
+        }
+    }
+
+private:
+    //! Intermediate values for each year
+    typename VCardType::IntermediateValuesType pValuesForTheCurrentYear;
+    unsigned int pNbYearsParallel;
+    const double eps = 1e-2;
+    const double margin = 5.;
```
flomnes: - `constexpr`
- at the top of the file, outside the class definition (no need for 1 `eps` per class instance)

guilpier-code: Done

flomnes (approved):

PR #2802, opened 2025-05-09
Binding constraints RHS : refactor to avoid repetitions

guilpier-code:
**Issue** : 
when setting the RHS of daily and weekly binding constraints, we repeat the same setting operation multiple times unnecessarily.

**What was done** : 
- [x] refactor : mainly switch loop over time step and loop over binding constraints. Try to separate responsibilities a bit more.
- [x] add several fast in memory tests for binding constraints working with must-run clusters 
- [x] add a bit of useful code for in memory tests  
- [x] make clarifications here and there


src/solver/simulation/include/antares/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -149,7 +149,6 @@ struct CONTRAINTES_COUPLANTES
 
     std::vector<double> SecondMembreDeLaContrainteCouplante;
 
-    int NombreDElementsDansLaContrainteCouplante;
```
guilpier-code, line 152: **NombreDElementsDansLaContrainteCouplante** : we dont' need this. Was removed.

JasonMarechal25 (approved):

PR #2803, opened 2025-05-12
Package OR-Tools shared libs [ANT-3227]

JasonMarechal25:
Since version 9.12 or-tools is not build as static library, instead ortools and it's dependencies are built as shared libraries.

This causes an issue where trying to execute antares-solver fail because of missing libraries. One way to locally solve this issue is to point LD_LIBRARY_PATH to ortools/install/lib directory.

With this PR we package ortools and subsequent dependencies in the assets. This way users can continue to use antares-solver as before without having to rely on an extarnal or-tools installation

### DIFF
--- a/.github/workflows/centos7.yml
+++ b/.github/workflows/centos7.yml
@@ -5,8 +5,6 @@ on:
     types: [ created ]
   push:
     branches:
-      - develop
-      - dependabot/*
   schedule:
     - cron: '21 2 * * *'
   workflow_call:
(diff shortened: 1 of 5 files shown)


flomnes:
In the "solver only" archives, paths are somewhat wrong
```
antares-solver
libsirius_solver.so
home/runner/work/Antares_Simulator/Antares_Simulator/or-tools/install/lib/libCbc.so
home/runner/work/Antares_Simulator/Antares_Simulator/or-tools/install/lib/libCbc.so.2
home/runner/work/Antares_Simulator/Antares_Simulator/or-tools/install/lib/libCbc.so.2.10.12
```
The .so files should be located in the same directory as `antares-solver`, exactly as libsirius_solver.so.

flomnes (approved):

PR #2809, opened 2025-05-19
Simplify linear optimization code

pet-mit:
The current code (weekly problem) uses classes PROBLEME_SIMPLEXE and PROBLEME_SIMPLEXE_NOMME to communicate with MPSolver.  
These classes are Sirius' representation and interface for the simplex algorithm.  
They are no longer needed since Sirius is now behind MPSolver, and since they are redundant with PROBLEME_HEBDO & PROBLEME_ANTARES_A_RESOUDRE objects.

This will then allow us to address unadressed comments in #2792


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -457,10 +393,8 @@ bool OPT_AppelDuSimplexe(const SingleOptimOptions& options,
             logs.info() << " Solver: Safe resolution failed";
         }
 
-        Probleme.SetUseNamedProblems(true);
```
guilpier-code, line 460: About  `Probleme.SetUseNamedProblems(true);` disappearance : 

We replaced this disappearance with adding a new boolean argument to **I_MPS_writer::runIfNeeded(...)**, meaning adding a new argument to every overridden **runIfNeeded(...)** function of **I_MPS_writer** child class.
Wouldn't it more simple to just replace : 
`Probleme.SetUseNamedProblems(true);`
with 
`NamedProblems->NamedProblems = true;` ?

Indeed, at that point of execution, we know we have a problem not feasible, and that we're about to stop.
**NamedProblems->NamedProblems**'s value does not seem to have any importance, so we change it at will, can't we ?

.github/workflows/ubuntu.yml
```diff
@@ -27,7 +27,7 @@ env:
   GITHUB_TOKEN: ${{ github.token }}
   IS_RELEASE: ${{ github.event_name == 'workflow_dispatch' }}
   RUN_SIMPLE_TESTS: ${{ github.event_name == 'push' || inputs.run-tests == 'true' }}
-  RUN_EXTENDED_TESTS: ${{ github.event_name == 'schedule' || inputs.run-tests == 'true' }}
```
guilpier-code: We shouldn't forget to revert this file.

.github/workflows/windows-vcpkg.yml
```diff
@@ -27,7 +27,7 @@ env:
   GITHUB_TOKEN: ${{ github.token }}
   IS_RELEASE: ${{ github.event_name == 'workflow_dispatch' }}
   RUN_SIMPLE_TESTS: ${{ github.event_name == 'push' || inputs.run-tests == 'true' }}
-  RUN_EXTENDED_TESTS: ${{ github.event_name == 'schedule' || inputs.run-tests == 'true' }}
+  RUN_EXTENDED_TESTS: ${{ github.event_name == 'push' || inputs.run-tests == 'true' }}
```
guilpier-code: To be reverted.

src/solver/optimisation/LegacyFiller.cpp
```diff
@@ -112,29 +117,27 @@ void LegacyFiller::UpdateContraints(unsigned idxRow, ILinearProblem& pb) const
...
 void LegacyFiller::CopyRows(ILinearProblem& pb) const
 {
-    for (int idxRow = 0; idxRow < problemeSimplexe_->NombreDeContraintes; ++idxRow)
+    for (int idxRow = 0; idxRow < problemeAResoudre_->NombreDeContraintes; ++idxRow)
     {
         UpdateContraints(idxRow, pb);
     }
 }
 
 std::string LegacyFiller::GetVariableName(unsigned int index) const
```
guilpier-code, line 130: I still have a problem with **GetVariableName** and **GetConstraintName** being here.
It seems they don't belong to this class. It seems more the responsibility of the problem itself.
Besides, as you well know, **GetConstraintName** is duplicated in **ComponentToAreaConnectionFiller**.

At least, instead of having :
`GetConstraintName(idxRow)`
We could have : 
**problemeAResoudre_->GetConstraintName(idxRow)**
(It would allow to remove the duplication).

But wouldn't it be possible to set the variable and constraints names (depending on we use human readable names or not) at some point,  for any use we want, either MPS printing, legacy filler or connection between areas and components ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
pet-mit: I actually think this is a non subject. fullMPSwriter seems to be unused. Do I remove the class ?

src/solver/optimisation/LegacyFiller.cpp
pet-mit: I already removed duplication from ComponentToAreaConnectionFiller

pet-mit: Actually, the names "x###" and "c###" for variables and constraints, when useNamedMps=false, are only important when **populating** the MPSolver (inside LegacyFiller).
This is done as a workaround, since MPSolver.Write does not permit obfuscating names (in opposition to ExportModelAsMpsFormat).
Names "x####" and "c####" are not important elsewhere. All other classes that act on legacy variables & constraints refer to them using (int) indices, including new fillers (I already replaced them with indices in ComponentToAreaConnectionFiller).

So, out options are:
- keep things as they are : only the LegacyFiller knows that the variables and constraints should or should not be named as in the business problem (human-readable names), on the MPSolver level.
- complexify LegacyOrtoolsLinearProblem by adding a compatibility dictionary that contains the business names and "MPSolver names" of every variable/constraint. Seen from the outside, we would only use business names and, under-the-hood, LegacyOrtoolsLinearProblem communicates with MPSolver either with business names, or with obfuscated names (using the dictionary). But this would add complexity & an overhead on the usage of the LinearProblem instance.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
guilpier-code: You're right.
And removing class **fullMPSwriter** has several consequences : 
- we no longer need **OPT_EcrireJeuDeDonneesLineaireAuFormatMPS**
- therefore we no longer need class **ProblemConverter**
- therefore we no longer need classes **NameTranslator**, **RealName**, **NullName**

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
```diff
@@ -25,18 +25,17 @@
...
 class ComponentToAreaConnectionFiller: public Optimisation::LinearProblemApi::LinearProblemFiller
 {
 public:
-    explicit ComponentToAreaConnectionFiller(const PROBLEME_SIMPLEXE_NOMME* problemeSimplexe,
-                                             unsigned int nTimestepsInProblem,
-                                             const ModelerStudy::SystemModel::System* modelerSystem,
+    explicit ComponentToAreaConnectionFiller(const PROBLEME_HEBDO* problemeHebdo,
                                              const VariableDictionary& modelerVariableDictionary);
+    std::string getLegacyConstraintName(const PROBLEME_ANTARES_A_RESOUDRE* problemeAResoudre,
+                                        unsigned index) const;
```
guilpier-code: `std::string getLegacyConstraintName(...)` undefined

guilpier-code (approved):

PR #2811, opened 2025-05-20
Fix prefix for short-term storage

flomnes:


src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -480,7 +480,7 @@ bool Rules::readLine(const AreaName::Vector& splitKey, const String& value, bool
     {
         return readBindingConstraints(splitKey, value);
     }
-    else if (kind_of_scenario == "st")
+    else if (kind_of_scenario == "sts")
```
guilpier-code, line 483: **Question** : how do we plan to call the scenario keyword for coming STS additional constraints ?

flomnes: `sta`

guilpier-code (approved):

PR #2813, opened 2025-05-20
Scenarize short-term storage additional constraints [ANT-3037]

flomnes:


src/libs/antares/study/include/antares/study/parts/short-term-storage/additionalConstraints.h
```diff
@@ -75,20 +79,22 @@ class AdditionalConstraints
...
     }
 
     const TimeSeries& rhs() const
     {
-        return rhs_;
+        return timeSeries;
     }
 
+    TimeSeriesNumbers timeseriesNumbers;
+    TimeSeries timeSeries;
```
guilpier-code, line 89: why do we change **rhs_** into **timeSeries** (whereas function that return the TS is still names **rhs()**) ?

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -30,6 +30,23 @@
 
 using namespace Yuni;
 
+namespace Internal
```
guilpier-code: I assume you defined namespace **Internal** to contain a single **reset** function that is not alone to wear this very common name.
**Internal** does not say much, and is kind of redundant with the **static** identifier in front of **reset(...)**, which says pretty much the same.
We could rename it (for instance) : **MultipleItemsByArea** (I know, it's quite longer and maybe clumsy, but more meaningful).

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -65,16 +65,16 @@ static void importShortTermStorages(
             toInsert.name = st.properties.name;
             for (const auto& constraint: st.additionalConstraints)
             {
-                if (constraint.enabled)
+                if (constraint->enabled)
                 {
                     auto newConstraint = constraint;
```
guilpier-code: `auto newConstraint = constraint;` : why do we make a copy here, even if it's a light `std::shared_ptr` copy ?
Do we need it ?

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -65,16 +65,16 @@ static void importShortTermStorages(
             toInsert.name = st.properties.name;
             for (const auto& constraint: st.additionalConstraints)
```
guilpier-code, line 66: `for (const auto& constraint: st.additionalConstraints)` : 
Instead, it could be : 
`for (const auto& additionalConstraint : st.additionalConstraints)`
so that `additionalConstraint->constraints` makes more sense than `newConstraint->constraints`

src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -30,6 +30,23 @@
...
+
+    v.clear();
+    v.resize(areaCount);
+    for (uint i = 0; i != areaCount; ++i)
+    {
+        v[i].attachArea(study.areas.byIndex[i]);
+        v[i].reset(study);
+    }
+}
+} // namespace Internal
```
flomnes: Small refactor to de-duplicate code

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -131,52 +131,52 @@ bool STStorageInput::loadAdditionalConstraints(const fs::path& parentPath)
...
-            !readRHS(rhsPath, additionalConstraints.rhs()))
+        if (const auto rhsPath = parentPath / ("rhs_" + additionalConstraints->name + ".txt");
+            !readRHS(rhsPath, additionalConstraints->rhs()))
         {
             logs.error() << "Error while reading rhs file: " << rhsPath;
             return false;
         }
 
-        if (auto [ok, error_msg] = additionalConstraints.validate(); !ok)
+        if (auto [ok, error_msg] = additionalConstraints->validate(); !ok)
```
guilpier-code: _**Cosmetic (and not exactly in the scope of this PR)**_ : 
I personally would have made a free function `void validate(const AdditionalConstraints& ac) `, this would have lightened the class **AdditionalConstraints** from private member functions and from `struct ValidateResult`, that are used only once upon reading input files.

Furthermore, we use more end more exceptions to handle problems (for instance in new modeler).
It allows to dramatically simplify the validate function return code and still keep code very clear.
As a consequence, we would no longer need `struct ValidateResult` : 
```c++
try
{
    validate(additionalConstraints);
}
catch (const std::invalid_argument& ex)
{
    logs.error() << "Invalid constraint in section: " << section->name;
    logs.error() << ex.what();
    return false;
}
```

guilpier-code:
**CAUTION** : **Antares** installed with the **Windows** installer crashes with error message :  

impossible-execution

src/libs/antares/study/parts/short-term-storage/additionalConstraints.cpp
```diff
@@ -26,7 +26,7 @@
 namespace Antares::Data::ShortTermStorage
```
guilpier-code, line 26: We should rename **enabledConstraints()** into **enabledConstraintsCount()**

src/solver/simulation/include/antares/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -182,7 +182,8 @@ struct PROPERTIES
     bool penalizeVariationInjection;
 
     std::shared_ptr<Antares::Data::ShortTermStorage::Series> series;
-    std::vector<Antares::Data::ShortTermStorage::AdditionalConstraints> additionalConstraints;
+    std::vector<std::shared_ptr<Antares::Data::ShortTermStorage::AdditionalConstraints>>
```
guilpier-code, line 186: A bit below, we use **::ShortTermStorage::**, so here (line 185), do we need **Antares::Data::** and be that thorough ?

src/tests/src/solver/optimisation/constraints/constraints_builder.cpp
```diff
@@ -51,24 +51,27 @@ struct BB
...
-                        "greater",
-                        true,
-                        addc2_injection_constraints),
-        addc3_netting("addc3_netting",
-                      "cluster_3",
-                      "netting",
-                      "equal",
-                      true,
-                      addc3_netting_constraints)
+        addc1_withdrawal(std::make_shared<Antares::Data::ShortTermStorage::AdditionalConstraints>(
```
guilpier-code: Do we need **Antares::Data::ShortTermStorage** everywhere ?
It's quite cumbersome.
Indeed, here we know we're in the context of tests about short term storage.

guilpier-code: **BB** : what does it mean ?

src/libs/antares/study/include/antares/study/parts/short-term-storage/additionalConstraints.h
flomnes: Because templated function `applyToMatrix` (ApplyToMatrix.hxx) expects an object with these two public members
- `timeSeries`
- `timeseriesNumbers`

Otherwise the template cannot be instantiated.

guilpier-code: ok, so rotten code somewhere forces us to lower the code quality somewhere else.
We could specialize **CheckValidity** of **AdditionalConstraints** as well, if we wanted.

src/libs/antares/study/scenario-builder/rules.cpp
flomnes: Yes, even at the root namespace it's good

src/solver/simulation/sim_calcul_economique.cpp
flomnes: No longer necessary, fixed by removing `newConstraint`

src/solver/simulation/sim_calcul_economique.cpp
flomnes: Yes, I don't want to change `AdditionalConstraints`'s layout

src/libs/antares/study/parts/short-term-storage/additionalConstraints.cpp
flomnes: done

src/libs/antares/study/parts/short-term-storage/container.cpp
flomnes: - Moved `validate` to free function
- Won't use exceptions

src/solver/simulation/include/antares/solver/simulation/sim_structure_probleme_economique.h
flomnes: I added the full namespace above

src/tests/src/solver/optimisation/constraints/constraints_builder.cpp
flomnes: Nothing

flomnes: done

src/libs/antares/study/include/antares/study/parts/short-term-storage/additionalConstraints.h
flomnes: Yes, it was done in this PR
```cpp
template<>
inline bool CheckValidity<ShortTermStorage::AdditionalConstraints>(
  uint value,
  const ShortTermStorage::AdditionalConstraints& ct,
  uint)
{
    return value < ct.timeSeries.numberOfColumns();
}
```

src/libs/antares/study/parts/short-term-storage/container.cpp
guilpier-code: the goal is to remove all validation function from class **AdditionalConstraints**. 
This is applicable to : 
- `bool isValidVariable() const;`
- `bool isValidOperatorType() const;`
- `bool isValidHours() const;`

src/libs/antares/study/include/antares/study/parts/short-term-storage/additionalConstraints.h
guilpier-code: You're not convinced by this suggestion. No problem.

src/libs/antares/study/parts/short-term-storage/container.cpp
flomnes: https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/2818

src/libs/antares/study/scenario-builder/rules.cpp
guilpier-code: ok

src/libs/antares/study/parts/short-term-storage/container.cpp
guilpier-code: I'm not sure this issue #2818 is useful : we'll forget it as soon as we resolve this conversation.

flomnes: > I'm not sure this issue #2818 is useful : we'll forget it as soon as we resolve this conversation.

done 62d4be573

guilpier-code (approved):

PR #2814, opened 2025-05-21
[ANT-3092] Valeur par defaut des bornes des variables

JasonMarechal25:
- Continuous or Integer variables :
	-  lower bound default value is -inf
	- upperbound default value is +inf
- Boolean variables :
	- lower bound default value is 0 (false)
	- upper bound default value is 1 (true)

### DIFF
--- a/docs/user-guide/modeler/02-inputs.md
+++ b/docs/user-guide/modeler/02-inputs.md
@@ -141,7 +141,7 @@ models:
     - **variable-type**: `continuous`, `integer`, or `binary`
     - **lower-bound** _(optional)_: an [expression](#expressions) representing the lower bound of the variable. Must use scalars
       and/or parameters only. If missing, defaults to -inf for continuous and integer types, or 0 for binary.
-    - **upper-bound**: an [expression](#expressions) representing the upper bound of the variable. Must use scalars
+    - **upper-bound** _(optional)_: an [expression](#expressions) representing the upper bound of the variable. Must use scalars
       and/or parameters only. If missing default to +inf for continuous and integer type or 1 for binary.
 - **constraints** _(optional)_: a collection of "internal" optimization constraints set by the model
     - **id**: an ID for the constraint. Must be unique inside the scope of the model, and
--- a/docs/user-guide/modeler/02-inputs.md
+++ b/docs/user-guide/modeler/02-inputs.md
@@ -142,7 +142,7 @@ models:
     - **lower-bound** _(optional)_: an [expression](#expressions) representing the lower bound of the variable. Must use scalars
       and/or parameters only. If missing, defaults to -inf for continuous and integer types, or 0 for binary.
     - **upper-bound** _(optional)_: an [expression](#expressions) representing the upper bound of the variable. Must use scalars
-      and/or parameters only. If missing default to +inf for continuous and integer type or 1 for binary.
+      and/or parameters only. If missing, defaults to +inf for continuous and integer types, or 1 for binary.
 - **constraints** _(optional)_: a collection of "internal" optimization constraints set by the model
     - **id**: an ID for the constraint. Must be unique inside the scope of the model, and
       respect [these rules](#rules-for-ids).
--- a/docs/user-guide/modeler/02-inputs.md
+++ b/docs/user-guide/modeler/02-inputs.md
@@ -139,7 +139,7 @@ models:
     - **id**: an ID for the variable. Must be unique inside the scope of the model, and
       respect [these rules](#rules-for-ids).
     - **variable-type**: `continuous`, `integer`, or `binary`
-    - **lower-bound**: an [expression](#expressions) representing the lower bound of the variable. Must use scalars
+    - **lower-bound** _(optional)_: an [expression](#expressions) representing the lower bound of the variable. Must use scalars
       and/or parameters only. If missing, defaults to -inf for continuous and integer types, or 0 for binary.
     - **upper-bound**: an [expression](#expressions) representing the upper bound of the variable. Must use scalars
       and/or parameters only. If missing default to +inf for continuous and integer type or 1 for binary.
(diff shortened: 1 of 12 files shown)


src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -163,10 +163,25 @@ void ComponentFiller::addVariables(Optimisation::LinearProblemApi::ILinearProble
...
+            return Expressions::Visitors::EvaluationResult(defaultValue);
+        }
+        return evaluator.dispatch(node.RootNode());
+    };
     for (const auto& variable: component_.getModel()->Variables() | std::views::values)
     {
-        const auto& lb = evaluator.dispatch(variable.LowerBound().RootNode());
-        const auto& ub = evaluator.dispatch(variable.UpperBound().RootNode());
+        namespace SM = ModelerStudy::SystemModel;
+        const auto& lb = valueOrDefault(variable.LowerBound(),
```
pet-mit, line 177: is it possible to do this only once, when creating the model in the library ? I feel like if we do it here, we might have an avoidable overhead during runtime, since the same operations will be done for every component

pet-mit (changes requested):
can you update the documentation please ?

src/solver/optim-model-filler/ComponentFiller.cpp
JasonMarechal25: modelConverter.cpp:convertVariables would be the best candidate but at this point we only have expression so it depends if we already support -inf/+inf expressions

docs/user-guide/modeler/02-inputs.md
```diff
@@ -108,41 +124,59 @@ models:
...
+      scenario, or is the same for all scenarios.
 - **variables** _(optional)_: a collection of optimization variables that are defined for this model
-    - **id**: an ID for the variable. Must be unique inside the scope of the model, and respect [these rules](#rules-for-ids).
-    - **variable-type**: `continuous`, `integer`, or `binary` 
-    - **lower-bound**: an [expression](#expressions) representing the lower bound of the variable. Must use scalars and/or parameters only.
-    - **upper-bound**: an [expression](#expressions) representing the upper bound of the variable. Must use scalars and/or parameters only.
+    - **id**: an ID for the variable. Must be unique inside the scope of the model, and
+      respect [these rules](#rules-for-ids).
+    - **variable-type**: `continuous`, `integer`, or `binary`
+    - **lower-bound**: an [expression](#expressions) representing the lower bound of the variable. Must use scalars
```
pet-mit:
```suggestion
    - **lower-bound** _(optional)_: an [expression](#expressions) representing the lower bound of the variable. Must use scalars
```

docs/user-guide/modeler/02-inputs.md
```diff
@@ -108,41 +124,59 @@ models:
...
+      and/or parameters only. If missing default to -inf for continuous and integer type or 0 for binary.
```
pet-mit:
```suggestion
      and/or parameters only. If missing, defaults to -inf for continuous and integer types, or 0 for binary.
```

docs/user-guide/modeler/02-inputs.md
```diff
@@ -108,41 +124,59 @@ models:
...
+    - **upper-bound**: an [expression](#expressions) representing the upper bound of the variable. Must use scalars
```
pet-mit:
```suggestion
    - **upper-bound** _(optional)_: an [expression](#expressions) representing the upper bound of the variable. Must use scalars
```

docs/user-guide/modeler/02-inputs.md
```diff
@@ -108,41 +124,59 @@ models:
...
+      and/or parameters only. If missing default to +inf for continuous and integer type or 1 for binary.
```
pet-mit:
```suggestion
      and/or parameters only. If missing, defaults to +inf for continuous and integer types, or 1 for binary.
```

src/solver/optim-model-filler/ComponentFiller.cpp
pet-mit: ok then maybe add a TODO note here.
I think a +inf expression can be useful in the future in yml files

pet-mit (changes requested):

src/solver/optim-model-filler/ComponentFiller.cpp
JasonMarechal25: https://gopro-tickets.rte-france.com/browse/ANT-3255

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -163,10 +163,25 @@ void ComponentFiller::addVariables(Optimisation::LinearProblemApi::ILinearProble
...
+                                        variable.Type() == SM::ValueType::BOOL
+                                          ? 0
+                                          : -std::numeric_limits<double>::infinity());
```
tbittar: @pet-mit  Shouldn't we define our own `infinity` ? As each solver may interpret it differently under the hood : for instance I think Xpress will put `1e20` whereas COIN something around `1e300`. This will result in different problems depending on the solvers and possibly behaviors that are solver-dependent

pet-mit: you're right.
@JasonMarechal25 you can use ILinearProblem::infinity() instead, which serves this purpose

pet-mit: FYI Google does not recommend / maintain MPSolver::solver_infinty() : https://github.com/google/or-tools/pull/4313

JasonMarechal25: We're already using infinity in a lot of other place though

pet-mit (approved):

flomnes (approved):

PR #2816, opened 2025-05-22
Remove SparseVector, remove data()[position]

flomnes:


JasonMarechal25:
Eigen provide a [Sparse vector](https://eigen.tuxfamily.org/dox/classEigen_1_1SparseVector.html), maybe it could be a good opportunity to introduce eigen ?

flomnes:
> Eigen provide a [Sparse vector](https://eigen.tuxfamily.org/dox/classEigen_1_1SparseVector.html), maybe it could be a good opportunity to introduce eigen ?

I removed `SpareVector`, since it wasn't really sparse. The only point was that it's size was unknown when starting to build the problem (number of matrix coefficients). The solution was to resize the vectors.

pet-mit (approved):

PR #2817, opened 2025-05-23
[ANT-3264] Default author is username

JasonMarechal25:
For GUI, default author name is now the local username
Add a new field "editor" with the name of last person to edit the study

### DIFF
--- a/src/libs/antares/study/header.cpp
+++ b/src/libs/antares/study/header.cpp
@@ -27,6 +27,7 @@
 
 #include <antares/logs/logs.h>
 #include "antares/study/version.h"
+#include "yuni/core/system/username.h"
 
 using namespace Yuni;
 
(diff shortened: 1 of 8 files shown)


src/libs/antares/study/header.cpp
```diff
@@ -54,7 +56,10 @@ void StudyHeader::reset()
     dateCreated = ::time(nullptr);
     dateLastSave = dateCreated;
     // Author
-    author = STUDYHEADER_DEFAULT_AUTHOR;
+    String username;
```
guilpier-code, line 59: If we wanted, we could try to get rid of `Yuni::String` s (at least **username**, **author** and **editor**), and replace them with `std::string`s.
Unfortunately, it may force us to make changes in **inifile.*** source files.
In case we want/could to do that, we can't replace `System::Username(username);` with a piece of code found [here](https://github.com/tanpero/getUserName).

guilpier-code (approved):

PR #2821, opened 2025-05-26
Force the export for criterion files in Expansion mode [ANT-3137]

flomnes:


src/solver/optimisation/opt_optimisation_lineaire.cpp
```diff
@@ -167,7 +167,8 @@ bool runWeeklyOptimization(const SingleOptimOptions& options,
             return false;
         }
 
-        if (problemeHebdo->ExportMPS != Data::mpsExportStatus::NO_EXPORT)
+        if (problemeHebdo->ExportMPS != Data::mpsExportStatus::NO_EXPORT
```
JasonMarechal25, line 170: Shouldn't there be a rule at a higher level which compute whether or not to export data like it is done here but in a more "centralized" way ?

flomnes: A class ? A function ?

JasonMarechal25: maybe just a boolean available in problem/study/options/whatever is available

payetvin (approved):

PR #2827, opened 2025-05-28
[ANT-2034] InMemory modeler test

JasonMarechal25:
Refactoring to allow in memory modeler test.

- Move _modeler-lib_ outside _modeler_ exe scope
- _modeler-lib_ now expose a _Modeler_ and some interfaces
- _modeler_ exe depends on modeler-lib for _Modeler_ and _IFileWriter_ interface
- _load-modeler-files_ depends on _modeler-libs_ for _ILoader_ interface


src/modeler/loadFiles/include/antares/solver/modeler/loadFiles/loadFiles.h
```diff
@@ -29,7 +29,7 @@
 #include <antares/study/system-model/library.h>
 #include <antares/study/system-model/system.h>
 
-#include "data.h"
+#include "../../../../../../modeler/include/antares/solver/modeler/data.h"
```
flomnes: Eliminate this relative path, it breaks our rules

src/modeler/FileWriter.cpp
flomnes, line 1: You may introduce a `NullWriter` with a factory function to avoid so many `if(!noOutput_)`

```cpp
std::unique_ptr<FileWriter> = writerFactory(path, noOutput);
```

src/study/system-model/library.cpp
```diff
@@ -89,6 +89,17 @@ LibraryBuilder& LibraryBuilder::withModels(std::vector<Model>&& models)
     return *this;
 }
 
+LibraryBuilder& LibraryBuilder::withModelsMap(std::unordered_map<std::string, Model>&& models)
+{
+    for (auto&& pair: models)
+    {
+        std::inserter(*library_.models_, library_.models_->end()) = std::make_pair(pair.first,
+                                                                                   std::move(
+                                                                                     pair.second));
```
flomnes:
```suggestion
        library_.models_->insert(std::make_pair(pair.first, std::move(pair.second)));
```

src/tests/end-to-end/modeler/test-modeler.cpp
```diff
@@ -0,0 +1,139 @@
...
+        {
+            components.emplace(component.Id(), copyComponent(component));
+        }
+    }
+
+    std::unordered_map<std::string, Antares::ModelerStudy::SystemModel::Component> components;
+    Fixture fixture;
+};
+
+class InMemoryWriter: public Antares::Solver::IWriter
```
flomnes: This is more of a `NullWriter` since it doesn't do anything.

src/tests/inmemory-modeler/inmemory-modeler.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include "antares/study/system-model/parameter.h"
+
+using namespace Antares::ModelerStudy::SystemModel;
+
+namespace Test::Modeler
+{
+auto build_context_parameter_with(const std::string& id,
+                                  const std::string& value,
+                                  const Antares::Expressions::Visitors::ParameterType& type)
+  -> std::pair<std::string, Antares::Expressions::Visitors::ParameterTypeAndValue>
```
flomnes:
```suggestion
std::pair<std::string, Antares::Expressions::Visitors::ParameterTypeAndValue> build_context_parameter_with(const std::string& id,
                                  const std::string& value,
                                  const Antares::Expressions::Visitors::ParameterType& type)
```

flomnes (commented):
Good job ! You may want to go a bit further with the `FileWriter` abstraction/implementation

src/tests/inmemory-modeler/inmemory-modeler.cpp
JasonMarechal25: Same thing

src/modeler/FileWriter.cpp
JasonMarechal25: It's a good idea.
Unfortunately with the current design we don't know if we will have a nullWriter or not at the time of instantiation.
We have access to parameters.noOutput value inside Modeler::solve().

src/tests/inmemory-modeler/inmemory-modeler.cpp
flomnes: More readable IMO

src/tests/end-to-end/modeler/test-modeler.cpp
```diff
@@ -0,0 +1,139 @@
...
+      .withScenarioGroupId(c.getScenarioGroupId())
+      .withParameterValues(c.getParameterValues())
+      .build();
+}
+
+using Models = std::unordered_map<std::string, Antares::ModelerStudy::SystemModel::Model>;
+
+template<class Fixture>
+class InMemoryLoader: public Antares::Solver::ILoader
+{
```
flomnes, line 57: This template is instanciated only with one type, considering removing template to make code easier to read

JasonMarechal25: One type so far.
For further test we may provide a different fixture.

.github/workflows/sonarcloud.yml
```diff
@@ -212,7 +212,7 @@ jobs:
           variant: "named-mps"
 
       - name: Run fast cucumber
-        if: ${{ env.RUN_SIMPLE_TESTS == 'true' && !cancelled() }}
+        if: ${{ !cancelled() }}
```
flomnes, line 215: revert ?

src/modeler/FileWriter.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void FileWriter::writeSolution(
+  const Optimisation::LinearProblemMpsolverImpl::OrtoolsMipSolution& solution)
+{
+    if (output)
+    {
+        logs.info() << "Writing objective & variable values...";
+        std::ofstream sol_out(outputPath_ / "solution.csv");
+        sol_out << std::setprecision(15) << "objective " << solution.getObjectiveValue()
```
flomnes, line 56: Slightly outside the scope of this PR, but we may have to specialize output file names for scenarios/weeks, to prevent solution files from being overwritten when there are multiple weeks and/or scenarios.

src/tests/inmemory-modeler/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+add_library(Antares::tests::inmemory-modeler ALIAS inmemory-modeler)
+target_sources(inmemory-modeler
+        PRIVATE
+    inmemory-modeler.cpp
+    include/inmemory-modeler.h
+)
+
+target_include_directories(inmemory-modeler
+        PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}/include
```
flomnes:
```suggestion
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
```

flomnes: That's how it's done for all other existing targets. The current version doesn't work well in "install" mode, which is used by antares-xpansion.

src/tests/inmemory-modeler/include/inmemory-modeler.h
```diff
@@ -0,0 +1,167 @@
...
+{
+class ComponentFiller;
+}
+
+namespace Antares::Optimisation::LinearProblemApi
+{
+class ILinearProblem;
+class FillContext;
+class LinearProblemFiller;
+} // namespace Antares::Optimisation::LinearProblemApi
```
flomnes: Those forward declarations are probably not needed, use declaration from headers when possible (special case: cyclic dependencies)

src/study/system-model/library.cpp
```diff
@@ -89,6 +89,15 @@ LibraryBuilder& LibraryBuilder::withModels(std::vector<Model>&& models)
     return *this;
 }
 
+LibraryBuilder& LibraryBuilder::withModelsMap(std::unordered_map<std::string, Model>&& models)
+{
+    for (auto&& pair: models)
+    {
+        library_.models_->insert(std::make_pair(pair.first, std::move(pair.second)));
```
flomnes:
```suggestion
        library_.models_->insert(std::move(pair));
```
seems to work as well considering function signatures (`unordered_map&&`, `auto&& pair`)

.github/workflows/sonarcloud.yml
pet-mit: no I think this was an error, these tests weren't executed

src/modeler/FileWriter.cpp
pet-mit: yes. I agree that should be outside the scope of this PR

.github/workflows/sonarcloud.yml
flomnes: Since then you tagged some tests as `@fast`, maybe that solved the issue ?

pet-mit: this workflow does not have a "RUN_SIMPLE_TESTS" variable defined. and this step is the only one to have such a condition in this file

flomnes (approved):

PR #2829, opened 2025-05-30
feat: Multiple Hydro Rule Curves

nikolaredstork:
This PR implements the Multiple Rule Curves for Hydro Reservoirs as an optional new feature, RTE-I's CR24.


flomnes:
The coverage is a bit under 60%, can you please increase the coverage on hydrorulecurves.cpp ?
image

src/solver/hydro/management/HydroInputsChecker.cpp
```diff
@@ -94,6 +98,50 @@ bool HydroInputsChecker::checkMinGeneration(uint year)
...
+          if ((tsIndexMin == tsIndexAvg) && (tsIndexAvg == tsIndexMax))
+          {
+              tsIndex = tsIndexMin;
+          }
+          else
+          {
+              errorCollector_(area.name)
+                << "Reservoir levels Time-Series indexes in area: " << area.id
+                << " for year: " << year << " are not equal. Something went wrong!";
+          }
```
flomnes: This check is both too strict and not useful
- Too strict : the user may want to have one column for either min, avg or max, this is usually valid since there is no ambiguity on what TS to chose since there is only one
- Not useful since `ruleCurves.min`, `ruleCurves.avg` and `ruleCurves.max` share the same `timeseriesNumbers`

Checks on the number of columns of co-scenarized TS are handled by `TimeSeriesNumbers::registerSeries` and `TimeSeriesNumbers::checkSeriesNumberOfColumnsConsistency`, so you don't need to do it yourself since you already added calls to `TimeSeriesNumbers::registerSeries`.

I suggest you remove this check.

src/tests/src/libs/antares/study/parts/hydro/test-hydro-series.cpp
```diff
@@ -121,8 +168,8 @@ BOOST_FIXTURE_TEST_CASE(Testing_load_power_credits_matrices_equal_width, Fixture
     maxHourlyGenPower.reset(3, HOURS_PER_YEAR);
     maxHourlyPumpPower.reset(3, HOURS_PER_YEAR);
 
-    fillTimeSeriesWithSpecialEnds(maxHourlyGenPower, 400.);
-    fillTimeSeriesWithSpecialEnds(maxHourlyPumpPower, 200.);
+    fillTimeSeriesWithSpecialEnds(maxHourlyGenPower, 401., 402.);
+    fillTimeSeriesWithSpecialEnds(maxHourlyPumpPower, 201., 202.);
```
flomnes, line 160: Why did you have to change these values ?

src/libs/antares/study/version.cpp
```diff
@@ -57,6 +57,12 @@ StudyVersion legacyVersionIntToVersion(unsigned version)
     // The list should remain ordered in the reverse order for performance reasons
     switch (version)
     {
+    case 920:
+        return StudyVersion(9, 2);
+    case 910:
+        return StudyVersion(9, 1);
+    case 900:
+        return StudyVersion(9, 0);
```
flomnes: The format changed from version 9.0, we no longer accept 900, 910, 920, etc.

We did that to avoid having problems for versions 10.x

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
```diff
@@ -0,0 +1,216 @@
...
+                                                                3,
+                                                                DAYS_PER_YEAR,
+                                                                Matrix<>::optFixedSize,
+                                                                &fileContent)
+              && ret;
+
+        if (enabledModeIsChanged)
+        {
+            JIT::enabled = true; // Back to the previous loading mode.
+        }
```
flomnes: I have a feeling we could use the `StandardRuleCurveLoader` here (code duplication)
```cpp
if (!usedBySolver)  hydroRuleCurves = single;
```

src/tests/src/libs/antares/study/parts/hydro/test-hydro-series.cpp
nikolaredstork: I wanted to reuse this function for filling up rule curves TS. Since `fillTimeSeriesWithSpecialEnds` was coded in the way that you need to parse float value and the function will add 1.0 to the parsed value and put it at the beginning of the TS and also it will add 2.0 to the same value and put it at the end of TS.

Rule Curves represent values between 0-1, so we couldn't reuse that function in previous implementation, so I've implemented in the way that I can now parse two values as arguments and put it at the beginning and at the end of the TS. 

Overall, only reason why I changed implementation of the function is to reuse it for filling up rule curve TS.

src/libs/antares/study/version.cpp
nikolaredstork: We need it for our Antares Web deployment. I will exclude it.

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
nikolaredstork: No we cannot, because this part of the code is for loading data when entering UI, same thing as for CR23. When we fully deprecate old UI, this part of the code can be excluded. Also when we fully deprecate old UI, `standardRuleCurvesGUI` class member can be also excluded and just use local variable to load data from standard rule curve file.

src/solver/hydro/management/HydroInputsChecker.cpp
nikolaredstork: Done!

src/libs/antares/study/parameters.cpp
```diff
@@ -235,6 +235,40 @@ bool StringToCompatibilityHydroPmax(Parameters::Compatibility::HydroPmax& mode,
...
+    }
+}
+
+bool StringToCompatibilityHydroRuleCurves(Parameters::Compatibility::HydroRuleCurves& mode,
+                                          const std::string& text)
+{
+    if (text.empty())
+    {
+        return false;
+    }
```
guilpier-code:
```c++
if (text.empty())
{
    return false;
}
```
is useless : if we remove this block, behavior does not change.

guilpier-code: Same for **StringToCompatibilityHydroPmax(...)**

src/libs/antares/study/include/antares/study/parameters.h
```diff
@@ -363,6 +363,12 @@ class Parameters final
             Hourly
         };
         HydroPmax hydroPmax = HydroPmax::Daily;
+        enum class HydroRuleCurves
+        {
+            Single,
+            Scenarized
+        };
+        HydroRuleCurves hydroRuleCurves = HydroRuleCurves::Single;
     };
```
guilpier-code, line 346: _**Personal remark**_ (not mandatory to apply it)

I personally don't like struct / class defined inside other struct / class declaration :  it makes surrounding class declaration bigger than it should.
Class Parameters is already very big (and should be split, by the way) and we inflate it we struct declaration.
A class declaration is a kind of summary of the class,  it should be as short as possible, so we can see at a glance what it does.
If, like me, you don't like big classes, help moving inner struct declarations outside the surrounding class.

src/libs/antares/study/include/antares/study/parameters.h
```diff
@@ -363,6 +363,12 @@ class Parameters final
             Hourly
         };
         HydroPmax hydroPmax = HydroPmax::Daily;
+        enum class HydroRuleCurves
+        {
+            Single,
+            Scenarized
+        };
+        HydroRuleCurves hydroRuleCurves = HydroRuleCurves::Single;
```
guilpier-code, line 345: If I understand correctly, either rules curves are scenarized, or it's not.
So why not a simple boolean for this information  ?
```c++
bool hydroRuleCurvesScenarized = false; // Default value
```

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
```diff
@@ -0,0 +1,179 @@
...
+**
+** Antares_Simulator is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#ifndef __ANTARES_LIBS_STUDY_PARTS_HYDRO_RESERVOIR_LEVELS_H__
```
guilpier-code: Please replace : 
```c++
#ifndef __ANTARES_LIBS_STUDY_PARTS_HYDRO_RESERVOIR_LEVELS_H__
#define __ANTARES_LIBS_STUDY_PARTS_HYDRO_RESERVOIR_LEVELS_H__
...
#endif
```
with just : 
```c++
#pragma once
```

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
```diff
@@ -0,0 +1,179 @@
...
+#define __ANTARES_LIBS_STUDY_PARTS_HYDRO_RESERVOIR_LEVELS_H__
+
+#include <antares/array/matrix.h>
+#include <antares/series/series.h>
+#include <antares/study/parameters.h>
+#include <antares/study/version.h>
+
+#include "../../fwd.h"
```
guilpier-code: We don't need to include these headers :
```c++
#include <antares/array/matrix.h>
#include <antares/study/version.h>
#include "../../fwd.h"
```
It slows down compilation time and makes code less simple than it could be.
Please remove them.

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
```diff
@@ -0,0 +1,179 @@
...
+
+namespace Antares
```
guilpier-code: Please concatenate namespaces : 
```c++
namespace Antares::Data
{
   ...
}
```

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
```diff
@@ -0,0 +1,179 @@
...
+/*!
+** \brief Rule Curves class that support reservoir levels data
+*/
+
+class RuleCurves
+{
+public:
+    enum
+    {
+        //! The minimum value
```
guilpier-code: Comments here bring nothing more than the **enum** items, we should remove them

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
```diff
@@ -0,0 +1,146 @@
...
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+
+#include <algorithm>
+
+#include <yuni/yuni.h>
```
guilpier-code: It seems we don't need all of these header inclusions : 
```c++
#include <algorithm>

#include <yuni/yuni.h>
#include <yuni/io/file.h>
#include <antares/exception/LoadingError.hpp>
#include <antares/inifile/inifile.h>
```
Please remove them if I'm right

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include <yuni/io/file.h>
+
+#include <antares/exception/LoadingError.hpp>
+#include <antares/inifile/inifile.h>
+#include <antares/logs/logs.h>
+#include <antares/study/parts/hydro/hydrorulecurves.h>
+#include "antares/study/study.h"
+
+namespace fs = std::filesystem;
```
guilpier-code, line 24: You specified : 
```c++
namespace fs = std::filesystem;
```
and yet barely use it : there are still many **std::filesystem::** in this source file.

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
```diff
@@ -0,0 +1,146 @@
...
+        throw std::invalid_argument("Value not supported for hydro rule curves compatibility");
+    }
+}
+
+bool RuleCurvesLoaderService::LoadFromFolder(
+  const std::string& areaID,
+  const std::filesystem::path& folder,
+  Parameters::Compatibility::HydroRuleCurves hydroRuleCurves)
+{
+    bool ret = true;
```
guilpier-code:
```c++
bool ret = true;
```
is useless here.
```c++
...
return loader->load();
```
is enough.

src/libs/antares/study/parameters.cpp
guilpier-code: Done

src/libs/antares/study/include/antares/study/parameters.h
nikolaredstork: This was proposed in previous implementation, if refactoring is necessary, it should be done in different PR.

src/libs/antares/study/parameters.cpp
nikolaredstork: This was proposed in previous implementation, if refactoring is necessary, it should be done in different PR.

src/libs/antares/study/include/antares/study/parameters.h
nikolaredstork: This was proposed in previous implementation, if refactoring is necessary, it should be done in different PR.

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
guilpier-code: Done

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
guilpier-code: Done

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
guilpier-code: Done

src/libs/antares/study/parameters.cpp
guilpier-code: Done in this [(corrective) PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3310)

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
guilpier-code: Done in this [(corrective) PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3310)

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
guilpier-code: Done in this [(corrective) PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3310)

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
guilpier-code: Done in this [(corrective) PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3310)

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
guilpier-code: Done in this [(corrective) PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3310)

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
guilpier-code: Done in this [(corrective) PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3310)

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+bool ScenarizedRuleCurvesLoader::load()
+{
+    const std::filesystem::path path = baseFolder_ / "series" / areaID_;
+    Matrix<>::BufferType fileContent;
+
+    bool ret = true;
+
+    fs::path filePath = path / "maxDailyReservoirLevels.txt";
+    ret &= max_.timeSeries.loadFromCSVFile(filePath.string(), 1, DAYS_PER_YEAR, &fileContent);
```
guilpier-code: Here operator `&=` is used.
This operator seems more associated to **bitwise AND** operator than to **logical AND** operator.
If a `&&=` operator existed, we would like to use it here, but it does not seem to exist.
So I guess `ret = expr && ret` should be used instead.

src/libs/antares/study/parts/hydro/hydrorulecurves.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+    max.averageTimeseries();
+    min.averageTimeseries();
+    avg.averageTimeseries();
+}
+
+bool ScenarizedRuleCurvesLoader::load()
+{
+    const std::filesystem::path path = baseFolder_ / "series" / areaID_;
+    Matrix<>::BufferType fileContent;
```
guilpier-code: **_Out of this PR : simple observation_**
```c++
Matrix<>::BufferType fileContent;
```
OMG : this variable is requested as argument of the loading function, but is never used in the calling scope. 
Why then pass it as argument ? 
It seems it does not need to be passed, and function signature should be simplified. 
To be looked at, some day (very soon), when legacy GUI is removed as planned.

nikolaredstork: This `Matrix<>::BufferType fileContent` is USED in all loading functions, I've just went with your approach.

For example:

image

guilpier-code: Yes, that's why my remark was tagged "out of this PR".
But still the question remains : why do we pass an argument to a function, here an empty string buffer, if it's not to do something with it after the call ?

guilpier-code: Let me be more clear : we need the string buffer in the body of the function, internally. That I don't deny.
But it does not seem to be used ever in the client side (the function caller). 
So this would tend to prove that this buffer should be defined as a local variable inside the loading function, and disappear from the list of its argument.
So do we ever do something with the argument on the client side ? 
I admit I did not check.

nikolaredstork: Totally agree. 
Function `internalLoadCSVFile` from **matrix.hxx** file should be reviewed:
```
template<class T, class ReadWriteT>
bool Matrix<T, ReadWriteT>::internalLoadCSVFile(const AnyString& filename,
                                                uint minWidth,
                                                uint maxHeight,
                                                uint options,
                                                BufferType* buffer) 
```

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
```diff
@@ -0,0 +1,139 @@
...
+                             TimeSeries& max,
+                             TimeSeries& avg,
+                             TimeSeries& min):
+        RuleCurvesLoader(baseFolder, areaID, max, avg, min)
+
+    {
+    }
+
+private:
+    bool load() final;
```
flomnes:
```suggestion
    bool load() override final;
```

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
```diff
@@ -0,0 +1,139 @@
...
+{
+public:
+    ScenarizedRuleCurvesLoader(const std::filesystem::path& baseFolder,
+                               const std::string& areaID,
+                               TimeSeries& max,
+                               TimeSeries& avg,
+                               TimeSeries& min):
+        RuleCurvesLoader(baseFolder, areaID, max, avg, min)
+    {
+    }
```
flomnes: Simplification - use constructor delegation
```suggestion
    using RuleCurvesLoader::RuleCurvesLoader;
```

src/libs/antares/study/include/antares/study/parts/hydro/hydrorulecurves.h
```diff
@@ -0,0 +1,139 @@
...
+public:
+    StandardRuleCurvesLoader(const std::filesystem::path& baseFolder,
+                             const std::string& areaID,
+                             TimeSeries& max,
+                             TimeSeries& avg,
+                             TimeSeries& min):
+        RuleCurvesLoader(baseFolder, areaID, max, avg, min)
+
+    {
+    }
```
flomnes: Use constructor delegation - see below

docs/user-guide/04-migration-guides.md
```diff
@@ -7,6 +7,24 @@ lower the costs of changing existing interfaces, both GUI and scripts.
 
 ### Input
 
+#### New optional feature and compatibility flag for scenarized hydro reservoir levels
```
flomnes, line 10: planned for 10.1.0

src/libs/antares/study/include/antares/study/parameters.h
```diff
@@ -468,6 +474,10 @@ bool StringToSimulationMode(SimulationMode& mode, Yuni::CString<20, false> text)
 const char* CompatibilityHydroPmaxToCString(const Parameters::Compatibility::HydroPmax);
 bool StringToCompatibilityHydroPmax(Parameters::Compatibility::HydroPmax&, const std::string& text);
 
+const char* CompatibilityHydroRuleCurvesToCString(
+  const Parameters::Compatibility::HydroRuleCurves mode);
```
flomnes: unnecessary (no more save functions)

src/libs/antares/study/parameters.cpp
```diff
@@ -202,6 +198,36 @@ bool StringToCompatibilityHydroPmax(Parameters::Compatibility::HydroPmax& mode,
...
+    switch (mode)
+    {
+    case Parameters::Compatibility::HydroRuleCurves::Single:
+        return "single";
+    case Parameters::Compatibility::HydroRuleCurves::Scenarized:
+        return "scenarized";
+    default:
+        return "Unknown";
+    }
+}
```
flomnes: unnecessary (no more save functions)

src/solver/hydro/management/daily.cpp
```diff
@@ -218,7 +218,8 @@ inline void HydroManagement::prepareDailyOptimalGenerations(
 
     double reservoirCapacity = area.hydro.reservoirCapacity;
 
-    auto& lowLevel = area.hydro.reservoirLevel[Data::PartHydro::minimum];
+    uint32_t seriesIndex = area.hydro.series->ruleCurves.min.getSeriesIndex(y);
+    const auto& lowLevel = area.hydro.series->ruleCurves.min.timeSeries[seriesIndex];
```
flomnes, line 222:
```suggestion
    const auto* lowLevel = area.hydro.series->ruleCurves.min.timeSeries[seriesIndex];
```

src/solver/hydro/management/HydroInputsChecker.cpp
```diff
@@ -77,6 +81,35 @@ bool HydroInputsChecker::checkMinGeneration(uint year)
...
 
+bool HydroInputsChecker::checkRuleCurves(uint year)
+{
+    bool ret = true;
+    areas_.each(
+      [this, &ret, &year](const Data::Area& area)
+      {
+          const auto& minRuleCurves = area.hydro.series->ruleCurves.min.getColumn(year);
+          const auto& avgRuleCurves = area.hydro.series->ruleCurves.avg.getColumn(year);
+          const auto& maxRuleCurves = area.hydro.series->ruleCurves.max.getColumn(year);
```
```suggestion
          const auto* minRuleCurves = area.hydro.series->ruleCurves.min.getColumn(year);
          const auto* avgRuleCurves = area.hydro.series->ruleCurves.avg.getColumn(year);
          const auto* maxRuleCurves = area.hydro.series->ruleCurves.max.getColumn(year);
```

src/solver/hydro/management/monthly.cpp
```diff
@@ -256,12 +255,12 @@ void HydroManagement::prepareMonthlyOptimalGenerations(
 
           else
           {
-              auto& reservoirLevel = area.hydro.reservoirLevel[Data::PartHydro::average];
+              const auto& ruleCurves = area.hydro.series->ruleCurves.avg.getColumn(y);
```
flomnes:
```suggestion
              const auto* ruleCurves = area.hydro.series->ruleCurves.avg.getColumn(y);
```

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -711,7 +711,9 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
                                 const uint nextWeekFirstDay = study.calendar
                                                                 .hours[PasDeTempsDebut + 7 * 24]
                                                                 .dayYear;
-                                auto& minLvl = area.hydro.reservoirLevel[Data::PartHydro::minimum];
+
+                                const auto& minLvl = area.hydro.series->ruleCurves.min.getColumn(
```
flomnes, line 715:
```suggestion
                                const auto* minLvl = area.hydro.series->ruleCurves.min.getColumn(
```

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -856,8 +858,9 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
                                     const uint nextWeekFirstDay = study.calendar
                                                                     .hours[PasDeTempsDebut + 7 * 24]
                                                                     .dayYear;
-                                    auto& maxLvl = area.hydro
-                                                     .reservoirLevel[Data::PartHydro::maximum];
+
+                                    const auto& maxLvl = area.hydro.series->ruleCurves.max
```
flomnes:
```suggestion
                                    const auto* maxLvl = area.hydro.series->ruleCurves.max
```

dusanparipovic:
Hi @flomnes I'll take a look asap.

flomnes:
@dusanparipovic I think I fixed most of the issues, the PR can finally be merged

dusanparipovic:
> @dusanparipovic I think I fixed most of the issues, the PR can finally be merged

If you want anything additional ping me, I'm looking for contribution on simulator :) @flomnes

JasonMarechal25 (approved):

PR #2819, opened 2025-05-23
Hydro monthly heuristic : add a new overflow optimization variable [ANT-3236]

guilpier-code:
Attempt to fix ticket [ANT-3236](https://gopro-tickets.rte-france.com/browse/ANT-3236), but for branch **8.8.x**

PR #2837, opened 2025-06-05
Hydro monthly heuristic : add variable overflow to problem, onto v9.3.0-rc1 [ANT-3236]

guilpier-code:
Attempt to fix ticket [ANT-3236](https://gopro-tickets.rte-france.com/browse/ANT-3236) for branch **v9.3.0-rc1**

**Note** :
This work was also done for branche **8.8.x** at[ this location](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2819)

### DIFF
--- a/src/solver/hydro/management/daily.cpp
+++ b/src/solver/hydro/management/daily.cpp
@@ -149,9 +149,8 @@ struct DebugData
                              uint y,
                              const Data::AreaName& areaName) const
     {
-        std::ostringstream filename;
-        filename << "daily." << areaName.c_str() << ".txt";
-        auto path = fs::path("debug") / "solver" / std::to_string(1 + y) / filename.str();
+        auto path = fs::path("debug") / "solver" / std::to_string(1 + y)
+                    / ("daily." + areaName + ".txt").c_str();
 
         std::ostringstream fileContent;
         fileContent << "\tNiveau init : " << hydro_specific.monthly[initReservoirLvlMonth].MOL
--- a/src/solver/hydro/management/daily.cpp
+++ b/src/solver/hydro/management/daily.cpp
@@ -149,11 +149,13 @@ struct DebugData
                              uint y,
                              const Data::AreaName& areaName) const
     {
-        std::ostringstream buffer;
-        auto path = fs::path("debug") / "solver" / std::to_string(1 + y) / "daily."
-                    / areaName.c_str() / ".txt";
+        std::ostringstream filename;
+        filename << "daily." << areaName.c_str() << ".txt";
+        auto path = fs::path("debug") / "solver" / std::to_string(1 + y) / filename.str();
 
-        buffer << "\tNiveau init : " << hydro_specific.monthly[initReservoirLvlMonth].MOL << "\n";
+        std::ostringstream fileContent;
+        fileContent << "\tNiveau init : " << hydro_specific.monthly[initReservoirLvlMonth].MOL
+                    << "\n";
         for (uint month = 0; month != MONTHS_PER_YEAR; ++month)
         {
             uint realmonth = (initReservoirLvlMonth + month) % MONTHS_PER_YEAR;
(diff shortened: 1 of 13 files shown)


src/solver/hydro/management/daily.cpp
```diff
@@ -149,11 +149,13 @@ struct DebugData
                              uint y,
                              const Data::AreaName& areaName) const
     {
-        std::ostringstream buffer;
-        auto path = fs::path("debug") / "solver" / std::to_string(1 + y) / "daily."
-                    / areaName.c_str() / ".txt";
+        std::ostringstream filename;
+        filename << "daily." << areaName.c_str() << ".txt";
+        auto path = fs::path("debug") / "solver" / std::to_string(1 + y) / filename.str();
```
a-zakir:
```suggestion
        auto path = fs::path("debug") / "solver" / std::to_string(1 + y) / "daily." / (areaName+=".txt").c_str();
```

src/solver/hydro/management/monthly.cpp
```diff
@@ -76,35 +77,40 @@ static void CheckHydroAllocationProblem(Data::Area& area,
...
     }
+    return yearInflows;
+}
+
+static std::array<double, 12> computeMonthlyTargetGenerations(
+  Area& area,
+  AreaDependantHydroManagementData& data,
+  std::array<MonthlyGenerationTargetData, 12>& areaMonthlyData)
+{
+    std::array<double, 12> MTG{0};
```
a-zakir, line 95: 1 - creation of the array

src/solver/hydro/management/monthly.cpp
```diff
@@ -76,35 +77,40 @@ static void CheckHydroAllocationProblem(Data::Area& area,
...
     if (not area.hydro.followLoadModulations)
     {
         for (uint realmonth = 0; realmonth != MONTHS_PER_YEAR; ++realmonth)
         {
-            hydro_specific.monthly[realmonth].MTG = data.inflows[realmonth];
+            MTG[realmonth] = data.inflows[realmonth];
         }
-
-        return total;
+        return MTG;
```
a-zakir, line 104: 2 - create another MTG + copy

src/solver/hydro/management/monthly.cpp
```diff
@@ -194,19 +197,20 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
 
           double solutionCost = 0.;
           double solutionCostNoised = 0.;
+          std::array<double, 12> MTG{0};
```
a-zakir, line 200: 3 - same

src/solver/hydro/management/monthly.cpp
```diff
@@ -220,21 +224,13 @@ void HydroManagement::prepareMonthlyOptimalGenerations(const double* random_rese
                                                                      calendar_,
                                                                      scratchmap);
 
-                  yearlyMaxGeneration += problem.TurbineMax[month];
-
                   problem.TurbineMin[month] = data.mingens[realmonth];
-                  problem.TurbineCible[month] = hydro_specific.monthly[realmonth].MTG;
+                  problem.TurbineCible[month] = MTG[realmonth];
```
a-zakir, line 228: the original struct contains one single instance of MTG

src/solver/hydro/management/daily.cpp
flomnes: I don't like `areaName += ".txt"`, I prefer `areaName + ".txt"`, having multiple instructions on a single line is error prone. Also, `areaName` might be re-used afterwards

guilpier-code: Done (see next push)

src/solver/hydro/management/monthly.cpp
guilpier-code: Not exactly, indeed see [this page](https://artificial-mind.net/blog/2021/10/23/return-moves).

**Excerpt** : 
_This is a classical “return-by-value” and (wrongfully) associated with copies and overhead._

_In many cases, this will actually move the result instead of copying it. For modern C++, one could even argue that this will move in most cases (or, as we will see, completely elide the copy and directly construct in the result memory)._

src/solver/hydro/management/monthly.cpp
guilpier-code: You're right, but the **MTG** array had nothing to do  inside the `hydro_specific.monthly[i]` struct.
Indeed, this latter structure exists because it stores data computed somewhere and used somewhere else.
On the contrary **MTG** is computed in a place and used right after. It's kind of a local variable.
So I removed it from `hydro_specific.monthly[realmonth]`

src/solver/hydro/management/daily.cpp
a-zakir: I didn’t find operator+ in Yuni::CString

src/solver/hydro/management/monthly.cpp
guilpier-code: I see your point

src/solver/hydro/management/daily.cpp
a-zakir: cool then

src/solver/hydro/management/monthly.cpp
a-zakir: This is compiler dependent not what C++ standard says

src/solver/hydro/management/daily.cpp
guilpier-code: What I did : 
```c++
auto path = fs::path("debug") / "solver" / std::to_string(1 + y)
        / ("monthly." + area.name + ".txt").c_str();
```
**Yuni** can be powerful sometimes

src/solver/hydro/management/monthly.cpp
guilpier-code: So, in the end, making **MTG** local and private to function **prepareMonthlyOptimalGenerations** leads to a small overhead : we're forced to create 2 (small) temporary arrays **MTG**, one more temporary that the other.

src/solver/hydro/management/monthly.cpp
a-zakir: The refacto is nice and if it not come with significant cost I would keep it as it.

src/solver/hydro/management/monthly.cpp
guilpier-code: This is known as **RVO**, if there are (on earth) compilers not using **RVO**, it's a chance we don't use them :-)

src/solver/hydro/management/monthly.cpp
guilpier-code: Is this significant ? 12 temporary doubles

guilpier-code: ok, in order to handle this, we've created an [associated ticket](https://gopro-tickets.rte-france.com/browse/ANT-3347)

flomnes (approved):

PR #2868, opened 2025-06-17
Rebuild MPSolver between solves & transmit basis when using sirius [ANT-1287]

pet-mit:
Concerns linear optimization:
- Rebuild MPSolver instance for each solve. This allows handling different weekly data in hybrid studies
- When using Sirius, set starting LP basis when possible
- Also, updated docs


payetvin:
One side effect of this change are the logs, there's a lot more outputs than before for both sirius and xpress.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -213,43 +213,14 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
...
-                                      ProblemeAResoudre->NombreDeVariables);
-
-            updateMeasure.tick();
-            timeMeasure.updateTime = updateMeasure.duration_ms();
-            optimizationStatistics.addUpdateTime(timeMeasure.updateTime);
-        }
-    }
+    solver = nullptr;
+
+    solver = convertToMPSolver(problemeHebdo, NumIntervalle, options);
```
```suggestion
    solver = convertToMPSolver(problemeHebdo, NumIntervalle, options);
```

src/tests/cucumber/features/steps/common_steps/solver_output_handler.py
```diff
@@ -66,6 +66,10 @@ def __if_none_then_parse(self, rs: result_type, area, year, file_name: str):
     def __get_values_hourly(self, area: str, year: int):
         return self.__if_none_then_parse(result_type.VALUES, area.lower(), year, "values-hourly.txt")
 
+    def __get_values_hourly_for_specific_week(self, area: str, year: int, week: int):
+        df = self.__if_none_then_parse(result_type.VALUES, area.lower(), year, "values-hourly.txt")
+        return df[(df['hourly']['Unnamed: 1_level_1'] > (week - 1) * 168) & (df['hourly']['Unnamed: 1_level_1'] <= week * 168)]
+
```
flomnes, line 72: 1. You can use file values-weekly.txt for that
2. I"m not sure I understand the special value `Unnamed: 1_level_1`

src/tests/cucumber/features/solver-features/hybrid_studies.feature
```diff
@@ -58,4 +58,18 @@ Feature: hybrid (simulator+modeler) studies
...
+    # copy of test 3_6_1, both weeks are clones
+    # except max_p of the generator component is 6200 in first week, 5900 in second
+    Given the solver study path is "Antares_Simulator_Tests_NR/hybrid/3_6_4"
+    When I run antares simulator
+    Then the simulation succeeds
+    And the simulation takes less than 5 seconds
+    # for now, modeler costs does not figure in system cost txt
+    And the annual system cost is 520000
+    And in area "AREA", during year 1, week 1, loss of load lasts 0 hours
+    And in area "AREA", during year 1, week 2, loss of load lasts 1 hours
```
flomnes, line 74: In my opinion you should compare the UNSP ENRG (volume of unsupplied energy), not the LOLD. The LOLD is too sensitive to equivalent solutions.

flomnes (changes requested):

src/tests/cucumber/features/solver-features/hybrid_studies.feature
pet-mit: it can't depend on equivalent solutions here, because load > Pmax, and there is no storage

src/tests/cucumber/features/steps/common_steps/solver_output_handler.py
pet-mit: ok but i think it should be done in another PR, to refactor these python scripts. the Unnamed level is because our output files contain empty column names

src/tests/cucumber/features/solver-features/hybrid_studies.feature
flomnes: For this specific example, OK. Generally speaking, prefer UNSP ENRG

flomnes (approved):

PR #2869, opened 2025-06-17
Add STS additional constraints to infeasible problem analysis [ANT-3386]

flomnes:


src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -159,17 +164,22 @@ std::string HydroProduction::infeasibilityCause()
...
+      {"::weekly::", std::make_unique<WeeklyBC, const std::string&, double>},
+      {"^FictiveLoads::", std::make_unique<FictitiousLoad, const std::string&, double>},
+      {"^AreaHydroLevel::", std::make_unique<HydroLevel, const std::string&, double>},
+      {"^Level::", std::make_unique<STS, const std::string&, double>},
+      {"^HydroPower::", std::make_unique<HydroProduction, const std::string&, double>},
+      {"^WithdrawalSum::", std::make_unique<STSWithdrawalSum, const std::string&, double>},
+      {"^InjectionSum::", std::make_unique<STSInjectionSum, const std::string&, double>},
+      {"^NettingSum::",
+       std::make_unique<STSNettingSum, const std::string&, double>}}; // namespace
+                                                                      // Antares::Optimization
```
guilpier-code: We should remove comments at end of line 177. It seems useless and causes clang to add a new line.

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -159,17 +164,22 @@ std::string HydroProduction::infeasibilityCause()
...
 }
 
 std::unique_ptr<WatchedConstraint> ConstraintsFactory::create(const std::string& name,
-                                                              const double value) const
+                                                              double value) const
```
guilpier-code, line 190: Here you removed a **const**. Is it on purpose ?

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -159,17 +164,22 @@ std::string HydroProduction::infeasibilityCause()
...
-      {"^HydroPower::", std::make_unique<HydroProduction, const std::string&, const double>}};
+      {"::hourly::", std::make_unique<HourlyBC, const std::string&, double>},
+      {"::daily::", std::make_unique<DailyBC, const std::string&, double>},
+      {"::weekly::", std::make_unique<WeeklyBC, const std::string&, double>},
+      {"^FictiveLoads::", std::make_unique<FictitiousLoad, const std::string&, double>},
+      {"^AreaHydroLevel::", std::make_unique<HydroLevel, const std::string&, double>},
+      {"^Level::", std::make_unique<STS, const std::string&, double>},
+      {"^HydroPower::", std::make_unique<HydroProduction, const std::string&, double>},
+      {"^WithdrawalSum::", std::make_unique<STSWithdrawalSum, const std::string&, double>},
+      {"^InjectionSum::", std::make_unique<STSInjectionSum, const std::string&, double>},
```
a-zakir: while this syntax is functional (function as value = std::make_unique) it's hard to read. why not be more explicit by replacing it with lambda and inside it a call to std::make_unique<T>(..) ?

src/solver/infeasible-problem-analysis/watched-constraints.cpp
flomnes: yes, `double` are passed by copy. It doesn't really make sense to specify `const`, it means that the copy used inside the function will be const, which is an implementation detail

```cpp

void f(int x)
{
    x++; // OK, I can change the copy
}

void f(const int x)
{
    x++; // ERROR
}
```

It's very different from e.g
```cpp

void f(int& x)
{
    x++; // OK, I can change pointed int
}

void f(const int& x)
{
    x++; // ERROR
}
```

src/solver/infeasible-problem-analysis/watched-constraints.cpp
flomnes: Yes I agree it's a bit hard to read as it is, but when taking a look at the rest of the code it's kind of obvious it is a function (`std::function`)

I don't like adding trivial lambda functions such as
```cpp
auto lambdaFunction = [&](Arg arg){return f(arg);}
```

src/solver/infeasible-problem-analysis/watched-constraints.cpp
```diff
@@ -159,17 +164,22 @@ std::string HydroProduction::infeasibilityCause()
...
+      {"^NettingSum::",
```
a-zakir: I introduced the use of regex here originally, but I’d suggest a small yet meaningful improvement for performance:

Change the map to a vector of precompiled regexes:
```
std::vector<std::pair<std::regex,
    std::function<std::unique_ptr<WatchedConstraint>(const std::string&, double)>>>
    regex_to_ctypes_;

```
And update the constructor accordingly:
```
ConstraintsFactory::ConstraintsFactory()
{
    regex_to_ctypes_ = {
        {std::regex("::hourly::"), [](const std::string& s, double d) {
            return std::make_unique<HourlyBC>(s, d);
        }},
        // ...
    };
}

```

This ensures that the regexes are compiled once during construction, instead of on every call to `create()`, which can significantly improve performance when matching large numbers of constraints.

Update the `create()` method as follows:
```
std::unique_ptr<WatchedConstraint> ConstraintsFactory::create(const std::string& name,
                                                              const double value) const
{
    for (const auto& [pattern, factory] : regex_to_ctypes_)
    {
        if (std::regex_search(name, pattern))
        {
            return factory(name, value);
        }
    }
    return nullptr;
}

```

flomnes: done (although much more simple than what you suggest) 4bef2629a

src/solver/infeasible-problem-analysis/watched-constraints.cpp
flomnes: done (not on purpose)

src/solver/infeasible-problem-analysis/watched-constraints.cpp
guilpier-code: ok, it's just that we make it clear that the parameters will not be changed (whether it's a copy or not)

guilpier-code (approved):

a-zakir (approved):

PR #2872, opened 2025-06-18
Infeasibility caused by hydro [ANT-3122]

guilpier-code:
Trial to fix [ticket ANT-3122](https://gopro-tickets.rte-france.com/browse/ANT-3122)

### DIFF
--- a/src/solver/simulation/common-hydro-levels.cpp
+++ b/src/solver/simulation/common-hydro-levels.cpp
@@ -63,11 +63,10 @@ void interpolateWaterValue(const Data::AreaList& areas,
         }
 
         double reservoirCapacity = area->hydro.reservoirCapacity;
-
+        double weekInitLevel = problem.CaracteristiquesHydrauliques[index].NiveauInitialReservoir;
         const std::vector<double>& niv = weeklyResults.niveauxHoraires;
 
-        waterVal[0] = Data::getWaterValue(100 * problem.previousSimulationFinalLevel[index]
-                                            / reservoirCapacity,
+        waterVal[0] = Data::getWaterValue(100 * weekInitLevel / reservoirCapacity,
                                           area->hydro.waterValues,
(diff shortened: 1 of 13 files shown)


src/solver/simulation/common-hydro-levels.cpp
```diff
@@ -94,6 +94,15 @@ void updatingWeeklyFinalHydroLevel(const Data::AreaList& areas, PROBLEME_HEBDO&
         const std::vector<double>& niv = weeklyResults.niveauxHoraires;
 
         problem.previousSimulationFinalLevel[index] = niv[nbHoursInAWeek - 1];
+
+        // Tiny numerical errors (for instance when using solver Xpress) can lead
+        // to a final level slightly above reservoir capacity.
+        // This causes an infeasibility when solving weekly optimization associated to
+        // week next to the current one. We make sure that final level remains correct.
+        if (problem.previousSimulationFinalLevel[index] > area->hydro.reservoirCapacity)
```
tbittar: Do the same if the final level falls below 0

tbittar (changes requested):

src/solver/simulation/common-hydro-levels.cpp
flomnes: done (see below), but we could clearly use C++17's [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp.html) for a nice one-liner.

```cpp
problem.previousSimulationFinalLevel[index] = std::clamp(niv[nbHoursInAWeek - 1],
                                                         0.,
                                                         area->hydro.reservoirCapacity);

```

guilpier-code: We're on discussion with @tbittar and @Juliette-Gerbaux , this change is about to be reverted.
I turned the PR into draft mode.

flomnes:
See #2874

PR #2874, opened 2025-06-19
Infeasibility caused by hydro (attempt 2) [ANT-3122]

guilpier-code:
Trial to fix [ticket ANT-3122](https://gopro-tickets.rte-france.com/browse/ANT-3122)

A first attempt was made [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2872).

Instead of changing the first PR, we created the new current PR because : 
- it felt easier to create a new PR
- we wanted to keep old changes contained in first PR for any purpose 

These (new) changes were imagined after failures occurring on the CI of first PR (see failures [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2872)).

**What was also done (regarding CI only)** : 
It's bad (we should have done that in separate PRs), but in this PR, we also : 
- moved archives (complete and solver only) creation and upload (**.zip** on **Win** and **.tar.gz** on **Ubuntu**) to right after the build step
- fix python scripts : about check integrity comparison, we switch actual and reference results so that logs are correct in case we get differences between the two.


src/solver/simulation/common-hydro-levels.cpp
```diff
@@ -57,14 +57,9 @@ void interpolateWaterValue(const Data::AreaList& areas,
             return;
         }
 
-        if (!area->hydro.useWaterValue)
```
guilpier-code, line 60:
```c++
if (!area->hydro.useWaterValue)
{
    return;
}
```
is useless if we consider the block just above

src/solver/simulation/common-eco-adq.cpp
```diff
@@ -349,7 +349,9 @@ void SetInitialHydroLevel(Data::Study& study,
     study.areas.each(
       [&problem, &firstDaySimu, &hydroVentilationResults](const Data::Area& area)
       {
-          if (area.hydro.reservoirManagement)
+          bool TurbinageEntreBornes = area.hydro.reservoirManagement
+                                      && (!area.hydro.useHeuristicTarget || area.hydro.useLeeway);
+          if (TurbinageEntreBornes)
           {
```
Juliette-Gerbaux, line 355: Is it possible to find a more explicit name for TurbinageEntreBornes ?

payetvin (approved):

PR #2887, opened 2025-06-26
Bypass JIT only for STS [ANT-3425]

payetvin:


flomnes (commented):
Instead of adding another new function, you could addd an optional argument to the existing function

src/libs/antares/series/include/antares/series/series.h
```diff
@@ -74,7 +74,7 @@ class TimeSeries
      ** \param average used to average timeseries
      ** \return A non-zero value if the operation succeeded, 0 otherwise
      */
-    bool loadFromFile(const std::filesystem::path& path, const bool average);
+    bool loadFromFile(const std::filesystem::path& path, const bool average, unsigned options = 0);
```
flomnes:
```suggestion
    bool loadFromFile(const std::filesystem::path& path, const bool average, unsigned options = Matrix<>::optNone);
```

flomnes (changes requested):

flomnes (approved):

PR #2891, opened 2025-06-30
Windows - ignore checks on optimal values

flomnes:


flomnes:
#2892 was implemented instead

PR #2899, opened 2025-07-02
Remove solver header logs

payetvin:


flomnes:
Not sure this is a good idea

PR #2901, opened 2025-07-03
Revert Objective

flomnes:
Hopefully it is no longer required


flomnes:
@a-zakir Bugfix still not deployed in MSVC 👎🏻

PR #2906, opened 2025-07-08
Removal of safe resolution [ANT-3566]

a-zakir:


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -246,7 +241,7 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
             optimizationStatistics.addUpdateTime(timeMeasure.updateTime);
         }
     }
-
+    // TODO could this condition be true?
```
flomnes: Yes, at the 1st resolution see #2868

flomnes (approved):

PR #2907, opened 2025-07-09
Add 2 future parameters in generaldata.ini [ANT-3504]

payetvin:


flomnes (commented):
Can you add an explanatory comment "to be used in a future version" ?

flomnes (approved):
~~Can you add an explanatory comment "to be used in a future version" ?~~ DONE

PR #2914, opened 2025-07-11
Use std::ifstream for Matrix<>::loadFromFileToBuffer [ANT-3426]

flomnes:


src/libs/antares/array/include/antares/array/matrix.h
```diff
@@ -228,10 +229,29 @@ class Matrix
...
     {
-        return Yuni::IO::File::LoadFromFile(buffer, filename, filesizeHardLimit);
+        // Open the file in binary mode and move to the end to get size
+        std::ifstream file(filepath, std::ios::binary | std::ios::ate);
+        if (!file.is_open())
+        {
+            return Yuni::IO::errNotFound;
+        }
+
+        std::streamsize size = file.tellg(); // Get size
```
a-zakir: tellg may not be the appropriate function see:
https://stackoverflow.com/a/5840160 
https://stackoverflow.com/a/22986486
actually you can use https://en.cppreference.com/w/cpp/filesystem/file_size.html

flomnes: done (use fs::file_size)

src/libs/antares/study/save.cpp
```diff
@@ -222,10 +222,6 @@ bool Study::saveToFolder(const AnyString& newfolder)
     buffer.clear() << folder << SEP << "layers" << SEP << "layers.ini";
     ret = saveLayers(buffer) and ret;
 
-    // Binding constraints
-    fs::path filpath = folder / "input" / "bindingconstraints";
-    ret = bindingConstraints.saveToFolder(filpath) and ret;
```
guilpier-code, line 227: This directory (**input/bindingconstraints**)  is already created when calling : 
```c++
ret = areas.saveToFolder(folder.string()) and ret;
```
just above

flomnes:
Won't fix (legacy GUI)

PR #2917, opened 2025-07-11
Test problems in-memory [ANT-3514]

flomnes:
Add observer implementation allowing to tests problems in-memory.

This way, tests are more unitary since we only test the problem creation from memory structures, not problem resolution, I/O and so forth.

Untitled Diagram drawio(4)

### DIFF
--- a/src/tests/inmemory-study/include/in-memory-study.h
+++ b/src/tests/inmemory-study/include/in-memory-study.h
@@ -20,6 +20,8 @@
  */
 #pragma once
 #define WIN32_LEAN_AND_MEAN
+#include <limits>
+
 #include "antares/solver/simulation/ISimulationObserver.h"
 #include "antares/solver/simulation/economy.h"
 #include "antares/solver/simulation/simulation.h"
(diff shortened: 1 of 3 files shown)


src/tests/inmemory-study/include/in-memory-study.h
```diff
@@ -295,6 +296,36 @@ class ScenarioBuilderRule
 // Simulation handler
 // =====================
 
+class TestingSimulationObserver: public Solver::Simulation::ISimulationObserver
+{
+public:
+    struct Variable
+    {
+        double Xmin;
```
JasonMarechal25: Please add default value to struct members.

JasonMarechal25 (changes requested):

src/tests/inmemory-study/include/in-memory-study.h
flomnes: Done, see [Use quiet_NaN for default values](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2917/commits/8c3a48e1bf36c5edca99636208314ba5decf1fa8)

JasonMarechal25 (approved):

PR #2922, opened 2025-07-16
Hydro remix algo : simplification

guilpier-code:
Fixes ticket [ANT-2619](https://gopro-tickets.rte-france.com/browse/ANT-2619), which is required for ticket [ANT-3175](https://gopro-tickets.rte-france.com/browse/ANT-3175).

**What was done** : 
- [x] **mainly** : separate remix algorithm from the storage it applies on. 
The algorithm remains a fonction, but the storage is no longer a reservoir : it's an instance of a more generic class, injected as argument to the algorithm. 
This instance could relate to another storage (like short term storage).
- [x] so, old algorithm was renamed from **shavePeaksByRemixingHydro** into **shavePeaksByRemixingStorageGen**, more accurate.
- [x] old code regarding the algorithm was split into multiple files, and code about algo remix has become drier. 
- [x] many renaming was done, to attempt to make code more understandable
- [ ] what else ? _To be completed_


src/solver/simulation/hydro-for-remix.cpp
```diff
@@ -0,0 +1,193 @@
...
+void HydroForRemix::checkInput(size_t size)
+{
+    std::vector<size_t> sizes = {generation_.size(), pmin_.size(), pmax_.size()};
+
+    if (!std::ranges::all_of(sizes, [&sizes](const size_t s) { return s == sizes.front(); }))
+    {
+        throw std::invalid_argument(error_msg_start + "arrays of different sizes");
+    }
+
+    if (!(generation_ <= pmax_))
```
Juliette-Gerbaux: Rajouter une tolérance (1e-6) pour éviter de renvoyer une erreur à cause de bruits du solver

src/solver/simulation/hydro-for-remix.cpp
```diff
@@ -0,0 +1,193 @@
...
+    {
+        throw std::invalid_argument(error_msg_start
+                                    + "Hydro generation not smaller than Pmax everywhere");
+    }
+
+    if (!(pmin_ <= generation_))
```
Juliette-Gerbaux: Idem

src/solver/simulation/hydro-for-remix.cpp
```diff
@@ -0,0 +1,193 @@
...
+    checkInput(generation_.size());
+    update();
+    checkLevels();
+}
+
+double HydroForRemixWithLevels::maxExchange(unsigned hourMax, unsigned hourMin)
+{
+    double bound = HydroForRemix::maxExchange(hourMax, hourMin);
+
+    unsigned smallestHour = std::min(hourMin, hourMax);
```
Juliette-Gerbaux: Peut-être renommer first/last hour pour éviter les confusions avec Min/Max ?

src/solver/simulation/hydro-for-remix.cpp
```diff
@@ -0,0 +1,193 @@
...
+void HydroForRemixWithLevels::checkLevels()
+{
+    if (!(levels_ <= capacity_ + LEVEL_TOLERANCE) || !(levels_ >= -LEVEL_TOLERANCE))
+    {
+        throw std::invalid_argument(error_msg_start
+                                    + "levels computed from input don't respect reservoir bounds");
+    }
+}
+
+void HydroForRemixWithLevels::update()
```
Juliette-Gerbaux, line 184: Renommer updateLevels ?

src/solver/simulation/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -0,0 +1,184 @@
...
+                    storage->update();
+
+                    TotalGen = updateTotalGen(DispatchGen, storage->generation());
+                    break;
+                }
+                triedMaxs[hourMax] = true;
+            }
+
+            UnsupE[hourMin] = storageGenInit[hourMin] + UnsupEinit[hourMin]
+                              - storage->generation()[hourMin];
```
Juliette-Gerbaux: Déplacer la mise à jour d'UnsupE hour Min avec celle d'hour Max ?

src/solver/simulation/include/antares/solver/simulation/hydro-for-remix.h
```diff
@@ -0,0 +1,60 @@
...
+protected:
+    void checkInput(size_t size) override;
+
+    std::vector<double>& generation_;
+    std::vector<double>& unsupE_;
+    const std::vector<double>& pmax_;
+    const std::vector<double>& pmin_;
+};
+
+class HydroForRemixWithLevels: public HydroForRemix
```
flomnes, line 29: For me it would be a bit clearer to have
```cpp
class HydroRemixNoLevels : public StorageForRemix {};
class HydroRemixLevels : public StorageForRemix {};
```
Also a virtual class probably does more than just storage, so you may consider renaming the base class.

src/solver/simulation/hydro-for-remix.cpp
guilpier-code: **hourMin** and **hourMax** were renamed into  **hourOfMinGen** and **hourOfMaxGen** to make it clear.
Are you ok with that ?

src/solver/simulation/hydro-for-remix.cpp
guilpier-code: The current remix algorithm is now more generic. 
This implies that the algo knows little about the storage it uses.
The algo only knows that, by contract : 
- the storage is generating power and how much is generated (function **generation()**), 
- the storage limits possible exchanges of power between 2 hours (function **maxExchange()**), 
- the storage can update itself (function **update()**).

But the algorithm does not know how this these quantities or operations are done / made.
In particular, the algorithm does not have to know if levels are involved somewhere during its own execution process. 
Indeed, what if there **reservoir management = false** ?

src/solver/simulation/shave-peaks-by-remix-storage-gen.cpp
guilpier-code: Done.

src/solver/simulation/include/antares/solver/simulation/hydro-for-remix.h
guilpier-code: > For me it would be a bit clearer to have
> 
> ```c++
> class HydroRemixNoLevels : public StorageForRemix {};
> class HydroRemixLevels : public StorageForRemix {};
> ```

I'm afraid there would be duplicate code if we do separate (as you suggest) hydro without levels and hydro with levels.
For instance we call **HydroForRemix :: maxExchange()** and **HydroForRemix :: checkInput(...)** inside methods of class **HydroForRemixWithLevels**.

Another way to say it is that the current class hierarchy makes sense : a reservoir with levels (reservoir management = true) **is** a reservoir (no levels) that also works with levels.

guilpier-code: > Also a virtual class probably does more than just storage, so you may consider renaming the base class.

I'm not sure we understand each other here : by **storage** I don't mean **data storage**, as you seem to suggest, but **storage** in a business sense : long term or short term storage.

src/solver/simulation/hydro-for-remix.cpp
guilpier-code: Done

src/solver/simulation/hydro-for-remix.cpp
guilpier-code: Done

JasonMarechal25 (approved):

src/solver/simulation/hydro-for-remix.cpp
```diff
@@ -0,0 +1,194 @@
...
+    double boundAtMax = generation_[hourOfMaxGen] - pmin_[hourOfMaxGen];
+    // Max slice we can add to hydro generation, at an hour when the total
+    // production reaches a min.
+    double boundAtMin = std::min(
+      {pmax_[hourOfMinGen] - generation_[hourOfMinGen], unsupE_[hourOfMinGen]});
+
+    return std::min(boundAtMax, boundAtMin);
+}
+
+void HydroForRemix::checkInput(size_t size)
```
guilpier-code, line 69: From @payetvin : parameters size unused

guilpier-code, line 69: comment to try to reopen the PR

guilpier-code (commented):
reopen the PR

PR #2926, opened 2025-07-17
Performance for problem building : avoid maps copy

tbittar:
Capture d’écran du 2025-07-17 18-39-45

Expensive copies were made when evaluating allTimeSumeOperators, especially in add_maps, using an inefficient += operator. This PR avoids some useless copies at this stage

### DIFF
--- a/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
+++ b/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
@@ -62,14 +62,14 @@ TimeDependentLinearExpression::TimeDependentLinearExpression(
 TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
   const TimeDependentLinearExpression& other)
 {
-    add_maps(linearExpressions_, std::move(other.GetLinearExpressions()));
+    add_maps(linearExpressions_, other.GetLinearExpressions());
     return *this;
 }
 
 TimeDependentLinearExpression& TimeDependentLinearExpression::operator-=(
   const TimeDependentLinearExpression& other)
 {
-    add_maps(linearExpressions_, std::move(other.GetLinearExpressions()), std::negate<>());
+    add_maps(linearExpressions_, other.GetLinearExpressions(), std::negate<>());
     return *this;
 }
 
--- a/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
+++ b/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
@@ -66,13 +66,6 @@ TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
     return *this;
 }
 
-// TimeDependentLinearExpression& TimeDependentLinearExpression::operator-=(
-//   const TimeDependentLinearExpression& other)
-// {
-//     add_maps(linearExpressions_, std::move(other.GetLinearExpressions()), std::negate<double>());
-//     return *this;
-// }
-
 TimeDependentLinearExpression TimeDependentLinearExpression::operator+(
   const TimeDependentLinearExpression& other) const
 {
--- a/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
+++ b/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
@@ -53,18 +53,40 @@ TimeDependentLinearExpression::TimeDependentLinearExpression(LinearExpressionMap
 {
 }
 
+TimeDependentLinearExpression::TimeDependentLinearExpression(
+  const TimeDependentLinearExpression& timeDependentLinearExpression):
+    linearExpressions_(timeDependentLinearExpression.linearExpressions_)
+{
+}
+
+TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
+  const TimeDependentLinearExpression& other)
+{
+    add_maps(linearExpressions_, std::move(other.GetLinearExpressions()));
+    return *this;
+}
+
+// TimeDependentLinearExpression& TimeDependentLinearExpression::operator-=(
+//   const TimeDependentLinearExpression& other)
+// {
+//     add_maps(linearExpressions_, std::move(other.GetLinearExpressions()), std::negate<double>());
+//     return *this;
+// }
+
 TimeDependentLinearExpression TimeDependentLinearExpression::operator+(
   const TimeDependentLinearExpression& other) const
 {
-    return TimeDependentLinearExpression(
-      add_maps(GetLinearExpressions(), other.GetLinearExpressions()));
+    auto result(*this);
+    result += other;
+    return result;
 }
 
 TimeDependentLinearExpression TimeDependentLinearExpression::operator-(
   const TimeDependentLinearExpression& other) const
 {
-    return TimeDependentLinearExpression(
-      add_maps(GetLinearExpressions(), other.GetLinearExpressions(), std::negate<>()));
+    auto result(*this);
+    result += -other;
+    return result;
 }
 
 template<typename BinaryOperator>
@@ -122,13 +144,6 @@ const LinearExpressionMap& TimeDependentLinearExpression::GetLinearExpressions()
 size_t TimeDependentLinearExpression::getSize() const
 {
     return linearExpressions_.size();
-}
-
-TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
-  const TimeDependentLinearExpression& other)
-{
-    linearExpressions_ = add_maps(linearExpressions_, other.GetLinearExpressions());
-    return *this;
 }
 
 TimeDependentLinearExpression TimeDependentLinearExpression::shiftLinearExpressions(
--- a/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
+++ b/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
@@ -66,6 +66,13 @@ TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
     return *this;
 }
 
+TimeDependentLinearExpression& TimeDependentLinearExpression::operator-=(
+  const TimeDependentLinearExpression& other)
+{
+    add_maps(linearExpressions_, std::move(other.GetLinearExpressions()), std::negate<>());
+    return *this;
+}
+
 TimeDependentLinearExpression TimeDependentLinearExpression::operator+(
   const TimeDependentLinearExpression& other) const
 {
@@ -78,7 +85,7 @@ TimeDependentLinearExpression TimeDependentLinearExpression::operator-(
   const TimeDependentLinearExpression& other) const
 {
     auto result(*this);
-    result += -other;
+    result -= other;
     return result;
 }
 
--- a/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
+++ b/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
@@ -53,12 +53,6 @@ TimeDependentLinearExpression::TimeDependentLinearExpression(LinearExpressionMap
 {
 }
 
-TimeDependentLinearExpression::TimeDependentLinearExpression(
-  const TimeDependentLinearExpression& timeDependentLinearExpression):
-    linearExpressions_(timeDependentLinearExpression.linearExpressions_)
-{
-}
-
 TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
   const TimeDependentLinearExpression& other)
 {
--- a/src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
+++ b/src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
@@ -101,10 +101,11 @@ void add_maps(MapType& left, const MapType&& right, UnaryOp op = std::identity{}
 {
     for (auto& [key, value]: right)
     {
-        if (left.find(key) != left.end())
+        auto it = left.find(key);
+        if (it != left.end())
         {
             // Key exists in left, add the values
-            left[key] += op(value);
+            it->second += op(value);
         }
         else
         {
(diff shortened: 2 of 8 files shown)


pet-mit (approved):

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -97,21 +97,21 @@ using FullKeyMap = std::unordered_map<FullKey, double, FullKeyHash>;
  * ```
  */
 template<typename MapType, typename UnaryOp = std::identity>
-MapType add_maps(const MapType& left, const MapType& right, UnaryOp op = std::identity{})
+void add_maps(MapType& left, const MapType&& right, UnaryOp op = std::identity{})
```
a-zakir: what is doing const MapType&& right ? you still have copy of 'right'

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -53,18 +53,33 @@ TimeDependentLinearExpression::TimeDependentLinearExpression(LinearExpressionMap
...
+TimeDependentLinearExpression::TimeDependentLinearExpression(
+  const TimeDependentLinearExpression& timeDependentLinearExpression):
+    linearExpressions_(timeDependentLinearExpression.linearExpressions_)
+{
+}
+
+TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
+  const TimeDependentLinearExpression& other)
+{
+    add_maps(linearExpressions_, std::move(other.GetLinearExpressions()));
```
a-zakir: this is not a move, still copying

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
```diff
@@ -65,6 +83,7 @@ class TimeDependentLinearExpression
     const LinearExpressionMap& GetLinearExpressions() const;
     size_t getSize() const;
     TimeDependentLinearExpression& operator+=(const TimeDependentLinearExpression& other);
+    TimeDependentLinearExpression& operator-=(const TimeDependentLinearExpression& other);
```
a-zakir, line 78: not defined

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -97,21 +97,21 @@ using FullKeyMap = std::unordered_map<FullKey, double, FullKeyHash>;
...
 {
-    auto result(left);
-    for (auto [key, value]: right)
+    for (auto& [key, value]: right)
     {
-        if (result.contains(key))
+        if (left.find(key) != left.end())
```
a-zakir: not efficient, you should find the element outside the if then compare it to .end()

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -97,21 +97,21 @@ using FullKeyMap = std::unordered_map<FullKey, double, FullKeyHash>;
...
         {
-            result[key] += op(value);
+            // Key exists in left, add the values
+            left[key] += op(value);
```
a-zakir:
```suggestion
        auto e = left.find(key);
        if (e != left.end())
{
            // Key exists in left, add the values
            e->second += op(value);
```

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
tbittar: Done

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
tbittar: ### Current Signature and Its Implications
#### Current: const MapType&& right
- This is an rvalue reference to const, which is unusual and rarely useful.
- In practice, this means you can only pass temporaries (rvalues) to right, and you cannot modify them (because of const).
- However, you also cannot move from them (because of const), so you lose the main benefit of rvalue references.

#### Typical Efficient Signature: MapType&& right
- This would allow you to move from temporaries, which can be more efficient if you want to "steal" their contents.

#### Proposed: const MapType& right
- This allows both lvalues and rvalues (as const references) to be passed.
- You cannot move from right, but you can read from it as many times as you want.

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
tbittar: Done

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
tbittar: Good catch ! There is no move (well in fact there was using the previous signature `const MapType&& right`, but no copy either

Great question! Let's analyze line 65 in `TimeDependentLinearExpression.cpp`:

```cpp
add_maps(linearExpressions_, std::move(other.GetLinearExpressions()));
```

### 1. **Is there a copy?**

- **`other.GetLinearExpressions()`** returns a `const LinearExpressionMap&` (as per the header).
- **`std::move(other.GetLinearExpressions())`** does **not** actually move anything here, because you cannot move from a const reference. The result is still a const reference, and the move is a no-op.
- **`add_maps`** now takes `const MapType& right`, so it will not move or copy the map itself, but will iterate over it and copy/assign the values as needed into `linearExpressions_`.

**Conclusion:**  
There is **no copy of the entire map**. The only copying that happens is for the individual elements that are inserted into or added to `linearExpressions_`.

---

### 2. **Is this the most efficient way?**

- **For the current use case:**  
  Yes, this is the most efficient and idiomatic way to add the contents of one map into another, given that you do not want to modify or "steal" from `other`.
- **Why?**  
  - You avoid unnecessary temporaries.
  - You do not create a copy of the whole map.
  - You only copy/assign the values that are actually needed for the addition.

- **Could it be more efficient?**  
  - Only if you could guarantee that `other` is a temporary and you wanted to move its contents (which is rare for mathematical expressions, as you usually want to preserve the original).
  - In that case, you could overload `add_maps` to take an rvalue reference and move elements, but this is not typical for this kind of operation.

---

### 3. **Should you remove `std::move`?**

- **Yes, you should remove `std::move`** from the call:
  ```cpp
  add_maps(linearExpressions_, other.GetLinearExpressions());
  ```
  - It is clearer and avoids confusion, since `std::move` on a const reference does nothing.

---

## **Summary Table**

| Code (current)                                   | Is there a copy? | Most efficient? | Suggestion                |
|--------------------------------------------------|------------------|-----------------|---------------------------|
| `add_maps(linearExpressions_, std::move(other.GetLinearExpressions()));` | No (not of map)  | Yes (for const input) | Remove `std::move`        |
| `add_maps(linearExpressions_, other.GetLinearExpressions());`            | No (not of map)  | Yes                  | Use this for clarity      |

---

## **Final Recommendation**

- **No, there is no copy of the entire map.**
- **Yes, this is the most efficient way for your use case.**
- **Remove `std::move` for clarity.**

If you want, I can make this small cleanup for you!

flomnes (changes requested):
The documentation for `add_maps` should be edited to reflect the changes on the function prototype.
from
```cpp
auto result = add_maps(map3, map4);
```
to
```cpp
add_maps(result, other);
```

flomnes:
Also, I heard that [absl::flat_hash_map](https://abseil.io/docs/cpp/guides/container) is quite faster than `std::map` or `std::unordered_map`. To be investigated

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
```diff
@@ -44,6 +44,24 @@ class TimeDependentLinearExpression
...
+
+    // Move assignment operator
+    TimeDependentLinearExpression& operator=(TimeDependentLinearExpression&& other) noexcept
+    {
+        if (this != &other)
+        {
+            linearExpressions_ = std::move(other.linearExpressions_);
+        }
+        return *this;
+    }
```
flomnes: This class only has a single member `linearExpressions_` and no parent class. Its move and copy constructors can be generated by the compiler, there is no need to specify them

```suggestion
    TimeDependentLinearExpression(TimeDependentLinearExpression&& other) noexcept = default;
    TimeDependentLinearExpression& operator=(TimeDependentLinearExpression&& other) noexcept = default;
```

~~Note: Move constructors for `std::unordered_map` are not marked `noexcept` and may throw.~~

Fun fact: in GCC's implementation, unordered_map's move constructor & move assignment are also compiler-generated
```cpp
      /// Move constructor.
      unordered_map(unordered_map&&) = default;
      /// Move assignment operator.
      unordered_map&
      operator=(unordered_map&&) = default;
```

flomnes (commented):
We might as well remove all operations that are not "in place", e.g
```
y = a + b
```
and only keep only "in place" operations to avoid copies
```
y += a;
y += b;
```

Example of a copy
```cpp
LinearExpression LinearExpression::operator+(const LinearExpression& other) const
{
    auto result(*this); // Copy
    result += other;
    return result;
}
```

tbittar:
> We might as well remove all operations that are not "in place", e.g
> 
> ```
> y = a + b
> ```
> 
> and only keep only "in place" operations to avoid copies
> 
> ```
> y += a;
> y += b;
> ```
> 
> Example of a copy
> 
> ```c++
> LinearExpression LinearExpression::operator+(const LinearExpression& other) const
> {
>     auto result(*this); // Copy
>     result += other;
>     return result;
> }
> ```

Yes that's the whole point of this PR, replacing the call to `operator+` by `+=` when adding maps. This should be done for all operators for which it is possible (`scale_map` ?)

flomnes:
> > We might as well remove all operations that are not "in place", e.g
> > ```
> > y = a + b
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and only keep only "in place" operations to avoid copies
> > ```
> > y += a;
> > y += b;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Example of a copy
> > ```c++
> > LinearExpression LinearExpression::operator+(const LinearExpression& other) const
> > {
> >     auto result(*this); // Copy
> >     result += other;
> >     return result;
> > }
> > ```
> 
> Yes that's the whole point of this PR, replacing the call to `operator+` by `+=` when adding maps. This should be done for all operators for which it is possible (`scale_map` ?)

Then there are two options
- Remove `operator+`, `operator-`, etc.
- Mark them as `[[deprecated]]` to display a warning whenever they're used

a-zakir:
in-place modification reduces thread safety compared to the original implementation

flomnes:
> in-place modification reduces thread safety compared to the original implementation

That's right, but this lib is currently not designed to be thread-safe.

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
```diff
@@ -44,6 +44,11 @@ class TimeDependentLinearExpression
       const Optimisation::LinearProblemApi::FillContext& fillContext,
       const LinearExpression& linearExpression);
     explicit TimeDependentLinearExpression(LinearExpressionMap linearExpressions);
+    explicit TimeDependentLinearExpression(
+      const TimeDependentLinearExpression& timeDependentLinearExpression);
+
+    TimeDependentLinearExpression(TimeDependentLinearExpression&& other) = default;
+    TimeDependentLinearExpression& operator=(TimeDependentLinearExpression&& other) = default;
```
flomnes:
```suggestion
    TimeDependentLinearExpression(TimeDependentLinearExpression&& other) noexcept = default;
    TimeDependentLinearExpression& operator=(TimeDependentLinearExpression&& other) = default;
```

flomnes (commented):
See https://next.sonarqube.com/sonarqube/coding_rules?open=cpp%3AS5018&rule_key=cpp%3AS5018

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
tbittar: Done

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/TimeDependentLinearExpression.h
tbittar: Done

tbittar:
> > > We might as well remove all operations that are not "in place", e.g
> > > ```
> > > y = a + b
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and only keep only "in place" operations to avoid copies
> > > ```
> > > y += a;
> > > y += b;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Example of a copy
> > > ```c++
> > > LinearExpression LinearExpression::operator+(const LinearExpression& other) const
> > > {
> > >     auto result(*this); // Copy
> > >     result += other;
> > >     return result;
> > > }
> > > ```
> > 
> > 
> > Yes that's the whole point of this PR, replacing the call to `operator+` by `+=` when adding maps. This should be done for all operators for which it is possible (`scale_map` ?)
> 
> Then there are two options
> 
>     * Remove `operator+`, `operator-`, etc.
> 
>     * Mark them as `[[deprecated]]` to display a warning whenever they're used

I choose to add `[deprecated]` because we may need that for specific use-case, and in this way we see all usage of `operator+` `operator-` during compilation (currently for substraction when reading a substraction node), but I think this is the work of another PR to modify that.

flomnes (approved):

PR #2927, opened 2025-07-21
Remove warnings on gcc

payetvin:


src/expressions/visitors/TimeIndexVisitor.cpp
```diff
@@ -89,7 +89,7 @@ TimeIndex TimeIndexVisitor::visit(const Nodes::NegationNode* neg)
     return dispatch(neg->child());
 }
 
-TimeIndex TimeIndexVisitor::visit(const Nodes::PortFieldNode* port_field_node)
+TimeIndex TimeIndexVisitor::visit([[maybe_unused]] const Nodes::PortFieldNode* port_field_node)
```
a-zakir: better use [unnamed parameter ](https://github.com/isocpp/CppCoreGuidelines/issues/808#issuecomment-263981643)
```suggestion
TimeIndex TimeIndexVisitor::visit( const Nodes::PortFieldNode* )
```

flomnes (approved):

PR #2928, opened 2025-07-22
CMake presets

JasonMarechal25:
Add a simple CMakePresets.json to derive from

Add CmakeUserPreset to git ignore to prevent leak of personal data


flomnes (commented):
Can you please document how to use this preset in the developer guide ?

flomnes (approved):

PR #2934, opened 2025-07-24
Forbid study path with non ascii chars [ANT-3635]

guilpier-code:
**Cases of loading / Saving a study we can think of** : 
- [x] Drag & drop of a study folder to GUI
- [x] Open a study by doing : "File > Open..."
- [ ] Open a study from the panel Welcome to **Antares Simulator > Open a recent study**
  **Nothing to do here** : GUI cannot add a study to "recent files" list if it wasn't correctly loaded first, which never happened for studies with non ASCII chars in their path.
- [x] Open a study on Windows, when user double-clicks a study folder or a file **study.antares**
- [x] **Save As** a study  into a folder whose path contains anon ASCII char
- [x] FYI, currently, there's no way to give a study a name with an non ASCII char : you can try typing such a char in the field **Name of the study** of panel **Save As**, this char is ignored and nothing is printed. 

**What was also done** : 
- [x] fix a crash in unit tests (see the comment on **tests-ts-numbers.cpp**).
- [x] many unit tests were added on functions contained in **src/solver/misc/options.cpp** to reduce the number of uncovered lines pointed by **Sonar Q**
- [x] unit tests were moved from  **tests-ts-numbers.cpp** to **test_utils.cpp**, because it's the place they belong.


src/ui/simulator/application/study.cpp
```diff
@@ -19,39 +19,36 @@
 ** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
 */
 #include "antares/study/study.h"
+
```
guilpier-code, line 22: The only real changes here are the definition and call of/to free function **containsOnlyASCIIcharsOnWindows(...)**.
The rest is IDE automatic reformatting.

guilpier-code (commented):
Just a comment to help review

flomnes:
> which never happened for studies with non ASCII chars in their path.

Simulator 8.8 could open studies with non-ascii characters, is the "recent studies" history shared across versions ?

==> I tried it : 
- creation of a simple study with 8.8.14 inside a folder with an accent in its path.
- check that the study can be found among recent studies for 8.8.14 GUI
- open **develop** GUI but can't find the previous study in the recent files

src/solver/misc/options.cpp
```diff
@@ -295,6 +295,15 @@ void checkAndCorrectSettingsAndOptions(Settings& settings, Data::StudyLoadOption
     }
 }
 
+static bool containsOnlyASCIIcharsOnWindows(const std::string& path)
+{
+#if defined(_WIN32) || defined(_WIN64)
```
```suggestion
#if defined(_WIN32)
```
> _WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, x64, or ARM64EC. Otherwise, undefined.

https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

flomnes (commented):
Please define `containsOnlyASCIIcharsOnWindows` only a single time. Even if we're changing code from a legacy & deprecated component, we are not savages.

==> done

src/solver/misc/options.cpp
guilpier-code: Done

guilpier-code:
> Please define `containsOnlyASCIIcharsOnWindows` only a single time. Even if we're changing code from a legacy & deprecated component, we are not savages.

Done

src/libs/antares/utils/utils.cpp
```diff
@@ -172,6 +173,15 @@ double floor(double d)
     return std::floor(std::round(d * largeValue) / largeValue);
 }
 
+bool containsOnlyASCIIcharsOnWindows(const std::string& path)
```
flomnes:
```suggestion
bool pathIsValid(const std::string& path)
```

flomnes (changes requested):

src/libs/antares/utils/utils.cpp
guilpier-code: Done

flomnes (changes requested):
The coverage is very poor (~18%). Please add unit tests in e.g src/tests/src/libs/antares/test_utils.cpp

flomnes (changes requested):
Please restore the --pid option

src/tests/src/solver/simulation/tests-ts-numbers.cpp
```diff
@@ -36,7 +36,7 @@ using namespace Antares::Solver::TimeSeriesNumbers;
 
 void initializeStudy(Study::Ptr study, unsigned int nbYears = 1)
 {
-    study->parameters.derated = false;
+    study->parameters.reset();
```
guilpier-code, line 39: We used to have a crash on **Win + Debug** mode, because **reset()** was not called on **parameters**.

Indeed, on test **load_wind_thermal_in_intra_and_inter_modal____check_all_ts_numbers_are_equal**,
line : 
```c++
TSGenerator::ResizeGeneratedTimeSeries(study->areas, study->parameters);
```
Inside **ResizeGeneratedTimeSeries(...)**, variables **params.timeSeriesToGenerate** and **params.nbTimeSeriesLoad** were undefined, causing the crash

So we fixed it by calling **reset()**.

guilpier-code:
> Please restore the --pid option

Done

src/tests/src/solver/misc/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+include(${CMAKE_SOURCE_DIR}/tests/macros.cmake)
+
+add_boost_test(tests-options
+        SRC
+        tests-options.cpp
+        INCLUDE
+        "${CMAKE_SOURCE_DIR}/solver/misc"
```
flomnes:
```suggestion
```

flomnes: This directory (src/solver/misc) does not contain any header, so this can be safely be removed (unused).

This would be useful to include private headers (src/solver/misc/private.h).

guilpier-code: Done

flomnes (approved):

PR #2936, opened 2025-07-24
Don't use basis in week 0

pet-mit:


pet-mit:
This version disables basis transmission during 2nd optimization of 1st week (it should only disable it during first optim).
Better version here : https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/2941

PR #2937, opened 2025-07-24
Add documentation for `include-export-solutions` for migration guide

flomnes:


MartinBelthle (approved):
You should also do a PR inside `antares-study-version` to change the v9.2 upgrader to add this field

flomnes:
> You should also do a PR inside `antares-study-version` to change the v9.2 upgrader to add this field

Not necessarily, it's an optional field

PR #2938, opened 2025-07-24
Revert hotfix for objective value

flomnes:
https://github.com/actions/runner-images/issues/12470

https://developercommunity.visualstudio.com/t/Code-optimization-bug-SIMD-std::transf/10912292


flomnes:
Still KO, maybe recompile sirius-solver ?

PR #2944, opened 2025-07-28
[ANT-3666] Move STS additional constraint data folder

JasonMarechal25:
- [ ] Add valid-v930 to CI


src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -108,96 +107,110 @@ static bool readRHS(const fs::path& rhsPath, TimeSeries& rhsSeries)
...
 
-        if (const auto rhsPath = parentPath / ("rhs_" + additionalConstraints->name + ".txt");
-            !readRHS(rhsPath, additionalConstraints->rhs()))
+        for (auto* section = ini.firstSection; section; section = section->next)
         {
-            logs.error() << "Error while reading rhs file: " << rhsPath;
-            return false;
-        }
+            auto additionalConstraints = std::make_shared<AdditionalConstraints>();
+            additionalConstraints->name = transformNameIntoID(section->name.c_str());
```
sylvlecl: I think we'd better keep the name AND the ID like for other objects, for consistency sake.

Otherwise, we'll use the ID in some places where we should use the name for consistency, see for example in the TS number data save function.

sylvlecl (approved):

PR #2945, opened 2025-07-28
[ANT-3666] 9.2 Move STS additional constraint data folder

JasonMarechal25:


JasonMarechal25:
Modeler tests fail on windows. It is already the case before this PR. Decision made to merge as is. Modeler is not "delivered" yet

payetvin (approved):

PR #2959, opened 2025-08-08
Hydro remix : more simplifications

guilpier-code:
Some more simplifications / clarifications of remix hydro algo + a step further towards [ticket ANT-3175](https://gopro-tickets.rte-france.com/browse/ANT-3175).

**Examples of things done** : 
- [ ] on study **short-tests/hydro_monthly_heuristic_gen_capa**, we had a regression that commit **"Hydro remix : fix a regression"** solves. No unit test could catch the problem. We should have such a unit test.
- [x] no more **DispatchGen** among the arguments of the algorithm : this vector was **Total Generation - hydro generation**, that is : **Load - Demand - Hydro - Unsupplied Energy**.
It was depending on **Hydro**, so on a particular  storage.
Instead, we have **Load** as argument. Unit tests were adapted accordingly.
- [x] checking algo input is now required outside (see **checkInput** function) algorithm, to separate responsibilities.
- [x] There used to be 2 kind of vectors allowing to filter hours where to find a **min** or a **max** for **TotalGen**. 
  - **validHours** (a **std::vector\<unsigned\>** of hours)
  - **triedMins** and **triedMaxs** (**std::vector\<bool\>**, telling if a min / max was already tried in the loop it belongs to)
 
  Now each of them are **std::set\<unsigned\>**, named **validHours**, **validHoursForMin** and **validHoursForMax**.
  The 2 last ones are built from the first.
- [x] updating impacted variable after finding a slice of energy to exchange was simplified
- [x] some free functions became so simple that they could be replaced with std algorithms
- [x] local variable **top**'s definition and use were removed : this variable turned out to be useless

### DIFF
--- a/src/solver/simulation/include/antares/solver/simulation/remix-utils.h
+++ b/src/solver/simulation/include/antares/solver/simulation/remix-utils.h
@@ -37,4 +37,18 @@ inline std::vector<double> operator-(const std::vector<double>& v, const double 
     return to_return;
 }
 
+inline std::vector<double> operator+(std::vector<double>& a, std::vector<double>& b)
+{
+    std::vector<double> out = a;
+    std::transform(out.begin(), out.end(), b.begin(), out.begin(), std::plus<>());
+    return out;
+}
+
+inline std::vector<double> operator-(std::vector<double>& a, std::vector<double>& b)
+{
+    std::vector<double> out = a;
+    std::transform(out.begin(), out.end(), b.begin(), out.begin(), std::minus<>());
+    return out;
+}
+
 }; // namespace Antares::Solver::Simulation
--- a/src/solver/simulation/include/antares/solver/simulation/remix-utils.h
+++ b/src/solver/simulation/include/antares/solver/simulation/remix-utils.h
@@ -37,14 +37,14 @@ inline std::vector<double> operator-(const std::vector<double>& v, const double 
     return to_return;
 }
 
-inline std::vector<double> operator+(std::vector<double>& a, std::vector<double>& b)
+inline std::vector<double> operator+(const std::vector<double>& a, const std::vector<double>& b)
 {
     std::vector<double> out = a;
     std::transform(out.begin(), out.end(), b.begin(), out.begin(), std::plus<>());
     return out;
 }
 
-inline std::vector<double> operator-(std::vector<double>& a, std::vector<double>& b)
+inline std::vector<double> operator-(const std::vector<double>& a, const std::vector<double>& b)
 {
     std::vector<double> out = a;
     std::transform(out.begin(), out.end(), b.begin(), out.begin(), std::minus<>());
--- a/src/solver/simulation/include/antares/solver/simulation/remix-utils.h
+++ b/src/solver/simulation/include/antares/solver/simulation/remix-utils.h
@@ -37,18 +37,16 @@ inline std::vector<double> operator-(const std::vector<double>& v, const double 
     return to_return;
 }
 
-inline std::vector<double> operator+(const std::vector<double>& a, const std::vector<double>& b)
+inline std::vector<double> operator+(std::vector<double> a, const std::vector<double>& b)
 {
-    std::vector<double> out = a;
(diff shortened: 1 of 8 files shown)


src/solver/simulation/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -99,87 +88,88 @@ static std::vector<bool> ValidHours(const std::vector<double>& Spillage,
...
+            exchange = std::max(std::min(maxExchangeFromStorage, maxVariation / 2.), 0.);
 
-            if (maxExchange > eps)
+            if (exchange > eps)
             {
-                break;
+                storage->generation()[hourOfMaxGen] -= exchange;
+                storage->generation()[hourOfMinGen] += exchange;
+
+                UnsupE[hourOfMaxGen] = storage->initialGen()[hourOfMaxGen]
```
Juliette-Gerbaux: Est ce que ce n'est pas possible de modifier UnsupE directement à partir de UnsupE et d'exchange ?

guilpier-code: C'est fait.

src/solver/simulation/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -65,121 +49,84 @@ static void checkInput(const std::vector<double>& DispatchGen,
...
-                                    const std::vector<double>& StorageGen,
-                                    const std::vector<double>& UnsupE)
-{
-    std::vector<bool> validHours(Spillage.size(), false);
-    for (unsigned h = 0; h < validHours.size(); h++)
+    std::set<unsigned> validHoursForMin(validHours);
+    while (true)
     {
-        if (Spillage[h] + DTG_MRG[h] == 0. && StorageGen[h] + UnsupE[h] > 0.)
+        std::erase_if(validHoursForMin, [&](int h) { return UnsupE[h] <= 0.; });
```
Juliette-Gerbaux: Use epsilon instead of 0 ?

Juliette-Gerbaux (approved):

src/solver/simulation/shave-peaks-by-remix-storage-gen.cpp
guilpier-code: Done

src/solver/simulation/include/antares/solver/simulation/remix-utils.h
```diff
@@ -37,4 +37,16 @@ inline std::vector<double> operator-(const std::vector<double>& v, const double
     return to_return;
 }
 
+inline std::vector<double> operator+(std::vector<double> a, const std::vector<double>& b)
+{
+    std::transform(a.begin(), a.end(), b.begin(), a.begin(), std::plus<>());
```
payetvin: 4rth arg is really a.begin() ? Just asking, i don't use transform often

flomnes: For binary operators, yes

flomnes (approved):

src/solver/simulation/include/antares/solver/simulation/remix-utils.h
guilpier-code: It's possible to do this, shorter : 
```c++
inline std::vector<double> operator+(std::vector<double> a, const std::vector<double>& b)
{
    std::range::transform(a, b, a.begin(), std::plus<double>());
    return a;
}
```

Juliette-Gerbaux (approved):

src/solver/simulation/include/antares/solver/simulation/remix-utils.h
```diff
@@ -4,48 +4,50 @@
 #include <ranges>
 #include <vector>
 
+namespace rng = std::ranges;
```
JasonMarechal25, line 7: Please don't shorten ranges as "rng". Could be confused with Random Number Generator

PR #2962, opened 2025-08-11
Performance optimizations in TimeDependentLinearExpression

a-zakir:
- single-pass sums
- fewer temporaries
- reduced map lookups

### DIFF
--- a/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
+++ b/src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
@@ -37,7 +37,7 @@ TimeDependentLinearExpression::TimeDependentLinearExpression(
     for (auto timestep(fillContext.getFirstTimeStep()); timestep <= fillContext.getLastTimeStep();
          ++timestep)
     {
-        linearExpressions_[timestep] = linearExpression;
+        linearExpressions_.emplace(timestep, linearExpression);
     }
 }
 
@@ -91,13 +91,13 @@ TimeDependentLinearExpression BinaryOpLinearExpression(const LinearExpressionMap
     auto result(left);
     for (const auto& [timeStep, other_linear_expression]: right)
     {
-        if (result.contains(timeStep))
+        if (auto it = result.find(timeStep); it == result.end())
         {
-            result[timeStep] = op(result.at(timeStep), other_linear_expression);
+            result.emplace(timeStep, other_linear_expression);
         }
         else
         {
-            result[timeStep] = other_linear_expression;
+            it->second = op(it->second, other_linear_expression);
         }
     }
     return TimeDependentLinearExpression(std::move(result));
@@ -121,11 +121,11 @@ TimeDependentLinearExpression TimeDependentLinearExpression::operator/(
 
 TimeDependentLinearExpression TimeDependentLinearExpression::operator-() const
 {
-    const auto& linear_expressions = GetLinearExpressions();
+    const auto& linearExpressions = GetLinearExpressions();
     LinearExpressionMap result;
-    for (const auto& timeStep: linear_expressions | std::views::keys)
+    for (const auto& [timeStep, linearExpression]: linearExpressions)
     {
-        result[timeStep] = -linear_expressions.at(timeStep);
+        result.emplace(timeStep, -linearExpression);
     }
     return TimeDependentLinearExpression(std::move(result));
 }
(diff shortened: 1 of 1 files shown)


src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -121,11 +121,11 @@ TimeDependentLinearExpression TimeDependentLinearExpression::operator/(
 
 TimeDependentLinearExpression TimeDependentLinearExpression::operator-() const
 {
-    const auto& linear_expressions = GetLinearExpressions();
+    const auto& linearExpressions = GetLinearExpressions();
     LinearExpressionMap result;
-    for (const auto& timeStep: linear_expressions | std::views::keys)
+    for (const auto& [timeStep, linearExpression]: linearExpressions)
```
flomnes, line 126:
```suggestion
LinearExpressionMap result;
    for (const auto& [timeStep, linearExpression]: GetLinearExpressions())
```

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -150,8 +150,9 @@ TimeDependentLinearExpression TimeDependentLinearExpression::shiftLinearExpressi
     LinearExpressionMap linearExpressions;
     for (const auto& timeStep: linearExpressions_ | std::views::keys)
     {
-        linearExpressions[timeStep] = linearExpressions_.at(
-          rotatedIndex(timeStep, shiftValue, fillContext));
+        linearExpressions.emplace(timeStep,
+                                  linearExpressions_.at(
+                                    rotatedIndex(timeStep, shiftValue, fillContext)));
     }
```
flomnes, line 156: `linearExpression` can be mutated in the loop

```suggestion
for (auto& [timeStep, linearExpression]: linearExpressions_)
{
        linearExpression = linearExpressions_.at(rotatedIndex(timeStep, shiftValue, fillContext)));
}
```

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -91,13 +91,13 @@ TimeDependentLinearExpression BinaryOpLinearExpression(const LinearExpressionMap
...
+        if (auto it = result.find(timeStep); it == result.end())
         {
-            result[timeStep] = op(result.at(timeStep), other_linear_expression);
+            result.emplace(timeStep, other_linear_expression);
         }
         else
         {
-            result[timeStep] = other_linear_expression;
+            it->second = op(it->second, other_linear_expression);
         }
```
flomnes, line 101: Not in the scope of this PR, but I don't think `a-b` is well defined when `a.timeSteps != b.timeSteps`

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -184,13 +185,13 @@ TimeDependentLinearExpression TimeDependentLinearExpression::allTimeSumLinearExp
     const Optimisation::LinearProblemApi::FillContext fillContext{
       linearExpressions_.begin()->first,
       linearExpressions_.rbegin()->first};
-    TimeDependentLinearExpression ret(fillContext);
-    for (auto timeStep = fillContext.getFirstTimeStep(); timeStep <= fillContext.getLastTimeStep();
-         ++timeStep)
+
+    LinearExpression sum;
+    for (const auto& expr: linearExpressions_ | std::views::values)
```
flomnes, line 190: # Before
1. Create a `TimeDependentLinearExpression ret` with `N=fillContext.getFirstTimeStep() - timeStep <= fillContext.getLastTimeStep()` zero through 

```cpp
TimeDependentLinearExpression::TimeDependentLinearExpression(
  const Optimisation::LinearProblemApi::FillContext& fillContext)
```
2. For each time-step, update all the identical components `N` times
3. Return the time-dependent expression composed (by construction) of `N` identical `LinearExpression`s

# After
1. Create a single (time-independent) linear expression `LinearExpression sum`
2. Add to it all `LinearExpression` for each time index
3. Duplicate this `LinearExpression` using this time
```cpp
TimeDependentLinearExpression::TimeDependentLinearExpression(
  const Optimisation::LinearProblemApi::FillContext& fillContext,
  const LinearExpression& linearExpression)
```

While an identical `LinearExpression` appears `N` times in the returned value in both cases, the 2nd version seems to perform fewer operations through `operator+=` and `add_maps` on `TimeDependentLinearExpression` objects, so in that sense it appears to be an improvement.

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
a-zakir: no it can't. if we want to do:
t  = 1 -> t = 3
t = 4 -> t = 1 (if t=1 is modified then the logic is broken )

flomnes: agreed

flomnes (approved):

flomnes (approved):

PR #2964, opened 2025-08-12
[ANT-3617] modeler 5.1 column base storage

a-zakir:


src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -0,0 +1,178 @@
...
+
+template<typename T>
+inline constexpr bool is_optional_v = is_optional<T>::value;
+
+// Primary template: applies to any type that's NOT an optional
+template<typename T>
+struct unwrap_optional
+{
+    using type = T; // just return the type unchanged
+};
```
flomnes: `T = std::optional<T>::value_type`

src/io/outputs/SimulationTableCsv.cpp
```diff
@@ -26,18 +26,42 @@
...
-    buffer_ << "block,component,output,"
-               "absolute_time_index,block_time_index,scenario_index,value,basis_status\n";
+    bool first = true;
+    for (const auto& col_name: storage_.columnNames())
+    {
+        if (!first)
+        {
+            buffer_ << ',';
+        }
+        first = false;
```
flomnes, line 62:
```suggestion
            first = false;
        }
```

src/io/outputs/include/antares/io/outputs/columns.h
```diff
@@ -0,0 +1,188 @@
...
+template<typename T>
+class TypedColumn: public IColumn
+{
+public:
+    explicit TypedColumn(std::function<std::string(const T&)> toStringFn = defaultToString,
+                         std::function<T(const std::string&)> fromStringFn = defaultFromString):
+        toStringFn_(toStringFn),
+        fromStringFn_(fromStringFn)
+    {
+    }
```
flomnes: Instead of providing `std::function`s at construction time, you can define non-pure `virtual` functions to allow derived classes to override these methods while still providing a default method.

```cpp
template<typename T>
class TypedColumn: public IColumn
{
virtual std::string toStringFn() const
{
    // paste defaultToString's implementation here
}
virtual T fromStringFn()
{
    // paste defaultFromString's implementation here
}
};
```

src/io/outputs/SimulationTableCsv.cpp
```diff
@@ -26,18 +26,42 @@
...
+    storage_.addStringColumn("output");
+    storage_.addOptionalColumn<unsigned int>("absolute_time_index");
+    storage_.addOptionalColumn<unsigned int>("block_time_index");
+    storage_.addOptionalColumn<unsigned int>("scenario_index");
+    storage_.addOptionalColumn<double>("value");
+    storage_.addStringColumn("basis_status");
     writeHeader();
 }
 
 void SimulationTableCsv::addEntry(SimulationTableEntry entry)
```
flomnes:
```suggestion
void SimulationTableCsv::addEntry(const SimulationTableEntry& entry)
```

src/io/outputs/include/antares/io/outputs/columns.h
```diff
@@ -0,0 +1,188 @@
...
+        if (option.has_value())
+        {
+            if constexpr (std::is_floating_point_v<T>)
+            {
+                return DoubleColumn::FromDouble(option.value());
+            }
+            else
+            {
+                return TypedColumn<T>::defaultToString(option.value());
+            }
```
flomnes: Should be simpler with the `virtual` functions suggestion above

```suggestion
             return TypedColumn<T>::toString(option.value());
```

src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -0,0 +1,178 @@
...
+        }
+        else if constexpr (is_optional_v<T>)
+        {
+            using Inner = unwrap_optional_t<T>;
+            getColumn<OptionalColumn<Inner>>(column_name).add(value);
+        }
+        else
+        {
+            throw std::runtime_error("Unsupported type"); // TODO
+        }
```
flomnes: this looks a bit too complex, probably too many cases to handle

src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -0,0 +1,178 @@
...
+        {
+            column->clear();
+        }
+    }
+
+private:
+    template<typename T>
+    struct always_false: std::false_type
+    {
+    };
```
flomnes: unused
```suggestion
```

src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -0,0 +1,178 @@
...
+
+    template<typename ColumnType>
+    void addColumn(const std::string& name)
+    {
+        // columns_[name] = std::make_unique<ColumnType>();
+        // column_order_.push_back(name);
```
flomnes:
```suggestion
```

flomnes (changes requested):
Add unit tests

src/io/outputs/include/antares/io/outputs/storage.h
flomnes: Also, `unwrap_optional_t` is useless, you can straight use `std::optional<T>::value_type` and
```cpp
template <typename T>
struct is_std_optional : std::false_type {};

template <typename T>
struct is_std_optional<std::optional<T>> : std::true_type {};
```
to determine if T is an `std::optional`

a-zakir: T is optional, so what is optional< T>? optional<optional< U> >

a-zakir: unwrap structs allows to extract the type behind optional.

flomnes: My point is that if `T = std::optional<U>`, then `U = unwrap_optional_t<T> = T::value_type`, so you don't really need `unwrap_optional_t`.

a-zakir: Just 4 cases, zero complexity

flomnes: In my opinion you might as well remove the template & overload `addValue`
```cpp
void addValue(const std::string& column_name, int value);
// same for double, std::string, std::optional<T> (templated)
```
So you can get rid of `if constexpr (...)`

src/io/outputs/include/antares/io/outputs/columns.h
a-zakir: that will turn TypedColumn::data_ into protected member and allow derived classes to modify it

src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -0,0 +1,141 @@
...
+
+    // Access column by index
+    template<typename ColumnType>
+    ColumnType& getColumn(const size_t index)
+    {
+        if (index >= columns_.size())
+        {
+            throw std::out_of_range("Column index out of range");
+        }
+        return dynamic_cast<ColumnType&>(*columns_[index]);
```
flomnes, line 132: This may result in a `nullptr` if the cast fails, so the caller must carefully check for the returned value. May I suggest returning an `IColumn*` instead ?
```cpp
dynamic_cast<DoubleColumn*>(new IntegralColumn()) = nullptr
```
EDIT Since this "find by index" method isn't used, I suggest removing it since it's dangerous & inconvenient.

As a consequence, you may remove `name_to_index_` and replace `columnNames_` and `columns_` by a single `your_preferred_map<std::string, std::unique_ptr<IColumn>>`.

a-zakir: ok

src/io/outputs/include/antares/io/outputs/columns.h
```diff
@@ -0,0 +1,131 @@
...
+
+template<typename U>
+struct is_optional<std::optional<U>>: std::true_type
+{
+};
+
+template<typename T>
+inline constexpr bool is_optional_v = is_optional<T>::value;
+
+static std::string FromDouble(const double value)
```
flomnes, line 53:
```suggestion
static std::string FromDouble(double value)
```

src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+
+    std::vector<std::unique_ptr<IColumn>> columns_;
+    std::vector<std::string> columnNames_;
+    std::unordered_map<std::string, size_t> name_to_index_;
```
flomnes, line 137:
```suggestion
    std::vector<std::pair<std::string, std::unique_ptr<IColumn>>> columns_;
```

flomnes (commented):
Remove `getColumn(index)`

flomnes (approved):
Explain why you keep a `std::map` and a `std::vector`

PR #2972, opened 2025-08-14
Modeler - fix various problems with time & scenarios [ANT-3712]

flomnes:
## Problem: invalid indices starting from the 2nd week in `expr[t-1]` expressions
Starting from the 2nd week, this expression makes the expression parser crash
```
gen1_prop_cost[t-1] + gen2_p * gen2_prop_cost
```
## Solution
In `FillContext`, exhibit two time indices
- local (within the timeblock, starting at 0) used to assign names to variables, and for the time operator `[]`
- global (within the year) used to retrieve timeseries coefficients

## Problem: invalid timeseries coefficients starting from the 2nd week
The coefficients are those used for the 1st week
## Solution
Use the global time index to retrieve timeseries coefficients

## Problem: invalid coefficients starting from the 2nd year (scenarization)
When using scenarized timeseries, 
```yaml
scenario-group: mygroup
...
parameters:
   - id: inflows
       - value: inflows_ts
       - scenario-dependent: true
```
## Solution
Stop using an empty scenario group repository, use the scenario group repository that was parsed at startup.

### DIFF
--- a/src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/ILinearProblemData.h
+++ b/src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/ILinearProblemData.h
@@ -33,26 +33,40 @@ namespace Antares::Optimisation::LinearProblemApi
 class FillContext
 {
 public:
-    FillContext(unsigned first, unsigned last, unsigned year):
-        firstTimeStep(first),
-        lastTimeStep(last),
-        year_{year}
+    FillContext(unsigned localFirstTimeStep,
+                unsigned localLastTimeStep,
+                unsigned globalFirstTimeStep,
+                unsigned globalLastTimeStep,
+                unsigned year):
+        local{localFirstTimeStep, localLastTimeStep},
+        global{globalFirstTimeStep, globalLastTimeStep},
+        year_(year)
     {
     }
 
-    [[nodiscard]] unsigned getFirstTimeStep() const
+    [[nodiscard]] unsigned getLocalFirstTimeStep() const
     {
-        return firstTimeStep;
+        return local.first;
     }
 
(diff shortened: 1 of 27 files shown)


src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/ILinearProblemData.h
```diff
@@ -33,26 +33,40 @@ namespace Antares::Optimisation::LinearProblemApi
 class FillContext
 {
 public:
-    FillContext(unsigned first, unsigned last, unsigned year):
-        firstTimeStep(first),
-        lastTimeStep(last),
-        year_{year}
+    FillContext(unsigned localFirstTimeStep,
```
JasonMarechal25, line 36: Maybe we should declare two structs to avoid confusion and mixing values inadvertently :
struct LocalTimeStep {
 unsigned first;
 unsigned last;
 };
 
 struct gloabalTimeStep {
 unsigned first;
 unsigned last;
 };

JasonMarechal25, line 36: Why not use LocalTimeInterval and GlobalTimeInterval in constructor ?

JasonMarechal25 (approved):

src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/ILinearProblemData.h
flomnes: No need to specify the constructors ?
```cpp
class Member
{
public:
    Member(Arg arg) {}
};

class Aggregate
{
public:
    Aggregate(Arg arg) : member(arg) {}
private:
    Member member;
};
```

PR #2982, opened 2025-08-20
5.2: Export ports in simulation table

payetvin:
- Add port entries
- Fix some issues with 5.1
- Needed updated test files : https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR/pull/77


src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -80,7 +92,17 @@ EvaluationResult EvalVisitor::visit(const Nodes::GreaterThanOrEqualNode* node)
 
 EvaluationResult EvalVisitor::visit(const Nodes::VariableNode* node)
 {
-    return EvaluationResult{context_.getVariableValue(node->value())};
+    if (component_ == nullptr || timeIndex_ == -1)
+    {
+        return EvaluationResult{context_.getVariableValue(node->value())};
+    }
+
+    std::string varName = component_->Id() + "." + node->value() + "_t"
```
a-zakir: scenario is missing. to build the name use the static function 
```
std::string VariableDictionary::buildVariableName(const PartialKey& key,
                                         std::optional<MCYearAndTime::MCYear> mcyear,
                                         std::optional<unsigned int> timestep);
```

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -95,10 +99,88 @@ std::string BuildModelerConstraintName(const std::string& cid,
...
+            TimeBlock tb = ts ? convertTimeStepToBlockTimeIndex(*ts, timeConversionMode)
+                              : TimeBlock{.block = currentBlock,
+                                          .blockTimeIndex = std::nullopt,
+                                          .absoluteTimeIndex = std::nullopt};
+
+            std::optional<double> value = std::nullopt;
+
+            std::map<std::string, double> variables;
+
+            const Antares::Expressions::Visitors::EvaluationContext
```
a-zakir: 1- don't rebuild evalContext for each variable/timeStep/sceneario. 
```
EvaluationContext::EvaluationContext(std::map<std::string, ParameterTypeAndValue> system_parameters,
                                     std::map<std::string, double> variables,
                                     const ILinearProblemData& data,
                                     const IScenario& scenario):
    parameters_types_and_values_(std::move(system_parameters)),
    variables_(std::move(variables)),
    data_(data),
    scenario_(scenario)
{
}
```
2- you could also transform evalcontext ctor arg into const refs and make 
```
EvaluationContext:: std::map<std::string, ParameterTypeAndValue> parameters_types_and_values_; ->EvaluationContext:: **const**  std::map<std::string, ParameterTypeAndValue> **&**parameters_types_and_values_;
 EvaluationContext::std::map<std::string, double> variables_;->  **const** EvaluationContext::std::map<std::string, double>**&** variables_;
```

src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -36,6 +37,17 @@ EvalVisitor::EvalVisitor(EvaluationContext context,
 {
 }
 
+EvalVisitor::EvalVisitor(EvaluationContext context,
+                         Optimisation::LinearProblemApi::FillContext fillContext,
+                         const ModelerStudy::SystemModel::Component* component,
```
a-zakir: timeIndex is only needed to evaluate variable. i think it should not be in ctor. this makes the class difficult to understand. Instead you could evaluate the variable for all timesteps (...) in fillcontext

a-zakir: see how parameter is processed

src/io/outputs/include/antares/io/outputs/SimulationTableGenerator.h
```diff
@@ -84,13 +89,39 @@ void addConstraintEntries(
...
+ * Fill modeler outputs in the simulation table
+ * @param simulationTable the simulation table to fill
+ * @param linearProblem the linear problem containing the optimal solution
+ * @param objectiveValue the overall objective value
+ * @param components list of modeler components
+ * @param dataSeries the input data series
+ * @param fillContext the fill context used to fill the linear problem
+ * @param currentBlock the current block that was optimized
+ * @param timeConversionMode the block to absolute time conversion mode
+ * @param forceScenarioDependency set to true if you want to force the scenario index to be exported
```
flomnes:
```suggestion
 * @param forceExportForScenarioIndex set to true if you want to force the scenario index to be exported
```

flomnes (commented):
Move `Component* component` to `EvaluationContext` ?

pet-mit:
> Move `Component* component` to `EvaluationContext` ?

yes but I prefer to do it in another PR following #3038

src/tests/inmemory-modeler/include/inmemory-modeler.h
```diff
@@ -31,6 +31,7 @@
 #include "antares/study/system-model/component.h"
 #include "antares/study/system-model/model.h"
 
+#include "../../../modeler/FileWriter.h"
```
flomnes: We try to avoid relative include paths as much as possible

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
```diff
@@ -21,6 +21,7 @@
 
 #define WIN32_LEAN_AND_MEAN
 
+#include <absl/container/internal/inlined_vector.h>
```
flomnes: Unused ? Also we don't directly depend on absl for now

flomnes (commented):
If you want to mock `EvaluationContextProvider`, you don't really need an interface class `IEvaluationContextProvider` for that, you can just mark the methods you want to call `virtual`
```cpp
class EvaluationContextProvider
{
public:
    virtual EvaluationContext provide(Args args)
    {
        // Impl
    }
};
class EvaluationContextProviderMock : public EvaluationContextProvider
{
    EvaluationContext provide(Args args) override
    {
        // Mock
    }
};
```

src/tests/inmemory-modeler/include/inmemory-modeler.h
pet-mit: I trust CLion too much :p

pet-mit:
> If you want to mock `EvaluationContextProvider`, you don't really need an interface class `IEvaluationContextProvider` for that, you can just mark the methods you want to call `virtual`
> 
> ```c++
> class EvaluationContextProvider
> {
> public:
>     virtual EvaluationContext provide(Args args)
>     {
>         // Impl
>     }
> };
> class EvaluationContextProviderMock : public EvaluationContextProvider
> {
>     EvaluationContext provide(Args args) override
>     {
>         // Mock
>     }
> };
> ```

It's actually to break the circular dependency between expressions and optim-model-filler:

TimeIndexVisitor needs something that can provide an EvaluationContext object for a given component, in order to be able to visit port connexions.
Rather than making antares::expressions depend on optim-model-filler (where the EvaluationContextProvider implementation is, with dependencies to LinearProblemData and ScenarioGroupRepository), I found it more elegant to create an abstraction for this provider.

flomnes (approved):

PR #2985, opened 2025-08-21
[On-hold][ANT-3625] Use int for keys

JasonMarechal25:
### Redesign completed:
* PartialKey modernized: internage with 32-bit IDs, string_view, noexcept, shared_mutex (shared lock reads), hash based on IDs (portable combination), external API (getComponent/getVariable) preserved.
* [[nodiscard]] additions implied by style, reduced caller-side repeated allocations possible.
* Introduces a slight potential controlled over-allocation (vector reverse in interneur).
### Criticism / points of attention:
* Current cost lookup: intern() creates a std::string(s) for map.find twice (read + write). Double construction can be eliminated by:
	* using an unordered_map with hashing/transparency (std::unordered_map<std::string, id_type, std::hashstd::string_view, std::equal_to<>>), find directly via string_view without allocation.
	* factorize by building the string key once (or move to single location when inserting).
* Lack of memory bounds: the interneur never releases; if a stream creates many distinct identifiers, unbounded growth. Minimum requirements: metrics, thresholds or lifecycle docs.
* Thread-safety: benign races scenario correct (double check pattern covered), but memory visibility order for reverse + map is ensured by lock, OK.
* Hash FullKey uses boost::hash_combine on PartialKey ⇒ fine, but PartialKeyHash is now O(1) on IDs; FullKeyHash could exploit IDs directly to limit string reads (current OK whereas boost::hash_combine(pk) calls our specialized hash so no regression).
* tryGet(): creates a new internage if the pair (component, variable) didn't exist; this pollutes the interneur for negative queries. Enhancements:
* Add a “non-internalizing” lookup method (e.g. findIds(component, variable) which doesn't create if absent) or
Provide a PartialKey::Probe constructor or a factory which doesn't modify the interneur.
* Performance fine: at(timestep) raises exception caught in tryGet; cost avoidable if check boundary (offset + size) and access without exception.
* VectorWithOffset: no explicit initialization of pointers (remain undefined after resize). Initialize to nullptr for safety (facilitates detection).
* Dimensions::getTimesteps() / getScenarioIndices(): default value IntegerInterval{0,0} when absent: size() returns 1 ⇒ consistent but clarify in doc (a “scalar”).
* buildVariableName: concat leads to many allocations; simple optimization: reserve ret.reserve(...).


pet-mit (dismissed):

pet-mit (changes requested):
use english comments instead of french

flomnes (changes requested):
- Methods `tryGet` seem to be unused
- Move function implementations to .cpp files

src/solver/optim-model-filler/PartialKey.cpp
```diff
@@ -25,31 +25,33 @@
...
 {
-    return variable_id;
+    return interner().get(variable_id_);
 }
 
-// PartialKeyhash
+PartialKey::StringInterner& PartialKey::interner() noexcept
+{
+    static StringInterner inst;
+    return inst;
```
flomnes: Global variable here, it may be better to have one instance for each `VariableDictionary`

JasonMarechal25:
New commit sets. Remove use of global variable (static) per @flomnes suggestion.

PR #2987, opened 2025-08-21
Replace binary minus with -= in ReadLinearConstraintVisitor

JasonMarechal25:


flomnes (commented):
LGTM.

Could you please eliminate the code duplication in these 3 functions ?
- `std::vector<LinearConstraint> ReadLinearConstraintVisitor::visit(const GreaterThanOrEqualNode* node)`
- `std::vector<LinearConstraint> ReadLinearConstraintVisitor::visit(const LessThanOrEqualNode* node)`
- `std::vector<LinearConstraint> ReadLinearConstraintVisitor::visit(const EqualNode* node)`

This block looks identical
```cpp
    auto left = linear_expression_visitor_.dispatch(node->left());
    left -= linear_expression_visitor_.dispatch(node->right());

    const auto& leftMinusRightLinearExpression = left.GetLinearExpressions();
```

You may use a templated free function
```cpp
template<class N>
LinearExpressionMap subtractLeftRight(N* node)
{

}
```

flomnes (approved):

PR #2988, opened 2025-08-22
[5.3 part 1/2] Add extra-outputs to models & ID unicity checks

pet-mit:
- Added the ability to define "extra-outputs" at the model level (yaml & cpp class), to define extra outputs that will be exported in the output files (actual export will come in an upcooming PR)
- Fixed the ID unicity check: currently, only some unicity checks are made at local scope (eg 2 constraints can't have the same ID). However, we want IDs to be unique across the whole model, so there is no ambiguity when interpreting expressions or output values.

### DIFF
--- a/src/study/system-model/include/antares/study/system-model/extraOutput.h
+++ b/src/study/system-model/include/antares/study/system-model/extraOutput.h
@@ -25,7 +25,6 @@
 
 #include <antares/expressions/expression.h>
 
-
 namespace Antares::ModelerStudy::SystemModel
 {
 
--- a/src/study/system-model/include/antares/study/system-model/extraOutput.h
+++ b/src/study/system-model/include/antares/study/system-model/extraOutput.h
@@ -1,0 +1,61 @@
+/*
+** Copyright 2007-2025, RTE (https://www.rte-france.com)
+** See AUTHORS.txt
+** SPDX-License-Identifier: MPL-2.0
+** This file is part of Antares-Simulator,
+** Adequacy and Performance assessment for interconnected energy networks.
+**
+** Antares_Simulator is free software: you can redistribute it and/or modify
+** it under the terms of the Mozilla Public Licence 2.0 as published by
+** the Mozilla Foundation, either version 2 of the License, or
+** (at your option) any later version.
+**
+** Antares_Simulator is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** Mozilla Public Licence 2.0 for more details.
+**
+** You should have received a copy of the Mozilla Public Licence 2.0
+** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+*/
+#pragma once
+
+#include <string>
+#include <unordered_map>
+
+#include <antares/expressions/expression.h>
+
+namespace Antares::Expressions::Visitors
+{
+enum class TimeIndex : unsigned int;
+}
+
+namespace Antares::ModelerStudy::SystemModel
+{
+
+/// An extra output expression
+class ExtraOutput
+{
+public:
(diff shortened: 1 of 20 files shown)


flomnes (changes requested):
`checkIdUnicity` is way too strict, as it enforces ID uniqueness at global scope. What we want is actually ID uniqueness at local scope.

pet-mit:
> `checkIdUnicity` is way too strict, as it enforces ID uniqueness at global scope. What we want is actually ID uniqueness at local scope.

we actually want IDs to be unique across the model. 
for example, if a parameter and a variable have the same ID "x", how do you know how to interpret contraint expression "5 * x < some_var" ? is x the parameter or the variable ?

src/study/system-model/include/antares/study/system-model/extraOutput.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include <string>
+#include <unordered_map>
+
+#include <antares/expressions/expression.h>
+
+namespace Antares::Expressions::Visitors
+{
+enum class TimeIndex : unsigned int;
+}
```
flomnes: Unused
```suggestion
```

src/study/system-model/include/antares/study/system-model/extraOutput.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Antares::ModelerStudy::SystemModel
+{
+
+/// An extra output expression
+class ExtraOutput
```
flomnes, line 32: This class looks very much like `class Constraint`

src/study/system-model/include/antares/study/system-model/model.h
```diff
@@ -121,13 +128,17 @@ class ModelBuilder
...
-
     ModelBuilder& withConstraints(std::vector<Constraint>&& constraints);
     ModelBuilder& withPortFieldDefinitions(std::vector<PortFieldDefinition>&& portFieldDefinitions);
+    ModelBuilder& withExtraOutputs(std::vector<ExtraOutput>&& extraOutputs);
+    void checkThatIdIsNotUsed(const std::string& id);
+    Model build();
 
 private:
     Model model_;
+    std::vector<std::string> attribute_ids_;
```
flomnes:
```suggestion
    // List of IDs used internally to check for uniqueness of IDs at component level
    std::vector<std::string> attribute_ids_;
```

src/study/system-model/model.cpp
```diff
@@ -38,6 +39,18 @@ std::size_t PortFieldKeyHash::operator()(const PortFieldKey& input) const
...
+void ModelBuilder::checkThatIdIsNotUsed(const std::string& id)
+{
+    if (std::ranges::find(attribute_ids_, id) != attribute_ids_.end())
+    {
+        std::string modelId = model_.id_;
+        reset();
+        throw Error::RuntimeError("Model \"" + modelId + "\" contains multiple objects with ID \""
+                                  + id + "\".");
+    }
+    attribute_ids_.emplace_back(id);
```
flomnes: For frequent searches, you might want to use a `std::unordered_set` for  `attribute_ids_`

src/study/system-model/include/antares/study/system-model/extraOutput.h
pet-mit: it's basically the same. but I think that we will do less checks on it because we allow more operators in its expression

src/study/system-model/model.cpp
```diff
@@ -38,6 +39,18 @@ std::size_t PortFieldKeyHash::operator()(const PortFieldKey& input) const
     return seed;
 }
 
+void ModelBuilder::checkThatIdIsNotUsed(const std::string& id)
+{
+    if (std::ranges::find(attribute_ids_, id) != attribute_ids_.end())
```
flomnes:
```suggestion
    if (attribute_ids_.contains(id))
```

flomnes (approved):

PR #2998, opened 2025-08-27
[On-hold] Vectorize variable access

JasonMarechal25:
### What changed

Keys changed from composite FullKey objects to compact 32‑bit VarIndex values.
Linear expressions no longer map with heavy string/struct keys; they reference variables by numeric indices.
How indices are obtained

A VariableDictionary resolves a variable “handle” once (componentId + variableId).
Given a handle and a (year, timeStep), it returns a stable VarIndex.
Constant variables map to a single index; time‑dependent variables map to one index per local time step.

### How a LinearExpression is stored

Offset is a double.
Terms are appended into a vector of pairs [VarIndex, coefficient] (fast write path, duplicates allowed).
On demand, a lazy cache aggregates duplicates into an unordered_map<VarIndex, double> for quick coefficient lookup.
Cache is invalidated on any modification and rebuilt only when needed.

### Why it is faster

No hashing/comparison of large composite keys; VarIndex is a small integer.
Appending to a vector is cache‑friendly and O(1) amortized.
Aggregation happens once per read, proportional to the number of terms actually present.
Downstream, expressions can be scattered into a dense std::vector<double> (size = number of variables) with O(1) index access and excellent memory locality.</double>

### Typical workflow

Build expressions by pushing terms [index, value] into the vector.
When reading or assembling into the solver, materialize the cache or directly scatter into a dense coefficient vector using indices.
The solver receives indices and values without any dictionary or key lookups on the hot path.
Complexity and memory

Append term: O(1) amortized.
Materialize/aggregate: O(T) where T is the number of raw terms.
Dense scatter: O(Nz) where Nz is the number of non‑zeros.
Memory per term is minimal (one uint32 + one double), versus larger FullKey objects.

### API impact

coefPerVar() becomes coefPerIndex() with the same semantics.
Tests and call sites must stop using FullKey and use VarIndex from the dictionary, or assert on the sole entry of coefPerIndex() when appropriate.
No more MCYearAndTime in the hot path; year/time are inputs to index resolution.

### Caveats

The lazy cache is mutable and not thread‑safe across threads; keep each LinearExpression confined to a thread or materialize before sharing.
VarIndex is 32‑bit; ensure total variable count fits in uint32_t.


src/solver/optim-model-filler/include/antares/solver/optim-model-filler/VariableDictionary.h
```diff
@@ -140,60 +163,113 @@ class Dimensions
...
-    const TwoIndexVectorByYear& operator[](const PartialKey& k) const; ///< Internal helper
-    mutable StringToIdMapper mapper_;                                  ///< For naming
+    // Legacy storage (kept for compatibility)
+    HashMapVector storageOfAddedMipVariables_;
+    // New optimized storage
+    std::vector<Value> storage_; ///< O(1) access by index
+    std::unordered_map<PartialKey, KeyHandle, PartialKeyHash>
+      keyHandleMap_; ///< PartialKey -> Handle resolution
+
+    mutable StringToIdMapper mapper_; ///< For naming
```
flomnes: For simplicity, I suggest keeping a small number of containers, instead of the 3 presented here.

flomnes (commented):
I suggest keeping a small number of containers, instead of

flomnes (commented):
LGTM, need to check if having multiple structures `byIndex`, `byKey` induces overhead

PR #3001, opened 2025-08-28
Delete ortools mpSolver 

a-zakir:


flomnes:
Please use smart pointers instead for `MPSolver` instances

a-zakir:
that’s what she said

PR #3009, opened 2025-08-29
Clean some warnings

JasonMarechal25:


src/expressions/antlr-interface/ExprParser.cpp
flomnes, line 1: It's best not to edit generated files
```cpp
// Generated from Expr.g4 by ANTLR 4.13.2
```

flomnes (approved):
- Please rebase on develop before merging
- Don't edit ANTLR-generated source files

PR #3011, opened 2025-08-29
Fix for scenarization & port issues [ANT-3750]

flomnes:
Pass the scenario repository to `EvaluationContext` for TS numbers, just like `ILinearProblemData` for TS.


src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
```diff
@@ -180,10 +181,13 @@ TimeDependentLinearExpression ReadLinearExpressionVisitor::visit(const PortField
...
-        const EvaluationContext connectedComponentEvalContext(component->getParameterValues(),
-                                                              {},
-                                                              evalContext_.data(),
-                                                              evalContext_.scenario());
+        const EvaluationContext connectedComponentEvalContext(
+          component->getParameterValues(),
+          {},
+          evalContext_.data(),
+          Antares::Optimisation::gScenarioGroupRepository->scenario(
+            component->getScenarioGroupId()));
```
flomnes: The important  change is here, where we get the appropriate `IScenario&` instance.

src/expressions/include/antares/expressions/visitors/EvaluationContext.h
```diff
@@ -95,6 +97,9 @@ class EvaluationContext
      */
     std::map<std::string, double> variables_;
     const Optimisation::LinearProblemApi::ILinearProblemData& data_;
-    const Optimisation::LinearProblemApi::IScenario& scenario_;
+
+public:
+    // TODO getter
```
pet-mit, line 102: complete this todo

JasonMarechal25 (approved):

pet-mit (changes requested):
in general it needs a test and I think we can simplify EvaluationContext design. I'll propose something

pet-mit:
> in general it needs a test and I think we can simplify EvaluationContext design. I'll propose something
check out
https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3043

PR #3015, opened 2025-08-29
"mc-" + int does not append to string

JasonMarechal25:


src/solver/ts-generator/hydro.cpp
```diff
@@ -284,7 +285,7 @@ bool GenerateHydroTimeSeries(Data::Study& study, Solver::IResultWriter& writer)
               [&writer, &progression](const Data::Area& area)
               {
                   const int precision = 0;
-                  std::string mcYear = "mc-" + 0;
+                  std::string mcYear = fmt::format("mc-{}", 0);
```
flomnes: No need for `fmt::format` here
```suggestion
                  const std::string mcYear = "mc-0";
```

flomnes (approved):

PR #3016, opened 2025-08-29
Throw exception properly

JasonMarechal25:


src/solver/lps/LpsFromAntares.cpp
```diff
@@ -21,6 +21,9 @@
 
 #include "antares/solver/lps/LpsFromAntares.h"
 
+#include "../../libs/antares/exception/include/antares/exception/LoadingError.hpp"
```
flomnes: Unused ?

src/solver/lps/LpsFromAntares.cpp
```diff
@@ -21,6 +21,9 @@
...
+#include "../../libs/antares/exception/include/antares/exception/RuntimeError.hpp"
```
flomnes:
```suggestion
#include <antares/exception/RuntimeError.hpp>
```

flomnes (changes requested):

src/solver/lps/LpsFromAntares.cpp
```diff
@@ -43,7 +45,8 @@ const WeeklyDataFromAntares& LpsFromAntares::weeklyData(WeeklyProblemId id) cons
     auto it = weeklyProblems.find(id);
     if (it == weeklyProblems.end())
     {
-        return WeeklyDataFromAntares(); // TODO Better error handling
+        throw Antares::Error::RuntimeError(
```
flomnes, line 50: You probably forgot to call `fmt::format` here. Don't forget to include it at the top.

flomnes (approved):

PR #3027, opened 2025-09-01
Fix ignoring return value of function declared with 'nodiscard'

JasonMarechal25:


flomnes (commented):
May I suggest `(void)noDiscardFunction()` ?

flomnes (approved):

PR #3035, opened 2025-09-02
[ANT-3638] Remove H cost from Ov cost

JasonMarechal25:
* Remove Hydro hourly values from overallCost computation


flomnes (commented):
That's a good start 👍🏻. Please also
- Remove file src/solver/variable/include/antares/solver/variable/economy/hydroCost.h
- Apply the same modification in src/solver/variable/include/antares/solver/variable/economy/overallCostCsr.h

flomnes:
Slight misunderstanding, please restore the H. COST column and leave it untouched for now. Sorry

flomnes (approved):

PR #3038, opened 2025-09-02
ANT-3750 - Alternative implementation

flomnes:


src/expressions/CMakeLists.txt
```diff
@@ -98,6 +99,7 @@ target_link_libraries(expressions
         linear-problem-api
         PRIVATE
         Antares::antares-study-system-model
+        Antares::optim-model-filler
```
pet-mit:
```suggestion
        Antares::antares-study-system-model
```
should be enough?

src/study/system-model/include/antares/study/system-model/system.h
```diff
@@ -42,6 +42,11 @@ class System
         return id_;
     }
 
+    std::unordered_map<std::string, Component>& Components()
```
pet-mit, line 45: dangerous indeed.
maybe load the scenario builder first, then send it as an argument to the system constructor instead ? then the scenario will be know in construction time

src/expressions/include/antares/expressions/visitors/EvaluationContext.h
```diff
@@ -96,5 +86,6 @@ class EvaluationContext
     std::map<std::string, double> variables_;
     const Optimisation::LinearProblemApi::ILinearProblemData& data_;
     const Optimisation::LinearProblemApi::IScenario& scenario_;
+    const Antares::ModelerStudy::SystemModel::Component* component_;
```
pet-mit: incomplete work (you have to add it in ctor & use it)

pet-mit (changes requested):

src/study/system-model/include/antares/study/system-model/system.h
flomnes: done

src/expressions/CMakeLists.txt
```diff
@@ -98,6 +99,7 @@ target_link_libraries(expressions
         linear-problem-api
         PRIVATE
         Antares::antares-study-system-model
+        Antares::antares-study-system-model
```
pet-mit: oops sorry, actually my last suggestion duplicated the line
```suggestion
        PRIVATE
        Antares::antares-study-system-model
```

src/expressions/include/antares/expressions/visitors/EvaluationContext.h
flomnes: done (not needed in the final solution)

src/io/inputs/yml-system/converter.cpp
```diff
@@ -95,14 +95,16 @@ static const SystemModel::Model& getModel(const std::vector<SystemModel::Library
...
+  const YmlSystem::Component& c,
+  const std::vector<SystemModel::Library>& libraries,
+  const Optimisation::ScenarioGroupRepository& scenarioRepository)
 {
     const auto [libraryId, modelId] = splitLibraryModelString(c.model);
 
     const SystemModel::Model& model = getModel(libraries, libraryId, modelId);
 
-    SystemModel::ComponentBuilder component_builder;
+    SystemModel::ComponentBuilder component_builder(scenarioRepository);
```
pet-mit, line 107: to stay coherent with the original builder design, you should instead add a "withScenario" function to ComponentBuilder

src/study/system-model/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ target_include_directories(antares-study-system-model
 target_link_libraries(antares-study-system-model
         PUBLIC
         Antares::expressions
+	Antares::optim-model-filler
```
pet-mit, line 38: if you apply my previous suggestion, system-model won't depend on optim-model-filler

src/study/system-model/include/antares/study/system-model/component.h
```diff
@@ -72,6 +80,11 @@ class Component
         return data_.model;
     }
 
+    const Antares::Optimisation::LinearProblemApi::IScenario* getScenario() const
```
pet-mit, line 83: actually, you need something like : 
```suggestion
    const Antares::Optimisation::LinearProblemApi::IScenario* getScenario(int MCyear) const
```
because the component should know which timeseries index to associate to every MC year

src/study/system-model/include/antares/study/system-model/system.h
pet-mit: then you can remove this new getter

pet-mit (changes requested):

flomnes:
See #3011

PR #3040, opened 2025-09-03
ANT-3697 : fix contraint on weekly hydro generation amount (ANT-3697)

guilpier-code:
Fixes [ticket ANT-3697](https://gopro-tickets.rte-france.com/browse/ANT-3697)

**Cause** : 
```c++
std::vector<double> v = {1, 1, 1, 1, 0.5};

double sum = std::accumulate(v.begin(), v.end(), 0);
std::cout << "sum = " << sum << std::endl;

sum = std::accumulate(v.begin(), v.end(), 0.);
std::cout << "sum = " << sum << std::endl;
```
**Output** 
> sum = 4
> sum = 4.5

In algorithm **std::accumulate**, type of init value (here **0** [int] or **0.** [double]) impacts return type, and an unwanted conversion into int occur.


flomnes (approved):
Well done

PR #3041, opened 2025-09-04
Logs time used to build modeler problem [ANT-3759]

payetvin:


flomnes (commented):
Consider extracting `class TimeMeasurement` and using it in both cases (display a duration in milliseconds)

a-zakir (approved):

flomnes (approved):

flomnes:
Logging in milliseconds seems better in my opinion

PR #3043, opened 2025-09-04
Evaluation context provider from Component [ANT-3750]

pet-mit:
Added a new class "EvaluationContextProvider" that allows to easily generate an EvaluationContext for a component.
ScenarioGroupRepository and LinearProblemData should be given to this class' ctor ; thus they are considered constant during the simulation.
WARNING : this is a change from before, since LinearProblemData has been an argument of LinearProblemFiller::addVariables / Constraints / Objective.
However, given our usage of LinearProblemData, I think we can consider it constant from now on (hence the TODO in the code)


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -133,14 +135,19 @@ MPSolver* fillAndGetMpSolver(LegacyOrtoolsLinearProblem& ortoolsProblem,
...
     VariableDictionary variableDictionary;
     std::vector<std::unique_ptr<Optimisation::ComponentFiller>> componentFillers;
-    ComponentToAreaConnectionFiller componentToAreaConnectionFiller(problemeHebdo,
-                                                                    variableDictionary);
+    ComponentToAreaConnectionFiller componentToAreaConnectionFiller(
+      problemeHebdo,
+      variableDictionary,
+      *problemeHebdo->linear_problem_data_,
+      *problemeHebdo->scenarioGroupRepository);
+    // TODO what happens if problemeHebdo->scenarioGroupRepository doesn't exist ?
```
flomnes: Same as `linear_problem_data_`, SIGSEGV

flomnes (approved):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
pet-mit: I think we should clean this up, might as well do it in this PR. I'll look into it tomorrow

flomnes (approved):

PR #3050, opened 2025-09-05
Feature/restudy model [ANT-3284] [ANT-3743]

a-zakir:


src/io/inputs/yml-system/converter.cpp
```diff
@@ -80,7 +80,7 @@ static const SystemModel::Model& getModel(const std::vector<SystemModel::Library
                                           const std::string& modelId)
 {
     auto lib = std::ranges::find_if(libraries,
-                                    [&libraryId](const auto& l) { return l.Id() == libraryId; });
+                                    [&libraryId](auto& l) { return l.Id() == libraryId; });
```
flomnes, line 83: `decltype(l)::Id()` no more `const` ?

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -87,19 +87,41 @@ IMipConstraint* ComponentToAreaConnectionFiller::getBalanceConstraint(ILinearPro
...
 {
     // Contribution is added to the left-hand side of the constraint
     // We invert the sign bc modeler is in "gen>0, load<0" convention
     // legacy constraint is in "gen<0, load>0" convention
-    for (const auto& [varKey, coef]: expression.coefPerVar())
+    // for (const auto& [varKey, coef]: coeffPerVar)
+    // {
+    //     auto* var = modelerVariableDictionary_[varKey];
+    //     areaBalanceConstraint->setCoefficient(var, -coef);
+    // }
```
flomnes: This breaks the "hybrid" mode

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpressionEigen.h
```diff
@@ -0,0 +1,88 @@
...
+        return coeffs_;
+    }
+
+    void setCoefPerVar(const Eigen::SparseMatrix<double>& coeffs)
+    {
+        coeffs_ = coeffs;
+    }
+
+    void setCol(int colIndex, const Eigen::SparseVector<double>& col);
+    void setRow(int rowIndex, const Eigen::SparseVector<double>& row);
```
flomnes, line 108: Do we need both methods ? Ideally we would only need one

a-zakir: as now only setCol is used. TODO

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
a-zakir: ~old code replaced few lines below~

src/io/inputs/yml-system/converter.cpp
a-zakir: to be reverted; i was on the idea to transform model into mutable

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
a-zakir: i forgot to multiply the coef by -1

PR #3053, opened 2025-09-08
Modeler 5.3: Add extra outputs to simulation table

payetvin:


src/io/outputs/include/antares/io/outputs/SimulationTableGenerator.h
```diff
@@ -104,6 +104,26 @@ void addPortEntries(ISimulationTable& simulationTable,
...
+
+void addEntriesForNode(ISimulationTable& simulationTable,
+                       const Antares::Optimisation::LinearProblemApi::FillContext& fillContext,
+                       const Antares::ModelerStudy::SystemModel::Component& component,
+                       unsigned currentBlock,
+                       const TimeConversionMode& timeConversionMode,
+                       std::optional<unsigned> scenario,
+                       bool forceExportForScenarioIndex,
+                       const Antares::Optimisation::EvaluationContextProvider& contextProvider,
+                       const std::string& cid,
```
flomnes: `cid` isn't very explicit

src/tests/resources/modeler/10_1/case_description.txt
```diff
@@ -0,0 +1,7 @@
+Test 10_1 : expected extra-outputs
+- spill_cost_contribution = [0 0 0 0 0]
+- ens_cost_contribution = [0 0 0 0 200000]
+- load_parameter = [20,40, 60, 80, 110]
+- generation_cost = [40, 80, 120, 160, 200]
+- squared_generation_cost [1600, 6400, 14400, 25600, 40000]
+- total_generation_cost = 600
```
guilpier-code, line 7: It's a bit short for a test description.
- What's the intention of the test ?
- To reach the goal of the test, what do we (at least roughly) do in the **yaml** files ?
- We should highlight input data and expected output. Here they seem all mixed.

src/tests/resources/modeler/10_1/input/system.yml
```diff
@@ -0,0 +1,74 @@
...
+
+   # Study case 10_1 : to test extra-output features
+   # Small test case with one node (1 thermal cluster, 1 one load)
+
+system:
+
+  id: case_10_1
+  description: To be added
+  components:
+    - id: base_zone
```
guilpier-code, line 21: **base_zone** ==> **base_area** ?
Everybody in the team is used to **area** instead of **zone**. 
Hence my suggestion.

src/tests/resources/modeler/10_1/input/system.yml
```diff
@@ -0,0 +1,74 @@
...
+      - id: spillage_cost
+        time-dependent: false
+        scenario-dependent: false
+        value: 0
+      - id: ens_cost
+        time-dependent: false
+        scenario-dependent: false
+        value: 20000
+    
+    - id: load_base_zone
```
guilpier-code, line 33: **load_base_zone** : **load** should be enough.
A component (like load) does not know it's related to a component **area**. 
So there should be no reference to an area here.

src/tests/resources/modeler/10_1/input/model-libraries/library-eo.yml
```diff
@@ -0,0 +1,154 @@
...
+      extra-outputs:
+        - id: generation_cost
+          expression: marginal_generation_cost * generation
+        - id: squared_generation_cost #test to manage a quadratic expression
+          expression: marginal_generation_cost * generation * marginal_generation_cost * generation
+        - id: total_generation_cost
+          expression: sum(marginal_generation_cost * generation)
+
+
+    - id: short-term-storage
```
guilpier-code, line 101: It seems we don't need  model **short-term-storage** in test **10_1**.
Let's remove it.

src/tests/resources/modeler/10_1/input/system.yml
guilpier-code, line 21: **Question** : what do we mean by **base** in **base_zone** ?

src/tests/resources/modeler/10_1/case_description.txt
```diff
@@ -0,0 +1,7 @@
+Test 10_1 : expected extra-outputs
+- spill_cost_contribution = [0 0 0 0 0]
```
guilpier-code, line 2: **spill_cost_contribution** : 
- it's unclear art first, but it's an output
- contribution to what ?
- rename into **output_spillage_cost** ?

src/tests/resources/modeler/10_1/case_description.txt
```diff
@@ -0,0 +1,7 @@
...
+- ens_cost_contribution = [0 0 0 0 200000]
```
guilpier-code, line 3: **ens_cost_contribution** : 
- as above : see comments on **spill_cost_contribution**

src/tests/resources/modeler/10_1/case_description.txt
```diff
@@ -0,0 +1,7 @@
...
+- load_parameter = [20,40, 60, 80, 110]
```
guilpier-code, line 4: **load_parameter** : 
When we describe a test, shouldn't we talk with business vocabulary only ?
If this is true, **load** seems enough. **parameter** is more technical and refers to **yaml** input structure.

src/tests/resources/modeler/10_1/input/model-libraries/library-eo.yml
```diff
@@ -0,0 +1,154 @@
...
+        - id: spill_cost_contribution
+          expression: spillage_cost * spillage
+        - id: ens_cost_contribution
+          expression: ens_cost * unsupplied_energy
+
+    - id: load
+      parameters:
+        - id: load
+          time-dependent: true
+          scenario-dependent: false
```
guilpier-code, line 56: There's a repetition of **load** keyword.
Maybe we could have :
```yaml
- id: load
      parameters:
        - id: time-series
          time-dependent: true
          scenario-dependent: false
```
or : 
```yaml
- id: load
      parameters:
        - id: values
          time-dependent: true
          scenario-dependent: false
```

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -61,11 +60,11 @@ TI updateTimeIndexIfShouldForceScenario(TI timeIndex, bool forceExportForScenari
     return forceExportForScenarioIndex ? timeIndex | TI::VARYING_IN_SCENARIO_ONLY : timeIndex;
 }
```
guilpier-code, line 60:
```c++
#include "../../modeler/FileWriter.h"
```
is useless. To be removed.

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -239,6 +238,48 @@ void addObjectiveValue(ISimulationTable& simulation,
...
+
+    auto handle = [&](std::optional<unsigned> ts, std::optional<unsigned> scenIdx)
+    {
+        TimeBlock tb = ts ? convertBlockTimeStepToAbsoluteTimeStep(*ts,
+                                                                   timeConversionMode,
+                                                                   currentBlock)
+                          : TimeBlock{.block = currentBlock + 1,
+                                      .blockTimeIndex = std::nullopt,
+                                      .absoluteTimeIndex = std::nullopt};
+        auto val = ts.has_value() ? value.valuesAsVector()[ts.value()] : value.valueAsDouble();
```
guilpier-code: **_Cosmetic :_** 
```c++
 auto val = ts.has_value() ? ...
```
could be replaced with (as done above) : 
```c++
auto val = ts ? ...
```

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -23,9 +23,8 @@
 #include <optional>
 
 #include <antares/solver/optim-model-filler/VariableDictionary.h>
```
guilpier-code, line 25: About function **handleDependingOnTimeIndex(...)** : 
In **std::function** argument : 
```c++
const std::function<void(std::optional<unsigned> ts, std::optional<unsigned> scenIdx)>& handle
```
the **scenIdx** is useless. 

Please see [this PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3057) for a fix

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -239,6 +236,48 @@ void addObjectiveValue(ISimulationTable& simulation,
...
+
+    auto handle = [&](std::optional<unsigned> ts, std::optional<unsigned> scenIdx)
+    {
+        TimeBlock tb = ts ? convertBlockTimeStepToAbsoluteTimeStep(*ts,
+                                                                   timeConversionMode,
+                                                                   currentBlock)
+                          : TimeBlock{.block = currentBlock + 1,
+                                      .blockTimeIndex = std::nullopt,
+                                      .absoluteTimeIndex = std::nullopt};
+        auto val = ts ? value.valuesAsVector()[ts.value()] : value.valueAsDouble();
```
a-zakir: valueAsVector is too expensive. do it once or add a const ref getter in EvalResult

payetvin: done on PR #3057

src/tests/resources/modeler/10_1/case_description.txt
flomnes: I think `expected extra-outputs` makes it clear that all of these are expected extra outputs

flomnes (approved):

PR #3056, opened 2025-09-09
Declare nested namespace in one line

JasonMarechal25:
Simpler

Doing it once to clear future PRs from modification


flomnes (approved):
Remove Antares copyright notices from Yuni's source (Yuni is a 3rd party component, not part of Antares)

PR #3057, opened 2025-09-11
Modeler > simulation table : trial for simplification

guilpier-code:


src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -273,11 +273,11 @@ void addEntriesForNode(ISimulationTable& simulationTable,
                                   .output = outputName,
                                   .absolute_time_index = tb.absoluteTimeIndex,
                                   .block_time_index = tb.blockTimeIndex,
-                                  .scenario_index = scenIdx,
+                                  .scenario_index = scenario,
```
pet-mit: in the original code, scenIdx here could be nullopt if (idxType == VARYING_IN_TIME_ONLY) or (idxType == CONSTANT_IN_TIME_AND_SCENARIO). this is not the case in what you propose so the behaviour may change. Are you sure it's ok?

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -196,7 +182,9 @@ void addConstraintEntries(ISimulationTable& simulationTable,
                        .dispatch(modelConstr.expression().RootNode());
         idxType = updateTimeIndexIfShouldForceScenario(idxType, forceExportForScenarioIndex);
 
-        auto handle = [&](std::optional<unsigned> ts, std::optional<unsigned> scenIdx)
+        scenario = isScenarioDependant(idxType) ? scenario : std::nullopt;
```
flomnes: The provided `scenario` is already an `std::optional`

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -257,27 +246,43 @@ void addEntriesForNode(ISimulationTable& simulationTable,
 
     TI idxType = Antares::Expressions::Visitors::TimeIndexVisitor(component, contextProvider)
                    .dispatch(rootNode);
+
     idxType = updateTimeIndexIfShouldForceScenario(idxType, forceExportForScenarioIndex);
 
-    auto handle = [&](std::optional<unsigned> ts, std::optional<unsigned> scenIdx)
+    scenario = isScenarioDependant(idxType) ? scenario : std::nullopt;
```
flomnes: The provided `scenario` is already an `std::optional`

src/io/outputs/SimulationTableGenerator.cpp
guilpier-code: Yes it is, but in order to maintain the previous code behavior, I need to to this. 
The  previous code used to do the same (locally update the **scenario**, depending **idxType**), more or less

a-zakir:
~i'd rename this pr "Modeler > simulation table : trial for **complexification**"~ humor

a-zakir:
> i'd rename this pr "Modeler > simulation table : trial for **complexification**"


There were four distinct cases; now there are only two. Why is this happening?

flomnes:
Many conflicts, more to come

PR #3063, opened 2025-09-15
Catch modeler loading error [ANT-3837]

payetvin:


src/libs/antares/study/load.cpp
```diff
@@ -232,6 +230,11 @@ void Study::loadModelerComponents()
         modelerInput_ = std::make_unique<Modeler::Data>(Solver::LoadFiles::loadAll(folder));
         checkModelerDataCompatibility();
     }
+    catch (const Error::LoadingError& e)
+    {
+        logs.error() << "Error while loading modeler components:";
+        logs.error() << e.what();
```
flomnes: One single log line
```suggestion
        logs.error() << "Error while loading modeler components: " << e.what();
```

flomnes (approved):

PR #3067, opened 2025-09-17
Multi storage remix : more tests [ANT-3175]

guilpier-code:


src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+#include <boost/test/unit_test.hpp>
+
+#include "antares/solver/simulation/remix-storage/create-storage-for-remix.h"
+#include "antares/solver/simulation/remix-storage/remix-utils.h"
+#include "antares/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.h"
+
+using namespace Antares::Solver::Simulation;
+
+template<unsigned int size>
```
flomnes: What is size ?

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
```diff
@@ -0,0 +1,266 @@
...
+        UnsupE.assign(size, 0.);
+        Spillage.assign(size, 0.);
+        DTG_MRG.assign(size, 0.);
+    }
+
+    void runRemixStorageAlgorithm()
+    {
+        Load = TotaGenWithoutStorage + UnsupE + sts_1.withdrawal + sts_2.withdrawal;
+
+        storagesForRemix.clear();
```
flomnes, line 71: Not needed, since the fixture object is used only once

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+        withdrawal.assign(size, 0);
+        injection.assign(size, 0);
+        levels.assign(size, 0.);
+        pmax.assign(size, std::numeric_limits<double>::max());
+        inflows.assign(size, 0.);
+        ovf.assign(size, 0.);
+        lowRuleCurve.assign(size, 0.);
+        upRuleCurve.assign(size, capacity);
+    }
```
flomnes: Initialize members with their constructors
```suggestion
    STS_setup(): withdrawal(size, 0),
    injection(size, 0),
    // etc.
    {
    }
```

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
```diff
@@ -0,0 +1,266 @@
...
+{
+    std::vector<double> Load(5);
+    std::vector<double> TotaGenWithoutStorage(5);
+    std::vector<double> Spillage(5);
+    std::vector<double> DTG_MRG(5);
+
+    // Creating sts_1 of size 5
+    std::vector<double> unsupE_1(5, 0.);
+    STS_setup<5> sts_setup_1;
+    auto sts_1 = sts_setup_1.createSTS(unsupE_1);
```
flomnes: `STS_setup` does nothing but hold data


```suggestion
    auto sts_1 = createSTS(unsupE_1);
```

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    // Creating sts_1 of size 3
+    std::vector<double> unsupE_2(3, 0.);
```
flomnes, line 114: Why give only `unsupE` as an input data ? What about inflows, etc. ?

flomnes (commented):
Please clarify input & output data, it's unclear what is an input and what is an output.

Ideally
```cpp
output = runAlgorithm(input);
assert(someFunction(output));
// or
assert(someFunction(input, output));
```

Feel free to create structs as necessary, for example `struct Input` and `struct Output`.

flomnes (changes requested):
Please see comments

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
guilpier-code: It's not, sorry.
Indeed, there are tests where we run the algorithm twice : between the 2 runs, only the value of a parameter changes and we're interested in the behavior difference.
We can change this and make 2 shorter tests instead if you like, but these tests would form a pair.

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
guilpier-code: Each **STS** (for remix) has its own **inflows**, **withdrawal**, ...
So the holder for a particular STS holds these vectors / scalars internally.
But **unsupE** is shared among all STS contributing to the remix algorithm.

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
guilpier-code: Done

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
guilpier-code: I changed : **size**  ==> **nb_hours**

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
guilpier-code: I renamed **STS_setup** into **STS_holder**, **STS_setup** being inaccurate indeed

guilpier-code:
> Please clarify input & output data, it's unclear what is an input and what is an output.
> 
> Ideally
> 
> ```c++
> output = runAlgorithm(input);
> assert(someFunction(output));
> // or
> assert(someFunction(input, output));
> ```
> 
> Feel free to create structs as necessary, for example `struct Input` and `struct Output`.

Suppose method **runRemixStorageAlgorithm()** disappears. Indeed, it only exists because we can have 2 runs of the algo in a test (but as I said above, we can change this).
Anyway, in this case, we would have the real and raw algorithm called directly from each tests : 
```c++
BOOST_FIXTURE_TEST_CASE(some_test, InputFixture<5>)
{
     // Set the input
     shavePeaksByRemixingStorageGen(Load, UnsupE, Spillage, DTG_MRG, storagesForRemix);
     // make checks
}
```
This algorithm itself does not separate **input** and **output** : most of the data it takes as argument is **input**, 
but some other (**unsupE** and each STS **withdrawals**) are also **output** (there are references and are changed while algo runs).

So **input** and **output** are kind of mixed in the algorithm itself.
Separate them for tests does not seem natural, at least to me.
So what do you think ?

flomnes:
> > Please clarify input & output data, it's unclear what is an input and what is an output.
> > Ideally
> > ```c++
> > output = runAlgorithm(input);
> > assert(someFunction(output));
> > // or
> > assert(someFunction(input, output));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Feel free to create structs as necessary, for example `struct Input` and `struct Output`.
> 
> Suppose method **runRemixStorageAlgorithm()** disappears. Indeed, it only exists because we can have 2 runs of the algo in a test (but as I said above, we can change this). Anyway, in this case, we would have the real and raw algorithm called directly from each tests :
> 
> ```c++
> BOOST_FIXTURE_TEST_CASE(some_test, InputFixture<5>)
> {
>      // Set the input
>      shavePeaksByRemixingStorageGen(Load, UnsupE, Spillage, DTG_MRG, storagesForRemix);
>      // make checks
> }
> ```
> 
> This algorithm itself does not separate **input** and **output** : most of the data it takes as argument is **input**, but some other (**unsupE** and each STS **withdrawals**) are also **output** (there are references and are changed while algo runs).
> 
> So **input** and **output** are kind of mixed in the algorithm itself. Separate them for tests does not seem natural, at least to me. So what do you think ?

I think it would be best to wrap inputs & outputs into appropriate structures. You may use references (`std::vector<double>&`, `std::span<double>`) as needed.

flomnes (approved):

PR #3068, opened 2025-09-18
Remove compile option unsuported on Clang

JasonMarechal25:
feliminate-unused-debug-symbols is a GCC only option


src/cmake/common-settings.cmake
```diff
@@ -70,11 +70,13 @@ else ()
     set(ANTARES_INSTALLER_EXTENSION "-debug")
 
     if (NOT WIN32)
```
guilpier-code: **Question** : 
Could we not remove the : 
```cmake
if (NOT WIN32)
    # case GNU
    # case other compiler, but not GNU
```
?
And have simply : 
```cmake
if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
    ...
if (CMAKE_CXX_COMPILER_ID STREQUAL "CLANG")
   ...
```

guilpier-code (approved):

PR #3069, opened 2025-09-18
Fix clang Wunused-local-typedef

JasonMarechal25:
It seems Clang is right on this and it's a bug on GCC.

Using a macro that does nothing for clang to erase the warning


guilpier-code (commented):
Sorry, I can't imagine the problem here.
A description of the issue would be appreciated, thanks.
I just wonder why we need to modify **Yuni**.

flomnes (approved):

PR #3070, opened 2025-09-18
Disable warning in antlr

JasonMarechal25:


src/expressions/antlr-interface/CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ Set(SRCS
 
 add_library(${PROJ} ${SRCS})
 add_library(Antares::${PROJ} ALIAS ${PROJ})
+if (NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
```
guilpier-code, line 18: **Question** : what if we compile on **Windows** ? What's happening then ?

flomnes: Nothing

guilpier-code (approved):

PR #3074, opened 2025-09-18
Remove Unused parameter

JasonMarechal25:


src/solver/simulation/include/antares/solver/simulation/remix-storage/create-storage-for-remix.h
```diff
@@ -16,7 +16,6 @@ std::shared_ptr<IStorageForRemix> makeHydroForRemix(std::vector<double>& generat
                                                     const std::vector<double>& pump,
                                                     const double initLevel,
                                                     const double reservoirCapacity,
-                                                    const double pumpEfficiency,
```
guilpier-code, line 19: Oops, my bad !
We need this **pumpEfficiency** and and the code looks like it doesn't.
I corrected this in [PR 3067](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3067)
More precisely, see [this commit](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3067/commits/670af6177be04f11166d7aaec81cb6cf7b87bbe1)
Thank you to point this !
I guess this PR does should be deleted now, what do you think ?

guilpier-code (changes requested):

PR #3085, opened 2025-09-22
Use simple structures for linear expression representation instead of Eigen's sparse matrices

flomnes:
TODO
- [ ] Assemble duplicate terms before adding contributions
- [ ] Constraints
- [ ] Split files + CMakefiles


src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -369,40 +755,16 @@ void ComponentFiller::addObjective(
     }
 
     const auto& solverVariables = optimEntityContainer_.getVariables();
-    Optimization::ReadLinearExpressionVisitor visitor(
-                                                      ctx,
-                                                      component_,
-                                                      optimEntityContainer_);
+    V::Visitor visitor(optimEntityContainer_, component_, 168); // TODO TimeSteps
```
flomnes: Here we use the newly-introduced visitor instead of `Optimization::ReadLinearExpressionVisitor`

a-zakir:
**feature/restudy-model** branch has been updated with **develop** and includes enhancements for the simulation table.
In my opinion to avoid conflict resolution, you should checkout a new branch from **feature/restudy-model**  and "paste" your changes into it.

PR #3086, opened 2025-09-22
Modeler 5.4: Evaluate comparisons [ANT-3620]

payetvin:


src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -73,6 +74,21 @@ class EvaluationResult
         return evaluateBinaryOperation(right, std::multiplies<>());
     }
 
+    EvaluationResult operator==(const EvaluationResult& right) const
+    {
+        return evaluateBinaryOperation(right, std::equal_to<>());
+    }
+
+    EvaluationResult operator<=(const EvaluationResult& right) const
```
a-zakir, line 82: if only one comparison (if any element in vect[i] vs vect[i] or vect[i] vs double) is false then return false. You don't have to call evaluateBinaryOperation or create specializations for this operator

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -73,6 +74,21 @@ class EvaluationResult
...
+    {
+        return evaluateBinaryOperation(right, std::less_equal<>());
+    }
+
+    EvaluationResult operator>=(const EvaluationResult& right) const
```
a-zakir, line 87: same

a-zakir (approved):

PR #3089, opened 2025-09-23
Use boost::iostreams::mapped_file_source to improve CSV loading performance [ANT-3915]

a-zakir:


src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -155,7 +181,8 @@ DataSeriesRepository DataSeriesRepoImporter::importFromDirectory(const std::file
         {
             continue;
         }
-        auto timeSeriesSet = std::make_unique<TimeSeriesSet>(importFromFile(entry, csvSeparators));
+        auto timeSeriesSet = std::make_unique<TimeSeriesSet>(entry.path().stem().string(),
+                                                             readCSV(entry, csvSeparator));
         repo.addDataSeries(std::move(timeSeriesSet));
     }
```
guilpier-code, line 210: We could even have : 
```c++
for (auto paths = std::filesystem::directory_iterator{path};
     const auto& entry: paths | pathFilter | hasRightExtension)
{
    auto timeSeriesSet = std::make_unique<TimeSeriesSet>(entry.path().stem().string(),
                                                             readCSV(entry, csvSeparator));
    ...
}
```
or something like that

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,142 @@
...
+        row.push_back(val);
+        ptr = next;
     }
-    return result;
+
+    return row;
 }
 
-static TimeSeriesSet matrixToTimeSeriesSet(std::string id, std::vector<std::vector<double>> matrix)
+std::vector<std::vector<double>> readCSV(const std::filesystem::path& filename, char sep)
```
guilpier-code: This function could even be **static**

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,142 @@
...
+    if (!file.is_open())
     {
-        return timeSeriesSet;
+        throw std::runtime_error("Failed to open file: " + fileName);
     }
-    int nSets = matrix[0].size();
-    for (int i = 0; i < nSets; ++i)
+
+    std::vector<std::vector<double>> columns;
+    const char* start = file.data();
```
guilpier-code, line 148:
```c++
const char* start = file.data();
```
**start** is **const** but we change it a bit further. This is strange, but ok.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,142 @@
...
+            newline = end;
+        }
+        // Handle Windows line endings
+        size_t line_len = newline - start;
+        if (line_len > 0 && start[line_len - 1] == '\r')
+        {
+            line_len--;
+        }
+
+        auto row = parseNumbersFast(start, start + line_len, sep, columns.size(), fileName);
```
guilpier-code: **parseNumbersFast** ==> **parseRowOfNumbers** would be better

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,142 @@
...
+    std::vector<std::vector<double>> columns;
+    const char* start = file.data();
+    const char* end = file.data() + file.size();
+
+    while (start < end)
     {
-        std::vector<double> set;
-        set.reserve(nTimesteps);
-        for (int j = 0; j < nTimesteps; ++j)
+        const char* newline = static_cast<const char*>(memchr(start, '\n', end - start));
```
guilpier-code:
```c++
const char* newline = ...
```
It seems that **newline** points actually at the end of a line. 
If it's true, why not call it **endLine** ?

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,142 @@
...
-    if (nTimesteps == 0)
+    // MappedFile file(fileName.c_str());
+    // Check file size first
+    std::error_code ec;
+    auto sz = std::filesystem::file_size(filename, ec);
+    if (ec || sz == 0)
+    {
+        return {}; // empty or inaccessible
+    }
+    const auto& fileName = filename.string();
```
guilpier-code, line 139: **fileName** : is mainly defined to supply exceptions raised with the current file name.
So instead of using **fileName**  in **readCSV(...)** or passing it further deep, you could try (if possible) to catch the exception thrown, and then rethrow it with the same message enriched with the file name.
To be more specific, we could try something like : 
```c++
try
{
    auto timeSeriesSet = std::make_unique<TimeSeriesSet>(entry.path().stem().string(),
                                                         readCSV(entry, csvSeparator));
}
catch(std::exception& e)
{
    std::string msg = entry.path().stem().string() + " : " + e.what();
    // rethrow the same exception type with the new msg
}
```
If this is not doable easily, then forget it.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
flomnes: Well that's the difference between `const char*` and `char * const`

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
flomnes: We had that but it resulted to a compilation error in one of our toolchains, can't remember which

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
a-zakir: I see what you mean. The only reason fileName is supplied to parseNumbersFast and parseOneDouble is to comply with existing tests, which are verifying exception messages.
I think it's not the best use of <code>try/catch</code> to 'just' append the error message.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
a-zakir: row of number does not make sense, what about parseRow ?

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
a-zakir: done

src/optimisation/linear-problem-data-impl/timeSeriesSet.cpp
```diff
@@ -21,6 +21,7 @@
 
 #include "antares/optimisation/linear-problem-data-impl/timeSeriesSet.h"
 
+#include <absl/strings/internal/str_format/extension.h>
```
flomnes: unused

src/optimisation/linear-problem-data-impl/timeSeriesSet.cpp
```diff
@@ -35,6 +36,16 @@ TimeSeriesSet::TimeSeriesSet(std::string name, unsigned int height):
...
+    tsSet_(std::move(tsSet))
+{
+    if (!tsSet_.empty())
+    {
+        height_ = tsSet_[0].size();
+    }
+}
+
 void TimeSeriesSet::add(const std::vector<double>& ts)
 {
```
flomnes, line 49: At line 55 `std::move` will fall back to a copy (ts is a const ref)

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,149 @@
...
 {
-    std::vector<std::vector<double>> result;
-    std::ifstream infile(path, std::ios_base::binary | std::ios_base::in);
-    if (!infile.is_open())
+    std::vector<double> row;
+    if (capacity > 0)
     {
-        throw std::invalid_argument("Could not open file " + path.filename().string());
+        row.reserve(capacity);
     }
```
flomnes: If `capacity=0`, this is a no-op.
```suggestion
    row.reserve(capacity);
```

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,158 @@
...
-    if (nTimesteps == 0)
+    // MappedFile file(fileName.c_str());
+    // Check file size first
+    std::error_code ec;
+    auto sz = std::filesystem::file_size(filename, ec);
+    if (ec || sz == 0)
     {
-        return timeSeriesSet;
+        return {}; // empty or inaccessible
     }
```
flomnes, line 138: IMO we should also throw an exception here, but display a more specific error message

the `return {}` is dangerous, it could lead to invisible errors.

My suggestion 
```cpp
if (sz == 0) return {}; // empty file -> empty matrix
if (ec) throw std::runtime_error("Error reading file " + filename + " (" + ec.message() +")");
```

src/optimisation/linear-problem-data-impl/timeSeriesSet.cpp
flomnes:
```suggestion
```

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -19,116 +19,158 @@
...
+                                            + ": rows have inconsistent number of columns");
+            }
         }
-        result.push_back(row);
     }
-    return result;
+    if (rowIndex != 0 && colIndex != columns.size())
+    {
+        throw std::invalid_argument(errorMessagePrefix
+                                    + ": rows have inconsistent number of columns");
```
flomnes: I suggest choosing a different error message to make debugging easier (see exception at l. 109)

src/optimisation/linear-problem-data-impl/timeSeriesSet.cpp
a-zakir: this method is not modified by this pr

flomnes (approved):

PR #3092, opened 2025-09-25
Modeler 6.2a : Data structures for optimization configuration [ANT-3885]

payetvin:


src/modeler/optimConfig/optimConfig.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+OptimConfigBuilder& OptimConfigBuilder::withObjectives(std::vector<Objective>&& objectives)
+{
+    config_.modelDecomposition.objectives = std::move(objectives);
+    return *this;
+}
+
+OptimConfig OptimConfigBuilder::build()
+{
+    return config_;
```
a-zakir: std::move ? you also need to check Ids uniqueness

src/modeler/optimConfig/optimConfig.cpp
```diff
@@ -0,0 +1,49 @@
...
+    for (const auto& model: models_)
+    {
+        modelIds[model.id()]++;
+    }
+
+    for (const auto& [id, count]: modelIds)
+    {
+        if (count > 1)
+        {
+            throw Error::RuntimeError("OptimConfig contains multiple models with ID \"" + id
```
JasonMarechal25: We will throw at the first duplication. Maybe stack messages and throw once for all ids

JasonMarechal25 (changes requested):

JasonMarechal25 (approved):

src/modeler/optimConfig/include/antares/modeler/optimConfig/optimConfig.h
```diff
@@ -0,0 +1,151 @@
...
+    ModelDecomposition modelDecomposition_;
+};
+
+class OptimConfig
+{
+public:
+    OptimConfig(std::vector<Model> models):
+        models_(std::move(models))
+    {
+        checkDuplicateModelIds();
```
guilpier-code, line 137: **checkDuplicateModelIds()** is used only internally. It should be **private**.

src/modeler/optimConfig/optimConfig.cpp
```diff
@@ -0,0 +1,49 @@
...
```
guilpier-code: I can't tell why **RuntimeError** exists : it inherits from `std::runtime_error` but does nothing more : no additional data member nor behavior.
Why not throw a `std::runtime_error` instead ?
We should remove this class.

src/tests/src/modeler/optimConfig/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+include(${CMAKE_SOURCE_DIR}/tests/macros.cmake)
+
+add_boost_test(optimConfig
```
guilpier-code: **optimConfig** ==> **test-on-modeler-optim_config** (would be more specific and clear). 
I don't know about Ubuntu, but on Windows and MSVC, it would be more easily identified and clear inside solution explorer.

src/tests/src/modeler/optimConfig/optimConfigBuilder.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#define WIN32_LEAN_AND_MEAN
+
+#define BOOST_TEST_MODULE optim config
+
+#include <boost/test/unit_test.hpp>
+
+#include "antares/modeler/optimConfig/optimConfig.h"
+
+BOOST_AUTO_TEST_SUITE(test_optim_config_builder)
```
guilpier-code: There is no **builder** here, we only call constructors. Builder is a design pattern.
Maybe **test-modeler-config-creation**  would be a better for source file and test suite

JasonMarechal25 (approved):

guilpier-code (approved):

PR #3093, opened 2025-09-25
Feature/restudy model

flomnes:


src/io/inputs/yml-system/converter.cpp
```diff
@@ -80,7 +80,7 @@ static const SystemModel::Model& getModel(const std::vector<SystemModel::Library
                                           const std::string& modelId)
 {
     auto lib = std::ranges::find_if(libraries,
-                                    [&libraryId](const auto& l) { return l.Id() == libraryId; });
+                                    [&libraryId](auto& l) { return l.Id() == libraryId; });
```
flomnes, line 83: revert

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -77,35 +84,44 @@ void addVariableEntries(ISimulationTable& simulationTable,
...
+    const auto& solverVariables = optimEntityContainer.getVariables();
+    const auto& optimComponent = optimEntityContainer.getOptimComponent(component.Index());
+    const auto& modelVariablesGlobalIndices = optimComponent.modelVariablesGlobalIndices;
+    unsigned variableLocalIndex = 0;
+    for (const auto& modelVar: component.getModel()->Variables())
     {
         bool scenDep = modelVar.IsScenarioDependent();
         bool timeDep = modelVar.isTimeDependent();
-
+        // this is the global Index
```
flomnes: remove

src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -281,39 +306,66 @@ void addEntriesForNode(ISimulationTable& simulationTable,
...
+                    const OptimEntityContainer& optimEntityContainer,
                     unsigned currentBlock,
                     const TimeConversionMode& timeConversionMode,
                     std::optional<unsigned> scenario,
-                    bool forceExportForScenarioIndex,
-                    const Antares::Optimisation::EvaluationContextProvider& contextProvider)
+                    bool forceExportForScenarioIndex)
 {
-    const auto& componentId = component.Id();
+    const auto& cid = component.Id();
```
flomnes, line 307:
```suggestion
    const auto& componentId = component.Id();
```

src/modeler-optimisation-container/include/antares/modeler-optimisation-container/OptimEntityContainer.h
```diff
@@ -0,0 +1,165 @@
...
+
+    // void IncrementVariableGLobalIndex()
+    // {
+    //     ++variableGlobalIndex_;
+    // }
+
+    void IncrementConstraintGLobalIndex()
+    {
+        ++constraintGlobalIndex_;
+    }
```
flomnes: unused ?

src/modeler-optimisation-container/include/antares/modeler-optimisation-container/OptimEntityContainer.h
```diff
@@ -0,0 +1,165 @@
...
+    std::vector<unsigned int> variableStartColumn_;
+    std::vector<OptimComponent> optimComponents_;
+    unsigned int variableGlobalIndex_ = 0;
+    //---
+    std::vector<LinearProblemApi::IMipConstraint*> constraints_;
+    std::vector<unsigned int> constraintStartLine_;
+    unsigned int constraintGlobalIndex_ = 0;
+    LinearProblemApi::ILinearProblem& linearProblem_;
+    const LinearProblemApi::ILinearProblemData* data_;
+    const ScenarioGroupRepository* scenarioGroupRepository_;
```
flomnes: Templated class, 1 for constraints, 1 for variables ?

src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
```diff
@@ -18,7 +18,7 @@
  * You should have received a copy of the Mozilla Public Licence 2.0
  * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
  */
-
+#if 0
```
guilpier-code, line 21: What do you want to do with these tests ?
Remove them completely or adapt them ?

src/tests/src/solver/optim-model-filler/test_linearExpression.cpp
```diff
@@ -18,7 +18,7 @@
  * You should have received a copy of the Mozilla Public Licence 2.0
  * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
  */
-
+#if 0
```
guilpier-code, line 21: Dangerous not to unit test classes **LinearExpression** and **TimeDependentLinearExpression**.
I supposed that it's on the way to be done while I writing this comment.

src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
flomnes: yes, work in progress

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ReadLinearExpressionVisitor.h
```diff
@@ -21,61 +21,289 @@
...
+    Antares::Optimization::TimeDependentLinearExpression visit(
+      const Nodes::VariableNode* node) override
+    {
+        const auto variableStart = optimEntityContainer_.getVariableStartColumn(component_,
+                                                                                node->Index());
+        if (node->timeIndex() == Antares::Optimisation::TimeIndex::CONSTANT_IN_TIME_AND_SCENARIO)
+        {
+            Antares::Optimization::LinearExpression out;
+            out.addVariable(variableStart, 1);
+            return Antares::Optimization::TimeDependentLinearExpression(std::move(out));
```
guilpier-code, line 145: We could avoid using **LinearExpression** in this file.
Indeed, we create **LinearExpression** only because we want to create (and return) a **TimeDependentLinearExpression**.

We could do this either by defining constructors for class **TimeDependentLinearExpression** : 
```c++
if (node->timeIndex() == Antares::Optimisation::TimeIndex::CONSTANT_IN_TIME_AND_SCENARIO)
{
    return Antares::Optimization::TimeDependentLinearExpression(variableStart);
}
```
or defining a factory public functions like : 
```c++
if (node->timeIndex() == Antares::Optimisation::TimeIndex::CONSTANT_IN_TIME_AND_SCENARIO)
{
    return Antares::Optimization::createConstantLinearExpression(variableStart);
}
```

If this change is possible and suitable, we should also remove : 
```c++
#include <antares/solver/optim-model-filler/LinearExpression.h>
```

guilpier-code: To be honest, I wondered if class **LinearExpression** could be totally private to class **TimeDependentLinearExpression**. 
I mean : would it be possible to create and use instances of **LinearExpression** only inside class **TimeDependentLinearExpression** ?
But it does not seem to be possible, because of an outside use of **LinearExpression** in **objective**, which is a non time depending expression.

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -18,162 +18,82 @@
...
+    using const_iterator = std::vector<std::pair<int, double>>::const_iterator;
+    const_iterator begin() const;
+    const_iterator end() const;
+    const std::pair<int, double>& operator[](std::size_t) const;
+    std::size_t size() const;
+
+private:
+    bool hasCoefs() const;
+    std::vector<std::pair<int, double>> coefs_;
+    double constant_ = 0.;
```
guilpier-code, line 55: About : 
```c++
double constant_ = 0.;
```
This is a bit confusing :  do we mean **offset_** ?
It seems that by **constant()**, we actually mean **offset()**, which seems to be a common and accepted business term. 
**Constant** is vague and we wonder what it is in the context of a linear expression

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,163 +19,420 @@
...
+{
+void removeDuplicates(std::vector<std::pair<int, double>>& v)
+{
+    // Most constraints have 1 or 2 terms, so we handle them here without sort & loops
+    switch (v.size())
+    {
+    case 0:
+    case 1:
+        return;
+    case 2:
```
guilpier-code, line 37: Are we sure that making a special treatment for **case 2** is useful ?
I mean : should the vector not be treated as if it was vector of size > 1 ?

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,163 +19,420 @@
...
 
+#include "antares/solver/optim-model-filler/LinearExpression.h"
+
+#include <algorithm>
 #include <stdexcept>
 
-#include <antares/solver/optim-model-filler/LinearExpression.h>
+namespace
+{
+void removeDuplicates(std::vector<std::pair<int, double>>& v)
```
guilpier-code, line 29: Do we remove duplicates or do we merge duplicates ?

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,163 +19,420 @@
...
     }
-    return result;
+    constant_ *= factor;
+
+    return *this;
+}
+
+LinearExpression& LinearExpression::operator+=(const LinearExpression& other)
+{
+    coefs_.reserve(coefs_.size() + other.coefs_.size());
```
guilpier-code, line 111: **Question** : is it possible that **other** contains terms with the same indices than the current linear expression ?

Same question for other operators.

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,163 +19,420 @@
...
+    {
+        // Multiplying two symbolic expressions would give quadratic terms,
+        // which this representation cannot hold.
+        throw std::runtime_error("Quadratic term detected");
+    }
+    else if (!hasCoefs() && !other.hasCoefs())
     {
-        terms_.emplace_back(k, v);
+        // constant * constant
+        constant_ *= other.constant_;
```
guilpier-code, line 171: It seems like : 
```c++
constant_ *= other.constant_;
```
is done for each case. 
It should be done once.

Applying this change, the case : 
```c++
else if (!hasCoefs() && !other.hasCoefs())
{
    // constant * constant
    constant_ *= other.constant_;
}
```
will disappear.

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -18,162 +18,82 @@
...
-    void invalidate()
-    {
-        am_I_valid_ = false;
-    }
-
-    double offset_ = 0;
-    mutable FullKeyMap unique_terms_; // aggregated unique sums
-    std::vector<RawTerm> terms_;      // may contain duplicates
-    mutable bool am_I_valid_ = false;
+    std::variant<LinearExpression, std::vector<LinearExpression>> v_;
```
guilpier-code, line 97: **Naive question** :
Why isn't **v_**  a simple `std::vector<LinearExpression>` instead of a `std::variant` ?
In other words, if case a **TimeDependentLinearExpression** contains a single **LinearExpression**,  couldn't it be contained in **v_** defined as a `std::vector<LinearExpression>`  ?
Where is the problem that prevent from doing that ?

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,163 +19,420 @@
...
-const FullKeyMap& LinearExpression::coefPerVar() const
+TimeDependentLinearExpression::TimeDependentLinearExpression(LinearExpression&& expr):
+    v_(std::move(expr))
 {
-    materialize();
-    return unique_terms_;
 }
 
-LinearExpression& LinearExpression::operator+=(const LinearExpression& other)
+void TimeDependentLinearExpression::expandTo(std::size_t nbTimesteps)
```
guilpier-code, line 234: It seems that **expandTo** should be **private**, as it is called only internally

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,163 +19,420 @@
...
+    }
+    throw std::runtime_error("Invalid variant");
 }
 
-double LinearExpression::offset() const
+TimeDependentLinearExpression& TimeDependentLinearExpression::operator+=(
+  const TimeDependentLinearExpression& other)
 {
-    return offset_;
+    if (other.size() > size())
```
guilpier-code, line 363: **Questions** : 
Is it possible that **other** and current **TimeDependentLinearExpression** are both vectors of different size ?

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,163 +19,420 @@
...
+    {
+        expandTo(other.size());
+    }
+    int t = 0;
+    for (auto& expr: *this)
+    {
+        expr *= other[t];
+        t++;
+    }
```
guilpier-code, line 424: If we need the index **t**, maybe better to keep the classic loop : 
```c++
for (std::size_t t = 0; t < size(); ++t)
{
    ...
}
```
as done above (see operator `+=`).

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ReadLinearConstraintVisitor.h
```diff
@@ -23,17 +23,22 @@
 
 #include <limits>
```
guilpier-code, line 24: It seems we don't need : 
```c++
#include <limits>
```
here, but rather in **ReadLinearConstraintVisitor.cpp**

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ReadLinearConstraintVisitor.h
```diff
@@ -23,17 +23,22 @@
...
 
-#include <antares/expressions/visitors/EvaluationContext.h>
 #include <antares/expressions/visitors/NodeVisitor.h>
 
+#include "LinearExpression.h"
```
guilpier-code, line 28:
```c++
#include "LinearExpression.h"
```
not needed. To be removed.

src/solver/optim-model-filler/ReadLinearConstraintVisitor.cpp
```diff
@@ -29,14 +29,16 @@
 using namespace Antares::Expressions::Nodes;
 using namespace Antares::ModelerStudy::SystemModel;
 
-namespace Antares::Optimization
+namespace Antares::Optimisation
 {
 
 ReadLinearConstraintVisitor::ReadLinearConstraintVisitor(
-  const Optimisation::EvaluationContextProvider& evalContextProvider,
+
```
guilpier-code, line 36: Remove empty lines **36** and **39**

a-zakir (approved):

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -18,162 +18,82 @@
  * You should have received a copy of the Mozilla Public Licence 2.0
  * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
  */
-
 #pragma once
```
guilpier-code, line 21: This file should be split in 2 parts.
One for class **LinearExpression** and one for **TimeDependentLinearExpression**

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
flomnes: yes, it is the constant term, as `b` in `a*x+b`

src/solver/optim-model-filler/LinearExpression.cpp
flomnes: Yes it slightly improves performance because we don't call `std::sort`, etc.

src/solver/optim-model-filler/LinearExpression.cpp
flomnes: The answer is yes, and that's why we need to remove (merge) duplicates before using the linear expression.

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
flomnes: Yes, it would be simpler indeed. Will look into it

src/solver/optim-model-filler/LinearExpression.cpp
flomnes: In practice, no

src/solver/optim-model-filler/LinearExpression.cpp
guilpier-code: Then we should rename  **removeDuplicates** as well as **LinearExpression :: removeDuplicateCoefficients**

src/solver/optim-model-filler/LinearExpression.cpp
guilpier-code: From what I understand, each time we use a binary operator (on 2 linear expressions), we have to merge duplicates.
So I would expect a call to this function inside these operators, no ?

src/tests/src/solver/optim-model-filler/test_linearExpression.cpp
flomnes: See #3120

src/solver/optim-model-filler/LinearExpression.cpp
flomnes: No, we merge at the end, before using the coefficients to build a constraint/objective

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
guilpier-code: Should we not rename **constant()** into **offset()** and **constant_** into **offset_** ?

src/solver/optim-model-filler/LinearExpression.cpp
guilpier-code: in this case (case 2), sort should be very fast and we will be quickly out of this function.
Besides, a linear expression of size 2 is unlikely to be met in the context of Antares Simulator.
I suppose the size is 1 or 168

PR #3105, opened 2025-09-29
Fix Sonar linking issues

flomnes:
For now, I just added an extra step to upload the full build directory in order to investigate the issue locally. More to come...


flomnes:
See #3103

PR #3118, opened 2025-10-02
Fix tests

a-zakir:


src/tests/src/io/outputs/testSimulationTable.cpp
```diff
@@ -744,8 +759,65 @@ struct MockMipSolution: IMipSolution
...
+            {
+                for (int t = 0; t < fillContext.getLocalNumberOfTimeSteps(); ++t)
+                {
+                    linearProblem->addVariable(0, 0, false, "");
+                }
+            }
+            else
+            {
+                linearProblem->addVariable(0, 0, false, "");
+                ;
```
flomnes:
```suggestion
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -79,7 +79,9 @@ static void fillModelerComponents(
   Modeler::Data* modelerData,
   OptimEntityContainer& optimEntityContainer)
 {
-    for (const auto& component: modelerData->system->Components())
+    const auto& components = modelerData->system->Components();
+    optimEntityContainer.reserveOptimComponents(components.size());
```
flomnes, line 83: Merge `addFromSystemComponent` and `reserveOptimComponents` into a single method, adding all components at once.
```suggestion
    optimEntityContainer.addFromSystemComponents(components);
```

src/optimisation/linear-problem-mpsolver-impl/include/antares/optimisation/linear-problem-mpsolver-impl/linearProblem.h
```diff
@@ -85,8 +93,8 @@ class OrtoolsLinearProblem: public LinearProblemApi::ILinearProblem
     operations_research::MPObjective* objective_;
     operations_research::MPSolverParameters params_;
 
-    std::vector<std::unique_ptr<OrtoolsMipVariable>> variables_;
-    std::vector<std::unique_ptr<OrtoolsMipConstraint>> constraints_;
+    std::vector<std::unique_ptr<LinearProblemApi::IMipVariable>> variables_;
+    std::vector<std::unique_ptr<LinearProblemApi::IMipConstraint>> constraints_;
```
flomnes, line 96: I know there is no such thing as `dynamic_cast<std::vector<Base>>(std::vector<Derived>)`, but having the derived type here may add vtable lookups in most function calls (setCoefficient, etc.)

a-zakir: i'm not sure i understand, there is no need for dynamic cast derived->base. In fact you can't avoid virtual call if everywhere pointers of base classes are used.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
flomnes: Done in the [next PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3120)

flomnes (approved):

PR #3122, opened 2025-10-06
Don't write simulation tables when --no-output is specified [ANT-3883]

payetvin:


flomnes (changes requested):
Create a `NullWriter` (writer that does nothing)

flomnes (commented):
Not sure this is the appropriate behavior. Simulation tables are not debug files.

Also, we shouldn't use/need a `noOutput()` method (the writer's role is to write, it shouldn't have getters).

PR #3124, opened 2025-10-07
Fixes after review for #3093

flomnes:
- Split LinearExpression.h / TimeDependentLinearExpression.h, same for .cpp
- Remove `std::variant`
- Rename `removeDuplicateCoefficients` -> `mergeDuplicateCoefficients`
- Other minor remarks


src/solver/optim-model-filler/include/antares/solver/optim-model-filler/LinearExpression.h
```diff
@@ -55,45 +54,4 @@ class LinearExpression final
     double constant_ = 0.;
```
guilpier-code, line 54: Should we not rename **constant()** into **offset()** and **constant_** into **offset_** ?

src/solver/optim-model-filler/LinearExpression.cpp
```diff
@@ -19,14 +19,14 @@
  * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
  */
 
-#include "antares/solver/optim-model-filler/LinearExpression.h"
-
 #include <algorithm>
 #include <stdexcept>
 
+#include "antares/solver/optim-model-filler/TimeDependentLinearExpression.h"
```
guilpier-code, line 25:
```c++
#include "antares/solver/optim-model-filler/TimeDependentLinearExpression.h"
```
I'm surprised that this header is included here, it should be **LinearExpression.h** instead.
class **LinearExpression** should know nothing about **TimeDependentLinearExpression**.

src/solver/optim-model-filler/ReadLinearConstraintVisitor.cpp
```diff
@@ -36,7 +36,6 @@ ReadLinearConstraintVisitor::ReadLinearConstraintVisitor(
```
guilpier-code, line 36: line **36** to be removed

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -0,0 +1,193 @@
...
+        return v_[idx];
+    }
+}
+
+const LinearExpression& TimeDependentLinearExpression::operator[](std::size_t idx) const
+{
+    if (size() == 1)
+    {
+        return v_[0];
+    }
```
guilpier-code: About : 
```c++
if (size() == 1)
{
    return v_[0];
}
```
I see your point / trick.
We should at least comment the code to avoid confusion to any reader of this code, which can be us.

**Questions** : 
Am I right when I say : 
- When 2 time dependent linear expressions have a different size, it's because one is **really time dependent** and the other is actually **constant in time** (contains only one **LinearExpression**) ?
- If 2 expressions are both really time dependent, they necessarily have the same size (168 time steps) ?

If the answer is yes to both questions, it should be more clear in the code, especially the class should have a method `bool isConstantInTime()`.
About expanding an expression, I suppose that we do it in case we needed it only, that is when have to combine (add or multiply) it with another one of different size..

src/solver/optim-model-filler/include/antares/solver/optim-model-filler/ReadLinearExpressionVisitor.h
guilpier-code: Please leave a comment [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3093#discussion_r2406462731), and say if you agree or disagree

guilpier-code (approved):

PR #3127, opened 2025-10-08
Simplify SpatialAggregateAll chain template

JasonMarechal25:
Use variadic template to generate spatial agregate type list.
Code is easier to read and diff to compare


flomnes:
Old & new type are indeed identical at compile time
```cpp
#include <type_traits>
static_assert(std::is_same<VariablesPerSetOfAreasOld, VariablesPerSetOfAreas>::value);
```

flomnes (approved):

PR #3128, opened 2025-10-08
Action to delete cache after merge

JasonMarechal25:
This action is triggered when a pr on develop is closed. This action delete all cache entry of the branch being merged.

Why : cache space is limited. Even if there is a default strategy of cache management, once a branch is merged its different cache entry don't serve any purpose anymore.
Instead of deleting potentially useful cache entry to make space for new entries, we opt to delete now useless cache entry beforehand and free space for new entries.


.github/workflows/delete_cache_after_merge.yml
```diff
@@ -0,0 +1,25 @@
+name: Delete cache after PR is merged
+on:
+  pull_request:
+    branches:
+    - develop
```
flomnes: This should be applied also on PRs to other branches (think PR in the PR)
```suggestion
```

.github/workflows/delete_cache_after_merge.yml
```diff
@@ -0,0 +1,25 @@
...
+    permissions:
+      # `actions:write` permission is required to delete caches
+      #   See also: https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-a-github-actions-cache-for-a-repository-using-a-cache-id
+      actions: write
+      contents: read
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+    - run: |
+        gh cache list --json id,key,ref | jq -r '.[] | select(.ref == "refs/heads/${{ github.event.pull_request.head.ref }}") | .id' | xargs -I {} gh cache delete {}
```
flomnes: 1. This is not very verbose. I'd like to know what cache entries were found & deleted, so I suggest that you `cat` the output of the 1st command. This would allow us to make sure that this job is useful
2. You could you option --ref to filter by ref. For example this seems to work
```
gh cache list  --json id --ref refs/heads/feature/var1 -q '[.[].id] | join(" ")'
1364693885 1364703825 1364693662 1364704382
```

.github/workflows/delete_cache_after_merge.yml
```diff
@@ -0,0 +1,25 @@
...
+  pull_request:
+    branches:
+    - develop
+    types:
+    - closed
+
+env:
+  GH_TOKEN: ${{ github.token }}
+  GH_REPO: ${{ github.repository }}
+  BRANCH: refs/pull/${{ github.event.pull_request.number }}/merge
```
flomnes: `BRANCH` is unused

```suggestion
```

flomnes (approved):

PR #3129, opened 2025-10-08
Simplify template chain economy sorties (area variables)

JasonMarechal25:
Use variadic template to generate area type list.
Code is easier to read and diff to compare


src/solver/variable/include/antares/solver/variable/commons/compose-all.h
```diff
@@ -0,0 +1,53 @@
...
+#include "antares/solver/variable/variable.h" // for Container::EndOfList
+
+namespace Antares
+{
+namespace Solver
+{
+namespace Variable
+{
+namespace Common
+{
```
flomnes: merge namespace

src/solver/variable/include/antares/solver/variable/economy/links.h
```diff
@@ -52,6 +52,8 @@ typedef FlowLinear             // Flow linear
           <>>>>>>>>>
     VariablePerLink;
 
+// Make Links a template alias to be used in Common::ComposeAll
```
flomnes, line 55: It looks like we could also use `Common::ComposeAll` to define `VariablePerLink`

flomnes (approved):

src/solver/variable/include/antares/solver/variable/economy/all.h
```diff
@@ -72,94 +81,64 @@
...
-                                                                                // detail
-                                         <NbOfDispatchedUnits         // Number of Units Dispatched
-                                          <NbOfDispatchedUnitsByPlant // Number of Units
-                                                                      // Dispatched by plant
-                                           <ProfitByPlant
-                                            // Links
-                                            <Variable::Economy::Links // All links
-                                             >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
-    VariablesPerArea;
+using VariablesPerArea = Common::ComposeAll<OverallCost,
```
guilpier-code, line 90: Seems better, simpler indeed !
This kind of list of types at compile time wears a name in meta programmation : static list.
So **ComposeAll** could even be renamed into **StaticList** or **MakeStaticList**

src/solver/variable/include/antares/solver/variable/commons/compose-all.h
```diff
@@ -0,0 +1,53 @@
...
+
+// Variadic composition of class templates: ComposeAll<Head, Tail...>::type =
```
guilpier-code: Shouldn't the word **recursive** be used as well in the comment ? 
It seems **variadic** is the way to reach the purpose of a recursive composition, but the composition is first recursive, isn't it ?

src/solver/variable/include/antares/solver/variable/economy/links.h
JasonMarechal25: You would think so, but it's not that easy. We need a terminal instanced type.
It's in my plan to also simplify links and ideally factorize template composer, but in later PRs

PR #3132, opened 2025-10-08
Interprétation des valeurs d'usage [ANT-3894]

payetvin:


src/libs/antares/study/parts/hydro/container.cpp
```diff
@@ -799,10 +799,16 @@ double getWaterValue(const double& level /* format : in % of reservoir capacity
                      const Matrix<double>& waterValues,
                      const uint day)
 {
-    assert((level >= 0. && level <= 100.) && "getWaterValue function : invalid level");
+    assert((level >= 0. && level <= 100. + 1e-6) && "getWaterValue function : invalid level");
```
JasonMarechal25: assert are not triggered in release mode.
Maybe we should convert this to a proper error

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -460,7 +458,7 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
                 problem.CaracteristiquesHydrauliques[k].WeeklyWaterValueStateRegular
                   = getWaterValue(nivInit * 100 / area.hydro.reservoirCapacity,
                                   area.hydro.waterValues,
-                                  weekFirstDay);
+                                  weekFirstDay + 6);
```
JasonMarechal25: Please used named variable instead of magic number "6".
If Variable name isn't enough add a comment why "6"

src/solver/simulation/sim_calcul_economique.cpp
```diff
@@ -510,9 +508,7 @@ void SIM_RenseignementProblemeHebdo(const Study& study,
                 for (uint layerindex = 0; layerindex < 100; layerindex++)
                 {
                     problem.CaracteristiquesHydrauliques[k].WaterLayerValues[layerindex]
-                      = 0.5
-                        * (area.hydro.waterValues[layerindex][weekFirstDay + 7]
-                           + area.hydro.waterValues[layerindex + 1][weekFirstDay + 7]);
+                      = area.hydro.waterValues[layerindex][weekFirstDay + 6]; // last day of week
```
JasonMarechal25: Please used named variable instead of magic number "6".
If Variable name isn't enough add a comment why "6" or "weekFirstDay+6"

JasonMarechal25 (changes requested):

JasonMarechal25 (approved):

PR #3133, opened 2025-10-09
[ANT-3760] NPCAP test cucumber

JasonMarechal25:


src/tests/cucumber/features/steps/common_steps/solver_steps.py
```diff
@@ -371,3 +373,17 @@ def ckeck_log_exists(context, log):
...
             return
     raise AssertionError(f"Log '{log}' is not reported in the logs")
+
+
+@then(
+    'in area "{area}", during year {year:d}, hourly value of "{var_name}" for hour {hour:d} is equal to {expected_value:d}')
+def check_hourly_variable_value(context, area, year, var_name, hour, expected_value):
+    actual_value = context.soh.get_hourly_value(area, year, var_name, hour)
+    assert expected_value == actual_value, \
+        f"Hourly value mismatch for {var_name}: expected {expected_value}, got {actual_value} (area={area}, year={year}, hour={hour})"
```
flomnes, line 383: A generic step is a good idea to avoid duplication (1 `@then` for each variable), but not used in this PR. Why ?

JasonMarechal25: Matter of expectation philosophy.
I originally went for the generic solution, but test expectation use natural language for variables

flomnes (approved):

PR #3139, opened 2025-10-09
Fix memory leak in class SpatialAggregate

flomnes:


src/solver/variable/include/antares/solver/variable/commons/spatial-aggregate.h
```diff
@@ -159,6 +159,14 @@ class SpatialAggregate
     : public Variable::IVariable<SpatialAggregate<VarT, NextT>, NextT, VCardProxy<VarT>>
 {
 public:
+    ~SpatialAggregate()
+    {
+        if (pValuesForTheCurrentYear)
```
JasonMarechal25: Why not use a uniqptr for pValuesForTheCurrentYear ?

flomnes: At first I tought it was impossible, but you opened my eyes. Thank you.

JasonMarechal25 (approved):

PR #3156, opened 2025-10-14
Extract random number handling from templated class ISimulation (technical preriquisite) [ANT-2334]

flomnes:
Extracting random number handling is necessary for the new `getProblem(year, week)` API since we need these functions that are private members of the `ISimulation<>` class.

### DIFF
--- a/src/solver/simulation/include/antares/solver/simulation/random.h
+++ b/src/solver/simulation/include/antares/solver/simulation/random.h
@@ -1,3 +1,24 @@
+/*
+ * Copyright 2007-2025, RTE (https://www.rte-france.com)
+ * See AUTHORS.txt
+ * SPDX-License-Identifier: MPL-2.0
+ * This file is part of Antares-Simulator,
+ * Adequacy and Performance assessment for interconnected energy networks.
+ *
+ * Antares_Simulator is free software: you can redistribute it and/or modify
+ * it under the terms of the Mozilla Public Licence 2.0 as published by
+ * the Mozilla Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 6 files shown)


src/solver/simulation/include/antares/solver/simulation/random.h
```diff
@@ -0,0 +1,38 @@
...
+ * Mozilla Public Licence 2.0 for more details.
+ *
+ * You should have received a copy of the Mozilla Public Licence 2.0
+ * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
+ */
+
+#pragma once
+
+#include <antares/mersenne-twister/mersenne-twister.h>
+#include <antares/solver/simulation/solver_utils.h>
```
guilpier-code, line 25: About : 
```c++
#include <antares/solver/simulation/solver_utils.h>
```

`class randomNumbers` is currently in **solver_utils.h**.
Functions creating random numbers are here, in **random.h**.

They should be located at the same place, it would make sense. 
It can even be confusing if they are not.

Furthermore, having a class (**randomNumbers**) on one hand and a creation / configuration  outside this class is strange. Everything should be gathered in the class.

So we should move class  **randomNumbers** to **random.h** at least, and maybe put everything in the class.

flomnes: [done](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3156/commits/76be8dc7be6fc3c6aa8fc55507331502e9534f67)

flomnes: also moved free functions to this class

guilpier-code (approved):

PR #3157, opened 2025-10-15
Hack once_flag

JasonMarechal25:
API clients can run multiple simulation in a single runtime. Static once_flag will be set at the first simulation and impair subsequent simulation.


flomnes:
Duplicate #3210

PR #3165, opened 2025-10-16
Getter for single problem [ANT-2334]

flomnes:
Add class `SingleProblemGetter` to allow for single week problem loading
```cpp
Antares::Solver::SingleProblemGetter getter;
getter.load("/path/to/study");
auto constant = getter.getConstantData();
auto weekly = getter.getWeeklyData({/*year*/0, /*week*/1});
```

# TODO
- [X] Load study from a filepath
- [X] Return proper constant data
- [X] Return proper weekly data from user-provided week & year indices
- [x] Check noises
- [x] Check hydro credits, that is RHS for weekly constraint `HydroPower`
- [x] Check timeseries generation (depends on #3182) 
- [x] Cache yearly data (no need to recompute yearly data)
- [X] Check call to `Study::computePThetaInfForThermalClusters`
- [ ] Check simulations not starting on Jan 1st
- [x] Add unit testing
- [x] Clean up code (duplication, TODOs, etc.)


src/api/include/antares/api/singleProblemGetter.h
```diff
@@ -0,0 +1,23 @@
...
+public:
+    void load(const std::filesystem::path& study_path);
+
+    ConstantDataFromAntares getConstantData();
+    WeeklyDataFromAntares getWeeklyData(WeeklyProblemId id);
+
+private:
+    HebdoProblemToLpsTranslator translator_;
+    std::unique_ptr<Antares::Data::Study> study_;
+    PROBLEME_HEBDO pb_;
```
JasonMarechal25: Not a reference ?

flomnes: What reference ?

JasonMarechal25: On PROBLEME_HEBDO.
We usually use reference on this struct. Not that I'm for it or against it. Just asking if having a local copy is a deliberate choice

src/api/include/antares/api/singleProblemGetter.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "antares/solver/hydro/management/management.h"
+#include "antares/solver/lps/LpsFromAntares.h"
+#include "antares/solver/optimisation/HebdoProblemToLpsTranslator.h"
+#include "antares/solver/simulation/random.h"
+#include "antares/solver/simulation/sim_structure_probleme_economique.h"
+#include "antares/study/study.h"
+
+// TODO split header
+namespace Details
```
JasonMarechal25: You wrote a todo but adding my 2 cents. Currently "Details" will be publicly visible by clients wheareas it seems you meant for it to be private implementation details

src/api/singleProblemGetter.cpp
```diff
@@ -0,0 +1,268 @@
...
+    Antares::HydroManagement hydroManagement(study_->areas,
+                                             study_->parameters,
+                                             study_->calendar,
+                                             resultWriter);
+
+    int nbYears = 0;
+    std::map<unsigned int, bool> isYearPerformed; // TODO check year number
+    for (uint year = 0; year < study_->parameters.nbYears; ++year)
+    {
+        isYearPerformed[year] = study_->parameters.yearsFilter[year];
```
JasonMarechal25: isYearPerformed feels redundant with study_->parameters.yearsFilter

src/api/singleProblemGetter.cpp
```diff
@@ -0,0 +1,268 @@
...
+            const auto& availableProduction = cluster->series.getColumn(year);
+            for (uint h = 0; h != cluster->series.timeSeries.height; ++h)
+            {
+                mrs[h] += availableProduction[h];
+            }
+        }
+    }
+}
+
+const Details::YearlyData& SingleProblemGetter::getYearlyData(unsigned year)
```
JasonMarechal25: Probably need to split this method

flomnes:
@JasonMarechal25 

> On PROBLEME_HEBDO.
We usually use reference on this struct. Not that I'm for it or against it. Just asking if having a local copy is a deliberate choice

Yes the getter owns an instance of `PROBLEME_HEBDO`. This is intended

src/api/include/antares/api/singleProblemGetter.h
flomnes: We will have to use multiple instances when running in parallel. I think for the current sprint, the single-threaded version is a reasonable goal.

src/api/singleProblemGetter.cpp
flomnes: done

src/api/include/antares/api/singleProblemGetter.h
flomnes: Yes, this will probably be reworked in the multi-threaded increment

JasonMarechal25 (approved):

PR #3175, opened 2025-10-20
Use boolean type instead of string type for "Run all tests"

JasonMarechal25:


flomnes:
image

That's better indeed

flomnes (approved):

PR #3178, opened 2025-10-20
On Windows : avoiding a compilation collision between antlr4 and Yuni

guilpier-code:
Everything is in the title.

### DIFF
--- a/src/tests/src/utils/unit_test_utils.h
+++ b/src/tests/src/utils/unit_test_utils.h
@@ -24,7 +24,7 @@
 #include <set>
 #include <string>
 
-#include <antares/logs/logs.h>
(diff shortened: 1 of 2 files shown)


src/tests/src/utils/unit_test_utils.h
```diff
@@ -24,7 +24,7 @@
 #include <set>
 #include <string>
 
-#include <antares/logs/logs.h>
+#include "yuni/core/event/interfaces.h"
```
guilpier-code, line 27: The **Yuni** header is narrowed from  **logs.h** to **interfaces.h**.
Apart from being a better practice,  it prevents another compilation error on **Windows**.

flomnes (approved):

PR #3187, opened 2025-10-24
6.2 E2E tests

JasonMarechal25:
End to End test for 6.2 : feature 13 : Export investment problem
Given the modeler system is described with an optim-config
Modeler should generate master, subproblems and structure file

We test that problems data are the same (number of rows, columns, ub, lb, type, etc) using xpress solver. Limit: We can't work with problems too big.
We test that structure files contains the same data 

Todos, pending 6.2c
- [ ] Fix master name (should be "master"
- [ ] Handle subproblem names. Either use names and append ".mps" when reading them or use file name directly.
- [ ] Use proper output directory. Use context.output_path, verify if correct


src/tests/resources/modeler/13_1/input/optim-config.yml
```diff
@@ -0,0 +1,23 @@
...
+  - id: lib_example_13_1.generator_with_continuous_invest
+    model-decomposition:
+      variables:
+        - id: pmax
+          location: master-and-subproblems
+        # Variable generation has default settings, no need to make it appear here
+      objective-contributions:
+        - id: invest_objective
+          location: master
+        - id: operational_objective
```
flomnes, line 22: The input yaml files contain typos, please update your branch when the rest is merged

src/tests/cucumber/features/steps/common_steps/modeler_mps_steps.py
```diff
@@ -0,0 +1,168 @@
+from behave import *
+import xpress as xp
```
flomnes, line 2: xpress requires a license and is nonfree, maybe we could find another python package for MPS parsing ?

flomnes: See https://pypi.org/project/pysmps/

PR #3191, opened 2025-10-27
testing cmake unity build

a-zakir:


src/CMakeLists.txt
```diff
@@ -276,6 +276,8 @@ if (BUILD_UI)
     add_compile_definitions(BUILD_UI)
     add_subdirectory(ui)  #all antares ui libs + antares simulator
 endif ()
+set(CMAKE_UNITY_BUILD ON)
```
flomnes: You should not impose unity builds for everyone. You may however enable it in a matrix for Ubuntu & Windows for GitHub Actions

```
cmake -S src -B build -CMAKE_UNITY_BUILD=${{matrix.unity-build}}
```


If devs want unity builds they can specify it
```
cmake -S src -B build -CMAKE_UNITY_BUILD=ON
```

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
```diff
@@ -18,32 +18,65 @@
  * You should have received a copy of the Mozilla Public Licence 2.0
  * along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
  */
-#include "antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.h"
-
-#include <boost/algorithm/string.hpp>
-
-#include "antares/study/parts/short-term-storage/HoursCollectorVisitor.h"
+//
+#pragma push_macro("ERROR")
```
flomnes, line 23: Care to explain ?

src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp
```diff
@@ -25,6 +25,8 @@
 #include <memory>
 #include <stdexcept>
 
+#include "antares/solver/simulation/sim_structure_probleme_economique.h"
```
flomnes: Why is this header necessary here ?

src/tests/src/solver/simulation/CMakeLists.txt
```diff
@@ -7,19 +7,14 @@ set(src_solver_simulation "${CMAKE_SOURCE_DIR}/solver/simulation")
 set(src_solver_hydro "${CMAKE_SOURCE_DIR}/solver/hydro")
 set(src_libs_antares_study "${CMAKE_SOURCE_DIR}/libs/antares/study")
 
-set(SRC_TS_NUMBERS
-        # For confort in IDE, but not necessary
```
flomnes, line 11: `SRC_TS_NUMBERS` removed but still used below

flomnes (commented):
Lots of effort was put into this ! Most often the code becomes more readable

You have to explain the "ERROR" macro though.

src/CMakeLists.txt
a-zakir: of course that's the plan

src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp
a-zakir: This change is for Windows. The definition of ERROR as 0 (`#define ERROR               0`)  is saved by #pragma push_macro and restored by #pragma pop_macro. 
Done to prevent compilation clash with an antlr enum:
  ```
enum class ParseTreeType : size_t {
    TERMINAL = 1,
    ERROR = 2,
    RULE = 3,
  };
```

File: src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp (hunk 3 above)
Comment (a-zakir) [reply src/solver/optimisation/constraints/ShortTermStorageCumulation.cpp]: removed

File: src/tests/src/solver/simulation/CMakeLists.txt (hunk 4 above)
Comment (a-zakir) [reply src/tests/src/solver/simulation/CMakeLists.txt:11]: let's remove it

File: src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp (hunk 2 above)
Comment (flomnes) [reply src/libs/antares/study/parts/short-term-storage/makeGroupsOfHoursFromString.cpp:23]: Then maybe only do this if the build platform is Windows ?

File: src/CMakeLists.txt [hunk 5]
```diff
@@ -276,6 +280,10 @@ if (BUILD_UI)
     add_compile_definitions(BUILD_UI)
     add_subdirectory(ui)  #all antares ui libs + antares simulator
 endif ()
+
+#restore the original flag
+set(CMAKE_UNITY_BUILD ${UNITY_BUILD})
+
```
flomnes, line 280: Can you try per-target properties ?
```cmake
set_target_properties(MyTargetA PROPERTIES UNITY_BUILD OFF)
```

https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html

Or you may explicitly disable it for each subdirectory, e.g **src/ui/CMakeLists.txt**
```cmake
set(CMAKE_UNITY_BUILD OFF)
```

It shouldn't propagate to the global scope.

flomnes (commented):
LGTM, 2 minor changes

src/CMakeLists.txt
a-zakir: Performing this for every target is inefficient. I'm not sure how many targets there are, but there are definitely a lot. The goal is for this option, when turned ON, to be applied to the entire code. Currently, it doesn't work for Yuni and UI. I plan to do it in an upcoming PR

a-zakir: For now if the user set ON the option, it is applied for all targets except Yuni and ui

flomnes: I understand. Now if you force CMAKE_UNITY_BUILD=OFF in ui/CmakeLists.txt and src/ext/yuni/src/CMakeLists.txt, the behavior is the same and you don't need to save the flag like here.

```cmake
set(CMAKE_UNITY_BUILD OFF)
```
CMake shouldn't propagate this value to the upper scope.

flomnes: Also, the UI code will be removed in the near future so we don't want to invest time into it.

a-zakir: ah ok i see

a-zakir: [bf3d4d2](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3191/commits/bf3d4d2c9fa6dd5fa727183c34e54c6ddf6a0c82)

flomnes (approved):

PR #3197, opened 2025-10-28
Fix delete cache

JasonMarechal25:
https://github.com/cli/cli/issues/8029

Potentialy need to encode gh parameters


JasonMarechal25:
Seems all is ok now.

PR #3198, opened 2025-10-29
Tests for modeler 5.5

payetvin:


flomnes (commented):
You could complexity the mock to account for time-dependency

flomnes (approved):

PR #3199, opened 2025-10-29
6.2-c : update libraries with optim config

guilpier-code:


src/io/inputs/yml-system/converter.cpp
```diff
@@ -101,7 +101,7 @@ static Component createComponent(const YmlSystem::Component& c,
 {
     const auto [libraryId, modelId] = splitLibraryModelString(c.model);
 
-    const Model& model = getModel(libraries, libraryId, modelId);
+    Model& model = getModel(libraries, libraryId, modelId);
```
flomnes, line 104: Note a non-const ref can be converted to `const`. A `const` ref can't be converted to non-const, except using `const_cast`.

You probably don't need to remove the `const` qualifier here.

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -424,7 +424,7 @@ void ComponentFiller::addObjectivesToMaster(const LinearProblemApi::FillContext&
 
 void ComponentFiller::addObjectives(const LinearProblemApi::FillContext& ctx)
 {
-    const auto* model = component_.getModel();
+    auto* model = component_.getModel();
```
flomnes, line 414: The `const` qualifier can probably be kept

src/io/inputs/yml-optim-config/converter.cpp
```diff
@@ -50,69 +49,95 @@ Modeler::Config::Location convertLocation(const std::string& locationStr)
...
+    return *sysVar;
 }
 
-std::vector<Modeler::Config::Objective> convertObjectives(const Model& ymlModel)
+
+SystemModel::Objective& findSystemObjective(std::string obj_id, SystemModel::Model& sysModel)
 {
-    std::vector<Modeler::Config::Objective> objectives;
-    for (const auto& obj: ymlModel.objectives)
+    auto filter = [&obj_id](const SystemModel::Objective& o) { return o.Id() == obj_id; };
```
flomnes:
```suggestion
    auto filter = [&obj_id](const SystemModel::Objective& objective) { return objective.Id() == obj_id; };
```

src/io/inputs/yml-optim-config/converter.cpp
```diff
@@ -50,69 +49,95 @@ Modeler::Config::Location convertLocation(const std::string& locationStr)
...
-        variables.push_back(convertVariable(var));
+        throw std::runtime_error("No variable found with this name: " + variable_id);
     }
-    return variables;
+    return *sysVar;
 }
 
-std::vector<Modeler::Config::Objective> convertObjectives(const Model& ymlModel)
+
+SystemModel::Objective& findSystemObjective(std::string obj_id, SystemModel::Model& sysModel)
```
flomnes:
```suggestion
SystemModel::Objective& findSystemObjective(const std::string& obj_id, SystemModel::Model& sysModel)
```

src/io/inputs/yml-optim-config/converter.cpp
```diff
@@ -50,69 +49,95 @@ Modeler::Config::Location convertLocation(const std::string& locationStr)
...
+SystemModel::Model& findSystemModel(const YmlOptimConfig::Model& ymlModel,
+                                    const std::vector<SystemModel::Library>& libraries)
+{
+    const auto [libraryId, modelId] = splitLibraryModelString(ymlModel.id);
+    return getModel(libraries, libraryId, modelId);
 }
 
-std::vector<Modeler::Config::Variable> convertVariables(const Model& ymlModel)
+SystemModel::Variable& findSystemVariable(std::string variable_id, SystemModel::Model& sysModel)
 {
```
flomnes:
```suggestion
SystemModel::Variable& findSystemVariable(const std::string& variable_id, SystemModel::Model& sysModel)
{
```

src/io/inputs/yml-optim-config/include/antares/io/inputs/yml-optim-config/converter.h
```diff
@@ -23,14 +23,16 @@
...
 
 namespace Antares::IO::Inputs::YmlOptimConfig
 {
 
 class OptimConfigConverter
 {
 public:
-    static Modeler::Config::OptimConfig convert(const OptimConfig& ymlOptimConfig);
+    static void updateLibrairies(const OptimConfig& ymlOptimConfig,
+                                 std::vector<ModelerStudy::SystemModel::Library>& libraries);
```
flomnes: A class with a single `static` method looks strange, is the `class` really needed ?

src/io/inputs/yml-optim-config/converter.cpp
```diff
@@ -50,69 +49,95 @@ Modeler::Config::Location convertLocation(const std::string& locationStr)
...
-        return {ymlObj.id, convertLocation(ymlObj.location)};
+        throw std::runtime_error("No library found with this name: " + libraryId);
     }
-    catch (const Error::RuntimeError& e)
+
+    auto search = lib->Models().find(modelId);
+    if (search == lib->Models().end())
     {
-        throw Error::RuntimeError("Error converting objective '" + ymlObj.id + "': " + e.what());
+        throw std::runtime_error("No library found with this name: " + modelId);
```
flomnes: Typo
```suggestion
        throw std::runtime_error("No model found with this name: " + modelId);
```

src/io/inputs/yml-optim-config/converter.cpp
```diff
@@ -50,69 +49,95 @@ Modeler::Config::Location convertLocation(const std::string& locationStr)
...
-    std::vector<Modeler::Config::Objective> objectives;
-    for (const auto& obj: ymlModel.objectives)
+    auto filter = [&obj_id](const SystemModel::Objective& o) { return o.Id() == obj_id; };
+    auto& sysObjectives = sysModel.Objectives();
+    auto sysObj = std::ranges::find_if(sysObjectives, filter);
+    if (sysObj == sysObjectives.end())
     {
-        objectives.push_back(convertObjective(obj));
-        std::vector<Modeler::Config::Objective> objectives;
+        throw std::runtime_error("No objetctive found with this name: " + obj_id);
```
flomnes: Typo
```suggestion
        throw std::runtime_error("No objective found with this name: " + obj_id);
```

flomnes (commented):
LGTM

A few typos

We should add unit tests for this

src/io/inputs/yml-system/converter.cpp
guilpier-code: Adding back a **const** causes a compilation error a bit further when calling  `component_builder.withModel(&model)`

guilpier-code: There would be a solution to all of this **const** problematic.
We could first load libraries and option config under the form of a yml data structures, and then build business data structure with these.

src/solver/optim-model-filler/ComponentFiller.cpp
guilpier-code: If we make this model **const**, then we cannot call **Objectives()** on it (turned to non-const function in this PR).
This causes a compilation error.

src/io/inputs/yml-optim-config/converter.cpp
guilpier-code: done

src/io/inputs/yml-optim-config/converter.cpp
guilpier-code: done

src/io/inputs/yml-optim-config/converter.cpp
guilpier-code: Done

src/io/inputs/yml-optim-config/include/antares/io/inputs/yml-optim-config/converter.h
guilpier-code: Indeed. Turned into a free function.

src/io/inputs/yml-optim-config/converter.cpp
guilpier-code: Done

src/io/inputs/yml-optim-config/converter.cpp
guilpier-code: Done

src/io/inputs/yml-system/converter.cpp
a-zakir: Have you examined the case of injecting optimConfig before libraries are created?

a-zakir: mostly to avoid de-consting Model

guilpier-code: Yes, but this way does not allow checking that all variables / objectives in the **optim-config.yml** exist in one of the libraries yml files

a-zakir: after you mapped libraries with optimConfig , you'd need a simple loop on optimConfig and raise exception if there is orphan optimConfig  .vars or objectives

guilpier-code: Indeed, we could fix it that way too.
I suggest we open a ticket on that matter and treat it at some point, as soon as possible.

guilpier-code: Here is [the ticket](https://gopro-tickets.rte-france.com/browse/ANT-4051)

a-zakir (approved):

PR #3204, opened 2025-10-30
Add test coverage for variable & objective decomposition

flomnes:
- Add unit test for variables
- Allow absence of attribute `variables` and `objective-contributions` when parsing optim-config.yml 

```yaml
models:
  - id: library.model
    model-decomposition:
      variables:
        - id: y
          location: master-and-subproblems
```

### DIFF
--- a/src/io/inputs/yml-optim-config/decoders.hxx
+++ b/src/io/inputs/yml-optim-config/decoders.hxx
@@ -30,6 +30,13 @@
 namespace YAML
 {
 
+// TODO this function is defined at least 3 times, deduplicate
+template<typename T>
+inline T as_fallback_default(const Node& n)
+{
+    return n.as<T>(T());
+}
+
 template<>
 struct convert<Antares::IO::Inputs::YmlOptimConfig::Variable>
 {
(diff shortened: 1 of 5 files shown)


src/io/inputs/yml-optim-config/decoders.hxx
```diff
@@ -30,6 +30,13 @@
 namespace YAML
 {
 
+// TODO this function is defined at least 3 times, deduplicate
+template<typename T>
+inline T as_fallback_default(const Node& n)
+{
+    return n.as<T>(T());
+}
```
a-zakir, line 38: this function is highly useless

flomnes: It avoid duplicating `T` I guess

a-zakir: before:
````
rhs.objectives = modelDecompositionNode["objective-contributions"]
                           .as<std::vector<Antares::IO::Inputs::YmlOptimConfig::Objective>>();
````
after:
```
rhs.objectives = as_fallback_default<
          std::vector<Antares::IO::Inputs::YmlOptimConfig::Objective>>(
          modelDecompositionNode["objective-contributions"]);
```

Comment (flomnes) [reply src/io/inputs/yml-optim-config/decoders.hxx:38]: No the behavior changes if the key is not found. Before, it crashes. After it yields an empty vector

Comment (guilpier-code) [reply src/io/inputs/yml-optim-config/decoders.hxx:38]: ok, so it does something more than it used to.
I may add : 
- **as_fallback_default** is a bit cryptic (at least to me). It's basically a conversion. So why not call it **convert** or **convert_to** ?
- Expanding namespace is particularly painful to read here. Couldn't we avoid it and have something like : 
```c++
rhs.objectives = convert_to<std::vector<YmlOptimConfig::Objective>>(modelDecompositionNode["objective-contributions"]);
```
?

src/tests/src/io/yml-importers/testYmlOptimConfig.cpp
```diff
@@ -44,7 +44,7 @@ BOOST_AUTO_TEST_CASE(parse_complete_valid_yaml)
           location: master
         - id: pmax_cluster
           location: master-and-subproblems
-      objectives:
+      objective-contributions:
```
guilpier-code, line 47: All these were also fixed in parent branch.

PR #3206, opened 2025-10-31
Add debug files to hydro remix [ANT-4028]

payetvin:


docs/user-guide/solver/04-parameters.md
```diff
@@ -254,13 +254,19 @@ These parameters are listed under the `[output]` section in the `.ini` file.
...
-- **Expected value:**
-- **Required:** **yes**
-- **Default value:**
-- **Usage:**
+---
+#### remix-storage-debug
+- **Expected value:** `true` or `false`
+- **Required:** no
+- **Default value:** `false`
+    - `true`: Output files in format remix-*year*-*week*.csv containing withdrawal for all storages. Format inside the file is *area* *storage* *hour* *withdrawal*
```
guilpier-code: About format, following sentence would be more clear : 
Each line of the file is associated to an hour and is formatted like this : *area*, *storage*, *hour*, *withdrawal*

src/solver/simulation/economy.cpp
guilpier-code, line 1: _Not really related to this PR, but can be quickly corrected :_ 

**private** data member **resultWriter** should be renamed into **resultWriter_** to make it clear it's **private**.

src/libs/antares/study/include/antares/study/parameters.h
guilpier-code, line 1: _Not really related to this PR, but can be quickly corrected :_
The 2 functions are declared but seem to be never defined.
```c++
void resetThresholdsAdqPatch();
void resetAdqPatchParameters();
```
We should remove them.

src/solver/simulation/economy.cpp
```diff
@@ -113,7 +113,8 @@ bool Economy::simulationBegin()
               numSpace,
               study.areas,
               study.parameters,
-              study.calendar);
+              study.calendar,
+              resultWriter);
```
guilpier-code: **Question about the result writer :** 
We're actually dumping "expert" information about changes in storage productions.
Do we need the **result writer** for this ? : 
- do we need to dump these pieces of data in a .zip ?
- do we need to pass the **result writer** all the way down from study loading to a particular post-treatment ?

We could locally (I mean locally to the remix algo) create it create it and use it

guilpier-code: Here we pass a writer to a **create(...)** method. 
At this level, it's easy to know if this writer is about to write something or not, depending on **study.parameters.remixStorageDebug**.
To be the less invasive as possible in the lower levels (and avoid the **if** statement, and the **nullptr**), I  recommend to pass to this method **create(...)** a **NullResultWriter** if we don't want to write.

So we could have something like : 
```c++
IResultWriter writerForRemixStorage = makeRemixStorageWriter(resultWriter, study);

postProcessesList_[numSpace] = interfacePostProcessList::create(
  ...
  study.calendar,
  writerForRemixStorage);
```
In this code, **makeRemixStorageWriter** returns the classic **resultWriter** if we actually write something, and a null writer otherwise.

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -454,9 +474,15 @@ void RemixHydroForAllAreas(const Data::AreaList& areas,
     else if (params.shedding.policy == Data::shpAccurateShavePeaks)
     {
         bool includeSTS = params.accurateShavePeaksIncludeShortTermStorage;
+        bool debugInfos = params.remixStorageDebug;
```
guilpier-code, line 475: If we follow [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3206#discussion_r2517586415), we don't need : 
`bool debugInfos = params.remixStorageDebug;` 
or
 `debugInfos ? &resultWriter : nullptr`

We just keep on passing the writer.

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -454,9 +474,15 @@ void RemixHydroForAllAreas(const Data::AreaList& areas,
...
         try
         {
-            RunAccurateShavePeaks(areas, problem, numSpace, hourInYear, includeSTS);
+            RunAccurateShavePeaks(areas,
+                                  problem,
+                                  numSpace,
+                                  hourInYear,
+                                  includeSTS,
+                                  debugInfos ? &resultWriter : nullptr);
```
guilpier-code, line 483: `writer ? &debugStream : nullptr` : same as above.

src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -64,6 +70,16 @@ void shavePeaksByRemixingStorageGen(const std::vector<double>& Load,
 
         if (!exchange.isPossible())
         {
+            if (debugStream)
+            {
+                StorageForRemixNoLevels* s = dynamic_cast<StorageForRemixNoLevels*>(
```
guilpier-code: We could avoid this **dynamic_cast**. 
It's the sign that something is wrong in the polymorphism design.
Every storage has a name. So it's expected that abstract parent class **IStorageForRemix** is supplied with a **name()** pure virtual method (attribute **name_** is then inherited in a **protected** way).

src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -64,6 +70,16 @@ void shavePeaksByRemixingStorageGen(const std::vector<double>& Load,
...
+                  (*cyclic_it).get());
+                auto& withdrawal = s->withdrawal();
+                for (unsigned h = 0; h < withdrawal.size(); ++h)
+                {
+                    *debugStream << s->name() << " " << h << " " << withdrawal[h] << std::endl;
+                }
+            }
```
guilpier-code: We should be less invasive to be more readable. We should hide implementation details behind a call). 
Instead of this block : 
```c++
if (debugStream)
{
    StorageForRemixNoLevels* s = dynamic_cast<StorageForRemixNoLevels*>(
      (*cyclic_it).get());
    auto& withdrawal = s->withdrawal();
    for (unsigned h = 0; h < withdrawal.size(); ++h)
    {
        *debugStream << s->name() << " " << h << " " << withdrawal[h] << std::endl;
    }
}
```

we should have : 
```c++
collectInfoForWriting(cyclic_it, remixStorageWriter);
```
Notice that now, we can the writer to the top **shavePeaksByRemixingStorageGen(...)** function, instead of the curious `std::stringstream*`.

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -420,13 +432,21 @@ static void RunAccurateShavePeaks(const Data::AreaList& areas,
               logs.warning(msg);
           }
       });
+    if (writer)
+    {
+        std::string filename("remix-" + std::to_string(problem.year) + "-"
+                             + std::to_string(problem.weekInTheYear) + ".csv");
+        std::string s = debugStream.str();
+        writer->addEntryFromBuffer(filename, s);
+    }
```
guilpier-code, line 439: We could avoid the **if** statement. See [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3206#discussion_r2517586415)

src/solver/simulation/economy.cpp
guilpier-code: _May not be treated in this PR._

In the end, I'm not a huge fan of this writer for everything, especially every file.
This "universal" writer may not be required when writing "debug" info, not intended for users.
One of the reasons we pass this writer all the way down  to remix algo is that it contains the study output path. 
We should pass this path down to where it's needed and use a more simple writer locally (inside or very next to the remix algo).

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -284,7 +287,8 @@ std::shared_ptr<IStorageForRemix> extractHydroForRemix(const Data::Area& area,
                              initLevel,
                              capacity,
                              efficiency,
-                             reservoirManagement);
+                             reservoirManagement,
+                             std::string(area.name) + " hydro");
```
guilpier-code, line 289: Appending "hydro" to storage name could be done inside **makeHydroForRemix**, this would avoid the caller to do it, make this caller a bit more simple.

src/solver/simulation/common-hydro-remix.cpp
```diff
@@ -357,7 +361,8 @@ std::shared_ptr<IStorageForRemix> extractSTSforRemix(const Data::Area& area,
                            upRuleCurve,
                            initLevel,
                            withdrawalEff,
-                           injectionEff);
+                           injectionEff,
+                           std::string(area.name) + " " + stsProperties.name);
```
guilpier-code, line 363: Same as above.

src/solver/simulation/include/antares/solver/simulation/remix-storage/storage-for-remix-no-levels.h
```diff
@@ -6,23 +6,27 @@
...
 
     double maxExchange(unsigned hourOfMaxGen, unsigned hourOfMinGen) const override;
     void update() override;
     const std::vector<double>& initWithdrawal() const override;
     std::vector<double>& withdrawal() override;
 
+    const std::string& name() const
+    {
+        return name_;
+    }
```
guilpier-code: To be moved to parent abstract class

src/solver/simulation/remix-storage/CMakeLists.txt
```diff
@@ -25,7 +25,10 @@ target_include_directories(${local_target}
 
 set_target_properties(${local_target} PROPERTIES FOLDER storage-remix)
 
-target_link_libraries(${local_target} PRIVATE storage-for-remix-list-sort)
+target_link_libraries(${local_target}
+    PRIVATE
+    storage-for-remix-list-sort
+    Antares::logs)
```
guilpier-code: `Antares::logs` : do we need this ?
In all files of target **shave-peaks-by-remix-storage-gen**, I could not find any usage of the logger. 
But maybe there is.

src/solver/simulation/economy.cpp
payetvin: If we pass a nullResultWriter we won't write (that's good) but we'll generate the debug infos for nothing
With the current implementation we don't generate anything if the debug isn't wanted

src/solver/simulation/common-hydro-remix.cpp
payetvin: We don't have access to sts properties in makeSTSforRemix

src/solver/simulation/include/antares/solver/simulation/remix-storage/storage-for-remix-no-levels.h
```diff
@@ -6,23 +6,27 @@
...
-                            const std::vector<double> Pmin);
+                            const std::vector<double> Pmin,
+                            const std::string& name = "");
 
     double maxExchange(unsigned hourOfMaxGen, unsigned hourOfMinGen) const override;
     void update() override;
     const std::vector<double>& initWithdrawal() const override;
     std::vector<double>& withdrawal() override;
 
+    const std::string& name() const override
```
guilpier-code: Any storage for remix has name, we know that.
Indeed, interface class **IStorageForRemix** has a method **name()**.
So why not have the interface class hold the data member **name_** and have the `const std::string& name()` method ?
A (small) benefit will be that when calling name(), we won't have the (small) overhead due to polymorphism.

src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -64,6 +78,7 @@ void shavePeaksByRemixingStorageGen(const std::vector<double>& Load,
 
         if (!exchange.isPossible())
         {
+            collectDebugInfos((*cyclic_it).get(), debugStream);
```
guilpier-code: So, according to the current code, what we want is : collect debug information about the current storage (in the loop) **only if** the exchange of production between 2 hours is **not** possible ?
It seems strange to me.

src/solver/simulation/economy.cpp
flomnes: I agree with @guilpier-code, it's a refactor for later since the code works well as is and is understandable.

flomnes (approved):

PR #3207, opened 2025-10-31
[ANT-4033] Feat: Support setObjectiveOffset

JasonMarechal25:
### Overview

The `objectiveOffset` represents the constant term in a linear programming objective function. For an objective function of the form `Σ(aᵢ·xᵢ) + b`, where `xᵢ` are decision variables with coefficients `aᵢ`, the offset corresponds to the constant term `b`.

This constant term aggregates all non-variable contributions to the objective value and is computed during the model building phase in `ComponentFiller::addObjectives()`.

#### 1. Variables Alone
**No offset contribution**: Variables and their coefficients do not contribute to the offset, regardless of their time/scenario dependency.

- `x` → offset = 0
- `param × x` → offset = 0
- `x(t)` over n time steps → offset = 0

#### 2. Literal Constants
**Offset = constant value**

- `x + 10` → offset = 10

#### 3. Parameters as Constant Terms
**Offset = parameter value** evaluated according to its time/scenario dependency. Parameters represent known values (not optimization variables) and thus contribute to the constant term.

##### Variable with constant offset
**Offset = n × constant_value** (where `n` is the number of time steps)

- Variable over 10 time steps with constant offset 10 → total offset = 100

##### Time-varying parameter
**Offset = Σₜ param(t)** (sum over all time steps in the optimization horizon)

- `param = [10, 11, 12]` over 3 time steps → offset = 33

#### 5. Scenario Dependency
**Offset = value for the selected scenario**

Parameters varying only by scenario contribute their value from the selected scenario's time series.

- Parameter with 3 scenarios `[15, 25, 35]`, scenario 2 selected → offset = 35

#### 6. Time AND Scenario Dependency
**Offset = Σₜ param(t, s)** (sum over all time steps for the selected scenario)

- Parameter in selected scenario with values `[15, 25, 35]` over 3 time steps → offset = 75

#### 7. Multiple Scenario Groups
When different model elements use different scenario groups:

- Each element selects its time series from **its own scenario group** independently
- Offsets from different elements **accumulate**
- Only **parameters** (constant terms) contribute to the offset
- **Variables** never contribute to the offset, even when using different scenario groups

**Example**: 
- Variable `x` using `scenarioX` with values `[11, 22, 33]`
- Parameter using `scenarioY` with values `[3, 6, 9]` over 3 time steps
- Variable contribution: 0 (variables don't affect offset)
- Parameter contribution: 3 + 6 + 9 = 18
- **Total offset: 18**


src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
```diff
@@ -738,4 +732,100 @@ BOOST_AUTO_TEST_CASE(one_var_with_param_objective)
...
+    createModelWithOneFloatVar("model", {"param"}, "x", literal(-50), literal(-40), {}, objective);
+    createComponent("model", "componentA", {build_context_parameter_with("param", "5")});
+    buildLinearProblem();
+    BOOST_CHECK_EQUAL(pb->getObjectiveOffset(), 0);
+}
+
+// One var with offset
+BOOST_AUTO_TEST_CASE(one_var_with_offset)
+{
+    const auto objective = add(variable("x", 0), literal(10));
```
guilpier-code: Adding a comment on objectives could make these unit tests even more readable, for example : 
```c++
// objective = x + 10.
const auto objective = add(variable("x", 0), literal(10));
```

src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
```diff
@@ -738,4 +732,100 @@ BOOST_AUTO_TEST_CASE(one_var_with_param_objective)
     BOOST_CHECK_EQUAL(pb->getObjectiveCoefficient(pb->lookupVariable("componentA.x")), -25);
 }
 
+BOOST_AUTO_TEST_CASE(one_var_no_offset)
```
guilpier-code: This file contains already many tests and many lines of code.
As its name says, it's supposed to test component fillers. 
It seems we're doing more than that in this file.
Maybe we should start thinking about splitting this file.

src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
```diff
@@ -738,4 +732,100 @@ BOOST_AUTO_TEST_CASE(one_var_with_param_objective)
...
+      "model",
+      {Parameter{"param", TimeDependent::YES, ScenarioDependent::NO}},
+      {{"x", ValueType::FLOAT, literal(-5), literal(10), true, false}},
+      {},
+      objective);
+    createComponent("model",
+                    "componentA",
+                    {build_context_parameter_with("param", "bounds", ParameterType::TIMESERIE)});
+
+    const vector<unsigned int> timeSteps{0, 1, 2};
```
guilpier-code:
```c++
const vector<unsigned int> timeSteps{0, 1, 2};
```
Do we need this vector ?
We could have (much shorter) : 
```c++
FillContext ctx{0, 2, 0, 2, 0};
```
But (as you may say), less expressive.
A problem comes from the class **FillContext**, which has a terrible name (difficult to be more meaningless).

Furthermore, as the global time interval is the same as the local time interval, I may recommend to have another constructor for class **FillContext**, so that we could have 

```c++
FillContext ctx{0, 2, 0};
```

src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
```diff
@@ -738,4 +732,100 @@ BOOST_AUTO_TEST_CASE(one_var_with_param_objective)
...
+    try
+    {
+        buildLinearProblem(ctx, data, scenarios);
+    }
+    catch (const Antares::Error::RuntimeError& e)
+    {
+        BOOST_CHECK(std::string(e.what()).find(
+                      "Trying to set multiple objective offset for the same objective.")
+                    != std::string::npos);
+    }
```
guilpier-code: Instead of this **try**... **catch**, we already have : 
```c++
BOOST_CHECK_EXCEPTION(buildLinearProblem(ctx, data, scenarios),
                      std::runtime_error,
                      containsMessage(err_msg));
```

src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
```diff
@@ -738,4 +732,100 @@ BOOST_AUTO_TEST_CASE(one_var_with_param_objective)
...
+                    "componentA",
+                    {build_context_parameter_with("param", "bounds", ParameterType::TIMESERIE)});
+
+    const vector<unsigned int> timeSteps{0, 1, 2};
+    FillContext ctx{timeSteps.at(0), timeSteps.at(2), timeSteps.at(0), timeSteps.at(2), 0};
+    auto bounds_time_series = std::make_unique<TimeSeriesSet>("bounds", 3);
+    // setting 3 hours (including h 1 and 2)
+    bounds_time_series->add({10., 11., 12.});
+    LinearProblemData data;
+    data.addDataSeries(std::move(bounds_time_series));
```
guilpier-code:
```c++
data.addDataSeries(std::move(bounds_time_series));
```
you don't have to move the `std::unique_ptr<TimeSeriesSet>` here, it will be automatically (as you know).

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -404,6 +407,60 @@ void ComponentFiller::addConstraints(const LinearProblemApi::FillContext& ctx)
...
 }
 
+namespace
+{
+std::optional<double> updateOffset(std::optional<double> objectiveOffset,
+                                   double offset,
+                                   std::string objectiveId,
+                                   LinearProblemApi::ILinearProblem& pb)
+{
+    constexpr double eps = 1e-6; // tolérance numérique pour les comparaisons
```
guilpier-code: Comments in English. 
Besides, we may not need this comment.

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -421,21 +478,23 @@ void ComponentFiller::addObjectives(const LinearProblemApi::FillContext& ctx)
           objective.expression().RootNode());
 
         auto& pb = optimEntityContainer_.Problem();
+        std::optional<double> objectiveOffset;
         for (const auto& expr: linearExpression)
         {
             for (const auto& [index, value]: expr)
             {
                 pb.setObjectiveCoefficient(solverVariables[static_cast<std::size_t>(index)].get(),
```
guilpier-code:
```c++
pb.setObjectiveCoefficient(solverVariables[static_cast<std::size_t>(index)].get()
```
Why this ugly `static_cast<std::size_t>(index)` ?
If I change this previous instruction into :
```c++
pb.setObjectiveCoefficient(solverVariables[index].get()
``` 
nothing special happens (it still compiles, without warning on **Windows**).

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -421,21 +478,23 @@ void ComponentFiller::addObjectives(const LinearProblemApi::FillContext& ctx)
           objective.expression().RootNode());
 
         auto& pb = optimEntityContainer_.Problem();
+        std::optional<double> objectiveOffset;
         for (const auto& expr: linearExpression)
         {
             for (const auto& [index, value]: expr)
```
guilpier-code:
```c++
for (const auto& [index, value]: expr)
```
Renaming : 
instead of **value**, **coef** is more consistant and precise with what we're doing here.
**value** is even misleading.

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -404,6 +407,60 @@ void ComponentFiller::addConstraints(const LinearProblemApi::FillContext& ctx)
...
+namespace
+{
+std::optional<double> updateOffset(std::optional<double> objectiveOffset,
+                                   double offset,
+                                   std::string objectiveId,
+                                   LinearProblemApi::ILinearProblem& pb)
+{
+    constexpr double eps = 1e-6; // tolérance numérique pour les comparaisons
+
+    if (!std::isfinite(offset))
```
guilpier-code:
```c++
if (!std::isfinite(offset))
```
Isn't there a contradiction between the proposition`!std::isfinite(offset)` and the message of exception raised ?
You probably mean : 
```c++
if (std::isfinite(offset))
```

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -404,6 +407,60 @@ void ComponentFiller::addConstraints(const LinearProblemApi::FillContext& ctx)
...
+                      offset));
+    }
+
+    if (std::abs(offset) < eps)
+    {
+        logs.debug() << fmt::format("Ignoring near-zero objective offset for objective: "
+                                    "{}.\n\tCurrent value: {}\n\tIgnored value: {}",
+                                    objectiveId,
+                                    objectiveOffset.value_or(0),
+                                    offset);
```
guilpier-code: Cases where offset is zero in an objective function will be very numerous in practice.
So we may have many many logs, for a default case.
Is this what we want ?

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -421,21 +478,23 @@ void ComponentFiller::addObjectives(const LinearProblemApi::FillContext& ctx)
...
         auto& pb = optimEntityContainer_.Problem();
+        std::optional<double> objectiveOffset;
         for (const auto& expr: linearExpression)
         {
             for (const auto& [index, value]: expr)
             {
                 pb.setObjectiveCoefficient(solverVariables[static_cast<std::size_t>(index)].get(),
                                            value);
             }
+            objectiveOffset = updateOffset(objectiveOffset, expr.constant(), objective.Id(), pb);
```
guilpier-code: I wondered about `updateOffset(...).`.
Roughly (and from what I understand), if the objective is time-dependent, the code adds the first of the **offset**'s component and raises an exception if there's another component that is different. I mean, if offset is **B = {b1, ...bn}** (**b1 = B[t1]**, **b2 = B[t2]**, ...), it adds **b1**,  **b2**, ... as long as 2 offsets components are different, otherwise fails.

@tbittar says that all of the **B**'s components have to be added to the objective function.
So I guess we should have something like : 
```c++
auto& pb = optimEntityContainer_.Problem();

for (const auto& objective: model->Objectives() | locationFilter)
{
    const auto linearExpression = visitor.visitMergeDuplicates(
      objective.expression().RootNode());

    for (const auto& expr: linearExpression)
    {
        for (const auto& [index, coef]: expr)
        {
            pb.setObjectiveCoefficient(solverVariables[index].get(), coef);
        }
        
        pb.setObjectiveOffset(expr.constant());
    }
}
```

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -421,21 +478,23 @@ void ComponentFiller::addObjectives(const LinearProblemApi::FillContext& ctx)
           objective.expression().RootNode());
 
         auto& pb = optimEntityContainer_.Problem();
```
guilpier-code:
```c++
auto& pb = optimEntityContainer_.Problem();
```
A reference to the linear problem is instantiated inside the loop on objectives. 
So we're doing the same instantiation  several times. 
It should be done before this loop.

src/tests/src/solver/optim-model-filler/test_componentFiller.cpp
JasonMarechal25: bounds_time_series is a unique_ptr of TimeSeriesSet
addDataSeries expect unique_ptr of IDataSeries
unique_ptr does not support implicite conversion, hence call to std::move

`fatal error: no viable conversion from 'unique_ptr<Antares::Optimisation::LinearProblemDataImpl::TimeSeriesSet>' to 'unique_ptr<IDataSeries>'
  204 |     data.addDataSeries(bounds_time_series);
      |                        ^~~~~~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/unique_ptr.h:320:12: note: candidate constructor template not viable: no known conversion from 'typename _MakeUniq<TimeSeriesSet>::__single_object' (aka 'unique_ptr<Antares::Optimisation::LinearProblemDataImpl::TimeSeriesSet>') to 'nullptr_t' (aka 'std::nullptr_t') for 1st argument
  320 |         constexpr unique_ptr(nullptr_t) noexcept`

src/optimisation/linear-problem-mpsolver-impl/linearProblem.cpp
```diff
@@ -231,6 +231,20 @@ OrtoolsMipSolution* OrtoolsLinearProblem::solve(bool verboseSolver)
...
+    {
+        solve(verboseSolver);
+    }
+    return solution_.get();
+}
+
+double OrtoolsLinearProblem::objectiveValue() const
+{
+    return objective_->Value();
+}
```
flomnes, line 246: The end-user doesn't want the objective offset, but rather the objective final value

I think we should have at least one unit test for this function.

src/optimisation/linear-problem-mpsolver-impl/include/antares/optimisation/linear-problem-mpsolver-impl/linearProblem.h
```diff
@@ -21,6 +21,8 @@
 
 #pragma once
 
+#include <functional>
+
```
flomnes:
```suggestion
```

src/optimisation/linear-problem-mpsolver-impl/linearProblem.cpp
JasonMarechal25: Done f098b6040565c2cc6fd3dc67f279e1661f01a9f9

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -424,7 +426,10 @@ void ComponentFiller::addObjectives(const LinearProblemApi::FillContext& ctx)
             throw Error::RuntimeError("Time dependent objectives are not supported in Antares.");
         }
         addStaticObjective(linearExpression);
+        objectiveOffset += linearExpression.constant()[0];
     }
```
guilpier-code, line 428: Only time independent expressions are accepted in objectives.
To make it even more clear in the code,  we could have, inside the loop over objectives : 
```c++
const auto root_node = objective.expression().RootNode();
const auto variability = getVariability(root_node, component_);
if (isTimeDependent(variability))
{
    throw Error::RuntimeError("Time dependent objectives are not supported in Antares.");
}

const auto linearExpression = visitor.visitMergeDuplicates(root_node)[0];
addStaticObjective(linearExpression);
objectiveOffset += linearExpression.constant();
``` 
Note that, in code above : 
- **linearExpression** is now of type **LinearExpression** (and no longer of type **TimeDependentLinearExpression**), so it has 1 dimension (not multiple dimensions).
- addStaticObjective(...) takes a **LinearExpression**

JasonMarechal25: indeed better : [`d6e0b98` (#3207)](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3207/commits/d6e0b9877f33ce31718b99798e4605f5271b4222)

guilpier-code (approved):

PR #3208, opened 2025-11-03
Using unique node for Functions

a-zakir:
it replaces reduced_cost and dual nodes and it will be used for other functions (max, min etc...)


a-zakir:
followed by https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3212

PR #3210, opened 2025-11-04
Add ExportBehavior for structure files

flomnes:
```cpp
enum class ExportBehavior
{
    // Never export structure problems
    Never,
    // Export once, on the first call
    Once,
    // Always export the structure problems
    Always
};
```

### DIFF
--- a/src/libs/antares/optimization-options/include/antares/optimization-options/options.h
+++ b/src/libs/antares/optimization-options/include/antares/optimization-options/options.h
@@ -52,8 +52,11 @@ struct CmdLineOptimOptions
 
 enum class ExportBehavior
 {
+    // Never export structure problems
     Never,
+    // Export once, on the first call
     Once,
+    // Always export the structure problems
     Always
 };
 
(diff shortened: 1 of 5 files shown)


src/libs/antares/optimization-options/include/antares/optimization-options/options.h
```diff
@@ -59,5 +69,6 @@ class OptimizationOptions final
     SingleOptimOptions secondOptimOptions;
     SingleOptimOptions quadraticOptimOptions;
     bool solverLogs = false;
+    ExportBehavior exportBehavior = ExportBehavior::Once;
```
JasonMarechal25: Should be "Never" to match old behaviour:
bool ExportStructure = false;

JasonMarechal25 (changes requested):

src/libs/antares/optimization-options/include/antares/optimization-options/options.h
flomnes: 394a0edda done

JasonMarechal25 (approved):

PR #3211, opened 2025-11-04
Final steps for 6.2c

flomnes:
- Fix glitches with `OptimEntityContainer`
- One `OptimEntityContainer` for each problem
- Write master.mps and 1-1.mps
- Write structure in `std::cout`

TODO
- [ ] Write structure into structure.txt

### DIFF
--- a/src/modeler-optimisation-container/OptimEntityContainer.cpp
+++ b/src/modeler-optimisation-container/OptimEntityContainer.cpp
@@ -37,7 +37,8 @@ OptimEntityContainer::OptimEntityContainer(LinearProblemApi::ILinearProblem& lin
 {
 }
 
-void OptimEntityContainer::addFromSystemComponents(const std::vector<Component>& components)
+void OptimEntityContainer::addFromSystemComponents(const std::vector<Component>& components,
+                                                   Modeler::Config::Location targetLocation)
 {
     optimComponents_.clear();
     optimComponents_.reserve(components.size());
@@ -47,13 +48,14 @@ void OptimEntityContainer::addFromSystemComponents(const std::vector<Component>&
         auto* model = component.getModel();
         const auto& variables = model->Variables();
         std::vector<unsigned int> modelVariableGlobalIndices;
-        modelVariableGlobalIndices.resize(variables.size());
 
-        for (auto variableLocalIndex = 0; variableLocalIndex < variables.size();
-             ++variableLocalIndex)
+        for (const auto& variable: variables)
         {
-            modelVariableGlobalIndices[variableLocalIndex] = variableGlobalIndex;
-            ++variableGlobalIndex;
+            if (AreLocationsCompatible(variable.location(), targetLocation))
+            {
+                modelVariableGlobalIndices.push_back(variableGlobalIndex);
+                ++variableGlobalIndex;
+            }
         }
         optimComponents_.push_back(
           {.index = component.Index(),
--- a/src/modeler-optimisation-container/OptimEntityContainer.cpp
+++ b/src/modeler-optimisation-container/OptimEntityContainer.cpp
@@ -56,6 +56,10 @@ void OptimEntityContainer::addFromSystemComponents(const std::vector<Component>&
                 modelVariableGlobalIndices.push_back(variableGlobalIndex);
                 ++variableGlobalIndex;
             }
+            else
+            {
+                modelVariableGlobalIndices.push_back(-1);
+            }
         }
         optimComponents_.push_back(
           {.index = component.Index(),
--- a/src/modeler-optimisation-container/OptimEntityContainer.cpp
+++ b/src/modeler-optimisation-container/OptimEntityContainer.cpp
@@ -49,6 +49,7 @@ void OptimEntityContainer::addFromSystemComponents(const std::vector<Component>&
         const auto& variables = model->Variables();
         std::vector<unsigned int> modelVariableGlobalIndices;
 
+        modelVariableGlobalIndices.reserve(variables.size());
         for (const auto& variable: variables)
         {
             if (AreLocationsCompatible(variable.location(), targetLocation))
@@ -58,6 +59,8 @@ void OptimEntityContainer::addFromSystemComponents(const std::vector<Component>&
             }
             else
             {
+                // We add dummy data to avoid "holes" in modelVariableGlobalIndices
+                // but these are not to be used
                 modelVariableGlobalIndices.push_back(-1);
             }
         }
--- a/src/modeler/FileWriter.cpp
+++ b/src/modeler/FileWriter.cpp
@@ -85,7 +85,7 @@ void FileWriter::writeProblem(
 {
     if (output)
     {
-        logs.info() << "Writing problem.lp...";
+        logs.info() << "Writing " << filename << "...";
         const auto lp_path = outputPath_ / filename;
         problem.Write(lp_path.string());
     }
--- a/src/modeler/FileWriter.cpp
+++ b/src/modeler/FileWriter.cpp
@@ -51,6 +51,11 @@ void FileWriter::init(bool setOutput, const std::string& simulationId)
     }
 }
 
+const std::filesystem::path& FileWriter::outputPath() const
+{
+    return outputPath_;
+}
+
 void FileWriter::writeSimulationTable(
   const Optimisation::LinearProblemApi::ILinearProblem& linearProblem,
   const Optimisation::LinearProblemApi::IMipSolution& solution,
@@ -79,15 +84,4 @@ FileWriter::FileWriter(std::filesystem::path path):
 {
 }
 
-void FileWriter::writeProblem(
-  const Optimisation::LinearProblemMpsolverImpl::OrtoolsLinearProblem& problem,
-  const std::string& filename)
-{
-    if (output)
-    {
-        logs.info() << "Writing " << filename << "...";
-        const auto lp_path = outputPath_ / filename;
-        problem.Write(lp_path.string());
-    }
-}
 } // namespace Antares::Modeler
--- a/src/modeler/FileWriter.cpp
+++ b/src/modeler/FileWriter.cpp
@@ -80,13 +80,14 @@ FileWriter::FileWriter(std::filesystem::path path):
 }
 
 void FileWriter::writeProblem(
-  const Optimisation::LinearProblemMpsolverImpl::OrtoolsLinearProblem& problem)
+  const Optimisation::LinearProblemMpsolverImpl::OrtoolsLinearProblem& problem,
+  const std::string& filename)
 {
     if (output)
     {
         logs.info() << "Writing problem.lp...";
-        const auto lp_path = outputPath_ / "problem.lp";
-        problem.WriteLP(lp_path.string());
+        const auto lp_path = outputPath_ / filename;
+        problem.Write(lp_path.string());
     }
 }
 } // namespace Antares::Modeler
--- a/src/modeler/FileWriter.cpp
+++ b/src/modeler/FileWriter.cpp
@@ -34,20 +34,16 @@
 
 namespace Antares::Modeler
 {
-void FileWriter::init(bool setOutput, const std::string& simulationId)
+void FileWriter::init(const std::string& simulationId)
 {
-    output = setOutput;
     outputPath_ = studyPath_ / "output";
     simulationId_ = simulationId;
-    if (output)
+    logs.info() << "Output folder : " << outputPath_;
+    if (!std::filesystem::is_directory(outputPath_)
+        && !std::filesystem::create_directory(outputPath_))
     {
-        logs.info() << "Output folder : " << outputPath_;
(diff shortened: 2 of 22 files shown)


src/modeler/modeler/Modeler.cpp
```diff
@@ -144,38 +142,52 @@ void Modeler::run() const
...
 
     // gp : class SystemLinearProblemBuilder should be renamed into ComponentFillersBuilder
     // gp : and build() should return the vector of component fillers
-    SystemLinearProblemBuilder system_linear_problem(data.system.get(),
-                                                     ortools_linear_problem,
-                                                     *data.dataSeries,
-                                                     data.scenarioGroupRepository,
-                                                     masterAndSubPbvars.get());
+    // Subproblem
+    SystemLinearProblemBuilder subproblem_builder(data.system.get(),
```
guilpier-code, line 165: I would expect that we first build + dump the **master problem**, and then build + dump+ solve the **sub-problem**.

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -104,11 +104,22 @@ void MasterAndSubPbVariables::add(const std::vector<std::string>& names, unsigne
...
+{
+    for (const auto& [problemId, v]: masterAndSubPbVars_)
+    {
+        for (const auto& [variableName, variableIndex]: v)
+        {
+            os << problemId << '\t' << variableName << '\t' << variableIndex << '\n';
+        }
+    }
+}
+
```
guilpier-code, line 124: Note that line (line # 295): 
```c++
const auto& evaluationContext = optimEntityContainer_.getEvaluationContext(component_);
```
is useless

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -334,7 +345,8 @@ void ComponentFiller::addVariables(const LinearProblemApi::FillContext& ctx)
                            dims);
         }
 
-        if (variable.location() == Modeler::Config::Location::MASTER_AND_SUBPROBLEMS
+        if ((variable.location() == Modeler::Config::Location::MASTER_AND_SUBPROBLEMS
```
guilpier-code: Why do we also add variables belonging to **MASTER** problem only to **masterAndSubPbvars_** ?
They seem to have nothing to do with content of ile **structure.txt**

src/modeler-optimisation-container/OptimEntityContainer.cpp
```diff
@@ -47,13 +48,21 @@ void OptimEntityContainer::addFromSystemComponents(const std::vector<Component>&
...
+            {
+                modelVariableGlobalIndices.push_back(variableGlobalIndex);
+                ++variableGlobalIndex;
+            }
+            else
+            {
+                // We add dummy data to avoid "holes" in modelVariableGlobalIndices
+                // but these are not to be used
+                modelVariableGlobalIndices.push_back(-1);
+            }
```
guilpier-code, line 65: Why do we need to add `-1`s ? 
Does it even work considering what is done in **ComponentFiller::addVariables(...)** regarding variables counts ?
I mean that, in this (previous) member function : 
```c++
optimEntityContainer_.addStartColumn();
```
is called only for variables that are "compatible" with the problem type.

src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/linearProblem.h
```diff
@@ -86,7 +86,7 @@ class ILinearProblem
     /// Solve the problem, returns a IMipSolution
     virtual IMipSolution* solve(bool verboseSolver) = 0;
 
-    virtual void WriteLP(const std::string& filename) const = 0;
+    virtual void Write(const std::string& filename) const = 0;
```
guilpier-code: This member function **Write** does not have to be part of **ILinearProblem** interface class : what if we want to write a linear problem in another way ? Would we add a new write member function to the interface ?
We should lighten this interface class by removing this function, and for class  `OrtoolsLinearProblem`, this **Write** function could be a free function : 
```c++
void writeMPS(const OrtoolsLinearProblem& problem, const std::string & filename);
```

src/optimisation/linear-problem-mpsolver-impl/include/antares/optimisation/linear-problem-mpsolver-impl/linearProblem.h
```diff
@@ -79,7 +79,7 @@ class OrtoolsLinearProblem: public LinearProblemApi::ILinearProblem
     bool isMaximization() const override;
 
     OrtoolsMipSolution* solve(bool verboseSolver) override;
-    void WriteLP(const std::string& filename) const override;
+    void Write(const std::string& filename) const override;
```
guilpier-code: see above comment on parent interface class

src/solver/optim-model-filler/ComponentFiller.cpp
guilpier-code, line 1: About filtering **variables** or **objectives**, we should consider using piping with a filter code style.
That's the reason why it was introduced in C++20, as you well know, and we naturally use "filter" when we talk about it.

src/modeler/FileWriter.h
```diff
@@ -54,8 +54,8 @@ class FileWriter: public Solver::IWriter
       const Optimisation::LinearProblemApi::FillContext& fillContext) const override;
     explicit FileWriter(std::filesystem::path path);
 
-    void writeProblem(
-      const Optimisation::LinearProblemMpsolverImpl::OrtoolsLinearProblem& problem) override;
```
guilpier-code, line 58: Why is class **FileWriter** for ? Do we need a class for what's done ?
- it has 2 member functions, doing different things, not consistent with one another, apart from "dumping".
- this class has 4 private data : 
  - **simulationId_** is shared by ... only one function (**writeSimulationTable(...)**), so what the point sharing this ?
  - **output** : `if(output)` is repeated in both functions. Then it should be used outside.
  - **studyPath_** : shouldn't be a member data because only used to build another data member.
- It seems the only reason why we put these 2 functions together in a class is because they share an output directory. So why not pass it as argument ?
- parent abstract class is useless : an interface class should not be created until we need it, that is if it has at least 2 children. So here it should be removed.

Instead, we could have : 
```c++
if (!parameters.noOutput)
{
	std::filesystem::path output_dir = createOuputDirectory(studyPath);
    Write(subproblem, "1-1.mps", output_dir);
	Write(master_problem, "master.mps", output_dir);
}
...
switch (solution->getStatus())
{
...
case MipStatus::FEASIBLE:
    WriteSimulationTable(...);
}
```

src/modeler/modeler/Modeler.cpp
```diff
@@ -144,38 +142,52 @@ void Modeler::run() const
...
 
     measure.tick();
     logs.info();
     logs.info() << "Modeler build took " << measure.toStringInSeconds();
 
     const auto simulationTableSuffix = formatTime(getCurrentTime(), "%Y%m%d-%H%M");
     writer_.init(!parameters.noOutput, simulationTableSuffix);
-    writer_.writeProblem(ortools_linear_problem);
+    writer_.writeProblem(subproblem, "1-1.mps");
+    writer_.writeProblem(master_problem, "master.mps");
```
guilpier-code: see [this comment ](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3211/files?show-viewed-files=true&file-filters%5B%5D=#r2494651820)

src/tests/resources/modeler/13_1/input/model-libraries/lib_example_13_1.yml
```diff
@@ -89,6 +89,7 @@ library:
           scenario-dependent: false
           time-dependent: false
       variables:
+        - id: master_only
```
guilpier-code, line 92: do we want to keep this variable **master_only** ?

flomnes: Yes ! This allowed us to spot a major bug with indices

src/modeler-optimisation-container/OptimEntityContainer.cpp
flomnes: We add dummy indices here to avoid index shifts (filtering)

src/modeler/modeler/Modeler.cpp
flomnes: All the same ! I can do the master problem first

src/solver/optim-model-filler/ComponentFiller.cpp
flomnes: Yes, I removed it on develop and can remove it here as well to avoid conflicts

src/solver/optim-model-filler/ComponentFiller.cpp
flomnes: You're absolutely right, I will make the change

src/optimisation/linear-problem-mpsolver-impl/include/antares/optimisation/linear-problem-mpsolver-impl/linearProblem.h
flomnes: done

src/solver/optim-model-filler/ComponentFiller.cpp
flomnes: Yes that can be done

src/modeler/FileWriter.h
flomnes: Yes, the idea of a `FileWriter` is quite nice but the actual design isn't very good

src/modeler/modeler/Modeler.cpp
flomnes: I introduced a free `Write` function

src/optimisation/linear-problem-api/include/antares/optimisation/linear-problem-api/linearProblem.h
flomnes: done

src/solver/optim-model-filler/ComponentFiller.cpp
guilpier-code: We could even define the new **locationFilter** at construction of **ComponentFiller**. Same for objectives.

src/optimisation/linear-problem-mpsolver-impl/include/antares/optimisation/linear-problem-mpsolver-impl/linearProblem.h
```diff
@@ -79,11 +79,12 @@ class OrtoolsLinearProblem: public LinearProblemApi::ILinearProblem
     bool isMaximization() const override;
 
     OrtoolsMipSolution* solve(bool verboseSolver) override;
-    void WriteLP(const std::string& filename) const override;
 
     double infinity() const override;
     bool isLP() const override;
 
+    friend void Write(const OrtoolsLinearProblem& problem, const std::filesystem::path& path);
```
guilpier-code, line 86: Here we still have dependency to an external **write** free function.
We should remove this friendship. 
This will cause a compilation error.
To make it go away,  we should make the following method **public** : 
```c++
operations_research::MPSolver* MpSolver() const;
```
and use it inside : 
```c++
void Write(const OrtoolsLinearProblem& problem, const std::filesystem::path& path)
```
If you hesitate to make **MpSolver()** public, don't : it **public** in class **LegacyOrtoolsLinearProblem**

src/modeler/modeler/Modeler.cpp
guilpier-code: Please see [this (further) comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3211#discussion_r2499093906)

src/modeler/FileWriter.cpp
guilpier-code, line 1: Remove the boolean data member **output**, and the `if (output)` in class **FileWriter** : we don't need it if the client code uses the writer inside a : 
```c++
if (!parameters.noOutput)
{
    ...
}
```

src/modeler/modeler/Modeler.cpp
```diff
@@ -143,39 +142,60 @@ void Modeler::run() const
...
     logs.info() << "Modeler build took " << measure.toStringInSeconds();
 
     const auto simulationTableSuffix = formatTime(getCurrentTime(), "%Y%m%d-%H%M");
-    writer_.init(!parameters.noOutput, simulationTableSuffix);
-    writer_.writeProblem(ortools_linear_problem);
+
+    if (!parameters.noOutput)
+    {
+        writer_.init(!parameters.noOutput, simulationTableSuffix);
+        auto output = writer_.outputPath();
```
guilpier-code, line 188:
```c++
auto output = writer_.outputPath();
```
not intuitive : a writer returning a output path.
We should compose the output path outside the writer and pass it as argument to writer, for example at construction.
And remove the member function `outputPath()`

flomnes: I agree but it's not blocking for the feature

PR #3212, opened 2025-11-05
[ANT-3642] 5.6, 5.7 max, min and pow operators

a-zakir:


src/expressions/visitors/LinearityVisitor.cpp
```diff
@@ -120,18 +123,103 @@ LinearStatus LinearityVisitor::visit([[maybe_unused]] const Nodes::AllTimeSumNod
...
+    }
+
+    // If base is constant, the result is constant
+    if (base == LinearStatus::CONSTANT)
+    {
+        return LinearStatus::CONSTANT;
+    }
+    EvalVisitor evalVisitor(optimEntityContainer_, fillContext_, component_);
+    auto exponentValue = evalVisitor.dispatch(operands[1]).valueAsDouble();
+    if (std::abs(exponentValue) < 1e-12)
```
flomnes: Magic value here

src/expressions/visitors/LinearityVisitor.cpp
```diff
@@ -120,18 +123,103 @@ LinearStatus LinearityVisitor::visit([[maybe_unused]] const Nodes::AllTimeSumNod
...
+    {
+        return LinearStatus::CONSTANT;
+    }
+    // If base is linear and exponent is 1 the result is linear
+    if (base == LinearStatus::LINEAR && std::abs(exponentValue - 1) < 1e-12)
```
flomnes: Magic value here

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -389,32 +397,137 @@ std::any ConvertorVisitor::visitDual(ExprParser::DualContext* context)
...
+    {
+        throw std::invalid_argument("reduced_cost operator expect only one variable id");
+    }
+
+    return static_cast<Node*>(
+      registry_.create<FunctionNode>(FunctionNodeType::reduced_cost, nodes.at(0)));
+}
+template<class T>
+std::any ConvertorVisitor::processPower(std::vector<T*> exprContexts,
+                                        const std::function<std::string()>& toStringTreeCallBack)
```
flomnes: Not sure if we really need a callback in this case. Since it accepts no arguments, it seems like a `std::string` would be enough.

Also, a `std::function` can be empty, `operator()` throws `std::bad_function_call` if `*this` does not store a callable function target, i.e. `!*this == true`, see [`operator()`](https://en.cppreference.com/w/cpp/utility/functional/function/operator().html).

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -389,32 +397,137 @@ std::any ConvertorVisitor::visitDual(ExprParser::DualContext* context)
...
+    if (exprContexts.at(0) == nullptr)
+    {
+        throw std::invalid_argument("bad power expression, the base is invalid in "
+                                    + toStringTreeCallBack());
+    }
+    if (exprContexts.at(1) == nullptr)
+    {
+        throw std::invalid_argument("bad power expression, the exponent is invalid in "
+                                    + toStringTreeCallBack());
+    }
```
flomnes: Not sure this can happen

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -389,32 +397,137 @@ std::any ConvertorVisitor::visitDual(ExprParser::DualContext* context)
...
+}
+
+std::any ConvertorVisitor::handleMin(ExprParser::ArgListContext* context)
+{
+    const auto nodes = std::any_cast<std::vector<Node*>>(context->accept(this));
+    if (nodes.size() != 2)
+    {
+        throw std::invalid_argument("min operator expect at least 2 operands got "
+                                    + std::to_string(nodes.size()));
+    }
```
flomnes: `min(a, b, c)` ?

src/modeler-optimisation-container/include/antares/modeler-optimisation-container/OptimEntityContainer.h
```diff
@@ -74,6 +74,17 @@ class OptimEntityContainer
...
+      const Antares::ModelerStudy::SystemModel::Component& component,
+      unsigned int index) const
+    {
+        auto* model = component.getModel();
+        auto& modelVar = model->Variables().at(index);
+        bool isVarTimeDep = modelVar.isTimeDependent();
+        bool isVarScenarioDep = modelVar.IsScenarioDependent();
+        return Optimisation::convertToTimeIndex(isVarTimeDep, isVarScenarioDep);
+    }
+
```
flomnes: Unused ?

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
```diff
@@ -210,6 +210,28 @@ TimeDependentLinearExpression TimeDependentLinearExpression::operator/(
...
+    return out;
+}
+
+TimeDependentLinearExpression TimeDependentLinearExpression::min(
+  const TimeDependentLinearExpression& other) const
+{
+    TimeDependentLinearExpression out(*this);
+    for (std::size_t t = 0; t < size(); t++)
+    {
+        out[t] = out[t].min(other[t]);
```
flomnes: It seems like this has useless copies (`operator=`). How about

```cpp
// (out[t] < other[t]) ? NOOP() : out[t] = other[t]
out[t].min(other[t]); // applyMin if you prefer
```
?

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -389,32 +397,137 @@ std::any ConvertorVisitor::visitDual(ExprParser::DualContext* context)
...
+    auto exprContexts = context->expr();
+    auto toStringTreeCallBack = [&context]() { return context->toStringTree(true); };
+    return processPower(exprContexts, toStringTreeCallBack);
+}
+
+std::any ConvertorVisitor::visitRightPower(ExprParser::RightPowerContext* context)
+{
+    auto exprContexts = context->right_expr();
+    auto toStringTreeCallBack = [&context]() { return context->toStringTree(true); };
+    return processPower(exprContexts, toStringTreeCallBack);
```
flomnes:
```suggestion
    return processPower(exprContexts, context->toStringTree(true));
```

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
flomnes: Also you may need to add a few
```cpp
if (other.size() > out.size())
{
    out.expandTo(other.size());
}
```
in order to resize `*this` if `other` is larger.

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
```diff
@@ -288,18 +288,85 @@ Antares::Optimization::TimeDependentLinearExpression ReadLinearExpressionVisitor
...
+        return handleDual(node);
+    case Nodes::FunctionNodeType::max:
+        return handleMax(node);
+    case Nodes::FunctionNodeType::min:
+        return handleMin(node);
+    case Nodes::FunctionNodeType::pow:
+        return handlePow(node);
+    default:
+        // TODO
+        throw std::runtime_error("");
```
flomnes: TODO

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
```diff
@@ -288,18 +288,85 @@ Antares::Optimization::TimeDependentLinearExpression ReadLinearExpressionVisitor
...
 
+Optimization::TimeDependentLinearExpression ReadLinearExpressionVisitor::handleMax(
+  const Nodes::FunctionNode* node)
+{
+    const auto& operands = node->getOperands();
+    // we know that max has at least two child
+    auto result(dispatch(operands.at(0)));
+    for (int i = 1; i < operands.size(); ++i)
+    {
+        const auto* operand = operands.at(i);
```
flomnes: I think `std::vector` has an `operator[] const`, but I may be wrong

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
```diff
@@ -288,18 +288,85 @@ Antares::Optimization::TimeDependentLinearExpression ReadLinearExpressionVisitor
...
 {
     throw Antares::Error::InvalidArgumentError(
       "A linear expression can't contain extra output operator dual.");
 }
 
+Optimization::TimeDependentLinearExpression ReadLinearExpressionVisitor::handleMax(
+  const Nodes::FunctionNode* node)
+{
+    const auto& operands = node->getOperands();
+    // we know that max has at least two child
```
flomnes:
```suggestion
    // we know that max has at least two children
```

src/expressions/include/antares/expressions/visitors/TimeIndexVisitor.h
```diff
@@ -57,8 +58,10 @@ class TimeIndexVisitor: public NodeVisitor<Optimisation::TimeIndex>
     Optimisation::TimeIndex visit(const Nodes::TimeIndexNode* timeIndexNode) override;
     Optimisation::TimeIndex visit(const Nodes::TimeSumNode* timeSumNode) override;
     Optimisation::TimeIndex visit(const Nodes::AllTimeSumNode* timeSumNode) override;
-    Optimisation::TimeIndex visit(const Nodes::ReducedCostNode* node) override;
-    Optimisation::TimeIndex visit(const Nodes::DualNode* node) override;
+    Optimisation::TimeIndex handleReducedCost(const Nodes::FunctionNode* node);
+    Optimisation::TimeIndex handleDual(const Nodes::FunctionNode* node);
+    Optimisation::TimeIndex handlePow(const Nodes::FunctionNode* node);
```
flomnes, line 63: These functions don't really need to be inside the class, they could be free functions

src/expressions/include/antares/expressions/visitors/LinearStatus.h
```diff
@@ -175,4 +175,15 @@ constexpr LinearStatus operator-(LinearStatus a)
...
+constexpr LinearStatus operator&(LinearStatus a, LinearStatus b)
+{
+    return operator+(a, b);
+}
+
+constexpr LinearStatus& operator&=(LinearStatus& a, LinearStatus b)
+{
+    a = a & b;
+    return a;
+}
```
flomnes: It seems that `operator&` really looks like `operator+`. Is there any reason to "duplicate" this operator ?

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -322,7 +328,26 @@ class EvalVisitor: public NodeVisitor<EvaluationResult>
...
+        const auto& operands = node->getOperands();
+        // we know that this function (max, min) has at least two child
+        auto result(dispatch(operands.at(0)));
+        for (int i = 1; i < operands.size(); ++i)
+        {
+            const auto* operand = operands.at(i);
+            result = result.applyOperation(dispatch(operand), op);
+        }
+        return result;
+    }
```
flomnes: I like that, it seems like another visitor has the same code, but with some duplication (min, max, etc.)

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -20,11 +20,11 @@
 */
 #pragma once
 
-#include <cmath>
 #include <functional>
 #include <sstream>
 #include <variant>
 
+#include <antares/expressions/nodes/FunctionNode.h>
```
flomnes: Shouldn't be necessary, it comes from ExpressionsNodes.h

src/expressions/include/antares/expressions/nodes/FunctionNode.h
```diff
@@ -22,45 +22,59 @@
...
+    {
+    }
+
+    explicit FunctionNode(FunctionNodeType type, const std::vector<Node*>& operands):
+        type_(type),
+        ParentNode(operands)
+    {
+    }
+
+    explicit FunctionNode(FunctionNodeType type, const std::vector<Node*>&& operands):
```
flomnes:
```suggestion
    explicit FunctionNode(FunctionNodeType type, std::vector<Node*>&& operands):
```

flomnes (commented):
Maybe document the procedure "how to add a new function node" ?

src/expressions/visitors/LinearityVisitor.cpp
a-zakir: Yes i don’t know what epsilon should i use

src/io/inputs/model-converter/convertorVisitor.cpp
a-zakir: Most Time it won’t be used that’s why it is not a string.

src/io/inputs/model-converter/convertorVisitor.cpp
a-zakir: Let me Check. However this is helpful in case of syntaxe error, could be used elsewhere

src/io/inputs/model-converter/convertorVisitor.cpp
a-zakir: Bad copy should be updated

src/io/inputs/model-converter/convertorVisitor.cpp
a-zakir: The string won’t be used most of the time

src/modeler-optimisation-container/include/antares/modeler-optimisation-container/OptimEntityContainer.h
flomnes: Unused.

src/io/inputs/model-converter/convertorVisitor.cpp
flomnes: I'm not sure I understand your argument

src/io/inputs/model-converter/convertorVisitor.cpp
a-zakir: don't worry about std::function being empty, the function is used correctly and privately two times.

src/io/inputs/model-converter/convertorVisitor.cpp
a-zakir: You want to replace the callback with a string for each call. The string will contain the tree structure and will only be used to provide more detailed error messages.

a-zakir: i'm avoiding the call to toStringTree for every time this operator is used.

src/expressions/include/antares/expressions/visitors/LinearStatus.h
a-zakir: the implementation is the same  
```
constexpr LinearStatus operator&(LinearStatus a, LinearStatus b)
{
    return operator+(a, b);
}

```
i don't want people to read linearStatus of max (a,b) = LinearStatus(a)+LinearStatus(b)

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
a-zakir: useless and and inefficient:
see 
```
const LinearExpression& TimeDependentLinearExpression::operator[](std::size_t idx) const
{
    if (isConstant())
    {
        return v_[0];
    }
    else
    {
        return v_[idx];
    }
}
```

src/modeler-optimisation-container/include/antares/modeler-optimisation-container/OptimEntityContainer.h
a-zakir: removed

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
a-zakir: yes if you don't want bound check [ ] is equivalent to .at()

src/solver/optim-model-filler/TimeDependentLinearExpression.cpp
a-zakir: you are right,

src/expressions/visitors/LinearityVisitor.cpp
a-zakir: replace with DEFAULT_THRESHOLD

src/io/inputs/model-converter/convertorVisitor.cpp
a-zakir: removed

src/expressions/include/antares/expressions/visitors/TimeIndexVisitor.h
a-zakir: free functions with the same names in different visitors break unity build

src/expressions/include/antares/expressions/visitors/LinearStatus.h
a-zakir: code is updated

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -122,7 +126,12 @@ class EvaluationResult
         return evaluateUnaryOperation(std::negate<>());
     }
 
-    [[nodiscard]] std::variant<double, std::vector<double>> value() const
+    // [[nodiscard]] std::variant<double, std::vector<double>> value() const
+    // {
+    //     return value_;
+    // }
```
flomnes:
```suggestion
```

src/expressions/include/antares/expressions/visitors/EvalVisitor.h
```diff
@@ -280,6 +290,34 @@ EvaluationResult EvaluationResult::evaluateUnaryOperation(Op op) const
...
+    {
+        throw std::invalid_argument("Expected at least two EvaluationResult");
+    }
+    const size_t size = getMaxSize(in);
+    std::vector<double> values(size);
+    std::vector<double> row(in.size());
+
+    for (size_t i = 0; i < size; ++i)
+    {
+        for (size_t j = 0; j < in.size(); ++j)
```
flomnes: I see two nested loops, it's not clear what `i` and `j` represent

flomnes (commented):
You only have time-independent tests, it would be nice to have at least a few tests with more than one timestep

a-zakir:
> You only have time-independent tests, it would be nice to have at least a few tests with more than one timestep
see
src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
 src\tests\src\solver\optim-model-filler\test_componentFiller.cpp

flomnes (approved):

PR #3213, opened 2025-11-05
Add 2 hydro remix tests [ANT-4022]

payetvin:


src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
guilpier-code, line 1: As you may know, there is another collection of unit tests on remix algorithm : **tests-multi-storage-remix.cpp**
I must admit that I would have introduced the new unit tests with 3 storage in this file, for the following reasons : 
- there are already 2 storage for each tests in **tests-multi-storage-remix.cpp**, whereas there's only one here.
- design is better to introduce multiple storage in **tests-multi-storage-remix.cpp**.

About this last reason : as we want to introduce 3 storage instead of 2, we could change a bit the design of the **fixture** : 
The idea would be to simply add a new template parameter to this fixture, representing the number of storage the a unit test allow to participate in the remix algorithm. 

```c++
template<unsigned nb_hours, unsigned nb_storage>
struct InputFixture
{
    InputFixture() : 
       ...
    {
        sts_holders.resize(nb_storage);
    }
    
    void runRemixStorageAlgorithm()
    {
        Load = TotaGenWithoutStorage + UnsupE;

        storagesForRemix.clear();
        for (unsigned sts_index = 0; sts_index < nb_storage; sts_index++)
        {
             Load += sts_holders[sts_index];
             storagesForRemix.push_back(sts_holders.createSTS(UnsupE));
        }

        shavePeaksByRemixingStorageGen(Load, UnsupE, Spillage, DTG_MRG, storagesForRemix);
    }
    ...
    std::vector<STS_holder<nb_hours>> sts_holders;
    ...
}
```

src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
```diff
@@ -53,15 +64,40 @@ struct InputFixture
...
         storagesForRemix.clear();
-        storagesForRemix.push_back(createHydroForRemix());
+        storagesForRemix.push_back(
+          createHydroForRemix(HydroGen, levels, inflows, init_level, capacity));
         shavePeaksByRemixingStorageGen(Load, UnsupE, Spillage, DTG_MRG, storagesForRemix);
     }
 
-    std::vector<double> TotalGenNoHydro, Load, HydroGen, UnsupE, levels, pmax, pmin, inflows, ovf,
-      pump, Spillage, DTG_MRG;
+    void callRemixStorageAlgorithmWith2or3storages(bool threeStorage = false)
```
guilpier-code:
```c++
void callRemixStorageAlgorithmWith2or3storages(bool threeStorage = false)
```
Design could be better : when there is a **or** or a **and** in the function name, this usually means it makes 2 things and could be split.
[Previous comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3213/files#r2519046254) tries to bring a solution to this : how to parameterize the number of storage.

src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
```diff
@@ -49,19 +60,43 @@ struct InputFixture
                                  reservoirManagement);
     }
 
-    void callRemixStorageAlgorithm()
+    void callRemixStorageAlgorithm(unsigned nbStorage = 1)
     {
-        Load = TotalGenNoHydro + UnsupE + HydroGen;
+        Load = TotalGenNoHydro + UnsupE + HydroGen + HydroGen2 + HydroGen3;
```
flomnes: strange that we take 3 storages into account even if `nbStorage < 3`

src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
```diff
@@ -49,19 +60,43 @@ struct InputFixture
...
+
+            if (nbStorage == 3)
+            {
+                auto storage3 = createHydroForRemix(HydroGen3,
+                                                    levels3,
+                                                    inflows3,
+                                                    init_level3,
+                                                    capacity3);
+                storagesForRemix.push_back(storage3);
+            }
```
flomnes, line 85: Can be improved, for example by creating a `struct` containing
- hydrogen
- levels
- inflows
- init level
- capacity

then `createHydroForRemix` takes this struct as it's only argument

src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
```diff
@@ -22,20 +23,30 @@ struct InputFixture
...
+        levels3.assign(size, 0.);
         pmax.assign(size, std::numeric_limits<double>::max());
         pmin.assign(size, 0.);
         inflows.assign(size, 0.);
+        inflows2.assign(size, 0.);
+        inflows3.assign(size, 0.);
         ovf.assign(size, 0.);
         pump.assign(size, 0.);
         Spillage.assign(size, 0.);
         DTG_MRG.assign(size, 0.);
```
flomnes, line 50: If you create a struct with constructor assign(size, 0) for each member, this can be shortened into
```cpp
InputFixture() : s1(size), s2(size), s3(size)
{
}
```

src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
```diff
@@ -517,3 +567,105 @@ BOOST_FIXTURE_TEST_CASE(comparison_of_results_with_python_algo,
...
+    // ------------------------------
+    // Run Remix algorithm
+    // ------------------------------
+    callRemixStorageAlgorithm(2);
+
+    // ------------------------------
+    // Compute flow balance after remix
+    // ------------------------------
+    double total_inflow_after = std::accumulate(inflows.begin(), inflows.end(), 0.0)
+                                + std::accumulate(inflows2.begin(), inflows2.end(), 0.0);
```
flomnes: wait, `inflows` is mutable ???

src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
payetvin: It's a vector of zero so no effect, avoid complicating the formula

src/tests/src/solver/simulation/remix-storage/test-hydro-remix.cpp
```diff
@@ -14,110 +15,139 @@
...
 using namespace Antares::Solver::Simulation;
 
+struct HydroStorageData
+{
+    HydroStorageData(size_t size)
+    {
+        hydroGen.assign(size, 0.);
+        levels.assign(size, 0.);
+        inflows.assign(size, 0.);
+    }
```
flomnes: Probably doesn't matter, but prefer member initialization
```suggestion
    HydroStorageData(size_t size) :
    hydroGen(size, 0.),
    levels(size, 0.),
    inflows(size, 0.)
    {
    }
```
this way you call this constructor `std::vector<T>::vector(std::size_t, const T&)` directly instead of `std::vector<T>::vector()` then `assign`.

flomnes (approved):

PR #3235, opened 2025-11-14
6.2c : towards restoring constness in system model objects

guilpier-code:
**What is done here** : 

Some work was done on branch **origin/feature/6.2c-optimConfigBehavior** in order to restore the "constness".
But this work was mixed with other changes.
This PR intends to extract the work on constness from the original work.


a-zakir:
👀

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/convertorVisitor.h
```diff
@@ -32,6 +38,42 @@ class NoPortWithThisId final: public std::runtime_error
...
+    {
+        it->second = true;
+
+        return convertLocation(it->first->location);
+    }
+    else
+    {
+        it->second = false;
+        return Modeler::Config::Location::SUBPROBLEMS;
+    }
```
a-zakir: may be this else is unnecessary

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/convertorVisitor.h
```diff
@@ -32,6 +38,37 @@ class NoPortWithThisId final: public std::runtime_error
...
+                                { return optimConfigEntity.first->id == id; });
+}
+
+template<class T>
+auto SearchEntity(const std::string id, const std::vector<std::pair<T*, bool>>& values)
+{
+    return std::ranges::find_if(values,
+                                [&](const auto& optimConfigEntity)
+                                { return optimConfigEntity.first->id == id; });
+}
```
flomnes, line 59: The `const` version seems to be unused

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/convertorVisitor.h
```diff
@@ -32,6 +38,37 @@ class NoPortWithThisId final: public std::runtime_error
...
+
+template<class T>
+Modeler::Config::Location updateLocation(std::string id, std::vector<std::pair<T*, bool>>& values)
```
flomnes, line 62: Naming: this function does not update the location of any object, it returns a Location. Consider a better name

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -331,11 +336,56 @@ std::vector<ModelerStudy::SystemModel::Objective> convertObjectives(
...
+class UnMatchedOptimConfig
+{
+public:
+    UnMatchedOptimConfig(const YmlOptimConfig::OptimConfig& optimConfig)
+    {
+        for (const auto& config: optimConfig)
+        {
+            for (const auto& variable: config.variables)
+            {
+                variables.emplace_back(std::make_pair(&variable, false));
```
flomnes, line 354:
```suggestion
                variables.emplace_back(&variable, false);
```

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -331,11 +336,56 @@ std::vector<ModelerStudy::SystemModel::Objective> convertObjectives(
...
+            }
+            for (const auto& objective: config.objectives)
+            {
+                objectives.emplace_back(std::make_pair(&objective, false));
```
flomnes, line 358:
```suggestion
                objectives.emplace_back(&objective, false);
```

src/io/inputs/model-converter/modelConverter.cpp
flomnes: No need to explicitly construct the object with `std::vector<T>::emplace_back`

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -344,22 +394,32 @@ std::vector<ModelerStudy::SystemModel::Objective> convertObjectives(
...
-        std::vector<ModelerStudy::SystemModel::Variable> variables = convertVariables(model);
+
+        std::vector<ModelerStudy::SystemModel::Variable> variables = convertVariables(
+          model,
+          unMatchedOptimConfig.variables);
+
         std::vector<ModelerStudy::SystemModel::Port> ports = convertPorts(model, portTypes);
         std::vector<ModelerStudy::SystemModel::PortFieldDefinition>
           portFieldDefinitions = convertPortFieldDefinitions(model, ports);
         std::vector<ModelerStudy::SystemModel::Constraint> constraints = convertConstraints(model);
```
flomnes, line 415: NOTE : constraints now have a location attribute

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/convertorVisitor.h
flomnes, line 62: If you want to clarify the intent you can add a const
```suggestion
Modeler::Config::Location updateLocation(std::string id, std::vector<std::pair<const T*, bool>>& values)
```

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/convertorVisitor.h
```diff
@@ -32,6 +38,37 @@ class NoPortWithThisId final: public std::runtime_error
...
+{
+    if (auto it = SearchEntity(id, values); it != values.end())
+    {
+        it->second = true;
+        return convertLocation(it->first->location);
+    }
+
+    return Modeler::Config::Location::SUBPROBLEMS;
```
flomnes, line 70:
```suggestion
    // By default, variables, constraints & objective contributions are in subproblems
    return Modeler::Config::Location::SUBPROBLEMS;
```

src/modeler/loadFiles/loadAll.cpp
```diff
@@ -37,10 +37,11 @@ Modeler::Data loadAll(const std::filesystem::path& studyPath)
     logs.info() << "Loading modeler files...";
     Modeler::Data data;
 
-    data.libraries = loadLibraries(studyPath);
-    logs.info() << "Libraries loaded";
+    const auto optimConfig = loadOptimConfig(studyPath);
+    logs.info() << "optim config is loaded";
```
flomnes, line 41:
```suggestion
    logs.info() << "Optim config loaded";
```

src/modeler/loadFiles/readOptimConfig.cpp
```diff
@@ -76,19 +76,18 @@ static void updateLibraries(const YmlOptimConfig::OptimConfig& obj,
...
         logs.info() << "Optim config file not found at " << configPath;
-        return;
+        return {};
     }
 
     std::string content = readOptimConfigFile(configPath);
-    const auto&& obj = parseOptimConfig(content, configPath);
-    updateLibraries(obj, libraries);
+    return parseOptimConfig(content, configPath);
+    // updateLibraries(obj, libraries);
```
flomnes, line 90:
```suggestion
```

src/modeler/loadFiles/include/antares/solver/modeler/loadFiles/loadFiles.h
```diff
@@ -39,7 +40,8 @@ Antares::Modeler::Data loadAll(const std::filesystem::path& studyPath);
 ModelerParameters loadParameters(const std::filesystem::path& studyPath);
 
 std::vector<ModelerStudy::SystemModel::Library> loadLibraries(
-  const std::filesystem::path& studyPath);
+  const std::filesystem::path& studyPath,
+  const IO::Inputs::YmlOptimConfig::OptimConfig& optimConfig = {});
```
flomnes, line 44:
```suggestion
  const IO::Inputs::YmlOptimConfig::OptimConfig& optimConfig);
```

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -344,22 +394,32 @@ std::vector<ModelerStudy::SystemModel::Objective> convertObjectives(
  */
 std::vector<ModelerStudy::SystemModel::Model> convertModels(
   const IO::Inputs::YmlModel::Library& library,
-  const std::vector<ModelerStudy::SystemModel::PortType>& portTypes)
+  const std::vector<ModelerStudy::SystemModel::PortType>& portTypes,
+  const YmlOptimConfig::OptimConfig& optimConfig = {})
```
flomnes, line 398:
```suggestion
  const YmlOptimConfig::OptimConfig& optimConfig)
```

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -331,11 +336,56 @@ std::vector<ModelerStudy::SystemModel::Objective> convertObjectives(
...
+            if (!matched)
+            {
+                throw std::invalid_argument(
+                  "objective : " + objective->id
+                  + " from optim-config.yml is not found in any library.");
+            }
+        }
+    }
+
+    std::vector<std::pair<const YmlOptimConfig::Variable*, bool>> variables;
```
flomnes, line 385: `private` members ?

flomnes: Also, consider using a `std::map<const T*, bool>` for ease of use & lookups

src/modeler/loadFiles/readLibraries.cpp
```diff
@@ -61,7 +63,7 @@ static ModelerStudy::SystemModel::Library loadSingleLibrary(const fs::path& file
 
     try
     {
-        return ModelConverter::convert(libraryObj);
+        return ModelConverter::convert(libraryObj, optimConfig);
```
guilpier-code, line 66:
```c++
return ModelConverter::convert(libraryObj, optimConfig);
```

Here we're supposed to **convert** yaml object into model business objects.
But we do something else :
- we check that elements in **optimConfig** already exist in libraries
- we merge **optimConfig** (more precisely its **location** part)  and libraries to build model variables (business objects)

It makes conversion more complicated than it should.
We could make the **check** and the **merge** from the 2 yaml sources in a previous step, and therefore simplify the **convert** step.

So we would have : 
```c++

try
{
    mergeOptimConfigToLibs(optimConfig, libraryObj);
}
catch(...)
...

try
{
    return ModelConverter::convert(libraryObj);
}
catch (...)
...
```

guilpier-code: We should have in mind that the merge yaml part will grow in the future, and will be easier to change and more clear if it's isolated from the conversion.

tbittar: Following a discussion with @guilpier-code , I agree that merging all required characteristics from the `optimConfig` and `libraryObj` seems clearer as operations and responsibilities are more clearly delimited. Especially, the `optimConfig` will be enriched in the future to include new characteristics and concepts, and possibly merging methods will be added and / or updated. With the design of Guillaume, it will be easy to identify which part of the code to modify and how, whereas the current design forces us to go at lower level and possibly make modifications in several methods which may not be clearly identified.

flomnes:
This PR is closed, since we decided to follow a different logic. Instead of parsing files separately, we will parse them together.

PR #3243, opened 2025-11-18
Storage remix : trial for fixes on debug print

guilpier-code:
What was done : 
- [x] remove collecting debug info from remix algorithm
- [x] make class **IStorageForRemix** hold the name of the storage

### DIFF
--- a/src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
+++ b/src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
@@ -14,14 +14,14 @@ const std::string shave_peak_remix_storage_error_msg_start = "Remix storage inpu
 namespace Antares::Solver::Simulation
 {
 
-void collectRemixDebugInfo(const ListStorageForRemix& storagesForRemix, std::stringstream* stream)
+void collectRemixDebugInfo(const ListStorageForRemix& storagesForRemix, std::stringstream& stream)
 {
     for (auto& storage: storagesForRemix)
     {
         unsigned hour = 0;
         for (auto& withdrawal: storage->withdrawal())
         {
-            *stream << storage->name() << " " << hour << " " << withdrawal << std::endl;
+            stream << storage->name() << " " << hour << " " << withdrawal << std::endl;
             hour++;
         }
     }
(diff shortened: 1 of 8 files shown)


src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -14,16 +14,16 @@ const std::string shave_peak_remix_storage_error_msg_start = "Remix storage inpu
...
-    {
-        *debugStream << s->name() << " " << h << " " << withdrawal[h] << std::endl;
+        unsigned hour = 0;
+        for (auto& withdrawal: storage->withdrawal())
+        {
+            *stream << storage->name() << " " << hour << " " << withdrawal << std::endl;
+            hour++;
+        }
     }
 }
```
flomnes: why not a reference `std::ostream&` ? I mean, it's not an optional argument and you don't check it against `nullptr`, so a reference would be more natural

payetvin (approved):

src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
guilpier-code: yes, good idea

src/solver/simulation/remix-storage/shave-peaks-by-remix-storage-gen.cpp
```diff
@@ -56,8 +56,7 @@ void shavePeaksByRemixingStorageGen(const std::vector<double>& Load,
                                     std::vector<double>& UnsupE,
                                     const std::vector<double>& Spillage,
                                     const std::vector<double>& DTG_MRG,
-                                    ListStorageForRemix& listStorage,
-                                    std::stringstream* debugStream)
+                                    ListStorageForRemix listStorage)
```
flomnes, line 59: No reference here ? Is this intentional ?

OK so this is a vector of shared_ptr. So we would copy this vector, copy the shared ptr inside and the effect looks the same. But still the copies are not needed

guilpier-code: yes it's intentional : we make a copy of the storage list when calling the remix algo.
Why ?
- this copy is light
- it allows to keep having access to all needed information about storage withdrawals outside the algorithm. With a reference, the storage list is changed by the algo, gets empty, and we loose this information.

flomnes (changes requested):
Use a reference for `ListStorageForRemix ` in function `shavePeaksByRemixingStorageGen`

flomnes (approved):

PR #3244, opened 2025-11-20
Storage remix tests : trial for improvement [ANT-4022]

guilpier-code:
What was done : 

- [x] move 2 tests to a unit tests context where many short term storage are tested, not reservoir levels. This context is more adapted to these 2 tests because, in real life, reservoir levels are alone, whereas there can be many STS.
- [x] Improve the fixture design, so that this fixture makes easy to involve as much STS as we want in a unit test
- [x] simplify these 2 tests a bit


src/solver/simulation/include/antares/solver/simulation/remix-storage/remix-utils.h
```diff
@@ -81,6 +82,11 @@ inline double min_on_subrange(std::vector<double>&& v, unsigned h, unsigned H)
     return *std::ranges::min_element(subset);
 }
 
+inline double sum(const std::vector<double>&& v)
+{
```
flomnes:
```suggestion
inline double sum(const std::vector<double>& v)
{
```

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
```diff
@@ -128,100 +134,109 @@ BOOST_AUTO_TEST_CASE(create_2_STS_for_nb_of_hours_not_equal___check_input_for_al
                           checkMessage(err_msg));
 }
 
+// Using directly InputFixture<n, m> in fixture test cases does not compile on Win MSVC.
```
flomnes, line 137: Yes, because `BOOST_FIXTURE_TEST_CASE` is a macro, <> messes with it

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
```diff
@@ -128,100 +134,109 @@ BOOST_AUTO_TEST_CASE(create_2_STS_for_nb_of_hours_not_equal___check_input_for_al
...
+// So we're forced to use typedefs instead.
+using InputFixture_5_2 = InputFixture<5, 2>;
+using InputFixture_5_3 = InputFixture<5, 3>;
+using InputFixture_8_2 = InputFixture<8, 2>;
```
flomnes: Please choose more explicit names (number of storages, number of timesteps, etc.)

flomnes (approved):

src/tests/src/solver/simulation/remix-storage/tests-multi-storage-remix.cpp
guilpier-code: Done

PR #3251, opened 2025-11-21
feat: [ANT-4168] setObjectiveOffset API

JasonMarechal25:
* Add setObjectiveOffset/getObjectiveOffset to ILinearProblem API
* In some case (Sirius) fallback to local handling when solver doesn't support setObjectiveOffset/getObjectiveOffset


src/optimisation/linear-problem-mpsolver-impl/linearProblem.cpp
```diff
@@ -42,11 +42,44 @@ void Write(const OrtoolsLinearProblem& problem, const std::filesystem::path& pat
...
     isLP_ = !isMip; // we don't care about pure integer prob
+
+    // Initialize objective offset callables based on solver type
+    if (solverName == "sirius")
+    {
+        // For Sirius: use a stateful functor
+        auto siriusHandler = std::make_shared<LocalObjectiveOffsetStorage>();
+        setOffsetCallable_ = [siriusHandler](double offset) { siriusHandler->setOffset(offset); };
+        getOffsetCallable_ = [siriusHandler]() { return siriusHandler->getOffset(); };
+    }
```
flomnes, line 97: The `std::function`s hold their own copy of `siriusHandler`, so all is good here. Capturing a raw pointer or a reference would cause a dangling pointer/reference though.

flomnes (commented):
LGTM, but it would be slightly more readable with an interface class, see #3265

flomnes (approved):

PR #3255, opened 2025-11-24
Small improvements for slack analysis

flomnes:
- Remove dependency on `std::type_index`
- Remove function used once
- Remove call `get()` for `std::shared_ptr`s

### DIFF
--- a/src/solver/infeasible-problem-analysis/report.cpp
+++ b/src/solver/infeasible-problem-analysis/report.cpp
@@ -28,6 +28,7 @@ namespace Antares::Optimization
 InfeasibleProblemReport::InfeasibleProblemReport(
   const std::vector<const operations_research::MPVariable*>& slackVariables)
 {
+    // Build constraints from slack variables (one constraint per slack variable)
     const ConstraintsFactory constraintsFactory;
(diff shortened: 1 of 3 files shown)


src/solver/infeasible-problem-analysis/report.cpp
```diff
@@ -22,18 +22,11 @@
...
 #include <algorithm>
 #include <regex>
-#include <typeindex>
 
 namespace Antares::Optimization
 {
 InfeasibleProblemReport::InfeasibleProblemReport(
   const std::vector<const operations_research::MPVariable*>& slackVariables)
-{
-    buildConstraintsFromSlackVars(slackVariables);
```
guilpier-code, line 32: ok, it's shorter now. 
That said, the advantage of having a **buildConstraintsFromSlackVars** method is to have a "title" for a treatment. It announces and clarifies the current task.

guilpier-code (approved):

src/solver/infeasible-problem-analysis/report.cpp
flomnes: I added a comment

PR #3257, opened 2025-11-25
feat: Time dependent objective is not supported

JasonMarechal25:


tbittar (approved):

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -396,33 +397,43 @@ void ComponentFiller::addConstraints(const LinearProblemApi::FillContext& ctx)
...
+
 void ComponentFiller::addObjectives(const LinearProblemApi::FillContext& ctx)
 {
     auto* model = component_.getModel();
-    const auto& solverVariables = optimEntityContainer_.getVariables();
     ReadLinearExpressionVisitor visitor(optimEntityContainer_, ctx, component_);
 
     for (const auto& objective: model->Objectives() | locationFilter())
     {
+        const auto root_node = objective.expression().RootNode();
```
guilpier-code, line 419:
```c++
const auto root_node = objective.expression().RootNode();
```
Now **root_node**  is defined, we should use it instead of `objective.expression().RootNode()` whenever it comes up.

src/solver/optim-model-filler/ComponentFiller.cpp
```diff
@@ -396,33 +397,43 @@ void ComponentFiller::addConstraints(const LinearProblemApi::FillContext& ctx)
...
         const auto linearExpression = visitor.visitMergeDuplicates(
           objective.expression().RootNode());
 
-        auto& pb = optimEntityContainer_.Problem();
-        for (const auto& expr: linearExpression)
+        const auto timeIndex = getConstraintTimeIndex(root_node, component_);
+        if (timeIndex == TimeIndex::VARYING_IN_TIME_ONLY
```
guilpier-code, line 423: Instead of : 
```c++
timeIndex == TimeIndex::VARYING_IN_TIME_ONLY
            || timeIndex == TimeIndex::VARYING_IN_TIME_AND_SCENARIO
```
we should have a : 
```c++
bool isVaryingWithTime(const TimeIndex& timeIndex)
{
    return timeIndex == TimeIndex::VARYING_IN_TIME_ONLY
            || timeIndex == TimeIndex::VARYING_IN_TIME_AND_SCENARIO;
}
```
defined in **TimeIndex.h**

guilpier-code: By the way : **TimeIndex** is a very misleading name for the associated **enum**.
Applied to an expression / node, It actually represents a variation property or a dependency status.
But **TimeIndex** suggests an index in time, that is a particular hour.
We should rename this whenever possible.

JasonMarechal25: Good suggestion
I've implemented it in a new PR  : https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3261

PR #3263, opened 2025-11-27
some improvement

a-zakir:
-. **Eliminates dangling pointer risk** - no raw pointers to map elements that could invalidate

-. **Single source of truth** - reserve participation data exists only in the map

-. **Cleaner duplicate detection** - dedicated `symmetricDuplicates()` method vs nested loops

-. **Improved `symmetricalIndices`** - early termination with direct name comparison

-. **Easier to maintain** - simpler data structure with fewer moving parts

-. **Type safety improvement** - `const std::set<ReserveName>&` prevents accidental modifications

### DIFF
--- a/src/libs/antares/study/include/antares/study/area/reserveParticipationContainer.h
+++ b/src/libs/antares/study/include/antares/study/area/reserveParticipationContainer.h
@@ -53,15 +53,8 @@ public:
 
     bool symmetricDuplicates(const std::set<ReserveName>& names) const
     {
-        for (const auto& sym: reserveParticipationsSymmetries)
-        {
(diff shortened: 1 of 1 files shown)


src/libs/antares/study/include/antares/study/area/reserveParticipationContainer.h
```diff
@@ -51,63 +51,56 @@ class ReserveParticipationContainerBase
...
-            }
-            else
-            {
-                throw std::out_of_range("This entity is not participating to reserve " + name);
+                return true;
             }
         }
-        for (auto existingSymmetry: reserveParticipationsSymmetries)
+        return false;
+    }
```
flomnes: use `std::any_of` instead of a loop here ?

PR #3271, opened 2025-12-02
Single problem getter - Week indexing starts at 1

flomnes:


src/api/singleProblemGetterImpl.cpp
```diff
@@ -151,9 +153,15 @@ ConstantDataFromAntares SingleProblemGetter::getConstantData()
 
 WeeklyDataFromAntares SingleProblemGetter::getWeeklyData(WeeklyProblemId id)
 {
-    const auto [year, week] = id;
+    auto [year, week] = id;
+    // by convention, weeks start at 1 from the caller's POV, but at 0 in Simulator
+    if (week == 0)
+    {
+        throw std::invalid_argument("Invalid week number 0 detected, week number must be >=1");
```
a-zakir: std::logic_error or domain_error

flomnes: [`std::out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range.html) maybe ?

JasonMarechal25 (approved):

PR #3274, opened 2025-12-03
Feature/fix bug multithread reserves

h-fournier:
A bug is happening when running the reserves code in parallel, this branch is here to solve this issue.

Bug explanation : 
The problem was that the object  `reserveParticipationIndexMaps`  was initialized during the output process.  `reserveParticipationIndexMaps` is in an Area, which is an object shared by multiple threads during parallel execution. 
When multiple threads were doings the output process, they could both corrupt data while the other was using it and it resulted in two bugs : a bad optionnal access and a BiMap error because keys were not found.

Solution:
 `reserveParticipationIndexMaps` is now initialized during the loading part of the study which is done sequentially, before sharing the data with all the threads.


src/solver/variable/include/antares/solver/variable/area.hxx
```diff
@@ -502,16 +502,6 @@ void Areas<NextT>::weekBegin(State& state)
...
-
-    if (state.study.parameters.reservesEnabled)
-    {
-        state.study.areas.each(
-          [&state](Data::Area& area)
-          {
-              state.area = &area;
-              state.initReserveParticipationIndexMaps();
-          });
-    }
```
flomnes, line 514: There should be no concurrent write on `state` since each thread has their own instance. See

https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/960b3703316f1e23afa1845814daeccc95719e98/src/solver/simulation/include/antares/solver/simulation/solver.hxx#L158-L175

h-fournier: Yes, that was the bug and removing it is the fix

.github/workflows/centos7.yml
```diff
@@ -8,6 +8,7 @@ on:
       - develop
       - dependabot/**
       - feature/reserve_model_lot3_scalian
+      - feature/fix_bug_multithread_reserves
```
JasonMarechal25, line 11: issue: to remove

```suggestion

```

JasonMarechal25 (approved):

PR #3292, opened 2025-12-12
Short term storage :  unit tests improve

guilpier-code:
What is done here : 
- [x] unit tests still do exactly the same things they used to do
-  [x] renaming
-  [x] removing code duplication (especially by introducing fixtures)
-  [x] make test more simple

As a result, the test file is much shorter.

### DIFF
--- a/src/libs/antares/study/parts/short-term-storage/container.cpp
+++ b/src/libs/antares/study/parts/short-term-storage/container.cpp
@@ -152,7 +152,7 @@ static bool loadAdditionalConstraintsProperties(AdditionalConstraints* additiona
     return true;
 }
 
-bool STStorageInput::loadAdditionalConstraints(const fs::path& parentPath)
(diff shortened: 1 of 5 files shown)


a-zakir (approved):

flomnes (approved):
Please apply clang-format before merging

src/libs/antares/study/parts/short-term-storage/container.cpp
```diff
@@ -142,21 +142,27 @@ static bool loadAdditionalConstraintsProperties(AdditionalConstraints* additiona
...
+        {
+            logs.error() << "Constraint " << additionalConstraints->name << " : "
+                         << "has a wrong key : " << key;
+            return false;
+        }
     }
     return true;
 }
 
 bool STStorageInput::loadAdditionalConstraints(const fs::path& parentPath)
```
guilpier-code: This function could be easily split into 2 parts : 
- load additional constraints properties from ini file
- load additional constraints associated RHS

PR #3299, opened 2025-12-16
Remove some ccache options

payetvin:


payetvin:
Work done and merged in #3318

PR #3301, opened 2025-12-16
Short term storage : split function that loads storage additional constraints

guilpier-code:


flomnes (approved):
Changes the logic a bit, but for the better
## Before

1. Load properties 1
2. Load RHS 1
3. Load properties 2
4. Load RHS 2
5. etc.

## After
1. Load properties 1
2. Load properties 2
3. etc.
4. Load RHS 1
5. Load RHS 2
6. etc.

PR #3307, opened 2025-12-18
[100%] Testing mps writer

a-zakir:


src/tests/src/io/outputs/testMpsWriter.cpp
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    static void checkProblemType(const unique_ptr<ILinearProblem>& originalProblem,
+                                 const operations_research::mb::ModelBuilderHelper& fromMps)
+    {
+        bool isLp = true;
+        for (int vi = 0; vi < fromMps.num_variables(); ++vi)
+        {
+            isLp &= !fromMps.VarIsIntegral(vi);
+        }
```
flomnes: std::all_of

src/tests/src/io/outputs/testMpsWriter.cpp
```diff
@@ -0,0 +1,158 @@
...
+    {
+        BOOST_CHECK(originalProblem->variableCount() == fromMps.num_variables());
+        const auto& origVariables = originalProblem->getVariables();
+        for (int vi = 0; vi < fromMps.num_variables(); ++vi)
+        {
+            const auto& origVariable = origVariables.at(vi);
+            BOOST_CHECK(origVariable->getName() == fromMps.VarName(vi));
+            BOOST_CHECK(origVariable->isInteger() == fromMps.VarIsIntegral(vi));
+            BOOST_CHECK(origVariable->getLb() == fromMps.VarLowerBound(vi));
+            BOOST_CHECK(origVariable->getUb() == fromMps.VarUpperBound(vi));
```
flomnes: BOOST_CHECK_EQUAL

src/tests/src/io/outputs/testMpsWriter.cpp
```diff
@@ -0,0 +1,158 @@
...
+    {
+        if (entry.path().stem() == "epic_2")
+        {
+            for (const auto& subEntry:
+                 fs::directory_iterator(resources / "epic_2" / "us2.5") | dirFilter)
+            {
+                run(subEntry.path());
+            }
+        }
+        run(entry.path());
```
flomnes: Not sure I understand why there are two distinct calls to `run`

src/tests/src/io/outputs/testMpsWriter.cpp
```diff
@@ -0,0 +1,158 @@
...
+    static void run(const fs::path& studyPath)
+    {
+        Antares::Solver::LoadFiles::FileLoader loader(studyPath);
+        Antares::Modeler::FileWriter writer(studyPath);
+        Antares::Solver::Modeler modeler(loader, writer);
+        modeler.run();
+
+        checkProblem(modeler.masterProblem(), studyPath / "output" / "master.mps");
+        checkProblem(modeler.subproblems().at(0), studyPath / "output" / "1-1.mps");
+    }
```
flomnes: IMO this method should be split
1. Create stuff
2. Check stuff

"run" is a bit confusing

src/tests/src/io/outputs/testMpsWriter.cpp
a-zakir: yes that would be the right way if i had a vector of variable fromMps

src/tests/src/io/outputs/testMpsWriter.cpp
a-zakir: that would be possible if `modeler.masterProblem()` and `modeler.subproblems()` are stored...

a-zakir: I can rename this method to checkProblems, in fact, I'll introduce a parameter that disables resolution

src/tests/src/io/outputs/testMpsWriter.cpp
a-zakir: Because the studies in src/tests/resources/modeler/epic2/us2.5/ are not at the same level as those in src/tests/resources/modeler/.

src/tests/src/io/outputs/testMpsWriter.cpp
a-zakir: ok without further modifications (like moving problems from modeler), i put build lines in separate func.

PR #3314, opened 2025-12-23
[100%] print mps in api mode

a-zakir:
usage: _build/api/antares-api -i /path/to/sttudy/ --write-mps --output /path/to/output/20210110-0900eco


src/solver/optimisation/HebdoProblemToLpsTranslator.cpp
```diff
@@ -71,6 +71,7 @@ WeeklyDataFromAntares HebdoProblemToLpsTranslator::translate(
     copy(problem->Sens, ret.Direction);
     resizeIfLargerThan(ret.Direction, problem->NombreDeContraintes);
     copy(name, ret.name);
+    ret.solver_ = problem->ProblemesSpx[0].get();
```
flomnes: dangling pointer if the owner of `problem->ProblemesSpx[0]` releases the `unique_ptr`

src/api/singleProblemGetterImpl.cpp
```diff
@@ -270,6 +268,41 @@ WeeklyDataFromAntares SingleProblemGetter::getWeeklyData(WeeklyProblemId id)
...
+                                                                modelerDataSeries,
+                                                                modelerScenarioGroupRepository);
+
+        auto* solver = fillAndGetMpSolver(ortoolsProblem,
+                                          fillCtx,
+                                          &pb_,
+                                          optimEntityContainer,
+                                          true); // TODO
+
+        ProblemeAResoudre->ProblemesSpx[numeroDeLIntervalle].reset(solver);
```
flomnes: useless in most cases

src/api/include/antares/api/singleProblemGetter.h
```diff
@@ -34,6 +35,11 @@ class SingleProblemGetter;
 
 namespace Antares::Solver
 {
+inline std::string problemName(const WeeklyProblemId& id)
+{
+    return fmt::format("problem-{}-{}--optim-nb-1", id.year + 1, id.week);
+}
+
```
flomnes: Why move the function def here ?

src/api/main.cpp
```diff
@@ -35,19 +37,39 @@ unsigned int toInt(const char* in)
 }
 
 int main(int argc, char** argv)
-{
-    if (argc != 4)
+{ // dirty options reader
+    if (argc < 4 || argc > 5)
```
flomnes: We need a proper option parser

src/api/main.cpp
```diff
@@ -58,5 +119,92 @@ int main(int argc, char** argv)
                        constant.ConstraintsMeaning,
                        weekly.Direction,
                        weekly.RHS);
+}
+
+
+template<typename T>
+struct Generator {
```
flomnes: Unless there are very good reasons to do it, I would rather not use coroutines for something as mundane as a counter. The reason is most team members are not familiar with them, including myself

src/api/include/antares/api/singleProblemGetter.h
```diff
@@ -50,7 +51,8 @@ class SingleProblemGetter final
     // TODO[FOM] This should not be necessary
     void writeNTCTimeSeries(const std::filesystem::path& outputDir);
     void writeStudyDescriptionFiles(const std::filesystem::path& outputDir);
-
+    int nbYears() const;
```
flomnes, line 38: Careful there may be playlists, so some years may be skipped

src/api/main.cpp
```diff
@@ -33,21 +40,75 @@ unsigned int toInt(const char* in)
 {
     return strtoul(in, nullptr, 10);
 }
+struct ApiOptions
+{
+    std::string studyFolder = "";
+    std::string outputFolder = "";
+    unsigned int year = -1;
```
flomnes, line 44: std::optional is the way to go to avoid magic values

src/api/main.cpp
```diff
@@ -33,21 +40,75 @@ unsigned int toInt(const char* in)
...
+    unsigned int week = -1;
+    bool writeMps = false;
+};
+
+// Boost.ProgramOptions is another candidate
+Yuni::GetOpt::Parser Parser(ApiOptions& options)
```
flomnes, line 50: Nice 👍🏻

src/api/main.cpp
```diff
@@ -33,21 +40,75 @@ unsigned int toInt(const char* in)
...
+}
 
-    const unsigned int year = toInt(argv[2]);
-    const unsigned int week = toInt(argv[3]);
+void ValidateOptions(const ApiOptions& options)
+{
+    if (options.studyFolder.empty())
+    {
+        throw std::invalid_argument(
+          "Study Folder is empty, please enter valid study path checkout --help");
```
flomnes: I would say "invalid directory" or even better "directory not provided". An empty directory is something else

src/api/main.cpp
```diff
@@ -33,21 +40,75 @@ unsigned int toInt(const char* in)
...
+    }
+}
+void printWeek(const ConstantDataFromAntares& constant, const WeeklyDataFromAntares& weekly, const ApiOptions& options, const WeeklyProblemId& id){
+      if (options.writeMps)
+    {
+        std::string mps;
+        weekly.solver_->ExportModelAsMpsFormat(false, false, &mps);
+        std::cout<<"Printing problem: "<<problemName(id)<<std::endl;
```
flomnes: Use a logger

src/api/main.cpp
```diff
@@ -1,7 +1,14 @@
 #include <iostream>
+#include <coroutine>
+#include <fstream>
+
+#include "yuni/core/getopt/parser.h"
 
 #include "antares/api/singleProblemGetter.h"
 
+#include "private/singleProblemGetterImpl.h"
```
flomnes: Why do we need to depend on the implementation here ? Work in progress?

src/api/include/antares/api/singleProblemGetter.h
a-zakir: i need it in the main

src/api/include/antares/api/singleProblemGetter.h
a-zakir: see where it is updated

src/api/main.cpp
a-zakir: The Yuni parser requires a reference to an object, so you can’t pass an uninitialized `std::optional<unsigned int>`. Initializing the optional just to satisfy the parser defeats its purpose.

JasonMarechal25 (approved):
soft approval

PR #3316, opened 2025-12-23
Modeler : operators ceil and floor [ANT-4295]

guilpier-code:
Adding the **ceil** and **floor** operators.

You should know that the current PR is followed by 2 more PRs (in a sense based on this one and on one another) : 
- [PR 3329](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3329), whose purpose is to simplify the forbidden nodes part
- [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335), whose purpose is to add features on **ceil** and **floor** operators that were forgotten / not yet done in the current PR (especially **ceil** operator and unit tests).


guilpier-code (commented):
blabla

src/expressions/include/antares/expressions/nodes/FunctionNode.h
guilpier-code: **FunctionNodeTypeToString** : this function should only be defined in **.cpp** source file, as it's called inside the class only.

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
```diff
@@ -661,181 +562,143 @@ BOOST_AUTO_TEST_CASE(ValidMaxExpression)
...
+
     std::string err_msg = "'expression with =' is not allowed in this context '" + expression + "'";
-    BOOST_CHECK_EXCEPTION(ModelConverter::NodeChecker(forbidden, expression).dispatch(node.node),
-                          ModelConverter::BadContextComposition,
+    BOOST_CHECK_EXCEPTION(NodeChecker(forbidden, expression).dispatch(node.node),
+                          BadContextComposition,
+                          checkMessage(err_msg));
+}
+
+BOOST_FIXTURE_TEST_CASE(floor_operator___nominal_case, SupplyModelForFunctionalOperator)
```
guilpier-code, line 672: Here we add several unit tests about wrong use of **floor** operator. 
For each wrong use, an exception is raised.
Do we want to use the **NodeChecker** as well (see **modelConverter.cpp**) ? Wouldn't it be a double check or a kind of duplication ?

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -482,12 +484,51 @@ std::any ConvertorVisitor::handleMin(ExprParser::ArgListContext* context)
...
+
+            return static_cast<Node*>(
+              registry_.create<FunctionNode>(FunctionNodeType::floor, paramNode));
+        }
+    }
+
+    // Try to create a literal node
+    try
+    {
+        const double arg_as_d = std::stod(arg_as_str);
```
flomnes: 1. Use `std::from_chars` instead
2. In the case of a literal, the `std::floor` function needs to be applied to the result of the `std::string` -> `double` conversion.

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
```diff
@@ -1462,7 +1460,7 @@ BOOST_FIXTURE_TEST_CASE(PrintDualNode, MyDummyFixture)
                                       create<ParameterNode>("constraint"),
                                       create<LiteralNode>(0));
     PrintVisitor printVisitor;
-    BOOST_CHECK(printVisitor.dispatch(dual) == "dual(constraint)");
+    BOOST_CHECK(printVisitor.dispatch(dual) == "dual(constraint, 0.000000)");
```
guilpier-code: **Careful here** : a test check was changed. Is this was we want ?

src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -244,6 +244,20 @@ EvaluationResult EvalVisitor::handlePow(const Nodes::FunctionNode* node)
...
+    {
+        return std::floor(d);
+    }
+};
+
+EvaluationResult EvalVisitor::evaluateFloorNode(const Nodes::FunctionNode* node)
+{
+    auto* floor_arg = node->getOperands()[0];
+    return dispatch(floor_arg).evaluateUnaryOperation(Floor());
+}
```
flomnes: It's not necessary to construct a temporary `Floor` object here

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -482,12 +484,51 @@ std::any ConvertorVisitor::handleMin(ExprParser::ArgListContext* context)
...
+    }
+}
+
 std::any ConvertorVisitor::visitFunction(ExprParser::FunctionContext* context)
 {
     const auto functionName = context->IDENTIFIER()->getText();
     auto* arglist = context->argList();
 
-    if (!arglist)
+    if (!arglist || !arglist->expr().size())
```
flomnes: `!arglist || arglist->expr().empty()`

flomnes (approved):

src/expressions/visitors/EvalVisitor.cpp
guilpier-code: ok, but it's a bit more clear to give the floor argument a name.

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -482,12 +484,51 @@ std::any ConvertorVisitor::handleMin(ExprParser::ArgListContext* context)
...
+    }
+
+    // Try to create a literal node
+    try
+    {
+        const double arg_as_d = std::stod(arg_as_str);
+        return registry_.create<LiteralNode>(arg_as_d);
+    }
+    catch (const std::invalid_argument&)
+    {
```
a-zakir: Why are combinations of parameters and literals not handled in the current implementation?
For example:

floor(param * 3)
floor(param1 * param2)
floor(param * lit1 + lit2)
Note:
Since floor is a non-linear function, the following does not hold:
floor(literal * param) != literal * floor(param)

a-zakir: also floor can be applied to variable in extra-outputs

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
a-zakir: no

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
```diff
@@ -1319,22 +1317,22 @@ BOOST_FIXTURE_TEST_CASE(functionNode_pow, MyDummyFixture)
     PrintVisitor printVisitor;
     const auto printed = printVisitor.dispatch(pow);
 
-    BOOST_CHECK_EQUAL(printed, "22.000000^(2.000000)");
+    BOOST_CHECK_EQUAL(printed, "pow(22.000000, 2.000000)");
```
a-zakir, line 1320: why do you change this print ?

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
a-zakir: you shouldn't check in the convertvisitor, his responsability is only to convert antlr tree.

a-zakir: for now it is hard coded that floor expects either param or literal node. why would user do floor(2.3) when he can simply write 2 ?

src/io/inputs/model-converter/convertorVisitor.cpp
guilpier-code: Done

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
guilpier-code: I can reverse this.
But I did it because I tried to replace these help functions : 
- **handleReducedCost**, 
- **handleDual**, 
- **ProcessOtherFunction**, 
- **handlePow**,
- **handleFloor**

with a unique (more generic) function. and it was easier to use 'pow' (as a result of `node->typeToString()`)

a-zakir: Ok , i understand. Pow is unique among function nodes. It is the only one with a symbol. The printvisitor tries to reflect what the user may enter.

src/io/inputs/model-converter/convertorVisitor.cpp
guilpier-code: Visiting floor operator (see **ConvertorVisitor::visitFloor**) is now much more simple (as it should be), and requires use of neither `std::floor` nor `std::stod`

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
```diff
@@ -312,31 +294,45 @@ Optimization::TimeDependentLinearExpression ReadLinearExpressionVisitor::handleP
...
 }
 
-Optimization::TimeDependentLinearExpression ReadLinearExpressionVisitor::visit(
+TimeDependentLinearExpression ReadLinearExpressionVisitor::visitFloor(
   const Nodes::FunctionNode* node)
+{
+    auto expressions = dispatch(node->getOperands()[0]);
+    for (auto& expression: expressions)
+    {
+        // gp : should be removed when arg of floor operator applies to extra output as well.
```
flomnes: https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/3a6ce6438168c953858faa9d07b75e46d4bc4d40/src/io/outputs/SimulationTableGenerator.cpp#L265-L280

This visitor isn't used to evaluate `ExtraOutputs`, we use `EvalVisitor` for that.

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
```diff
@@ -161,6 +161,33 @@ BOOST_FIXTURE_TEST_CASE(visit__min_literal__param, VisitorFixture<ReadLinearExpr
     BOOST_CHECK_EQUAL(linear_expression[0].size(), 0);
 }
 
+BOOST_FIXTURE_TEST_CASE(linear_expr_from_floor_of_a_literal,
+                        VisitorFixture<ReadLinearExpressionVisitor>)
```
flomnes, line 165: You may also add `floor(4.5)*x`, where `x` is a variable node.

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
guilpier-code: ok, but why this remark ?

flomnes: You can remove the comment

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
guilpier-code: But a node of type `FunctionNodeType::dual` takes 2 arguments (a constraint and an index), and the **PrintVisitor** should print both. 
So what do we do ?
**dual(contraint)[index]** ?

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
guilpier-code: Comment was removed in PR [3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335)
See file **src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp**

src/expressions/include/antares/expressions/nodes/FunctionNode.h
guilpier-code: Was moved to corresponding  cpp file in PR [3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335)

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
guilpier-code: I changes dual node print into following format **dual(contraint)[index]**.
But I did this in [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335)

src/tests/src/solver/optim-model-filler/test_readLinearExpressionVisitor.cpp
guilpier-code: Done in [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335)

src/solver/optim-model-filler/ReadLinearExpressionVisitor.cpp
guilpier-code: Was removed in [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335)

src/tests/src/io/yml-importers/testConvertorVisitor.cpp
guilpier-code: Checks done in **convert visitor** turn out to be useful.
Indeed, there are 2 kind of prohibition about floor operators (an ceil) : 
- we have to control the number of arguments of floor (less than one and more than one is not accepted)
- we have to control the type of arguments floor can take and in which conditions.

First checks are done in the **convert visitor**, and I can see a better place to do them.

PR #3320, opened 2025-12-26
Adding ceil and floor operators : cleaning before work [ANT-4295]

guilpier-code:
This cleaning work comes before actual [work on ceil and floor operators](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3316)


src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -647,4 +626,32 @@ std::any ConvertorVisitor::visitRightExpression(ExprParser::RightExpressionConte
     return context->expr()->accept(this);
 }
 
+template<class T>
+std::any ConvertorVisitor::processPower(std::vector<T*> exprContexts)
```
flomnes:
```suggestion
std::any ConvertorVisitor::processPower(const std::vector<T*>& exprContexts)
```

flomnes (approved):

PR #3321, opened 2025-12-29
Add unit testing for adq patch debug files [ANT-4034]

flomnes:


src/tests/solver/optimisation/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+include(${CMAKE_SOURCE_DIR}/tests/macros.cmake)
+
+add_boost_test(tests-post-process
+        SRC test-post-process.cpp
+        LIBS
+        Antares::tests::in-memory-study
+        Antares::logs
```
payetvin: We need logs ?

payetvin (approved):

src/tests/solver/optimisation/CMakeLists.txt
flomnes: Indeed

PR #3323, opened 2026-01-06
Fixing jenkins structure.txt tests

payetvin:


payetvin:
fixed by using python 3.11 on jenkins side

PR #3324, opened 2026-01-06
Remove useless `std::unique_ptr` around temporary objects

flomnes:
1. Remove dependency on `std::unique_ptr`
2. Avoid dynamic allocations where not needed
3. Simplify the code overall

### DIFF
--- a/src/solver/hydro/daily/h2o_j_resoudre_le_probleme_lineaire.cpp
+++ b/src/solver/hydro/daily/h2o_j_resoudre_le_probleme_lineaire.cpp
@@ -19,8 +19,6 @@
 ** along with Antares_Simulator. If not, see <https://opensource.org/license/mpl-2-0/>.
 */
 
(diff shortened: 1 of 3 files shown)


JasonMarechal25 (approved):
I agree and approve.
However, Probleme is now on the stack, we just need to be sure that it is not an issue, for example is it goes above stack size limit

flomnes:
> I agree and approve. However, Probleme is now on the stack, we just need to be sure that it is not an issue, for example is it goes above stack size limit

On my Linux x86_64 machine,
```cpp
sizeof(PROBLEME_SIMPLEXE) = 264
```
so stack size shouldn't be an issue. See [here](https://github.com/rte-france/sirius-solver/blob/2580549970df6b975cac91936bfcd8d48fe86bc1/src/SIMPLEXE/spx_definition_arguments.h#L63-L177) for the structure definition.

PR #3329, opened 2026-01-09
Modeler : simplify forbidden nodes

guilpier-code:
This PR is the 2nd PR out of 3 PR about adding operators **ceil** and **floor** : 
- [PR 3316](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3316) (the main work on adding these operators)
- [PR 3329](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3329), (the current one)
- [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335) 

It intends to simplify and make more readable code about forbidden nodes. 

**What was done** :  some PR comments were left below. Please read them.

### DIFF
--- a/src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
+++ b/src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
@@ -78,21 +78,21 @@ void ForbiddenNodesVisitor::visit(const DivisionNode* divisionNode)
 void ForbiddenNodesVisitor::visit(const EqualNode* equalNode)
 {
     std::type_index nodeTypeId = typeIndexOf<EqualNode>();
-    checkIsForbidden(nodeTypeId, equalNode);
+    checkIsForbidden(equalNode, nodeTypeId);
     visitChildren(equalNode, nodeTypeId);
 }
 
 void ForbiddenNodesVisitor::visit(const LessThanOrEqualNode* lessThanOrEqualNode)
 {
     std::type_index nodeTypeId = typeIndexOf<LessThanOrEqualNode>();
-    checkIsForbidden(nodeTypeId, lessThanOrEqualNode);
+    checkIsForbidden(lessThanOrEqualNode, nodeTypeId);
     visitChildren(lessThanOrEqualNode, nodeTypeId);
 }
 
 void ForbiddenNodesVisitor::visit(const GreaterThanOrEqualNode* greaterThanOrEqualNode)
 {
     std::type_index nodeTypeId = typeIndexOf<GreaterThanOrEqualNode>();
-    checkIsForbidden(nodeTypeId, greaterThanOrEqualNode);
+    checkIsForbidden(greaterThanOrEqualNode, nodeTypeId);
     visitChildren(greaterThanOrEqualNode, nodeTypeId);
 }
 
@@ -108,7 +108,7 @@ void ForbiddenNodesVisitor::visit(const LiteralNode*)
 
 void ForbiddenNodesVisitor::visit(const VariableNode* variableNode)
 {
-    checkIsForbidden(typeIndexOf<VariableNode>(), variableNode);
+    checkIsForbidden(variableNode, typeIndexOf<VariableNode>());
 }
 
 void ForbiddenNodesVisitor::visit(const ParameterNode*)
@@ -118,39 +118,39 @@ void ForbiddenNodesVisitor::visit(const ParameterNode*)
 
 void ForbiddenNodesVisitor::visit(const PortFieldNode* portFieldNode)
 {
-    checkIsForbidden(typeIndexOf<PortFieldNode>(), portFieldNode);
+    checkIsForbidden(portFieldNode, typeIndexOf<PortFieldNode>());
 }
 
 void ForbiddenNodesVisitor::visit(const PortFieldSumNode* portFieldSumNode)
 {
-    checkIsForbidden(typeIndexOf<PortFieldSumNode>(), portFieldSumNode);
+    checkIsForbidden(portFieldSumNode, typeIndexOf<PortFieldSumNode>());
 }
 
 void ForbiddenNodesVisitor::visit(const TimeShiftNode* timeShiftNode)
 {
     std::type_index nodeTypeId = typeIndexOf<TimeShiftNode>();
-    checkIsForbidden(nodeTypeId, timeShiftNode);
+    checkIsForbidden(timeShiftNode, nodeTypeId);
     visitChildren(timeShiftNode, nodeTypeId);
 }
 
 void ForbiddenNodesVisitor::visit(const TimeIndexNode* timeIndexNode)
 {
     std::type_index nodeTypeId = typeIndexOf<TimeIndexNode>();
-    checkIsForbidden(nodeTypeId, timeIndexNode);
+    checkIsForbidden(timeIndexNode, nodeTypeId);
     visitChildren(timeIndexNode, nodeTypeId);
 }
 
 void ForbiddenNodesVisitor::visit(const TimeSumNode* timeSumNode)
 {
     std::type_index nodeTypeId = typeIndexOf<TimeSumNode>();
-    checkIsForbidden(nodeTypeId, timeSumNode);
+    checkIsForbidden(timeSumNode, nodeTypeId);
     visitChildren(timeSumNode, nodeTypeId);
 }
 
 void ForbiddenNodesVisitor::visit(const AllTimeSumNode* allTimeSumNode)
 {
     std::type_index nodeTypeId = typeIndexOf<AllTimeSumNode>();
-    checkIsForbidden(nodeTypeId, allTimeSumNode);
+    checkIsForbidden(allTimeSumNode, nodeTypeId);
     visitChildren(allTimeSumNode, nodeTypeId);
 }
 
@@ -182,12 +182,12 @@ void ForbiddenNodesVisitor::visit(const FunctionNode* functionNode)
         break;
     }
 
-    checkIsForbidden(nodeTypeId, functionNode);
+    checkIsForbidden(functionNode, nodeTypeId);
     visitChildren(functionNode, nodeTypeId);
 }
 
-void ForbiddenNodesVisitor::checkIsForbidden(const std::type_index& nodeTypeId,
-                                             const Node* node) const
+void ForbiddenNodesVisitor::checkIsForbidden(const Node* node,
+                                             const std::type_index& nodeTypeId) const
 {
     checkIGloballyForbidden(nodeTypeId, node);
     checkIsForbiddenByParent(nodeTypeId, node);
--- a/src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
+++ b/src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
@@ -154,7 +154,7 @@ void ForbiddenNodesVisitor::visit(const AllTimeSumNode* allTimeSumNode)
     visitChildren(allTimeSumNode, nodeTypeId);
 }
 
-std::type_index getTypeIndex(const FunctionNode* functionNode)
+std::type_index functionNodeTypeIndex(const FunctionNode* functionNode)
 {
     switch (functionNode->type())
     {
@@ -177,7 +177,7 @@ std::type_index getTypeIndex(const FunctionNode* functionNode)
 
 void ForbiddenNodesVisitor::visit(const FunctionNode* functionNode)
 {
-    std::type_index nodeTypeId(getTypeIndex(functionNode));
+    std::type_index nodeTypeId(functionNodeTypeIndex(functionNode));
     checkIsForbidden(functionNode, nodeTypeId);
     visitChildren(functionNode, nodeTypeId);
 }
--- a/src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
+++ b/src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
@@ -154,34 +154,30 @@ void ForbiddenNodesVisitor::visit(const AllTimeSumNode* allTimeSumNode)
     visitChildren(allTimeSumNode, nodeTypeId);
 }
 
-void ForbiddenNodesVisitor::visit(const FunctionNode* functionNode)
+std::type_index getTypeIndex(const FunctionNode* functionNode)
 {
-    std::type_index nodeTypeId(typeid(int)); // Must be default constructed here
-
     switch (functionNode->type())
     {
     case FunctionNodeType::reduced_cost:
-        nodeTypeId = typeIndexOf<FunctionNodeType::reduced_cost>();
-        break;
+        return typeIndexOf<FunctionNodeType::reduced_cost>();
     case FunctionNodeType::dual:
-        nodeTypeId = typeIndexOf<FunctionNodeType::dual>();
-        break;
+        return typeIndexOf<FunctionNodeType::dual>();
     case FunctionNodeType::min:
-        nodeTypeId = typeIndexOf<FunctionNodeType::min>();
-        break;
+        return typeIndexOf<FunctionNodeType::min>();
     case FunctionNodeType::max:
-        nodeTypeId = typeIndexOf<FunctionNodeType::max>();
-        break;
+        return typeIndexOf<FunctionNodeType::max>();
     case FunctionNodeType::pow:
-        nodeTypeId = typeIndexOf<FunctionNodeType::pow>();
-        break;
+        return typeIndexOf<FunctionNodeType::pow>();
     case FunctionNodeType::floor:
-        nodeTypeId = typeIndexOf<FunctionNodeType::floor>();
-        break;
+        return typeIndexOf<FunctionNodeType::floor>();
     default:
-        break;
+        return typeid(int); // Supposed to be dead code (never reached)
     }
+}
 
+void ForbiddenNodesVisitor::visit(const FunctionNode* functionNode)
+{
+    std::type_index nodeTypeId(getTypeIndex(functionNode));
     checkIsForbidden(functionNode, nodeTypeId);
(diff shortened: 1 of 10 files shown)


src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
```diff
@@ -33,154 +33,63 @@ namespace Antares::IO::Inputs::ModelConverter
...
-    case NT::pow:
-        return forbiddenNodeKey<NT::pow>();
-    case NT::dual:
-        return forbiddenNodeKey<NT::dual>();
-    case NT::reduced_cost:
-        return forbiddenNodeKey<NT::reduced_cost>();
-    default:
-        throw std::runtime_error("ForbiddenNodeKey is not implemented");
-    }
-}
```
guilpier-code, line 70: This function was removed (unused).

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
```diff
@@ -33,154 +33,63 @@ namespace Antares::IO::Inputs::ModelConverter
...
-inline std::type_index forbiddenNodeKey(const Expressions::Nodes::Node& node)
-{
-    if (auto* funcNode = dynamic_cast<const Expressions::Nodes::FunctionNode*>(&node))
-    {
-        return forbiddenNodeKey(funcNode->type());
-    }
-    return {typeid(node)};
+    using enum_item_as_type = std::integral_constant<Expressions::Nodes::FunctionNodeType, item>;
+    return std::type_index(typeid(enum_item_as_type));
 }
```
guilpier-code, line 79: This function was removed (unused).

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
```diff
@@ -33,154 +33,63 @@ namespace Antares::IO::Inputs::ModelConverter
...
-    [[nodiscard]] bool isForbidden() const
-    {
-        return global_.contains(forbiddenNodeKey<NodeType>());
-    }
-
-    template<Expressions::Nodes::FunctionNodeType NodeType>
-    [[nodiscard]] bool isForbidden() const
-    {
-        return global_.contains(forbiddenNodeKey<NodeType>());
-    }
```
guilpier-code, line 158: Here we have 2 versions of the same template function, and besides they are redundant with function **check(...)** (see below).
They were removed.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
```diff
@@ -33,154 +33,63 @@ namespace Antares::IO::Inputs::ModelConverter
...
-
-        // parent-specific forbidden child?
-        const auto& it = rules_.find(parentKey);
-        if (it == rules_.end())
-        {
-            return false;
-        }
-
-        return it->second.contains(childKey);
-    }
```
guilpier-code, line 183: This **check(...)** function checks both if a node is forbidden globally and if it is forbidden by a parent node.
So it does 2 things.
The function was split in 2 two functions : **isGloballyForbidden** and **isForbiddenByParent**

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
```diff
@@ -1,194 +0,0 @@
...
-#include <optional>
-#include <ranges>
-
-#include <antares/expressions/nodes/NodesForwardDeclaration.h>
-#include <antares/expressions/visitors/NodeVisitor.h>
-#include <antares/io/inputs/model-converter/ForbiddenNodes.h>
-
-namespace Antares::IO::Inputs::ModelConverter
-{
-class NodeChecker final: public Expressions::Visitors::NodeVisitor<void>
```
guilpier-code, line 33: class **NodeChecker** was renamed into **ForbiddenNodesVisitor**, more expressive.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
```diff
@@ -1,194 +0,0 @@
...
-        if (forbid_.isForbiddenFor<func>(typeIndex))
-        {
-            throw BadContextComposition(
-              {.expression = expression_, .childName = childName, .parentName = parentName});
-        }
-    }
-}
-
-template<typename Parent>
-void NodeChecker::checkChildren(const std::string& parentName,
```
guilpier-code, line 150: Template method **checkChildren(...)** : 
- has 2 versions for 2 types. Leads to code duplication.
- Makes 2 things (make a check on node prohibition **and** visit children), and this function does not do what its name says.

To fix these problems, the method was split and, as the template parameter turned out to be unnecessary, it was removed.
Now we have **visitChildren** instead.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
```diff
@@ -1,194 +0,0 @@
...
-    }
-
-    explicit BadContextComposition(const BadExpression& context):
-        invalid_argument(BuildMessage(context))
-    {
-    }
-};
-
-template<typename Child>
-void NodeChecker::checkConsistencyWithParents(const std::string& childName) const
```
guilpier-code, line 114: Template method **checkConsistencyWithParents** : same problem than **checkChildren**.
- code duplication
- makes 2 things and does not do what its name suggests.

It was renamed into **checkIsForbidden** and the template parameter was removed (to avoid code duplication).

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
```diff
@@ -1,194 +0,0 @@
...
-};
-
-struct BadExpression
-{
-    std::string expression;
-    std::string childName;
-    std::optional<std::string> parentName;
-};
-
-class BadContextComposition final: public std::invalid_argument
```
guilpier-code, line 90: **Cosmetic** : 
`struct BadExpression` was removed and exception class **BadContextComposition** was renamed into **ForbiddenNodeFound**, more expressive.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
```diff
@@ -1,194 +0,0 @@
...
-    void checkChildren(const std::string& parentName,
-                       const std::vector<Expressions::Nodes::Node*>& children,
-                       bool validateConsistencyWithParents);
-    template<Expressions::Nodes::FunctionNodeType func>
-    void checkChildren(const std::string& parentName,
-                       const std::vector<Expressions::Nodes::Node*>& children,
-                       bool validateConsistencyWithParents);
-    template<class NodeType>
-    void handleComparisonNode(const std::string& op,
-                              const std::vector<Expressions::Nodes::Node*>& children);
```
guilpier-code, line 74: **handleComparisonNode** : turned out to be unnecessary. Was removed.

src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
```diff
@@ -0,0 +1,228 @@
...
+{
+    visitChildren(divisionNode, typeIndexOf<DivisionNode>());
+}
+
+void ForbiddenNodesVisitor::visit(const EqualNode* equalNode)
+{
+    std::type_index nodeTypeId = typeIndexOf<EqualNode>();
+    checkIsForbidden(nodeTypeId, equalNode);
+    visitChildren(equalNode, nodeTypeId);
+}
```
guilpier-code, line 83: Code here is now more detailed than it used to be.
 It's clear that 2 distinct things are done : first we check if current node is forbidden, then we visit its children.

src/io/inputs/model-converter/NodeChecker.cpp
```diff
@@ -1,162 +0,0 @@
...
-}
-
-void NodeChecker::visit(const MultiplicationNode* multiplicationNode)
-{
-    checkChildren<MultiplicationNode>("multiplication", multiplicationNode->getOperands(), false);
-}
-
-void NodeChecker::visit(const DivisionNode* divisionNode)
-{
-    checkChildren<DivisionNode>("division", divisionNode->getOperands(), false);
```
guilpier-code, line 56: Here **checkChildren** takes an extra string argument (**"division"**).
It seems this information is (more or less) contained in the **DivisionNode** by calling `divisionNode->name()`.
So we used this to make this arg disappear.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodesVisitor.h
guilpier-code, line 1: This file was originally named **NodeChecker.h**

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
guilpier-code, line 1: This file is now renamed **ForbiddenNodesVisitor.h**

src/io/inputs/model-converter/NodeChecker.cpp
guilpier-code, line 1: This file was renamed into **ForbiddenNodesVisitor.cpp**

src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
guilpier-code, line 1: This file used to be named **NodeChecker.cpp**, and was renamed.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
```diff
@@ -33,154 +33,63 @@ namespace Antares::IO::Inputs::ModelConverter
...
+    // gp : we originally used unordered set and map, do we restore these types ?
```
flomnes, line 90: both are fine, since these containers store very few entries (<100)

src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
```diff
@@ -0,0 +1,224 @@
...
+    case FunctionNodeType::min:
+        return typeIndexOf<FunctionNodeType::min>();
+    case FunctionNodeType::max:
+        return typeIndexOf<FunctionNodeType::max>();
+    case FunctionNodeType::pow:
+        return typeIndexOf<FunctionNodeType::pow>();
+    case FunctionNodeType::floor:
+        return typeIndexOf<FunctionNodeType::floor>();
+    default:
+        return typeid(int); // Supposed to be dead code (never reached)
```
flomnes, line 174: This is a case of error, throw an exception here

flomnes (approved):

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
a-zakir: no they aren't the same, one has full template, the other is non template argument,

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
a-zakir: don't say unused, say you moved the logic into nodechecker visitor

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
a-zakir: no it does not do 2 things, telling if node is forbidden (globally or by parents)

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
a-zakir: i don't think the new name is more expressive

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
a-zakir: checkChildren was good

a-zakir: don't say duplication, it is 
non template argument and template arg. in fact you have noticed that


```
template<typename NodeType>
std::type_index typeIndexOf()
{
    return std::type_index(typeid(NodeType));
}

template<Expressions::Nodes::FunctionNodeType item>
std::type_index typeIndexOf()
{
    using enum_item_as_type = std::integral_constant<Expressions::Nodes::FunctionNodeType, item>;
    return std::type_index(typeid(enum_item_as_type));
}
```

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
a-zakir: check consistency with parents is clear enough for me. there's no duplication

src/io/inputs/model-converter/NodeChecker.cpp
a-zakir: there is a reason why i didn't use the name given by node. the error message is for the user who likely to understand that division can not hold dual rather that divisionnode can not hold dualnode

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
guilpier-code: From what I understand, this logic was already there in the node checker visitor (see old **NodeChecker::visit**), I changed it a bit, but it was there. About this function, I removed it and saw compilation still passed.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
guilpier-code: My first comment was not clear enough, for that I apologize.
What I meant is that **isForbiddenFor** is intends to check if a node is **forbidden by parent**, and **isForbidden** checks if a node is **globally forbidden**.
**isForbiddenFor** calls the **check(...)** function which checks if a node is both **forbidden by parents** and **globally**.
So **isForbiddenFor** does something that **isForbidden** is supposed to do.

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/NodeChecker.h
guilpier-code: As every visitor in the modeler has a name ending by **Visitor**, I did the same here, so that it's clear by the name of source file it is a visitor.
As of what it does, it searches for forbidden nodes, so I thought it was more clear for this information to be in the name of the visitor.

src/io/inputs/model-converter/NodeChecker.cpp
guilpier-code: ok. in order to make the code more simple, I made the bet that user would understand **DivisionNode** as easily as **division**.
If an extra clarity for users is required, shouldn't this new piece of information (a new id for each node) be held by the node itself ? 
Meaning : shouldn't we have another method than **name()** for each node (like **nameForUser()** or **id()** or whatever);
We could even change the returned value of **name()**. In this case, returned **"DivisionNode"** would be replaced with **"division"**

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
guilpier-code: ok, I removed the comment in [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335)

src/io/inputs/model-converter/ForbiddenNodesVisitor.cpp
guilpier-code: fixed in [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335), by following code : 
```c++
std::type_index functionNodeTypeIndex(const FunctionNode* functionNode)
{
    switch (functionNode->type())
    {
    ...
    case FunctionNodeType::ceil:
        return typeIndexOf<FunctionNodeType::ceil>();
    default:
        std::string err_msg = "ForbiddenNodesVisitor > ";
        err_msg += "function '" + functionNode->name() + "' is unknown.";
        throw std::invalid_argument(err_msg);
    }
}
```

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/ForbiddenNodes.h
```diff
@@ -33,154 +33,63 @@ namespace Antares::IO::Inputs::ModelConverter
 {
 
 template<typename NodeType>
-std::type_index forbiddenNodeKey()
+std::type_index typeIndexOf()
 {
-    static_assert(!std::is_same_v<NodeType, Expressions::Nodes::FunctionNode>,
-                  "Use Expressions::Nodes::FunctionNodeType enum values or "
-                  "forbiddenNodeKey(Expressions::Nodes::FunctionNodeType) "
-                  "instead of FunctionNode for forbidden rules.");
```
a-zakir, line 41: this was important, it forbids FunctionNode as key

guilpier-code: ok, I fixed it in [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335)

PR #3332, opened 2026-01-12
Refactor AverageData year management

payetvin:


src/solver/variable/include/antares/solver/variable/storage/averagedata.h
```diff
@@ -47,13 +47,11 @@ class AverageData
     void merge(unsigned int year, const IntermediateValues& rhs);
 
 public:
-    std::vector<HighPrecision> year;
+    std::vector<HighPrecision> year{1};
```
flomnes: Even better
```suggestion
    HighPrecision yearly;
```

flomnes (commented):
Good job !

Using a 1-size `std::vector` seems a bit overkill

flomnes (dismissed):

payetvin:
Using a highPrecision for year values cause smalls differences in std deviation for yearly results

flomnes (approved):

PR #3333, opened 2026-01-12
Fix the linear problem type for decomposed problems

a-zakir:


src/modeler/modeler/Modeler.cpp
```diff
@@ -98,6 +99,22 @@ class SystemLinearProblemBuilder final
     BendersDecomposition* bendersDecomposition_ = nullptr;
 };
 
+bool isMip(const Antares::Modeler::Data& data, Location location)
```
flomnes, line 102:
```suggestion
bool isProblemLocationMip(const Antares::Modeler::Data& data, Location location)
```

flomnes (approved):

PR #3336, opened 2026-01-12
Add custom weight to averagedata and unit tests

payetvin:


src/solver/variable/storage/averagedata.cpp
```diff
@@ -25,7 +25,15 @@
...
 namespace Antares::Solver::Variable::R::AllYears
 {
+
 AverageData::AverageData() = default;
+
+AverageData::AverageData(unsigned size):
+    yearsWeightSum(size)
+{
+    yearsWeight.assign(size, 1.);
+}
```
flomnes, line 35: 1. You can use the `yearsWeight` constructor
2. `yearsWeightSum` is not necessarily `size`, since some weights may be `!= 1`

src/tests/src/solver/variable/test_averageData.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include "antares/solver/variable/storage/averagedata.h"
+
+constexpr double TOLERANCE = 1.e-6;
+
+using namespace Antares::Solver::Variable;
+
+BOOST_AUTO_TEST_SUITE(AverageData_suite)
+
+auto all_close = [](auto val, double expected, double tol = 1e-6)
```
flomnes:
```suggestion
auto check_close = [](auto val, double expected, double tol = TOLERANCE)
```

src/solver/variable/storage/averagedata.cpp
payetvin: For sets weight will be 1, need to see how we'll handle set + years

payetvin:
Not needed

PR #3346, opened 2026-01-16
[100%- optional] Use customized mps generator in the simulator

a-zakir:


src/solver/utils/include/antares/solver/utils/mps_utils.h
```diff
@@ -75,6 +81,6 @@ class mpsWriterFactory
     // Member data...
     Data::mpsExportStatus export_mps_;
     bool export_mps_on_error_;
-    MPSolver* solver_ = nullptr;
+    const Optimisation::LinearProblemApi::ILinearProblem* linearProblem_;
```
flomnes: const ref here would be better, since `linearProblem` is already provided to the constructor

flomnes (approved):

PR #3335, opened 2026-01-12
ceil and floor operator : add missing features

guilpier-code:
This PR is the 3rd PR out of 3 PR about adding operators **ceil** and **floor** : 
- [PR 3316](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3316) (the main work on adding these operators)
- [PR 3329](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3329), its purpose is to simplify the forbidden nodes part
- [PR 3335](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3335) (the current PR)

This last PR's purpose is to add features on **ceil** and **floor** operators that were forgotten / not yet done in the current PR (especially **ceil** operator and unit tests). 

Note : 
- [x] Linearity visitor turns out to be useless, so it was removed (unit tests as well)
- [ ] what else ?

PR #3351, opened 2026-01-18
Add code, documentation & tests for the ceil operator [ANT-4295]

flomnes:
- Add a test to ensure `ReadLinearExpressionVisitor` rejects floor when its argument has non-zero coefficients (i.e. depends on variables)
- Add a test to ensure `floor(variable)` is rejected through `ForbiddenNodes`
- Update the user documentation to mention the `floor` operator
- Add missing `ceil` operator


flomnes:
See #3335

PR #3355, opened 2026-01-19
hybrid studies : bounds from modeler [ANT-4033]

guilpier-code:
This work is about **part 1** of ticket [ANT-4033](https://gopro-tickets.rte-france.com/browse/ANT-4033)
Related specifications area [here](https://gopro-collaboratif.rte-france.com/pages/viewpage.action?pageId=527087892)
Why **part 1** ? 
- **Part 1** is about **to-area-bound** field of area connections port, very recently renamed into **spillage-bound**.
So, in the code, first name is still applied.
- **Part 2** : we add a new constraint in the legacy linear problem to replace the upper bound of **unsupplied energy** variable. Indeed, having a bound defined by a constraint (and releasing the variable actual bound to infinity) allows some other optimization variable contribute limiting value of this variable.  See [part 2](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3378)
- **Part 3** :  is about taking into account **from-area-bound** field of area connections port, very recently renamed into **unsupplied-energy-bound**. See [part 3](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3382).

This 3 points will be / are treated each in a different pull request.

**Note** : 
Some renaming was done in the **VariableManager**, **VariableNamer** and the **ConstraintBuilder** : 
- **PositiveUnsuppliedEnergy** is actually the **UnsuppliedEnergy**
- **NegativeUnsuppliedEnergy** is actually the **Spillage**


src/study/system-model/portType.cpp
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    if (!(a.has_value() || b.has_value()))
+    {
+        return false;
+    }
+
+    return a->injection == b->injection && a->to_area_bound == b->to_area_bound
+           && a->from_area_bound == b->from_area_bound;
+}
```
flomnes, line 48: Note `std::optional` has operator== with this spec
```
lhs.has_value() != rhs.has_value() ? false :
    (lhs.has_value() == false ? true : *lhs == *rhs)
```
see https://en.cppreference.com/w/cpp/utility/optional/operator_cmp.html

guilpier-code: I tried this, but it causes a crash (on instruction `*lhs == *rhs`). I'll try harder at some point.

src/io/inputs/yml-model/include/antares/io/inputs/yml-model/Library.h
```diff
@@ -98,15 +98,19 @@ struct Model
...
+
 struct PortType
 {
     std::string id;
     std::string description;
-    // Small optimization: we only need the name of the fields
-    // No need for an intermediate struct "field" with just a string "name" member
     std::vector<std::string> fields;
-    // Also, we only need the injection-field of the area-connection
-    std::string area_connection_injection_field;
```
guilpier-code, line 109: Here `std::string area_connection_injection_field;` is replaced 
with a `AreaConnection area_connection;`

src/io/inputs/yml-model/decoders.hxx
```diff
@@ -223,15 +237,13 @@ struct convert<Antares::IO::Inputs::YmlModel::PortType>
...
-                // Must have exactly one area connection field definition
                 return false;
             }
-            for (const auto& field: node["area-connection"])
-            {
-                rhs.area_connection_injection_field = field["injection-field"].as<std::string>("");
-            }
+            rhs.area_connection.injection = findValue("injection-field", node);
+            rhs.area_connection.to_area_bound = findValue("to-area-bound-field", node);
+            rhs.area_connection.from_area_bound = findValue("from-area-bound-field", node);
```
guilpier-code: Now we read 3 fields

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
```diff
@@ -35,33 +37,53 @@ class ComponentToAreaConnectionFiller final
     explicit ComponentToAreaConnectionFiller(
       const PROBLEME_HEBDO* problemeHebdo,
       Optimisation::OptimEntityContainer& variableContainer,
-      const Optimisation::LinearProblemApi::ILinearProblemData& linearProblemData,
```
guilpier-code, line 38: **ILinearProblemData** not required in this class. Was removed.

src/solver/optimisation/include/antares/solver/optimisation/ComponentToAreaConnectionFiller.h
guilpier-code, line 1: Some refactoring was made here, to try to avoid duplicate code.
It works like this, but there is still duplicate code. 
For example, member functions **addInjectionPortToLinearProblem** and **addToAreaBoundPortInLinearProblem** are siblings.

src/solver/optimisation/variables/VariableManagement.h
```diff
@@ -16,6 +16,8 @@ class VariableManager final
                     std::vector<std::vector<int>>& NumeroDeVariableDeTrancheDeStock,
                     const int32_t& NombreDePasDeTempsPourUneOptimisation);
 
+    VariableManager(PROBLEME_HEBDO* pb_hebdo);
```
guilpier-code: Here's a new constructor for **VariableManager**.
It's still unused for now but aims at replacing the content of **VariableManagerUtils.h/.cpp** which seems overkill.
This constructor should be used instead.

src/study/system-model/include/antares/study/system-model/portType.h
```diff
@@ -3,64 +3,37 @@
...
+    std::string to_area_bound;
+    std::string from_area_bound;
+};
 
 class PortType final
 {
 public:
     PortType(const std::string& id,
              std::vector<PortField>&& fields,
-             const std::string& areaConnectionFieldId = ""):
```
guilpier-code, line 22: In this header, many definitions were moved (and refactored a bit) to an associated cpp file

src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -344,6 +343,16 @@ EvaluationResult::EvaluationResult(const std::variant<double, std::vector<double
...
 
+std::vector<double> EvaluationResult::asVector(const unsigned& size) const
+{
+    if (std::holds_alternative<double>(value_))
+    {
+        return std::get<std::vector<double>>(value_);
+    }
+    double value = std::get<double>(value_);
+    return std::vector(size, value);
+}
```
a-zakir: a lot of copies is done here

a-zakir: seems to be unused

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -139,33 +139,38 @@ static ForbiddenNodes ForbiddenInExtraOutput()
...
         std::vector<PortField> fields;
-        for (const auto& field: portType.fields)
+        for (const auto& field: ymlPortType.fields)
         {
             fields.emplace_back(field);
         }
 
         // Can't have port types with the same ID
-        if (std::ranges::find_if(out, [&portType](const auto& p) { return p.Id() == portType.id; })
+        if (std::ranges::find_if(out, [&](const auto& p) { return p.Id() == ymlPortType.id; })
```
a-zakir: just capture ymlPortType as it was

src/tests/src/io/yml-importers/test-system-with-connexions.cpp
```diff
@@ -13,6 +13,92 @@ using namespace std::string_literals;
...
+    - id: power_flow
+      description: power flow
+      fields:
+        - id: flow
+        - id: to-area-bound
+        - id: from-area-bound
+      area-connection:
+        - injection-field: flow
+        - to-area-bound-field: to-area-bound
+        - from-area-bound-field: from-area-bound
```
a-zakir: area-connection is a single coherent object i would have expected this structure
```suggestion
      area-connection:
        injection-field: flow
        to-area-bound-field: to-area-bound
        from-area-bound-field: from-area-bound
```

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -3,127 +3,192 @@
...
+    return component.getModel()->Ports().at(portId).Type().areaConnection()->to_area_bound;
+}
+
+std::map<std::string, unsigned> associateIndicesToAreas(const PROBLEME_HEBDO* problemeHebdo_)
+{
+    std::map<std::string, unsigned> areaIndices;
+    unsigned index = 0;
+    for (auto name: problemeHebdo_->NomsDesPays)
+    {
+        areaIndices.insert({name, index++});
```
a-zakir: try_emplace

src/solver/optimisation/variables/VariableManagement.h
a-zakir: overkill ? no; the goal is to not have PROBLEM_HEBDO in this file

a-zakir:
```suggestion

```

src/study/system-model/portType.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+const std::vector<PortField>& PortType::Fields() const
+{
+    return fields_;
+}
+
+const std::optional<AreaConnection>& PortType::areaConnection() const
+{
+    return areaConnection_;
+}
```
a-zakir, line 81: AreaConnection is never an optional; it never takes nullopt

src/expressions/visitors/EvalVisitor.cpp
guilpier-code: Good eyes : not required indeed. 
In the course of this ticket, specifications changed.
Before this change, this function was useful, but not anymore.

src/study/system-model/portType.cpp
guilpier-code: **std::nullopt_t** is the default value for an `std::optional`.
In class **PortType's** constructor, **areaConnection_** can remain untouched, and so remains **nullopt**

a-zakir: Return statement in ctor is uncommun.

a-zakir:
The filler part of the pr is fine, i approve the Pr

a-zakir (approved):

src/study/system-model/portType.cpp
guilpier-code: no problem, I changed it : no more **return** in the ctor

src/io/inputs/model-converter/modelConverter.cpp
guilpier-code: done

src/tests/src/io/yml-importers/test-system-with-connexions.cpp
guilpier-code: Good point.
Changing **area-connections** from a **list** to a **map** (in yaml world of course) means also changing **hybrid studies**.
Anyway, I tried what you asked in [this PR](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3381).
This PR's CI is expected to fail because of **cucmber hybrid studies**, not updated yet.

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
guilpier-code: Done

src/solver/optimisation/variables/VariableManagement.h
guilpier-code: ok, the intention was to make **VariableManager** independent from the problem hebdo, right ?
It is a good intention.
But actually, the **VariableManager** remains closely related to the **problem hebdo**, because it holds references to **problem hebdo** data members. So having a factory to create the **variable manager** or not does not seem to change this dependency very much, does it ?

guilpier-code: Related to [this discussion](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3355#discussion_r2746681436)

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -3,127 +3,191 @@
...
+    for (auto h(0); h <= ctx.getLocalLastTimeStep(); ++h)
     {
-        IMipConstraint* areaBalanceConstraint = getBalanceConstraint(pb, lowerAreaId, localIndex);
-
-        for (const auto& [index, coef]: linearExpression[localIndex])
+        IMipConstraint* constraint = constraints[h];
+        for (const auto& [index, coef]: linearExpression[h])
         {
-            areaBalanceConstraint->setCoefficient(solverVariables.at(index).get(), -coef);
+            constraint->setCoefficient(solverVariables.at(index).get(), -coef);
```
tbittar, line 114: Big warning here !! As noted in the comment of the previous, we heavily rely on the convention that in the legacy balance constraint, generations are < 0 and loads are > 0. As before there was just this constraint to modify with the AreaConnection there was no problem. But have you checked that for the new constraints involving AreaBounds, this is the correct way to set the coefficients (i.e is the sign convention the same in the legacy spillage constraint than in the balance one) ?

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -3,127 +3,191 @@
...
         }
 
-        double constant = linearExpression[localIndex].constant();
-        areaBalanceConstraint->setBounds(areaBalanceConstraint->getLb() + constant,
-                                         areaBalanceConstraint->getUb() + constant);
+        double c = linearExpression[h].constant();
+        constraint->setBounds(constraint->getLb() + c, constraint->getUb() + c);
```
tbittar, line 118: Moreover as we use `-coef`, why not `-c` here (the orignal code may be wrong as well) ? I think this use case has never been tested and it should be...

src/tests/src/solver/optimisation/test_area_connections--spillage--unsupE.cpp
```diff
@@ -0,0 +1,211 @@
...
+    filler.addConstraints(fill_ctx);
+
+    // ---------
+    // Checks
+    // ---------
+    const auto* fictive_ct_t0 = linearProblem.lookupConstraint(
+      "FictiveLoads::area<area1>::hour<0>");
+    const auto* var1_t0 = linearProblem.lookupVariable("component_with_vars.var_1_t0");
+
+    BOOST_CHECK_EQUAL(fictive_ct_t0->getCoefficient(var1_t0), -2.);
```
tbittar: Add a check on the value of the constant ?

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
guilpier-code: I checked this, and documentation was added to [PR for part 3](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3382)

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
guilpier-code: Yes we're sure, see documentation added to [PR for part 3](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3382)

src/tests/src/solver/optimisation/test_area_connections--spillage--unsupE.cpp
guilpier-code: Done. See [PR for part 3](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3382) and particularly [this commit](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3382/changes/b56ebe54ba4890e12e63bbe3f6a90ac1aafca39d)

src/solver/optimisation/constraints/MaxUnsuppliedEnergy.cpp
```diff
@@ -0,0 +1,19 @@
+#include "antares/solver/optimisation/constraints/MaxUnsuppliedEnergy.h"
+
+void MaxUnsuppliedEnergy::add(int pdt, int pays)
```
a-zakir: We should avoid adding new constraints with old fashion.

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -248,13 +248,17 @@ void VariableNamer::FinalStorage(unsigned varIndex)
...
+    // TODO : what is "PositiveUnsuppliedEnergy" ?
+    // We mean "UnsuppliedEnergy". To be replaced.
     SetAreaElementNameHour(varIndex, "PositiveUnsuppliedEnergy");
 }
 
-void VariableNamer::NegativeUnsuppliedEnergy(unsigned varIndex)
+void VariableNamer::Spillage(unsigned varIndex)
 {
+    // TODO : what is "NegativeUnsuppliedEnergy" ?
+    // Should be replaced with "Spillage".
```
flomnes, line 261: Remove comment

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -248,13 +248,17 @@ void VariableNamer::FinalStorage(unsigned varIndex)
     SetAreaElementNameHour(varIndex, "FinalStorage");
 }
 
-void VariableNamer::PositiveUnsuppliedEnergy(unsigned varIndex)
+void VariableNamer::UnsuppliedEnergy(unsigned varIndex)
 {
+    // TODO : what is "PositiveUnsuppliedEnergy" ?
+    // We mean "UnsuppliedEnergy". To be replaced.
```
flomnes, line 254: Remove comment

src/study/system-model/include/antares/study/system-model/component.h
```diff
@@ -93,7 +93,7 @@ class Component final
     const Expression& expressionAtPortField(const std::string& portId,
                                             const std::string& fieldId) const;
 
-    void addAreaConnection(const std::string& localPortId, const std::string& areaId);
+    void addAreaConnection(const std::string& localPortId, std::string areaId);
```
flomnes: const ref

flomnes (changes requested):
Good job
Remove `VariableManager(PROBLEME_HEBDO*)`

Some unnecessary comments

src/solver/optimisation/opt_rename_problem.cpp
guilpier-code: Please explain why this comment should be removed, sorry I'm confused :
it seems the renaming should take a step further and be applied to variable complete names in **mps** files, isn't it ?
This would make mps files lines a bit shorter, and this change does not seem to have an impact on users. 
On **Xpansion** ?

src/study/system-model/include/antares/study/system-model/component.h
guilpier-code: Done

src/solver/optimisation/ComponentToAreaConnectionFiller.cpp
```diff
@@ -3,127 +3,247 @@
...
+
+std::string componentUnsupEnergyBound(const Component& component, const std::string& portId)
+{
+    return component.getModel()
+      ->Ports()
+      .at(portId)
+      .Type()
+      .areaConnection()
+      ->unsupplied_energy_bound;
+}
```
flomnes: Expression `x = component.getModel()->Ports().at(portId).Type().areaConnection()`, is used 3 times.

Make a private method that returns & checks this value is valid before calling `x->member_value`

guilpier-code: To avoid code duplication, I made a new public method for class **Component** : 
```c++
const std::optional<AreaConnection>& Component::areaConnectionAtPort(
  const std::string& portId) const
{
    if (!getModel()->Ports().contains(portId))
    {
        std::string errMsg = "Port '" + portId + "' not found in component '" + data_.id + "'"; 
        throw std::invalid_argument(errMsg);
    }
    return getModel()->Ports().at(portId).Type().areaConnection();
}
```
And, in **ComponentToAreaConnectionFiller.cpp**, called it instead of the free functions **componentInjectionField**, **componentSpillageBound** and **componentUnsupEnergyBound**, that were removed.

About validity, for example let's look at 
```c++
void ComponentToAreaConnectionFiller::addInjectionPortToLinearPb(const FillContext& ctx,
                                                                 const Component& component,
                                                                 const std::string& portId,
                                                                 const unsigned& areaIndex)
{
    std::string portField = component.areaConnectionAtPort(portId)->injection;
    if (portField.empty())
    {
        // area connection does not specify this port field
        return;
    }

    // 1. GEMS side : get time-dependent linear expression at a component port field
    auto linearExpression = linearExpressionAtPortField(portId, portField, component, ctx);
   ...
}
```
- call to `component.areaConnectionAtPort(portId)` checks that the component has a area connection to **portId** (otherwise, exception thrown)
- look if `component.areaConnectionAtPort(portId)->injection` is used (in port type definition). If not, nothing to be done, exit.
- when calling linearExpressionAtPortField, a check is made to see if there is an expression / node at `component.areaConnectionAtPort(portId)->injection`.

So  guess validity checking is enough.

flomnes (approved):

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -139,33 +139,38 @@ static ForbiddenNodes ForbiddenInExtraOutput()
...
         {
-            throw PortTypeWithThisIdAlreadyExists(portType.id);
+            throw PortTypeWithThisIdAlreadyExists(ymlPortType.id);
         }
 
-        PortType portTypeModel(portType.id,
-                               std::move(fields),
-                               portType.area_connection_injection_field);
-        out.emplace_back(std::move(portTypeModel));
+        PortType portType(ymlPortType.id,
```
guilpier-code, line 170: Shouldn't we check that every key of **area-connection** section of a part type (**injection**, **spillage_bound** or **unsupplied_energy_bound**) has a value that correspond to a port id in section **fields** of the same port type ?
This check should be done as soon as possible in the code execution. 
**convertPortTypes** function is a good place.

guilpier-code: already done

PR #3365, opened 2026-01-25
Use MPSolver for the "daily" hydro heuristic

flomnes:


flomnes:
Hey @flomnes, is this still relevant?

PR #3369, opened 2026-01-26
[ANT-4342 - 100%] Thermal capacity Investment

a-zakir:


src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,196 @@
...
+          component.Id()));
+    }
+    return field.value();
+}
+
+void ThermalCapacityFiller::addConstraints(const FillContext& ctx)
+{
+    for (const auto& component: modelerSystem_->Components())
+    {
+        for (const auto& [portId, thermalConnection]: component.portToThermalConnections())
```
tbittar: As there might be other thermal hybrid connections in the future (for binding constraints for instance)
```suggestion
        for (const auto& [portId, thermalCapacityConnection]: component.portToThermalCapacityConnections())
```

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -164,7 +164,8 @@ std::vector<PortType> convertTypes(const ::YmlModel::Library& library)
 
         PortType portTypeModel(portType.id,
                                std::move(fields),
-                               portType.area_connection_injection_field);
+                               portType.area_connection_injection_field,
+                               portType.thermal_capacity_connection_field);
```
guilpier-code: **Question** : 
In the yaml file (in specs), the **field** seems not required / redundant.
Moreover, this new functionality is about investment, and there's no name mentioning that.
So my question is : instead of having :  
```yaml
 port-types:
    - id: capacity_port
      fields:
        - id: capacity
      thermal-capacity-connection: 
        - capacity-field: capacity
```
Couldn't we have something like : 
```yaml
 port-types:
    - id: thermal_capacity_port_type
      fields:
        - id: thermal_capacity_port
      thermal-invest-connection: 
        - capacity: thermal_capacity_port
```

src/io/inputs/yml-model/decoders.hxx
```diff
@@ -209,6 +209,41 @@ struct convert<Antares::IO::Inputs::YmlModel::Model>
 template<>
 struct convert<Antares::IO::Inputs::YmlModel::PortType>
 {
+    static bool processAreaInjectionField(const Node& node,
```
guilpier-code: Here **area-connection** and **thermal-capacity-connection** are still yaml collections, and should be a dictionaries (as you rightfully suggested somewhere else)

src/io/inputs/yml-model/decoders.hxx
```diff
@@ -209,6 +209,41 @@ struct convert<Antares::IO::Inputs::YmlModel::Model>
...
+                                          Antares::IO::Inputs::YmlModel::PortType& rhs)
+    {
+        if (node["area-connection"].IsDefined())
```
guilpier-code: reading these 2 sections (of port-type) is very similar and lead to suspect code duplication. 
Couldn't we find a way to remove duplication (if any) ?

src/io/inputs/yml-model/decoders.hxx
guilpier-code: Renaming : the 2 functions start by **process**, when we actually **read** or **convert**.

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -164,7 +164,8 @@ std::vector<PortType> convertTypes(const ::YmlModel::Library& library)
 
         PortType portTypeModel(portType.id,
                                std::move(fields),
-                               portType.area_connection_injection_field);
+                               portType.area_connection_injection_field,
```
guilpier-code: **Question** (draft, I'll come back to this question after more review) :
It's more a question on specifications.
As we already have an **area-connection** section both in **my-library.yml** (portType) and in **system.yml**,  why don't we use it ? Is connection for investing on thermal capacity so different from other area connections ? 
What would be wrong with : 
```yaml 
port-types:
    - id: area-connection-port-type
      fields:
        - id: capacity
        - id: to-balance
        - id: to-area-bound
        - id: from-area-bound
        - id: to-thermal-capacity
      area-connection: 
        - thermal-capacity: to-thermal-capacity
        - injection-field: field_to_balance
        - ...
```
and in system.yml : 

```yaml
area-connections:
    - component: my_load
      port: area-connection-port 
      area: area1
    - thermal-capacity-for-investment:
      port: area-connection-port
      area: fr
      cluster-id: nuclear1
```

src/io/inputs/yml-system/converter.cpp
```diff
@@ -119,18 +119,6 @@ static Component& findComponent(const std::string& id, std::vector<Component>& c
     return *it;
 }
 
-static const Port& findPort(const Component& component, const std::string& portId)
```
guilpier-code, line 122: ok, **findPort** was moved to **Component.cpp**.
We could even make it a method of class **Component**  : 
- this function is only about a component
- it would be easier to use (an argument disappears), especially when calling it from class Component itself (**\*this** disappears as an argument). 

Last argument could be a default one (as it is sometimes empty).

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,197 @@
...
+// capacity = availability_factor * invested_capacity + already_installed_availability_factor *
+// already_installed_capacity
+// TODO
+void ThermalCapacityFiller::processThermalCapacityField(
+  const TimeDependentLinearExpression& linearExpression,
+  const ModelerStudy::SystemModel::Component::ThermalConnection& thermalConnection,
+  const FillContext& ctx)
+{
+    auto areaId = thermalConnection.areaId;
+    boost::algorithm::to_lower(areaId);
```
guilpier-code: Turning area id into lower case should be done in class **Component**, at the end of **Component::addThermalCapacityConnection** method.

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,197 @@
...
+      .NumeroDuPalierDansLEnsembleDesPaliersThermiques[clusterLocalIndex];
+}
+
+// set up DispatchableProduction up to max and add
+// DispatchableProduction[t] <= component.port.capacity-field (qui vaut
+// my_thermal_invest.capacity_port.
+// capacity = availability_factor * invested_capacity + already_installed_availability_factor *
+// already_installed_capacity
+// TODO
+void ThermalCapacityFiller::processThermalCapacityField(
```
guilpier-code, line 154: We could make this **cpp** file more readable by removing too long namespace heads.

src/study/system-model/include/antares/study/system-model/component.h
```diff
@@ -95,10 +95,24 @@ class Component final
...
+                                      const std::string& clusterId);
     std::optional<std::string> areaConnectedToPort(const std::string& portId) const;
 
     const std::map<std::string, std::string>& portToAreaConnections() const;
 
+    struct ThermalConnection
+    {
+        std::string areaId;
+        std::string clusterId;
+    };
```
guilpier-code: Maybe move `struct ThermalConnection` out of class **Component** (above ?), so that : 
- when you need it elsewhere (like in class **ThermalCapacityFiller**), we wouldn't have to write **Component::ThermalConnection** but simply **ThermalConnection**
-  to clear a bit class **Component**'s interface

src/study/system-model/include/antares/study/system-model/component.h
```diff
@@ -95,10 +95,24 @@ class Component final
...
+
+    const std::map<std::string, ThermalConnection>& portToThermalConnections() const;
+
+    std::optional<::Antares::ModelerStudy::SystemModel::Component::ThermalConnection>
+    thermalCapacityConnectedToPort(const std::string& portId) const;
```
guilpier-code: Instead of : 
```c++
 std::optional<::Antares::ModelerStudy::SystemModel::Component::ThermalConnection>
    thermalCapacityConnectedToPort(const std::string& portId) const;
```
We could have : 
```c++
 std::optional<ThermalConnection> thermalCapacityConnectedToPort(const std::string& portId) const;
```
more clear (**ThermalConnection** being in the same namespace scope as class **Component**).

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -126,6 +127,8 @@ void fillLinearProblem(FillContext& fillCtx,
           optimEntityContainer,
           *problemeHebdo->modelerData->dataSeries,
           problemeHebdo->modelerData->scenarioGroupRepository));
+        fillersCollection.push_back(
+          std::make_unique<ThermalCapacityFiller>(problemeHebdo, optimEntityContainer));
```
guilpier-code, line 131: Discussion :
Note that in **ComponentToAreaConnectionFiller**  constructor, arguments 3 and 4 are never used.
So argument list of constructors of classes **ComponentToAreaConnectionFiller** and **ThermalCapacityFiller**, are identical.
And fundamentally, do these 2 classes do something different ?
They both modify the linear problem (variables bounds or constraints).
So if we wanted, we could modify the LP in the **thermal capacity connection** way inside the **ComponentToAreaConnectionFiller** (and remove the **ThermalCapacityFiller**).
Problem is : do we want this ?
If **area-connections** and **thermal-capacity-connections** are, and should  stay separated in **yaml** files (as @tbittar says), shouldn't they remain separated when filling the LP as well ?
**But we may have duplicate code if code stays as it is**

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,197 @@
...
+    {
+        for (const auto& [portId, thermalConnection]: component.portToThermalConnections())
+        {
+            addComponentPortContributionToThermalCapacity(ctx,
+                                                          component,
+                                                          portId,
+                                                          thermalConnection);
+        }
+    }
+}
```
guilpier-code, line 76: I would not hesitate to do : 

```c++
void ThermalCapacityFiller::addConstraints(const FillContext& ctx)
{
    for (const auto& component: modelerSystem_->Components())
    {
        for (const auto& [portId, thermalConnection]: component.portToThermalConnections())
        {
            std::string& areaId = thermalConnection.areaId;
            std::string& clusterId = thermalConnection.clusterId;
            addComponentPortToLinearPb(ctx, component, portId, areaId, clusterId);
        }
    }
}
```
because it seems more clear what we pass and, in the lower level, code will get simpler.

Moreover, as we are in the context of **Thermal capacity connection**, we don't need to repeat it in the function name, so we can shorten a bit this name, from **addComponentPortContributionToThermalCapacity** into **addComponentPortToLinearPb**

src/solver/optimisation/ThermalCapacityFiller.cpp
guilpier-code, line 154: Maybe renaming suggestion : 
**processThermalCapacityField** may not be a meaningful name (because of **process**).
It seems this function does 2 things.
So maybe we could have (for example) : 
```c++
removeThermalProdsUpBound(ctx, areaId, clusterId);
limitingThermalProdsWithConstraint(ctx, linearExpression, areaId, clusterId)
```

src/tests/cucumber/features/steps/common_steps/solver_steps.py
```diff
@@ -13,9 +13,12 @@
 from common_steps.solver_input_handler import solver_input_handler
 from common_steps.solver_output_handler import solver_output_handler
 
-from features.steps.common_steps.assertions import assert_double_close
-from features.steps.common_steps.modeler_output_handler import modeler_output_handler
-from features.steps.common_steps.table_compare import *
+from common_steps.assertions import assert_double_close
```
guilpier-code, line 15: If you add 
```py
from common_steps.assertions import assert_double_close
```
then you should remove : 
```py
from common_steps.assertions import *
```

src/tests/cucumber/features/solver-features/hybrid_studies.feature
```diff
@@ -94,3 +94,26 @@ Feature: hybrid (simulator+modeler) studies
     When I run antares simulator on all studies
     Then all simulations succeed
     And for each time step, all studies have the same objective value
+
+
+  @fast @short
+  Scenario: Thermal capacity connection: on one node (= area, thermal cluster = continuous_generator_candidate), one timestep and one scenario with one candidate.
+
+    Given the solver study path is "Antares_Simulator_Tests_NR/hybrid/13_1_thermal_add_on/"
```
guilpier-code, line 102: Study **Antares_Simulator_Tests_NR/hybrid/13_1_thermal_add_on** has 1 year of 365 days. 
Is it required that the study has so many weeks ?

src/tests/cucumber/features/steps/common_steps/parse_linear_expression.py
```diff
@@ -0,0 +1,62 @@
+import sympy as sp
+import re
+
+
+def parse_linear_expression(expression: str):
```
a-zakir, line 5: /!\  this will be removed, the test can hard code expected coeffs and rhs

a-zakir: #TODO

src/io/inputs/yml-model/decoders.hxx
a-zakir: renamed convert*

src/io/inputs/yml-model/decoders.hxx
a-zakir: it is possible. done

src/io/inputs/yml-system/converter.cpp
a-zakir: converted into component method, default arg shadows behavior

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: does not change anything, but why not !

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
a-zakir: _And fundamentally, do these 2 classes do something different ?_ Yes one modifies balance constraints, the other add new constraint with polymorphic terms and rhs to connect component and thermal cluster. This separate matters.
_They both modify the linear problem_ if we follow this assumption all modifications of LP should be in one big giant function as it was few years ago...

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: it is kind of primitive obsession.

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: you like one line function ?
```
void removeThermalProdsUpBound(auto*dispatchableProduction){
dispatchableProduction->setUb(infinity);
}
```

src/tests/cucumber/features/solver-features/hybrid_studies.feature
a-zakir: 1 week is suffcient to prove the constraint existence, 2 weeks to prove that it expand to more than one week etc...

a-zakir: i'll ask to reduce the simulated time laps

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "antares/solver/optimisation/variables/VariableManagerUtils.h"
+#include "antares/solver/simulation/sim_structure_probleme_economique.h"
+
+using namespace Antares::Optimisation;
+using namespace Antares::Optimisation::LinearProblemApi;
+using namespace Antares::ModelerStudy::SystemModel;
+
+namespace Antares::Optimization
+{
+ThermalCapacityFiller::ThermalCapacityFiller(PROBLEME_HEBDO* problemeHebdo,
```
guilpier-code, line 18: Several times in this class we use `optimEntityContainer_.Problem()`, even when we call : 
```c++
optimEntityContainer_.getVariables()
```
it's done behind the scene.
So maybe we could have a private data `ILinearProblem& linearProblem_;` set at construction of class **ThermalCapacityFiller** to `optimEntityContainer_.Problem()`

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,193 @@
...
+                                             OptimEntityContainer& optimEntityContainer):
+    problemeHebdo_(problemeHebdo),
+    modelerSystem_(problemeHebdo->modelerData->system.get()),
+    optimEntityContainer_(optimEntityContainer),
+    variableManager_(VariableManagerFromProblemHebdo(problemeHebdo))
+{
+    unsigned int i = 0;
+    for (auto name: problemeHebdo_->NomsDesPays)
```
guilpier-code: We should avoid using french in new code : 
- `const auto& nomPalier`
- `const auto& palierThermiques`

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,193 @@
...
+namespace Antares::Optimization
+{
+ThermalCapacityFiller::ThermalCapacityFiller(PROBLEME_HEBDO* problemeHebdo,
+                                             OptimEntityContainer& optimEntityContainer):
+    problemeHebdo_(problemeHebdo),
+    modelerSystem_(problemeHebdo->modelerData->system.get()),
+    optimEntityContainer_(optimEntityContainer),
+    variableManager_(VariableManagerFromProblemHebdo(problemeHebdo))
+{
+    unsigned int i = 0;
```
guilpier-code: local indices **i** and **j** : they could have a better name.

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,193 @@
...
+        throw Error::RuntimeError(
+          fmt::format(" area '{}' has not thermal cluster by the name '{}' ", areaId, clusterId));
+    }
+    return itv->second;
+}
+
+int ThermalCapacityFiller::getClusterIndex(const std::string& areaId, const std::string& clusterId)
+{
+    //--- area
+    const auto& [pays, clusters] = *areaClusters(areaId);
```
guilpier-code:
```c++
const auto& [pays, clusters] = *areaClusters(areaId);
```
in this instruction, **areaId** and **pays** are identical. 
So one of them seems useless.

And due to the function name, I would expect a list of clusters

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,193 @@
...
+          fmt::format("unknown area '{}', is found in thermal-connection-capacity ", areaId));
+    }
+    return &it->second;
+}
+
+unsigned int getClusterLocalIndex(const std::string& areaId,
+                                  const std::string& clusterId,
+                                  const std::unordered_map<std::string, unsigned>& clusters)
+{
+    if (clusters.empty())
```
guilpier-code, line 135: Case `clusters.empty()` is impossible, by construction (see class constructor).

src/solver/optimisation/ThermalCapacityFiller.cpp
```diff
@@ -0,0 +1,193 @@
...
+    optimEntityContainer_(optimEntityContainer),
+    variableManager_(VariableManagerFromProblemHebdo(problemeHebdo))
+{
+    unsigned int i = 0;
+    for (auto name: problemeHebdo_->NomsDesPays)
+    {
+        unsigned int j = 0;
+        const auto& palierThermiques = problemeHebdo_->PaliersThermiquesDuPays[i];
+        std::unordered_map<std::string, unsigned int> clusters;
+        for (const auto& nomPalier: palierThermiques.NomsDesPaliersThermiques)
```
guilpier-code: If what we eventually need is a global number for a thermal cluster, should we continue storing local indices ?
Or should we store cluster global indices directly ?

src/solver/optimisation/ThermalCapacityFiller.cpp
guilpier-code: In the end, what is done in the constructor's body is selecting and storing information held in the problem hebdo.
And what we need is being able to give an area id and and a cluster id in input, and get a global number (for the cluster).
Today, all this work is done in class **ThermalCapacityFiller**.
But it could be encapsulated in a new class (whatever the name).
This class could be instantiated in the **ThermalCapacityFiller**'s constructor as a private data member.
And one of its method (**clusterGlobalNumber(areaId, clusterId)**) could be called at some point.

src/solver/optimisation/include/antares/solver/optimisation/ThermalCapacityFiller.h
```diff
@@ -0,0 +1,76 @@
...
+                                                                                    unsigned pdt);
+    void addCapacityFieldConstraint(const TimeDependentLinearExpression& linearExpression,
+                                    const Optimisation::LinearProblemApi::FillContext& ctx,
+                                    int clusterIndex,
+                                    const std::string& namePrefix);
+
+    struct AreaAndClusters
+    {
+        unsigned int areaIndex;
+        std::unordered_map<std::string, unsigned> clusters;
```
guilpier-code: `std::unordered_map<std::string, unsigned> clusters;` : Is a unordered map required here ? 
Would a simple std::map be enough ?

src/io/inputs/yml-model/decoders.hxx
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <optional>
```
guilpier-code, line 7: It seems we don't need **optional** here

src/tests/src/solver/optimisation/FillerFixture.cpp
```diff
@@ -0,0 +1,160 @@
...
+// this should be called before setUpModelerVariables
+void FillerFixture::addLegacyVariables(const std::vector<std::string>& variableNames)
+{
+    for (const auto& variable: variableNames)
+    {
+        linearProblem.addVariable(-999, 999, false, variable);
+    }
+}
+
+void FillerFixture::setUpModelerVariables(unsigned int ts_start,
```
guilpier-code: `void FillerFixture::setUpModelerVariables(...)` : 
This functions adds variables to the LP.

We have a `void FillerFixture::addLegacyVariables` function, so why not have `void FillerFixture::addModelerVariables(...)`  (instead of setUp) ?

a-zakir: i just finished to resolve conflicts, wait please

src/solver/optimisation/include/antares/solver/optimisation/ThermalCapacityFiller.h
a-zakir: this is not an arg, unordered_map is faster than map. the question is why you prefer map ?

src/io/inputs/yml-model/decoders.hxx
a-zakir: deleted

src/solver/optimisation/include/antares/solver/optimisation/ThermalCapacityFiller.h
guilpier-code: ok, I can't a reason in this situation.

src/tests/src/solver/optimisation/FillerFixture.cpp
guilpier-code: I have no problem with waiting

guilpier-code: Now that you've merged from develop, there are regressions on CI, which is expected.

a-zakir: renamed add*

src/tests/cucumber/features/solver-features/hybrid_studies.feature
a-zakir: 13* hybrid studies have been reduced to 3 weeks

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: english Imperialism

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: renamed

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: renamed

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: this review is becoming interesting..

a-zakir: ~Although the constructor prevents empty clusters, the user might select an existing area that contains none. This check ensures that this scenario is handled and, if it occurs, provides a helpful error message.~

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: it is possible to contain local indices in the ctor

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: yes, in upcoming Pr.

src/solver/optimisation/ThermalCapacityFiller.cpp
a-zakir: i think this case can be handled  in areaClusters with certainly less precise message ""unknown area '{}', is found in thermal-connection-capacity or the area has not thermal clusters."

guilpier-code (approved):

PR #3377, opened 2026-01-29
fix: remove unnecessary global time step adjustment

JasonMarechal25:
## Explanation
The local -> global conversion is already done here, no need to convert twice. Otherwise the time labels are wrong.
```cpp
TimeBlock convertBlockTimeStepToAbsoluteTimeStep(unsigned int timeStep,
                                                 const TimeConversionMode& mode,
                                                 const unsigned currentBlock)
{
    switch (mode)
    {
    case TimeConversionMode::WeeklyBlocks:
        return {.block = currentBlock + 1,
                .blockTimeIndex = timeStep + 1, // here
                .absoluteTimeIndex = currentBlock * 168 + timeStep + 1}; // here
    case TimeConversionMode::DailyBlocks:
        return {.block = currentBlock + 1,
                .blockTimeIndex = timeStep + 1, // here
                .absoluteTimeIndex = currentBlock * 24 + timeStep + 1}; // here
    case TimeConversionMode::SingleBlock:
    default:
        return {.block = 1, .blockTimeIndex = timeStep + 1, .absoluteTimeIndex = timeStep + 1}; // here
    }
}
```


src/io/outputs/SimulationTableGenerator.cpp
```diff
@@ -92,7 +92,7 @@ void addVariableEntries(ISimulationTable& simulationTable,
         auto handle = [&](std::optional<unsigned> timeStep, std::optional<unsigned> scenIdx)
         {
             TimeBlock tb = timeStep ? convertBlockTimeStepToAbsoluteTimeStep(
```
a-zakir: are sure about this ?

JasonMarechal25: No.
It works on the exemple I used but I don't yet know about all impacts

flomnes (approved):

PR #3393, opened 2026-02-06
fix: properly handle modeler error

JasonMarechal25:
In some cases modeler error where ignored. It was the case for yml convertion error. In this case error were caught in load.cpp but only logs.infos were triggered.

We choose to trigger logs.error instead to abort the simulation and display error to user.

A side effect is that in the case of legacy study without model library we now trigger error which we don't want. 
In the absence of model-library a warning is triggered instead of bubbling up an exception.

Implementation note: we use std::optional. With c++23 we would have used the optional::transform pattern.


src/modeler/loadFiles/readLibraries.cpp
```diff
@@ -164,13 +169,20 @@ Solver::ResolutionMode convertResolutionMode(std::string ymlMode)
...
+    if (ymlLibraries.has_value())
+    {
+        const auto ymlOptimConfig = loadOptimConfigFromYaml(studyPath);
+        updateLibrariesWithOptimConfig(ymlLibraries.value(), ymlOptimConfig);
+        // Avoid ambiguous brace-init by explicitly creating the optional
+        // using std::make_optional/std::make_pair.
+        return std::make_optional(
+          std::make_pair(convertIntoSystemLibs(ymlLibraries.value()),
+                         convertResolutionMode(ymlOptimConfig.resolution_mode)));
+    }
```
flomnes: `std::make_pair` is probably not needed here

```cpp
template< class T, class... Args >
constexpr std::optional<T> make_optional( Args&&... args );
```

https://en.cppreference.com/w/cpp/utility/optional/make_optional.html

flomnes (commented):
I'm not a fan of mixing exceptions and `std::optional` for error handling. For me it's either one or the other
Let's discuss it

flomnes (approved):

PR #3397, opened 2026-02-09
feat: Add sts prod to fictitious load [ANT-3739]

payetvin:


src/solver/optimisation/constraints/FictitiousLoad.cpp
```diff
@@ -19,6 +19,15 @@ void FictitiousLoad::add(int pdt, int pays)
...
 
+    if (data.DefaillanceNegativeUtiliserHydro[pays])
+    {
+        for (const auto& storage: data.ShortTermStorage[pays])
+        {
+            builder.ShortTermStorageWithdrawal(storage.clusterGlobalIndex, -1.0);
+            builder.ShortTermStorageInjection(storage.clusterGlobalIndex, 1.0);
+        }
+    }
+
```
flomnes, line 29: Two remarks
1. l. 19 is crappy, we should never add zero coefficients to constraints. I suggest putting everything into an `if` clause instead of the ternary that's currently in place
2. Avoid duplication and re-use this function in the same way as `ExportPaliers`
https://github.com/AntaresSimulatorTeam/Antares_Simulator/blob/c32b046566ff8b9fb5aff85005d0cc0470300f2e/src/solver/optimisation/constraints/AreaBalance.cpp#L6-L15

flomnes (commented):
+ documentation
+ unit tests

src/solver/optimisation/constraints/FictitiousLoad.cpp
```diff
@@ -16,8 +16,12 @@ void FictitiousLoad::add(int pdt, int pays)
 
     builder.updateHourWithinWeek(pdt);
     ExportPaliers(data.PaliersThermiquesDuPays[pays], builder);
-    auto coeff = data.DefaillanceNegativeUtiliserHydro[pays] ? -1 : 0;
-    builder.HydProd(pays, coeff).Spillage(pays, 1.0);
+
+    if (data.DefaillanceNegativeUtiliserHydro[pays])
+    {
+        builder.HydProd(pays, -1.0).NegativeUnsuppliedEnergy(pays, 1.0);
+        shortTermStorageBalance(data.ShortTermStorage[pays], builder);
```
flomnes:
```suggestion
        builder.HydProd(pays, -1.0).NegativeUnsuppliedEnergy(pays, 1.0).ShortTermStorageWithdrawal(index, -1.0);
```

flomnes (approved):

JasonMarechal25 (approved):

PR #3407, opened 2026-02-16
Fix MILP problems in hybrid studies [ANT-4551]

flomnes:


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -325,11 +347,13 @@ bool OPT_AppelDuSimplexe(const SingleOptimOptions& options,
     }
     else
     {
-        LegacyOrtoolsLinearProblem infeasibleProblem(problemeHebdo->ProblemeAResoudre->isMIP(),
-                                                     options.solverName);
-        FillContext fillCtx = buildFillContext(problemeHebdo, NumIntervalle);
         const auto& modelerData = problemeHebdo->modelerData;
         bool hasModelerData = modelerData != nullptr;
+        const bool isMip = problemeHebdo->ProblemeAResoudre->isMIP()
```
a-zakir, line 333: While this code is clean, there’s some duplication with the pure modeler. It might be worth constructing this information in the optimEntity container instead.

a-zakir (approved):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
a-zakir: actually, forget about my last statement, this milp flag is used to select the right and available solver for the problem

PR #3416, opened 2026-02-17
 optim entity container : trial to simplify

guilpier-code:
Trial to simplify and clarify class  **OptimEntityContainer**'s mission.
This class is supposed to store modeler variables / constraints numbering in the LP, and give these numbers to client code. 
It appears it's also used to avoid passing some other things : if these things are stored inside class **OptimEntityContainer**, they don't have to be passed as long as we pass an instance of **OptimEntityContainer**.
This is a great way for coupling code and make it more entangled.

What was done : 
- [x] remove **scenario** and **evaluation context** reference from class **OptimEntityContainer**
- [x] move methods definitions from .h to .cpp 

What should be done : 
- [ ] rename class **OptimEntityContainer** with a better name (it's about numbering in LP)
- [ ] the LP is still given by OptimEntityContainer (via **Problem()**). Not good, if LP has to be passed, it shouldn't be passed through passing an instance of **OptimEntityContainer** 
- [ ] When we pass the scenario group repo in fillers / visitors, do we mean actually passing a scenario (and not the whole scenario repo) ?


flomnes (approved):
LGTM

PR #3429, opened 2026-02-19
Fix dynamic district aggregation

payetvin:


src/solver/variable/dynamicAggregation/dynamicAggregationSingleYear.cpp
```diff
@@ -23,6 +23,14 @@ void DynamicAggregationSingleYear::addResultsToSets(const PROBLEME_HEBDO& pb)
     }
 }
 
+void DynamicAggregationSingleYear::reset()
+{
+    for (auto& [_, setData]: setsData_)
```
JasonMarechal25, line 28: note: you can use setsData_ | views::values

JasonMarechal25 (approved):

PR #3441, opened 2026-02-26
Remove progression display

payetvin:


guilpier-code:
About progressions, class that can be found in : 
- **src\libs\antares\study\include\antares\study\progression\progression.h**
- **src\libs\antares\study\include\antares\study\progression\progression.hxx**
- **src\libs\antares\study\progression\progression.cpp**

Could be removed (especially, **Progression::Task** used to be called from legacy GUI).
We may want to remove any references to this class.
Do we want do something with it ?

JasonMarechal25:
> About progressions, class that can be found in :
> 
>     * **src\libs\antares\study\include\antares\study\progression\progression.h**
> 
>     * **src\libs\antares\study\include\antares\study\progression\progression.hxx**
> 
>     * **src\libs\antares\study\progression\progression.cpp**
> 
> 
> Could be removed (especially, **Progression::Task** used to be called from legacy GUI). We may want to remove any references to this class. Do we want do something with it ?

@copilot implement suggestion

JasonMarechal25 (approved):

PR #3442, opened 2026-02-26
Remove unused code from class Study

payetvin:


src/libs/antares/study/include/antares/study/study.h
```diff
@@ -268,7 +262,7 @@ class Study: public Yuni::NonCopyable<Study>
     ** \param output True for checking output filenames, false for input
     ** \param chfolder The study folder to take into consideration
     */
-    bool checkForFilenameLimits(bool output, const YString& chfolder = nullptr) const;
```
guilpier-code, line 271: There are even more simplifications to be done in function **checkForFilenameLimits**, as it's called only once : 
```c++
if (!study.checkForFilenameLimits(true))
{
    ...
}
``` 
and in this case argument **chfolder** has always the same default value **nullptr**.

So now, this function should have no argument.

src/libs/antares/study/study.cpp
```diff
@@ -543,211 +536,70 @@ bool Study::checkForFilenameLimits(bool output, const String& chfolder) const
         studyfolder = chfolder;
     }
```
guilpier-code: Bye bye **chfolder**

src/tests/src/libs/antares/study/test_study.cpp
```diff
@@ -331,19 +331,18 @@ BOOST_AUTO_TEST_CASE(version_parsing)
...
 {
     auto s = std::make_unique<Study>();
-    BOOST_CHECK(s->checkForFilenameLimits(true)); // empty areas should return true
+    BOOST_CHECK(s->checkForFilenameLimits()); // empty areas should return true
 
-    BOOST_CHECK(study->checkForFilenameLimits(true));
-    BOOST_CHECK(study->checkForFilenameLimits(false));
-    BOOST_CHECK(study->checkForFilenameLimits(true, "abc"));
+    BOOST_CHECK(study->checkForFilenameLimits());
+    BOOST_CHECK(study->checkForFilenameLimits("abc"));
```
guilpier-code:
```c++
BOOST_CHECK(study->checkForFilenameLimits("abc"));
```
should be removed, because this function looses all its arguments now

guilpier-code (approved):

PR #3444, opened 2026-02-26
Remove functions related to rename

payetvin:


src/libs/antares/study/include/antares/study/binding_constraint/BindingConstraintsRepository.h
guilpier-code: What we do with rename, we'll have to do it with **remove** and **add** I reckon.

For **reverseWeightSign** as well (made a quick search). Same for **AreaList::fixOrientationForAllInterconnections** by the way.

src/libs/antares/study/include/antares/study/binding_constraint/BindingConstraintsRepository.h
```diff
@@ -122,11 +122,6 @@ class BindingConstraintsRepository final: public Yuni::NonCopyable<BindingConstr
     */
     void removeConstraintsWhoseNameConstains(const AnyString& filter);
```
guilpier-code: In the list of **what to remove with legacy GUI ?**, we also have : 
```c++
void removeConstraintsWhoseNameConstains(const AnyString& filter);
```
It involves removing struct **WhoseNameContains**.
`CBuilder::deletePreviousConstraints()` should also be removed, by the way.

src/libs/antares/study/include/antares/study/binding_constraint/BindingConstraintsRepository.h
guilpier-code: `void clear();` should probably be removed.

src/libs/antares/study/include/antares/study/parts/common/cluster_list.h
guilpier-code, line 62: Same as for binding constraints : 
- remove functions will have to be removed someday.
- **saveDataSeriesToFolder** too
- what about `void rebuildIndexes();` ? (I assume this function is called only if changes were made in the list of clusters via old GUI)

src/libs/antares/study/include/antares/study/study.h
guilpier-code, line 1: `Area* areaAdd(const AreaName& name);` to be removed

src/libs/antares/study/include/antares/study/parts/hydro/allocation.h
```diff
@@ -39,8 +39,6 @@ class HydroAllocation final
     */
     void remove(const AreaName& areaid);
```
guilpier-code, line 40: Same thing for remove().

What about all functions **clear()** & **reset()** ? Without GUI, do they have an interest : if we only load once, what's the point of clearing or resetting ?

src/libs/antares/study/include/antares/study/scenario-builder/sets.h
guilpier-code, line 83: Same for **remove()**, **createNew()**. 
**exists()** ?
**clear()** ?

flomnes: Add/remove functions might still be useful for in-memory studies

src/libs/antares/study/include/antares/study/study.h
flomnes: 99% sure it is used for in-memory studies

flomnes (approved):

PR #3457, opened 2026-03-04
Remove usedByTheSolver

payetvin:


src/libs/antares/study/area/links.cpp
```diff
@@ -496,70 +496,65 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* areaList, Area* area, const
         ret = link.loadTimeSeries(study.header.version, folder) && ret;
```
guilpier-code, line 585: Let's take advantage of this, and have : 
```c++
...
ret = link.loadTimeSeries(study.header.version, folder) && ret;
checkLoadedTimeSeries(link);

// Inifile
const IniFile::Property* p = s->firstProperty;
for (; p; p = p->next)
{
    key = p->key;
    key.toLower();
    value = p->value;
    if (!AreaLinksInternalLoadFromProperty(link, key, value))
    {
        logs.warning() << '`' << p->key << "`: Unknown property";
    }
}

postConfigureLoadedData(link);
// end of AreaLinksLoadFromFolder(...)
```
About last call `postConfigureLoadedData(link);` , [further comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3457#discussion_r2885087190) is helpful to avoid returning `false`

src/libs/antares/study/area/links.cpp
```diff
@@ -576,40 +571,37 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* areaList, Area* area, const
             }
         }
 
-        // From the solver only
-        if (study.usedByTheSolver)
-        {
-            link.overrideTransmissionCapacityAccordingToGlobalParameter(
-              study.parameters.transmissionCapacities);
+        // From the solver
```
guilpier-code:
```c++
// From the solver
```
we don't need that anymore, it's implicit

src/libs/antares/study/area/links.cpp
```diff
@@ -576,40 +571,37 @@ bool AreaLinksLoadFromFolder(Study& study, AreaList* areaList, Area* area, const
...
+        case Data::LocalTransmissionCapacities::infinite:
+        {
+            // Copper plate mode
+            auto infinity = +std::numeric_limits<double>::infinity();
+            link.directCapacities.fill(infinity);
+            link.indirectCapacities.fill(infinity);
+            break;
+        }
+        default:
+            return false;
```
guilpier-code: default case is impossible : `link.transmissionCapacities` is an **enum**, it can't take a value out of its value range.

guilpier-code (commented):
Unfinished review (only a few comments)

src/libs/antares/study/area/links.cpp
JasonMarechal25: defaulting anyway is good. If you add a new enum value you may not want to fallthrough the switch cases.

src/tests/src/solver/simulation/test-time_series.cpp
```diff
@@ -21,6 +21,7 @@ namespace fs = std::filesystem;
 
 void initializeStudy(Study& study)
 {
+    study.parameters.include.constraints = true;
```
guilpier-code:
```c++
study.parameters.include.constraints = true;
```
Strange, why do we need to add this now ?
It seems totally unrelated to **usedByTheSolver**

guilpier-code: For  `include.constraints` the default value is **true** (see **parameters**' `reset()` method).
So why don't we force `include.constraints`  to **true** at construction ? 
```c++
class Parameters final
{
...
    struct IncludeOptions
    {
        bool constraints = true:
        ....
    }
}
```

guilpier-code (approved):

PR #3468, opened 2026-03-05
 Fix UI in CR24 Multiple Rule Curves

nikolaredstork:
Initial testing!


flomnes:
Starting from version 10.0, there is no GUI, save functions are being removed

PR #3470, opened 2026-03-06
Extract utils for std::vector<double>

guilpier-code:
At 3 places in the code (at least), we need easy-to-use unary / binary operators on `std::vector<double>` : **operator+/-**, **operator+=**, **operator***, sum, ...
This PR intends to start making a common `std::vector` / standard containers  utils lib.

You'll notice that all vector operator definitions disappeared from source files : 
- **src/solver/simulation/include/antares/solver/simulation/remix-storage/remix-utils.h**
- **src/solver/simulation/sim_binding_constraints_rhs.cpp**
- **src/expressions/visitors/EvalVisitor.cpp**


src/solver/simulation/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ target_sources(antares-solver-simulation
 
 target_include_directories(antares-solver-simulation
         PUBLIC
+        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/libs/antares/utils/include>
```
JasonMarechal25: suggestion: Instead of target_include_directories and full path, couldn't we make vector utils a header only library?

JasonMarechal25 (changes requested):

src/libs/antares/utils/include/antares/utils/vector-utils.h
```diff
@@ -0,0 +1,74 @@
...
+}
+
+inline std::vector<double> operator-(const std::vector<double>& v, const double c)
+{
+    std::vector<double> to_return = v;
+    rng::for_each(to_return, [&c](double& e) { e = e - c; });
+    return to_return;
+}
+
+inline std::vector<double> operator+(std::vector<double> a, const std::vector<double>& b)
```
payetvin, line 44: maybe add checks for size ?

JasonMarechal25: I concoure. 
Or at least explicite the behaviour.

src/solver/simulation/CMakeLists.txt
guilpier-code: Yes, wanted to, but later. 
You gave me push to do it.
Done now.

src/libs/antares/utils/include/antares/utils/vector-utils.h
guilpier-code: As it is now, the transform algo processes the min(a.size(), b.size()) first element and stops.
If both input vectors are equally sized, every elements are processed in both vectors.
There will never be a exception due to different size.
I'll leave a comment in the code... now done.

src/libs/antares/utils/utils.cpp
```diff
@@ -11,50 +11,6 @@ using namespace Yuni;
 
 namespace Antares
 {
-void BeautifyName(YString& out, AnyString oldname)
```
guilpier-code, line 14: **BeautifyName** not used anymore (due to GUI removal)

src/libs/antares/utils/CMakeLists.txt
```diff
@@ -1,25 +1,47 @@
-set(PROJ StudyUtils)
-set(SRC_PROJ
+# =========================
```
guilpier-code, line 1: Note that there are now 2 independent targets : **study-utils** and **vector-utils** instead of one **StudyUtils** previously

JasonMarechal25 (approved):

PR #3483, opened 2026-03-12
Modeler > Port field definition : allow non linear expressions [ANT-4646]

guilpier-code:
Fixes [ticket ANT-4646](https://gopro-tickets.rte-france.com/browse/ANT-4646)

Ticket **ANT-4646** is addressed (hopefully) : we relaxed the prohibition in expression of models' **port-field-definitions** (any non-linear operator is allowed), and we made sure that, once components and connections are read, port field resolution in binding constraint section leads to linear expressions only.

While addressing **ANT-4646**, what was done (more precisely) : 
- [x] code about forbidden node was moved in a independent target **forbidden-nodes** so that it can easily be accessible from both **convert model** (model creation) and **convert system** (components creation and connections). In this context, source files about forbidden nodes were moved to another location.
- [x] Class `ForbiddenNodesVisitor`  was modified to be able to make port field resolution and check linearity.
- [x] we need to modify class **SystemModel::Constraint** in order to be able to distinguish simple constraints and binding constraints 
- [x] determination of port field role (sender or receiver) was moved from **system conversion** into **model conversion** (thus, earlier) : we don't have to wait until **system conversion** in order to find out these roles.
- [x] Tests : 
  - [x] study (full modeler) for cucumber, to check that defining a **thermal cluster income** works 
  - [x] Unit tests to check that prohibition works  
- [x] updating documentation


src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -87,58 +87,43 @@ static void CommonPreSolve(ForbiddenNodes& f)
 
 static ForbiddenNodes ForbiddenInConstraint()
```
guilpier-code: Code about defining node prohibition in an expression (depending on use context) was simplified. 
- What remains : static functions are still called once.
- What we loose : calls to these functions are eager (they are called once exactly, even if not needed). Code used to be lazy (calls were made once, and only if needed).
- What we gain : code is more readable, more simple, has less lines

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -59,86 +59,93 @@ PortInDefinition::PortInDefinition(const std::string& portId, const std::string&
...
+{
+    return {};
+}
+
+const ForbiddenNodes forbiddenInConstraint = ForbidNodesInConstraint();
+const ForbiddenNodes forbiddenInBindingConstraint = ForbidNodesInBindingConstraint();
+const ForbiddenNodes forbiddenInVariableBounds = ForbidNodesInVariableBounds();
+const ForbiddenNodes forbiddenInPortFieldDef = ForbidNodesInPortFieldDef();
+const ForbiddenNodes forbiddenInObjective = ForbidNodesInObjective();
+const ForbiddenNodes forbiddenInExtraOutput = ForbidNodesInExtraOutput();
```
flomnes: I'm not sure why this needs to be added

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -346,12 +352,17 @@ std::vector<PortFieldDefinition> convertPortFieldDefinitions(const YmlModel::Mod
...
+        ForbiddenNodesVisitor(forbiddenInPortFieldDef, pfdefinition.definition)
           .dispatch(nodeRegistry.node);
+
         portFieldDefinitions.emplace_back(*itPort,
                                           *itField,
                                           Expression(pfdefinition.definition,
                                                      std::move(nodeRegistry)));
+
+        // A definition for a port field means this field is a sender
+        itPort->setFieldRole(itField->Id(), FieldRole::Sender);
```
flomnes, line 293: unclear

flomnes (commented):
The logic for port field role is unclear

flomnes (commented):
Lots of refactoring / renaming, can you split this PR in 2 ?
1. Refactor / rename
2. Implement the actual feature

Thank you

guilpier-code:
> The logic for port field role is unclear

what do you mean by unclear ? unclear in what way ?

src/io/inputs/model-converter/modelConverter.cpp
guilpier-code: ok, so at least, here is an explanation : 
About port field roles, we make 2 things (before and after this work, same) : 
1. we determine the role of every port field (sender / receiver) for every component
2. we check that 2 connected port field are compatible : one must be a sender and the other a receiver.

Before this work, step 1 and 2 were made after components and connections creation.
After this work, step 1 is made after models are created, and step 2 after components and connections creation.

Why did I do that ?
For 2 reasons : 
- I first thought I would need it that way (but I eventually didn't)
- As port field roles can be determined sooner (as soon as required information is available), it better to do that. The sooner the better.

src/io/inputs/model-converter/modelConverter.cpp
guilpier-code: I did that for performance reason : otherwise we would have to call the  same function each time we need it.
Here the result of this call is store in a constant.

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -359,28 +280,44 @@ std::vector<PortFieldDefinition> convertPortFieldDefinitions(const YmlModel::Mod
...
 }
 
-static void addSingleConstraint(std::vector<Constraint>& constraints,
-                                const YmlModel::Constraint& constraint,
-                                const YmlModel::Model& model,
-                                const ForbiddenNodes& forbiddenNodes)
+static Constraint createConstraint(const YmlModel::Constraint& constraint,
+                                   const YmlModel::Model& model,
+                                   const ForbiddenNodes& forbiddenNodes)
+{
```
flomnes: Avoid code duplication, add a `bool` parameter
```suggestion
static Constraint createConstraint(const YmlModel::Constraint& constraint,
                                   const YmlModel::Model& model,
                                   const ForbiddenNodes& forbiddenNodes,
                                   bool isBinding)
{
```

src/io/inputs/forbidden-nodes/ForbiddenNodesVisitor.cpp
```diff
@@ -33,6 +34,16 @@ ForbiddenNodesVisitor::ForbiddenNodesVisitor(const ForbiddenNodes& forbiddenNode
 {
 }
 
+// Constructor called to check expressions after component were built, so we have access to the
+// component and its ports fields definitions.
+ForbiddenNodesVisitor::ForbiddenNodesVisitor(const ForbiddenNodes& forbid,
+                                             const std::string& expression,
+                                             const ModelerStudy::SystemModel::Component* component):
```
flomnes: Now in some cases, you have `component_ = nullptr` and other cases `component_ != nullptr`

This is a bit dangerous, since you have to check for `component_` before using it

src/io/inputs/forbidden-nodes/ForbiddenNodesVisitor.cpp
```diff
@@ -101,11 +112,36 @@ void ForbiddenNodesVisitor::visit(const ParameterNode*)
...
 void ForbiddenNodesVisitor::visit(const PortFieldNode* portFieldNode)
 {
+    // gp : could be dead code, considering visit(const PortFieldSumNode* portFieldSumNode)
     checkIsForbidden(portFieldNode, typeIndexOf<PortFieldNode>());
 }
 
 void ForbiddenNodesVisitor::visit(const PortFieldSumNode* portFieldSumNode)
 {
+    if (component_)
+    {
```
flomnes: The behavior of this visitor is quite different if `component_ == nullptr` vs `component_ != nullptr`

I suggest transferring `component_` to a reference and making it mandatory

src/io/inputs/forbidden-nodes/ForbiddenNodesVisitor.cpp
flomnes: See https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3512

src/io/inputs/model-converter/modelConverter.cpp
guilpier-code: Done

flomnes (approved):

PR #3498, opened 2026-03-17
Remove yuni strings for alias areaname, clustername... 

payetvin:


src/libs/antares/study/scenario-builder/rules.cpp
```diff
@@ -147,21 +149,25 @@ bool Rules::readRenewableCluster(const AreaName::Vector& splitKey,
     }
     else
     {
-        bool isTheActiveRule = (pName.toLower() == study_.parameters.activeRulesScenario.toLower());
+        std::string pNameLower = boost::to_lower_copy(pName);
+        std::string activeRuleLower = boost::to_lower_copy(study_.parameters.activeRulesScenario);
+        bool isTheActiveRule = (pNameLower == activeRuleLower);
```
flomnes: Add a `compareCaseInsensitive` function ?

src/libs/antares/study/scenario-builder/sets.cpp
```diff
@@ -115,16 +117,16 @@ bool Sets::internalLoadFromINIFile(const AnyString& filename)
           }
 
           RulesScenarioName name = section.name;
-          name.trim(" \t");
-          if (!name)
+          boost::trim(name);
```
flomnes, line 120: May differ for some characters

src/solver/application/application.cpp
```diff
@@ -163,10 +163,6 @@ void Application::readDataForTheStudy(Data::StudyLoadOptions& options)
         loadingException = std::current_exception();
     }
 
-    // This settings can only be enabled from the solver
-    // Prepare the output for the study
-    study.prepareOutput();
-
```
flomnes, line 169: are you sure about removing this ?

src/libs/antares/date/date.cpp
```diff
@@ -544,112 +545,114 @@ void Calendar::reset()
...
             }
-            hourtext << ":00";
+            oss << ":00";
+            hourtext = oss.str();
         }
 
         // days
         for (uint d = 0; d != maxDaysInYear; ++d)
         {
             auto& str = text.daysYear[d];
```
flomnes: This reference adds nothing, you can just remove it and at the end,
```cpp
text.daysYear[d] = oss.str()
```

src/libs/antares/correlation/correlation.cpp
```diff
@@ -108,8 +108,8 @@ static inline void ReadCorrelationCoefficients(Correlation& correlation,
             continue;
         }
 
-        const AreaName::Size offset = p->key.find('%');
-        if (offset == AreaName::npos or !offset or offset == p->key.size() - 1)
+        const size_t offset = p->key.find('%');
+        if (offset == std::string::npos or !offset or offset == p->key.size() - 1)
```
flomnes, line 115: Maybe add a comment here for clarity
- First case corresponds to "not found"
- Second corresponds to "%" in first position
- Third correspond to "%" in last position
All these cases are non-valid

src/libs/antares/study/include/antares/study/fwd.h
```diff
@@ -40,13 +41,13 @@ class StudyRuntimeInfos;
...
+using AreaName = std::string;
 //! Name of a single link
-using AreaLinkName = Yuni::CString<ant_k_area_name_max_length * 2 + 1, true>;
+using AreaLinkName = std::string;
 
-using ConstraintName = Yuni::CString<ant_k_constraint_name_max_length, true>;
+using ConstraintName = std::string;
 
-using RulesScenarioName = Yuni::CString<64, true>;
+using RulesScenarioName = std::string;
```
flomnes: You can just remove these aliases and use `std::string`

flomnes (changes requested):
Check
- Removal of `study.prepareOutput();`
- Replacing `Yuni::trim` with `boost::trim`, they may behave differently

src/solver/application/application.cpp
payetvin: It's already called in Study::parameterFiller which is called when loading the study

flomnes (approved):

PR #3507, opened 2026-03-20
Refactoring findExistingLinkWith function to reduce duplicate code with const version

YoshuaLisasi:
Refactor findExistingLinkWith: remove duplication via const delegation

Both the const and non-const overloads of Area::findExistingLinkWith contained the same search logic, which is a maintenance risk — any change would need to be applied in both places.

To address this, the implementation now lives only in the const version. The non-const overload simply delegates to it:

AreaLink* Area::findExistingLinkWith(Area& with)
{
    return const_cast<AreaLink*>(std::as_const(*this).findExistingLinkWith(with));
}

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ _build*
 .vscode/
 *.dll
 .idea/
+src/tests/resources/modeler/**/output/
 
 # pytest
 __pycache__
--- a/src/libs/antares/study/area/area.cpp
+++ b/src/libs/antares/study/area/area.cpp
@@ -64,64 +64,29 @@ void Area::clearAllLinks()
     }
 }
 
-AreaLink* Area::findExistingLinkWith(Area& with)
+const AreaLink* Area::findExistingLinkWith(const Area& with) const
 {
     if (&with == this)
     {
         return nullptr;
     }
 
-    if (not links.empty())
+    for (const auto& [key, link]: links)
     {
-        const AreaLink::Map::iterator end = links.end();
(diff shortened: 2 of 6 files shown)


guilpier-code:
It turns out we don't need the **non-const** version of function **findExistingLinkWith(...)**.
Indeed, removing it leads to : 
- remove a non const version of calling function **AreaList::findLink(...)**
- add a **const** to the returned pointer where this last function is called. 
- make one small adaptation on a **getLink** function in **rules.cpp**

And we're done.

By the way, we can also : 
- remove **AreaListAddLink(...)**, called nowhere.
- probably easily remove argument  **updaterMode** of `Rules::getLink`, because this argument has always the same value. 

These removals are probably made possible because of recent GUI removal.

.gitignore
```diff
@@ -55,6 +55,7 @@ _build*
 .vscode/
 *.dll
 .idea/
+src/tests/resources/modeler/**/output/
```
guilpier-code, line 58: Why do we need this ?

guilpier-code (approved):

.gitignore
YoshuaLisasi: These output folders are generated locally when running modeler tests. 
They shouldn't be tracked by git.

``` c++
...
  src/tests/resources/modeler/10_1/output/
        src/tests/resources/modeler/10_2/output/
        src/tests/resources/modeler/10_3/output/
        src/tests/resources/modeler/10_4/output/
        src/tests/resources/modeler/10_5/output/
...
```

PR #3512, opened 2026-03-24
Refactor `ForbiddenNodesVisitor`, split for system/libary

flomnes:


src/io/inputs/yml-system/converter.cpp
```diff
@@ -271,7 +271,8 @@ void checkForNonLinearityBehindConnections(const std::vector<Component>& compone
         for (const auto& binding_constraint: component.getModel()->Constraints() | filterBC)
         {
             const auto& expression = binding_constraint.expression();
-            ForbiddenNodesVisitor visitor(forbidNonLinearNodes, expression.Value(), &component);
+            Antares::IO::Inputs::ForbidNodes::ForbiddenNodesVisitorWithComponent
```
guilpier-code: We don't need `Antares::IO::Inputs::ForbidNodes::` here.

src/io/inputs/forbidden-nodes/include/antares/io/inputs/forbidden-nodes/ForbiddenNodesVisitor.h
```diff
@@ -65,11 +62,24 @@ class ForbiddenNodesVisitor final: public Expressions::Visitors::NodeVisitor<voi
...
 
+protected:
     // Data members
     const ForbiddenNodes& forbiddenNodes_;
     std::vector<std::pair<std::string, std::type_index>> parentsStack_;
     const std::string& expression_;
-    const ModelerStudy::SystemModel::Component* component_ = nullptr;
+};
+
+class ForbiddenNodesVisitorWithComponent final: public ForbiddenNodesVisitor
```
guilpier-code: ok, why not ?
I would call this class **ForbiddenNodesInComponentVisitor**, because we have a component and want to see if it contains forbidden nodes.

src/io/inputs/yml-system/converter.cpp
flomnes: done

src/io/inputs/forbidden-nodes/include/antares/io/inputs/forbidden-nodes/ForbiddenNodesVisitor.h
flomnes: done

guilpier-code (approved):

PR #3523, opened 2026-03-25
refacto: better error handling in input

JasonMarechal25:
Simplify and homognise error handling in modeler/inputs. Use one exception type instead of several ones. Using a different type for each error only cluttered the code and didn't bring any value in itself. The useful information was contained in the exception message.

**Error handling unification:**

* Introduced a new header-only INTERFACE target, `input-error`, which defines the `InputError` exception class as the base for all input-related errors in the `io/inputs` layer. Updated CMake files to install and link this target where necessary.
* Refactored all code in `data-series-csv-importer`, `forbidden-nodes`, and `model-converter` to throw `IO::Inputs::InputError` instead of standard exceptions like `std::invalid_argument`, `std::runtime_error`, or custom exceptions.

**Codebase cleanup and dependency updates:**

* Removed obsolete custom exception classes such as `ForbiddenNodeFound`, `NoPortWithThisId`, and `NotImplemented` that are now replaced by `InputError`
* Updated CMake files for `forbidden-nodes`, `model-converter`, and `data-series-csv-importer` to link against the new `input-error` target, ensuring all components use the unified error type.

**Header inclusions and namespace adjustments:**

* Added necessary includes for `InputError` in public headers and removed unnecessary includes. Adjusted namespaces to ensure correct usage of the new error type.

These changes collectively standardize error handling for input modules, making the codebase more robust and easier to maintain.


src/io/inputs/data-series-csv-importer/include/antares/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.h
```diff
@@ -5,6 +5,7 @@
 
 #include <filesystem>
 
+#include <antares/io/inputs/InputError.h>
```
guilpier-code: It seems we don't need this **include** here but only in the associated **cpp** file.
We should keep the header as simple as possible, ads it's the public interface of  class **DataSeriesRepoImporter**.

src/io/inputs/data-series-csv-importer/DataSeriesRepoImporter.cpp
```diff
@@ -91,7 +91,7 @@ static bool ParseRow(const char* first,
                 std::ostringstream oss;
                 oss << errorMessagePrefix << ": row (" << rowIndex << ") has more columns ("
                     << colIndex + 1 << ") than the expected (" << columns.front().size() << ").";
-                throw std::invalid_argument(oss.str());
+                throw IO::Inputs::InputError(oss.str());
```
guilpier-code: As we are in a cpp file, we could avoid using expanded namespace **IO::Inputs::InputError** and use simply **InputError**, shorter and clear enough.
Not mandatory, up to you.

src/io/inputs/CMakeLists.txt
```diff
@@ -1,5 +1,21 @@
 OMESSAGE("Antares IO-Inputs")
 
+# Header-only INTERFACE target exposing InputError
```
guilpier-code, line 3: As you know, I work on Windows (and MSVC) : 
When search for **Antares::input-error** in my IDE, search does not spot an occurrence in this file.
So it's as if this **target** was never defined, although it clearly is.

So would it makes sense to move this definition in a new sub folder (for example named **input-error**) ?
It would certainly make this target visible for research on **MSCV**.

But maybe I should just stop using **MSVC** and use another **IDE**, like **VSCode** or **CLion**.

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -187,7 +182,8 @@ Node* ConvertorVisitor::convertIdentifier(const std::string& identifier) const
             return static_cast<Node*>(registry_.create<VariableNode>(var.id, index, v));
         }
     }
-    throw NoParameterOrVariableWithThisName(identifier);
+    throw IO::Inputs::InputError("No parameter or variable found for this identifier: "
```
guilpier-code: Do we need **IO::Inputs::** ?

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/modelConverter.h
```diff
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: MPL-2.0
 
 #pragma once
-#include <stdexcept>
 
+#include "antares/io/inputs/InputError.h"
 #include "antares/io/inputs/yml-model/Library.h"
 
 namespace Antares
```
guilpier-code: Here, there are 3 forward declarations we don't need, for various reasons.
Either declared type are simply not used, or there is an included header already declares the type.

src/io/inputs/yml-model/decoders.cpp
```diff
@@ -57,7 +33,7 @@ void checkMandatoryIdField(const Node& node, const std::string& nodeName)
         throw KeyNotFound(node.Mark(),
                           fmt::format("{} id is mandatory in library\n{}",
                                       nodeName,
-                                      printPathTree(nodePath)));
+                                      Antares::IO::Inputs::YmlUtils::printPathTree(nodePath)));
```
guilpier-code, line 36: Do we need **Antares::IO::Inputs::YmlUtils::** ?

src/io/inputs/yml-system/include/antares/io/inputs/yml-system/converter.h
```diff
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <antares/io/inputs/InputError.h>
```
guilpier-code:
```c++
#include <antares/io/inputs/InputError.h>
```
Should not be here, but in a **cpp** source file.

src/io/inputs/yml-system/converter.cpp
```diff
@@ -22,39 +23,13 @@ namespace Antares::IO::Inputs::SystemConverter
...
-    }
-};
 
 std::pair<std::string, std::string> splitLibraryModelString(const std::string& s)
 {
     size_t pos = s.find('.');
     if (pos == std::string::npos)
     {
-        throw ErrorWhileSplittingLibraryAndModel(s);
+        throw IO::Inputs::InputError("'.' not found while splitting library and model: " + s);
```
guilpier-code: **IO::Inputs::** useful here ?

src/io/inputs/yml-model/decoders.cpp
JasonMarechal25: yes

src/io/inputs/model-converter/include/antares/io/inputs/model-converter/modelConverter.h
guilpier-code: I removed all the forward declarations, and it still compiles, at least on Windows.

payetvin (approved):

PR #3526, opened 2026-03-26
refacto: improve error handling in GEMS yml loading

JasonMarechal25:
This pull request introduces several improvements focused on YAML input decoding, error handling, and code organization for the Antares input modules. The main changes include enhanced validation and error reporting for YAML parsing, refactoring of decoder implementations, and minor code cleanups.

**YAML Decoding and Error Handling Improvements:**

* Introduced a `requireMap` helper function in both `yml-model` and `yml-optim-config` decoders to enforce that YAML nodes are mappings, throwing a detailed `InputError` if not, and updated all relevant decode functions to use this helper for more consistent and informative error handling.

**Code Refactoring and Organization:**

* Moved the implementation of the YAML decoder specializations for `yml-optim-config` types from the header to a new `decoders.cpp` file, and updated the CMake configuration to include this new source file. The header now only contains function declarations.

**Minor Cleanups:**

* Fixed formatting and removed unnecessary blank lines in several files for better code readability.
These changes collectively improve the robustness, maintainability, and clarity of YAML input parsing and error reporting in the Antares codebase.


src/io/inputs/forbidden-nodes/ForbiddenNodesVisitor.cpp
```diff
@@ -182,8 +182,8 @@ std::type_index functionNodeTypeIndex(const FunctionNode* functionNode)
     case FunctionNodeType::ceil:
         return typeIndexOf<FunctionNodeType::ceil>();
     default:
-        throw IO::Inputs::InputError("ForbiddenNodesVisitor > function '"
-                                     + functionNode->name() + "' is unknown.");
+        throw IO::Inputs::InputError("ForbiddenNodesVisitor > function '" + functionNode->name()
```
guilpier-code: Do we need **IO::Inputs::** ?

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -230,9 +230,8 @@ std::vector<PortFieldDefinition> convertPortFieldDefinitions(const YmlModel::Mod
                                             { return field.Id() == pfdefinition.field; });
         if (itField == portFields.end())
         {
-            throw IO::Inputs::InputError("In port-field-definitions, for port: "
-                                         + pfdefinition.port + " , field not found: "
-                                         + pfdefinition.field);
+            throw IO::Inputs::InputError("In port-field-definitions, for port: " + pfdefinition.port
```
guilpier-code: do we need **IO::Inputs::** ?

src/io/inputs/yml-model/decoders.cpp
```diff
@@ -5,9 +5,30 @@
 
 #include <unordered_set>
 
+#include <antares/io/inputs/InputError.h>
+
 namespace YAML
 {
 
+namespace
```
guilpier-code, line 15: We should remove **Antares::IO::Inputs::**, there are countless occurrences of it here, they make code harder to read than it should.

src/io/inputs/yml-model/decoders.cpp
```diff
@@ -5,9 +5,30 @@
...
+{
+/// Throws InputError if the node is present but is not a YAML map.
+/// Returns false silently when the node is absent or null (permitting as_fallback_default).
+bool requireMap(const Node& node, const char* typeName)
```
guilpier-code: **Question** : why a `const char* typeName` and not a `const std::string&` ?

src/io/inputs/yml-model/decoders.cpp
```diff
@@ -118,7 +139,8 @@ bool convert<Antares::IO::Inputs::YmlModel::PortType>::convertAreaConnectionFiel
 
     if (!area_conn_node.IsMap())
     {
-        return false;
+        throw Antares::IO::Inputs::InputError(
+          "Expected a YAML mapping for 'area-connection' in port-type");
     }
```
guilpier-code: Shouldn't we use **requireMap** here instead ?
```c++
if (!requireMap(area_conn_node, "area-connection"))
{
    return false;
}
```

src/io/inputs/yml-model/decoders.cpp
```diff
@@ -146,7 +168,8 @@ bool convert<Antares::IO::Inputs::YmlModel::PortType>::convertThermalCapacityFie
 
     if (!child_node.IsMap())
     {
-        return false;
+        throw Antares::IO::Inputs::InputError(
```
guilpier-code: Same as above (use of **requireMap**)

src/io/inputs/yml-model/decoders.cpp
guilpier-code: About : 
```c++
checkFields(area_conn_node,
            std::unordered_set<std::string>{"injection-to-balance",
                                            "spillage-bound",
                                            "unsupplied-energy-bound"});
```
We could make it more simple : 
```c++
checkFields(area_conn_node,
            {"injection-to-balance", "spillage-bound", "unsupplied-energy-bound"});

```
Besides, do we need `std::unordered_set<std::string>` here ? Are we that concerned with performance that we can't use simple `std::set<std::string>` ?

src/io/inputs/yml-model/decoders.cpp
guilpier-code: I would rename **requireMap(...)** into **checkIsMap(...)**, more accurate.

src/io/inputs/yml-model/decoders.cpp
guilpier-code, line 100: The lambda **diffSet** is defined inside **checkFields(...)**.
We should move **diffSet** out of **checkFields(...)**, this last function would be shorter and a bit more understandable.

guilpier-code, line 54: We could move all helper functions regarding **yaml maps** (**checkKeysIfMap**, **checkFields**, **diffSet**) into a new couple of (.h|.cpp) files (**check-yaml-map.h** ?).
It would lighten current source file.

guilpier-code, line 61: This **if** statement may be useless : we already are in the context of a map : we are supposed to know that **node is** a map.

guilpier-code, line 78: More generally on **checkKeysIfMap(...)**, several remarks can be made (in my opinion): 
```c++
YmlMapMarker marker(node);

// If allowedFields provided, check equality of key sets
if (!allowedFields.empty())
{
    if (marker.actualSet() == allowedFields)
    {
        return std::nullopt; // valid
    }
    // invalid map: return marker so caller can build an error message
    return std::make_optional<YmlMapMarker>(marker);
}

// invalid size: return marker so caller can build an error message
return std::make_optional<YmlMapMarker>(marker);
```
The main mission of **checkKeysIfMap(...)** seems to be checking if a map holds expected fields or not. So it should return a bool, and it should not involve a `YmlMapMarker`, and even less a `std::optional<YmlMapMarker>`. 
`YmlMapMarker` comes after, when we're sure that map does not hold the right fields.
To me, **checkKeysIfMap(...)** is even useless. 
Please let me explain : here's how I see its caller (**checkFields(...)**) : 
```c++
void checkFields(const Node& node, const std::set<std::string>& expectedFields)
{
    const std::set<std::string> actual_fields =  actualFields(node);
    if (expectedFields == actual_fields )
    {
        return;
    }
    
    const auto unexpected = diffSet(actual_fields, expectedFields);
    const auto missing = diffSet(expectedFields, actual_fields);
    YmlMapMarker marker(node.Tag(), unexpected, missing);
    
    const std::string message = makeErrorMsg(unexpected, missing, marker);
    throw Exception(node.Mark(), message);
}
```
Doing it that way : 
- code is more clear
- there is no strange `std::optional`
- YmlMapMarker does nothing else than deliver a tree under the form of a string.
- **checkKeysIfMap(...)** disappears

By the way, note that : 
- **YmlMapMarker::buildMarkedTreeAllPresent()** is useless and should be removed
- **YmlMapMarker::buildMarkedTreeForUnexpectedAndMissing()** has a uselessly long name and could be called **buildMarkedTree**
- class **YmlMapMarker** is entirely defined is a header file (should be split into .h .cpp)
- class **YmlMapMarker** could be renamed into **YmlTreeDisplayer**, more expressive and accurate 
- There does seem to be unit tests on that class. Should we make some ?

src/io/inputs/yml-optim-config/include/antares/io/inputs/yml-optim-config/decoders.h
```diff
@@ -21,122 +21,43 @@ template<>
 class convert<Antares::IO::Inputs::YmlOptimConfig::Variable>
```
guilpier-code, line 21: About : 
```c++
#include "yaml-cpp/yaml.h"
```
If not needed here, to be moved to associated **.cpp**

src/io/inputs/yml-system/converter.cpp
```diff
@@ -250,59 +250,50 @@ void checkForNonLinearityBehindConnections(const std::vector<Component>& compone
 
 System convert(const YmlSystem::System& ymlSystem, const std::vector<Library>& libraries)
```
guilpier-code, line 251: **IO::Inputs::** useless ?

src/io/inputs/yml-utils/include/antares/io/inputs/yml-utils/path_utils.h
```diff
@@ -6,11 +6,15 @@
...
+#include "yaml-cpp/yaml.h"
+
 namespace Antares::IO::Inputs::YmlUtils
 {
 
 std::string printPathTree(const std::filesystem::path& p);
 
 std::string getBaseTree(const std::filesystem::path& nodeTagPath);
 
+bool requireMap(const YAML::Node& node, const char* typeName);
```
guilpier-code, line 18: Devs may be in progress because  **requireMap** exists elsewhere (in some **decoders.cpp** file), so there's a code duplication.

src/io/inputs/yml-model/decoders.cpp
JasonMarechal25: Good catch, I've changed to use string_view

src/io/inputs/yml-model/decoders.cpp
JasonMarechal25: I've simplified call sites.

We don't "need" unordered_set, but it is always a good habit to have. Since I have simplified the call sites it's now less intrusive

src/io/inputs/yml-model/decoders.cpp
JasonMarechal25: I beg to defer. It would be the case if the function returned true/false, however in the current implementation will throw an error if node is not a map but a value.

src/io/inputs/yml-model/decoders.cpp
JasonMarechal25: I've refactored in 0f6efc8eb6970e02256b563cde476c0001c4881e

JasonMarechal25: Indeed.
Could be done in a new PR

JasonMarechal25: @copilot This comment make a lot of sense. Review it to see if we missed something. Otherwise open a PR following the suggestions.

src/io/inputs/yml-utils/include/antares/io/inputs/yml-utils/path_utils.h
JasonMarechal25: Good catch, fixed in 974203bdd042344f0a947da1a6579690f0d2da96

src/io/inputs/yml-model/decoders.cpp
guilpier-code: no problem, we can differ

src/io/inputs/yml-model/decoders.cpp
guilpier-code: ok

guilpier-code (approved):

PR #3527, opened 2026-03-26
Remove member function Study::areaAdd [ANT-4815]

YoshuaLisasi:
- Remove the function
- Transform loadListFromFile to allow a call to addAreaToListOfAreas
- move the case lname empty to AreaListAddFromNames (because duplicate code)
- remove CheckForbiddenCharacterInAreaName, useless (due to previous call of transformNameIntoID)
- Replace calls to areaAdd in tests by calls to addAreaToListOfAreas

Possible follow ticket :  Area::createMissingData and Area::resetToDefaultValues are unused. Can they be removed safely ?

### DIFF
--- a/src/libs/antares/study/area/list.cpp
+++ b/src/libs/antares/study/area/list.cpp
@@ -268,10 +268,8 @@ bool AreaList::loadListFromFile(const fs::path& filename)
     AreaName name;
     // Each lines in the file
     std::string buffer;
-    uint line = 0;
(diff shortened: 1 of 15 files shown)


flomnes (changes requested):
`CheckForbiddenCharacterInAreaName` becomes unused, so invalid area names are no longer checked after this PR

Please use this method after loading the area

flomnes (approved):

PR #3530, opened 2026-03-30
Districts: fix offset when no output [ANT-4830]

payetvin:


src/solver/variable/include/antares/solver/variable/container.hxx
```diff
@@ -230,7 +247,8 @@ void List<NextT>::buildSurveyReport(SurveyResults& results,
     if (dataLevel == Category::DataLevel::setOfAreas && fileLevel == Category::FileLevel::va
         && dynamicAggregationAllYears_)
     {
-        const auto& setName = pStudy->setsOfAreas.nameByIndex(results.data.setOfAreasIndex);
+        uint originalIdx = findOriginalSetIndex(*pStudy, results.data.setOfAreasIndex);
+        const auto& setName = pStudy->setsOfAreas.nameByIndex(originalIdx);
```
flomnes, line 251: While I understand the logic, this seems quite complex
1. Find the original index, taking disabled sets into account
2. Find the name of the set with that index

There should be a way with `results.data.setOfAreasIndex` directly

flomnes (commented):
+ Add test case with disabled district

payetvin:
done in #3533

PR #3535, opened 2026-04-01
Fix: remove dynamique variables from digest

JasonMarechal25:
When there are dynamique variable activated, digest is wrong. Some columns have the wrong values.
We remove dynamique variables from digest and fix values for other columns.
Dynamique variables will be reintegrated properly in digest later

### DIFF
--- a/.github/workflows/centos7.yml
+++ b/.github/workflows/centos7.yml
@@ -7,7 +7,6 @@ on:
     branches:
     - develop
     - dependabot/**
-    - fix/*
   schedule:
   - cron: '21 2 * * *'
   workflow_call:
--- a/.github/workflows/centos7.yml
+++ b/.github/workflows/centos7.yml
@@ -7,6 +7,7 @@ on:
     branches:
     - develop
     - dependabot/**
+    - fix/*
   schedule:
   - cron: '21 2 * * *'
   workflow_call:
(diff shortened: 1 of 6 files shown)


src/solver/variable/include/antares/solver/variable/variable.hxx
```diff
@@ -315,7 +315,7 @@ inline void IVariable<ChildT, NextT, VCardT>::buildDigest(SurveyResults& results
                                                           int dataLevel) const
 {
     // Generate the Digest for the local results (areas part)
-    if (VCardType::columnCount != 0
+    if (VCardType::columnCount > 0
```
flomnes: Add a comment here, since you use the fact `dynamicColumn = -1`

For example "exclude dynamic variables & no columns"

JasonMarechal25: 4d253a9abbb1f71656b455db5af675a9ac11a121

src/solver/variable/include/antares/solver/variable/variable.hxx
```diff
@@ -309,13 +309,24 @@ inline void IVariable<ChildT, NextT, VCardT>::buildAnnualSurveyReport(SurveyResu
...
 
+template<class ChildT, class NextT, class VCardT>
+inline bool IVariable<ChildT, NextT, VCardT>::hasColumn() const
+{
+    // Leverage the fact that dynamicType has columnCount = -1
+    if (VCardType::columnCount > 0)
+    {
+        return true;
+    }
+    return false;
```
flomnes:
```suggestion
    return VCardType::columnCount > 0;
```

.github/workflows/centos7.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches:
     - develop
     - dependabot/**
+    - fix/*
   schedule:
```
flomnes:
```suggestion
schedule:
```

flomnes (approved):

PR #3555, opened 2026-04-08
Merge two conditions, simplify

flomnes:


flomnes:
The two-pass is necessary
1. Read all reserves
2. Read symmetries related to previously loaded reserves

PR #3563, opened 2026-04-13
Allow problem-generator to load GEMS studies [ANT-4847]

payetvin:
Fixes [ticket ANT-4847](https://gopro-tickets.rte-france.com/browse/ANT-4847)


src/api/main.cpp
```diff
@@ -14,7 +20,28 @@ int main(const int argc, const char** argv)
...
+    const fs::path studyPath(argv[1]);
+
+    const fs::path antaresMarker = studyPath / "study.antares";
+    const fs::path modelerMarker = studyPath / "parameters.yml";
+
+    if (fs::exists(antaresMarker))
+    {
+        logs.info() << "Loading Antares study...";
+        const SingleProblemGetter getter(studyPath);
+        getter.printProblems();
```
guilpier-code, line 32: **_This remark is irrelevant here, but it should be addressed at some point_**

Why do we call this object **SingleProblemGetter** whereas we obviously print MPS and Structure files for many problems ?
I talked to @tbittar, he's not sure as well.

src/api/main.cpp
```diff
@@ -14,7 +20,28 @@ int main(const int argc, const char** argv)
...
+    }
+    else if (fs::exists(modelerMarker))
+    {
+        logs.info() << "Loading Modeler study...";
+        const SingleProblemGetterModeler getter(studyPath);
```
guilpier-code, line 37: Technically, **SingleProblemGetterModeler** should be renamed into **ModelerProblems**.
We really should ban **single** from the class name, misleading.
Changing class name implies also changing source files names.

Less important :
It could also have a get() and a print() methods : 
```c++
...
else if (fs::exists(modelerMarker))
{
    logs.info() << "Loading Modeler study...";
    const ModelerProblem modelerProblems(studyPath);
    modelerProblems.getProblems();
    modelerProblems.printProblems();
}
```

src/api/singleProblemGetterModelerImpl.cpp
```diff
@@ -0,0 +1,44 @@
...
+    modeler_->exportMps();
+    modeler_->exportStructureFile();
+}
+
+void SingleProblemGetterModeler::printProblems() const
+{
+    const auto& master = modeler_->masterProblem();
+    const auto& subproblems = modeler_->subproblems();
+
+    auto printProblem = [](const auto& problem, const char* name)
```
guilpier-code, line 28: About **printProblem** lambda :
-  Defining a lambda inside a caller function makes caller less readable. Lambdas are handy in other context, but not necessarily in this one.
```c++
void printProblem(const auto& problem, const char* name)
{
    if (problem)
    {
        logs.info() << name << ": " << problem->variableCount() << " variables, "
                    << problem->constraintCount() << " constraints";
    }
};

void SingleProblemGetterModeler::printProblems() const
{
    const auto& master = modeler_->masterProblem();
    const auto& subproblems = modeler_->subproblems();
    printProblem(master, "Master problem");
    for (size_t i = 0; i < subproblems.size(); ++i)
    {
        printProblem(subproblems[i], ("Subproblem " + std::to_string(i + 1)).c_str());
    }
}
```
is easier to read
- **printProblems()** don't really **print** the problems (as it's done in the legacy side), it just **logs** the problems.
  The real printing (mps and structure file) is done in the constructor of **SingleProblemGetterModeler**, which is confusing.
  I suggest you **print** and **log** in the same **printProblems()** method.

- Very similar logs are already made by modeler when it builds the problems. We have a kind of duplication here. 
-  In this lambda, we check if the problem is not null before logging. After all what was done before coming in this code, we may not need this check, it should have been done before : null problem issues should have been taken care of sooner. If a problem is null, we're not supposed to come here.

src/api/main.cpp
```diff
@@ -14,7 +20,28 @@ int main(const int argc, const char** argv)
...
+    const fs::path antaresMarker = studyPath / "study.antares";
+    const fs::path modelerMarker = studyPath / "parameters.yml";
+
+    if (fs::exists(antaresMarker))
+    {
+        logs.info() << "Loading Antares study...";
+        const SingleProblemGetter getter(studyPath);
+        getter.printProblems();
+    }
+    else if (fs::exists(modelerMarker))
```
guilpier-code, line 34: In [ticket  ANT-4847](https://gopro-tickets.rte-france.com/browse/ANT-4847) are only specified **legacy** or full **GEMS** cases.
What if we are in the **hybrid** case ?
With this code, we will print only the **legacy** problems.

guilpier-code (commented):
Unfinished review, continues tomorrow.

src/modeler/modeler/Modeler.cpp
```diff
@@ -253,27 +251,28 @@ void Modeler::run()
       parameters_.lastTimeStep,  // global = local
       0};
 
-    // Sub problem
+    timeScenarioCtx_ = std::make_unique<FillContext>(timeScenarioCtx);
```
guilpier-code, line 254: This is a bit shorter : 
```c++
    timeScenarioCtx_ = std::make_unique<FillContext>(parameters_.firstTimeStep,
                                                     parameters_.lastTimeStep,
                                                     parameters_.firstTimeStep,
                                                     parameters_.lastTimeStep,
                                                     0);
```
Besides, if you move this instruction into the **Modeler**'s constructor (as it should IMO), then you'll be able to split **buildProblems()** into **buildMaster()** and **buildSubProblems()** (maybe private methods) and have : 
```c++
void Modeler::buildProblems()
{
    Utils::TimeMeasurement measure;
    logs.info() << "linear problem of System loaded";

    buildMaster();
    buildSubProblems();

    logs.info();
    logs.info() << "Modeler build took " << measure.toStringInSeconds();
}
```

src/api/main.cpp
payetvin: Legacy and hybrid are both handled by SingleProblemGetter as they must contain a study.antares

src/api/main.cpp
payetvin: It use to be one problem, It should indeed be renamed.

src/api/singleProblemGetterModelerImpl.cpp
guilpier-code: I can see that the 3 last points were left aside. 
Any particular reason ?

payetvin: Master problem isn't printed in modeler logs, and I think it's fine to have a final log even if it was logged before.

Name changed from print() to logSize()

flomnes (approved):

PR #3565, opened 2026-04-14
parquet part 2 : parquet for modeler

guilpier-code:
This PR is about [ticket ANT-4633](https://gopro-tickets.rte-france.com/browse/ANT-4633).

As we choose to break **ticket ANT-4633** into several parts, this is **part 2**.

What was done : 
- what's required by **ticket ANT-4633** on **modeler** side : 
  - we need a modeler cmd line option to force parquet for simulation table print
  - Columns of simulation table hold elements of various type (a given column is homogeneous regarding the type of its element). So we need to store each column of simulation table in a different way in an **arrow** table before printing it. 
- refactor to allow previous changes and for code more simple and more clear. For a discussion (in french) about what was improved in the code, see file : 
  **docs/developer-guide/comments-on-parquet-subject.txt** (this is intended to be removed before closing the **ticket 4633**)


src/io/outputs/include/antares/io/outputs/ISimulationTable.h
```diff
@@ -1,20 +0,0 @@
-// Copyright 2007-2026, RTE (https://www.rte-france.com)
```
guilpier-code, line 1: This class was useless (we don't do polymorphism with simulation table), so it was removed.

src/io/outputs/include/antares/io/outputs/columns.h
```diff
@@ -14,11 +15,30 @@ namespace Antares::IO::Outputs
 class IColumn
 {
 public:
+    IColumn(const std::string name):
```
guilpier-code, line 18: Giving a name to columns simplifies the code and avoids having an external vector of names for columns, side by side with a vector of nameless columns.
That's we made this change.

src/io/outputs/include/antares/io/outputs/SimulationTable.h
```diff
@@ -0,0 +1,37 @@
+// Copyright 2007-2026, RTE (https://www.rte-france.com)
```
guilpier-code, line 1: old class **SimulationTableCsv** was renamed into **SimulationTable** : it's just a  simulation table content, without choosing the writing format (csv / parquet)

src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -1,4 +1,6 @@
 #pragma once
+#include <algorithm>
```
guilpier-code, line 2: Storage is a column-based storage for simulation table.
Changes here are the consequences of giving each column a name (as a member data). See [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3565/changes#r3080362381)

src/io/outputs/CMakeLists.txt
```diff
@@ -1,38 +1,59 @@
-set(SOURCES
-        SimulationTableCsv.cpp
-        SimulationTableCsvFile.cpp
+# ===================================
```
guilpier-code, line 1: here we needed to split old target **modeler-simulation-table** into 2 new targets : 
- **simulation-table**
- **simu-table-generator**

for many reasons : 
- any new target needing to build upon **simulation table** don't need to be linked with code that generates the **simulation target**.
- it's more clear and more flexible

src/io/outputs/SimulationTable.cpp
```diff
@@ -1,13 +1,13 @@
 // Copyright 2007-2026, RTE (https://www.rte-france.com)
 // SPDX-License-Identifier: MPL-2.0
 
-#include "antares/io/outputs/SimulationTableCsv.h"
+#include "antares/io/outputs/SimulationTable.h"
```
guilpier-code, line 4: Here some renaming was done for more accuracy.

Note that many free functions are related to csv writing format, which not good : a **csv writer** (for simulation tables) already exists, and that's the place where this code should be moved.

src/libs/antares/study/include/antares/study/parameters.h
```diff
@@ -420,6 +420,9 @@ class Parameters final
     bool noOutput = false;
     //@}
 
+    // In case we print simulation tables, do we print it in csv or parquet ?
+    bool parquetFmtForSimuTables = false;
```
guilpier-code, line 430: adding a new legacy solver option for parquet writing.

src/libs/antares/writer/include/antares/writer/i_table_writer.h
```diff
@@ -8,20 +8,22 @@
 #include <string>
 #include <vector>
 
+#include "antares/io/outputs/SimulationTable.h"
+
 namespace Antares::Writer
 {
 
 class ITableWriter
```
guilpier-code, line 16: **ITableWriter** intends to be the parent class for any writer that writes a **simulation table** (for now : csv / parquet).
That's why **writeTable(...)** only takes a reference to a **SimulationTable**

src/libs/antares/writer/include/antares/writer/table_format.h
```diff
@@ -1,14 +0,0 @@
-// Copyright 2007-2026, RTE (https://www.rte-france.com)
-// SPDX-License-Identifier: MPL-2.0
-
-#pragma once
-
-namespace Antares::Writer
-{
-// Table output formats
-enum class TableFormat
```
guilpier-code, line 9: We don't need a enum class for now, a boolean is enough.
But this enum class was very clear, we may want it back.

src/io/outputs/include/antares/io/outputs/SimulationTableCsvFile.h
```diff
@@ -1,23 +0,0 @@
...
-
-#pragma once
-#include <filesystem>
-#include <fstream>
-
-#include "SimulationTableCsv.h"
-
-namespace Antares::IO::Outputs
-{
-class SimulationTableCsvFile final: public SimulationTableCsv
```
guilpier-code, line 12: class **SimulationTableCsvFile** was a csv writer for simulation table.
it inherited from class **SimulationTableCsv**, which was the real simulation table class.

We chose to rearrange this in a more natural way : now we have **class SimulationTable** on one side, and we have writer for simulation tables (csv or parquet), and no more inheritance between to 2 : we only pass **SimulationTable** to **writers**.

src/io/outputs/SimulationTableCsvFile.cpp
```diff
@@ -1,44 +0,0 @@
-// Copyright 2007-2026, RTE (https://www.rte-france.com)
-// SPDX-License-Identifier: MPL-2.0
-
-#include "antares/io/outputs/SimulationTableCsvFile.h"
```
guilpier-code, line 4: see [this comment ](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3565/changes#r3080696707), it explains we removed this source file.

src/libs/antares/writer/CMakeLists.txt
```diff
@@ -5,50 +5,57 @@ include(GenerateExportHeader)
 find_package(Parquet REQUIRED)
 find_package(Arrow REQUIRED)
 
-add_library(parquet_writer SHARED
+add_library(simulation-table-writers SHARED
```
guilpier-code, line 8: old **parquet_writer** target was renamed into **simulation-table-writers**; much more accurate.
The parquet and the csv writers are embedded here.

src/modeler/modeler/Modeler.cpp
```diff
@@ -302,8 +328,19 @@ void Modeler::run()
     }
     if (data_.resolutionMode == ResolutionMode::SEQUENTIAL_SUBPROBLEMS)
     {
-        auto* solution = solveSubproblem();
-        writeSubProblemSimulationTable(solution, *subproblemOptimEntityContainer, timeScenarioCtx);
+        subProbSolution_ = solveSubproblem();
```
guilpier-code, line 323: Before this change, old function **writeSubProblemSimulationTable** was doing 3 different things (not exactly related to its name) : 
- checking the solution of LP (does solution exists ? Is problem feasible ? is solution optimal ?)
- make the simulation table from this solution
- write the simulation, and in the write format (csv / parquet)

This 3 tasks deserve 3 steps (meaning 3 function calls), not to be done in a single function with a misleading name.
This is what was done here.

src/modeler/main.cpp
```diff
@@ -14,7 +14,23 @@ using namespace Antares;
 static void usage()
 {
     std::cout << "Usage:\n"
-              << "antares-modeler <path/to/study>\n";
+              << "antares-modeler <path/to/study> [--parquet]\n";
+}
+
+bool parquetFormatRequired(int argc, const char** argv)
```
guilpier-code, line 21: Modeler options : here's a new option for writing simulation table in parquet format.

src/solver/misc/options.cpp
```diff
@@ -67,6 +67,10 @@ void addParameterOptions(Yuni::GetOpt::Parser& parser,
                    'z',
                    "zip-output",
                    "Force the write output into a single zip archive");
+    parser.addFlag(settings.parquetFmtForSimuTables,
```
guilpier-code, line 70: Here's a new cmd line option for legacy solver. Code is clear enough.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -546,41 +548,33 @@ void ISimulation<ImplementationType>::storeYearBuffers(uint year,
...
+
+        writerEntry = header + std::move(globalSecondBuffer);
         pResultWriter.addEntryFromBuffer("simulation_table--optim-nb-2.csv", writerEntry);
     }
-
     yearSimulationBuffers_.clear();
 }
-
-template<class ImplementationType>
-OptimisationsSimulationTable& ISimulation<ImplementationType>::getSimulationTable(uint numSpace)
```
guilpier-code, line 578: **getSimulationTable** function calls another **getSimulationTable** function. Too complicated, we don't need this nested calls. That's why this function was removed.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -546,41 +548,33 @@ void ISimulation<ImplementationType>::storeYearBuffers(uint year,
 template<class ImplementationType>
 void ISimulation<ImplementationType>::aggregateAndWriteSimulationTables()
 {
-    std::lock_guard lock(buffersMutex_);
```
guilpier-code, line 549: We don't need a lock guard here : we're not in a multi thread sequence. 
So we removed it.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -546,41 +548,33 @@ void ISimulation<ImplementationType>::storeYearBuffers(uint year,
...
-    std::string globalFirstBuffer;
-    std::string globalSecondBuffer;
-    // dans l'ordre des années
-    for (uint year = 0; year < study.parameters.nbYears; ++year)
```
guilpier-code, line 553: We don't need to make a loop over years and then find the year in a map, it's expensive and  complicated.
We just need a loop over the map keys (years) : as keys, years are already ordered by increasing value.

src/libs/antares/writer/columnToArrowAdapter.h
```diff
@@ -0,0 +1,130 @@
...
+
+// Arrow / Parquet
+#include <arrow/api.h>
+#include <arrow/io/api.h>
+
+#include "antares/io/outputs/columns.h"
+
+namespace Antares::Writer
+{
+
```
guilpier-code, line 14: Here we can find polymorphic classes.
**Indeed** : 
In simulation table, columns can be of different types, because they can contain int, double, strings, enum class, and also std::optional of all these previous types.
Each column of the simulation table is stored differently (depending on the type of element it contains), so they all need a specific adapter for this. 
that's why each of current adapter class is associated to a column type in simulation table.
Each of these adapter must provide a field (for arrow schema) and must be able to make and return an array to be put in the arrow table thta will be written in a **parquet** file.

That's why they all have a common (interface) parent class says, they all must implement methods **makeField()** and **makeArray()**

src/libs/antares/writer/csv_table_writer.h
```diff
@@ -16,9 +16,8 @@ namespace Antares::Writer
 class CsvTableWriter final: public ITableWriter
 {
 public:
-    void writeTable(const std::filesystem::path& filePath,
-                    const std::vector<std::string>& header,
-                    const std::vector<std::vector<std::string>>& rows) override;
```
guilpier-code, line 21: Class **CsvTableWriter::writeTable** no longer needs header vector or rows  made somewhere else and passed to it.
It's passed the simulation table, and extracts from it what it needs.
More natural like this.

src/libs/antares/writer/parquet_table_writer.cpp
```diff
@@ -30,123 +12,70 @@
...
+    output_file_.replace_extension(".parquet");
+}
 
-    auto st_write = parquet::arrow::WriteTable(*table,
-                                               arrow::default_memory_pool(),
-                                               outfile,
-                                               chunk_size,
-                                               props);
-    if (!st_write.ok())
+void ParquetTableWriter::writeTable(const SimulationTable& simuTable) const
```
guilpier-code, line 69: Now code for writing a table in parquet format is very simple and readable.
Note that it used to convert in column format a vector of rows, previously made from a simulation table organized in column format. Very complicated. 
We no longer do that, we directly extract columns from simulation table.

src/modeler/fileWriter/include/antares/solver/modeler/fileWriter/FileWriter.h
```diff
@@ -5,44 +5,23 @@
...
-{
-class ILinearProblem;
-class ILinearProblemData;
-class FillContext;
-} // namespace Antares::Optimisation::LinearProblemApi
-
-namespace Antares::Optimization
-{
-class VariableDictionary;
-}
```
guilpier-code, line 19: These forward declarations are useless, removed.

src/modeler/fileWriter/FileWriter.cpp
```diff
@@ -7,57 +7,51 @@
...
     return outputPath_;
 }
 
-void FileWriter::writeSimulationTable(
-  const Optimisation::LinearProblemApi::ILinearProblem& linearProblem,
-  const Optimisation::LinearProblemApi::IMipSolution& solution,
-  const ModelerData& modelerData,
-  const Optimisation::OptimEntityContainer& variableContainer,
-  const Optimisation::LinearProblemApi::FillContext& fillContext) const
+void FileWriter::writeSimulationTable(SimulationTable& SimulationTable) const
```
guilpier-code, line 41: **FileWriter::writeSimulationTable(...)** takes now only the simulation (as argument) and does exactly what its name suggests, no more, no less (as opposed to previously).

src/libs/antares/writer/table_writer_factory.cpp
```diff
@@ -24,20 +24,18 @@
 #include "csv_table_writer.h"
 #include "parquet_table_writer.h"
 
+namespace fs = std::filesystem;
+
 namespace Antares::Writer
 {
 
-ITableWriter::Ptr makeTableWriter(TableFormat fmt)
+ITableWriter::Ptr makeTableWriter(bool parquetFormatRequired, fs::path& filePath)
```
guilpier-code, line 32: Table writing format used to be an **enum**, and was replaced with a **boolean** (= parquet format or not).
Maybe not a good idea.

src/libs/antares/writer/i_table_writer.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+
+static void ensureParentDir(const std::filesystem::path& file)
+{
+    auto p = file.parent_path();
+    if (!p.empty())
+    {
+        std::error_code ec;
+        std::filesystem::create_directories(p, ec);
+        // Best effort; ignore error as writer will fail on open if directory truly invalid
```
flomnes, line 16: warning in case of error ?

flomnes (approved):

PR #3567, opened 2026-04-14
refacto: Clean up scenario-builder

flomnes:
We no longer save scenariobuilder.dat, so save functions can be removed safely (dead code).

- `hydroLevelsData::addToPrefix_` no longer useful
- `ShortTermAdditionalConstraintsTSNumberData::saveToINIFile` removed
- `ShortTermInflowsTSNumberData::saveToINIFile` removed


YoshuaLisasi:
SonarCloud already flagged this, but just to confirm: this constructor takes a single parameter, so it should be marked `explicit`. 
Without it, the compiler is allowed to implicitly convert a `std::function<void(Study&, MatrixType&)>` into a `hydroLevelsData` object, which is almost certainly not intended.

Simple fix:
```cpp
explicit hydroLevelsData(std::function<void(Study&, MatrixType&)> applyToTarget);
```

YoshuaLisasi (approved):

PR #3579, opened 2026-04-20
Pack minmax data to improve synthesis RAM

payetvin:


JasonMarechal25 (approved):

src/solver/variable/include/antares/solver/variable/storage/minmax-data.h
```diff
@@ -13,7 +13,7 @@ namespace Antares::Solver::Variable::R::AllYears
 class MinMaxData
 {
 public:
-    struct Data
+    struct __attribute__((packed)) Data
```
flomnes, line 16:
```cpp
#pragma pack(1)
```
may be more portable (supported by MSVC, clang, gcc)

> #pragma pack is a Microsoft syntax that has been [ported to GCC](https://gcc.gnu.org/onlinedocs/gcc/Structure-Layout-Pragmas.html) for compatibility reasons.
> __attribute__((aligned)) is a GCC-specific syntax (unsupported by MSVC).

[source](https://stackoverflow.com/a/40043130)

flomnes:
Hi @payetvin  👋

Thanks for this PR! Here are a few thoughts:

---

## 🟡 Memory savings may not justify the cost

The `packed` attribute reduces `sizeof(Data)` from 16 to 12 bytes (25% reduction), but the actual RAM savings depend on how many `Data` instances exist in a real study. More importantly, unaligned `double` accesses can be slower on x86 and may even fault on ARM. I'd recommend profiling the actual memory footprint and performance before and after to confirm the trade-off is worth it.

💡 **Suggestion**: Consider a benchmark showing the real RAM savings and any performance impact.

---

## 🔴 Unaligned memory access risk

`__attribute__((packed))` removes struct padding, meaning `double` may be accessed at a non-8-byte-aligned offset. On x86 this works but is slower; on ARM it can cause a hardware fault.

💡 **Fix**: If packing is strictly needed for I/O, use a separate packed struct for serialization and a naturally-aligned one for computation. Or better yet, avoid packing altogether.

---

## 🔴 Portability concern

`__attribute__((packed))` is GCC/Clang-specific and won't compile on MSVC. If the project targets multiple compilers, this could be a problem.

💡 **Fix**: Use a cross-platform macro like:
```cpp
#if defined(_MSC_VER)
    #define PACKED [[msvc::pack(1)]]
#else
    #define PACKED __attribute__((packed))
#endif
```
Or, ideally, avoid packing entirely.

---

## 🟢 SoA layout could be a better long-term fix

The class already uses `std::vector<Data>` per time resolution. Consider splitting into `std::vector<double> values` and `std::vector<uint32_t> indices` (Structure of Arrays). This gives:

- ✅ Natural alignment (no `packed` needed)
- ✅ Better cache locality for vectorized operations
- ✅ Potentially even better memory savings

---

## 🟡 Minor: add documentation

The `Data` struct is public but undocumented. A brief comment explaining *why* packing is needed would help future readers.

---

### Summary

Small, targeted change with a clear intent — nice work on the focus! 🎯 The main concerns are around **portability** and **potential performance regressions from unaligned accesses**. If the team is confident in an x86-only deployment target and the benchmark confirms the benefit, this is a low-risk change. Otherwise, the SoA refactor might be the cleaner long-term solution.

Happy to discuss! 😊

payetvin:
done in #3588

PR #3580, opened 2026-04-20
ANT-4899 - Not loading binding constraints when enabled is set to false in INI file

YoshuaLisasi:
### Refactor BindingConstraintLoader::load and add early return when disabled
**Context**

The initial goal was to add an early return when enabled is set to false in the INI file, so that we skip time series loading entirely for disabled constraints. Since the load function was doing too many things in a single method, I took the opportunity to refactor it into smaller, focused functions.
Changes

_populateConstraint_ — Parses all key-value pairs from the INI section and fills the corresponding fields of the BindingConstraint.
_parseWeightAndOffset_ — Resolves link or cluster references and assigns their weight and offset to the constraint.
_validate_ — Checks that all required fields (name, id, type, operator) are set and valid.
_loadByOperator_ — Loads time series and builds the final constraint(s) depending on the operator type (single or both).
Early return — If pEnabled is false (either from the INI file or because no weights were found), load now returns an empty vector before reaching the time series loading step.

### DIFF
--- a/src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
+++ b/src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
@@ -43,9 +43,11 @@ std::vector<std::shared_ptr<BindingConstraint>> BindingConstraintLoader::load(En
 
     if (!bc->pEnabled)
     {
-        return {};
+        /// This BC won't be used, return it without loading time series
+        logs.info() << "DEBUG: BC " << bc->name() << " is disabled, skipping time series";
+        return {bc};
     }
-
+    logs.info() << "DEBUG: BC " << bc->name() << " is enabled, loading time series";
     return loadByOperator(env, bc);
 }
 
@@ -190,7 +192,7 @@ bool BindingConstraintLoader::loadTimeSeriesLegacyStudies(
     return false;
 }
 
-void BindingConstraintLoader::populateConstraint(EnvForLoading& env,
+void BindingConstraintLoader::populateConstraint(const EnvForLoading& env,
                                                  std::shared_ptr<BindingConstraint>& bc)
 {
     // Foreach property in the section...
@@ -251,7 +253,7 @@ void BindingConstraintLoader::populateConstraint(EnvForLoading& env,
     }
 }
 
-void BindingConstraintLoader::parseWeightAndOffset(EnvForLoading& env,
+void BindingConstraintLoader::parseWeightAndOffset(const EnvForLoading& env,
                                                    const IniFile::Property* p,
                                                    std::shared_ptr<BindingConstraint>& bc)
 {
@@ -323,7 +325,7 @@ void BindingConstraintLoader::parseWeightAndOffset(EnvForLoading& env,
     }
 }
 
-bool BindingConstraintLoader::validate(EnvForLoading& env,
+bool BindingConstraintLoader::validate(const EnvForLoading& env,
                                        const std::shared_ptr<BindingConstraint>& bc)
 {
     if (bc->pName.empty())
@@ -371,17 +373,18 @@ std::vector<std::shared_ptr<BindingConstraint>> BindingConstraintLoader::loadByO
     }
     case BindingConstraint::opBoth:
     {
-        auto greater_bc = std::make_shared<BindingConstraint>();
-        greater_bc->copyFrom(bc.get());
-        greater_bc->name(bc->name() + "_sup");
-        greater_bc->pID = bc->pID;
-        greater_bc->operatorType(BindingConstraint::opGreater);
-        bc->name(bc->name() + "_inf");
+        auto greaterBc = std::make_shared<BindingConstraint>();
+        const auto originalName = bc->name();
+        greaterBc->copyFrom(bc.get());
+        greaterBc->name(originalName + "_sup");
+        greaterBc->pID = bc->pID;
+        greaterBc->operatorType(BindingConstraint::opGreater);
+        bc->name(originalName + "_inf");
         bc->operatorType(BindingConstraint::opLess);
 
-        if (loadTimeSeries(env, bc.get()) && loadTimeSeries(env, greater_bc.get()))
+        if (loadTimeSeries(env, bc.get()) && loadTimeSeries(env, greaterBc.get()))
         {
-            return {bc, greater_bc};
+            return {bc, greaterBc};
         }
         break;
     }
--- a/src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
+++ b/src/libs/antares/study/binding_constraint/BindingConstraintLoader.cpp
@@ -27,152 +27,11 @@ std::vector<std::shared_ptr<BindingConstraint>> BindingConstraintLoader::load(En
     auto bc = std::make_shared<BindingConstraint>();
     bc->clear();
 
-    // Foreach property in the section...
-    for (const IniFile::Property* p = env.section->firstProperty; p; p = p->next)
-    {
-        if (p->key.empty())
-        {
-            continue;
-        }
-
-        if (p->key == "name")
-        {
-            bc->pName = p->value;
-            continue;
-        }
-        if (p->key == "id")
-        {
-            bc->pID = p->value;
-            boost::to_lower(bc->pID);
-            continue;
-        }
-        if (p->key == "enabled")
-        {
-            bc->pEnabled = p->value.to<bool>();
-            continue;
-        }
-        if (p->key == "type")
-        {
-            bc->pType = BindingConstraint::StringToType(p->value);
-            continue;
-        }
-        if (p->key == "operator")
-        {
-            bc->pOperator = BindingConstraint::StringToOperator(p->value);
-            continue;
-        }
-        if (p->key == "filter-year-by-year")
-        {
-            bc->pFilterYearByYear = stringIntoDatePrecision(p->value);
-            continue;
-        }
-        if (p->key == "filter-synthesis")
-        {
-            bc->pFilterSynthesis = stringIntoDatePrecision(p->value);
-            continue;
-        }
-        if (p->key == "comments")
-        {
-            bc->pComments = p->value;
-            continue;
-        }
-        if (p->key == "group")
-        {
-            bc->group_ = p->value.c_str();
-            continue;
-        }
-
-        // initialize the values
-        double w = .0;
-        int o = 0;
-
-        // Separate the value
-        if (auto setKey = p->key.find('%'); setKey != 0 && setKey != String::npos) // It is a link
-        {
-            if (bool ret = SeparateValue(env, p, w, o); !ret)
-            {
-                continue;
-            }
-
-            const AreaLink* lnk = env.areaList.findLinkFromINIKey(p->key);
-            if (!lnk)
-            {
-                logs.error() << env.iniFilename << ": in [" << env.section->name << "]: `" << p->key
-                             << "`: link not found";
-                continue;
-            }
-            if (!Utils::isZero(w))
-            {
-                bc->weight(lnk, w);
-            }
-
-            if (!Utils::isZero(o))
-            {
-                bc->offset(lnk, o);
-            }
-
-            continue;
-        }
-        else // It must be a cluster
-        {
-            // Separate the key
-            setKey = p->key.find('.');
-            if (0 == setKey || setKey == String::npos)
-            {
-                logs.error() << env.iniFilename << ": in [" << env.section->name << "]: `" << p->key
-                             << "`: invalid key";
-                continue;
-            }
-
-            if (bool ret = SeparateValue(env, p, w, o); !ret)
-            {
-                continue;
-            }
-
-            const ThermalCluster* clstr = env.areaList.findClusterFromINIKey(p->key);
-            if (!clstr)
-            {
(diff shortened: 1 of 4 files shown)


flomnes:
## 1. Summary

This PR refactors `BindingConstraintLoader::load()` — a ~180-line monolithic function — into four smaller, focused helper methods:

| Method | Responsibility |
|---|---|
| `populateConstraint()` | Iterates INI properties and fills `BindingConstraint` fields |
| `parseWeightAndOffset()` | Resolves link/cluster references and assigns weight/offset |
| `validate()` | Checks required fields (name, id, type, operator) |
| `loadByOperator()` | Loads time series and builds final constraint(s) |

It also adds an early return when `pEnabled` is `false`, skipping time series loading entirely for disabled constraints.

**Verdict: Overall positive.** The refactoring is a clear improvement in readability and maintainability. Several issues remain that should be addressed before merge.

---

## 2. Strengths

### 2.1 Meaningful Method Extraction
The four new methods each have a single, well-defined responsibility. This is textbook [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) and makes the code significantly easier to reason about.

### 2.2 Early Return for Disabled Constraints
The new early return at the top of the `load()` flow (after `validate()`) is the core bug fix and is placed correctly. Disabled constraints no longer waste resources attempting to load time series files that may not exist.

### 2.3 `validate()` Returns `bool`
Extracting validation into a method that returns `bool` is a good design choice — it enables the caller to branch on success/failure cleanly.

### 2.4 `loadByOperator()` Preserves Original Logic
The `opBoth` case (which splits one constraint into two: `_inf` and `_sup`) is preserved exactly as before, which is important for correctness.

---

## 3. Issues

### 3.1 🔴 Critical — `load()` Passes `EnvForLoading` by Value, But New Methods Take Non-const Lvalue References

**Severity: High**

The `load()` method takes `EnvForLoading env` by **value** (it is move-constructed at the call site in `BindingConstraintsRepository::LoadBindingConstraint`). All four new helper methods take `EnvForLoading& env` as a non-const lvalue reference.

This is **syntactically fine** (an lvalue can bind to an lvalue reference), but it creates a subtle API inconsistency: the new methods *could* modify `env`, and there is no indication from their signatures whether they do. Consider:

- `populateConstraint()` — does not modify `env` → should take `const EnvForLoading&`
- `parseWeightAndOffset()` — does not modify `env` → should take `const EnvForLoading&`
- `validate()` — does not modify `env` → should take `const EnvForLoading&`
- `loadByOperator()` — calls `loadTimeSeries(env, ...)` which takes `EnvForLoading&` → needs non-const

**Fix:** Mark the first three as `const EnvForLoading& env`. This documents intent and prevents accidental mutation.

### 3.2 🔴 Critical — `parseWeightAndOffset()` Has Silent Failure via `return` Instead of `continue`

**Severity: High**

In the original code, when `SeparateValue()` failed, the loop used `continue` to skip to the next INI property. In the refactored `parseWeightAndOffset()`, the same failure path uses `return`, which exits the helper function and returns control to `populateConstraint()`.

While this **appears to work** (the loop in `populateConstraint()` continues to the next property), the behavior is subtly different:

- **Original:** `continue` in the loop body → next iteration of the `for` loop in `populateConstraint()`
- **Refactored:** `return` from `parseWeightAndOffset()` → falls through to the next iteration of the `for` loop in `populateConstraint()`

The end result is the same, but the control flow is less obvious. A reader scanning `parseWeightAndOffset()` might not immediately realize that `return` is being used as a loop-continuation mechanism.

**Recommendation:** Either:
1. Keep the `continue` pattern by returning a `bool` from `parseWeightAndOffset()` and using `if (!parseWeightAndOffset(...)) continue;` in the caller, or
2. Add a comment at the top of `parseWeightAndOffset()` explaining that `return` is used to skip to the next property.

### 3.3 🟡 Medium — `loadByOperator()` Mutates `bc` In-Place for the `opBoth` Case

**Severity: Medium**

In the `opBoth` branch:

```cpp
auto greater_bc = std::make_shared<BindingConstraint>();
greater_bc->copyFrom(bc.get());
greater_bc->name(bc->name() + "_sup");   // reads bc->name()
greater_bc->pID = bc->pID;
greater_bc->operatorType(BindingConstraint::opGreater);
bc->name(bc->name() + "_inf");            // mutates bc->name()
bc->operatorType(BindingConstraint::opLess);
```

The `greater_bc` is created from a copy of `bc`, then `bc` is **mutated in place** (name changed to `"_inf"`, operator to `opLess`). This is a source of confusion:

- `greater_bc` gets the **original** name + `"_sup"` suffix
- `bc` gets the **original** name + `"_inf"` suffix

But `greater_bc->name()` was set using `bc->name()` **before** the mutation. If `bc->name()` were called again after the mutation, it would return a different value. This is fragile.

**Fix:** Capture the original name in a local variable before any mutation:

```cpp
const auto originalName = bc->name();
auto greater_bc = std::make_shared<BindingConstraint>();
greater_bc->copyFrom(bc.get());
greater_bc->name(originalName + "_sup");
greater_bc->pID = bc->pID;
greater_bc->operatorType(BindingConstraint::opGreater);
bc->name(originalName + "_inf");
bc->operatorType(BindingConstraint::opLess);
```

### 3.4 🟡 Medium — `populateConstraint()` Falls Through to `parseWeightAndOffset()` for Unknown Keys

**Severity: Medium**

In the original code, any property that didn't match a known key fell through to the weight/offset parsing block (the `if (p->key.find('%'))` / `else` block). In the refactored version, unknown keys fall through to `parseWeightAndOffset()`, which will:

1. Try `find('%')` — if not found, fall to the `else` branch
2. Try `find('.')` — if not found, log an error
3. Try `SeparateValue()` — which will also likely fail

This means **unknown INI keys produce an error log**. This is arguably correct behavior (catch typos in INI files), but the original code had the same behavior, so this is not a regression. However, it's worth noting that the refactoring makes this error path less obvious — a reader might not realize that `parseWeightAndOffset()` is also the "unknown key handler."

**Recommendation:** Add a comment in `populateConstraint()` noting that unrecognized keys are delegated to `parseWeightAndOffset()` for weight/offset parsing, and will produce an error if they don't match either pattern.

### 3.5 🟡 Medium — `validate()` Logs Errors But Doesn't Include Constraint ID

**Severity: Medium**

Each validation error logs `env.iniFilename` and `env.section->name`, but **not** `bc->pID`. When a study has many binding constraints, it can be hard to identify which specific constraint failed validation. The original code had the same issue, but this is an opportunity to improve.

**Fix:** Add `bc->pID` (or `bc->name()`) to the error messages:

```cpp
if (bc->pName.empty())
{
    logs.error() << env.iniFilename << ": in [" << env.section->name
                 << "] (id: " << bc->pID << "): Invalid binding constraint name";
    return false;
}
```

### 3.6 🟢 Low — `parseWeightAndOffset()` Takes `const IniFile::Property* p` But Could Be `const auto&`

**Severity: Low**

The parameter `const IniFile::Property* p` is fine, but the original code used `p->key` and `p->value` extensively. Since `IniFile::Property` is a legacy struct, this is acceptable. No action needed, but worth noting for consistency.

### 3.7 🟢 Low — `populateConstraint()` and `parseWeightAndOffset()` Are Not `static`

**Severity: Low**

`SeparateValue()` is declared `static` in the header. The new methods (`populateConstraint`, `parseWeightAndOffset`, `validate`, `loadByOperator`) are **not** `static`. Since they don't access any non-static member state of `BindingConstraintLoader`, they could be `static`.

This is a minor point — the class is `final` and has no members, so the distinction is moot at runtime. But for API clarity, making them `static` would signal that they are utility functions.

### 3.8 🟢 Low — Missing Unit Tests

**Severity: Low (for new code)**

SonarQube reports 52.2% coverage on new code (below the 60% threshold). There are **no unit tests** for `BindingConstraintLoader` anywhere in the test suite. The refactoring touches ~210 new lines and should have corresponding tests.

**Recommendation:** Add unit tests covering:
1. `populateConstraint()` — valid properties, unknown keys
2. `parseWeightAndOffset()` — valid link, valid cluster, invalid key, missing separator
3. `validate()` — empty name, empty id, unknown type, unknown operator
4. `loadByOperator()` — `opBoth` splitting behavior
5. `load()` — early return when `pEnabled` is false

---

## 4. Style & Conventions

### 4.1 ✅ Brace Style
All methods follow the Allman brace style (braces on new lines). Consistent with project conventions.

### 4.2 ✅ Naming Conventions
Method names follow `lowerCamelCase`. Good.

### 4.3 ✅ Include Ordering
The new methods don't introduce new includes. The existing includes are correctly ordered.

### 4.4 ⚠️ `using namespace Yuni;`
The file still has `using namespace Yuni;` at namespace scope. This is a legacy pattern. The PR doesn't introduce new Yuni usage, so this is not a regression, but it's worth noting that new code should avoid YNI types where possible per the C++ guidelines.

### 4.5 ✅ Error Handling
Error messages use `logs.error()` consistently. No use of `logs.fatal()`. Good.

---

## 5. Potential Bugs

### 5.1 `opBoth` Name Mutation Race Condition (Revisited)

In `loadByOperator()`, the `opBoth` case modifies `bc` in place after creating `greater_bc`. If `loadTimeSeries()` fails for `bc`, the function returns `{}` and `bc` is left in a modified state (name suffixed with `_inf`, operator set to `opLess`). Since `bc` is a `shared_ptr`, if any other code holds a reference to it, that reference will see the mutated state.

In practice, `bc` is created fresh in `load()` and is not shared elsewhere, so this is unlikely to cause a real bug. But it's a code smell.

### 5.2 `pEnabled` Set to `false` When No Weights Found — Then Early Return

The flow is:
1. Parse INI → `bc->pEnabled` may be `true` from INI
2. If no weights found → `bc->pEnabled = false`
3. Validate → passes
4. **Early return** if `!bc->pEnabled`

This is correct behavior. However, the log message is absent — when a constraint is silently skipped because no weights were found, there's no log entry. Consider adding an info-level log:

```cpp
if (!bc->pEnabled)
{
    logs.info() << "Skipping disabled binding constraint `" << bc->name() << "`";
    return {};
}
```

---

## 6. Recommendations Summary

| Priority | Issue | Action |
|---|---|---|
| 🔴 High | 3.1 | Add `const` to `EnvForLoading&` in `populateConstraint`, `parseWeightAndOffset`, `validate` |
| 🔴 High | 3.2 | Clarify control flow in `parseWeightAndOffset()` (comment or return bool) |
| 🟡 Medium | 3.3 | Capture original name before mutation in `opBoth` branch |
| 🟡 Medium | 3.5 | Include constraint ID in validation error messages |
| 🟡 Medium | 3.8 | Add unit tests to meet SonarQube coverage threshold |
| 🟢 Low | 3.7 | Consider making helper methods `static` |
| 🟢 Low | 5.2 | Add log when skipping disabled constraint |

---

## 7. Overall Assessment

**This is a good refactoring.** The extraction of `validate()` and `loadByOperator()` is particularly clean. The early return for disabled constraints is the right fix for the underlying bug.

The main concerns are:
1. **API consistency** — `const` correctness on the `EnvForLoading` parameter
2. **Test coverage** — SonarQube is failing, and there are no tests at all
3. **`opBoth` mutation** — fragile in-place modification of `bc`

Addressing items 3.1, 3.3, and 3.8 would make this PR merge-ready. The rest are nice-to-have improvements.

**Recommendation: Approve with conditions.** Request changes for the `const` correctness and test coverage issues; the rest can be addressed in follow-up PRs.

flomnes:
## Review Feedback Follow-up

Thank you for your efforts on this PR, @YoshuaLisasi ! The refactoring is a clear improvement and the recent commit addresses several key concerns.

### ✅ Addressed in latest commit

| Review Issue | What Changed |
|---|---|
| **3.1 🔴** — `const` correctness on `EnvForLoading&` | ✅ All three methods now take `const EnvForLoading& env`: `populateConstraint`, `parseWeightAndOffset`, `validate` |
| **3.3 🟡** — `opBoth` name mutation | ✅ Captured `originalName` before mutation, renamed `greater_bc` → `greaterBc` (better naming) |
| **3.8 🟢** — Test coverage | ✅ Tests updated: `enabled = false` → `enabled = true`, added `AreaAddLinkBetweenAreas` calls, added time series file writing |

### ❌ Not Addressed

| Review Issue | Status |
|---|---|
| **3.2 🔴** — Control flow clarity in `parseWeightAndOffset()` | ❌ No comment added, no change to the `return` pattern |
| **3.5 🟡** — Include constraint ID in validation errors | ❌ No change to error messages |
| **3.7 🟢** — Make helper methods `static` | ❌ No change |
| **5.2 🟢** — Log when skipping disabled constraint | ❌ Added a `DEBUG` log but returns `{bc}` instead of `{}` — behavioral change worth confirming |

### ⚠️ Behavioral Change to Confirm

In `load()`, the disabled constraint path now does:
```cpp
logs.debug() << "BC " << bc->name() << " is disabled, skipping time series";
return {bc};  // returns the constraint, does not skip it
```

The original code returned `{}` (empty vector), effectively discarding disabled constraints. Now it returns `{bc}`, meaning disabled constraints are still loaded and included. This could be the **intended fix** (disabled = skip time series, but keep the constraint), but it is a behavioral change worth confirming.

### Overall

The two critical issues (3.1 and 3.3) and the test coverage (3.8) are addressed. The remaining items are lower priority. The PR is closer to merge-ready but the behavioral change in the disabled path should be verified.

Thanks again for the great work!

flomnes (approved):

PR #3586, opened 2026-04-21
parquet part 3 : parquet for legacy solver

guilpier-code:
This PR is about [ticket ANT-4633](https://gopro-tickets.rte-france.com/browse/ANT-4633).

As we choose to break **ticket ANT-4633** into several parts, this is **part 3**.
This PR intends to write simulation tables for legacy solver.

**Please note** : list of what was done can be found below, but some github comments were left to ease review.
**Caution** : many review comments made here are addressed in this [PR 3657](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3657).

What was done :
- [x] remove the aggregation of all years' simulation tables into only 2 blocks (one per optimization), as well as printing them into 2 **CSV** files only.  Instead, make each year print its own simulation table.
- [x] class **OptimisationsSimulationTable** was (physically) moved to where it more likely belongs (next to class **SimulationTable**)
- [x] class **OptimisationsSimulationTable** was very much simplified : it now only contains data (could now be simple **struct**) and the writing part was transferred to class **LegacySimulationTablesWriter**.
- [x] New class **LegacySimulationTablesWriter** now uses intensively  the table writers developed on modeler side (in [PR 3565](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3565)), instead of keeping old code from class **OptimisationsSimulationTable**

What remains to be done (if item is checked, then it's done) : 
- [x] continue simplifying class **SimulationTable** (especially there is code duplication between class **SimulationTable** [writeToBuffer(), buffer(), buffer_] and class **CsvTableWriter** [make_line(...), ...])
- [x] We removed unit tests on printing simulation tables into CSV format. They may be replaced by tests printing into files. We could even add more tests for thorough checks.
- [x] **OptimisationsSimulationTable** could be renamed into **LegacySimulationTables**, maybe more clear.
- [x] Consider removing class **FileWriter**. Indeed, it does little now, and calls another polymorphic writer (csv or parquet). So what  **FileWriter** does could be done in the parent (abstract) class of the polymorphic writer.
- [x] Cucumber scripts :  I made changes in class **solver_input_handler** and **solver_output_handler**, in the name of the simu table files name. are this files useful ? please check.
- [x] target **parquet_writer** was reintroduced by a merge from develop, it used to be named **simulation-table-writers**, much better (more accurate). Let's rename this target back.
- [ ] _(maybe fixed with a fresh clone)_ in debug (at least on Win), **zlibd.lib** not copied into build folder at configuration step. And dll like **parquet**, **arrows** + dependencies are not copied next to .exe needing them (even if they exist somewhere in the build folder after **cmake** config step). 
- [ ] what else ?


src/io/outputs/include/antares/io/outputs/OptimisationsSimulationTable.h
```diff
@@ -10,23 +10,15 @@ namespace Antares::Solver
 class IResultWriter;
 }
 
+// gp : no namespace here. We should ass one (Antares::IO::Outputs).
+
 class OptimisationsSimulationTable
```
guilpier-code, line 10: As you can see, this class was : 
- moved to another location (next to class **SimulationTable**)
- simplified : indeed another class (**LegacySimulationTablesWriter**) takes care of the writing part

src/io/outputs/CMakeLists.txt
```diff
@@ -3,6 +3,8 @@
 # ===================================
 set(SRC_SIMU_TABLE
         SimulationTable.cpp
+        OptimisationsSimulationTable.cpp
```
guilpier-code, line 6: class **OptimisationsSimulationTable** was moved here

src/io/outputs/OptimisationsSimulationTable.cpp
```diff
@@ -0,0 +1,16 @@
+// Copyright 2007-2026, RTE (https://www.rte-france.com)
+// SPDX-License-Identifier: MPL-2.0
+
+#include "include/antares/io/outputs/OptimisationsSimulationTable.h"
```
guilpier-code, line 4: this class is now so short that it could now be a struct

src/libs/antares/writer/include/antares/writer/LegacySimulationTablesWriter.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include <filesystem>
+
+#include "antares/io/outputs/OptimisationsSimulationTable.h"
+#include "antares/io/outputs/SimulationTable.h"
+#include "i_table_writer.h"
+
+namespace Antares::Writer
+{
+class LegacySimulationTablesWriter
```
guilpier-code: This is the class that writes the 2 simulation tables contained in class **OptimisationsSimulationTable** (for a given year, one per optimization).

src/libs/antares/writer/LegacySimulationTablesWriter.cpp
```diff
@@ -0,0 +1,53 @@
...
+    writeForOptim(tables.firstOptimSimulationTable(), 1);
+    writeForOptim(tables.secondOptimSimulationTable(), 2);
+}
+
+void LegacySimulationTablesWriter::writeForOptim(const SimulationTable* table,
+                                                 unsigned optim_number)
+{
+    auto filepath = makeSimuTableFilePath(folder_, year_, optim_number);
+    ITableWriter::Ptr writer = makeTableWriter(parquetFormat_, filepath);
+    writer->writeTable(*table);
```
guilpier-code, line 53: This is where we actually write one the the 2 legacy simulation tables.
call to **makeTableWriter()** sets the king of **writer** we want (CSV or parquet).
Then the **writer** writes.

src/solver/optimisation/include/antares/solver/optimisation/weekly_optimization.h
```diff
@@ -18,15 +17,16 @@ class WeeklyOptimization final
...
     void solve();
+    OptimisationsSimulationTable* simulationTables();
 
 private:
     Antares::Solver::Optimization::OptimizationOptions options_;
     PROBLEME_HEBDO* const problemeHebdo_ = nullptr;
     IResultWriter& writer_;
     std::reference_wrapper<Simulation::ISimulationObserver> simulationObserver_;
-    OptimisationsSimulationTable* simulationTables_ = nullptr;
+    std::shared_ptr<OptimisationsSimulationTable> simulationTables_;
```
guilpier-code: here I tried an unique smart pointer, but for reason, it does not compile.
I'll retry another time

src/solver/optimisation/include/antares/solver/optimisation/weekly_optimization.h
```diff
@@ -4,10 +4,9 @@
 #pragma once
 
 #include "antares/solver/simulation/ISimulationObserver.h"
+#include "antares/io/outputs/OptimisationsSimulationTable.h"
```
guilpier-code, line 8: Changes here are consequences of the fact that, now, we have a pair of simulation tables per year, and we no longer need to keep track of every year simulation tables.

src/solver/simulation/include/antares/solver/simulation/solver.hxx
```diff
@@ -155,15 +155,6 @@ public:
                                              pDurationCollector,
                                              scratchmap);
 
-        if (!study.parameters.noOutput)
```
guilpier-code, line 158: All the removed code in this file was related to the fact that we used to write only a pair of simulation tables, forcing us to keep in memory the content of every year's simulation table.
This would make consumed RAM grow dangerously, and is not acceptable.

src/solver/simulation/adequacy.cpp
```diff
@@ -6,19 +6,21 @@
 #include <antares/exception/AssertionError.hpp>
 #include <antares/exception/UnfeasibleProblemError.hpp>
 #include "antares/solver/simulation/solver_utils.h"
+#include "antares/writer/LegacySimulationTablesWriter.h"
```
guilpier-code, line 10: Writing a pair of simulation tables per year makes code more simple (shorter).

src/tests/src/io/outputs/testSimulationTable.cpp
```diff
@@ -210,49 +212,13 @@ BOOST_AUTO_TEST_CASE(MultipleEntries)
     BOOST_CHECK_EQUAL(lineCount, numEntries);
 }
 
-BOOST_AUTO_TEST_CASE(MultipleWriteCalls_AccumulateData)
```
guilpier-code, line 215: Here is a unit test file.
As class **OptimisationsSimulationTable**  was simplified, some tests were either changed, and more likely removed. See below for other comments.

src/tests/src/io/outputs/testSimulationTable.cpp
```diff
@@ -210,49 +212,13 @@ BOOST_AUTO_TEST_CASE(MultipleEntries)
...
-    SimulationTableEntry entry1{.block = 1,
-                                .component = "comp1",
-                                .output = "var1",
-                                .absolute_time_index = 1,
-                                .block_time_index = 1,
-                                .scenario_index = 0,
-                                .value = 10.0,
-                                .status = MipBasisStatus::BASIC};
-    tables.firstOptimSimulationTable()->addEntry(entry1);
-    tables.write();
```
guilpier-code, line 229: We no longer write the legacy **OptimisationsSimulationTable** like that in a buffer.
Instead we use class **LegacySimulationTablesWriter**, which is tested below (see test named **WriteTo_CreatesCorrectFiles**)

flomnes (commented):
Code Review: feature/parquet-for-legacy-solver ==> **See [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3586#issuecomment-4378614701) for an answer**

### 🟡 Bug: Typo in error message

**File:** `LegacySimulationTablesWriter.cpp`

```cpp
std::string msg = "Output folder " + folder.string() + "is does not exist.";
```

Missing space: `"is does not exist."` → `" is does not exist."` (should be `"does not exist."`)

---

### 🟡 Bug: Missing `#include <memory>` in table_writer_factory.cpp

The develop version uses `std::make_shared<>()` but doesn't `#include <memory>`. It likely works because of transitive includes, but this is fragile.

---

### 🟡 Design: `makeTableWriter` has no `[[noreturn]]` / unreachable warning

**File:** `table_writer_factory.cpp` (develop version)

```cpp
ITableWriter::Ptr makeTableWriter(TableFormat fmt)
{
    switch (fmt) {
    case TableFormat::CSV: return ...;
    case TableFormat::Parquet: return ...;
    }
    return std::make_shared<CsvTableWriter>();  // unreachable but silences compiler
}
```

If someone adds a new `TableFormat` value and forgets the case, the compiler won't warn (no `default` or `[[fallthrough]]`). Should either add a `default` that throws, or add `[[noreturn]]` + `std::terminate()`.

---

### 🟡 Design: `dynamic_cast` chain with no fallback for unknown types

**File:** `columnToArrowAdapter.cpp`

```cpp
std::shared_ptr<IColumnAdapter> makeColumnAdapter(
  const std::unique_ptr<IO::Outputs::IColumn>& column)
{
    if (auto* c = dynamic_cast<StringColumn*>(column.get())) { ... }
    if (auto* c = dynamic_cast<DoubleColumn*>(column.get())) { ... }
    // ... 5 more dynamic_casts ...

    throw std::invalid_argument("makeColumnAdapter: column type unknown: " + column->name());
}
```

This is a fragile anti-pattern. If a new column type is added to `IColumn` and its adapter isn't registered here, it will throw at runtime with a cryptic message. Consider:
- Using a `std::map<std::type_index, factory_fn>` for registration
- Adding a virtual method on `IColumn` itself (visitor pattern)
- At minimum, logging the type name in the error message

---

### 🟡 Potential issue: `std::filesystem::path&` passed by non-const reference

Both `ParquetTableWriter` and `CsvTableWriter` on the branch take `std::filesystem::path&` (non-const reference):

```cpp
ParquetTableWriter(std::filesystem::path& filePath);
CsvTableWriter(std::filesystem::path& filePath);
```

This is unusual — the path is only read, never modified. Should be `const std::filesystem::path&` or better, take ownership via `std::filesystem::path` by value. The non-const ref also prevents passing temporaries.

---

### 🟡 Test gap: No test for the `LegacySimulationTablesWriter` class itself

The test file `testSimulationTable.cpp` was refactored to test `LegacySimulationTablesWriter`, but:
- The old test `MultipleWriteCalls_AccumulateData` was deleted — this tested the accumulation behavior of `OptimisationsSimulationTable::write()`. It's unclear if this behavior is still covered.
- No test verifies that Parquet output is correctly read back and validated
- No test for the case when `noOutput=true` (tables should be `nullptr`)

---

### 🟡 Memory: `std::shared_ptr` vs `std::unique_ptr` for simulation tables

**File:** `weekly_optimization.h`

```cpp
std::shared_ptr<OptimisationsSimulationTable> simulationTables_;
```

A `shared_ptr` is used but there's only one owner (`WeeklyOptimization`). A `unique_ptr` would be more appropriate and avoid the overhead of reference counting. The only reason to use `shared_ptr` would be if the table is shared across threads, but the solver runs per-week sequentially within a year.

---

### 🟡 Behavior change: Per-year file output vs. single file

The old code aggregated all year buffers into two files (`simulation_table--optim-nb-1.csv` and `simulation_table--optim-nb-2.csv`). The new code creates one file per year (`simulation-table-{year}-optim-nb-{N}`).

This is a breaking change for:
- Any downstream tool/script that reads these files by the old naming convention
- Any study that expects a single monolithic file

There's no migration path or deprecation warning mentioned.

---

### 🟢 Minor: Typo in variable name `addOptionalsToBuider`

**File:** `columnToArrowAdapter.cpp`

```cpp
void addOptionalsToBuider(const std::vector<std::optional<T>>& in, B& builder)
```

`Buider` → `Builder`. Called from `OptDoubleColumnAdapter::makeArray()` and `OptIntColumnAdapter::makeArray()`.

---

### 🟢 Minor: Blank line left after removing code

**File:** `economy.cpp`

```cpp
weeklyOptProblems_.emplace_back(study.parameters.optOptions,
                                &pProblemesHebdo[numSpace],
                                resultWriter_,
                                simulationObserver_.get(),
                                !study.parameters.noOutput);
// ← stray blank line here
```

---

### Summary

| Severity    | Count | Issue                                                                                         |
|-------------|-------|-----------------------------------------------------------------------------------------------|
| 🟡 Bug      | 2     | Typo in error message; dynamic_cast chain is fragile                                          |
| 🟡 Design   | 4     | shared_ptr overuse; path by non-const ref; breaking file naming change; switch without default|
| 🟡 Test     | 1     | Missing Parquet roundtrip test; deleted accumulation test                                     |
| 🟢 Minor    | 2     | Typo in function name; stray whitespace                                                       |

---
See [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3586#issuecomment-4378614701) for an answer

guilpier-code:
About this [big comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3586#pullrequestreview-4157164618) : 
- [x] **Bug: Typo in error message** 
- [x] **Bug: Missing #include <memory> in table_writer_factory.cpp**
- [ ] **Design: makeTableWriter has no [[noreturn]] / unreachable warning**
  ==> Can be adressed in further a dev / PR if necessary
- [ ] **Design: dynamic_cast chain with no fallback for unknown types**
  ==> Can be adressed in further a dev / PR if necessary
- [ ] **Potential issue: std::filesystem::path& passed by non-const reference** : 
  We're forced to do that : unlike what is said in this remark, the path is modified.
  See TODO in class FileWriter's constructor.
- [ ] **Test gap: No test for the LegacySimulationTablesWriter class itself** : remains to be done
- [x] **Memory: std::shared_ptr vs std::unique_ptr for simulation table**
- [x] **Behavior change: Per-year file output vs. single file** : expected
- [x] **Minor: Typo in variable name addOptionalsToBuider**
- [x] **Minor: Blank line left after removing code**

src/tests/cucumber/features/solver-features/hybrid_studies.feature
```diff
@@ -1,12 +1,12 @@
 Feature: hybrid (simulator+modeler) studies
 
-  @fast @short
+  @hybrid
```
guilpier-code: by changing **@fast @short** into **@hybrid**, we avoid that **cucumber** studies are run twice on CI.
Indeed, they were run with fast cucumber step and hybrid cucumber step.

flomnes (commented):
## Comprehensive C++ Review: PR #3586 — "parquet part 3: parquet for legacy solver"

**Scope**: Full code review of the C++ changes across 49 files (535 additions, 323 deletions). This is a significant refactoring that moves simulation table writing from the legacy solver's `OptimisationsSimulationTable` class to a new `LegacySimulationTablesWriter`, while also adding Parquet support via the modeler-side table writers.

---

### Review: LegacySimulationTablesWriter.cpp / .h

### Issues

1. **[High]** Line 35-36: `write()` takes a non-const reference to `OptimisationsSimulationTable` but the `writeForOptim()` method takes a `const SimulationTable*` — inconsistent const-correctness.
   - **Issue**: `write()` calls `writeForOptim()` on both tables, but `writeForOptim` takes `const SimulationTable*`. This means the `write()` method can't modify the table, yet the caller (`economy.cpp` line ~221) calls `simulationTables->clear()` *after* `write()`. The `clear()` call is outside the writer, which is fine, but the API is slightly confusing.
   - **Fix**: Consider making `write()` take `const OptimisationsSimulationTable&` for clarity, or document that the caller must handle `clear()`.

2. **[Medium]** Line 14: `check_is_existing_folder` throws `std::invalid_argument` for a filesystem error.
   - **Issue**: `std::invalid_argument` is meant for invalid function arguments, not filesystem errors. A `std::runtime_error` or `std::filesystem::filesystem_error` would be more appropriate.
   - **Fix**: Use `std::runtime_error` or `std::filesystem::filesystem_error`.

3. **[Low]** Line 20-23: `makeSimuTableFilePath` constructs a filename via string concatenation.
   - **Issue**: The filename format `simulation-table-{year}-optim-nb-{N}` is hardcoded as a string template. If the naming convention changes, it's easy to make a typo.
   - **Suggestion**: Consider using `std::format` (C++20) or `fmt::format` for clarity, or at least a named constant for the prefix.

### Suggestions

- Consider adding `[[nodiscard]]` to `makeSimuTableFilePath` since its return value is always used.

---

### Review: columnToArrowAdapter.cpp / .h

### Issues

1. **[Critical]** Lines 1-131 of `columnToArrowAdapter.h`: Header includes `<arrow/api.h>` and `<arrow/io/api.h>` directly in a public header.
   - **Issue**: This exposes Arrow/Parquet dependencies through the public API. Any consumer of `IColumnAdapter` must include Arrow headers. This creates a hard dependency and increases compile times for all consumers.
   - **Fix**: Use the PIMPL idiom or forward declarations. Move the Arrow-specific `makeField()`/`makeArray()` methods to a private implementation layer.

2. **[High]** `makeColumnAdapter` uses a chain of `dynamic_cast` (lines 206-240 of the diff).
   - **Issue**: If a new column type is added to `IColumn` but its adapter isn't registered here, it will throw at runtime with a cryptic message. This is a fragile anti-pattern (Open-Closed Principle violation).
   - **Fix**: Consider a factory registration pattern or a virtual `accept()` method on `IColumn` (Visitor pattern). At minimum, the error message now includes `typeid(*column).name()` which helps debugging (this was already fixed in the PR).

3. **[Medium]** Line 203-206: Null check added for `makeColumnAdapter`.
   - **Issue**: Good addition, but the check should ideally be at the call site to prevent null from ever reaching this function. If null can legitimately occur, consider returning `std::nullopt` or a null adapter instead of throwing.
   - **Fix**: The null check is appropriate as a defensive measure.

4. **[Low]** `IColumnAdapter` has a virtual destructor with `= default` — correct, but the class has no `final` specifier.
   - **Suggestion**: If these adapter classes are not intended for further inheritance, add `final` to `StringColumnAdapter`, `DoubleColumnAdapter`, etc.

### Suggestions

- The adapter classes are clean and follow the expected pattern.
- Consider adding `noexcept` to destructors (they're `= default` so this is automatic).
- The `makeField()`/`makeArray()` return types use `std::shared_ptr<arrow::Field>` and `std::shared_ptr<arrow::Array>` — appropriate for Arrow's ownership model.

---

### Review: table_writer_factory.cpp / .h

### Issues

1. **[High]** `makeTableWriter` has no `default` case in the switch-like `if` chain.
   - **Issue**: If `parquetFormatRequired` is the only branch, this is effectively a binary choice. However, the function returns `std::make_shared<CsvTableWriter>()` as a fallback. If someone adds a new format enum later, the compiler won't warn.
   - **Fix**: Add a `default` branch that throws, or use `[[noreturn]]` + `std::terminate()` for the unreachable path. The current code silences the compiler but is fragile.

2. **[Medium]** `makeTableWriter` takes `std::filesystem::path` by value (not reference).
   - **Issue**: This is actually correct — the factory needs to take ownership of the path to pass it to the concrete writer. However, the old API took `std::filesystem::path&` (non-const reference), which was a bug.
   - **Fix**: The change is correct. No fix needed.

3. **[Low]** Missing `#include <memory>` in `table_writer_factory.cpp`.
   - **Issue**: `std::make_shared` is used but `<memory>` is not included. It works due to transitive includes but is fragile.
   - **Fix**: The PR already adds `#include <memory>` — good catch.

---

### Review: economy.cpp / adequacy.cpp

### Issues

1. **[High]** **Memory leak in `economy.cpp`**: `simulationTables_` is a `std::vector<OptimisationsSimulationTable>` in the `Economy` class, but the code now creates a *new* local `simulationTables` variable per year and writes it. The `simulationTables_` vector member is still resized in `simulationBegin()` but never used for its actual purpose (storing per-year tables).
   - **Issue**: The `simulationTables_` vector is allocated but effectively dead code. It's resized in `simulationBegin()` but the per-year tables are now created as local `std::unique_ptr<OptimisationsSimulationTable>` variables.
   - **Fix**: Remove the `simulationTables_` member from `Economy` and `Adequacy` classes entirely. It's no longer needed.

2. **[High]** **Inconsistency between Economy and Adequacy**: 
   - In `Economy::year()`, the simulation tables are retrieved from `weeklyOptProblems_[numSpace].simulationTables()` and written after the week loop.
   - In `Adequacy::year()`, a new `simulationTables` is created locally at the start of the year, passed to each weekly optimization, and written after the week loop.
   - **Issue**: These two paths have different lifetimes. In `Economy`, the `WeeklyOptimization` owns the table (via `unique_ptr`). In `Adequacy`, the table is created fresh each year. This inconsistency could lead to subtle bugs if the code paths diverge.
   - **Fix**: Consider unifying the approach — either both should create tables locally per year, or both should use the `WeeklyOptimization`-owned approach.

3. **[Medium]** `economy.cpp` line ~132: `auto* simulationTables = weeklyOptProblems_[numSpace].simulationTables();`
   - **Issue**: This raw pointer is used after the `solve()` calls. If `solve()` throws, `simulationTables` is still valid (it's owned by `weeklyOptProblems_`), but the writer is not called in the exception path.
   - **Fix**: Consider wrapping the write in a try-catch or using RAII to ensure cleanup.

4. **[Medium]** `adequacy.cpp` line ~142-145: `simulationTables` is created locally in each year iteration.
   - **Issue**: This is fine for correctness, but it means the table data is accumulated across all weeks of the year in a single `OptimisationsSimulationTable` object. If a week fails, the accumulated data for previous weeks is lost.
   - **Suggestion**: Document this behavior. Consider whether partial data should be written even on failure.

5. **[Low]** `economy.cpp` line ~132: The comment `// In order to avoid slight differences in parallel/sequential, we clear the basis at the start of each year` is followed by `auto* simulationTables = ...` on the next line. There's a blank line between the comment and the code.
   - **Fix**: Remove the stray blank line.

### Suggestions

- The per-year file naming (`simulation-table-{year}-optim-nb-{N}.csv`) is a breaking change from the old monolithic files. This should be documented in the PR description (it is, but consider a migration note).
- Consider adding a `[[nodiscard]]` to `WeeklyOptimization::simulationTables()` since the return value matters.

---

### Review: weekly_optimization.h / .cpp

### Issues

1. **[Medium]** `WeeklyOptimization` constructor takes `bool writeSimuTable` instead of a pointer or `std::optional`.
   - **Issue**: A boolean flag is less expressive than `std::optional<OptimisationsSimulationTable*>` or `std::unique_ptr<OptimisationsSimulationTable>`. It's easy to pass the wrong value.
   - **Fix**: Consider `std::optional<std::reference_wrapper<OptimisationsSimulationTable>>` or keep the bool but add a comment explaining its meaning.

2. **[Low]** `WeeklyOptimization` is marked `final` but has a virtual destructor (`~WeeklyOptimization() = default`).
   - **Issue**: A `final` class doesn't need a virtual destructor since it can't be inherited.
   - **Fix**: Remove `virtual` keyword (though `= default` on a destructor is implicitly virtual if any base class has one — in this case there's no base, so it's fine to just have `~WeeklyOptimization() = default;`).

3. **[Medium]** `WeeklyOptimization` is non-copyable (`delete` copy constructor/assignment) but movable (`default` move constructor/assignment).
   - **Issue**: The class contains `std::unique_ptr<OptimisationsSimulationTable>` which is movable, so this is correct. However, it's stored in a `std::vector<WeeklyOptimization>` in `Economy::weeklyOptProblems_`, which requires move construction. This is fine, but the vector is resized in `simulationBegin()` which triggers move operations.
   - **Suggestion**: Ensure `reserve()` is called before resizing to avoid unnecessary moves.

---

### Review: FileWriter.cpp / .h (modeler)

### Issues

1. **[Medium]** `FileWriter::init()` removed the `generatePathWithSuffix` logic that prevented overwriting existing output folders.
   - **Issue**: The old code would append `-2`, `-3`, etc. to avoid overwriting. This safety mechanism is now gone.
   - **Fix**: Consider whether this is intentional. If so, document it. If not, restore the suffix logic.

2. **[Medium]** `FileWriter` constructor changed from `Writer::TableFormat fmt` to `bool parquetFormatRequired`.
   - **Issue**: A `bool` parameter is less self-documenting than an enum. Callers now pass `parquetFormat` (a bool) which is less clear than `Writer::TableFormat::Parquet`.
   - **Fix**: Consider keeping the enum or using a named parameter pattern. However, this is a minor point.

3. **[Low]** `FileWriter::init()` changed `throw Modeler::ModelerError(...)` to `throw std::runtime_error(...)`.
   - **Issue**: Mixing exception types (`Modeler::ModelerError` in one place, `std::runtime_error` in another) is inconsistent.
   - **Fix**: Use `Modeler::ModelerError` consistently, or use `std::runtime_error` everywhere.

---

### Review: CMakeLists.txt changes

### Issues

1. **[Medium]** `src/libs/antares/writer/CMakeLists.txt`: `LegacySimulationTablesWriter.cpp` was added to the `parquet_writer` target, but the target name is misleading.
   - **Issue**: The PR description mentions that the target `parquet_writer` should be renamed to `simulation-table-writers` (it was renamed from that, but the PR description says to rename it back). The target name `parquet_writer` is too narrow since it now also handles CSV writing.
   - **Fix**: As noted in the PR description, rename to `simulation-table-writers`.

2. **[Low]** Multiple `CMakeLists.txt` files have trailing newline removed (tools/batchrun, tools/finder, tools/vacuum, tools/yby-aggregator).
   - **Issue**: POSIX requires text files to end with a newline. Removing it may cause warnings with some tools.
   - **Fix**: Restore trailing newlines.

3. **[Low]** `src/packaging/CMakeLists.txt`: A tab character was introduced instead of spaces.
   - **Issue**: Inconsistent indentation (tabs vs spaces).
   - **Fix**: Use spaces consistently.

---

### Review: Test changes (testSimulationTable.cpp)

### Issues

1. **[High]** Multiple test cases were removed: `Constructor_WritesHeader`, `MultipleWriteCalls_AccumulateData`, `WritePerformance_LargeDataSet`, `TemplateFunction_VariableEntries_AllCombinations`, `Constructor_InitializesEmptyTables`, `AddEntriesToBothTables`, `Clear_ResetsAllTables`.
   - **Issue**: The PR description acknowledges this ("We removed unit tests on printing simulation tables into CSV format"). However, removing tests without replacements is risky. The `MultipleWriteCalls_AccumulateData` test verified that `write()` accumulates data across multiple calls — this behavior is now in `OptimisationsSimulationTable::write()` but the test was deleted.
   - **Fix**: Restore at least the accumulation test, adapted to the new API.

2. **[Medium]** The new test `WriteTo_CreatesCorrectFiles` tests `LegacySimulationTablesWriter` but only tests CSV output.
   - **Issue**: No test verifies Parquet output is correctly written and readable.
   - **Fix**: Add a Parquet roundtrip test.

3. **[Low]** Test file path expectations changed from `simulation_table.csv` to `simulation-table.csv` (with hyphen instead of underscore).
   - **Issue**: This is consistent with the new naming convention, but the test file `testMpsWriter.cpp` also needs to be checked for consistency.

---

### Review: Python test changes (cucumber)

### Issues

1. **[Medium]** `modeler_output_handler.py`: New function `accumulate_simu_table_files()` reads all `simulation-table*.csv` files and concatenates them.
   - **Issue**: This function is a workaround for the per-year file naming. It reads all files matching the pattern and concatenates them, stripping headers from subsequent files. This is fragile — if files are added/removed, the concatenation order matters.
   - **Fix**: Consider using `sorted()` with a stable key that includes the year number to ensure correct ordering.

2. **[Medium]** `solver_steps.py`: Changed from comparing individual simulation table files to comparing entire folders.
   - **Issue**: The old approach compared specific files (`simulation_table--optim-nb-1.csv`, `simulation_table--optim-nb-2.csv`). The new approach compares all files in the output folder against the reference folder. This is broader but may catch unrelated file differences.
   - **Fix**: Consider narrowing the comparison to only simulation-table files.

---

### Review: solver.hxx changes

### Issues

1. **[Medium]** `solver.hxx` still contains `storeYearBuffers` and `aggregateAndWriteSimulationTables` methods that are no longer called from the main simulation loop.
   - **Issue**: These methods are dead code — they're declared in `solver.h` and defined in `solver.hxx` but never called. The `yearSimulationBuffers_` map and `buffersMutex_` are also dead.
   - **Fix**: Remove these methods and members. They were intended for the old aggregation approach and are no longer needed.

2. **[Low]** `#include "antares/io/outputs/OptimisationsSimulationTable.h"` added to `solver.hxx`.
   - **Issue**: This creates a dependency from the core simulation infrastructure to the IO outputs layer. Consider whether this include is still needed (it was added for the now-removed `storeYearBuffers`/`aggregateAndWriteSimulationTables` methods).
   - **Fix**: If the methods are removed, the include should also be removed.

---

### Summary

| Severity | Count | Issue |
|----------|-------|-------|
| **Critical** | 1 | Arrow headers in public header (`columnToArrowAdapter.h`) |
| **High** | 6 | Dead `simulationTables_` members; dead `storeYearBuffers`/`aggregateAndWriteSimulationTables`; missing `default` in factory; `dynamic_cast` chain fragility; inconsistent table ownership between Economy/Adequacy; removed accumulation test |
| **Medium** | 10 | Exception type inconsistency; `bool` parameter vs enum; missing Parquet roundtrip test; folder overwrite safety; `simulationTables_` resize without reserve; Python file concatenation fragility; CMake naming; trailing newlines; tab indentation; exception type in FileWriter |
| **Low** | 5 | `std::invalid_argument` for filesystem error; missing `#include <memory>` (already fixed); non-virtual destructor on final class; stray blank line; `makeSimuTableFilePath` naming |

### Overall Assessment

This is a well-structured refactoring that cleanly separates concerns. The new `LegacySimulationTablesWriter` is a good addition. However, there are several areas that need attention:

1. **Dead code cleanup**: The `storeYearBuffers`/`aggregateAndWriteSimulationTables` methods and `simulationTables_` members in `Economy`/`Adequacy` should be removed.
2. **Test coverage**: The removed tests should be restored or replaced with equivalent coverage.
3. **API design**: The Arrow dependency leak through the public header and the `dynamic_cast` chain should be addressed in follow-up PRs.
4. **Consistency**: Exception types and naming conventions should be unified across the changes.

src/libs/antares/writer/columnToArrowAdapter.h
```diff
@@ -0,0 +1,131 @@
+// Copyright 2007-2026, RTE (https://www.rte-france.com)
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+// Arrow / Parquet
+#include <arrow/api.h>
+#include <arrow/io/api.h>
+
```
flomnes, line 8: Not good, don't expose this in a public header

flomnes:
## Updated C++ Review: PR #3586 — "parquet part 3: parquet for legacy solver"

**Branch state**: 53 files changed, 533 additions, 681 deletions. Significant cleanup since the initial review.

---

### ✅ Issues That Were Fixed

| # | Previous Issue | Status |
|---|---------------|--------|
| 1 | Typo `Buider` → `Builder` in `addOptionalsToBuider` | ✅ **FIXED** — renamed to `addOptionalsToBuilder` |
| 2 | `std::invalid_argument` typo in error message (`"is does not exist."`) | ✅ **FIXED** — now `" does not exist."` |
| 3 | Missing `#include <memory>` in `table_writer_factory.cpp` | ✅ **FIXED** — include added |
| 4 | `storeYearBuffers`/`aggregateAndWriteSimulationTables` dead code in `solver.h`/`solver.hxx` | ✅ **FIXED** — both methods and `yearSimulationBuffers_`/`buffersMutex_` members removed |
| 5 | `simulationTables_` dead member in `Economy` and `Adequacy` classes | ✅ **FIXED** — both removed. `getSimulationTable()` and `getSimulationTableHeader()` also removed |
| 6 | `#include "antares/io/outputs/OptimisationsSimulationTable.h"` in `solver.hxx` (dead include) | ✅ **FIXED** — include removed |
| 7 | `table_writer_factory.cpp` still had unreachable fallback code | ✅ **FIXED** — the binary `if/return` is now clean with no unreachable fallback |
| 8 | `FileWriter` constructor still threw `Error::RuntimeError` | ✅ **FIXED** — constructor no longer throws at all (removed the `RuntimeError` for missing folder) |
| 9 | `FileWriter::init()` still removed `generatePathWithSuffix` | ✅ **FIXED** — the suffix logic is restored |

---

### 🔴 Issues Still Present

#### 1. **[Critical]** Arrow headers in public header (`columnToArrowAdapter.h`)

**File**: `src/libs/antares/writer/columnToArrowAdapter.h` (lines 7-8)

```cpp
#include <arrow/api.h>
#include <arrow/io/api.h>
```

- **Issue**: Arrow/Parquet dependencies leak through the public API. Any consumer of `IColumnAdapter` must include Arrow headers, increasing compile times and creating a hard dependency.
- **Fix**: Use PIMPL or move Arrow-specific methods (`makeField()`, `makeArray()`) to a private implementation layer.

#### 2. **[High]** `dynamic_cast` chain with no registration mechanism

**File**: `src/libs/antares/writer/columnToArrowAdapter.cpp` (lines 206-240)

```cpp
std::shared_ptr<IColumnAdapter> makeColumnAdapter(const std::unique_ptr<IColumn>& column)
{
    if (!column) { throw ...; }
    if (auto* c = dynamic_cast<StringColumn*>(column.get())) { ... }
    if (auto* c = dynamic_cast<DoubleColumn*>(column.get())) { ... }
    // ... 7 more dynamic_casts ...
    throw std::invalid_argument("makeColumnAdapter: column type unknown: " + column->name()
                                + " (dynamic type: " + typeid(*column).name() + ")");
}
```

- **Issue**: If a new column type is added to `IColumn` but its adapter isn't registered here, it throws at runtime. This is an Open-Closed Principle violation.
- **Note**: The error message was improved to include `typeid(*column).name()` — good improvement.
- **Fix**: Consider a factory registration pattern or Visitor pattern on `IColumn`.

#### 3. **[High]** `OptimisationsSimulationTable` no longer accumulates data across `write()` calls

**File**: `src/io/outputs/OptimisationsSimulationTable.cpp`

```cpp
void OptimisationsSimulationTable::clear()
{
    firstOptimSimulationTable_.clear();
    secondOptimSimulationTable_.clear();
}
```

- **Issue**: The old `write()` method accumulated buffers (`firstOptimBuffer_ += firstOptimSimulationTable_.buffer()`). The new `clear()` just calls `SimulationTable::clear()` on the internal tables. The `write()` method and `moveBuffers()` method were **removed entirely**. This is a **behavioral change** — the old code accumulated data across multiple `write()` calls, but the new code has no accumulation. The removed test `MultipleWriteCalls_AccumulateData` verified this behavior.
- **Impact**: If callers previously relied on multiple `write()` calls accumulating data, this is a breaking change. However, looking at the new usage in `economy.cpp` and `adequacy.cpp`, each year creates a fresh `OptimisationsSimulationTable` and writes once, so the old accumulation behavior was unused. Still, the behavioral change should be documented.

#### 4. **[High]** `OptimisationsSimulationTable` header has no namespace and a TODO comment

**File**: `src/io/outputs/include/antares/io/outputs/OptimisationsSimulationTable.h`

```cpp
// gp : No namespace here. We should add one (Antares::IO::Outputs).
// gp : Besides, this class could be a struct now

class OptimisationsSimulationTable
{
```

- **Issue**: The class is missing a namespace (should be `Antares::IO::Outputs`) and has a TODO comment in the header. The class is essentially a POD struct with no invariants.
- **Fix**: Either add the namespace and make it a `struct`, or remove the TODO comment and add the namespace.

#### 5. **[Medium]** `ITableWriter::output_file_` is a non-const reference

**File**: `src/libs/antares/writer/include/antares/writer/i_table_writer.h`

```cpp
protected:
    std::filesystem::path& output_file_;
```

- **Issue**: A non-const reference member is unusual. It means the base class holds a reference to a path owned by the caller, which is a dangling reference risk if the caller's path goes out of scope. The derived classes (`CsvTableWriter`, `ParquetTableWriter`) take `std::filesystem::path&` in their constructors, confirming this pattern.
- **Fix**: Either take ownership via `std::filesystem::path` (by value) or use `std::string_view`/`std::string` for the path.

#### 6. **[Medium]** `makeTableWriter` takes `std::filesystem::path&` (non-const reference)

**File**: `src/libs/antares/writer/include/antares/writer/table_writer_factory.h`

```cpp
SIMULATION_TABLE_WRITERS_EXPORT ITableWriter::Ptr makeTableWriter(bool parquetFormatRequired,
                                                                  std::filesystem::path& filePath);
```

- **Issue**: Same as #5 — non-const reference prevents passing temporaries and is unusual for a read-only parameter.
- **Fix**: Take by value (the factory needs to own the path anyway) or use `const std::filesystem::path&`.

#### 7. **[Medium]** `LegacySimulationTablesWriter` uses `std::invalid_argument` for filesystem error

**File**: `src/libs/antares/writer/LegacySimulationTablesWriter.cpp` (line 15)

```cpp
throw std::invalid_argument(msg);
```

- **Issue**: `std::invalid_argument` is for invalid function arguments, not filesystem errors.
- **Fix**: Use `std::runtime_error` or `std::filesystem::filesystem_error`.

#### 8. **[Medium]** `LegacySimulationTablesWriter` constructor takes `const fs::path&` but stores by copy

**File**: `src/libs/antares/writer/LegacySimulationTablesWriter.cpp` (lines 29-33)

```cpp
LegacySimulationTablesWriter::LegacySimulationTablesWriter(const fs::path& folder,
                                                           const unsigned year,
                                                           bool parquetFormat):
    ...
    folder_(folder)  // copy
```

- **Issue**: The parameter is `const fs::path&` but the member is stored by copy. Taking by value with `std::move` would be more efficient when the caller has an rvalue.
- **Fix**: Change constructor to take `std::filesystem::path folder` and use `std::move(folder)`.

#### 9. **[Medium]** `OptimisationsSimulationTable` no longer has `write()` or `moveBuffers()` — tests reference removed methods

**File**: `src/tests/src/io/outputs/testSimulationTable.cpp`

- **Issue**: The removed test `MultipleWriteCalls_AccumulateData` tested accumulation behavior. The new `OptimisationsSimulationTable` has no `write()` method and no `moveBuffers()`. The test `WriteTo_CreatesCorrectFiles` was rewritten to use `LegacySimulationTablesWriter` directly, which is good, but:
  - No test verifies `OptimisationsSimulationTable::clear()` works correctly
  - No test for `OptimisationsSimulationTable` when `noOutput=true` (tables should be `nullptr`)
  - The `Clear_ResetsAllTables` test was removed without replacement

#### 10. **[Medium]** `FileWriter` constructor no longer validates the study path exists

**File**: `src/modeler/fileWriter/FileWriter.cpp`

```cpp
FileWriter::FileWriter(const std::filesystem::path& studyPath, bool parquetFormatRequired):
    studyPath_(studyPath),
    parquetFormatRequired_(parquetFormatRequired)
{
}
```

- **Issue**: The old constructor threw `Error::RuntimeError("Could not find output Folder: " + studyPath.string())` if the path didn't exist. This validation was removed entirely. The error will now manifest later (in `init()`) with a less clear message.
- **Fix**: Restore the validation in the constructor, or document that validation happens in `init()`.

#### 11. **[Medium]** `FileWriter::init()` uses `fs::create_directories` (plural) instead of `fs::create_directory` (singular)

**File**: `src/modeler/fileWriter/FileWriter.cpp` (line 29)

```cpp
if (!fs::is_directory(outputPath_) && !fs::create_directories(outputPath_))
```

- **Issue**: `create_directories` creates all missing parent directories, while `create_directory` creates only the final directory. The old code used `create_directories` too, so this is consistent. However, if the `outputPath_` is `studyPath/output/time`, creating all parents is probably fine. No actual bug, just noting the behavior.

#### 12. **[Low]** `simulation-table-writers_export.h` used in public headers but target is named `parquet_writer`

**File**: `src/libs/antares/writer/CMakeLists.txt`

```cmake
add_library(parquet_writer SHARED ...
```

- **Issue**: The target name `parquet_writer` is misleading since it also handles CSV writing. The PR description notes this should be renamed to `simulation-table-writers`.
- **Fix**: Rename as noted in the PR description.

#### 13. **[Low]** Tab character in `src/packaging/CMakeLists.txt`

**File**: `src/packaging/CMakeLists.txt` (line 115)

- **Issue**: A tab character was introduced instead of spaces for indentation.
- **Fix**: Replace with spaces.

#### 14. **[Low]** Trailing newlines removed from multiple CMakeLists.txt files

**Files**: `src/tools/config/CMakeLists.txt`, `src/tools/vacuum/CMakeLists.txt`, `src/modeler/fileWriter/CMakeLists.txt`

- **Issue**: POSIX requires text files to end with a newline.
- **Fix**: Restore trailing newlines.

---

### 📊 Updated Summary

| Severity | Count | Issue |
|----------|-------|-------|
| **Critical** | 1 | Arrow headers in public header (`columnToArrowAdapter.h`) |
| **High** | 3 | `dynamic_cast` chain fragility; removed accumulation behavior without test; `OptimisationsSimulationTable` no namespace + TODO in header |
| **Medium** | 7 | Non-const ref for path in `ITableWriter`/`makeTableWriter`; `std::invalid_argument` for filesystem error; `std::filesystem::path` by value in constructor; missing test coverage; removed path validation in `FileWriter`; `LegacySimulationTablesWriter` path copy inefficiency |
| **Low** | 3 | Target name `parquet_writer` misleading; tab in CMakeLists.txt; trailing newlines removed |

---

### 🆕 New Issues Found

| # | Issue |
|---|-------|
| 15 | **[Medium]** `OptimisationsSimulationTable::clear()` calls `SimulationTable::clear()` on internal tables, but the old `clear()` cleared string buffers. The old `SimulationTable::clear()` cleared the `ostringstream buffer_` and the `ColumnBasedStorage`. The new `OptimisationsSimulationTable::clear()` delegates to the internal `SimulationTable::clear()`, which is correct now that buffers were removed. **But** — the old `OptimisationsSimulationTable::clear()` cleared `firstOptimBuffer_` and `secondOptimBuffer_` which were separate from the `SimulationTable` objects. This is a **behavioral change** that should be verified. |
| 16 | **[Low]** `LegacySimulationTablesWriter` constructor parameter `bool parquetFormat` — the parameter name `parquetFormat` doesn't match the member name `parquetFormat_` and the convention in `FileWriter` which uses `parquetFormatRequired_`. |

---

### Overall Assessment

The branch has been **significantly cleaned up** since the initial review. Dead code (`storeYearBuffers`, `aggregateAndWriteSimulationTables`, `simulationTables_` members) has been removed, and the `OptimisationsSimulationTable` class has been simplified (buffers removed, delegation to `SimulationTable::clear()`).

The remaining issues are primarily:
1. **API design**: Arrow dependency leak and `dynamic_cast` chain (should be addressed in follow-up PRs)
2. **Test coverage**: Several removed tests need replacement
3. **Minor style**: Path parameter passing, exception types, trailing newlines

guilpier-code:
Answers to the [big comment above](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3586#issuecomment-4487770446) : 

1. Was addressed in **merged** [PR 3654](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3654)
2. **dynamic_cast chain with no registration mechanism** : addressed in separate [PR 3657](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3657)
3. **OptimisationsSimulationTable no longer accumulates data across write() calls** : it's a warning (we change of behavior), not a fix requirement.
4. **OptimisationsSimulationTable header has no namespace and a TODO comment** : addressed in separate [PR 3657](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3657)
5. **ITableWriter::output_file_ is a non-const reference** : Addressed in separate [PR 3657](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3657)
6. **makeTableWriter takes std::filesystem::path& (non-const reference)** : addressed in separate [PR 3657](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3657)
7. **LegacySimulationTablesWriter uses std::invalid_argument for filesystem error** : it's done in this PR
8. **LegacySimulationTablesWriter constructor takes const fs::path& but stores by copy** : addressed in separate [PR 3657](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3657)

src/modeler/main.cpp
```diff
@@ -47,7 +46,7 @@ int main(int argc, const char** argv)
 
     // Options parsing
     std::filesystem::path studyPath(argv[1]);
-    Writer::TableFormat tableFormat = getTableFormat(argc, argv);
+    bool parquetFormat = parquetFormatRequired(argc, argv);
```
JasonMarechal25: Please keep format enum and not a boolean.

JasonMarechal25 (changes requested):

src/modeler/main.cpp
guilpier-code: I knew you would say that, and I think you're right, I should have left it.
I wanted to make it as simple as possible by introducing a **boolean**, but it's less clear.

guilpier-code: Done

JasonMarechal25 (approved):

PR #3588, opened 2026-04-22
refacto: reduce memory consumption of minmax

JasonMarechal25:


src/solver/variable/storage/minmax-data.cpp
```diff
@@ -3,50 +3,47 @@
...
-static void initArray(bool opInferior, std::vector<MinMaxData::Data>& array)
+static void initArray(bool opInferior, MinMaxData::Data& data)
 {
-    for (auto& data: array)
-    {
-        data.value = opInferior ? DBL_MAX : -DBL_MAX; // +inf or -inf
-        data.index = static_cast<uint32_t>(-1);
-    }
+    const double init = opInferior ? DBL_MAX : -DBL_MAX;
+    std::fill(data.values.begin(), data.values.end(), init);
```
flomnes: use [`std::numeric_limits`](https://en.cppreference.com/cpp/types/numeric_limits)

flomnes (approved):

flomnes (commented):
See below

flomnes (commented):
## Review: PR #3588 — "refacto: reduce memory consumption of minmax"

Two high-priority issues that should be addressed before merging:

### 1. Inconsistent use of `Memory::RawPointer` (lines 60, 90 in `minmax.h`)

The `hourly` cases wrap `.data()` with `Memory::RawPointer(...)` while all other categories pass `.data()` directly. Since `Memory::RawPointer` is an identity function (just returns the pointer), this is unnecessary and inconsistent. It looks like leftover from a template deduction workaround that no longer applies with the new function signatures.

**Fix**: Remove the `Memory::RawPointer(...)` wrappers from both `hourly` cases for consistency:

```cpp
// Before:
InternalExportIndices<HOURS_PER_YEAR, VCardT>(report,
                                              Memory::RawPointer(
                                                minmax.hourly.indices.data()),
                                              fileLevel);

// After:
InternalExportIndices<HOURS_PER_YEAR, VCardT>(report,
                                              minmax.hourly.indices.data(),
                                              fileLevel);
```

Same for `hourly.values.data()` at line 90.

### 2. `uint16_t` sentinel value collision (line 26 in `minmax-data.h`, line 18 in `minmax-data.cpp`)

`uint16_t` has a max value of 65535. The code stores `year + 1` as the index, and uses `std::numeric_limits<uint16_t>::max()` (65535) as the sentinel for "never updated". This means value 65535 is ambiguous — it could mean "year 65534" or "never updated". If a simulation runs more than 65534 years, the index will overflow silently.

The old code used `uint32_t` (max ~4 billion years) with `static_cast<uint32_t>(-1)` as sentinel, which was unambiguous.

**Options**:
- **Preferred**: Keep `uint32_t` for indices. The memory savings from `uint16_t` vs `uint32_t` are only 4 bytes per entry × 8760 = ~34 KB per category, which is marginal.
- **Alternative**: If `uint16_t` is kept, reserve 65535 as a sentinel and add a `static_assert` or runtime check that the number of years won't exceed 65534.

PR #3591, opened 2026-04-22
antares-modeler: add time in output

payetvin:


flomnes (dismissed):
Here is my review of the PR "antares-modeler: add time in output".

## Summary

This PR refactors the FileWriter to use a time-based output directory instead of a simulation ID, adds automatic directory suffixing to avoid overwrites, and passes the simulation table name as a parameter. Overall a reasonable refactoring, but there are several concerns below.

---

## 🔴 Issues (must fix before merge)

### 1. SonarQube Quality Gate — 14.7% new code coverage (required ≥ 60%)
The automated check already flagged this. The PR touches FileWriter.cpp and FileWriter.h significantly but adds only 5 lines of test changes in testMpsWriter.cpp without any new unit tests for the directory creation / suffix logic. **This needs dedicated tests** for:
- The auto-increment suffix logic (e.g. when outputPath/ exists, it should become outputPath-1, then outputPath-2, etc.)
- The boundary condition at index < 2000 (what happens when all 2000 slots are taken?)
- The create_directories call replacing create_directory

### 2. Silent failure when all 2000 suffixes are exhausted
If all 2000 candidates exist, the loop exits with index == 2000 and still appends -2000, creating a potentially very deep directory. There is no error or warning logged. Consider:
- Throwing a ModelerError after exhausting candidates, or
- At minimum logging a warning() and falling back to the base path.

### 3. Race condition: TOCTOU in directory existence check
The check std::filesystem::exists(outputPath_) followed by create_directories is a classic TOCTOU race. Between the check and the creation, another process could create the directory. Since create_directories already returns false if the directory exists, the current logic is safe for creation but the existence check before the loop is not atomic. Not a critical issue for single-threaded use, but worth noting.

---

## 🟡 Suggestions (nice to have)

### 4. Magic number 2000
The 2000 limit is a magic number. Consider making it a named constant or parameter, e.g.:
constexpr unsigned int kMaxOutputSuffixAttempts = 2000;

### 5. Hardcoded "1-1" string in Modeler.cpp
This is a magic string. The subproblem identifier should ideally come from the subproblem itself (e.g. subproblem_1_1->getId() or similar), not hardcoded. This couples the writer call to a specific subproblem index.

### 6. Parameter rename: simulationId → time in init()
The parameter was renamed from simulationId to time, which is more accurate. However, this is a breaking API change on IWriter::init(). Ensure all downstream consumers are updated (they appear to be within this PR, but worth confirming).

### 7. Formatting — line width
The reformatted writeSimulationTable signature in FileWriter.h uses a compact style that fits on fewer lines. This is fine, but ensure it is consistent with the project formatting conventions (the code style guide advises ~100 chars). The new formatting looks good.

---

## 🟢 Positive observations

- The auto-increment suffix logic is a good addition to prevent accidental overwrites of existing outputs.
- Replacing create_directory with create_directories is more robust (creates parent directories if needed).
- Removing the simulationId_ member variable and passing simulationTableName as a parameter is a cleaner design — it makes the function more functional and less stateful.
- The test changes in testMpsWriter.cpp correctly adapt to the new outputPath()-based approach, which is actually more robust than hardcoding paths.

---

## Verdict

**Changes requested.** The core logic is sound, but the missing test coverage for the new directory handling logic and the silent failure mode when all suffixes are exhausted need to be addressed before this can be merged. The SonarQube gate will also block CI.

src/modeler/fileWriter/FileWriter.cpp
```diff
@@ -17,13 +17,27 @@
...
+    // avoid overwriting existing output by adding a suffix
+    if (std::filesystem::exists(outputPath_))
+    {
+        std::string candidate;
+        uint index = 1;
+        do
+        {
+            ++index;
+            candidate = outputPath_.string() + '-' + std::to_string(index);
+        } while (std::filesystem::exists(candidate) && index < 2000);
```
JasonMarechal25: Question: Why 2000 ?
Issue: Avoid magic number, prefere constant declaration

JasonMarechal25 (changes requested):

JasonMarechal25 (approved):

PR #3595, opened 2026-04-22
chore: update changelog

JasonMarechal25:


flomnes (approved):
LGTM — changelog looks good. Minor note: verify parquet PR number (#3499) and whether #3581/#3582 should also be included, but nothing blocking.

PR #3602, opened 2026-04-23
refactor(variable): simplify DispatchableGenMargin structure [13]

JasonMarechal25:
This pull request refactors the implementation of the dispatchable generation margin variable in the Antares Solver. The main change is a significant simplification and modernization of the code, moving from a custom class-based approach to a traits-based system that leverages code reuse and reduces boilerplate. The update also replaces the classic include guard with `#pragma once` and adapts the code to use a new `Economy_Base` template.

**Refactoring and modernization:**

* Replaced the old `DispatchableGenMargin` class and its associated logic with a traits-based approach (`DispatchableGenMarginTraits`), which is now used with the generic `Economy_Base` template for variable definition. This removes a large amount of boilerplate and duplicated logic.
* Updated the way results types and intermediate value handling are defined, using modern C++ type aliases and static methods within the traits struct.
* Simplified the code by removing explicit method implementations for initialization, simulation steps, and result merging, as these are now handled by the base template.

**Header and dependency improvements:**

* Replaced the traditional include guard with `#pragma once` for better clarity and to prevent multiple inclusions.
* Updated includes to use `economy_base.h` instead of the previous `variable.h

## Diff
[diff.txt](https://github.com/user-attachments/files/27041040/diff.txt)


src/solver/variable/include/antares/solver/variable/economy/avail-dispatchable-generation.h
```diff
@@ -1,255 +1,85 @@
...
-        VariableAccessorType::InitializeAndReset(results, study);
-    }
-
-    void initializeFromArea(Data::Study* study, Data::Area* area)
-    {
-        pArea = area;
-        // Next
-        NextType::initializeFromArea(study, area);
-    }
+    using AuxiliaryDataType = Data::Area*;
```
payetvin, line 33: AuxilaryDataType is an area or a vector<double>
I don't find the name clear, if you can't use the types directly rename it maybe

payetvin (changes requested):

PR #3609, opened 2026-04-27
refacto(sorties): largely refactor "variable"

JasonMarechal25:
This pull request introduces several improvements and maintenance changes across the codebase and CI workflows. The most notable changes include enhancements to the build and packaging process, stricter class semantics, and the addition of a utility script for managing git branches. Below are the most important changes grouped by theme:

**Build and Packaging Improvements:**

* The `Config.cmake.in` file now ensures that the correct versions of `absl` and `re2` dependencies are used by pinning them to the OR-Tools co-installed versions, preventing linker errors when using vcpkg. It also adds `Arrow` and `Parquet` as required dependencies.
* The Ubuntu CI workflow (`.github/workflows/ubuntu.yml`) has removed the steps for creating and uploading `.tar.gz` and `.deb` archives, as well as the publishing of these assets to GitHub releases. Additionally, the `run-tests` input now defaults to `'true'`. 
* The CentOS CI workflow (`.github/workflows/centos7.yml`) now also triggers on branches matching the `fix/*` pattern.

**Code Quality and Maintenance:**

* The `EnvForLoading` class now explicitly deletes copy and move assignment operators to prevent accidental assignment, enforcing stricter object semantics.
* A minor fix in `common-eco-adq.cpp` replaces direct member access with a function call to `avgdata()`, improving code correctness and maintainability.

**Tooling:**

* Added a new script `push_all.sh` that automates pushing a range of commits to remote branches with a customizable prefix, improving workflow efficiency for managing multiple fix branches.


src/solver/variable/include/antares/solver/variable/adequacy/all.h
guilpier-code, line 1: **To be done someday**

File :  **spatial-aggregate.h**

As already done for struct **ComposeAll** and class **IVariable**, we could separate struct **SpatialAggregateAll** and class **SpatialAggregate** into different files

guilpier-code, line 5: #pragma once

src/solver/variable/include/antares/solver/variable/adequacy/links.h
guilpier-code, line 28: Strangely, the economy counterpart of this file (**../economy/links.h**) does not use `Common::ComposeAll`.

src/solver/variable/include/antares/solver/variable/adequacy/all.h
guilpier-code, line 1: **To be done someday**

File : **economy_base.h**

I can see that this file is related to economy mode, starting with its name.
What is the relationship between the code it contains and this mode.
It seems to me that the class could be called **variable_base**, isn't it ?

src/solver/variable/include/antares/solver/variable/container.h
```diff
@@ -21,80 +21,62 @@
...
+** \brief Top-level container for all output variables
 **
-** This structure is merely a static linked list with all variables
+** Owns the inner aggregator (a JoinAll<…> of scope wrappers) by composition
+** and forwards every hook to it. Adds dynamic-aggregation bookkeeping that
+** isn't part of the inner aggregator's responsibilities.
 */
-template<class NextT = Container::EndOfList>
-class List: public NextT
+template<class Inner = Container::EndOfList>
```
guilpier-code: About : 
```c++
template<class Inner = Container::EndOfList>
class List
{
    ...
}
```
It seems that class `Container::EndOfList` is now useless. 
If so, we should totally remove it.

src/solver/variable/include/antares/solver/variable/commons/rowBalance.h
guilpier-code, line 86: **To be done someday**  _(and not finished yet)_
 
About `struct Statistics` : making searches over the code base, it seems it's not useful (to be checked more deeply).
Indeed, it seems only used to count the max number of columns to be printed in a given output file, and what is tested is if this number is 0 or not. For this, see **reportbuilder.hxx** and its functions **RunForEachThermalCluster** and **RunForEachLink**.

**_Not finished, to be continued._**

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
```diff
@@ -0,0 +1,303 @@
...
+{
+public:
+    using VCardType = VCardDynamicMultiColumn<Traits>;
+    using AncestorType = Variable::IVariable<DynamicMultiColumnBase<Traits>, VCardType>;
+    using ResultsType = typename VCardType::ResultsType;
+    using VariableAccessorType = VariableAccessor<ResultsType, VCardType::columnCount>;
+
+    using AuxiliaryDataType = typename detail::AuxiliaryDataType<Traits>::type;
+
+    static constexpr std::size_t count = 1;
```
guilpier-code:
```c++
static constexpr std::size_t count = 1;
```
useless : removing it causes no compilation damage

src/solver/variable/include/antares/solver/variable/economy/dispatchableGeneration.h
```diff
@@ -3,295 +3,76 @@
...
+};
 
-private:
-    //! Intermediate values for each year
-    typename VCardType::IntermediateValuesType pValuesForTheCurrentYear;
-    std::vector<std::string> groupNames_; // Names of group containing the clusters of the area
-    std::map<std::string, unsigned int> groupToNumbers_; // Gives to each group (of area) a number
-    size_t nbColumns_ = 0;
-    unsigned int pNbYearsParallel;
+using VCardDispatchableGeneration = VCardDynamicMultiColumn<DispatchableGenerationTraits>;
```
guilpier-code: **VCardDispatchableGeneration** : no need for this variable, defined but not used.
There are many variables for which a card type is defined and never used.
For some variables, card is used in tests, not in production code.
For these variables, prefer defining card in the test code, not in production.

src/solver/variable/include/antares/solver/variable/economy/avail-dispatchable-generation.h
```diff
@@ -1,255 +1,73 @@
...
+        intermediateValues.computeStatisticsForTheCurrentYear();
     }
+};
 
-    void hourBegin(unsigned int hourInTheYear)
-    {
-        // Next variable
-        NextType::hourBegin(hourInTheYear);
-    }
+using VCardAvailableDispatchGen = EconomyVariableCard<AvailableDispatchGenTraits>;
```
guilpier-code: `using VCardAvailableDispatchGen = EconomyVariableCard<AvailableDispatchGenTraits>;`
Defined but not used in production code, but only in tests.
If you really need it in tests, I suggest you define it in tests, not in production code.

src/solver/variable/include/antares/solver/variable/adequacy/overallCost.h
```diff
@@ -229,46 +61,31 @@ class OverallCost: public Variable::IVariable<OverallCost<NextT>, NextT, VCardOv
...
-            results.variableUnit = VCardType::Unit();
-
-            pValuesForTheCurrentYear[numSpace]
-              .template buildAnnualSurveyReport<VCardType>(results, fileLevel, precision);
+            values[i] += state.thermalClusterOperatingCostForYear[i];
         }
     }
+};
+
+using VCardOverallCost = Economy::EconomyVariableCard<OverallCostTraits>;
```
guilpier-code: **VCardOverallCost** unused in production code.
Should be removed, and defined in tests if needed in tests.

src/solver/variable/include/antares/solver/variable/adequacy/overallCost.h
```diff
@@ -4,222 +4,54 @@
 #ifndef __SOLVER_VARIABLE_ADEQUACY_OverallCost_H__
```
guilpier-code, line 4:
```c++
#pragma once
```

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -3,6 +3,61 @@
...
+/*!
+** \file economy_base.h
+**
+** \brief Base class for area-level economy variables
+**
+** ## Traits Contract
+**
+** A valid Economy variable Traits must provide:
+** - Required static methods:
+**   - \c Caption() -> std::string
```
guilpier-code: What is `\c`  for (at beginning of many lines) ?

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -18,16 +73,70 @@
...
-// The `detail` namespace contains implementation details and helper
-// utilities for the Economy variable system. Symbols placed in this
-// namespace are internal implementation artifacts (traits, fallbacks,
-// SFINAE helpers) and are not part of the public API. Using a dedicated
-// `detail` namespace makes the intent explicit: these types help detect
-// optional nested types or member functions in `Traits` (for example
-// `AuxiliaryDataType`) and provide safe defaults when those optional
-// members are absent. This enables generic code to compile whether or
-// not a particular `Traits` class provides extra auxiliary data.
+namespace Hooks_
```
guilpier-code: Why append a **_** to a namespace **Hooks_** ? 
Appended **_** normally reserved to private data members.

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -44,10 +153,16 @@ struct AuxiliaryDataType<TraitsT, std::void_t<typename TraitsT::AuxiliaryDataTyp
...
+template<class VCardType, class ResultsType, int CDataLevel, int CFile>
+inline constexpr int statisticsCount =
+  ((VCardType::categoryDataLevel & CDataLevel && VCardType::categoryFileLevel & CFile)
+   ? VCardType::columnCount * ResultsType::count
+   : 0);
 } // namespace detail
 
 template<class Traits>
-struct VCard_Base
+struct EconomyVariableCard
```
guilpier-code, line 121: **EconomyVariableCard** : I'm not sure that mentioning  **Economy** in the name is relevant.
For mode **adequacy**, there's an associated list of variables, and for mode **economy**, there's another list.
Any variable' class / struct is supposed to be isolated from the rest of the world, and no supposed to care if it's **adequacy** or **economy**. It does care about if it's dynamic or not though.
So : 
- **EconomyVariableCard** should be renamed into **VariableCard** or **VCard**.
- **EconomyVariableBase** should be renamed into **VariableBase**

That said, there are a very few cases of variables specific to both modes (see **SpilledEnergy** and **OverallCost**, and that's it).
For these 2 cases, we can name them  **SpilledEnergy_Adq** and **SpilledEnergy_Eco**, or **OverallCost_Adq** and **OverallCost_Eco**. 
And all variables header files should be contained in the same folder.

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -18,16 +73,70 @@
...
-// SFINAE helpers) and are not part of the public API. Using a dedicated
-// `detail` namespace makes the intent explicit: these types help detect
-// optional nested types or member functions in `Traits` (for example
-// `AuxiliaryDataType`) and provide safe defaults when those optional
-// members are absent. This enables generic code to compile whether or
-// not a particular `Traits` class provides extra auxiliary data.
+namespace Hooks_
+{
+template<class Traits, class AuxData>
+void initializeFromAreaIfSupported(AuxData& aux, Data::Study* study, Data::Area* area)
```
guilpier-code, line 38: **initializeFromAreaIfSupported** : 
- not sure **IfSupported** is necessary in the function name. **initializeFromArea** is shorter and does the job.
- Auxiliary data is particularly meaningless. Actually, we initialize a private data member / attibute. So **dataMember** would be better. And function could be called `initMemberFromArea(DataMember& dt, study, ...)`

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
guilpier-code: In hook function names, **IfSupported** is cumbersome.
We could remove it.

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -3,6 +3,61 @@
...
 #pragma once
 
+/*!
+** \file economy_base.h
+**
+** \brief Base class for area-level economy variables
+**
+** ## Traits Contract
+**
+** A valid Economy variable Traits must provide:
```
guilpier-code: This comment is both too thorough and flat. It makes it unclear.
That's the problem with comments generated by AIs.
There many pieces of information we don't need.
What is a **hook** and what is a **policy** in an output variable ?
This information could be useful for a developer, but not given here.

src/solver/variable/include/antares/solver/variable/economy/dispatchable-generation-margin.h
```diff
@@ -1,259 +1,57 @@
...
-            pValuesForTheCurrentYear[numSpace].initializeFromStudy(study);
-        }
-
-        // Next
-        NextType::initializeFromStudy(study);
-    }
 
-    template<class R>
-    static void InitializeResultsFromStudy(R& results, Data::Study& study)
+    static void setHourlyValue(IntermediateValues& yearlyValues,
```
guilpier-code: We should remove `detail::EmptyAuxiliaryData&,` as argument of **setHourlyValue**.
Indeed, if we do, the 3 arguments function will be called (see **economy_base.h**).

This remark also applies at least to the following variables : 
- dispatchable-generation-margin.h
- domesticUnsuppliedEnergy.h
- lold.h
- loldCsr.h
- lolp.h
- price.h
- spilledEnergy.h (economy version)
- unsupliedEnergy.h

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -18,16 +73,70 @@
...
+void weekForEachAreaIfSupported(IV& iv, State& state, uint numSpace)
+{
+    if constexpr (requires { Traits::weekForEachArea(iv, state, numSpace); })
+    {
+        Traits::weekForEachArea(iv, state, numSpace);
+    }
+}
+
+template<class Traits, class IV, class Aux, class State>
+void setHourlyValueIfSupported(IV& iv, Aux& aux, State& state, uint numSpace)
```
guilpier-code: **setHourlyValueIfSupported** : 
there are many calls of this function (for example in **UnsuppliedEnergyTrait**, but not only) where last argument **numSpace** remains unused.
So we could have (for example) : 
```c++
template<class Traits, class IV, class Aux, class State>
void setHourlyValueIfSupported(IV& iv, Aux& aux, State& state, uint numSpace)
{
    if constexpr (requires { Traits::setHourlyValue(iv, aux, state, numSpace); })
    {
        Traits::setHourlyValue(iv, aux, state, numSpace);
    }
    else if constexpr (requires { Traits::setHourlyValue(iv, state, numSpace); })
    {
        Traits::setHourlyValue(iv, state, numSpace);
    }
    else if constexpr (requires { Traits::setHourlyValue(iv, state); })
    {
        Traits::setHourlyValue(iv, state);
    }
}
```

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
guilpier-code, line 38: Still on **initializeFromAreaIfSupported** : its a hook called below 
```c++
void initializeFromArea(Data::Study* study, Data::Area* area)
{
    Hooks::initializeFromAreaIfSupported<Traits>(auxiliaryData_, study, area);
}
```
This function could instead return the auxiliary / data member type, for more clarity : 
```c++
void initializeFromArea(Data::Study* study, Data::Area* area)
{
    auxiliaryData_ = Hooks::initializeFromArea<Traits>(study, area);
}
```

src/solver/variable/include/antares/solver/variable/adequacy/spilledEnergy.h
```diff
@@ -28,13 +28,8 @@ struct VCardSpilledEnergy
...
     }
 
-    //! The expecte results
-    typedef Results<R::AllYears::Average< // The average values throughout all years
-      R::AllYears::StdDeviation<          // The standard deviation values throughout all years
-        R::AllYears::Min<                 // The minimum values throughout all years
-          R::AllYears::Max<               // The maximum values throughout all years
-            >>>>>
-      ResultsType;
+    //! Expected results configuration
```
guilpier-code, line 31: There are several variables remaining almost unchanged.
For example, this one.
Its **economy** counterpart (**economy/spilledEnergy.h**) was deeply changed.

Why keeping variables unchanged ? Are they specific ?

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -92,55 +207,57 @@ struct VCard_Base
     //! Intermediate values
     static constexpr uint8_t hasIntermediateValues = 1;
```
guilpier-code: **hasIntermediateValues** : is defined in many places, but seems never used.
Is it ? (searches on Visual Studio can be incomplete).

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -83,7 +198,7 @@ struct VCard_Base
     static constexpr uint8_t nodeDepthForGUI = +0;
```
guilpier-code: **nodeDepthForGUI** : as name suggests, we may no longer need this.

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -165,92 +280,100 @@ class Economy_Base
 
     void initializeFromArea(Data::Study* study, Data::Area* area)
     {
-        // Next
-        NextType::initializeFromArea(study, area);
+        Hooks_::initializeFromAreaIfSupported<Traits>(auxiliaryData_, study, area);
     }
 
-    void initializeFromLink(Data::Study* study, Data::AreaLink* link)
+    void initializeFromLink(Data::Study* /*study*/, Data::AreaLink* /*link*/)
```
guilpier-code: **initializeFromLink** : this method does nothing.
As the current class inherits from **IVariable**, which already defines an empty **initializeFromLink**, I suggest we remove **initializeFromLink** from class **EconomyVariableBase**.

This remark applies to any other empty method of this class (**simulationEnd**, **hourBegin**).

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
```diff
@@ -165,92 +280,100 @@ class Economy_Base
...
-        NextType::yearBegin(year, numSpace);
+        Hooks_::yearBeginIfSupported<Traits>(
+          pValuesForTheCurrentYear[numSpace],
+          auxiliaryData_,
+          year,
+          numSpace);
     }
 
-    void yearEndBuild(State& state, unsigned int year, unsigned int numSpace)
+    void yearEndBuild(State& state, uint year, uint numSpace)
```
guilpier-code: **yearEndBuild** : at any place where it's defined, it's empty (does nothing).
I suggest we remove any reference to this function.

What's true for this function may be true for another one.
We should check that.

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
```diff
@@ -0,0 +1,303 @@
...
+#pragma once
+
+/*!
+** \file dynamic_multi_column_base.h
+**
+** \brief Base class for dynamic multi-column variables (runtime column count)
+**
+** ## Traits Contract
+**
+** A valid DynamicMultiColumn Traits must provide:
```
guilpier-code, line 13: This (AI generated) comment is of no use : we don't care about returned types, it's obvious.

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
```diff
@@ -0,0 +1,303 @@
...
+    static constexpr uint8_t hasIntermediateValues = 1;
+    static constexpr uint8_t isPossiblyNonApplicable = 0;
+
+    using IntermediateValuesDeepType = IntermediateValues;
+    using IntermediateValuesBaseType = std::vector<IntermediateValues>;
+    using IntermediateValuesType = std::vector<IntermediateValuesBaseType>;
+
+    using IntermediateValuesTypeForSpatialAg = std::unique_ptr<IntermediateValuesDeepType[]>;
+
+    struct Multiple
```
guilpier-code: `struct Multiple` : this struct seems private to dynamic variables. It does not seem to be used outside of this header file.
Unlike **VCardDynamicMultiColumn**, which public.
If this is true, we should move `struct Multiple` outside **VCardDynamicMultiColumn**.

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
```diff
@@ -0,0 +1,303 @@
...
+            AncestorType::pResults[i].reset();
+        }
+    }
+
+    size_t getMaxNumberColumns() const
+    {
+        return nbColumns_ * ResultsType::count;
+    }
+
+    void simulationBegin()
```
guilpier-code: **simulationBegin** : 
I'm confused : this function calls `Traits::onSimulationBegin`.
Searching for **onSimulationBegin** over code base gives that only **FlowQuadTraits** defines this function.
Problem : **FlowQuadTraits** is associated to a **non dynamic** variable.

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
```diff
@@ -0,0 +1,303 @@
...
+class DynamicMultiColumnBase
+    : public Variable::IVariable<DynamicMultiColumnBase<Traits>, VCardDynamicMultiColumn<Traits>>
+{
+public:
+    using VCardType = VCardDynamicMultiColumn<Traits>;
+    using AncestorType = Variable::IVariable<DynamicMultiColumnBase<Traits>, VCardType>;
+    using ResultsType = typename VCardType::ResultsType;
+    using VariableAccessorType = VariableAccessor<ResultsType, VCardType::columnCount>;
+
+    using AuxiliaryDataType = typename detail::AuxiliaryDataType<Traits>::type;
```
guilpier-code: **AuxiliaryDataType** : removing this definition causes no **antares-solver** compilation damage.
It seems useless.

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
```diff
@@ -0,0 +1,303 @@
...
+                          Traits::onSimulationBegin(pValuesForTheCurrentYear, pNbYearsParallel);
+                      })
+        {
+            Traits::onSimulationBegin(pValuesForTheCurrentYear, pNbYearsParallel);
+        }
+    }
+
+    void simulationEnd()
+    {
+    }
```
guilpier-code: For empty methods, see [this remark](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3609#discussion_r3265163255)

src/solver/variable/include/antares/solver/variable/storage/results.h
```diff
@@ -4,87 +4,174 @@
...
+template<class T, class Head, class... Tail>
+struct TupleContains<T, std::tuple<Head, Tail...>>
+  : std::conditional_t<std::is_same_v<T, Head>,
+                       std::true_type,
+                       TupleContains<T, std::tuple<Tail...>>>
+{
+};
+
+template<class T, class Tuple>
+inline constexpr bool tuple_contains_v = TupleContains<T, Tuple>::value;
```
guilpier-code: `constexpr bool tuple_contains_v = TupleContains<T, Tuple>::value;` : 
misleading name, it gives the impression that we want to know if a `std::tuple` contains a value, though we actually want to know if the a `std::tuple` can contain elements of type T, which is different.

**tuple_can_contains_T** or **tuple_can_contains_type** would be a better name.

We desperately need clarity in this very fuzzy code.

src/solver/variable/include/antares/solver/variable/economy/avail-dispatchable-generation.h
JasonMarechal25: Good catch. I've cleaned up in 967a0e8b1bff631600dab85ccc18f62074ca1dcf

src/solver/variable/include/antares/solver/variable/economy/dispatchable-generation-margin.h
JasonMarechal25: You're right. Fixed in 54cfe29c2d1be10013276f54c2dccb546b82d5cd

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
JasonMarechal25: Move to a proper document: 56c87fd6df9afc8cdc116a65823a30af294c6f3d

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/issues/3668

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
JasonMarechal25: The name was misleading. 
Nonetheless I've cleanup : 41d530d2294727e5e0fb5502f6df6567f5e059cb

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
JasonMarechal25: Removed in 283623d0ddb38e539224856060a9421aa15bd0f3

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
JasonMarechal25: Removed empty methods in b73b93a153b954631c38c29e00b44971c772a6b2

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
JasonMarechal25: Done in 48cf28a66b8a0767a36820d5a4d2df16b90662e9

src/solver/variable/include/antares/solver/variable/storage/results.h
```diff
@@ -4,87 +4,172 @@
...
-        count = DecoratorType::count,
-    };
+    static constexpr std::size_t count = std::tuple_size_v<DecoratorTuple>;
+
+private:
+    template<class Tup>
+    struct CategoryFileFold;
 
-    enum
+    template<class... Ds>
```
guilpier-code: In : 
```c++
template<class... Ds>
struct CategoryFileFold<std::tuple<Ds...>>
{
    static constexpr int value = (0 | ... | Ds::categoryFile);
};
```
What is **Ds** ? I guess it means **Decorators** (result decorators : average, std deviation, ...), but it's not obvious.
Renaming this template parameter would be welcome.

src/solver/variable/include/antares/solver/variable/adequacy/spilledEnergy.h
JasonMarechal25: I focused on economy first. I didn't target adequacy to remove duplication.

src/solver/variable/include/antares/solver/variable/commons/links/links.h
```diff
@@ -152,9 +113,6 @@ class Links
 
     void buildDigest(SurveyResults& results, int digestLevel, int dataLevel) const;
 
-    template<class I>
-    static void provideInformations(I& infos);
-
     template<class VCardToFindT>
     inline const double* retrieveHourlyResultsForCurrentYear(uint) const
```
guilpier-code: **retrieveHourlyResultsForCurrentYear** : here and everywhere else, this function does nothing when it's defined, and it's called nowhere.
 So we can wonder why we have it.
My guess is : in tests **simple-study.cpp**, we run a simulation and we retrieve some simulation results to test them. To do that, we indirectly call **retrieveResultsForArea** and **retrieveResultsForThermalCluster**.
We certainly intended to use **retrieveHourlyResultsForCurrentYear** at some point, but we never did.

Whatever the intention, we define "retrieve" functions for tests that remain unused in production code.
This is **bad design**, but we all know it regarding output variables.

src/solver/variable/include/antares/solver/variable/commons/links/links.hxx
```diff
@@ -5,186 +5,181 @@
...
+template<class VariablePerLinkList>
+inline void Links<VariablePerLinkList>::initializeFromLink(Data::Study*, Data::AreaLink*)
 {
-    // Do nothing
 }
 
-template<class VariablePerLink>
-inline void Links<VariablePerLink>::initializeFromAreaLink(Data::Study*, Data::AreaLink*)
+template<class VariablePerLinkList>
+inline void Links<VariablePerLinkList>::initializeFromThermalCluster(Data::Study*,
```
guilpier-code: **initializeFromThermalCluster** : this function is declared and defined in multiple places in code base, but : 
- it does nothing everywhere
- it's never called

We should think of removing it.

src/solver/variable/include/antares/solver/variable/commons/links/links.hxx
```diff
@@ -5,186 +5,181 @@
...
+template<class VariablePerLinkList>
+inline void Links<VariablePerLinkList>::initializeFromStudy([[maybe_unused]] Data::Study& study)
 {
-    // Do nothing
 }
 
-template<class VariablePerLink>
-inline void Links<VariablePerLink>::initializeFromStudy([[maybe_unused]] Data::Study& study)
+template<class VariablePerLinkList>
+inline void Links<VariablePerLinkList>::initializeFromLink(Data::Study*, Data::AreaLink*)
```
guilpier-code, line 18: **initializeFromLink** : this function is declared, defined, empty and never called (all over the code base).
We should remove it.

src/solver/variable/include/antares/solver/variable/economy/all.h
```diff
@@ -160,20 +161,11 @@ using VariablesPerSetOfAreas = Common::SpatialAggregateAll<OverallCost,
...
-
-typedef Variable::Join<
-  // Variables for each area / links attached to the areas
-  Variable::Areas<VariablesPerArea>,
-  // Variables for each set of areas
-  Variable::Join<Variable::SetsOfAreas<VariablesPerSetOfAreas>,
-                 // Variables for each binding constraint
-                 Variable::BindingConstraints<VariablesPerBindingConstraints>>>
-  ItemList;
+using ItemList = Variable::JoinAll<Variable::Areas<VariablesPerArea>,
```
guilpier-code, line 185: I've 2 things to say here : 

**First**  (more a **question**) : 
type **Links** is currently embedded inside **VariablesPerArea**. Would it be possible (I mean quite easily) to separate **Links** from areas, and eventually have something like : 
```c++
using ItemList = Variable::JoinAll<Variable::Areas<VariablesPerArea>,
                                   Variable::Links<VariablesPerLinks>,
                                   Variable::SetsOfAreas<VariablesPerSetOfAreas>,
                                   Variable::BindingConstraints<VariablesPerBindingConstraints>>;
```
?
As **JoinAll** accepts a undefined number of template parameters, at least we don't have to change **JoinAll**

**Second** : 
More about class **JoinAll** :  all of its functions make a (static) loop over its wrappers.
But quite often, the called functions (for wrappers **Areas**, **Links**, **SetOfAreas**, ...) do nothing. But with this design, we're forced to define them, even empty. So we have many empty function all over code base.
So in **JoinAll**, could we not call a function in a wrapper only if it is defined ? I saw that this mecanism exists in **economy_base.h** : 
```c+
if constexpr (requires { Traits::initializeFromArea(aux, study, area); })
{
    Traits::initializeFromArea(aux, study, area);
}
```

src/solver/variable/include/antares/solver/variable/commons/generation.h
```diff
@@ -162,15 +162,6 @@ struct GenerationTraits<WindTag>: public WindTraits
 {
```
guilpier-code: The current source file does not exist in any **CMakeLists.txt**.
And it's not the only one.
Consequences : searching over the whole project on MSVC miss symbols in this files.

src/solver/variable/include/antares/solver/variable/commons/generation.h
```diff
@@ -182,7 +173,6 @@ using VCardTimeSeriesValuesGeneration = VCardTimeSeriesBase<GenerationTraits<Tag
  * generation variables. It handles the specific behavior of renewable
```
guilpier-code: So many comments in this source file, and many duplications.
Prefer quality over quantity.

src/solver/variable/include/antares/solver/variable/commons/join_all.h
```diff
@@ -0,0 +1,273 @@
...
+#include "join.h"
+
+namespace Antares::Solver::Variable
+{
+
+template<class... Wrappers>
+class JoinAll
+{
+public:
+    using VCardType = VCardJoin;
```
guilpier-code:
```c++
using VCardType = VCardJoin;
```
This **VCardType** is unused, neither inside the current source file, nor outside.
We should remove this line.

src/solver/variable/include/antares/solver/variable/commons/join_all.h
```diff
@@ -0,0 +1,273 @@
...
+
+#ifndef __SOLVER_VARIABLE_JOIN_ALL_H__
+#define __SOLVER_VARIABLE_JOIN_ALL_H__
+
+#include <tuple>
+#include <utility>
+
+#include "antares/solver/variable/variable.h"
+
+#include "join.h"
```
guilpier-code:
```c++
#include "join.h"
```
Does not seem to be useful.
We should remove this inclusion.
Furthermore, I can't see any place else where we include this  header.
We may remove **join.h** from code base.

src/solver/variable/include/antares/solver/variable/commons/spatial-aggregate.h
```diff
@@ -3,12 +3,29 @@
 
 #pragma once
 
+#include <type_traits>
```
guilpier-code, line 6: Something is unclear here : 
- this header seems useless where it's included (at least in **economy/all.h**)
- I can'see that it's useful in **adequacy/all.h**, which raises a question : why **adequacy/all.h** is different  from **economy/all.h** in that matter ? Should not both **all.h** be alike ?
- Furthermore, nearly all function of class **SpatialAggregate** are empty. Can we not find a way to remove them ?

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
JasonMarechal25: d6419eba7fb65d8396b8e879834264845c47d5e0

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
JasonMarechal25: indeed, it was dead code left over from old design

src/solver/variable/include/antares/solver/variable/commons/links/links.h
JasonMarechal25: Indeed, it was a scafild mfunction to unit test during refactor

src/solver/variable/include/antares/solver/variable/commons/links/links.hxx
JasonMarechal25: Yes. It's a remnant of the old desing

src/solver/variable/include/antares/solver/variable/economy/all.h
JasonMarechal25: For your first question, maybe. But I would open a new PR since it seems like quite a substential change

For the second ond, it's not quite as simple as that.

src/solver/variable/include/antares/solver/variable/commons/generation.h
JasonMarechal25: Cleaned up

src/solver/variable/include/antares/solver/variable/economy/dynamic_multi_column_base.h
guilpier-code: Easy to fix (if justified)

src/solver/variable/include/antares/solver/variable/economy/economy_base.h
guilpier-code: There are remaining **nodeDepthForGUI**. Ex : in **reserveParticipationBase.h**

guilpier-code (approved):

PR #3617, opened 2026-04-29
test: run test also on top

JasonMarechal25:
Split ubuntu into two runs:
* One will use the committed TestNR submodule
* One will use the latest TestNR version


.github/workflows/ubuntu.yml
```diff
@@ -386,19 +400,21 @@ jobs:
       run: exit 1
 
     - name: Installer .deb creation
+      if: ${{ success() && matrix.isUploadTarget }}
       run: |
         cd _build
         cpack -G DEB -D CPACK_INSTALLED_DIRECTORIES="${{ env.ORTOOLS_DIR }}/install/lib;bin"
 
     - name: Installer deb upload push
+      if: ${{ matrix.isUploadTarget }}
```
YoshuaLisasi, line 409: Maybe add && success() ?

.github/workflows/ubuntu.yml
```diff
@@ -181,10 +189,16 @@ jobs:
       run: |
         echo "SIMTEST=${{ fromJson(env.SIMTEST_JSON).version }}" >> $GITHUB_ENV
 
-    - name: Init submodule Antares_Simulator_Tests_NR
+    - name: Init submodule Antares_Simulator_Tests_NR (commited submodule)
```
YoshuaLisasi: Typo "commited" → "committed"

.github/workflows/ubuntu.yml
JasonMarechal25: It's not needed. There is barrier above that exit if not success.
None the way I did removed the conditional success on the step "Installer .deb creation" for homogeneity

YoshuaLisasi (approved):

PR #3619, opened 2026-04-30
9.3 Fix thermal memory leaks

payetvin:


flomnes (dismissed):

flomnes (changes requested):
target branch must be 9.3.x

flomnes (approved):

PR #3621, opened 2026-05-04
Fix double free on OL8

flomnes:
Avoid mixing static / shared symbols by adding a PRIVATE link in CMake.

### DIFF
--- a/.github/workflows/oracle8.yml
+++ b/.github/workflows/oracle8.yml
@@ -9,6 +9,8 @@ on:
     - dependabot/*
     - release/*
     - copilot/*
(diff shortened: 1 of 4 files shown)


payetvin (approved):

.github/workflows/oracle8.yml
```diff
@@ -9,6 +9,8 @@ on:
     - dependabot/*
     - release/*
     - copilot/*
+    # TODO remove
+    - fix/segfault-ol8
```
flomnes:
```suggestion
```

PR #3627, opened 2026-05-07
fix: Use simulation week for GEMS component in hybrid [ANT-4930]

JasonMarechal25:
This pull request introduces improvements to the calculation of global time indices and block indices in the linear solver, adds a targeted test scenario for hybrid studies, and enhances error reporting in modeler output validation. The main focus is on ensuring correct time/block handling for weekly and daily optimization modes, improving test coverage, and making debugging easier.

Key changes include:

### Core logic and calculation improvements

* Updated the calculation of `globalFirst` and `currentBlock` in `opt_appel_solveur_lineaire.cpp` to use `HeureDansLAnnee` for more accurate and robust time indexing, replacing previous calculations based on week numbers and magic numbers. This ensures correct handling of weekly and daily optimization intervals. [[1]](diffhunk://#diff-e4122da783d1e5790a77aa098a9ef9cb7b697d0d3cfe1c6c10a26cefc689dd1fL85-R92) [[2]](diffhunk://#diff-e4122da783d1e5790a77aa098a9ef9cb7b697d0d3cfe1c6c10a26cefc689dd1fL243-R253)

### Testing and validation

* Added a new test scenario in `hybrid_studies.feature` to verify that simulation weeks are handled properly for GEMS components, including checks for correct output values and objective values.
* Updated the Antares simulator test resources submodule to the latest commit, ensuring the new scenario is supported.

### Developer experience

* Improved error messages in `modeler_output_handler.py` for simulation table lookups, providing detailed information about available components, outputs, and time indices to aid in debugging test failures.

### DIFF
--- a/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
+++ b/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
@@ -239,10 +239,18 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
 
     if (simulationTable && modelerData)
     {
-        unsigned currentBlock = problemeHebdo->OptimisationAuPasHebdomadaire
-                                  ? static_cast<unsigned>(problemeHebdo->HeureDansLAnnee) / 168
-                                  : static_cast<unsigned>(problemeHebdo->HeureDansLAnnee) / 24
-                                      + NumIntervalle;
+        // Compute the current block index (weekly blocks if optimization is weekly,
+        // daily blocks otherwise). Replace magic numbers with named constants.
+        unsigned currentBlock;
+        const unsigned heure = static_cast<unsigned>(problemeHebdo->HeureDansLAnnee);
+        if (problemeHebdo->OptimisationAuPasHebdomadaire)
+        {
+            currentBlock = heure / Antares::Constants::nbHoursInAWeek;
+        }
+        else
+        {
+            currentBlock = heure / HOURS_PER_DAY + static_cast<unsigned>(NumIntervalle);
+        }
         TimeConversionMode timeConversionMode = problemeHebdo->OptimisationAuPasHebdomadaire
                                                   ? TimeConversionMode::WeeklyBlocks
                                                   : TimeConversionMode::DailyBlocks;
--- a/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
+++ b/src/solver/optimisation/opt_appel_solveur_lineaire.cpp
@@ -82,15 +82,14 @@ FillContext buildFillContext(const PROBLEME_HEBDO* problemeHebdo, int NumInterva
     auto nTsInDay = static_cast<unsigned>(problemeHebdo->NombreDePasDeTempsDUneJournee);
     if (problemeHebdo->OptimisationAuPasHebdomadaire)
     {
-        globalFirst = problemeHebdo->weekInTheYear * nTsInDay * problemeHebdo->NombreDeJours;
+        globalFirst = static_cast<unsigned>(problemeHebdo->HeureDansLAnnee);
         globalLast = globalFirst + nTsInDay * problemeHebdo->NombreDeJours - 1;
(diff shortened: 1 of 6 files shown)


flomnes:
Le fix ne semble corriger pas le problème
> Errored scenarios:
>  features/solver-features/hybrid_studies.feature:132  Use simulation week properly for GEMS components

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
guilpier-code, line 104: Adding a **const** : 
`About void fillLinearProblem(FillContext& fillCtx, ...)`
We should replace this it with : 
`void fillLinearProblem(const FillContext& fillCtx, ...)`

guilpier-code, line 103: About comment : 
```c++
// Returns a shared_ptr to the solver
``` 
This comment seems wrong and misleading. We should remove it.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -82,15 +82,14 @@ FillContext buildFillContext(const PROBLEME_HEBDO* problemeHebdo, int NumInterva
     auto nTsInDay = static_cast<unsigned>(problemeHebdo->NombreDePasDeTempsDUneJournee);
```
guilpier-code, line 82: About : 
```c++
auto nTsInDay = static_cast<unsigned>(problemeHebdo->NombreDePasDeTempsDUneJournee);
```
**NombreDePasDeTempsDUneJournee** is always **24**.
It's a constant and makes no sense as a data member of **problemeHebdo**.
Besides, constant **HOURS_PER_DAY** already exist, and is even already used in this source file.
So we should used it here too.
In longer term, we should use it at any place : 
- where **NombreDePasDeTempsDUneJournee** is used
- where **24** is used (when it actually means the nb of hours in a day)
We should create a ticket for this.

src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
```diff
@@ -69,8 +69,27 @@ def get_simulation_table_entry(self, component: str, output: str, block: int, ti
         else:
             df = df[df["scenario_index"] == scenario]
         if len(df) != 1:
+            all_components = self.simulation_table["component"].dropna().unique().tolist()
```
guilpier-code, line 72: Here we're building an error message from the content of a simulation table and we raise an error.
The new code should reflect this and have 2 lines.
Otherwise, so many lines are added to an already fat function.
We should create a **make_error_msg** function, and have : 
```python
if len(df) != 1:
    error_msg = make_error_msg(self.simulation_table)
    raise LookupError(error_msg)
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -240,9 +239,18 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
 
     if (simulationTable && modelerData)
     {
-        unsigned currentBlock = problemeHebdo->OptimisationAuPasHebdomadaire
-                                  ? problemeHebdo->weekInTheYear
-                                  : problemeHebdo->weekInTheYear * 7 + NumIntervalle;
+        // Compute the current block index (weekly blocks if optimization is weekly,
```
guilpier-code, line 242: Here I don't understand the difference between old and new code, functionally.
Sorry. Can you explain ?

JasonMarechal25: problemeHebdo->weekInTheYear is "relative" to the optimisation. If we do 1 week starting at week N, problemeHebdo->weekInTheYear is 0 independently of N.
Here we align output timeblock to the timeblock effectively used before in buildFillContext

flomnes (approved):

PR #3629, opened 2026-05-11
Fix MPS export for integer variables [ANT-4925]

payetvin:


flomnes:
## PR Review: Fix MPS export for integer variables [ANT-4925]

### Summary

This PR adds 5 lines to `src/io/outputs/MPSGenerator.cpp` in the `writeBounds()` method. The change handles a subtle but important edge case: when an **integer variable** has bounds `lb=0` and `ub=+∞`, the code previously just `continue`d (relying on the MPS default of `lb=0`). However, some MPS solvers interpret an **implicit** lower bound of 0 for integer variables as a **binary variable** (`lb=0, ub=1`). The fix explicitly writes a `LI` (Lower Integer) bound to clarify that the variable is integer with the stated lower bound.

---

## Review: `src/io/outputs/MPSGenerator.cpp`

### Issues

1. **[High] The fix is incomplete — it only handles the `lb=0, ub=+∞` case**

   The PR comment says: *"explicit bounds to avoid upper bound interpreted as 1 by some solver (binary vars)"*. This correctly identifies the root cause, but the fix only covers the specific case where `lbIsZero && ubIsPlusInfinity`. 

   The same issue could theoretically occur in other scenarios. For example, if a solver sees an integer variable with no explicit upper bound, it might default to `ub=1`. However, in the MPS format, `ub=+∞` is explicitly represented (via `MI` for lower or no upper bound line), so this shouldn't be an issue there.

   **Assessment**: The fix is actually **correct and targeted** — it only touches the exact problematic case. No expansion is needed.

2. **[Medium] The `LI` bound alone doesn't fully disambiguate from binary**

   The MPS `LI` bound type sets the **lower bound** of an integer variable. The generated line is:
   ```
       BND1 LI varname 0
   ```
   This tells the solver: "this is an integer variable with lower bound 0". But it does **not** explicitly state the upper bound. The upper bound is implicitly `+∞` (no upper bound line is written), which is correct.

   However, some solvers might still interpret `LI var 0` as potentially binary if no upper bound is given. The comment acknowledges this by saying the purpose is to avoid the upper bound being "interpreted as 1", but the `LI` directive only sets the lower bound.

   **Assessment**: Looking at the MPS format specification and major solver behavior (HiGHS, SCIP, Gurobi, CPLEX), `LI` with a lower bound of 0 is the **standard and correct** way to declare a non-negative integer variable. The fix is correct. The comment could be slightly misleading though.

3. **[Low] Comment wording could be clearer**

   Current comment:
   ```cpp
   // explicit bounds to avoid upper bound interpreted as 1 by some solver (binary vars)
   ```
   
   This is slightly confusing. The issue isn't that the *upper bound* is interpreted as 1 — it's that an integer variable with implicit `lb=0` and no explicit upper bound might be **assumed to be binary** (`lb=0, ub=1`) by some solvers when no bounds section is present at all.

   **Suggested fix**:
   ```cpp
   // Explicit LI bound to prevent some solvers from assuming binary (lb=0, ub=1)
   // when no bounds are specified for integer variables.
   ```

### Suggestions

1. **Add a test case** for this specific scenario. The existing test `MPSGenerator_MixedIntegerProblem` creates an integer variable with `lb=0, ub=10`, which doesn't hit this code path. A test like:

   ```cpp
   BOOST_AUTO_TEST_CASE(MPSGenerator_UnboundedIntegerVariable)
   {
       OrtoolsLinearProblem problem(true, "highs");
       auto* x = problem.addIntVariable(0, problem.infinity(), "nonneg_int");
       problem.setObjectiveCoefficient(x, 1.0);
       problem.setMinimization();
       // ... write and verify LI bound is present
   }
   ```

   Interestingly, the existing test file already has `MPSGenerator_UnboundedIntegerVariable`, but it tests the case `lb=-∞, ub=2`, not `lb=0, ub=+∞`. Adding coverage for the `lb=0, ub=+∞` integer case would be valuable.

2. **Consider consistency**: The `isBinary` check on line 290 already handles the `lb=0, ub=1` integer case with `BV`. The new `LI` fix handles `lb=0, ub=+∞`. The code flow is:
   - `lb=-∞, ub=+∞` → `FR` (free)
   - `lb==ub` → `FX` (fixed)
   - `isBinary` (i.e., `lb=0, ub=1`) → `BV` (binary)
   - `lb=0, ub=+∞` for int → `LI` (new fix)
   - Otherwise → `LO`/`UP` or `LI`/`UI`

   This is logically sound. No changes needed, but worth noting the clean structure.

### Overall

| Aspect | Rating | Notes |
|--------|--------|-------|
| **Correctness** | ✅ Good | The fix targets the right code path and uses the correct MPS directive (`LI`) |
| **Safety** | ✅ Good | No regressions — only adds output, doesn't change existing behavior for other cases |
| **Testing** | ⚠️ Needs work | No new test case for the specific `lb=0, ub=+∞` integer variable scenario |
| **Documentation** | ⚠️ Minor | Comment is slightly misleading; could be clearer |
| **Code style** | ✅ Good | Follows project conventions (indentation, formatting, `fmt::format_to`) |

**Recommendation: Approve with minor comment improvement and a request to add a regression test.** The fix itself is sound and addresses a real solver compatibility issue. The 5-line change is minimal and well-targeted.

flomnes (approved):

PR #3636, opened 2026-05-11
Skip time series loading for disabled clusters and short-term storage [ANT-4899]

YoshuaLisasi:
### Skip time series loading for disabled clusters and short-term storage
**Context**

Following the same approach as the binding constraints refactoring _(ANT-4899)_, this PR extends the optimization to other study objects. Previously, time series were loaded for all clusters and short-term storage, even when disabled, wasting time and memory.
Changes

Thermal/Renewable clusters — Added an early return in Cluster::loadDataSeriesFromFolder when enabled is false. 
Since both thermal and renewable clusters inherit from Cluster, both are covered.
Short-term storage — Added an early return in STStorageCluster::loadSeries when enabled() returns false.

In both cases, the function returns true (no error) without reading any file from disk.

**Tests**
Added tests in test-cluster-series.cpp covering both thermal and renewable clusters with enabled = true (series are loaded) and enabled = false (series are skipped).
Also added a test in short-term-storage-input-output.cpp for the disabled short-term storage case.


JasonMarechal25 (approved):

flomnes (changes requested):
Add a `debug` log message when skipping
   ```cpp                                                                                                                                                                                                   
     if (!enabled)                                                                                                                                                                                          
     {                                                                                                                                                                                                      
         logs << "Skipping time series loading for disabled cluster: " << id();                                                                                                                             
         return true;                                                                                                                                                                                       
     }                                                                                                                                                                                                      
   ```

flomnes (approved):

PR #3644, opened 2026-05-14
refactor: Remove yuni, cleaning

payetvin:
This PR continues the "remove Yuni / cleanup" effort by migrating several path/time/file utilities to the C++ standard library, simplifying cluster APIs, and adapting cluster list indexing behavior across loaders and tests.

## Changes

- Make cluster list indexing explicit via `buildIndexes()` (instead of doing it implicitly on `addToCompleteList()`), updating loaders and tests accordingly.
- Replace Yuni-based path building / time formatting / timers with `std::filesystem`, `std::time`, and `std::chrono`.
- Remove GUI-only APIs (`reset`, `copyFrom`, `remove`, some save helpers) from cluster-related types.
- Add tests for renewable cluster loading with invalid properties.
- Add unit tests for `study.importprepro.cpp` covering `importTimeseriesIntoInput()`: no-op behaviour when export is disabled, correct output paths for load/solar/wind/hydro/thermal time-series, and that only configured types produce output files.


JasonMarechal25:
@copilot Add or improve unit tests for src/libs/antares/study/study.importprepro.cpp

flomnes (approved):

PR #3646, opened 2026-05-15
test: Add unit tests for YmlSystem decoder error paths

JasonMarechal25:
Improve coverage

### DIFF
--- a/src/tests/src/io/yml-importers/testYmlSystem.cpp
+++ b/src/tests/src/io/yml-importers/testYmlSystem.cpp
@@ -5,6 +5,8 @@
 
 #include <boost/test/unit_test.hpp>
 
+#include <antares/io/inputs/InputError.h>
+#include <antares/io/inputs/yml-system/decoders.h>
 #include <antares/io/inputs/yml-system/parser.h>
 
 using namespace std::string_literals;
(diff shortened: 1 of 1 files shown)


src/tests/src/io/yml-importers/testYmlSystem.cpp
```diff
@@ -167,3 +169,252 @@ BOOST_AUTO_TEST_CASE(component_two_parameters)
...
+    // silent return-false path of requireMap and yields a default-constructed
+    // ThermalComponent rather than an exception.
+    YmlSystem::Parser parser;
+    const auto system = R"(
+        system:
+            id: base_system
+            thermal-capacity-connections:
+                - component: c1
+                  port: p1
+                  thermal-component: ~
```
YoshuaLisasi, line 332: Is thermal-component intentionally optional in a thermal-capacity-connection? 

When set to null (~), requireMap silently returns false and as_fallback_default yields an empty ThermalComponent.
If that's by design, a comment would help clarify it.

JasonMarechal25: Good catch. Feels like a real issue

YoshuaLisasi (approved):

src/tests/src/io/yml-importers/testYmlSystem.cpp
tbittar: Yes real issue, this should be mandatory !

PR #3654, opened 2026-05-19
Fix - don't expose arrow's symbols in parquet_writer's API

flomnes:


src/libs/antares/writer/columnToArrowAdapter.h
```diff
@@ -3,12 +3,19 @@
 
 #pragma once
 
-// Arrow / Parquet
-#include <arrow/api.h>
-#include <arrow/io/api.h>
+#include <memory>
```
guilpier-code, line 6: It seems : 
```c++
#include <optional>
#include <vector>
```
can both be moved to associated .cpp file

guilpier-code (approved):

PR #3658, opened 2026-05-21
Do not override mps with the one used in infeasibility analysis

tbittar:


flomnes:
Throwaway branch, closing

PR #3659, opened 2026-05-21
Add tests for ANT-4421

tbittar:
They must all pass


src/expressions/visitors/EvalVisitor.cpp
```diff
@@ -425,11 +425,13 @@ EvaluationResult EvaluationResult::timeSumOnVector(int from, int to) const
 {
     const std::vector<double> values = valuesAsVector(); // Exception throw if value_ not a vector
     std::vector<double> to_return(values.size(), 0.);
-
-    for (int shift = from; shift <= to; ++shift)
+    if (from <= to)
```
flomnes: The added `if` is not necessary, since if the condition evaluates to `false` the for loop won't be executed

YoshuaLisasi: done

flomnes (approved):

PR #3676, opened 2026-05-28
feat: add legacy variable in simulation table [ANT-4962]

JasonMarechal25:
* aggregate values are not saved in simulation table, only variable of the problem (problem.NombreDeVariables)
* Legacy variables are saved in simulation table at the same time as modeler variables
* LegacyNameMapper map Raw name to name to display in simulation table.
  * For now it's a static map, will later be changed for something reading a json, yml or whatever


src/libs/antares/reservesParsing/SymmetryFieldBaseListener.cpp
tbittar, line 1: Is having reformatting on reserves file relevant in this PR ?

src/solver/optimisation/LegacyNameMapper.cpp
```diff
@@ -0,0 +1,23 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+#include "antares/solver/optimisation/LegacyNameMapper.h"
+
+namespace Antares::Optimization
+{
+
+const std::unordered_map<std::string, std::string> LegacyNameMapper::kOutputMap = {
+  {"UnsuppliedEnergy", "unsupplied_energy"},
+  {"Spillage", "spillaged_energy"},
```
tbittar: spilled_energy

src/tests/cucumber/features/solver-features/legacy_simulation_table.feature
```diff
@@ -0,0 +1,28 @@
...
+    # `UnsuppliedEnergy::area<area>::hour<33>` (0-based legacy hour); after
+    # the parser/mapper it becomes a simulation-table row with
+    # component=area<area>, output=unsupplied_energy, absolute_time_index=34.
+    Given the solver study path is "Antares_Simulator_Tests_NR/short-tests/002 Thermal fleet - Base"
+    When I run antares simulator
+    Then the simulation succeeds
+    And in area "AREA", unsupplied energy on "2 JAN 09:00" of year 1 is of 52 MW
+    And the modeler outputs contain the following entries
+      | block | component  | output            | timestep | scenario | value |
+      | 1     | area<area> | unsupplied_energy | 34       | 0        | 52    |
```
tbittar: Here in the component column I would only expect `area` i.e. for any component named `Model<Component_name>` in the legacy nomenclature, extract only the component name (i.e. `Thermal<nuc-fr>` -> `nuc-fr` in the ST). The link between a component name and its model will be done in the 3rd ticket handling the hybridation of the ST (generation of a system file)

src/tests/src/solver/optimisation/test_legacy_name_mapper.cpp
```diff
@@ -0,0 +1,45 @@
...
+BOOST_AUTO_TEST_CASE(known_unsupplied_energy_is_mapped)
+{
+    const LegacyNameMapper mapper;
+    BOOST_CHECK_EQUAL(mapper.mapOutput("UnsuppliedEnergy"), "unsupplied_energy");
+}
+
+BOOST_AUTO_TEST_CASE(known_spillage_is_mapped)
+{
+    const LegacyNameMapper mapper;
+    BOOST_CHECK_EQUAL(mapper.mapOutput("Spillage"), "spillaged_energy");
```
tbittar: spilled_energy

tbittar (changes requested):

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -56,6 +59,59 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+        if (name.empty())
+        {
+            continue;
+        }
+
+        const auto parsed = Antares::Optimization::ParseLegacyVariableName(name);
+        if (!parsed)
+        {
+            continue;
+        }
```
```suggestion
        const auto& metadata = problem.variablesMetadata[index];
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -255,15 +311,26 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
...
+                                ::getObjectiveValue(solver.get()),
+                                *modelerData,
+                                optimEntityContainer,
+                                fillCtx,
+                                currentBlock,
+                                timeConversionMode,
+                                true);
+        }
+
+        static const Antares::Optimization::LegacyNameMapper kLegacyNameMapper;
```
guilpier-code: **kLegacyNameMapper** : 
- why is there a k as a first letter ? Typo ? Shouldn't we have **legacyNameMapper** instead ?
- Besides, what's the point in passing this **name mapper** as a argument, if we only use it inside function **FillLegacySimulationTable** ? We should create and use it inside this function's body.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -255,15 +311,26 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
...
+        FillLegacySimulationTable(*simulationTable,
```
guilpier-code, line 316: If we already pass **fillCtx** as argument, why also pass `fillCtx.getYear()` ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -56,6 +59,59 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+{
+    const unsigned globalFirstTimeStep = fillContext.getGlobalFirstTimeStep();
+    const unsigned globalLastTimeStep = fillContext.getGlobalLastTimeStep();
+    const unsigned int block = currentBlock + 1;
+
+    const auto namesSize = problem.NomDesVariables.size();
+    const auto valuesSize = problem.X.size();
+    for (int index = 0; index < problem.NombreDeVariables; ++index)
+    {
+        if (static_cast<std::size_t>(index) >= namesSize
```
guilpier-code: In **opt_optimisation_lineaire.cpp**, we have : 
```c++
void resizeProbleme(PROBLEME_ANTARES_A_RESOUDRE* ProblemeAResoudre,
                    unsigned nombreDeVariables,
                    unsigned nombreDeContraintes)
{
    ...
    ProblemeAResoudre->X.resize(nombreDeVariables);
    ...
    ProblemeAResoudre->NomDesVariables.resize(nombreDeVariables);
    ...
}
```
So here, do we need block : 
```c++
if (static_cast<std::size_t>(index) >= namesSize
    || static_cast<std::size_t>(index) >= valuesSize)
{
    continue;
}
```
?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
guilpier-code, line 42: We could have : 
```c++
using namespace Antares::Optimization;
```
in order to have make some type identifier much shorter (for example **LegacyNameMapper** instead of **Antares::Optimization::LegacyNameMapper**).

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -56,6 +59,59 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+            || static_cast<std::size_t>(index) >= valuesSize)
+        {
+            continue;
+        }
+
+        const auto& name = problem.NomDesVariables[index];
```
guilpier-code: Can a variable not have a name ?
If it's not possible, we should remove this` if... continue` block.

src/tests/src/solver/optimisation/test_legacy_variable_name_parser.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+BOOST_AUTO_TEST_CASE(empty_angle_bracket_value_returns_nullopt)
+{
+    BOOST_CHECK(!ParseLegacyVariableName("output::Area<loc>::hour<>"));
+}
+
+BOOST_AUTO_TEST_CASE(unit_other_than_hour_returns_nullopt)
+{
+    BOOST_CHECK(!ParseLegacyVariableName("output::Area<loc>::day<1>"));
```
guilpier-code: We could add a line which is correct, so that we can compare with the 3 incorrect tests.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -255,15 +302,25 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
...
+                                getObjectiveValue(solver.get()),
+                                *modelerData,
+                                optimEntityContainer,
+                                fillCtx,
+                                currentBlock,
+                                timeConversionMode,
+                                true);
+        }
+
+        static constexpr LegacyNameMapper legacyNameMapper;
```
guilpier-code, line 315:
```c++
static constexpr LegacyNameMapper legacyNameMapper;
```
if we don't need this **legacyNameMapper** in this scope, why do we pass it to de **fill** function ?
Why not create it inside **FillLegacySimulationTable**'s body ?

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -56,11 +59,57 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+
+    const auto infoSize = problem.LegacyVariablesInfo.size();
+    const auto valuesSize = problem.X.size();
+    for (int index = 0; index < problem.NombreDeVariables; ++index)
+    {
+        if (static_cast<std::size_t>(index) >= infoSize
+            || static_cast<std::size_t>(index) >= valuesSize)
+        {
+            continue;
+        }
```
guilpier-code: About :
```c+
const auto infoSize = problem.LegacyVariablesInfo.size();
const auto valuesSize = problem.X.size();
for (int index = 0; index < problem.NombreDeVariables; ++index)
{
    if (static_cast<std::size_t>(index) >= infoSize
        || static_cast<std::size_t>(index) >= valuesSize)
    {
        continue;
    }
    ...
}
```
`problem.LegacyVariablesInfo` and `problem.X` were precisely sized to `problem.NombreDeVariables`, so I'm not sure these checks are useful.
We could remove the **if** statement, as well as **infoSize** and **valuesSize**

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -75,12 +75,24 @@ std::vector<std::string>& Namer::names()
...
 }
 
+void Namer::RecordLegacyVariableInfo(unsigned index,
+                                     const std::string& output,
+                                     const std::string& component)
+{
+    if (legacyInfo_ == nullptr)
+    {
+        return;
+    }
```
guilpier-code: We should make this check (`legacyInfo_ `not null) at construction (constructor), otherwise this check is done many times, which is : 
- is not effective
- is not the mission of a **RecordLegacyVariableInfo** function

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -75,12 +75,24 @@ std::vector<std::string>& Namer::names()
...
+    (*legacyInfo_)[index] = Antares::Optimization::LegacyVariableInfo{output, component, timeStep_};
```
guilpier-code:
```c++
(*legacyInfo_)[index] = = {output, component, timeStep_};
```
is shorter.

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
```diff
@@ -63,14 +70,26 @@ class Namer
...
 class VariableNamer: public Namer
 {
 public:
-    using Namer::Namer;
+    VariableNamer(
+      std::vector<std::string>& names,
+      std::vector<std::optional<Antares::Optimization::LegacyVariableInfo>>& legacyInfo):
+        Namer(names)
+    {
+        legacyInfo_ = &legacyInfo;
```
guilpier-code: Why making **legacyInfo_** a raw pointer ?
Making it a reference would make it easier to use in method **RecordLegacyVariableInfo** : 
`legacyInfo_[index]` is simpler than` (*legacyInfo_)[index]`

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -75,12 +75,24 @@ std::vector<std::string>& Namer::names()
     return names_;
 }
 
+void Namer::RecordLegacyVariableInfo(unsigned index,
```
guilpier-code: In the end, if we simplify  **RecordLegacyVariableInfo** enough, it's a one liner method and we don't need it.

src/solver/optimisation/include/antares/solver/optimisation/LegacyVariableInfo.h
```diff
@@ -0,0 +1,20 @@
...
+#include <string>
+
+namespace Antares::Optimization
+{
+
+// Structured legacy description of an optimization variable, built by VariableNamer
+// at the same time as the variable name and stored parallel to NomDesVariables.
+struct LegacyVariableInfo
+{
+    std::string output;
```
guilpier-code:
```c++
std::string output;
```
For a legacy variable, what is an **output** ?
By digging, we find out that it is the name of the variable in the simulation table.
But we should not have to search, we should know right away what it is.
**output** is way too inexpressive, even as a field of a simulation table.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -56,11 +59,57 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+        if (!info)
+        {
+            continue;
+        }
+
+        std::optional<unsigned> blockTimeIndex;
+        if (info->timeIndex >= globalFirstTimeStep && info->timeIndex <= globalLastTimeStep)
+        {
+            blockTimeIndex = info->timeIndex - globalFirstTimeStep + 1;
+        }
```
guilpier-code: Same remark : the **if** statement seems useless : we always fulfill the conditions : violating them is impossible.

src/solver/optimisation/LegacyNameMapper.cpp
```diff
@@ -0,0 +1,23 @@
+// Copyright 2007-2026, RTE (https://www.rte-france.com)
+// SPDX-License-Identifier: MPL-2.0
+
+#include "antares/solver/optimisation/LegacyNameMapper.h"
+
+namespace Antares::Optimization
+{
+
+const std::unordered_map<std::string, std::string> LegacyNameMapper::kOutputMap = {
```
guilpier-code: Why would we have 2 names for the same variable ?
It makes the code more complicated and less effective at runtime.
**==> answer** : because of **view builder** requirements

src/solver/optimisation/include/antares/solver/optimisation/LegacyVariableInfo.h
JasonMarechal25: renamed: name

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
JasonMarechal25: Because I had put it in Namer, the base class. It could be null pointer for constraintNamer.

I changed it to use virtualization and having the variable only in VariableNamer

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
JasonMarechal25: I've replaced with assert in debug

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
JasonMarechal25: To eventually allow dependency injection

src/solver/optimisation/opt_rename_problem.cpp
JasonMarechal25: Now a no-opt virtual function

guilpier-code (approved):

src/solver/optimisation/opt_rename_problem.cpp
```diff
@@ -75,12 +75,27 @@ std::vector<std::string>& Namer::names()
...
 
+void Namer::RecordLegacyVariableInfo(unsigned /*index*/,
+                                     const std::string& /*output*/,
+                                     const std::string& /*component*/)
+{
+    // No-op for the base namer (e.g. ConstraintNamer); only VariableNamer records legacy info.
+}
+
+void VariableNamer::RecordLegacyVariableInfo(unsigned index,
+                                             const std::string& output,
```
tbittar, line 86: `name` seems more natural than `output` (and is already used in the `LegacyVariableInfo` structure)

tbittar (approved):

PR #3681, opened 2026-05-29
format: reserve symetrie

JasonMarechal25:


flomnes:
Can you also check format-code.sh ?

flomnes (approved):

PR #3682, opened 2026-06-01
feat: allow dynamic mapping of legacy variable names for simulation table [ANT-4962]

JasonMarechal25:
* Introduced LegacyVariableNameParser for parsing legacy variable names.
* Updated CMakeLists.txt to include new source and header files.
* Enhanced fillLegacySimulationTable function to utilize the new parser.


flomnes (commented):
Can you document the YAML file format or provide examples ? Also document where to place this file and how it will be used

PR #3687, opened 2026-06-01
Avoid column lookup for simulation tables

flomnes:
# The good
- Simplify the code by removing multiple `getColumn()` methods
- Avoid 1 linear search for every entry add
# The bad
- Hard-coded columns

### DIFF
--- a/src/io/outputs/include/antares/io/outputs/SimulationTable.h
+++ b/src/io/outputs/include/antares/io/outputs/SimulationTable.h
@@ -17,12 +17,21 @@ public:
     SimulationTable();
     SimulationTable(SimulationTable&& other) noexcept;
     void addEntry(const SimulationTableEntry& entry);
-    const std::vector<std::unique_ptr<IColumn>>& columns() const;
+    const std::vector<std::shared_ptr<IColumn>>& columns() const;
     size_t rowCount() const;
     void clear();
     std::vector<std::vector<std::string>> storageIntoRows() const;
 
 private:
     ColumnBasedStorage storage_;
+    std::shared_ptr<IntegralColumn> block_;
+    std::shared_ptr<OptionalColumn<std::string>> component_;
+    std::shared_ptr<StringColumn> output_;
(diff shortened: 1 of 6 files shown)


src/io/outputs/include/antares/io/outputs/storage.h
```diff
@@ -11,70 +11,33 @@ namespace Antares::IO::Outputs
 class ColumnBasedStorage
 {
 public:
-    void addStringColumn(const std::string& name)
+    std::shared_ptr<StringColumn> addStringColumn(const std::string& name)
```
guilpier-code, line 17: Do we still need **addStringColumn**, **addIntegralColumn**, ... ?
Can we not replace them with `addColumn<StringColumn>(...) `, `addColumn<IntegralColumn>(...)`, ... ?

src/io/outputs/include/antares/io/outputs/SimulationTable.h
```diff
@@ -17,12 +17,21 @@ class SimulationTable
...
-    const std::vector<std::unique_ptr<IColumn>>& columns() const;
+    const std::vector<std::shared_ptr<IColumn>>& columns() const;
     size_t rowCount() const;
     void clear();
     std::vector<std::vector<std::string>> storageIntoRows() const;
 
 private:
     ColumnBasedStorage storage_;
+    std::shared_ptr<IntegralColumn> block_;
+    std::shared_ptr<OptionalColumn<std::string>> component_;
```
guilpier-code, line 28: ok, data members added for performance I suppose.

flomnes: yes, so we don't need to lookup at every add

src/io/outputs/include/antares/io/outputs/storage.h
flomnes: Yes, these methods can be removed if we expose the templated method and template types

guilpier-code (approved):

PR #3692, opened 2026-06-02
feat: Export original problem for unfeasible problems [ANT-4987]

flomnes:
### What it does

This PR cleans up how **MPS solver files** (a standard format for linear/mixed-integer programming problems) are exported from Antares. It removes a confusing boolean flag (`namedProblems`) that controlled whether variables and constraints kept their human-readable names or got renamed to generic index-based labels like `x0`, `c1`, etc.

### The Problem It Solves

That flag was scattered across multiple layers of the codebase — the problem filler, the MPS generator, the MPS writer — making it hard to reason about which path was taken and why. It added unnecessary branching logic everywhere.

### What Changes

- **The flag is removed entirely.** Instead, each call site that exports an MPS file now explicitly decides: "do I want readable names?" — and that decision is visible right where the export happens.
- **Error-case exports always keep names** (since debugging is the whole point of exporting a failed problem).
- **The code gets simpler overall** — about 28 fewer lines, fewer member variables, fewer parameters to pass around.

### In Practice

If you're using Antares and exporting MPS files for debugging or external solver use, the behavior should be the same or better (more readable names in error cases). If you're maintaining the solver code, you'll find the export path easier to follow since the naming decision isn't hidden behind a flag passed through 4-5 function calls.


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -324,7 +324,7 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
     }
 
     return {.timeMeasure = timeMeasure,
-            .mps_writer_factory = mps_writer_factory,
+            .originalProblem = ortoolsProblem,
```
flomnes, line 327: The most significant line: we pass the **original problem** inside a `std::shared_ptr` here to be used by the unfeasible problem analyzer.

flomnes:
Using massif to measure the RAM occupation shows very little difference between this branch and develop for study [short-tests/001 One node - passive](https://github.com/AntaresSimulatorTeam/Antares_Simulator_Tests_NR/tree/main/short-tests/001%20One%20node%20-%20passive)
image

JasonMarechal25 (approved):

PR #3696, opened 2026-06-02
Add cucumber test for antares-modeler + parquet simulation tables

flomnes:
Add cucumber test-case for antares-modeler --parquet

Using an existing study, read the values in the parquet simulation table and compare them with existing ones (CSV).

### DIFF
--- a/src/solver/variable/include/antares/solver/variable/commons/spatial-aggregate.h
+++ b/src/solver/variable/include/antares/solver/variable/commons/spatial-aggregate.h
@@ -206,7 +206,9 @@ public:
     }
 
     template<class V>
-    void computeSpatialAggregatesSummary(V& allVars, unsigned int year, unsigned int numSpace)
+    void computeSpatialAggregatesSummary([[maybe_unused]] V& allVars,
+                                         unsigned int year,
+                                         unsigned int numSpace)
     {
         if (VCardType::VCardOrigin::spatialAggregateMode & Category::spatialAggregateEachYear)
         {
--- a/src/solver/variable/include/antares/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
+++ b/src/solver/variable/include/antares/solver/variable/economy/bindingConstraints/bindingConstraintsMarginalCost.h
@@ -248,15 +248,16 @@ public:
     }
 
     template<class VCardToFindT>
-    static void retrieveResultsForLink(typename Storage<VCardToFindT>::ResultsType** result,
-                                       const Data::AreaLink* link)
+    static void retrieveResultsForLink([[maybe_unused]]
+                                       typename Storage<VCardToFindT>::ResultsType** result,
+                                       [[maybe_unused]] const Data::AreaLink* link)
     {
     }
 
     template<class VCardToFindT>
     static void retrieveResultsForThermalCluster(
-      typename Storage<VCardToFindT>::ResultsType** result,
-      const Data::ThermalCluster* cluster)
+      [[maybe_unused]] typename Storage<VCardToFindT>::ResultsType** result,
+      [[maybe_unused]] const Data::ThermalCluster* cluster)
     {
     }
 
--- a/src/solver/variable/include/antares/solver/variable/storage/raw.h
+++ b/src/solver/variable/include/antares/solver/variable/storage/raw.h
@@ -41,8 +41,8 @@ public:
 
     template<class S, class VCardT>
     void buildSurveyReport(SurveyResults& report,
-                           const S& results,
-                           int dataLevel,
+                           [[maybe_unused]] const S& results,
+                           [[maybe_unused]] int dataLevel,
                            int fileLevel,
                            int precision) const
     {


guilpier-code:
**Caution** : a fix is about to be made (if not already made) to properly write the simulation in parquet format (if required by option) in case of running an hybrid study. This fix can be found [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3698).

flomnes:
> **Caution** : a fix is about to be made (if not already made) to properly write the simulation in parquet format (if required by option) in case of running an hybrid study. This fix can be found [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3698).

Good point, this PR only uses antares-modeler but having antares-solver + hybrid studies would be nice, too

src/tests/cucumber/features/steps/common_steps/modeler_steps.py
guilpier-code: _**Out of this PR, but should be addressed some day**_

**modeler_output_values** is messy. 
Unless you already know this function, you may make some digging to understand what it does.
Indeed : 
-  name **modeler_output_values** is not expressive. Should be renamed into **check_simulation_table_content**
- the code should be looking more like this : 
```python
@step('the modeler outputs contain the following entries')
check_simulation_table_content(context):
    expected_entries = read_expected_entries(context.table)
    simulation_table = context.moh.simulation_table
    check_ST_entries(simulation_table, expected_entries)
```
much more clear.
- Too many **if...else...** : we should have a function / class called for every column (**time step**, **scenario**, **block**)
- Columns related **component** or **output** are supposed to be always present (unlike the previous ones). Is it expected ?

src/tests/cucumber/features/steps/common_steps/modeler_steps.py
```diff
@@ -92,13 +99,18 @@ def run_modeler(context):
     else:
         context.output_path = os.path.join(context.study_path,
                                            "output")  # TODO : fixme parse_output_folder_from_logs(out)
+        use_parquet = hasattr(context, "_use_parquet") and context._use_parquet
```
guilpier-code: I'd like to say 2 things here, but it's all about splitting code : 
1. `context.moh` (of class **modeler_output_handler**) is in fact 2 things : a **simulation table** and an **invest problem**. So there should be 2 classes instead of one, and `context.moh` should be replaced with `context.simu_table` and `context.invest_pb`
2. About reading simulation table, we currently have a bit of code here (in **run_modeler** function) and a bit of code inside **modeler_output_handler**'s constructor. So this reading is spread among 2 entities. Instead, we should make 2 readers (csv and parquet) and a factory of readers.

To sum up previous points, we could have something like : 
```python
context.output_path = ...

simu_table_reader = make_simu_table_reader(context.output_path, context._use_parquet)
context.simu_table = simu_table_reader.read()

context.invest_pb = read_invest_pb(context.output_path)
```

**filePattern** disappears from this level and is defined and used privately inside each ST reader.

src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
```diff
@@ -1,150 +1,72 @@
...
-            f"  block: '{block}'\n"
-            f"  timestep: '{timestep}'\n"
-            f"  scenario: '{scenario}'\n"
-            f"Found {len(df)} row(s) (expected 1).\n"
-            f"Available components: {available_components}\n"
-            f"Available outputs: {available_outputs}\n"
-            f"Available timesteps for component '{component}': {available_for_comp}\n"
-            f"Available absolute_time_index values: {available_timesteps}")
-
-    def get_simulation_table_entry(self, component: str, output: str, block: int, timestep: int, scenario: int):
```
guilpier-code, line 115: I see **get_simulation_table_entry** was moved to **simulation_table_checker.py**.
I'm not sure its place is in a "checker" class. A checker class can check, but here, it also seems to retrieve / fetch.

If (as recommended above) we have a `context.simu_table` object (of class **SimlationTable**), a method **get_entry(...)** could be very relevant.

The same remark applies to other **get** functions.

src/tests/cucumber/features/steps/common_steps/simulation_table_reader.py
```diff
@@ -0,0 +1,89 @@
...
+# They return pandas DataFrames that can be consumed by shared check methods.
+
+from io import StringIO
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+
+import pyarrow as pa
+import pyarrow.parquet as pq
```
guilpier-code: _**Cosmetic**_
I think we could have (for more clarity in places were they're used) : 
```python
import pyarrow as arrow
import pyarrow.parquet as parquet
```
**pa** and **pq** are very short, ok, but also not expressive.

src/tests/cucumber/features/steps/common_steps/simulation_table_reader.py
```diff
@@ -0,0 +1,89 @@
...
+    if not files:
+        raise FileNotFoundError(
+            f"No simulation table files matching '{filePattern}' found in '{outputPath}'"
+        )
+
+    tables = [pq.read_table(f) for f in files]
+    if len(tables) == 1:
+        pa_table = tables[0]
+    else:
+        pa_table = pa.concat_tables(tables)
```
guilpier-code: **Question** : 
what if we have simply : 
```python
tables = [pq.read_table(f) for f in files]
pa_table = pa.concat_tables(tables)
``` 
Are we in trouble in this case ?
Isn't the case` len(tables) == 1` already handled by **arrow** ?

src/tests/cucumber/features/steps/common_steps/simulation_table_reader.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def accumulate_simu_table_files(directory: Path, filePattern: str) -> str:
+    """
+    Accumulate contents of all simulation-table*.csv files into a single string.
+    Keeps the CSV header from the first file and removes it from subsequent files.
+    """
+    files = sorted(directory.glob(filePattern))
+    if not files:
+        return ""
```
guilpier-code:
```python
if not files:
    return ""
```
We don't need that.
Indeed, if **files** is empty, the following loop on **enumerate(files)** makes no iteration and the returned `"".join(accumulated)` is an empty string, as expected.

src/tests/cucumber/features/steps/common_steps/simulation_table_reader.py
flomnes: It's quite usual in python to have 2-letters aliases for modules, e.g
```py
import numpy as np
```

src/tests/cucumber/features/steps/common_steps/modeler_steps.py
flomnes: done

src/tests/cucumber/features/steps/common_steps/modeler_steps.py
flomnes: done

src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
flomnes: done

src/tests/cucumber/features/steps/common_steps/simulation_table_reader.py
flomnes: done

src/tests/cucumber/features/steps/common_steps/simulation_table_reader.py
flomnes: done

src/tests/cucumber/features/steps/common_steps/simulation_table_reader.py
flomnes: done

flomnes:
@guilpier-code 
| # | Comment | Action |
|---|---------|--------|
| **#3** | Rename `modeler_output_values` → `check_simulation_table_content`, refactor with fewer if/else | ✅ Renamed step function; extracted `read_expected_entries()` and `check_st_entries()` helpers |
| **#4** | Split `context.moh` into `context.simu_table` + `context.invest_pb`, introduce reader factory | ✅ `context.simu_table` holds `SimulationTable`, `context.invest_pb` holds investment problems; `make_simulation_table_reader()` factory; `filePattern` defined privately inside each reader |
| **#5** | Move `get_*` functions into a `SimulationTable` class | ✅ Created `SimulationTable` class with `get_entry()`, `get_objective_value()`, `get_objective_values_by_block()`, `get_dataframe()`, `get_column_names()` as methods |
| **#6** | Rename `pa`/`pq` → `arrow`/`parquet` | ✅ Imports renamed to `import pyarrow as arrow` / `import pyarrow.parquet as parquet` |

src/tests/cucumber/features/steps/common_steps/modeler_steps.py
guilpier-code: Much better indeed.

src/tests/cucumber/features/steps/common_steps/simulation_table_checker.py
```diff
@@ -0,0 +1,145 @@
+# Simulation table class and checkers (business logic)
+# The SimulationTable class wraps a pandas DataFrame and provides
+# query methods. These are format-agnostic: the same methods work
+# for CSV or Parquet data.
+
+from typing import Optional
```
guilpier-code: Looking at its content, I would now rename this source file into **simulation_table.py** (instead of **simulation_table_checker.py**)

src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
```diff
@@ -1,150 +1,62 @@
...
-        if df.empty:
-            raise LookupError(f"Simulation table contains no objective values")
-        # Group by block and get the objective value for each block
-        result = {}
-        for _, row in df.iterrows():
-            block = int(row["block"])
-            value = row["value"]
-            result[block] = value
-        return result
+        return self.simulation_table.get_objective_values_by_block()
```
guilpier-code: These 2 getters now seem useless.
After these removals, class **modeler_output_handler** is so small that we can wonder if it's useful, and if it should be removed.

src/tests/cucumber/features/steps/common_steps/simulation_table_checker.py
flomnes: done

src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
flomnes: done

PR #3700, opened 2026-06-05
feat(variable-names): implement default variable name mapping

JasonMarechal25:


src/modeler/loadFiles/CMakeLists.txt
```diff
@@ -26,6 +26,17 @@ target_include_directories(load-modeler-files
...
 
+include(GNUInstallDirs)
+# Bake two paths into the library so loadAll.cpp can find variable-names.yml
+# without platform-specific executable-directory detection:
+#   - INSTALL path: where `cmake --install` places the file (FHS data dir)
+#   - SOURCE path:  absolute path in the source tree, used as a fallback for
+#                   development builds where the install prefix may not exist
+target_compile_definitions(load-modeler-files PRIVATE
+    "ANTARES_VARIABLE_NAMES_INSTALL_PATH=\"${CMAKE_INSTALL_FULL_DATADIR}/antares/variable-names.yml\""
+    "ANTARES_VARIABLE_NAMES_SOURCE_PATH=\"${CMAKE_CURRENT_SOURCE_DIR}/../resources/variable-names.yml\""
```
flomnes, line 37: Does this mean that antares-modeler / antares-solver will require a .yaml file shipped ?

Can we not embed the content of this file into a .cpp ?

1. CMake reads variable-names.yml
2. Places it's content into variable-names.cpp
```cpp
std::string defaultVariableNames()
{
return R"(<content here>)";
}
```
3. This file is linked into a library

PR #3703, opened 2026-06-08
doc: remove pdf user guide build

slcorrensi:
At each release a pdf user guide was generated with sphinx, myst-parser and latex.
It added complexity to the documentation maintenance alongside the site built with mkdocs and the doxygen documentation.
Moreover, it wasn't really used anymore.


flomnes (approved):
LGTM — this is a safe cleanup. The PDF build pipeline was unused and the MkDocs website remains fully functional. No unintended references found.

PR #3704, opened 2026-06-08
feat: map legacy variable names [ANT-4962]

JasonMarechal25:
Map all relevant names


src/solver/optimisation/LegacyNameMapper.cpp
```diff
@@ -7,8 +7,32 @@ namespace Antares::Optimization
...
+  {"NODU", "num_units_on"},
+  {"NumberStartingDispatchableUnits", "num_units_starting"},
+  {"NumberStopingDispatchableUnits", "num_units_stopping"},
+  {"NumberBreakingDownDispatchableUnits", "num_units_falling"},
+
+  // Short term storage
+  {"Injection", "injection_power"},
+  {"Withdrawal", "withdrawal_power"},
+  {"Level", "level"},
+  // CostVariationInjection
```
YoshuaLisasi, line 26: Question : Why the comments ?

JasonMarechal25: Specifications mention this variables but give no value. The coment is meant to show we didn't forget those, but chose to not map them

YoshuaLisasi (approved):

PR #3705, opened 2026-06-08
fix: handle empty/whitespace-only optim-config.yml gracefully

flomnes:
## Problem

When the `optim-config.yml` file exists but is empty (0 bytes) or contains only whitespace, yaml-cpp's `YAML::Load()` returns a null node. Calling `root.as<OptimConfig>()` on this null node throws the cryptic error:

```
invalid node; this may result from using a map iterator as a sequence iterator, or vice-versa
```

This causes the simulation to abort with exit code 1.

## Root Cause

`loadOptimConfigFromYaml()` reads the file content and immediately passes it to the YAML parser. An empty file produces a null YAML node, which cannot be deserialized into the `OptimConfig` struct.

## Fix

Treat empty or whitespace-only `optim-config.yml` files the same as missing files — return a default `OptimConfig{}` and log an info message. This is consistent with the existing behavior when the file does not exist.

## Changes

- **src/modeler/loadFiles/readOptimConfig.cpp**: Added empty/whitespace check before parsing
- **src/tests/src/io/yml-importers/testYmlOptimConfig.cpp**: Added `parse_whitespace_only_string_throws` test
- **src/tests/src/modeler/loadFiles/test-load-optim-config.cpp**: Added two integration tests for empty and whitespace-only files

## Verification

- All existing unit tests pass (TestYmlModel, test-load-yaml-files, tests-modeler)
- Solver runs successfully with the previously-failing test study `001 One node - passive` (exit code 0)
- Logs a clear info message: `Optim config file is empty at <path>`


JasonMarechal25:
"Treat empty or whitespace-only optim-config.yml files the same as missing files " will conflict with https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3695 where we want to concider empty optim-config.yml file as intended empty config and not a missing one

flomnes:
> "Treat empty or whitespace-only optim-config.yml files the same as missing files " will conflict with https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3695 where we want to concider empty optim-config.yml file as intended empty config and not a missing one

No, this PR handles empty optim-config.yml files. This is more of a consequence of #3695, no real conflict IMO

PR #3717, opened 2026-06-11
feat(extra-outputs): constraints and duals [ANT-5006]

JasonMarechal25:
Implement new extra outputs. In the same way we already have variable information we add constraint information. This is particulary useful to retrieve dual information (CoutMarginalDesContraintes)

#### Spec
dual(balance)

#### Implementation
-constraintDuals[index]

#### Infos
price = -dual(area balance constraint): the stored dual is the negative of the marginal price (the legacy outputs print -CoutsMarginauxHoraires).

#### Spec
dual(balance) > unsupplied_energy_cost - 5

#### Implementation
price > *unsuppliedCost - 5. ? 1. : 0.

#### Spec
max(dual(flow_decomposition),-dual(flow_decomposition))
=> |dual(flow dissociation constraint)|

#### Implementation
std::abs(constraintDuals[index])

#### Infos
Anchor: _FlowDissociation_

#### Spec
dual(FinalStockExpression)

#### Implementation
constraintDuals[index]

#### Infos
Anchor: _FinalStockExpression_


src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
```diff
@@ -273,7 +284,15 @@ class ConstraintNamer: public Namer
                                     const std::string& constrIndex_name);
 
 private:
+    void RecordLegacyVariableInfo(unsigned index,
```
guilpier-code, line 287: Now that we defined **RecordLegacyVariableInfo** for class **ConstraintNamer**, there may be something wrong with comment inside `Namer::RecordLegacyVariableInfo` : 
```c++
void Namer::RecordLegacyVariableInfo(unsigned /*index*/,
                                     const std::string& /*name*/,
                                     const std::string& /*component*/)
{
    // No-op for the base namer (e.g. ConstraintNamer); only VariableNamer records legacy info.
}
```

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
```diff
@@ -273,7 +284,15 @@ class ConstraintNamer: public Namer
...
+                                  const std::string& output,
+                                  const std::string& component) override;
+
     void BindingConstraint(unsigned constrIndex,
                            const std::string& name,
                            const std::pair<std::string, std::string>& timeGranularity);
+
+    // Structured legacy description, parallel to names_. Null when this namer
+    // was built without a recording target (the default for constraints).
+    std::vector<std::optional<Antares::Optimization::LegacyVariableInfo>>* legacyInfo_ = nullptr;
```
guilpier-code, line 297: It does not seem to be any difference between `ConstraintNamer::legacyInfo_`  and  `VariableNamer::legacyInfo_`, so why don't they have the same type ?
Please also see remark on methods **RecordLegacyVariableInfo** in the same source file's review.

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
guilpier-code, line 287: Both methods **RecordLegacyVariableInfo** for classes **VariableNamer** and **ConstraintNamer** seem to be very similar.
Could we not define just one **RecordLegacyVariableInfo** in class **Namer** ?

src/solver/optimisation/LegacyExtraOutputs.cpp
guilpier-code, line 276: Instead of : 
```c++
if (info->name == "DispatchableProduction")
{
    AddThermalPropCost(simulationTable,
                      *info,
                       index,
                       solutionValues,
                       linearCosts,
                       fillContext,
                       currentBlock);
}
```
we could have : 
```c++
SimulationTableEntry entry;
....
if (info->name == "DispatchableProduction")
{
    entry = thermalPropCost(*info,
                            index,
                            solutionValues,
                            linearCosts,
                            fillContext,
                            currentBlock);
}
...
simulationTable.addEntry(entry);
```

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -167,12 +167,97 @@ void AddLinkPropCost(SimulationTable& simulationTable,
                          + indirectCost.value() * indirect.value();
     AddExtraOutputEntry(simulationTable, "prop_cost", info, value, fillContext, currentBlock);
 }
+
+// price = -dual(area balance constraint): the stored dual is the negative of
```
guilpier-code, line 171: Many IA generated comments are useless : they add nothing more than what the code already says more clearly, and even slows the reader.

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
JasonMarechal25: Made the function pure virtual 2460361d42580b63389c3166b32fe8fc60efd0e3

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
JasonMarechal25: ConstraintNamer uses Namer 1 parameter constructor. If I wanted to enforce using a 2 parameters constructor to pass a legacyInfo vector I would need to update 50 call sites of ConstrateNamer constructor usage.

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: No.
Not all variables produce an entry so you would add empty entry in simulation table

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: I beg to differ on this one.
The comment are more "spec" oriented and in natural language.

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: Nonetheless a refactor is forecoming https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3765

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
guilpier-code: The current design of **namers** is strange (it always has been) : behavior is different depending on the constructor that is called. It's confusing.

What we could **possibly** do : 
- have a unique **legacyInfo_** in the parent class **Namer**.
- have a new data member for class **Namer** that would be a `std:function` named **treatVariableInfo_**. 
- if constructor **Namer::Namer** is called, then **treatVariableInfo_** is assigned a `doNothing(....)`
- if any other constructor is called (**ConstraintNamer** or **VariableNamer**), **treatVariableInfo_** is assigned a `Namer::RecordVariableInfo(....)` that actually do the job.

Doing this, we have a unique `legacyInfo_` data member instead of 2, and we have 2 possible way of treating the variables info : 
- do nothing
- record them

instead of the currently strange `if (legacyInfo_ != nullptr)` in **ConstraintNamer::RecordLegacyVariableInfo**

guilpier-code (approved):

PR #3726, opened 2026-06-15
Add cucumber test for antares-modeler + parquet simulation tables

flomnes:
Add cucumber tests for antares-modeler with Parquet output format support.


src/tests/cucumber/features/steps/common_steps/solver_steps.py
```diff
@@ -330,7 +324,8 @@ def run_simulation(context):
     # For hybrid studies:
     outputPath = Path(context.output_path)
     if any(outputPath.glob("simulation-table*.csv")):
-        context.moh = modeler_output_handler(outputPath, "simulation-table*-optim-nb-1.csv")
+        context.moh = modeler_output_handler(outputPath)
```
guilpier-code, line 333: We may no longer need `context.moh` and have instead : 
- `context.simu_table `(already exists)
- `context.invest_pb`

In that case, class **modeler_output_handler** would disappear.

If removing class **modeler_output_handler** is suitable, source **modeler_output_handler.py** should be renamed into **invest_problem.py**

src/tests/cucumber/features/steps/common_steps/modeler_output_handler.py
```diff
@@ -1,152 +1,61 @@
...
+
+    Solver tests still reference context.moh (modeler_output_handler),
+    so we keep this class for compatibility. It delegates to SimulationTable.
+    """
+
+    def __init__(self, outputPath: Path):
+        self.simulation_table = SimulationTable(
+            make_simulation_table_reader(outputPath, OutputFormat.CSV)()
+        )
+        self.problems = read_invest_problems(outputPath)
```
guilpier-code: `self.problems` : seems useless

flomnes: done

PR #3730, opened 2026-06-16
ci: add Ubuntu Clang workflow with sanitizers and strict warnings

JasonMarechal25:
Add a dedicated quality-gate workflow building Antares with an up-to-date
Clang and the following checks enabled:
- AddressSanitizer + UndefinedBehaviorSanitizer (fatal on first finding)
- compiler warnings treated as errors (-Werror)
- hardened libstdc++ assertions (_GLIBCXX_ASSERTIONS)
- unit and end-to-end test suite executed under the sanitizers


.github/workflows/ubuntu-clang-sanitizers.yml
```diff
@@ -0,0 +1,184 @@
...
+      run: |
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh ${{ env.CLANG_VERSION }}
+        sudo apt-get install -y clang-${{ env.CLANG_VERSION }} \
+          clang-tools-${{ env.CLANG_VERSION }} \
+          libclang-rt-${{ env.CLANG_VERSION }}-dev \
+          lld-${{ env.CLANG_VERSION }}
+        echo "CC=clang-${{ env.CLANG_VERSION }}" >> $GITHUB_ENV
+        echo "CXX=clang++-${{ env.CLANG_VERSION }}" >> $GITHUB_ENV
```
flomnes, line 68: Why use apt-get and the prebuilt version from llvm.org ? One is not enough ?

JasonMarechal25: Good catch — they serve different purposes, but there was some redundancy. `clang-22` isn't in the Ubuntu 22.04 repositories, so `llvm.sh` is needed to add the apt.llvm.org repository (and it installs `clang`/`lld` for that version). The follow-up `apt-get` was only needed for the packages `llvm.sh` doesn't install but that this job requires: the sanitizer runtime `libclang-rt-22-dev` and `clang-tools`.

The `clang-22`/`lld-22` entries in the `apt-get` line were indeed redundant since `llvm.sh` already installs them. I've trimmed those and added a comment explaining the split (961de48).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3735, opened 2026-06-16
fix(warnings): misc cleanups (unused, tautological compare, typeid, nodiscard)

JasonMarechal25:
Part of the Clang `-Werror` clean-up (batch 4/4: miscellaneous cleanups).

## Changes
- **`list.cpp` — latent bug (`-Wtautological-constant-out-of-range-compare`)**: `findLinkFromINIKey`/`findClusterFromINIKey` compared the result of `key.find(...)` (type `AnyString::Size`, 32-bit) against `AreaName::npos` (64-bit `SIZE_MAX`). The comparison was **always false**, so the "separator not found" guard never fired. Now compares against `AnyString::npos`, matching the `find()` return type.
- **`correlation.cpp` (`-Wunused-function`)**: removed the dead `static inline ExportCorrelationCoefficients` (no callers anywhere in the tree).
- **`report.cpp` (`-Wpotentially-evaluated-expression`)**: bind the dereferenced `shared_ptr` to a reference before `typeid`, so the operand is a side-effect-free glvalue. Dynamic type resolution is unchanged.
- **`testSimulationTable.cpp` (`-Wunused-lambda-capture`)**: dropped the needless capture of the compile-time constant `entriesPerThread`.
- **`test_structured_linear_problem.cpp` (`-Wunused-result`)**: cast the `[[nodiscard]]` `dual()` result to `void` inside `BOOST_CHECK_THROW`.

## Context
Surfaced by the new Ubuntu + up-to-date-Clang CI job that builds with `-Werror`.

---

### DIFF
--- a/src/libs/antares/correlation/correlation.cpp
+++ b/src/libs/antares/correlation/correlation.cpp
@@ -157,36 +157,6 @@ static inline void ReadCorrelationCoefficients(Correlation& correlation,
     }
 }
 
-static inline void ExportCorrelationCoefficients(Study& study,
-                                                 const Matrix<>& m,
-                                                 std::ostream& file,
-                                                 const std::string& name)
-{
-    if (m.empty() or m.width != m.height)
-    {
-        return;
-    }
-
-    file << '[' << name << "]\n";
-
-    // For each column
-    for (uint x = 0; x != m.width; ++x)
-    {
-        const AreaName& from = study.areas.byIndex[x]->id;
-
-        auto& col = m.entry[x];
-        for (uint y = 0; y < x; ++y)
-        {
-            if (!Utils::isZero(col[y]))
-            {
-                file << from << '%' << study.areas.byIndex[y]->id << " = " << col[y] << '\n';
-            }
-        }
-    }
-
-    file << '\n';
-}
-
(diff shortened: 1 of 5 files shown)


JasonMarechal25:
LTGM

src/libs/antares/correlation/correlation.cpp
```diff
@@ -157,36 +157,6 @@ static inline void ReadCorrelationCoefficients(Correlation& correlation,
     }
 }
 
-static inline void ExportCorrelationCoefficients(Study& study,
```
flomnes, line 160: nice catch

src/tests/src/io/outputs/testSimulationTable.cpp
flomnes, line 324: OK so apparently you don't need to capture `entriesPerThread` because it's a constant expression. I suggest using `constexpr` here to make it perfectly clear

flomnes, line 346:
```cpp
BOOST_AUTO_TEST_CASE(ConcurrentAccess_MultipleThreads)
```

Is it really concurrent access if the caller protects the resource with a mutex lock ? This test tests out-of-order access, but certainly not concurrent access

flomnes (approved):

PR #3739, opened 2026-06-18
feat: extra-outputs

JasonMarechal25:


src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -62,14 +63,61 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+        std::vector<double> indirectCapacity(nPdt);
+        for (std::size_t pdt = 0; pdt < nPdt; ++pdt)
+        {
+            directCapacity[pdt] = problemeHebdo.ValeursDeNTC[pdt]
+                                    .ValeurDeNTCOrigineVersExtremite[interco];
+            indirectCapacity[pdt] = problemeHebdo.ValeursDeNTC[pdt]
+                                      .ValeurDeNTCExtremiteVersOrigine[interco];
+        }
+        context.directCapacityByLink[linkKey] = std::move(directCapacity);
+        context.indirectCapacityByLink[std::move(linkKey)] = std::move(indirectCapacity);
```
flomnes: 1. Group `directCapacityByLink` and `indirectCapacityByLink` together in the same container
2. There is a slight risk of collision with `area1 + "$$" + area2`, you may use 

```cpp
class LegacyExtraOutputsContext
{
  struct LinkData {
    std::vector<double> direct;
    std::vector<double> indirect;
  };
  std::map<std::pair<std::string, std::string>, LinkData> capacityByLink;
};
```

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -250,6 +251,99 @@ void AddHydroShadowPrice(SimulationTable& simulationTable,
...
+// upper bound of the DirectFlow variable — and is carried per-pdt in
+// LegacyExtraOutputsContext. Skipped when the link's capacity is not
+// known to the context.
+void AddLinkIsDirectlyCongested(SimulationTable& simulationTable,
+                                const LegacyVariableInfo& info,
+                                std::size_t index,
+                                const std::vector<double>& solutionValues,
+                                const LegacyExtraOutputsContext& context,
+                                const FillContext& fillContext,
+                                unsigned currentBlock)
```
flomnes: 1. Code duplication with `AddLinkIsIndirectlyCongested`, refactor
2. `index` is unclear, rename or pass `solutionValues[index]` directly instead of `{solutionValue, index}`
3. Define `constexpr double saturationEpsilon = 1e-5;` at the top in the anonymous namespace

src/solver/optimisation/include/antares/solver/optimisation/opt_fonctions.h
```diff
@@ -65,13 +66,20 @@ void OPT_VerifierPresenceReserveJmoins1(PROBLEME_HEBDO*);
 **
```
guilpier-code, line 65: This header file is a god header file, it's an example of what not to do.
Change it a bit and you have to re-compile the whole base code.
We should not add anything in this header.
Prefer creating a small new header (declaration) + source file (for definition).

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -62,14 +63,61 @@ static void logProblemSize(const MPSolver* mpSolver)
...
 
+// Snapshot the per-problem study data the legacy extra outputs need but cannot
+// read off the problem's variables. Built once per week from problemeHebdo (the
+// snapshotted data — reservoir capacities and NTC — is week-wide and constant
+// across the daily/weekly blocks of that week), then passed to
+// AddLegacyExtraOutputs. Keys mirror the `component` field on LegacyVariableInfo
+// (lowercased area names; for links, `origin$$destination` matching AREA_SEP in
+// opt_rename_problem.cpp). Declared at global scope in opt_fonctions.h so the
+// per-week caller (runWeeklyOptimization) can build it.
+LegacyExtraOutputsContext BuildLegacyExtraOutputsContext(const PROBLEME_HEBDO& problemeHebdo)
```
guilpier-code: **BuildLegacyExtraOutputsContext** is defined here, and it does seem to make sense.
This free function is actually a constructor of struct / class **LegacyExtraOutputsContext**.
So we should make it a constructor, in a separate **LegacyExtraOutputsContext.cpp** file

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -250,6 +251,99 @@ void AddHydroShadowPrice(SimulationTable& simulationTable,
...
+                        "is_indirectly_congested",
+                        info,
+                        saturated,
+                        fillContext,
+                        currentBlock);
+}
+
 } // namespace
 
 void AddLegacyExtraOutputs(SimulationTable& simulationTable,
```
guilpier-code, line 287: This function (**AddLegacyExtraOutputs**(...)) is long, with 2 loops and many **if** statements.
At least it could be split in 2 functions, one to add extra outputs built from variables, the other to add extra outputs built from constraints.
These 2 functions would take less arguments.

src/tests/src/solver/optimisation/constraints/CMakeLists.txt
```diff
@@ -12,3 +12,9 @@ add_boost_test(constraints_namer
         LIBS
 	model_antares)
 
+add_boost_test(legacy_variable_info_namer
```
guilpier-code, line 15: **legacy_variable_info_namer.cpp** does not exist in code base, so cmake config fails

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+  "op5_emissions"};
+
+// Emits the legacy extra outputs for one weekly solve. The constructor captures
+// everything that is invariant across the dispatch loop (the table, the solution
+// view, the study-data context, the derived area-price map, the raw value / cost
+// / dual arrays, the fill context and the block); each emitter method then takes
+// only the two things that vary per row: the anchor's LegacyVariableInfo and its
+// index. Variable-anchored methods index the value / cost arrays; constraint-
+// anchored methods index the dual array.
+class LegacyExtraOutputEmitter
```
guilpier-code, line 50: **_Cosmetic_** : 

```c++
class LegacyExtraOutputEmitter
```
It's clear that we're in the context of legacy extra output.
So **legacy** is not needed.
This class name could be shortened into **ExtraOutputEmitter**

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -62,14 +63,119 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+        {
+            const std::string key = std::string(areaName) + "$$"
+                                    + paliers.NomsDesPaliersThermiques[palier];
+            context.emissionFactorsByCluster[key] = paliers.emissionFactors[palier];
+
+            LegacyExtraOutputsContext::ThermalMarginData margin;
+            margin.unitSize = paliers.TailleUnitaireDUnGroupeDuPalierThermique[palier];
+            margin.minStablePower = paliers.PminDuPalierThermiquePendantUneHeure[palier];
+            const auto& disp = paliers.PuissanceDisponibleEtCout[palier];
+            if (disp.PuissanceDisponibleDuPalierThermique.size() >= nPdt
```
guilpier-code:
```c++
if (disp.PuissanceDisponibleDuPalierThermique.size() >= nPdt
    && disp.PuissanceMinDuPalierThermique.size() >= nPdt)

```
This **if** statement is **not** needed, as **PuissanceDisponibleDuPalierThermique** is sized to **nPdt**.
Same for **PuissanceMinDuPalierThermique**.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -62,14 +63,119 @@ static void logProblemSize(const MPSolver* mpSolver)
...
+// read off the problem's variables. Built once per week from problemeHebdo (the
+// snapshotted data — reservoir capacities and NTC — is week-wide and constant
+// across the daily/weekly blocks of that week), then passed to
+// AddLegacyExtraOutputs. Keys mirror the `component` field on LegacyVariableInfo
+// (lowercased area names; for links, `origin$$destination` matching AREA_SEP in
+// opt_rename_problem.cpp). Declared at global scope in opt_fonctions.h so the
+// per-week caller (runWeeklyOptimization) can build it.
+LegacyExtraOutputsContext BuildLegacyExtraOutputsContext(const PROBLEME_HEBDO& problemeHebdo)
+{
+    LegacyExtraOutputsContext context;
```
guilpier-code: Furthermore, this build function is quite long and contains very different things : thermal, hydro, links, ...
We should try not to make the same design errors that were made for the **OPT_** part and try to separate these things.
A way to separate things could be to have class **ExtraOutputsContext** inherit from class **ThermalContext**, **HydroContext**, **LinksContext**, ... 
Each of these classes would have its own constructor.
Of course,  **ExtraOutputsContext** would call multiple constructors.

If we do that, we shouldn't hesitate to have multiple loops on areas (**pays**).
But each loop would do much less things.

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+// generation_power is the DispatchableProduction value (by index, area-safe); the
+// rates come from the context keyed by "area$$cluster". Skipped for clusters
+// absent from the context.
+void LegacyExtraOutputEmitter::thermalEmissions(const LegacyVariableInfo& info, std::size_t index)
+{
+    const auto it = context_.emissionFactorsByCluster.find(clusterKey(info));
+    if (it == context_.emissionFactorsByCluster.end())
+    {
+        return;
+    }
```
guilpier-code: About :  
```c++
const auto it = context_.emissionFactorsByCluster.find(clusterKey(info));
if (it == context_.emissionFactorsByCluster.end())
{
    return;
}
```
This block of code raises several questions / remarks for me : 
- is there some cases where we can't find a cluster by its key in `context_.emissionFactorsByCluster` ?
   It seems not : we have to find a cluster by key in `context_.emissionFactorsByCluster`.
   Otherwise, its unexpected and it's an error.
- Same question as above but for `context_.thermalMarginByCluster`
- if `context_.emissionFactorsByCluster` does not contain a cluster key, then don't `context_.thermalMarginByCluster` : these 2 maps are built together.
- this block (find a cluster by key, whatever the map) is duplicated several time, we'd like to have it once instead. 

For these remarks, I suggest : 
- Adopt the design suggested earlier, where class **ExtraOutputsContext** is split into multiple inherited classes
- For thermal part, the thermal class holds a unique map associating a cluster key to the complete data for the cluster (margin, emissionFactor, ...)
- This allows to find a cluster once, not 3 times.

src/solver/optimisation/LegacyExtraOutputs.cpp
guilpier-code, line 50: Can you explain why you called it an **emitter** ?

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
-                        info,
-                        solutionValues[index] > lossOfLoadthreshold ? 1. : 0.,
-                        fillContext,
-                        currentBlock);
+    emit("is_loss_of_load", info, values_[index] > lossOfLoadthreshold ? 1. : 0.);
+}
+
+// actual_load = load: the area's input residual-load series, carried per-hour in
+// the context. Skipped when the area is unknown to the context.
+void LegacyExtraOutputEmitter::areaActualLoad(const LegacyVariableInfo& info, std::size_t)
```
guilpier-code: Why second argument ?
To be removed ?

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+        fillContext_(fillContext),
+        block_(currentBlock)
+    {
+    }
+
+    // One method per anchor: the recorded variable / constraint name that
+    // AddLegacyExtraOutputs dispatches on. Each emits every output derived from
+    // that anchor. Variable anchors index the value / cost arrays; constraint
+    // anchors index the dual array.
+    void dispatchableProduction(const LegacyVariableInfo& info, std::size_t index);
```
guilpier-code: `std::size_t index` : you mean `std::size_t varIndex` or `std::size_t var`, more clear when you navigate in the current code.

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
-                         const std::string& output,
-                         const LegacyVariableInfo& info,
-                         double value,
-                         const FillContext& fillContext,
-                         unsigned currentBlock)
+// Reads a per-hour study series carried in the context (load, inflows, loop
+// flow): finds the component's vector, then indexes it by hour-in-week. Returns
+// nullopt when the component is unknown to the context or the hour is out of
+// range, so the caller can skip the row.
+std::optional<double> ContextValueAtHour(
```
guilpier-code: `std::optional<double> ContextValueAtHour(...)` : this is supposed to be a **LegacyExtraOutputsContext**'s method. Its' a getter.
So instead of : 
```c++
const auto load = ContextValueAtHour(context_.loadByArea,
                                     info.component,
                                     info.timeIndex,
                                     context_.weekFirstTimeStep);
```
we could have : 
```c++
const auto load = context_.load(info.component, info.timeIndex)
```
shorter.

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
-                              .absolute_time_index = info.timeIndex,
-                              .block_time_index = BlockTimeIndex(fillContext, info.timeIndex),
-                              .scenario_index = fillContext.getYear(),
-                              .value = value,
-                              .status = std::nullopt});
+    const auto it = byKey.find(key);
+    if (it == byKey.end())
+    {
+        return std::nullopt;
+    }
```
guilpier-code:
```c++
const auto it = byKey.find(key);
if (it == byKey.end())
{
    return std::nullopt;
}
```
As you know, we're currently filling ST with legacy data and results.
So I'm telling to myself : "when we look for something (for example the load of an area), we're supposed to find it, and if we don't, this means there is a problem".
As a consequence, I don't understand that **ContextValueAtHour(...)** returns an **optional**.
It should return a double, or throw an error.

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+    return it->second[pdt];
+}
+
+// area -> price (= -dual(AreaBalance)), built up front from the recorded
+// AreaBalance constraints so the variable-anchored consumers (thermal profit,
+// link congestion fees, ...) can look up their area's price before the variable
+// dispatch loop. Zero on MIP weeks, where the duals are not extracted (same
+// limitation as the price output).
+std::unordered_map<std::string, double> BuildPriceByArea(
```
guilpier-code: **BuildPriceByArea(...)** : here we build a complete new variable in function **AddLegacyExtraOutputs(...)**.
Not sure it's the mission / responsibility of this function.
To me, mission of **AddLegacyExtraOutputs** is to fetch data, make entries and add it to ST.
This is already something. Also building can be considered too much.  
Shouldn't it be done in the **LegacyExtraOutputsContext** ?
Or in some place else ?

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+// Splits a link component "origin$$destination" (see AREA_SEP in
+// opt_rename_problem.cpp) into its origin and extremity area names. Returns
+// nullopt when the separator is absent.
+std::optional<std::pair<std::string, std::string>> SplitLinkEndpoints(const std::string& component)
+{
+    constexpr std::string_view sep = "$$";
+    const auto pos = component.find(sep);
+    if (pos == std::string::npos)
+    {
+        return std::nullopt;
```
guilpier-code: If we can't find **$$** here, we have a problem, don't we ?
I'm not sure returning an optional solves the problem.

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+    // Pushes one row into the simulation table for the given anchor and value.
+    void emit(const std::string& output, const LegacyVariableInfo& info, double value);
+
+    // Context key for per-cluster study data: the thermal anchor's component is
+    // the cluster name alone, which is not unique across areas, so the area
+    // qualifier is folded in (see AREA_SEP in opt_rename_problem.cpp).
+    std::string clusterKey(const LegacyVariableInfo& info) const
+    {
+        return info.area + "$$" + info.component;
+    }
```
guilpier-code: **clusterKey(...)** does not use any data member, it's not a method but a free function.
To be defined outside the current class.

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+// above. Skipped when the cluster is absent from the context or its per-hour
+// vectors do not cover the anchor's hour.
+void LegacyExtraOutputEmitter::thermalMargins(const LegacyVariableInfo& info, std::size_t index)
+{
+    const auto it = context_.thermalMarginByCluster.find(clusterKey(info));
+    if (it == context_.thermalMarginByCluster.end())
+    {
+        return;
+    }
+    const auto& data = it->second;
```
guilpier-code: `const auto& data = it->second;` : **data** not expressive. 
Rename it into **margin**.

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+    const unsigned pdt = hourInWeek(info);
+    if (pdt >= data.availability.size() || pdt >= data.minGenPower.size())
+    {
+        return;
+    }
```
guilpier-code: This case is not supposed to happen. If it does, something is wrong. Throw an exception ?
I'm not even sure that this kind of check has to be done in class **LegacyExtraOutputEmitter** or that we have to do it. It slows execution and bloats the code.

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
JasonMarechal25: Good catch: [`bb7f10f` (this PR)](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3739/changes/bb7f10f8f3cface8bcb2909a4f6b0baded2217c3)

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
-                          const std::vector<double>& linearCosts,
-                          const LegacySolutionView& solution,
-                          const FillContext& fillContext,
-                          unsigned currentBlock)
-{
-    const auto spilled = solution.value("Spillage", info.component, info.timeIndex);
-    const auto spillageCost = solution.linearCost("Spillage", info.component, info.timeIndex);
+void LegacyExtraOutputEmitter::areaImbalanceCost(const LegacyVariableInfo& info, std::size_t index)
+{
+    const auto spilled = solution_.value("Spillage", info.component, info.timeIndex);
```
guilpier-code: `solution_.value()` should return a double, not an `std::optional`.
If we get a `std::null_opt`, then something is wrong.

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3739/changes/2258cdd629bdb101dc9096085f81d4f706e05809

src/tests/src/solver/optimisation/constraints/CMakeLists.txt
JasonMarechal25: Fixed

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
 
 // is_loss_of_load = 1 when the area has unsupplied energy (above the 0.5 MW
 // solver-noise threshold), 0 otherwise.
-void AddAreaIsLossOfLoad(SimulationTable& simulationTable,
-                         const LegacyVariableInfo& info,
-                         std::size_t index,
-                         const std::vector<double>& solutionValues,
-                         const FillContext& fillContext,
-                         unsigned currentBlock)
+void LegacyExtraOutputEmitter::areaIsLossOfLoad(const LegacyVariableInfo& info, std::size_t index)
```
guilpier-code: This variable is the already existing **LOLD**. 
**Naive question** : why don't we use it ?
(design problem ?)

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
JasonMarechal25: Code changed substantially. Could you confirme if it is still relevant please?

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+// started-units count compares ceil(NODU) against the previous hour; at the
+// first hour of the week the previous NODU is not part of this weekly solution,
+// so the start-up term is dropped.
+void LegacyExtraOutputEmitter::thermalNonPropCost(const LegacyVariableInfo& info, std::size_t index)
+{
+    const double unitsOn = std::ceil(values_[index]);
+    const double fixedCost = costs_[index];
+
+    double startupCost = 0.;
+    if (const auto cost = solution_.linearCost("NumberStartingDispatchableUnits",
```
guilpier-code: All these **if** statements... 
For me : IA useless checks. They make code more complex and slower.

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: Because it "emits" a value for the simulation table

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,471 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
-    AddExtraOutputEntry(simulationTable,
-                        "actual_num_units_on",
-                        info,
-                        std::ceil(solutionValues[index]),
-                        fillContext,
-                        currentBlock);
+    emit("actual_num_units_on", info, std::ceil(values_[index]));
+}
+
+// non_prop_cost = startup_cost * started_units + fixed_cost * units_on, with
```
guilpier-code: We should think of removing all unnecessary comments.
Most of the time, code is more clear than these comments.

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: fixed https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3739/changes/b7356a17ea57297558bbcfd5fd862648921796be

src/solver/optimisation/include/antares/solver/optimisation/LegacyVariableInfo.h
```diff
@@ -15,6 +15,14 @@ struct LegacyVariableInfo
     std::string name;
     std::string component;
     unsigned timeIndex;
+
+    // Area the variable belongs to, captured from the namer's current area at
```
guilpier-code: One data member added : 6 lines of comments. Could be much shorter :-)

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: Done

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,117 +30,485 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
-                              .block_time_index = BlockTimeIndex(fillContext, info.timeIndex),
-                              .scenario_index = fillContext.getYear(),
-                              .value = value,
-                              .status = std::nullopt});
+    const auto it = byKey.find(key);
+    if (it == byKey.end())
+    {
+        return std::nullopt;
+    }
+    const unsigned pdt = timeIndex - weekFirstTimeStep;
```
flomnes: Underflow risk if timeIndex < weekFirstTimeStep

Example - this snippet compiles
```cc
static_assert(1ul - 2ul == 18446744073709551615ul);
```

Add a check or avoid using `unsigned int` for substraction

src/solver/optimisation/LegacyExtraOutputsContext.cpp
```diff
@@ -0,0 +1,112 @@
...
+            directCapacity[pdt] = problemeHebdo.ValeursDeNTC[pdt]
+                                    .ValeurDeNTCOrigineVersExtremite[interco];
+            indirectCapacity[pdt] = problemeHebdo.ValeursDeNTC[pdt]
+                                      .ValeurDeNTCExtremiteVersOrigine[interco];
+            loopFlow[pdt] = problemeHebdo.ValeursDeNTC[pdt]
+                              .ValeurDeLoopFlowOrigineVersExtremite[interco];
+        }
+        directCapacityByLink[linkKey] = std::move(directCapacity);
+        indirectCapacityByLink[linkKey] = std::move(indirectCapacity);
+        loopFlowByLink[std::move(linkKey)] = std::move(loopFlow);
```
flomnes: This strongly suggests having a single struct instead of 3 x `map`

```cc
struct myStruct{A directCapacity, B indirectCapacity, C loopFlow};
std::map<std::string, myStruct} byLink;
```

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -25,347 +28,364 @@ std::optional<unsigned> BlockTimeIndex(const FillContext& fillContext, unsigned
...
+    //   down_margin          = generation_power - min(cluster_availability, minGenPower)
+    // availability, unitSize and minStablePower are the spinning-adjusted
+    // weekly-problem quantities; the spec's raw cluster_max_generation /
+    // max_power_per_unit differ by the same (1 - spinning/100) factor, which
+    // cancels in the formulas above.
+    const auto& disp = paliers.PuissanceDisponibleEtCout[index];
+    const auto& availabilitySeries = disp.PuissanceDisponibleDuPalierThermique;
+    const auto& minGenSeries = disp.PuissanceMinDuPalierThermique;
+    if (static_cast<std::size_t>(pdt) < availabilitySeries.size()
+        && static_cast<std::size_t>(pdt) < minGenSeries.size())
```
guilpier-code:
```c++
if (static_cast<std::size_t>(pdt) < availabilitySeries.size()
    && static_cast<std::size_t>(pdt) < minGenSeries.size())
```
not sure this **if** statement is useful : 
- **availabilitySeries** is sized to 168, insuring that it's always true
- if this condition is not realized, code would crash earlier
- if this is useful, why don't we have this **if** statement each time we use **PuissanceDisponibleDuPalierThermique** ?

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: acb6fafe0e3c05b8e341a220d4e1c37934b35844

flomnes (approved):

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: 827a51e10815e2be33039125654adffa380ef294

PR #3740, opened 2026-06-19
fix(warnings): silence deprecated get_temporary_buffer from stable_sort

JasonMarechal25:
Next `-Werror` warning surfaced after the first four batches merged (seen in the Clang job of #3730, run 27769442216).

## Issue
Under Clang + libstdc++‑12, `std::ranges::stable_sort` (in `InfeasibleProblemReport::filterConstraintsToOneByType`) instantiates the **C++17‑deprecated** `std::get_temporary_buffer<std::shared_ptr<WatchedConstraint>>`. With `-Werror -Wdeprecated-declarations` this is a hard error:

```
.../c++/12/bits/stl_tempbuf.h:263:8: fatal error:
  'get_temporary_buffer<std::shared_ptr<...WatchedConstraint>>' is deprecated [-Wdeprecated-declarations]
```

## Fix
The deprecation originates in the standard‑library internals, not our code, so it's silenced with a scoped `#pragma GCC diagnostic` (honored by both Clang and GCC) around the single call. `std::sort` is **not** a substitute: stability is required because step 2 (`unique`) keeps the first instance of each type group, so the stable ordering determines which constraint is retained.

## Note
This was the only error the `-Werror` build reached (it halts at the first one per TU via `-Wfatal-errors`). Once merged, the Clang job on #3730 should be re-run to confirm a clean build or surface the next warning, if any.

---

### DIFF
--- a/src/solver/infeasible-problem-analysis/report.cpp
+++ b/src/solver/infeasible-problem-analysis/report.cpp
@@ -50,7 +50,14 @@ bool greaterValue(const std::shared_ptr<WatchedConstraint> a, std::shared_ptr<Wa
 void InfeasibleProblemReport::filterConstraintsToOneByType()
 {
     // 1. Grouping constraints by C++ type (inside a group, order of instances remains unchanged)
+    // NOTE: libstdc++'s std::stable_sort instantiates std::get_temporary_buffer, which is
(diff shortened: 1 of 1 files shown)


JasonMarechal25 (commented):
LTGM

flomnes (approved):

PR #3749, opened 2026-06-22
Create a system for the view builder [ANT-5007]

payetvin:
### TODO
- [x] Merge #3783
- [ ] Check for duplicate names between legacy and GEMS components


src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
```diff
@@ -10,6 +10,16 @@
 #include <antares/study/fwd.h>
 #include "antares/solver/optimisation/LegacyVariableInfo.h"
 
+inline const std::string SEP = "::";
+inline const std::string AREA_SEP = "$$";
+inline const std::string LINK("link");
+inline const std::string AREA("area");
```
flomnes, line 16: **Issue**: These were previously file-scoped `const std::string` in opt_rename_problem.cpp (internal linkage, safe). Moving them to namespace-scope `inline const std::string` gives them external linkage and requires dynamic initialization (heap allocation) for every translation unit that includes this header. This introduces potential static initialization order issues and unnecessary runtime overhead.     
                                                                                                                                                                                                            
 **Fix**: Use `inline constexpr std::string_view` for the constants:                                                                                                                                              
                                                                                                                                                                                                            
 ```cpp                                                                                                                                                                                                     
   inline constexpr std::string_view SEP = "::";                                                                                                                                                            
   inline constexpr std::string_view AREA_SEP = "$$";                                                                                                                                                       
   inline constexpr std::string_view LINK = "link";                                                                                                                                                         
   inline constexpr std::string_view AREA = "area";                                                                                                                                                         
 ```

src/view-builder/generateSystemForView.cpp
```diff
@@ -0,0 +1,232 @@
...
+            for (const auto& [mId, model] : library.Models())
+            {
+                if (&model == component.getModel())
+                {
+                    return library.Id() + "." + mId;
+                }
+            }
+        }
+        return "unknown.unknown";
+    };
```
flomnes: Use `Component::getModel` instead of this double linear search
```cpp
class Component final
{
public:
    // ...
    const Model* getModel() const
    {
        return data_.model;
    }
    // ...
};
```

src/view-builder/generateSystemForView.cpp
```diff
@@ -0,0 +1,232 @@
...
+    YAML::Node root;
+    root["system"] = system;
+    return root;
+}
+
+YAML::Node mergeHybridSystemYaml(const Antares::Data::Study& study, const YAML::Node& legacyYaml)
+{
+    auto* modelerData = study.getModelerData();
+    YAML::Node legacyComponents = legacyYaml["system"]["components"];
+    YAML::Node legacyConnections = legacyYaml["system"]["connections"];
```
flomnes: **Issue**: These create deep copies of the YAML nodes from `legacyYaml`. The nodes are only read from afterward, so references would suffice.
**Fix**:
```cpp
const auto& legacyComponents = legacyYaml["system"]["components"];
const auto& legacyConnections = legacyYaml["system"]["connections"];
```

src/view-builder/generateSystemForView.cpp
```diff
@@ -0,0 +1,232 @@
...
+    }
+    bool hasLegacy = false;
+    for (const auto& lib : libs)
+    {
+        if (lib.as<std::string>() == "antares_legacy_models")
+        {
+            hasLegacy = true;
+            break;
+        }
+    }
```
flomnes: **Issue**: O(n) linear scan to check for a single string. While `n` is small here, `std::ranges::any_of` (C++20) is clearer and idiomatic.

**Fix**:
```cpp
#include <ranges>
bool hasLegacy = std::ranges::any_of(libs, [](const auto& lib) {
    return lib.as<std::string>() == "antares_legacy_models";
});
```

src/view-builder/generateSystemForView.cpp
```diff
@@ -0,0 +1,232 @@
...
+            connections.push_back(makeConnection(makeComponentId(area.id, "ShortTermStorage", st.id), "balance_port", areaLoc, "balance_port"));
+        }
+
+        components.push_back(longTermStorageToYaml(area));
+        connections.push_back(makeConnection(makeComponentId(area.id, "Hydro"), "balance_port", areaLoc, "balance_port"));
+
+        for (const auto& [linkName, link] : area.links)
+        {
+            components.push_back(linkToYaml(*link));
+
```
flomnes: **Issue**: `linkName` is destructured but never referenced. This generates a compiler warning with `-Wunused-variable`.

**Fix**:
```cpp
for (const auto& [linkName, link] : area.links)
{
    (void)linkName;
```

src/view-builder/generateSystemForView.cpp
```diff
@@ -0,0 +1,232 @@
...
+    YAML::Node system;
+    system["id"] = "legacy_converted";
+    YAML::Node libs = YAML::Node(YAML::NodeType::Sequence);
+    libs.push_back("antares_legacy_models");
+    system["model-libraries"] = libs;
+
+    YAML::Node components = YAML::Node(YAML::NodeType::Sequence);
+    YAML::Node connections = YAML::Node(YAML::NodeType::Sequence);
+
+    study.areas.each([&](const Area& area)
```
flomnes: Explicitly capture the required scope variables.

src/io/inputs/yml-system/include/antares/io/inputs/yml-system/system.h
```diff
@@ -23,6 +23,7 @@ struct Component
     std::string model;
     std::string scenarioGroup;
     std::vector<Parameter> parameters;
+    std::vector<std::string> properties;
```
flomnes, line 29: What is this newly read field ? Is it documented ?

src/solver/application/application.cpp
```diff
@@ -403,6 +404,11 @@ void Application::execute()
 
     // Save about-the-study files (comments, notes, etc.)
     pStudy->saveAboutTheStudy(*resultWriter);
+
+    YAML::Node viewsYaml = Antares::ViewBuilder::generateSystemForView(*pStudy);
+    std::string yamlContent = YAML::Dump(viewsYaml);
+    resultWriter->addEntryFromBuffer("system-for-views.yml", yamlContent);
```
flomnes: Encapsulate into a function
```cpp
exportSystemView(*pStudy, resultWriter);
```

src/view-builder/legacyToYaml.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+namespace
+{
+
+YAML::Node makeComponent(const std::string& id, const std::string& modelId, const std::vector<std::string>& properties)
+{
+    YAML::Node component;
+    component["id"] = id;
+    component["model"] = modelId;
+    component["parameters"] = YAML::Node(YAML::NodeType::Sequence);
```
flomnes, line 24: Are we sure that we want to export empty parameters ?

src/view-builder/legacyToYaml.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+    std::string result = areaLocation(areaId) + SEP + childType;
+    if (!childId.empty())
+    {
+        result += "<" + childId + ">";
+    }
+    return result;
+}
+
+std::string miscGenTypeName(int index)
```
flomnes, line 91: Use the `enum class` type, not `int`
```suggestion
std::string miscGenTypeName(MiscGenIndex  index)
```

src/view-builder/legacyToYaml.cpp
payetvin: Yes it's in the spec

src/solver/optimisation/include/antares/solver/optimisation/opt_rename_problem.h
payetvin: I will re-use naming from the simulation table, will look into it then

src/view-builder/legacyToYaml.cpp
payetvin: The enum is not an enum class and switching to it would cause changes on several files in the rest of the code

enum in constants.h should be switched to enum class in another PR

src/io/inputs/yml-system/include/antares/io/inputs/yml-system/system.h
payetvin: It is now, I will add after the behavior of the PR is validated

flomnes (approved):

PR #3754, opened 2026-06-22
fix(sanitizer): name FillContext local in EvalVisitor tests (stack-use-after-scope)

JasonMarechal25:
Second batch of **real bugs** from the ASan/UBSan test phase (#3730).

## Issue
ASan reported `stack-use-after-scope` in `FillContext::getYear()` during `_PrintAndEvalNodes_/evaluate_shifted_param`:

```
#0 FillContext::getYear()  ILinearProblemData.h:67
#1 EvalVisitor::visit(ParameterNode const*)  EvalVisitor.cpp:114
...
#8 evaluate_shifted_param::test_method()  test_PrintAndEvalVisitors.cpp:864
```

`EvalVisitor` stores its `FillContext` **by reference** — and that's intentional: production `ComponentFiller` constructs a single visitor and mutates the context across timesteps, relying on the visitor seeing the updates (so switching to by-value is not an option). The bug is in the tests: several passed a **brace-temporary** `FillContext` straight to the constructor. The temporary is destroyed at the end of the constructor expression, so the later `visitor.dispatch()` read a destroyed stack object.

## Fix
Bind the `FillContext` to a named local that outlives the visitor at each affected site (7 sites), matching the pattern already used by production and by the other tests. Only the `evaluate_shifted_param` path tripped ASan (the binary aborts on first error), but all sites shared the same latent dangling-reference pattern, so all are fixed to avoid follow-up failures.

## Note
"Fix real bugs first" track; sanitizers stay blocking. Remaining: the `date.cpp` out-of-bounds month index needs deeper calendar-build tracing (flagging separately), and the OR-Tools `MPObjective` SEGV cluster looks test-harness-related.

---

### DIFF
--- a/src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
+++ b/src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
@@ -849,8 +849,13 @@ EvaluationResult CreateAndEvaluateTimeNode(Node* p)
     MockLinearProblem linearProblem = MockLinearProblem(true);
     OptimEntityContainer optimContainer(linearProblem);
     optimContainer.addFromSystemComponents(components);
+    const Antares::Optimisation::LinearProblemApi::FillContext fillContext{first,
+                                                                           last /*three hours*/,
+                                                                           first,
+                                                                           last,
+                                                                           0};
     EvalVisitor visitor(optimContainer,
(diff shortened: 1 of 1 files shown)


JasonMarechal25:
LTGM

flomnes (approved):

PR #3760, opened 2026-06-25
fix(warning): use override

JasonMarechal25:


src/libs/antares/study/include/antares/study/scenario-builder/RenewableTSNumberData.h
```diff
@@ -16,7 +16,7 @@ class renewableTSNumberData final: public TSNumberData
 public:
     renewableTSNumberData() = default;
 
-    virtual ~renewableTSNumberData()
+    ~renewableTSNumberData() override
     {
     }
```
flomnes: For consistency
```suggestion
    ~renewableTSNumberData() override = default;
```

src/libs/fswalker/filejob.inc.hxx
```diff
@@ -18,7 +18,7 @@ public:
     {
     }
 
-    virtual ~FileJob()
+    ~FileJob() override
     {
     }
```
flomnes: For consistency
```suggestion
    ~FileJob() override = default;
```

flomnes (approved):
LGTM
Minor style remarks

PR #3761, opened 2026-06-25
Extra-outputs : Quick fixes

guilpier-code:
What was done : 
- source files **constraint_builder_utils.cpp** are deleted : their content can be replaced by a constructors in class **ConstraintBuilder**
- renaming in **opt_rename_problem.cpp** : output ==> output (more accurate)
- what else ?


src/solver/optimisation/opt_rename_problem.cpp
guilpier-code, line 1: Here we just move code so that classes methods definitions are together.
Aso renaming : output ==> name (more accurate)

JasonMarechal25:
I wouldn't accept the PR as it is. I'm not against refactoring constraint builder, but it is outside extra-outputs scope, so it should be based on develop.
Also be mindful that there are at least 3 other PR opened for extra-output, meaning that much more conflict to solve.

guilpier-code:
> I wouldn't accept the PR as it is. I'm not against refactoring constraint builder, but it is outside extra-outputs scope, so it should be based on develop. Also be mindful that there are at least 3 other PR opened for extra-output, meaning that much more conflict to solve.

I will base this PR on **develop** ASAP

PR #3776, opened 2026-07-01
Trial for improvement for in opt_rename_problem

guilpier-code:
Make class hierarchy a bit more clear : 
- class **VariableNamer** and class **ConstraintNamer** delegate construction to theit parent (class **Namer**), so they no longer have a constructor on their own
- Method **Namer::RecordLegacyVariableInfo** is shorter and looses the **if** statement it used to have, and that was confusing (**legacyInfo_** could be **null_ptr** in this method)
- clang formatting was applied (and lines were added because of that)


flomnes (changes requested):
The name ON/OFF setting is handled by MPSGenerator since #3692

You can safely remove the single-parameter constructor for `Namer` and method `RecordNothing`.

flomnes:
Too many conflicts

PR #3782, opened 2026-07-02
feat: remove JIT [ANT-4954]

JasonMarechal25:
Lot's of unrelated change due to having to fix namespace leaking


src/libs/antares/array/include/antares/array/matrix.h
```diff
@@ -56,8 +55,6 @@ class Matrix
         optFixedSize = 1,
         //! Do not produce warnings/errors
         optQuiet = 2,
-        //! Do not postpone the loading
```
payetvin, line 59: goes from 2 to 16, not critical but nice to clean

payetvin (approved):

PR #3784, opened 2026-07-03
refacto: move ReserveIndexMap

JasonMarechal25:
Move ReserveIndexMap from runtime.h to study. Does not belong to runtime information

Use shared_ptr to avoid having to include boost::bimap in a header.

boost::bimap is a heavy template expensive at build time.


payetvin (approved):
Fix sonar issues other than implicit conversion before merging

PR #3789, opened 2026-07-03
refactor(yuni): replace Yuni::Job::QueueService with a std thread pool

JasonMarechal25:
## What

Introduce **`Antares::Concurrency::ThreadPool`** — a small std-based worker pool (`std::jthread` + `condition_variable` + FIFO deque of `packaged_task`s, ~70 lines) — and use it everywhere the `Antares::Concurrency` wrapper path used the yuni job queue:

- solver MC-year loop (`ISimulation::pQueueService`)
- zip result-writer io queue (`Application::ioQueueService`, `APIInternal::execute`)
- `resultWriterFactory` / `ZipWriter`
- the concurrency & writer tests

The **writer library is now fully yuni-free** and no longer links `yuni-static-core`.

Out of scope (separate future PR): the raw `QueueService` users — fswalker, vacuum, yby-aggregator — which use custom counter-based draining.

## Behavioural equivalence

- **Exceptions propagate through the returned `std::future`** (the wrapper's whole purpose); a throwing task doesn't kill its worker (new test).
- **`FutureSet` untouched** — first-submitted exception rethrow and join-all semantics stay test-pinned.
- **Deferred start**: tasks added before `start()` are queued but don't run — matching the solver's add-all-then-`start()` sequence with the old QueueService (`add(); …; start(); wait(qseIdle); stop();` becomes `add(); …; start(n);` + the pre-existing `FutureSet::join()`, which was already the real drain).
- **FIFO execution**; the io writer queue keeps its single worker (`maximumThreadCount(1)` → `ThreadPool(1)`), new test pins single-worker FIFO order.

## Simplifications this enables

- **Job priorities dropped**: the only non-default use (`priorityLow` for zip write jobs) targeted the single-worker io queue, where FIFO makes priorities meaningless. No test referenced priorities.
- **`EnsureQueueStartedIfNeeded` deleted**: its single call site always received an already-started queue, making it a provable no-op.
- `nullptr` remains a valid "no pool" argument to `resultWriterFactory` (7 existing call sites); the solver's `NoQueueService` guard is unchanged.

## Known-acceptable deltas

- The solver's idle workers now live until `ISimulation` destruction instead of being joined right after the year loop — not observable.
- A never-started pool destroyed with queued tasks would leave broken-promise futures — impossible at current call sites, documented in the header.

## Tests

The 5 existing concurrency cases pass unchanged (only the pool-creation helper changed). **4 new cases** pin the pool itself: deferred start, destructor drains pending tasks, worker survives a task exception, single-worker FIFO order. The writer zip end-to-end test (2 threads) exercises the pool under real concurrent zip writes.

Part of the incremental removal of the bundled libYuni dependency (follows #3769, #3770, #3774, #3775).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01E6D8kGQNL4xBgwTFgW1WGi)_


src/solver/application/include/antares/application/application.h
```diff
@@ -3,11 +3,12 @@
 
 #pragma once
 
+#include <yuni/core/event/event.h>
```
flomnes: Is yuni's event.h still used ? If not, remove it

Maybe `Application` no longer needs to derive from `Yuni::IEventObserver<Application, Yuni::Policy::SingleThreaded>`

src/libs/antares/concurrency/include/antares/concurrency/concurrency.h
```diff
@@ -4,38 +4,81 @@
...
+ * The pool supports deferred start: tasks added before start() are queued
+ * and only begin executing once start() spawns the workers. The destructor
+ * completes all queued tasks before joining the workers, provided the pool
+ * was started; destroying a never-started pool with queued tasks leaves
+ * their futures with a broken promise.
+ */
+class ThreadPool
+{
+public:
+    ThreadPool() = default;
```
flomnes: The default constructor is kind of dangerous since is doesn't start threads (empty `tasks_`) and there is no possible recovery (if `ThreadPool` is created with this constructor, it's a no-op)

Either `delete` the default ctor or make it `private`.

**EDIT**
OK there is a `start` method, so we can either 
1. start the queue immediately (`ThreadPool(unsigned threadCount)`) 
2. delay it with `ThreadPool() + start(threadCount)`.

I think it would be best to keep only one possibility for clarity.

Since you removed the `ensureQueueStarted` helper, ensuring that the queue is always started wouldn't be a bad idea. And I can't see a reason why we would want a delayed start for a tread pool.

src/solver/application/include/antares/application/application.h
JasonMarechal25: 5b90970b8fbd59f75bd7d9dc6794d9cf0fc5c618

I didn't check for inheritance. To do in another PR

src/libs/antares/concurrency/include/antares/concurrency/concurrency.h
JasonMarechal25: 63a479ab8eb502f8b81fb339f48114582ed68bd5

flomnes (approved):

PR #3792, opened 2026-07-06
Add expression in error log [ANT-4940]

payetvin:
Example log:
`Error while converting library : library-eo: in file 'resources/modeler/10_1/input/model-libraries/library-eo.yml', line 50 : Error while converting expression 'ens_cost * unspplied_energy': No parameter or variable found for this identifier: unspplied_energy`


src/io/inputs/yml-model/decoders.cpp
```diff
@@ -279,6 +294,7 @@ bool convert<YmlModel::Constraint>::decode(const Node& node, YmlModel::Constrain
 
     rhs.id = node["id"].as<std::string>("");
     rhs.expression = node["expression"].as<std::string>();
+    rhs.expressionLineNb = node["expression"].Mark().line + 1;
```
flomnes: Add filename + line number to class `Expression` (metadata) and create a function for these 2 lines
```cpp
rhs.expression = import(node["expression"]);
```

src/io/inputs/model-converter/convertorVisitor.cpp
```diff
@@ -148,7 +150,18 @@ Expressions::NodeRegistry convertExpressionToNode(const std::string& exprStr,
...
+    {
+        root = std::any_cast<Node*>(visitor.visit(tree));
+    }
+    catch (InputError& e)
+    {
+        throw InputError(fmt::format("in file {} : Error while converting expression '{}': {}",
+                                     fileAndLineNb,
+                                     exprStr,
+                                     e.what()));
+    }
```
flomnes, line 172: When `fileAndLineNb` is empty (e.g., in unit tests where `filename = ""`), the formatted message reads:
```
in file  : Error while converting expression 'abc': No parameter or variable found...
```

The double space and missing file info is confusing.

**Fix:** Conditionally include the file context:
```cpp
if (fileAndLineNb.empty())
{
    throw InputError(fmt::format("Error while converting expression '{}': {}",
                                 exprStr,
                                 e.what()));
}
else
{
    throw InputError(fmt::format("in {} : Error while converting expression '{}': {}",
                                 fileAndLineNb,
                                 exprStr,
                                 e.what()));
}
```

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -21,6 +21,18 @@ using namespace Antares::ModelerStudy::SystemModel;
...
 
+namespace
+{
+std::string buildFileAndLineNb(const std::string& filename, size_t lineNb)
+{
+    if (lineNb == 0)
+    {
+        return {};
+    }
+    return fmt::format("'{}', line {}", filename, std::to_string(lineNb));
```
flomnes: `fmt::format` can format integers directly. The `std::to_string` call is redundant and creates an unnecessary temporary `std::string`.

**Fix (C++20):**
```cpp
return fmt::format("'{}', line {}", filename, lineNb);
```

src/io/inputs/model-converter/modelConverter.cpp
```diff
@@ -351,8 +383,9 @@ std::vector<Model> convertModels(const YmlModel::Library& library,
 {
     std::vector<Model> models;
     models.reserve(library.models.size());
-    for (const auto& model: library.models)
+    for (auto model: library.models)
```
flomnes: This forces a copy, probably not something we want

Just pass `library.filename` to `convertVariables`.
```suggestion
    for (auto& model: library.models)
```

src/io/inputs/yml-model/include/antares/io/inputs/yml-model/Library.h
```diff
@@ -50,6 +50,9 @@ struct Variable
     bool time_dependent;
     bool scenario_dependent;
     std::string location;
+
+    size_t lower_bound_lineNb = 0;
+    size_t upper_bound_lineNb = 0;
```
flomnes: Improvement
```cpp
struct ExpressionLineNumber
{
    std::string inputExpr;
    size_t lineNumber;
};
struct Variable
{
    ExpressionLineNumber lower_bound;
    ExpressionLineNumber upper_bound;
};
```
In decoders.cpp
```cpp
bool convert<YmlModel::ExpressionLineNumber>::decode(const Node& node, YmlModel::ExpressionLineNumber& rhs)
{
        if (node.IsDefined() && !node.IsNull())
        {
            rhs.line_number = node.Mark().line + 1;
        } else {
            rhs.line_number = 0;
        }
        rhs.inputExpr = node.as<std::string>("");
}
```
So that you only need to call
```cpp
lower_bound = node["lower_bound"].as<ExpressionLineNumber>();
upper_bound = node["upper_bound"].as<ExpressionLineNumber>();
```

And you can eliminate "weird" `expressionLineNb`, `lower_bound_lineNb`, `upper_bound_lineNb`, etc.

flomnes (changes requested):

flomnes (approved):

PR #3795, opened 2026-07-08
Remove `boost::bimap`, use 2 `std::map` instead

flomnes:


JasonMarechal25:
https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3784

PR #3799, opened 2026-07-09
feat: extra-output  [ANT-5006]

JasonMarechal25:


src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -123,28 +125,48 @@ void LegacyExtraOutputEmitter::areaOutputs(uint32_t pays, int pdt)
...
+                           + problemeHebdo_.AllMustRunGeneration[pdt]
+                               .AllMustRunGenerationOfArea[pays];
+    emit("actual_load", area, pdt, rawLoad);
 
     const double price = areaPrice(pays, pdt);
     emit("price", area, pdt, price);
+    emit("balance_port.price", area, pdt, price);
 
     constexpr double nearLossOfLoadCutoff = 5.;
     emit("is_near_loss_of_load",
```
flomnes, line 192: `cost(unsupplied)` returns `problem_.CoutLineaire[unsupplied]`, which has noise added to avoid same-cost solutions. I think it would be best not to include this numerical trick into the computation of outputs.

@Juliette-Gerbaux what do you think ?

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -123,28 +125,48 @@ void LegacyExtraOutputEmitter::areaOutputs(uint32_t pays, int pdt)
...
+    const double rawLoad = problemeHebdo_.ConsommationsAbattues[pdt].ConsommationAbattueDuPays[pays]
+                           + problemeHebdo_.AllMustRunGeneration[pdt]
+                               .AllMustRunGenerationOfArea[pays];
+    emit("actual_load", area, pdt, rawLoad);
 
     const double price = areaPrice(pays, pdt);
     emit("price", area, pdt, price);
+    emit("balance_port.price", area, pdt, price);
 
     constexpr double nearLossOfLoadCutoff = 5.;
```
flomnes, line 191: Add a comment to explain the value, e.g
```cpp
  // 5 MW threshold for near LoL detection
  constexpr double nearLossOfLoadCutoff = 5.;
```

src/solver/optimisation/LegacyExtraOutputs.cpp
```diff
@@ -280,6 +343,15 @@ void LegacyExtraOutputEmitter::weeklyHydroOutputs(uint32_t pays) const
...
          pdt,
          dual(problemeHebdo_.NumeroDeContrainteExpressionStockFinal[pays]));
+
+    const std::size_t layerCount = problemeHebdo_.NumeroDeVariableDeTrancheDeStock[pays].size();
+    double bellmanValue = 0.;
+    for (std::size_t layer = 0; layer < layerCount; ++layer)
+    {
+        const int layerStorage = variableManager_.LayerStorage(pays, layer);
+        bellmanValue += cost(layerStorage) * x(layerStorage);
+    }
```
flomnes, line 404: In some cases, `layerStorage = -1` (`CaracteristiquesHydrauliques[pays].AccurateWaterValue = false`, see opt_construction_variables_optimisees_lineaire.cpp).

Since `cost` and `x` don't check that their argument is >=0, I strongly suggest adding a guard here to avoid `std::out_of_range` thrown by `std::vector<double>::operator[]`.

src/solver/simulation/include/antares/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -589,6 +598,19 @@ struct PROBLEME_HEBDO
...
+    // simulation-table extra outputs can emit rows for components that are not
+    // LP variables. Outer index: area; availablePower is indexed by hour in
+    // the week.
+    struct INPUT_GENERATION
+    {
+        std::string componentName;
+        std::vector<double> availablePower;
+    };
+
+    std::vector<std::vector<INPUT_GENERATION>> InputGenerationOfArea;
```
flomnes, line 614: Add a comment about indexing here e.g `InputGenerationOfArea[area][timeStep]`.

src/libs/antares/exception/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ source_group("exception" FILES ${SRC_PROJ})
 
 add_library(${PROJ} ${SRC_PROJ})
 add_library(Antares::${PROJ} ALIAS ${PROJ})
+set_target_properties(${PROJ} PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
 
 set_target_properties(${PROJ} PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
flomnes: Those two lines look redundant

flomnes (commented):
Nice job @claude !

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: When in accurate we early return so we wont emit output in such case
```
    if (!problemeHebdo_.CaracteristiquesHydrauliques[pays].AccurateWaterValue)
    {
        return;
    }
```

src/solver/simulation/include/antares/solver/simulation/sim_structure_probleme_economique.h
JasonMarechal25: 0119b6ea33802640f0293c345cd609986caaf740

src/solver/optimisation/LegacyExtraOutputs.cpp
JasonMarechal25: 91adefa77dc2d6259798a9da04757f9be0773670

flomnes (approved):

PR #3801, opened 2026-07-09
fix: improve CSV writing [ANT-5271]

JasonMarechal25:
* Construct output in a single buffer instead of copying strings for each cells
* Iterate over IColumns instead of copying the content of the simulation table in a vector<vector<string>>
* Reworked SurveyResult dumping to use fmt


payetvin (approved):

flomnes:
Remove custom CSV writer changes then merge

PR #3805, opened 2026-07-13
fix: retrieveAverageNTC computed at each time step

JasonMarechal25:
retrieveAverageNTC was computed at each time step


flomnes:
Benefit is unclear, to be investigated

JasonMarechal25:
> Benefit is unclear, to be investigated

The average NTC only depends on the link and on the hour, not on the time step being filled in : compute it once per link for the whole week.

flomnes (approved):

PR #3818, opened 2026-07-17
refactor(yuni): replace Yuni::Logs backend with a std-only implementation

JasonMarechal25:
## What

Replace the `Yuni::Logs` backend of the Antares logging facade (`src/libs/antares/logs`) with a self-contained std-only implementation. `Antares::logs` keeps its exact public surface:

- nine level methods — `fatal/error/warning/checkpoint/notice/progress/info/compatibility/debug()`, each with a `(const U&)` overload
- a streamable `LogBuffer` supporting the same **closed** set of operand types Yuni supported (Yuni's `Append<>` template has no generic definition — its supported-type set is closed and was fully enumerable by inspection, so nothing could silently change format)
- `appendFormat`/`vappendFormat` with identical printf semantics
- `logfile()`/`logfileIsOpened()`/`closeLogfile()` — same append-mode semantics
- a `callback` hook with the same raw-message / skip-empty / skip-debug contract
- `verbosityLevel` gating, `applicationName`, and the thread-local `threadNumber()` that produces the `[app-N]` suffix

**The `logs` library no longer links `yuni-static-core`.**

## Two-commit structure (tests first)

1. **`test(logs): characterize the on-disk/console log format before backend swap`** — a new characterization suite (`log-format-test.cpp`) pinning the exact decorated line format, ANSI console colors + stderr/stdout routing, printf `%f` float formatting, raw (unquoted) `filesystem::path` streaming, and the callback contract. Verified locally to pass against the **old** Yuni-backed implementation first.
2. **`refactor(yuni): replace Yuni::Logs backend...`** — the actual swap. Same test suite, same assertions, passes unchanged against the new implementation.

## Verification (done locally, no vcpkg toolchain needed for the logs lib itself)

Compiled and linked the characterization suite against both the old (yuni) and new (std) backends, in both NDEBUG and debug configurations — **all pass, byte-identical output, zero Yuni objects in the new build.**

## Mechanical fallout from the swap

- `Yuni::Logs::Verbosity::*` / `Yuni::Logs::threadNumber()` → `Antares::Logs::` at their ~10 call sites (`application.cpp`'s `onLogMessage`, `resources.cpp`/`gamma-euler.cpp`'s Debug-enabled guards, `cleaner-v20.cpp`/`in-memory-study.cpp`'s `verbosityLevel` assignments, `solver.hxx`'s thread-number tagging).
- `if (!logs.logfile())` → `if (logs.logfile().empty())` at 3 sites (`logfile()` now returns `std::string`).
- `Application` and the test-only `CaptureAntaresLogs` no longer derive from `Yuni::IEventObserver` — that base only existed to satisfy `Yuni::Event::connect`'s requirements; the new `LogCallback::connect` takes a plain object pointer + member pointer.
- `src/io/file.cpp` had a local `uint` (Yuni's typedef, previously reached transitively through `logs.h → yuni.h`) with no other Yuni include in its chain → `unsigned int`. Audited the other 20 files using `uint` while including `logs.h`: all of them reach `yuni.h` through another Antares header they already include (`study.h`, `economy_base.h → area.h`, etc.), so no other change was needed.
- Dead code removed: `LogCompatibility` (0 callers), the standalone `LogLevel` enum (0 users), `AppendHostName` (0 external callers). `hostname.hxx` is **kept as-is** — it's still used directly by `libs/antares/sys/policy.cpp` with a Yuni string, independently of the logging facade.

Part of the incremental removal of the bundled libYuni dependency (follows #3769, #3770, #3774, #3775, #3789).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01E6D8kGQNL4xBgwTFgW1WGi)_


src/libs/antares/resources/resources.cpp
```diff
@@ -5,6 +5,9 @@
 
 #include <yuni/yuni.h>
 #include <yuni/core/string.h>
+#include <yuni/io/directory.h>
```
payetvin, line 8: Are those includes needed if we remove some yuni objects ?

src/libs/antares/logs/logs.cpp
```diff
@@ -3,29 +3,482 @@
...
+#endif
+};
+
+#ifdef _WIN32
+#define ANTARES_LOGS_WIN_COLOR(attr) attr,
+#else
+#define ANTARES_LOGS_WIN_COLOR(attr)
+#endif
+
+#ifdef _WIN32
```
payetvin, line 46: Lots of specific windows code in this PR, should be tested by @guilpier-code

payetvin (changes requested):
Several sonarcloud issues including some criticals, fix them or explain if they can be ignored

src/tests/src/libs/antares/logs/log-format-test.cpp
```diff
@@ -0,0 +1,471 @@
...
+    std::streambuf* previousBuffer_;
+};
+
+// Saves/restores global logger state, opens a temporary log file
+class LogsUnderTest
+{
+public:
+    LogsUnderTest():
+        workingDir_(
+          generateAndCreateDirName(boost::unit_test::framework::current_test_case().p_name))
```
guilpier-code: _**Out of scope, but easy fix**_

**generateAndCreateDirName(...)** : misleading name, should be **createTmpDirWithName(...)**

src/tests/src/libs/antares/logs/log-format-test.cpp
```diff
@@ -0,0 +1,471 @@
...
+        std::error_code ec;
+        std::filesystem::remove_all(workingDir_, ec);
+    }
+
+    LogsUnderTest(const LogsUnderTest&) = delete;
+    LogsUnderTest& operator=(const LogsUnderTest&) = delete;
+    LogsUnderTest(LogsUnderTest&&) = delete;
+    LogsUnderTest& operator=(LogsUnderTest&&) = delete;
+
+    std::vector<std::string> fileLines() const
```
guilpier-code: **_Cosmetic_**

```c++ 
std::vector<std::string> fileLines() const
```
Could be renamed into : 
```c++ 
std::vector<std::string> readLogFile() const
```
more accurate

src/tests/src/libs/antares/logs/log-format-test.cpp
```diff
@@ -0,0 +1,471 @@
...
+    std::filesystem::path logFilePath_;
+};
+
+} // namespace
+
+BOOST_AUTO_TEST_SUITE(file_line_format)
+
+BOOST_FIXTURE_TEST_CASE(each_level_has_exact_tag_bytes, LogsUnderTest)
+{
+    logs.fatal() << "message";
```
guilpier-code, line 142: Many log levels. Do we need so many ?

For example :
- **progress()** is never used in **Antares**, whereas it should (see next bullet)
- **notice()** : in **Antares**, often used when **progress()** should be used
- **compatibility()** : is never used though we have version compatibility issues
- what's the difference between **notice** and **info** ?

At some point, we should clarify our logging level usage and : 
- change some logging calls into the right logging level
- remove logging levels we don't need

src/libs/antares/logs/logs.cpp
```diff
@@ -3,29 +3,482 @@
...
+                                       ANTARES_LOGS_WIN_COLOR(winDefault)};
+const LevelInfo levelInfoInfo = {Verbosity::Info::level,
+                                 "infos",
+                                 false,
+                                 true,
+                                 nullptr,
+                                 nullptr,
+                                 ANTARES_LOGS_WIN_COLOR(winDefault)
+                                   ANTARES_LOGS_WIN_COLOR(winDefault)};
+const LevelInfo levelInfoCompatibility = {Verbosity::Compatibility::level,
```
guilpier-code: Tag for compatibility is "compatibility notice".
"compatibility" seems enough, and is shorter.

src/tests/src/libs/antares/logs/log-format-test.cpp
```diff
@@ -0,0 +1,471 @@
...
+    checkLine(lines[3], R"(\[test\]\[check\] message)");
+    checkLine(lines[4], R"(\[test\]\[notic\] message)");
+    checkLine(lines[5], R"(\[test\]\[progress\] message)");
+    checkLine(lines[6], R"(\[test\]\[infos\] message)");
+    checkLine(lines[7], R"(\[test\]\[compatibility notice\] message)");
+}
+
+BOOST_FIXTURE_TEST_CASE(no_ansi_escape_bytes_in_file, LogsUnderTest)
+{
+    logs.error() << "colored on console, plain in file";
```
guilpier-code, line 161: It seems that something's wrong here.
Printing in file : no problem.
Printing in console : we get no color, despite what error message says.
So, in **logs.cpp**, we should have : 
```c++
const LevelInfo levelInfoError = {Verbosity::Error::level,
                                  "error",
                                  true,
                                  true,
                                  "\x1b[0;31m",
                                  "\x1b[0;31m",
                                  ANTARES_LOGS_WIN_COLOR(winRed)
                                    ANTARES_LOGS_WIN_COLOR(winRed)};
```

src/libs/antares/logs/include/antares/logs/logs.h
```diff
@@ -4,72 +4,351 @@
...
+    std::string logfile() const;
+    bool logfileIsOpened() const;
+    void closeLogfile();
+
+    void applicationName(std::string name);
+
+    //! Minimum level required for a message to be dispatched (see Verbosity::*::level)
+    int verbosityLevel;
+    //! Connect to be notified of every dispatched message (see LogCallback)
+    LogCallback callback;
```
guilpier-code, line 269:
```c++
LogCallback callback;
```
It seems we don't need this data member, along with the class **LogCallback**.
Indeed, there's some cleaning to do in class **Application** : for example, method `Application::onLogMessage` is no longer called since the removal of legacy **GUI**.
Consequences of this should be, at least :
- in **logs** : 
  -  no need for class **LogCallback** or data member `callback`
  - no need for call to this callback function inside **Logger::dispatch**
 - in class **Application** : 
   - no need for method `Application::onLogMessage`
   - no need for data members **pErrorCount**, **pWarningCount**, **messagesStack**
   - **Application::writeExecutionInfo()**'s body should be shortened

src/libs/antares/resources/resources.cpp
JasonMarechal25: In this PR we still use Yuni so yes for now. Eventually we can remove other Yuni dependencies

src/tests/src/libs/antares/logs/log-format-test.cpp
JasonMarechal25: I've removed progress and compatibility.

I left the other and their usage untouched. Can be discussedin a dedicated issue

ea2909d82a20a0d7e4326cc362b8b3b99abe2bee

src/tests/src/libs/antares/logs/log-format-test.cpp
JasonMarechal25: I don't understand.
I do get color in console.
Capture d&#39;écran 2026-08-18 104344

Only checkpoints message are all colored. Other levels only the tag is colored

src/libs/antares/logs/include/antares/logs/logs.h
JasonMarechal25: This goes beyond the scope of this PR. Let's open a dedicated issue

JasonMarechal25: I think onLogMessage is used during loading to check whether or not we had warning or loading error and abort

src/tests/src/libs/antares/logs/log-format-test.cpp
JasonMarechal25: 47ee0ba389b17666ba80787d441d4c31093a367e

src/tests/src/libs/antares/logs/log-format-test.cpp
JasonMarechal25: a66be4d08b6b800cb1577565291557e286c13a51

src/libs/antares/logs/include/antares/logs/logs.h
guilpier-code: I can't see any place where it's called.
Actually, when removing it (definition and connection to logger : `logs.callback.connect(this, &Application::onLogMessage);`), the **Antares solver** compiles just fine.
So what am I missing ?

src/tests/src/libs/antares/logs/log-format-test.cpp
guilpier-code: When I test it : 
- **tag** is colored correctly in **console** for me (and for you as well)
- **text** is **never** colored in **console**

Is that what we expect ? That **text** is never colored ?

src/libs/antares/logs/include/antares/logs/logs.h
guilpier-code: ok, my bad.
Indeed, while reading input data, we have : 
```c++
logs.callback.connect(this, &Application::onLogMessage);
... // reading input data
logs.callback.clear();
```
So we're sure that the **callback** in no longer in use after this reading step.

src/tests/src/libs/antares/logs/log-format-test.cpp
JasonMarechal25: Only checkpoint level has colored messages

guilpier-code: ok. 
If we want to keep the message (as well as the **tag**) colored in **checkpoint**  logging, maybe we could test it (have a unit test on this) ?

src/tests/src/libs/antares/logs/log-format-test.cpp
```diff
@@ -0,0 +1,466 @@
...
+    BOOST_REQUIRE_EQUAL(lines.size(), 6);
+    checkLine(lines[0], R"(\[test\]\[fatal\] message)");
+    checkLine(lines[1], R"(\[test\]\[error\] message)");
+    checkLine(lines[2], R"(\[test\]\[warns\] message)");
+    checkLine(lines[3], R"(\[test\]\[check\] message)");
+    checkLine(lines[4], R"(\[test\]\[notic\] message)");
+    checkLine(lines[5], R"(\[test\]\[infos\] message)");
+}
+
+BOOST_FIXTURE_TEST_CASE(no_ansi_escape_bytes_in_file, LogsUnderTest)
```
guilpier-code, line 159: Test **no_ansi_escape_bytes_in_file** : very technical name for a test. Could be renamed (for example) : **logging_an_error_in_file_implies_no_color_byte**

guilpier-code (approved):

payetvin (approved):

PR #3829, opened 2026-07-22
[ANT-5179] Update annual costs in reserve cucumber tests

flomnes:
OV COST = SOME COST + RES PARTICIPATION COST

There was a bug before that lead to RES PARTICIPATION COST being computed to 0. Since we fixed this bug, we also need to update the reference OV COST.


src/tests/cucumber/features/solver-features/reserves_tests.feature
```diff
@@ -31,7 +31,7 @@ Feature: reserves tests
     And in area "AREA", during year 1, for cluster "thermal_all_cheap" and reserve "Res_1", reserve participation power is always equal to 20 MWh
     And in area "AREA", during year 1, for cluster "thermal_expensive_prod" and reserve "Res_1", reserve participation power is always equal to 20 MWh
     And in area "AREA", during year 1, for cluster "thermal_expensive_res_part" and reserve "Res_1", reserve participation power is always equal to 0 MWh
-    And the annual system cost is 7.04928e+07
+    And the annual system cost is 7.04995e+07
```
flomnes, line 34: NEW VALUE - OLD VALUE = 6700 (EUR)
RESERVE PARTICIPATION COST = 6720 (EUR)

This is consistent. The 20 EUR difference is caused by rounding error.

src/tests/cucumber/features/solver-features/reserves_tests.feature
guilpier-code, line 20: **l'odre** ==> **l'ordre**

guilpier-code (approved):

PR #3835, opened 2026-07-27
Remove unused function storageIntoRows

payetvin:


flomnes (approved):
You can also remove `enum VarOffset` from test_legacy_extra_outputs.cpp, it is also unused

PR #3836, opened 2026-07-28
MPS comparison - use pytest.approx, relative tol=1e-6

flomnes:


src/tests/run-study-tests/check_on_results/compare_mps_files.py
```diff
@@ -7,8 +6,7 @@
 from utils.find_output import find_dated_output_folder, find_simulation_folder
 from utils.assertions import check
 from shared_utils.mps_utils import *
-
-tol = 1e-12
+from pytest import approx
```
guilpier-code, line 9: **_out of scope, but very easy to fix_**

```python
def compare_files(self, pair, ref_mps, out_mps):
    ...
```
This function do not use its first argument (pair). 
We should remove it.

guilpier-code (approved):

PR #3837, opened 2026-07-28
Increase Simulation table disk storage and writing speed

payetvin:
ST components will be in lower case and this avoids using a function to lower case each time we add a component


flomnes:
@payetvin What is this for ?

YoshuaLisasi (approved):

src/io/outputs/include/antares/io/outputs/columns.h
```diff
@@ -7,11 +7,11 @@
 #include <limits>
 #include <optional>
 #include <string>
-#include <unordered_map>
 #include <vector>
 
+#include <boost/unordered/unordered_flat_map.hpp>
```
flomnes, line 12: Add dependency on Boost::unordered

https://vcpkg.io/en/package/boost-unordered

https://www.boost.org/doc/libs/latest/libs/unordered/doc/html/unordered/reference/unordered_flat_map.html

flomnes (changes requested):

src/io/outputs/include/antares/io/outputs/columns.h
payetvin: done

flomnes (commented):
Add boost::unordered to CMake deps

flomnes (approved):

PR #3839, opened 2026-07-28
Remove SHARED simulation-table-writers, export symbols, etc.

flomnes:
Having SHARED `simulation-table-writers` causes double free errors because of static symbols. We try to have it as a static library instead (the default)

## Before
image

## After
image

### DIFF
--- a/.github/workflows/oracle8.yml
+++ b/.github/workflows/oracle8.yml
@@ -9,6 +9,7 @@ on:
     - dependabot/*
     - release/*
     - copilot/*
+    - fix/rm-shared # TODO remove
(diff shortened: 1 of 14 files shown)


YoshuaLisasi (approved):

.github/workflows/oracle8.yml
```diff
@@ -9,6 +9,7 @@ on:
     - dependabot/*
     - release/*
     - copilot/*
+    - fix/rm-shared # TODO remove
```
flomnes:
```suggestion
```

PR #3841, opened 2026-07-29
Fix: change node to load and lower component name [ANT-5250]

YoshuaLisasi:
feat: apply ST naming convention to legacy component IDs [ANT-5250]

Component names now follow the {area}_{type}_{cluster} pattern to
guarantee uniqueness in the Simulation Table (e.g. area_load,
area_thermal_cluster, area_hydro_storage, area1_area2_link).
All names are lowercase. Fixes non-deterministic link naming
caused by const char* comparison.


payetvin:
done in #3799 and #3847

flomnes:
@payetvin Can you check the result is identical ?

PR #3843, opened 2026-07-30
Export system-for-views in full GEMS

payetvin:


src/modeler/main.cpp
```diff
@@ -37,6 +37,13 @@ TableFormat getTableFormat(int argc, const char** argv)
     return TableFormat::CSV;
 }
 
+void exportSystemForViews(fs::path studyPath, fs::path outputPath)
+{
+    fs::path systemInput = studyPath / "input" / "system.yml";
+    fs::path systemOutput = outputPath / "system-for-views.yml";
+    fs::copy(systemInput, systemOutput);
+}
```
flomnes: Handle errors (e.g source file doesn't exist, can't write)

flomnes (commented):
Add
- unit tests (add checks to existing tests)
- documentation

flomnes (approved):

src/modeler/main.cpp
payetvin: done

PR #3850, opened 2026-08-06
fixed race condition in tests-simle-study

YoshuaLisasi:
**Context**

The replacement of QueueService by a ThreadPool introduced an intermittent segfault (~2% of runs) in the parallel test case.

**Root cause**

TestingSimulationObserver::notifyHebdoProblem is called concurrently by multiple worker threads and performs unsynchronized insertions into a shared std::map. When two threads insert simultaneously, the internal red-black tree gets corrupted during rebalancing, causing a segfault in std::_Rb_tree_insert_and_rebalance.

This was not an issue with the previous QueueService, which serialized task execution. The ThreadPool runs MC years truly in parallel, exposing the missing synchronization.

**Fix**

Add a std::mutex to TestingSimulationObserver and lock it in notifyHebdoProblem.

**Diagnosis tools used**
Helgrind to rule out data races in the ThreadPool and solver.hxx synchronization logic
GDB with --catch_system_errors=no to capture the crash backtrace
Stress testing (500 runs loop) to confirm the fix

### DIFF
--- a/src/tests/inmemory-study/in-memory-study.cpp
+++ b/src/tests/inmemory-study/in-memory-study.cpp
@@ -280,6 +280,7 @@ void TestingSimulationObserver::notifyHebdoProblem(const PROBLEME_HEBDO& problem
                                                    int optimizationNumber,
                                                    std::string_view name)
 {
+    std::lock_guard<std::mutex> lock(mutex);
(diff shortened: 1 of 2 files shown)


flomnes (approved):
well done 👍🏻

PR #3853, opened 2026-08-12
Feat: add gems to adqpatch [ANT-5264]

YoshuaLisasi:
Add GEMS contribution to adequacy patch CSR bounds

In hybrid studies (Legacy + GEMS), the CSR adequacy patch did not account for GEMS components when computing ENS bounds and constraint RHS values. This PR adds a helper computeGemsContributionForArea() that loops over GEMS components connected to each area via portToAreaConnections, retrieves the port field expression, and evaluates it using EvalVisitor.

Modified constraints:

setBoundsOnENS() — adds unsupplied_energy_bound to Xmax
setRHSfictitiousLoadValue() — adds spillage_bound to RHS
setRHSMaxEnsLoadValue() — adds unsupplied_energy_bound to RHS


src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
 #include "antares/solver/optimisation/opt_structure_probleme_a_resoudre.h"
 #include "antares/solver/simulation/adequacy_patch_runtime_data.h"
 #include "antares/solver/simulation/sim_structure_probleme_economique.h"
 
 using namespace Yuni;
 
+double computeGemsContributionForArea(const PROBLEME_HEBDO* problemeHebdo,
+                                      uint32_t area,
+                                      int triggeredHour,
+                                      const std::string& portFieldName)
```
flomnes: While this works, it's a bit inefficient. What would be best is
1. Get the list of trigerred hours
2. Evaluate the expression **once** on those timesteps, return a list of values

But step #2 requires that we allow `EvalVisitor` to evaluate an expression on non-consecutive timesteps, which it can't right now.

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
+#include <antares/solver/modeler/ModelerData.h>
+#include <antares/study/system-model/component.h>
+#include <antares/study/system-model/portType.h>
+#include <antares/study/system-model/system.h>
 #include "antares/solver/optimisation/opt_fonctions.h"
 #include "antares/solver/optimisation/opt_structure_probleme_a_resoudre.h"
 #include "antares/solver/simulation/adequacy_patch_runtime_data.h"
 #include "antares/solver/simulation/sim_structure_probleme_economique.h"
 
 using namespace Yuni;
```
guilpier-code:
```c++
using namespace Yuni;
```
We should remove this, useless

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
+              triggeredHour + problemeHebdo->HeureDansLAnnee, // globalLast
+              problemeHebdo->year);
+            Expressions::Visitors::EvalVisitor evalVisitor(*optimEntityContainer,
+                                                           fillContext,
+                                                           component,
+                                                           modelerData->dataSeries.get(),
+                                                           scenario);
+            contribution += evalVisitor.dispatch(expressionNode).valuesAsVector()[0];
+
+            (void)expressionNode;
```
guilpier-code:
```c++
(void)expressionNode;
```
seems useless

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
```diff
@@ -175,9 +180,13 @@ void HourlyCSRProblem::setRHSMaxEnsLoadValue()
...
 
-                if (load >= 0.)
+                double gemsEnsBound = computeGemsContributionForArea(problemeHebdo_,
+                                                                     Area,
+                                                                     triggeredHour,
+                                                                     "unsupplied_energy_bound");
+                if (load + gemsEnsBound >= 0.)
                 {
-                    SecondMembre[Cnt] = load + 1e-5;
+                    SecondMembre[Cnt] = load + gemsEnsBound + 1e-5;
```
flomnes: Just increase `load` by `computeGemsContributionForArea(...)`
```suggestion
                load += computeGemsContributionForArea(problemeHebdo_,
                                                                     Area,
                                                                     triggeredHour,
                                                                     "unsupplied_energy_bound");
                if (load >= 0.)
{
                    SecondMembre[Cnt] = load + 1e-5;
```

src/solver/optimisation/include/antares/solver/optimisation/adequacy_patch_csr/hourly_csr_problem.h
```diff
@@ -46,6 +46,12 @@ struct LinkVariable
 };
 
 struct PROBLEME_HEBDO;
+// GEMS contribution for hybrid studies
+double computeGemsContributionForArea(const PROBLEME_HEBDO* problemeHebdo,
+                                      uint32_t area,
+                                      int triggeredHour,
+                                      const std::string& portFieldName);
+;
```
flomnes: stray semicolon
```suggestion
```

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
+              triggeredHour + problemeHebdo->HeureDansLAnnee, // globalFirst
+              triggeredHour + problemeHebdo->HeureDansLAnnee, // globalLast
+              problemeHebdo->year);
+            Expressions::Visitors::EvalVisitor evalVisitor(*optimEntityContainer,
+                                                           fillContext,
+                                                           component,
+                                                           modelerData->dataSeries.get(),
+                                                           scenario);
+            contribution += evalVisitor.dispatch(expressionNode).valuesAsVector()[0];
+
```
flomnes: 1. The expression may not evaluate to a vector (e.g the port has a literal value such as `44`), in which case an `EvalResultTypeError` exception will be thrown (unintended)
2. In case it does, we want the value at hour `triggeredHour`

```suggestion
            contribution += evalVisitor.dispatch(expressionNode).value(triggeredHour);

```

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
flomnes:
```suggestion
```

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -23,8 +104,13 @@ void HourlyCSRProblem::setBoundsOnENS()
...
-            problemeAResoudre_.Xmax[var] = problemeHebdo_->ResultatsHoraires[area]
-                                             .ValeursHorairesDENS[triggeredHour]
+            double ensLegacy = problemeHebdo_->ResultatsHoraires[area]
+                                 .ValeursHorairesDENS[triggeredHour];
+            double gemsContribution = computeGemsContributionForArea(problemeHebdo_,
+                                                                     area,
+                                                                     triggeredHour,
+                                                                     "unsupplied_energy_bound");
+            problemeAResoudre_.Xmax[var] = ensLegacy + gemsContribution
                                            + belowThisThresholdSetToZero;
```
flomnes: May result in Xmax < Xmin, infeasible quadratic problem

clamp


```suggestion
            problemeAResoudre_.Xmax[var] = std::max(problemeAResoudre_.Xmin[var], ensLegacy + gemsContribution + belowThisThresholdSetToZero);
```

src/solver/optimisation/opt_appel_solveur_lineaire.cpp
```diff
@@ -222,6 +222,7 @@ static SimplexResult OPT_TryToCallSimplex(const SingleOptimOptions& options,
     const ILinearProblemData* modelerDataSeries = hasModelerData ? modelerData->dataSeries.get()
                                                                  : nullptr;
     OptimEntityContainer optimEntityContainer(*ortoolsProblem);
+    problemeHebdo->optimEntityContainer = &optimEntityContainer;
```
flomnes: This is really bad, `optimEntityContainer` object will be destroyed, but it's address kept. Possible dangling pointer

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,193 @@
...
+    problem.modelerData = nullptr;
+    problem.NomsDesPays.push_back("area1");
+
+    double result = computeGemsContributionForArea(&problem, 0, 0, "unsupplied_energy_bound");
+    BOOST_CHECK_EQUAL(result, 0.0);
+}
+
+BOOST_AUTO_TEST_CASE(no_optimEntityContainer_returns_zero)
+{
+    GemsContributionFixture f;
```
flomnes: Use `BOOST_FIXTURE_TEST_CASE`

flomnes (changes requested):

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
+                continue;
+            }
+
+            auto* expressionNode = component.nodeAtPortField(portId, fieldId);
+
+            auto* optimEntityContainer = problemeHebdo->optimEntityContainer;
+            if (!optimEntityContainer)
+            {
+                continue;
+            }
```
guilpier-code: Checks inside function **computeGemsContributionForArea(...)** : 
- **First** : what could be enhanced
- **Second** : suggestions for improvement

**What could be enhanced** : 
- `optimEntityContainer` : if this data is **nullptr**, we are in trouble : we're supposed to never have this case. So this should have been checked earlier in the **adq patch**, and this should raise an error. So it's a first reason for which this check should not be done in this function
- Besides, we're doing this (previous) check multiple times : at each area.
- Checking **modelerData** and the associated **system** are not null could be moved up : if there's no GEMS use, we should escape its contribution earlier in the code.

**Suggestions** : 
- Removing check on `optimEntityContainer` from **computeGemsContributionForArea(...)**, and move this check (for example) to the class **HourlyCSRProblem**'s constructor body. If this check is negative (`nullptr`), raise an error instead of doing nothing
- In order to have more clarity and only one check over **modelerData**, we could split **HourlyCSRProblem::setBoundsOnENS()** in 2 smaller functions : 
```c++
void HourlyCSRProblem::setBoundsOnENS()
{
    setENSboundsFromLegagy();

    // If there no modelerData, no need to go further.
    auto* modelerData = problemeHebdo_->modelerData;
    if (!modelerData || !modelerData->system)
    {
        setENSboundsFromGEMS()
    }
}
```
> So we would make a loop over areas inside each of these 2 functions. And **computeGemsContributionForArea** gets simpler without checks.

> I would even go further and define a new **private** data member for class **HourlyCSRProblem** : `bool GemsUse_ = !modelerData || !modelerData->system`. This will prove to be useful in next comments
- **Cosmetic** : instead of having : 
```c++
const std::string areaName = problemeHebdo->NomsDesPays[area];

for (const auto& component: modelerData->system->Components())
{
    for (const auto& [portId, connectedArea]: component.portToAreaConnections())
    {
        if (connectedArea != areaName)
        {
               continue;
        }
        ...
    }
}
```
We could have : 
```c++
const std::string areaName = problemeHebdo->NomsDesPays[area];
const auto isConnectedToArea = [&](const auto& p) { return p.second == areaName;};
const auto filterPort = std::views::filter(isConnectedToArea) | std::views::keys;
...
for (const auto& component: modelerData->system->Components())
{
    for (const auto& portId : component.portToAreaConnections() | filterPort)
    {
        ...
    }
}
```
More c++ 20 style, and maybe a bit more readable. Up to you.

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
+
+            const auto& scenario = modelerData->scenarioGroupRepository.scenario(
+              component.getScenarioGroupId());
+            Antares::LinearProblem::Api::FillContext fillContext(
```
guilpier-code: This **fillContext** is created for every **triggered hour**, every **area**, and possibly every **component** and **area connection**.
But it seems that it could be created once for a given **triggered hour** (that is also an instance of **HourlyCSRProblem**).
For example, it could be created as an `std::optional`, given that the `GemsUse_` (defined above) is **true**.
Supposing it's created as a **private** data member of class **HourlyCSRProblem**, it could be passed to function **computeGemsContributionForArea(...)**, except if we make this function a method (please see next comment)

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
+#include <antares/study/system-model/portType.h>
+#include <antares/study/system-model/system.h>
 #include "antares/solver/optimisation/opt_fonctions.h"
 #include "antares/solver/optimisation/opt_structure_probleme_a_resoudre.h"
 #include "antares/solver/simulation/adequacy_patch_runtime_data.h"
 #include "antares/solver/simulation/sim_structure_probleme_economique.h"
 
 using namespace Yuni;
 
+double computeGemsContributionForArea(const PROBLEME_HEBDO* problemeHebdo,
```
guilpier-code: Still about **computeGemsContributionForArea(...)** : 
given that 2 of it's arguments (**PROBLEME_HEBDO** and **triggeredHour**) are actually data members of class  **HourlyCSRProblem**, it would make sense to turn this function into a **HourlyCSRProblem**'s method.
It would make these 2 args disappear.

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -3,13 +3,94 @@
...
+            }
+            else if (portFieldName == "spillage_bound")
+            {
+                fieldId = areaConnection->spillage_bound;
+            }
+
+            if (fieldId.empty())
+            {
+                continue;
+            }
```
guilpier-code: I'm not a fan of this block **if... else if... if...**. We could make function (or method ?) a bit more generic, and smaller. Consider this : 
```c++
double gemsContributionForArea(...
                               const std::function<std::string(const AreaConnection&)> & getFieldId)
{
    ...
    for (... /* component */)
    {
        for (... /* portId */)
        {
             ...
             std::string fieldId = getFieldId(areaConnection);
             if (fieldId.empty())
             {
                 continue;
             }
             ...
        }
    }
    return contribution;
}

// Defined in portType.h/.cpp
std::string getUnsupEnergyBound(const AreaConnection& ac)
{
    return ac.unsupplied_energy_bound;
}

// Defined in portType.h/.cpp
std::string getSpilledBound(const AreaConnection& ac)
{
    return ac.spillage_bound;
}

// Defined in portType.h/.cpp
double gemsUnsupEnergyForArea(const PROBLEME_HEBDO* problemeHebdo,
                              uint32_t area,
                              int triggeredHour)
{
    return gemsContributionForArea(problemeHebdo, area, triggeredHour, getUnsupEnergyBound);
}

double gemsSpilledForArea(const PROBLEME_HEBDO* problemeHebdo,
                          uint32_t area,
                          int triggeredHour)
{
    return gemsContributionForArea(problemeHebdo, area, triggeredHour, getSpilledBound);
}
```
ok, code is a bit longer now in total, but : 
- **computeGemsContributionForArea** gets even shorter
- **GemsUnsupEnergyForArea** and **GemsSpilledForArea**, as functions actually called by  class **HourlyCSRProblem** are more specific and loose an argument.
- functions **getSpilledBound** and **GemsUnsupEnergyForArea** should be declared / defined in **portType.h/.cpp** (where `struct AreaConnection` is defined), not in the adequacy patch code area, to be possibly reused. So we get 3 functions instead of one.
- note that I shortened **computeGemsContributionForArea** into **gemsContributionForArea**
- note that, if we take [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3853#issuecomment-5506827800) into account, this 3 last functions are turned into methods, which reduces their input arguments to **area** and a **std::function**

src/solver/optimisation/adequacy_patch_csr/construct_problem_constraints_RHS.cpp
```diff
@@ -142,7 +142,12 @@ void HourlyCSRProblem::setRHSfictitiousLoadValue()
                 }
 
                 // RHS = STt - (1-BT)*STmint + BH*Ht + BF*(Ft - Lt) + BH*STS_net_production
-                double rhs = stt - stmint + ht + bfTerm + stsNetProduction;
+                // setRHSfictitiousLoadValue()
```
guilpier-code:
```c++
// setRHSfictitiousLoadValue()
```
to be removed

guilpier-code:
I also think we could separate more deeply **Legacy** and **Gems** contribution to **adq patch**.
It would be nice to have, at the top of class **HourlyCSRProblem**, something like : 
```c++
void HourlyCSRProblem::setVariableBounds()
{
    ...
    setBoundsOnENS();
    gemsPart->setBoundsOnENS();
    ...
}

void HourlyCSRProblem::buildProblemConstraintsRHS()
{
    ...
    
    setRHSfictitiousLoadValue();
    gemsPart_->setRHSfictitiousLoadValue();

    setRHSbindingConstraintsValue();
    gemsPart_->setRHSbindingConstraintsValue();
    ...
}
```
Note that : 
- **gemsPart_** would be a private data member of **HourlyCSRProblem**, and would be instance of a class **GemsPart**
- Class **GemsPart** could be a **friend** of **HourlyCSRProblem**, so it could have access to the latter class data member , if necessary (I think of the triggered hour for instance). But we could also keep them completely separated (no **friend** key word), and create the **GemsPart** instance with a bit more input data. It would allow to test **GemsPart** class more easily.
- **gemsPart_** could do nothing if **GEMS** does not exist (pure legacy study). In that case **gemsPart_** could be a null object. So, depending on if **GEMS** part exists (`!modelerData || !modelerData->system` is **true** or **false**), `gemsPart_` would hold a **null** or **active** object
- This class would be the place to create a **fillContext_** (as a private data member) useful only for GEMS contributions
- This class would be defined in a separate pair of **gems-part.h/.cpp**

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
guilpier-code: This function is called here and also called in another source file.
It would make sense to declare / define it in a separate pair of source file.
Please see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3853#issuecomment-5506827800) as well (about **GEMS** contribution as a class)

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -2,30 +2,183 @@
...
+#include <antares/study/system-model/component.h>
+#include <antares/study/system-model/portType.h>
+#include <antares/study/system-model/system.h>
 #include "antares/solver/optimisation/opt_fonctions.h"
 #include "antares/solver/optimisation/opt_structure_probleme_a_resoudre.h"
 #include "antares/solver/simulation/adequacy_patch_runtime_data.h"
 #include "antares/solver/simulation/sim_structure_probleme_economique.h"
 
-using namespace Yuni;
+// double HourlyCSRProblem::computeGemsContributionForArea(uint32_t area,const std::string&
```
guilpier-code: We should remove this commented code

guilpier-code (approved):
Remaining unresolved conversations from @guilpier-code are taken care of in [PR ](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3927)

src/solver/optimisation/adequacy_patch_csr/set_variable_boundaries.cpp
```diff
@@ -2,30 +2,183 @@
...
+//                                                            fillContext,
+//                                                            component,
+//                                                            modelerData->dataSeries.get(),
+//                                                            scenario);
+//             contribution += evalVisitor.dispatch(expressionNode).value(triggeredHour);
+
+//         }
+//     }
+//     return contribution;
+// }
```
flomnes: Remove this commented code

src/solver/simulation/include/antares/solver/simulation/sim_structure_probleme_economique.h
```diff
@@ -775,6 +776,8 @@ struct PROBLEME_HEBDO
 
     // TODO: 1 study but several PROBLEME_HEBDO, may cause race conditions
     Antares::Solver::ModelerData* modelerData = nullptr;
+    std::unique_ptr<Antares::LinearProblem::OptimEntityContainer> optimEntityContainer;
+    std::shared_ptr<Antares::LinearProblem::Api::ILinearProblem> ortoolsProblem;
```
flomnes: Parallel lifecycle here, this is pretty bad

See #3922 for explanation

flomnes (approved):

src/solver/optimisation/adequacy_patch_csr/gems-part.cpp
```diff
@@ -0,0 +1,161 @@
...
+            auto* expression = component.nodeAtPortField(portId, fieldId);
+
+            const auto& scenario = modelerData->scenarioGroupRepository.scenario(
+              component.getScenarioGroupId());
+            Visitors::EvalVisitor evalVisitor(*problemeHebdo_->optimEntityContainer,
+                                              fillContext_,
+                                              component,
+                                              modelerData->dataSeries.get(),
+                                              scenario);
+            contribution += evalVisitor.dispatch(expression).value(0);
```
flomnes, line 90: Ajouter un commentaire ici, car c'est un point assez délicat

On évalue bien l'expression au pas de temps `triggeredHour_`, et pas au premier instant. Cette information est "cachée" dans `fillContext_`.

src/solver/optimisation/adequacy_patch_csr/gems-part.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+void ActiveGemsPart::setRHSMaxEnsLoadValue()
+{
+    for (uint32_t area = 0; area < problemeHebdo_->NombreDePays; ++area)
+    {
+        if (problemeHebdo_->adequacyPatchRuntimeData->areaMode[area] == physicalAreaInsideAdqPatch)
+        {
+            auto it = constraintCsrMaxEnsLoad_.find(area);
+            if (it != constraintCsrMaxEnsLoad_.end())
```
flomnes: style
```suggestion
            if (auto it = constraintCsrMaxEnsLoad_.find(area); it != constraintCsrMaxEnsLoad_.end())
```

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
flomnes, line 1: 1. Beaucoup de duplication et de code inutile ("boilerplate")
2. On ne teste que l'absence de levée d'exception, pas la valeur numérique des bornes. Pour tester la fonctionnalité, il faut tester la valeur numériques des bornes dans différents cas :
  - conso résiduelle < 0, conso GEMS = 0
  - conso résiduelle > 0, conso GEMS > 0
  - combinaisons des différents cas

guilpier-code: In these tests on **ActiveGemsPart**, public methods effect is not really tested.
For instance, the effect of **ActiveGemsPart::setBoundsOnENS()** on unsupplied energy max bound could be tested.

src/solver/optimisation/adequacy_patch_csr/gems-part.cpp
```diff
@@ -0,0 +1,165 @@
...
+                int Cnt = it->second;
+                problemeAResoudre_.SecondMembre[Cnt] += gemsUnsupEnergyForArea(area);
+            }
+        }
+    }
+}
+
+double ActiveGemsPart::gemsSpilledForArea(uint32_t area) const
+{
+    return gemsContributionForArea(area, Antares::ModelerStudy::SystemModel::getSpilledBound);
```
guilpier-code: We should remove **Antares::ModelerStudy::SystemModel::**

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+        ScenarioGroupRepository scenarioGroupRepository;
+        auto scenario = std::make_unique<Scenario>("SG");
+        scenario->setTimeSerieNumber(0, 1);
+        scenarioGroupRepository.addScenario("SG", std::move(scenario));
+        modelerData->scenarioGroupRepository = std::move(scenarioGroupRepository);
+
+        modelerData->dataSeries = std::make_unique<LinearProblemData>();
+
+        auto linearProblem = std::make_shared<StructuredLinearProblem>();
+        problemeHebdo.ortoolsProblem_ = linearProblem;
```
guilpier-code:
```c++
problemeHebdo.ortoolsProblem_ = linearProblem;
```
Data member **ortoolsProblem_** is never used in the code base. 
It should be removed. 
And of course we should remove this line.

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+        ScenarioGroupRepository scenarioGroupRepository;
+        auto scenario = std::make_unique<Scenario>("SG");
+        scenario->setTimeSerieNumber(0, 1);
+        scenarioGroupRepository.addScenario("SG", std::move(scenario));
+        modelerData->scenarioGroupRepository = std::move(scenarioGroupRepository);
+
+        modelerData->dataSeries = std::make_unique<LinearProblemData>();
+
+        auto linearProblem = std::make_shared<StructuredLinearProblem>();
```
guilpier-code:
```c++
auto linearProblem = std::make_shared<StructuredLinearProblem>();
problemeHebdo.optimEntityContainer = std::make_shared<OptimEntityContainer>(*linearProblem);
```
To be noted :
- **linearProblem** : 
  - remains empty for our tests (never filled)
  - only used for building an empty **OptimEntityContainer**
  - should be renamed into **emptyLinearProblem**
- This can be explained : For our **GEMS** models / components, the expressions we evaluate to adapt **adq patch CSR hourly problems** are based on components parameters. These parameters don't need an **OptimEntityContainer** to be evaluated. If we decided to evaluate expressions  based on component **variables** (given as a solution of a linear problem), we would need non empty **OptimEntityContainer** and **linearProblem**.
- The previous consideration deserves a comment in tests code, in my opinion.

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+    // Sets the values of the GEMS parameters (one time series per parameter, one value per hour).
+    void setGemsParameters(const std::vector<double>& residualLoad, const std::vector<double>& load)
+    {
+        auto data = std::make_unique<LinearProblemData>();
+        addTimeSeries(*data, "residual_load", residualLoad);
+        addTimeSeries(*data, "load", load);
+        modelerData->dataSeries = std::move(data);
+    }
+
+    std::unique_ptr<Solver::ModelerData> buildModelerSystem()
```
guilpier-code: **buildModelerSystem()** : this function returns a **ModelerData** instance, not a **System**;
It should be renamed into **buildModelerData()**

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+        problemAResoudre.Xmax.resize(20, 0.0);
+        problemAResoudre.Xmin.resize(20, 0.0);
+        problemAResoudre.SecondMembre.resize(20, 0.0);
+
+        ScenarioGroupRepository scenarioGroupRepository;
+        auto scenario = std::make_unique<Scenario>("SG");
+        scenario->setTimeSerieNumber(0, 1);
+        scenarioGroupRepository.addScenario("SG", std::move(scenario));
+        modelerData->scenarioGroupRepository = std::move(scenarioGroupRepository);
```
guilpier-code:
```c++
ScenarioGroupRepository scenarioGroupRepository;
auto scenario = std::make_unique<Scenario>("SG");
scenario->setTimeSerieNumber(0, 1);
scenarioGroupRepository.addScenario("SG", std::move(scenario));
modelerData->scenarioGroupRepository = std::move(scenarioGroupRepository);
```
Here we actually complete **modelerData**.
So this code could be moved to **buildModelerSystem()** (or **buildModelerData()** as recommended [here](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3853/changes#r4035108112)).

If we do this, **buildModelerSystem()** would have 2 parts : 
```c++
std::unique_ptr<Solver::ModelerData> buildModelerSystem()
{
    data->system = createSystemFromYml();
    data->scenarioGroupRepository = createScenarioGroupRepo();
}
```
Clear, and simple.

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+const int maxEnsLoadArea2 = 13;
+} // namespace
+
+struct GemsContributionFixture
+{
+    GemsContributionFixture()
+    {
+        modelerData = buildModelerSystem();
+
+        problemeHebdo.modelerData = modelerData.get();
```
guilpier-code: - we should gather all lines about populating **problemeHebdo** in a single call : 
```c++
problemeHebdo = makeWeeklyProblem();
```
- same for **problemAResoudre** (**makeProblemToSolve()**)

Eventually the **GemsContributionFixture**'s constructor gets short and much more clear.

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+BOOST_AUTO_TEST_SUITE(gems_contribution_tests)
+
+// Residual load > 0: the spillage bound increases the RHS of the fictitious load constraint.
+BOOST_FIXTURE_TEST_CASE(spillage_bound_with_positive_residual_load, GemsContributionFixture)
+{
+    // residual_load = 10 -> 2 * 10 + 30 = 50
+    // load = 20          -> 20 / 2 - 10 = 0
+    setGemsParameters({10.0, 10.0}, {20.0, 20.0});
+    setLegacyBounds();
```
guilpier-code:
```c++
setLegacyBounds();
```
Called in every tests (code duplication).
Could be moved to **makeProblemToSolve()** : see [this comment](https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3853/changes#r4035242516)

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+};
+
+BOOST_AUTO_TEST_SUITE(gems_contribution_tests)
+
+// Residual load > 0: the spillage bound increases the RHS of the fictitious load constraint.
+BOOST_FIXTURE_TEST_CASE(spillage_bound_with_positive_residual_load, GemsContributionFixture)
+{
+    // residual_load = 10 -> 2 * 10 + 30 = 50
+    // load = 20          -> 20 / 2 - 10 = 0
+    setGemsParameters({10.0, 10.0}, {20.0, 20.0});
```
guilpier-code:
```c+
// load = 20          -> 20 / 2 - 10 = 0
setGemsParameters({10.0, 10.0}, {20.0, 20.0});
```
We could make it more simple : 
```c+
// load = 0         -> 0
setGemsParameters({10.0, 10.0}, {0., 0.});
```

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+    setLegacyBounds();
+
+    gemsPart->setHour(0);
+    gemsPart->setBoundsOnENS();
+    gemsPart->setRHSfictitiousLoadValue();
+    gemsPart->setRHSMaxEnsLoadValue();
```
guilpier-code:
```c++
 gemsPart->setBoundsOnENS();
 gemsPart->setRHSfictitiousLoadValue();
 gemsPart->setRHSMaxEnsLoadValue();
```
Every tests makes all of these calls.
Tests should have a more focused purpose, testing only one thing, and get shorter, and so more clear about what is tested.
Production code is split into pieces, each piece could be tested in isolation from the rest.

for example, this modified test works fine : 
```c++
BOOST_FIXTURE_TEST_CASE(spillage_bound_with_positive_residual_load, GemsContributionFixture)
{
    // residual_load = 10 -> 2 * 10 + 30 = 50
    // load = 0          -> 0 / 2 - 0 = 0
    setGemsParameters({10.0, 10.0}, {0., 0.});
    setLegacyBounds(); // To be moved elsewhere

    gemsPart->setHour(0);
    gemsPart->setRHSfictitiousLoadValue();

    // Fictitious load constraint (spillage bound): only area1 is connected to the GEMS component
    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[fictitiousLoadArea1], 500.0 + 50.0);
    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[fictitiousLoadArea2], 500.0);
}
```

A new name for this test could be : 
**positive_residual_load__gems_part_sets_rhs_fictitious_load__rhs_correct**

Test is much shorter and has a clear objective

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[maxEnsLoadArea1], 300.0);
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[maxEnsLoadArea2], 300.0);
+
+    // Bounds not touched by the GEMS contribution remain at their legacy value
+    BOOST_CHECK_EQUAL(problemAResoudre.Xmax[5], 0.0);
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[5], 0.0);
+}
+
+// Residual load < 0: the spillage bound decreases the RHS of the fictitious load constraint.
+BOOST_FIXTURE_TEST_CASE(spillage_bound_with_negative_residual_load, GemsContributionFixture)
```
guilpier-code: Same test as above, but with different residual load.
This test could be shorter and more focused.

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[maxEnsLoadArea1], 300.0);
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[maxEnsLoadArea2], 300.0);
+}
+
+// GEMS load > 0: the unsupplied energy bound increases the ENS variable bound and the RHS of the
+// max ENS load constraint.
+BOOST_FIXTURE_TEST_CASE(unsupplied_energy_bound_with_positive_gems_load, GemsContributionFixture)
+{
+    // load = 50           -> 50 / 2 - 10 = 15
+    // residual_load = -15 -> 2 * (-15) + 30 = 0
```
guilpier-code:
```c++
// residual_load = -15 -> 2 * (-15) + 30 = 0
```
Why not : 
```c++
// residual_load = 0 -> 2 * 0 + 0 = 0
```
?

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+    // Unsupplied energy bound: contribution is zero
+    BOOST_CHECK_EQUAL(problemAResoudre.Xmax[ensVarArea1Hour0], 100.0);
+    BOOST_CHECK_EQUAL(problemAResoudre.Xmax[ensVarArea2], 100.0);
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[maxEnsLoadArea1], 300.0);
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[maxEnsLoadArea2], 300.0);
+}
+
+// GEMS load > 0: the unsupplied energy bound increases the ENS variable bound and the RHS of the
+// max ENS load constraint.
+BOOST_FIXTURE_TEST_CASE(unsupplied_energy_bound_with_positive_gems_load, GemsContributionFixture)
```
guilpier-code: Here it seems this tests coule be split in 2 shorter tests : 
- one modifying **Xmax** of ens variables : calling **setBoundsOnENS** only
- the other  modifying rhs of ens variables : calling **setRHSMaxEnsLoadValue** only

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
```diff
@@ -0,0 +1,387 @@
...
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[maxEnsLoadArea2], 300.0);
+
+    // Fictitious load constraint (spillage bound): contribution is zero
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[fictitiousLoadArea1], 500.0);
+    BOOST_CHECK_EQUAL(problemAResoudre.SecondMembre[fictitiousLoadArea2], 500.0);
+}
+
+// GEMS load < 0: the unsupplied energy bound decreases the ENS variable bound and the RHS of the
+// max ENS load constraint.
+BOOST_FIXTURE_TEST_CASE(unsupplied_energy_bound_with_negative_gems_load, GemsContributionFixture)
```
guilpier-code: Same remarks as above : 
- make residual load profile more simple with zeros
- split this test into multiple if needed, calling one **gemsPart** method at a time

src/tests/src/solver/optimisation/adequacy_patch/test_compute_gems_contribution.cpp
guilpier-code: Maybe for another PR
---
```c++ 
auto linearProblem = std::make_shared<StructuredLinearProblem>();
problemeHebdo.optimEntityContainer = std::make_shared<OptimEntityContainer>(*linearProblem);
```
Still about the **linear problem** : 
**OptimEntityContainer** is built with a LP. But its mission is to ask question to the LP, and certainly not to change it.
So we should build a  **OptimEntityContainer** with a **const** LP.
Problem : it's impossible to do that because of : 
`ILinearProblem& OptimEntityContainer::Problem()` 
which does not return `const &`.

This is a design problem : this previous **OptimEntityContainer** method should not exist
This issue comes up often these days.
**OptimEntityContainer** and **LP** are closely related.

A possible immediate solution could be to have a new class, child of **LegacyOrtoolsLinearProblem**, that would have the methods and data members that **OptimEntityContainer** has, doing its job.
So : no more  **OptimEntityContainer**, only a special LP instead.

To be looked at.

PR #3855, opened 2026-08-13
refactor: Remove yuni from src/libs/antares/study/

payetvin:


src/libs/antares/utils/include/antares/utils/utils.h
JasonMarechal25, line 1: suggestion: add tests for those utilities

src/tests/src/libs/antares/array/tests-matrix-save.h
```diff
@@ -9,14 +9,6 @@ using namespace std;
 using namespace Yuni;
 using namespace Antares;
 
-struct TSNumbersPredicate
```
JasonMarechal25, line 12: Question: was this unused?

src/solver/application/application.cpp
```diff
@@ -550,8 +550,6 @@ Application::~Application()
             pStudy->importLogsToOutputFolder(*resultWriter);
         }
 
-        // release all reference to the current study held by this class
-        pStudy->clear();
```
JasonMarechal25, line 554: Question: why the removal of clear ?

payetvin: clear was only used to shrink and clear some yuni strings

src/tests/src/libs/antares/array/tests-matrix-save.h
payetvin: Yes

src/libs/antares/utils/include/antares/utils/utils.h
payetvin: done

JasonMarechal25 (approved):

PR #3858, opened 2026-08-17
feat(optimisation): refactor simulation table handling [ANT-5256]

JasonMarechal25:
Refactor simulation table filling. Introduce "stages".  A stage is a time at wich we want to write a simulation table. For now there are only optim-1 and optim-2 but it will enable other stages like adequacy patch or remix hydro


src/solver/optimisation/LegacySimulationTableSnapshot.cpp
```diff
@@ -0,0 +1,54 @@
...
+                               const LegacyNameMapper& nameMapper,
+                               unsigned currentBlock)
+{
+    const PROBLEME_ANTARES_A_RESOUDRE& problem = *problemeHebdo.ProblemeAResoudre;
+
+    // LegacyVariablesInfo, X and CoutLineaire are all sized to NombreDeVariables
+    // in resizeProbleme, so the index-based reads below are always in bounds.
+    assert(problem.LegacyVariablesInfo.size() == static_cast<std::size_t>(problem.NombreDeVariables)
+           && problem.X.size() == static_cast<std::size_t>(problem.NombreDeVariables)
+           && problem.CoutLineaire.size() == static_cast<std::size_t>(problem.NombreDeVariables));
```
flomnes: `problem.CoutLineaire` is not used in this function, consider removing the assert

src/io/outputs/include/antares/io/outputs/OptimisationsSimulationTable.h
```diff
@@ -3,19 +3,39 @@
...
+// writer emits one file per stage, named after it.
 class OptimisationsSimulationTable
 {
 public:
-    Antares::IO::Outputs::SimulationTable* firstOptimSimulationTable();
-    Antares::IO::Outputs::SimulationTable* secondOptimSimulationTable();
+    // Stage names of the two optimisation passes. They are part of the output
+    // file names, so they must not change.
+    static constexpr const char* firstOptimStage = "optim-nb-1";
+    static constexpr const char* secondOptimStage = "optim-nb-2";
```
flomnes: Can be `private`

src/libs/antares/writer/LegacySimulationTablesWriter.cpp
```diff
@@ -43,16 +42,18 @@ LegacySimulationTablesWriter::LegacySimulationTablesWriter(const fs::path& folde
 
 void LegacySimulationTablesWriter::write(OptimisationsSimulationTable& tables)
 {
-    writeForOptim(tables.firstOptimSimulationTable(), 1);
-    writeForOptim(tables.secondOptimSimulationTable(), 2);
+    for (const auto& [stage, table]: tables.stages())
```
flomnes: `OptimisationsSimulationTable& tables` can probably be passed as `const`

flomnes (commented):
The refactored `write()` now only writes files for stages that were actually requested during the year, instead of always writing both `optim-nb-1` and `optim-nb-2`. Since the `optim-nb-2` stage is only created when the 2nd optimization pass runs, in years where it's skipped (no thermal heuristic, expansion mode, or all weeks failing the 1st pass), the `simulation-table-<year>-optim-nb-2.*` file silently goes missing breaking the stable per-year file list downstream tools may depend on.

src/solver/optimisation/LegacySimulationTableSnapshot.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3858/commits/88f4d3523b14fa0054397ac6a0ddd3709b2ebfb5

src/io/outputs/include/antares/io/outputs/OptimisationsSimulationTable.h
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3858/commits/11b4844fa3be037c214c3cef8ec5450ed3b1cc04

src/libs/antares/writer/LegacySimulationTablesWriter.cpp
JasonMarechal25: https://github.com/AntaresSimulatorTeam/Antares_Simulator/pull/3858/commits/4a41303459e22b351a6a2481eb897fbe54d3e4a3

flomnes (approved):

PR #3867, opened 2026-08-19
feat(simulation): add support for selecting simulation table stages [ANT-5234]

JasonMarechal25:


flomnes (approved):

docs/user-guide/solver/10-command-line.md
```diff
@@ -40,6 +40,7 @@ hide:
 | --derated                | Force the [derated](04-parameters.md#derated) mode                                                |
 | -z, --zip-output         | Write the results into a single zip archive                                                       |
 | --parquet                | Write simulation table in format parquet instead of csv format                                    |
+| --simulation-table-stages=VALUE | Override [which resolution stages get a simulation table](04-parameters.md#simulation-table-stages): `all`, or a comma-separated list among `optim-nb-1`, `optim-nb-2`, `remix-hydro`, `adq-patch-csr`. Default: `all` |
```
flomnes: Has no effect unless `--output all` or `--output simulation-tables` is provided

src/io/outputs/OptimisationsSimulationTable.cpp
```diff
@@ -3,8 +3,70 @@
...
+        {
+            everyStage = true;
+            continue;
+        }
+
+        const auto& known = allStages();
+        if (std::find(known.begin(), known.end(), name) == known.end())
+        {
+            std::ostringstream message;
+            message << "Invalid value for " << source << ": '" << name << "' (expected all";
```
flomnes:
```suggestion
            message << "Invalid value for " << source << ": '" << name << "' (expected all, ";
```

src/io/outputs/OptimisationsSimulationTable.cpp
```diff
@@ -3,8 +3,70 @@
...
+    std::string name;
+    while (std::getline(stream, name, ','))
+    {
+        // Tolerate the spaces a user naturally writes after a comma.
+        const auto first = name.find_first_not_of(" \t");
+        if (first == std::string::npos)
+        {
+            continue;
+        }
+        name = name.substr(first, name.find_last_not_of(" \t") - first + 1);
```
flomnes: 1. Split the string by `,`, ignoring \t, " ", etc.
2. Apply business logic "all", `everyStage`

I would have 2 functions here
`std::string -> std::vector<std::string> -> std::vector<enum>`

Consider using `enum` for stages instead of `const char*`

JasonMarechal25: Indeed cleaner with enum

flomnes (approved):

PR #3889, opened 2026-08-24
Fix: Correct units in districts [ANT-5626]

payetvin:


JasonMarechal25 (approved):
Code doesn't say much, I'll trust tests

PR #3899, opened 2026-08-26
Add check for connections [ANT-4385]

payetvin:


src/io/inputs/yml-system/converter.cpp
```diff
@@ -122,6 +123,27 @@ void CheckPortsType(const Port& firstPort, const Port& secondPort)
...
+    {
+        const auto portFieldRole_1 = port_1.fieldRole(field.Id());
+        const auto portFieldRole_2 = port_2.fieldRole(field.Id());
+
+        if (portFieldRole_1 == portFieldRole_2)
+        {
+            std::ostringstream msg;
+            msg << "In connection between components '" << componentId_1 << "' and '"
+                << componentId_2 << "': Field '" << field.Id() << "' is " << portFieldRole_1
+                << " in both ports '" << port_1.Id() << "' and '" << port_2.Id() << "'";
```
flomnes: Use `fmt::format`

flomnes (changes requested):
Add at least one unit test

flomnes (approved):

PR #3909, opened 2026-08-31
refactor(reserves): replace std::optional .value() chains with -> operator

flomnes:
## Motivation

`std::optional::operator->` is the idiomatic way to access the contained value. Chained `obj.value().member` calls are more verbose and obscure that the object is an optional.

## Method

Mechanical replacement of `obj.value().member` / `obj.value().method()` with `obj->member` / `obj->method()` for `std::optional` objects — 264 occurrences in 47 files across the study library, solver (simulation, variable, optimisation) and tests. Covers same-line, end-of-line and multi-line expression layouts.

`absl::StatusOr` usages in `ortools_quadratic_wrapper.cpp` are intentionally untouched (no `operator->`).

Formatted with the project clang-format (18.1.3). No behavior change.

## Verification

- Full build: `cmake --build build-debug` — success
- Full test suite: `ctest` — 90/90 passed


guilpier-code:
Searching for `value().` in my IDE, I get occurrences in **test_study_reserves_load.cpp**.
Looking at them more closely, it's about optionals

guilpier-code:
I can't see the point in defining : 
```c++
template<class T>
using ReserveOpt = std::optional<T>;
```
is it that useful ? It seems to me it's not.

src/solver/simulation/sim_alloc_probleme_hebdo.cpp
guilpier-code: I can see that there are many : `.value()[some index]` in the base code.
For example here, instead of : 
```c++
auto& res = resultats.Reserves.value()[j];
```
couldn't we have : 
```c++
auto& res = resultats.Reserves->at(j);
```
?

_One subtlety: v->at(index) checks both that the optional contains a vector and that index is valid. If the optional is empty, it throws std::bad_optional_access; if the index is invalid, at() throws std::out_of_range._

flomnes: You're right, the `value()` can be removed as well

flomnes:
> I can't see the point in defining :
> 
> ```c++
> template<class T>
> using ReserveOpt = std::optional<T>;
> ```
> 
> is it that useful ? It seems to me it's not.

Not really, previously it was a real custom type, but now it's only a typedef. It can be removed safely

guilpier-code (approved):

PR #3921, opened 2026-09-07
refactor: avoid leaking file context in ILoader

JasonMarechal25:
ILoader is meant to be an abstraction of loader. By addig a "studyPath" method to it we tie ILoader to file system domain.
Even if we exclusively use file system for inputs, it is still good to keep this abstraction. At least to prevent file system dependency in unit tests down the road.

- Remove studyPath() from ILoader
- Encapsulate paths used by FileLoader in a dedicated data structure to prevent constructor to explode in arg number.
- Some minor refactoring, clening up namespaces, anonymous namespace, nodiscard, constness...


flomnes (approved):
`studyPath` is now dead code, consider removing it

PR #3922, opened 2026-09-07
Make OptimEntityContainer own the linear problem lifetime

flomnes:
## PR Title

**Share `ILinearProblem` ownership with `OptimEntityContainer` to support post-solve GEMS evaluation**

### Summary

Refactors `ILinearProblem` ownership from `std::unique_ptr` to `std::shared_ptr` so that `OptimEntityContainer` co-owns the linear problem it references, and removes the redundant problem pointers that had to keep the problem alive elsewhere.

### Motivation

Hourly adequacy-patch GEMS contribution evaluations run **after** the weekly simplex solve and read variable solution values through the `OptimEntityContainer`. Previously the container only held a reference to the problem, so the problem had to be kept alive by out-of-band owners (`PROBLEME_HEBDO::ortoolsProblem`, `SimplexResult::originalProblem`, and the `Modeler`'s `unique_ptr` members). That was fragile: any consumer relying on the container could see a dead problem. The container now owns the problem's lifetime for all post-solve consumers.

### Changes

- **`OptimEntityContainer`**: constructor now takes a `std::shared_ptr<ILinearProblem>`; the container keeps the problem alive.
- **`Modeler`** (`Modeler.h` / modeler problem building): `masterProblem_`, `subproblems_`, and `ProblemEntity.problem` changed from `unique_ptr` to `shared_ptr`. `buildProblem` shares the freshly built problem with the container it creates before returning.
- **`PROBLEME_HEBDO`**: dropped the now-redundant `ortoolsProblem` member — the weekly problem is only reachable via `optimEntityContainer`.
- **`OPT_TryToCallSimplex`**: removed `SimplexResult::originalProblem`. The infeasible-problem path builds its problem as a `shared_ptr` and retrieves the problem for MPS export from `optimEntityContainer->Problem()`.
- **Adequacy patch CSR** (`set_variable_boundaries.cpp`): added a debug log line printing the GEMS contribution per area (supporting work for the GEMS-in-adequacy-patch feature).

### Tests

Mechanical ownership updates across fixtures that previously held the problem by value or `unique_ptr`; they now use `std::shared_ptr` and pass it to `OptimEntityContainer` (direct `OptimEntityContainer(pb)` construction instead of `OptimEntityContainer(*pb)`):

- Expression evaluation tests (`test_PrintAndEvalVisitors`, `test_evaluate-function-operators`, `test_evaluate-sum-operators`)
- Simulation table tests (`testSimulationTable`)
- Modeler tests (`UtilMocks`, `inmemory-modeler`, `testLinearProblemBuilder`, `test-component-filler-to-master-pb`, `visitorFixture`)
- Optimisation filler tests (`FillerFixture`, `test_ThermalCapacityFiller`, `test_area_connections--spillage--unsupE`, `test_componentToAreaConnectionFiller`, `test_compute_gems_contribution`)

### DIFF
--- a/src/modeler/modeler/include/antares/solver/modeler/Modeler.h
+++ b/src/modeler/modeler/include/antares/solver/modeler/Modeler.h
@@ -3,6 +3,7 @@
 
 #pragma once
 #include <filesystem>
+#include <memory>
 
 #include <antares/optimisation/linear-problem-api/linearProblem.h>
 #include "antares/io/outputs/SimulationTable.h"
@@ -35,7 +36,8 @@ class ILoader;
 
 struct ProblemEntity
 {
-    std::unique_ptr<LinearProblem::Api::ILinearProblem> problem;
+    // Shared with the container, which references the problem's entities.
+    std::shared_ptr<LinearProblem::Api::ILinearProblem> problem;
     std::unique_ptr<LinearProblem::OptimEntityContainer> optimEntityContainer;
 };
 
@@ -73,13 +75,13 @@ public:
 
     ILoader& loader_; // gp : make it private
 
-    [[nodiscard]] const std::unique_ptr<LinearProblem::Api::ILinearProblem>& masterProblem() const
+    [[nodiscard]] const std::shared_ptr<LinearProblem::Api::ILinearProblem>& masterProblem() const
     {
         return masterProblem_;
     }
 
-    [[nodiscard]] const std::vector<std::unique_ptr<LinearProblem::Api::ILinearProblem>>&
-    subproblems() const
+    [[nodiscard]]
+    const std::vector<std::shared_ptr<LinearProblem::Api::ILinearProblem>>& subproblems() const
     {
         return subproblems_;
     }
@@ -95,8 +97,9 @@ private:
       const LinearProblem::OptimEntityContainer& subproblemOptimEntityContainer,
       const LinearProblem::Api::FillContext& timeScenarioCtx) const;
 
-    std::unique_ptr<LinearProblem::Api::ILinearProblem> masterProblem_ = nullptr;
-    std::vector<std::unique_ptr<LinearProblem::Api::ILinearProblem>> subproblems_;
+    // Shared with the containers referencing them.
+    std::shared_ptr<LinearProblem::Api::ILinearProblem> masterProblem_ = nullptr;
+    std::vector<std::shared_ptr<LinearProblem::Api::ILinearProblem>> subproblems_;
     std::unique_ptr<LinearProblem::OptimEntityContainer> subproblemOptimEntityContainer_ = nullptr;
     std::unique_ptr<LinearProblem::Api::FillContext> timeScenarioCtx_ = nullptr;
     LinearProblem::Api::IMipSolution* subProbSolution_ = nullptr;
--- a/src/modeler/modeler/include/antares/solver/modeler/Modeler.h
+++ b/src/modeler/modeler/include/antares/solver/modeler/Modeler.h
@@ -34,20 +34,22 @@ namespace Antares::Solver
 {
 class ILoader;
 
-struct ProblemEntity
-{
-    // Shared with the container, which references the problem's entities.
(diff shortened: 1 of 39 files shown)


src/modeler/modeler/include/antares/solver/modeler/Modeler.h
```diff
@@ -35,7 +36,8 @@ class ILoader;
 
 struct ProblemEntity
```
guilpier-code: It seems that, in `struct ProblemEntity`, we only use the public data **problem**, and not the data **optimEntityContainer**.
To be accurate, there is one place where we use the **optim entity container** : 
```c++
void Modeler::buildSubProblem()
{
    auto [subproblem, subproblemOptimEntityContainer] = buildProblem(
                                                                     ...
                                                                     );
    subproblems_.emplace_back(std::move(subproblem));
    subproblemOptimEntityContainer_ = std::move(subproblemOptimEntityContainer);
}
```
But this usage could easily be removed, to finally have : 
```c++
void Modeler::buildSubProblem()
{
    subproblems_ = buildProblem(
                               ...
                               );
}
```

If I'm correct, we can remove this class.

src/modeler/modeler/include/antares/solver/modeler/Modeler.h
```diff
@@ -95,8 +97,9 @@ class Modeler final
       const LinearProblem::OptimEntityContainer& subproblemOptimEntityContainer,
       const LinearProblem::Api::FillContext& timeScenarioCtx) const;
 
-    std::unique_ptr<LinearProblem::Api::ILinearProblem> masterProblem_ = nullptr;
```
guilpier-code, line 98: Do we need the unique / shared pointer to be initialized as **nullptr** ?
Is it not **nullptr** by default ?

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
```diff
@@ -771,7 +771,7 @@ struct TimeDependentParameterFixture
     std::string compoName = "1245";
     std::vector<Component> components;
     Antares::LinearProblem::ScenarioGroupRepository scenarioGroupRepo;
-    MockLinearProblem linearProblem = MockLinearProblem(true);
+    std::shared_ptr<MockLinearProblem> linearProblem = std::make_shared<MockLinearProblem>(true);
```
guilpier-code, line 774:
```c++
std::shared_ptr<MockLinearProblem> linearProblem = std::make_shared<MockLinearProblem>(true);
```
or : 
```c++
auto linearProblem = std::make_shared<MockLinearProblem>(true);
```

src/tests/src/modeler/UtilMocks.h
```diff
@@ -461,7 +461,7 @@ struct MyDummyFixture: Antares::Expressions::Registry<Antares::Expressions::Node
     Antares::LinearProblem::ScenarioGroupRepository scenarioGroupRepository = makeScenarioGroupRepo(
       components.front());
 
-    MockLinearProblem linearProblem = MockLinearProblem(true);
+    std::shared_ptr<MockLinearProblem> linearProblem = std::make_shared<MockLinearProblem>(true);
```
guilpier-code, line 464: use of `auto` ?

flomnes: not allowed for class members as of C++20

src/tests/src/expressions/test_PrintAndEvalVisitors.cpp
flomnes: Not allowed in class member declaration

src/modeler/modeler/include/antares/solver/modeler/Modeler.h
flomnes: done

src/modeler/modeler/include/antares/solver/modeler/Modeler.h
flomnes: done

PR #3926, opened 2026-09-07
Add filtering for district output data

flomnes:
### Context

Districts (sets of areas) always export **all five time granularities** (`values-hourly/daily/weekly/monthly/annual.txt`), both for the synthesis report (`mc-all`) and the year-by-year reports (`mc-ind`). On large studies, this is often wasteful: a user may only need annual synthesis results for an "all system" district, while keeping the full granularity for individual areas.

This PR adds two optional keys, `filter-synthesis` and `filter-year-by-year`, to the existing `INPUT/areas/sets.ini` file to restrict the exported granularities **per district and per report type**. They reuse the exact same syntax and semantics as the `filter-synthesis` / `filter-year-by-year` keys already available for areas, links and binding constraints.

> **Note:** this replaces the earlier `sets-outputs.yaml` approach (which required a separate optional file and a `yaml-cpp` dependency). The filtering now lives entirely inside `sets.ini`, and the YAML file has been fully removed.

### Usage

```ini
; INPUT/areas/sets.ini
[all system]
apply-filter = add-all
output = true
filter-synthesis = annual
filter-year-by-year = hourly, daily
```

Each district (INI section) may define the two optional keys. `filter-synthesis` applies to the synthesis report (`mc-all`) and `filter-year-by-year` to the year-by-year reports (`mc-ind`). Each value is a list (separated by commas and/or spaces) of granularities: `hourly`, `daily`, `weekly`, `monthly`, `annual`.

| Input | Behavior |
|---|---|
| Key missing (or district not defined) | default behavior: all granularities exported |
| `filter-synthesis = annual` | only `values-annual.txt` for that district in `mc-all` |
| `filter-year-by-year = hourly, daily` | only `values-hourly.txt` and `values-daily.txt` in `mc-ind` |
| key with no valid token (e.g. `filter-synthesis = none`) | the district output is skipped entirely for that report |
| unknown token (e.g. `bogus`) | ignored |

### Implementation

The feature deliberately reuses the existing area/link filtering machinery so that there is a single source of truth:

- **`study/sets.{h,cpp}`** — `Sets::Options` carries `filterSynthesis` / `filterYearByYear` (`unsigned int` bitmasks of `Data::FilterFlag`, default `filterAll`), mirroring `Area::filterSynthesis` and `AreaLink::filterSynthesis`. The two new keys are parsed inside `Sets::loadFromFile()` with the existing **`stringIntoDatePrecision()`** (the same function as the `filter-synthesis` / `filter-year-by-year` keys of areas and links). A `Sets::outputFilter(id, synthesis)` accessor returns the effective mask. The old `Sets::loadOutputPrecisionsFromFile()` and its `yaml-cpp` include are removed.
- **`solver/variable/.../surveyresults/`** — `SurveyResultsData` carries a `setOfAreasPrecisionFilter` (default `filterAll`). In the report builder, `RunForEachSetOfAreas` skips the district directory when the filter is empty (same `skipDirectory` behavior as areas) and sets the mask for the duration of the run; the per-granularity guards in `RunGlobalResults` / `RunAnnual` use the `Data::filter*` constants, in the same style as `area.hxx` / `links.hxx`. Areas, links and binding constraints are untouched (their default mask is `filterAll`).
- **`study/load.cpp`** — no change to the loading entry point is needed beyond the removed `sets-outputs.yaml` call: the filters are read together with the rest of `sets.ini`.
- **Logging** — the solver log summarizes each applied (non-default) filter, e.g.:
  ```
  sets: district `west and east`: mc-all granularities = annual
  sets: district `west and east`: mc-ind granularities = hourly, daily
  ```

### Tests

**Unit** (`test-setofareas-reports`, all passing): the districts are now driven through a real `sets.ini` (written to a temp file and loaded via `Sets::loadFromFile`) instead of the YAML file:
- loader: `filter-synthesis` / `filter-year-by-year` parsed per report type, unknown tokens ignored, `none` → empty filter, valid INI list → corresponding mask, district without the keys → `filterAll`
- export (synthesis): filtered district gets only the selected `values-*.txt`, an unfiltered district still gets all five
- export (synthesis): `none` → no district file at all
- export (year-by-year): filtered district gets only the selected `values-*.txt`

**End-to-end**: ran `antares-solver` on a `short-tests` study (`021 Four areas - DC law`, economy mode, 1 MC year) with two districts — an unfiltered `all areas` and a filtered `west and east` (`filter-synthesis = annual`, `filter-year-by-year = hourly, daily`). The unfiltered district exported all five granularities for both reports, while the filtered district exported only `annual` in `mc-all` and only `hourly` + `daily` in `mc-ind`; per-area outputs remained unfiltered. The solver log reported the applied filters and no "Invalid property" warnings were emitted.

### Documentation

`docs/user-guide/solver/02-inputs.md`: the `sets.ini` section now documents the two new keys (in the PRINCIPLE list and with an example), and the previous `sets-outputs.yaml` section has been removed.

### Notes / non-goals

- Output **variables** (columns) per district are intentionally *not* part of this PR; column selection remains globally handled by the variables inspector.
- Granularity filtering for areas/links already exists via their INI keys — this PR only extends the same concept to districts, using the same keys.
- The `sets-outputs.yaml` file (and the `yaml-cpp` dependency it required) is fully removed, with no backward-compatibility path.

### DIFF
--- a/src/libs/antares/study/area/sets.cpp
+++ b/src/libs/antares/study/area/sets.cpp
@@ -4,7 +4,6 @@
 #include "antares/study/sets.h"
 
 #include <string>
-
 #include <yaml-cpp/yaml.h>
 
 #include <antares/utils/utils.h>
@@ -253,7 +252,8 @@ bool Sets::loadOutputPrecisionsFromFile(const std::filesystem::path& filename)
 
     if (!root || !root.IsMap())
     {
-        logs.warning() << "sets: `" << filename << "`: the root node must be a mapping of "
+        logs.warning() << "sets: `" << filename
+                       << "`: the root node must be a mapping of "
                           "districts";
         return true;
     }
@@ -291,8 +291,9 @@ bool Sets::loadOutputPrecisionsFromFile(const std::filesystem::path& filename)
             }
             if (!node.IsSequence())
             {
-                logs.warning() << "sets: `" << filename << "`: district `" << districtKey
-                               << "`, `" << reportType << "` must be a sequence of "
+                logs.warning() << "sets: `" << filename << "`: district `" << districtKey << "`, `"
+                               << reportType
+                               << "` must be a sequence of "
                                   "granularities, ignored";
                 continue;
             }
@@ -326,19 +327,17 @@ bool Sets::loadOutputPrecisionsFromFile(const std::filesystem::path& filename)
     {
         if (pair.second.filterSynthesis != filterAll)
         {
-            logs.info() << "sets: district `" << pair.first
-                        << "`: mc-all granularities = "
+            logs.info() << "sets: district `" << pair.first << "`: mc-all granularities = "
                         << (pair.second.filterSynthesis == filterNone
-                                ? std::string("none")
-                                : datePrecisionIntoString(pair.second.filterSynthesis));
+                              ? std::string("none")
+                              : datePrecisionIntoString(pair.second.filterSynthesis));
         }
         if (pair.second.filterYearByYear != filterAll)
         {
-            logs.info() << "sets: district `" << pair.first
-                        << "`: mc-ind granularities = "
+            logs.info() << "sets: district `" << pair.first << "`: mc-ind granularities = "
                         << (pair.second.filterYearByYear == filterNone
-                                ? std::string("none")
-                                : datePrecisionIntoString(pair.second.filterYearByYear));
+                              ? std::string("none")
+                              : datePrecisionIntoString(pair.second.filterYearByYear));
         }
     }
     return true;
--- a/src/libs/antares/study/area/sets.cpp
+++ b/src/libs/antares/study/area/sets.cpp
@@ -5,6 +5,8 @@
 
 #include <string>
 
+#include <yaml-cpp/yaml.h>
+
 #include <antares/utils/utils.h>
 
 namespace Antares::Data
@@ -21,6 +23,7 @@ std::string trim(const std::string& s)
     const auto end = s.find_last_not_of(" \t");
     return s.substr(begin, end - begin + 1);
 }
+
 } // namespace
 
 Sets::Sets(const Sets& rhs):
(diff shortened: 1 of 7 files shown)


src/libs/antares/study/area/sets.cpp
```diff
@@ -227,6 +255,16 @@ void Sets::rebuildAllFromRules(SetHandlerAreas& handler)
     }
 }
 
+unsigned int Sets::outputFilter(const IDType& id, bool synthesis) const
```
JasonMarechal25: suggestion: use enum instead of bool for synthesis/yby filter.

Synthesis and YearByYear are not couple by a on/off sementics.

src/solver/variable/include/antares/solver/variable/surveyresults/reportbuilder.hxx
```diff
@@ -440,7 +470,18 @@ private:
...
+            // district directory (same behavior as for areas).
+            const unsigned int filter = sets.outputFilter(setName, GlobalT);
+            if (filter == Data::filterNone)
+            {
+                continue;
+            }
+            results.data.setOfAreasPrecisionFilter = filter;
             SurveyReportBuilderFile<GlobalT, NextT, CDataLevel>::Run(list, results, numSpace);
+            // Reset the filter: it must not leak to the other data levels
+            results.data.setOfAreasPrecisionFilter = Data::filterAll;
```
JasonMarechal25: Question: reset to filterAll or to the previous value? I don't know if it can be different from All at this point

JasonMarechal25 (approved):
I've tested with TP2_Viz, all good

src/solver/variable/include/antares/solver/variable/surveyresults/reportbuilder.hxx
flomnes: I think it's not necessary to set `results.data.setOfAreasPrecisionFilter` after calling `SurveyReportBuilderFile<GlobalT, NextT, CDataLevel>::Run`

Think of it as a function argument (and it should be !)

src/libs/antares/study/area/sets.cpp
flomnes: Done in b6cf39e4cdc5a53876ed56d8644e8780d670b1c0: `Sets::outputFilter` now takes an `enum class ReportType { synthesis, yearByYear }` instead of a bool, so the parameter selects one of the two report types rather than being an on/off flag.

src/solver/variable/include/antares/solver/variable/surveyresults/reportbuilder.hxx
flomnes: Done in e390cc7af862228166b89c561bc24ead66193ca9: the filter is now passed as an argument to `SurveyReportBuilderFile::Run` (threaded through `RunGlobalResults` / `RunAnnual` and the recursive next-level call, with a `filterAll` default so areas/links/thermal clusters are untouched). The `SurveyResultsData::setOfAreasPrecisionFilter` field and the reset are both gone, so there is no state to reset.

PR #3930, opened 2026-09-08
refactor(optimisation): avoid modifying `problem_.X` and `problem_.CoutsMarginauxDesContraintes`, use copies

flomnes:


flomnes:
Identical work was performed independently by @JasonMarechal25

PR #3934, opened 2026-09-10
fix: keep IScenario as pure as possible

JasonMarechal25:


src/modeler-optimisation-container/scenarioGroupRepo.cpp
```diff
@@ -28,7 +30,13 @@ class DefaultScenario final: public LinearProblem::Api::IScenario
     {
         return 1; // Default rank for empty groupId
     }
+
+    [[nodiscard]] bool hasYear(Year) const override
+    {
+        return true;
+    };
```
flomnes: stray semicolon
```suggestion
    }
```

flomnes (approved):

